PR #239, opened 2026-03-19
fix: use bash shell for cargo build steps on Windows

NicholaiVogel:
## Summary

- Windows runners default to PowerShell, which treats `$RELEASE_TARGET` as a PowerShell variable (empty) instead of the env var
- Cargo gets `--target ""` and fails with `error: target was empty`
- All 4 Windows build legs (predictor x64/arm64, daemon x64/arm64) fail every release
- Fix: add `shell: bash` to both cargo build steps, matching the upload steps which already use it

## Test plan

- [ ] Next release produces Windows binaries on the GitHub release page
- [ ] All 10 matrix legs pass (linux x64, darwin x64/arm64, win32 x64/arm64 for both predictor and daemon)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -183,6 +183,7 @@ jobs:
           if ($LASTEXITCODE -notin @(0, 3010)) { exit $LASTEXITCODE }
 
       - name: Build predictor
+        shell: bash
         run: cargo build --release --target "$RELEASE_TARGET"
         working-directory: packages/predictor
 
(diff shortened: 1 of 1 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: use bash shell for cargo build steps on Windows` (commit `a6805bb3`) now and I'll follow up shortly with feedback.

PR #241, opened 2026-03-19
fix: prevent CMD window flashing and fix workspace path matching on Windows

aaf2tbz:
## Summary
- Add `windowsHide: true` to all subprocess spawns (`daemon.ts`, `predictor-client.ts`) to prevent console windows from flashing on Windows
- Rewrite `spawnHidden` in `provider.ts` to use `node:child_process` with `windowsHide` on Windows, keeping `Bun.spawn` on other platforms
- Normalize backslashes to forward slashes in workspace scope matching for both TypeScript (`marketplace.ts`) and Rust (`marketplace.rs`) routes

## Test plan
- [ ] Verify no CMD window flash when daemon spawns subprocesses on Windows
- [ ] Verify predictor subprocess starts without visible window
- [ ] Verify pipeline LLM provider calls (claude, codex) don't flash windows
- [ ] Verify marketplace workspace scope matching works with Windows-style backslash paths
- [ ] Verify no regression on Linux/macOS (Bun.spawn path unchanged)

### DIFF
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -133,14 +133,18 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 		const toWebStream = (nodeStream: import("node:stream").Readable): ReadableStream<Uint8Array> =>
 			Readable.toWeb(nodeStream) as ReadableStream<Uint8Array>;
 
-		const exitPromise = new Promise<number>((resolve) => {
+		const exitPromise = new Promise<number>((resolve, reject) => {
 			child.on("exit", (code) => resolve(code ?? 1));
-			child.on("error", () => resolve(1));
+			child.on("error", (err) => reject(err));
 		});
 
+		if (!child.stdout || !child.stderr) {
+			throw new Error("spawnHidden(win32): stdout/stderr unexpectedly null");
+		}
+
 		return {
-			stdout: toWebStream(child.stdout!),
-			stderr: toWebStream(child.stderr!),
+			stdout: toWebStream(child.stdout),
+			stderr: toWebStream(child.stderr),
 			exited: exitPromise,
 			kill(signal?: string) {
 				if (signal === "SIGKILL") {


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: prevent CMD window flashing and fix workspace path matching on Windows` (commit `46006185`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +119,47 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+			env: options?.env ? sanitizedEnv : undefined,
+			windowsHide: true,
+		});
+
+		// Convert Node.js Readable streams to Web ReadableStreams
+		const toWebStream = (nodeStream: import("node:stream").Readable): ReadableStream<Uint8Array> =>
+			Readable.toWeb(nodeStream) as ReadableStream<Uint8Array>;
+
+		const exitPromise = new Promise<number>((resolve) => {
+			child.on("exit", (code) => resolve(code ?? 1));
```
NicholaiVogel, line 137: **Behavioral asymmetry: errors silently become exit code 1.**

```typescript
child.on("error", () => resolve(1));
```

If `nodeSpawn` emits an `error` event (e.g. `ENOENT` at the OS level — possible on Windows when a `.cmd` wrapper exists on PATH but the underlying binary is missing), this resolves the exit promise with `1` instead of rejecting. The Bun path's `proc.exited` presumably rejects in the same scenario. The callers (`callClaude`, `callCodex`) catch exit-code errors with `throw new Error("claude-code exit 1: ...")`, so it degrades gracefully, but the stderr message will be empty and the diagnosis will be harder. Consider:

```typescript
child.on("error", (err) => {
  reject(err);  // or at minimum log the error before resolve(1)
});
```

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +119,47 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+			child.on("error", () => resolve(1));
+		});
+
+		return {
```
NicholaiVogel, line 145: **Non-null assertions on `child.stdout` / `child.stderr`.**

`child.stdout!` and `child.stderr!` are safe *now* because `stdio: ["ignore", "pipe", "pipe"]` guarantees both streams. However, unlike the Bun path which has an explicit runtime guard (`if (!proc.stdout || !proc.stderr) throw ...`), this path silently assumes. If `stdio` options ever change or if a future refactor splits this helper, the assertions will produce a runtime `TypeError` that bypasses the error handling. Recommend adding the same guard:

```typescript
if (!child.stdout || !child.stderr) {
  throw new Error("spawnHidden(win32): stdout/stderr unexpectedly null");
}
```

packages/daemon-rs/crates/signet-daemon/src/routes/marketplace.rs
```diff
@@ -152,11 +152,13 @@ fn scope_matches(scope: &McpScope, ctx: &ContextQuery) -> bool {
             .workspace
             .as_ref()
             .map(|w| {
+                let wn = w.replace('\\', "/");
                 scope.workspaces.iter().any(|s| {
-                    if s.ends_with('*') {
-                        w.starts_with(s.trim_end_matches('*'))
+                    let sn = s.replace('\\', "/");
```
NicholaiVogel, line 157: **Pre-existing wildcard asymmetry between Rust and TypeScript (not introduced here, but worth tracking).**

The Rust wildcard check `wn.starts_with(sn.trim_end_matches('*'))` will match `"C:/Users/foobar"` against the pattern `"C:/Users/foo*"` (prefix match without requiring a path separator). The TypeScript `dimensionMatches` adds a `/` boundary: `currentLower.startsWith("${wildcard}/")`, so it does *not* match `foobar`. The two implementations will behave differently for patterns like `"path/prefix*"` (no slash before `*`). Not a bug introduced by this PR, but the normalization added here may surface this discrepancy more on Windows where paths are more likely to be passed unnormalized.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: prevent CMD window flashing and fix workspace path matching on Windows` (commit `fac09712`); taking another look at the fixes and I will report back shortly.

PR #242, opened 2026-03-19
fix: add missing --project option to pre-compaction hook

NicholaiVogel:
## Summary

- The Claude Code connector generates `--project "$(pwd)"` for all three hook commands, but the CLI's `pre-compaction` subcommand never declared a `--project` option
- Commander throws `error: unknown option '--project'` on every compaction, silently breaking the entire pre-compaction hook
- This means pre-compaction checkpoints (`consumeState()`) never fire, degrading session continuity across compaction boundaries

## Test plan

- [ ] `bun run build` passes
- [ ] `bun run typecheck` passes (pre-existing opencode-plugin error unrelated)
- [ ] `signet hook pre-compaction -H claude-code --project /tmp` no longer errors on the flag
- [ ] Trigger a compaction in Claude Code — hook should succeed and checkpoint should be written

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -6098,7 +6098,7 @@ hookCmd
 	.command("pre-compaction")
 	.description("Get summary instructions before session compaction")
 	.requiredOption("-H, --harness <harness>", "Harness name")
-	.option("--project <project>", "Project path")
+	.option("--project <project>", "Project path") // accepted for connector parity; project flows via continuity state from session-start
 	.option("--message-count <count>", "Number of messages in session", Number.parseInt)
 	.option("--json", "Output as JSON")
 	.action(async (options) => {


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: add missing --project option to pre-compaction hook` (commit `224b2dcb`) now and I'll follow up shortly with feedback.

packages/cli/src/cli.ts
```diff
@@ -6098,6 +6098,7 @@ hookCmd
 	.command("pre-compaction")
 	.description("Get summary instructions before session compaction")
 	.requiredOption("-H, --harness <harness>", "Harness name")
+	.option("--project <project>", "Project path")
```
NicholaiVogel: The option declaration is correct and stops Commander from rejecting the flag. However, the diff does not show the `.action` handler body for `pre-compaction`. Please confirm that `options.project` is actually consumed in the handler (e.g., passed to `consumeState()` or whatever project-scoping logic the other hook commands use). If the handler ignores it, the crash is fixed but project-aware checkpointing may still be a no-op for this hook. A follow-up task or a brief comment in the handler noting that `options.project` is intentionally unused (if that is the case) would prevent future confusion.

NicholaiVogel: Good catch — options.project is intentionally not forwarded to the daemon. The project for pre-compaction checkpoints flows through continuity state initialized during session-start (via initContinuity() at hooks.ts:955). The PreCompactionRequest interface deliberately omits project — both in TS and the Rust shadow daemon.

The option is declared purely so Commander doesn't reject the flag that the connector passes. Added an inline comment in 8f622dd explaining this data flow.

PR #243, opened 2026-03-19
fix: MCP stdio server process leak on session end

NicholaiVogel:
## Summary

- **signet-mcp processes were never exiting** after Claude Code sessions ended, accumulating as orphans (~125MB RAM each)
- Root cause: shutdown handler in `mcp-stdio.ts` cleared the refresh timer but never called `server.close()` or `process.exit()`
- Added graceful server shutdown, process exit, and a `stdin` end listener as a fallback for ungraceful parent death

## Test plan

- [ ] Start a Claude Code session with signet MCP enabled
- [ ] Verify `signet-mcp` process is running (`ps aux | grep signet-mcp`)
- [ ] Exit the Claude Code session
- [ ] Verify the `signet-mcp` process is gone
- [ ] Repeat several times to confirm no orphan accumulation

### DIFF
--- a/packages/daemon/src/mcp-stdio.ts
+++ b/packages/daemon/src/mcp-stdio.ts
@@ -35,8 +35,14 @@ const refreshTimer = setInterval(() => {
 	void refreshMarketplaceProxyTools(server, { notify: true });
 }, refreshMs);
 
+let closing = false;
 const shutdown = () => {
+	if (closing) return;
+	closing = true;
 	clearInterval(refreshTimer);
+	// Hard deadline: exit even if server.close() hangs
+	const deadline = setTimeout(() => process.exit(0), 3000);
+	deadline.unref();
 	void server.close().finally(() => process.exit(0));
 };
 


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: MCP stdio server process leak on session end` (commit `f26bd2a6`) now and I'll follow up shortly with feedback.

packages/daemon/src/mcp-stdio.ts
```diff
@@ -37,7 +37,11 @@ const refreshTimer = setInterval(() => {
 
 const shutdown = () => {
 	clearInterval(refreshTimer);
+	void server.close().finally(() => process.exit(0));
 };
```
NicholaiVogel, line 47: **Double-invocation risk**: `shutdown` has no re-entrancy guard. During normal session teardown the parent may close stdin *and* send SIGTERM in quick succession, causing both `process.stdin.on('end', shutdown)` and `process.on('SIGTERM', shutdown)` to fire. `server.close()` will be called twice on the same server object; whether the MCP SDK tolerates that is not guaranteed and any rejection is silently discarded via `void`. Add a one-shot guard:
```ts
let shuttingDown = false;
const shutdown = () => {
  if (shuttingDown) return;
  shuttingDown = true;
  clearInterval(refreshTimer);
  void server.close().finally(() => process.exit(0));
};
```

packages/daemon/src/mcp-stdio.ts
```diff
@@ -37,7 +37,11 @@ const refreshTimer = setInterval(() => {
...
```
NicholaiVogel, line 48: **Hung `server.close()` recreates the leak**: If `server.close()` never settles (e.g. in-flight request keeps the transport open) the `finally` callback never runs and the process stays alive — the same orphan symptom the PR is fixing, just via a different path. Add an unref'd escape hatch so the process always exits within a bounded time:
```ts
void server.close().finally(() => process.exit(0));
setTimeout(() => process.exit(1), 5_000).unref();
```
The `.unref()` means the timer won't keep the event-loop alive on a clean fast shutdown.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix: MCP stdio server process leak on session end` (commit `03bf3641`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/mcp-stdio.ts
```diff
@@ -35,9 +35,19 @@ const refreshTimer = setInterval(() => {
 	void refreshMarketplaceProxyTools(server, { notify: true });
 }, refreshMs);
 
+let closing = false;
 const shutdown = () => {
+	if (closing) return;
+	closing = true;
 	clearInterval(refreshTimer);
```
NicholaiVogel, line 42: Previously unresolved: `server.close()` hang left the process alive. Fixed — `deadline` fires after 3 s regardless. `.unref()` is the right call here so the timer never keeps an otherwise-idle process alive past a clean exit.

packages/daemon/src/mcp-stdio.ts
```diff
@@ -35,9 +35,19 @@ const refreshTimer = setInterval(() => {
 	void refreshMarketplaceProxyTools(server, { notify: true });
 }, refreshMs);
 
+let closing = false;
```
NicholaiVogel, line 38: Previously unresolved: re-entrant `shutdown` invocations (e.g. stdin end + SIGTERM arriving within the same tick) would call `server.close()` twice. Fixed — `closing` flag short-circuits all subsequent callers after the first.

PR #244, opened 2026-03-19
feat: retroactive memory supersession

NicholaiVogel:
## Summary

- Adds retroactive supersession system that detects when newly stored facts contradict existing entity attributes on the same entity/aspect, and marks the old attributes as superseded
- Two processing paths: inline pass (post-structural_classify) and periodic sweep (maintenance worker)
- Four-signal heuristic detection (negation polarity, antonym pairs, value conflict, temporal supersession) with optional LLM semantic fallback
- Extracts shared antonym/tokenizer module from worker.ts for reuse across contradiction detection and supersession
- Adds planning spec, dependencies.yaml entry, and cross-references in KNOWLEDGE-GRAPH.md and PIPELINE.md

## Motivation

When a memory like "broke up with Amari on March 14" was stored, older memories like "apartment hunting with Amari" and "be responsive to Amari's messages" remained `status='active'`. The supersession primitives existed (`supersedeAttribute`, `propagateMemoryStatus`) but only fired on explicit UPDATE/DELETE proposals — never retroactively. Every production agent memory system (Graphiti/Zep, Mem0, LangMem) uses prospective-only checking, but Signet's entity+aspect scoping makes retroactive scanning tractable (2-10 siblings per aspect).

## Key Design Decisions

- **No new migration** — uses existing `entity_attributes.status`/`superseded_by` schema
- **Constraints protected** — `kind='constraint'` attributes are never auto-superseded
- **Shadow mode respected** — proposals recorded without mutation when shadow mode is active
- **Idempotent** — `supersedeAttribute()` on already-superseded rows is a no-op
- **Lossless** — superseded attributes keep their rows; source memories archive to cold tier

## Test plan

- [x] 13 behavioral contract tests in `supersession.test.ts` (all passing)
- [x] `bun run build` passes
- [x] `bun run typecheck` passes (no new type errors)
- [x] Verified pre-existing worker.test.ts failures are unchanged by antonym extraction refactor
- [ ] Manual validation: store contradictory memories, verify old attributes get `status='superseded'` after structural_classify tick
- [ ] Monitor shadow mode proposals before enabling live supersession

### DIFF
--- a/packages/daemon/src/pipeline/worker.ts
+++ b/packages/daemon/src/pipeline/worker.ts
@@ -801,8 +801,8 @@ function runStructuralPass1(
 			// Resolve source entity ID from the entities table
 			const canonical = matchedTriple.source.trim().toLowerCase().replace(/\s+/g, " ");
 			const entityRow = db
-				.prepare("SELECT id, entity_type FROM entities WHERE canonical_name = ? LIMIT 1")
-				.get(canonical) as { id: string; entity_type: string } | undefined;
+				.prepare("SELECT id, entity_type, agent_id FROM entities WHERE canonical_name = ? LIMIT 1")
+				.get(canonical) as { id: string; entity_type: string; agent_id: string } | undefined;
 			if (!entityRow) continue;
 
 			// Skip if this memory already has a structural attribute row (classified or stub)
@@ -818,9 +818,9 @@ function runStructuralPass1(
 				`INSERT INTO entity_attributes
 				 (id, aspect_id, agent_id, memory_id, kind, content, normalized_content,
 				  confidence, importance, status, created_at, updated_at)
-				 VALUES (?, NULL, 'default', ?, 'attribute', ?, ?, ?, 0.5, 'active', ?, ?)`,
+				 VALUES (?, NULL, ?, ?, 'attribute', ?, ?, ?, 0.5, 'active', ?, ?)`,
 			).run(
-				attrId, fact.memoryId, fact.content, fact.normalizedContent,
+				attrId, entityRow.agent_id, fact.memoryId, fact.content, fact.normalizedContent,
 				fact.confidence, now, now,
 			);
 			stats.attributesCreated++;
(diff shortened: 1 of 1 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: retroactive memory supersession` (commit `fa9277c0`) now and I'll follow up shortly with feedback.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `feat: retroactive memory supersession` (commit `3fd10017`); taking another look at the feature work and I will report back shortly.

packages/daemon/src/pipeline/antonyms.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export function overlapCount(
+	left: readonly string[],
+	right: readonly string[],
+): number {
+	const rightSet = new Set(right);
+	let overlap = 0;
+	for (const token of left) {
+		if (rightSet.has(token)) overlap++;
+	}
```
NicholaiVogel, line 97: ANTONYM_SET is exported but never imported by supersession.ts or any other module (hasAntonymConflict iterates ANTONYM_PAIRS directly). Not a bug, but it's dead surface area. Remove or use it.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `feat: retroactive memory supersession` (commit `572c4853`); taking another look at the feature work and I will report back shortly.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -895,6 +833,7 @@ function runStructuralPass1(
 				entity_type: entityRow.entity_type,
 				fact_content: fact.content,
 				attribute_id: attrId,
+				agent_id: "default",
```
NicholaiVogel: Still hardcoding `agent_id: "default"` here means the fix in `processClassifyBatch` (`payloads[0].agent_id ?? "default"`) is a no-op for all agents: the payload will always carry `"default"`, so the `??` fallback never fires. The `entityRow` available in `runStructuralPass1` should have an `agent_id` field — forward it here: `agent_id: entityRow.agent_id ?? "default"`. Without this, supersession (and the `getAspectsForEntity` fix) will continue silently operating under the wrong agent scope for any non-default agent.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `feat: retroactive memory supersession` (commit `7b3bc8d9`); taking another look at the feature work and I will report back shortly.

PR #245, opened 2026-03-19
feat: add scope column for memory isolation

NicholaiVogel:
## Summary

- Adds nullable `scope TEXT` column to memories table (migration 033) with a partial index on non-NULL values — zero overhead for normal queries
- Default behavior: all searches add `AND scope IS NULL`, making scoped memories invisible unless explicitly requested
- remember/recall/forget endpoints accept optional `scope` parameter for clean namespace isolation
- Enables MemoryBench provider to isolate benchmark data per-question without polluting real memory

## Context

MemoryBench (`references/memorybench/`) needs per-question memory isolation via `containerTag`. Rather than hacking isolation through tags, this adds a proper nullable column. The Signet MemoryBench provider (in `references/`, gitignored) uses this to POST memories with `scope: containerTag` and recall only within that scope.

This also lays groundwork for future features needing memory namespacing (multi-agent isolation, Desire Paths stories).

## Test plan

- [ ] Restart daemon, verify migration 033 runs (`PRAGMA table_info(memories)` shows `scope`)
- [ ] POST a memory with `scope: "test"`, verify normal recall (no scope) doesn't return it
- [ ] POST recall with `scope: "test"`, verify only scoped memories return
- [ ] POST forget with `scope: "test"`, verify only scoped memories deleted
- [ ] Verify existing memory creation/search unchanged (no regression)
- [ ] `bun run build && bun run typecheck` pass

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -1946,7 +1946,7 @@ interface ForgetCandidatesRequest {
 	sourceType: string;
 	since: string;
 	until: string;
-	scope: string;
+	scope: string | null;
 	limit: number;
 }
 
@@ -2146,9 +2146,11 @@ function buildForgetCandidatesWhere(req: ForgetCandidatesRequest, alias: string)
 		parts.push(`${prefix}source_type = ?`);
 		args.push(req.sourceType);
 	}
-	if (req.scope) {
+	if (req.scope !== null) {
 		parts.push(`${prefix}scope = ?`);
 		args.push(req.scope);
+	} else {
+		parts.push(`${prefix}scope IS NULL`);
 	}
 	if (req.since) {
 		parts.push(`${prefix}created_at >= ?`);
@@ -3362,7 +3364,7 @@ app.post("/api/memory/forget", async (c) => {
 		sourceType: parseOptionalString(payload.source_type) ?? "",
 		since: parseOptionalString(payload.since) ?? "",
 		until: parseOptionalString(payload.until) ?? "",
-		scope: parseOptionalString(payload.scope) ?? "",
+		scope: parseOptionalString(payload.scope) ?? null,
 		limit,
 	};
 
@@ -3374,7 +3376,7 @@ app.post("/api/memory/forget", async (c) => {
 		request.sourceType.length > 0 ||
 		request.since.length > 0 ||
 		request.until.length > 0 ||
-		request.scope.length > 0;
+		request.scope !== null;
 	if (ids.length === 0 && !hasQueryScope) {
 		return c.json(
 			{
--- a/packages/daemon/src/memory-search.ts
+++ b/packages/daemon/src/memory-search.ts
@@ -173,6 +173,11 @@ export async function hybridRecall(
 	const minScore = cfg.search.min_score;
 
 	const filter = buildFilterClause(params);
+	// When scope filtering is active, vector search can't filter at query
+	// time so out-of-scope IDs get dropped at hydration. Over-fetch to
+	// compensate so the final result count still hits `limit`.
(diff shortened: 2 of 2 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: add scope column for memory isolation` (commit `6283ea76`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -2145,6 +2146,10 @@ function buildForgetCandidatesWhere(req: ForgetCandidatesRequest, alias: string)
 		parts.push(`${prefix}source_type = ?`);
 		args.push(req.sourceType);
 	}
+	if (req.scope) {
```
NicholaiVogel: **Bug (high): unscoped forget deletes scoped memories.**

`buildForgetCandidatesWhere` adds `scope = ?` only when `req.scope` is truthy, but adds no `scope IS NULL` guard when it is falsy. This is the opposite of how recall behaves: recall *always* defaults to `AND scope IS NULL` when no scope is provided.

Consequence: calling `/forget` without a `scope` parameter will match and delete scoped memories (e.g. MemoryBench rows) that the same caller could never read via `/recall`. A routine memory-cleanup operation silently nukes benchmark namespaces.

Fix — mirror the recall default:
```typescript
if (req.scope) {
  parts.push(`${prefix}scope = ?`);
  args.push(req.scope);
} else {
  parts.push(`${prefix}scope IS NULL`);
}
```
You will also need to adjust `ForgetCandidatesRequest.scope` to `string | null` (or keep it `string` and use `null` as the sentinel) so the handler can distinguish "caller passed null" from "caller passed a scope value".

packages/daemon/src/daemon.ts
```diff
@@ -1946,6 +1946,7 @@ interface ForgetCandidatesRequest {
 	sourceType: string;
 	since: string;
 	until: string;
+	scope: string;
```
NicholaiVogel: **Type inconsistency (low): `ForgetCandidatesRequest.scope` is `string`, not `string | null`.**

Every other scope-bearing type in this PR (`RecallParams`, `IngestEnvelope`, the remember body) uses `string | null`. Using `string` here (with `?? ""` at the parse site) means there is no way for a caller to explicitly request "filter to only NULL-scoped memories" via the forget path — the empty-string sentinel conflates "no filter" with "filter to unscoped". This should be `string | null` and the parse site should map absent/null payload values to `null` rather than `""`.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `feat: add scope column for memory isolation` (commit `9b51ae50`); taking another look at the feature work and I will report back shortly.

packages/daemon/src/memory-search.ts
```diff
@@ -159,6 +173,11 @@ export async function hybridRecall(
 	const minScore = cfg.search.min_score;
 
 	const filter = buildFilterClause(params);
+	// When scope filtering is active, vector search can't filter at query
```
NicholaiVogel, line 176: Previously flagged: vector search bypassed scope at query time, causing hydration to silently return fewer results than `limit`. Fixed: `vecTopK = top_k * 2` when `params.scope !== undefined`. The 2x factor is a heuristic — if the scoped fraction of the vector index is high (e.g. heavy benchmark usage) results can still come in short. For the current MemoryBench use-case this is fine. If stricter guarantees are needed later, consider a configurable multiplier or iterative over-fetch.

PR #246, opened 2026-03-19
fix(openclaw): harden plugin sync and patch plugins.allow

NicholaiVogel:
## Summary
- validate `@signetai/signet-memory-openclaw` global installs in `signet sync` by requiring `dist/index.js` before marking the sync version as current
- ensure OpenClaw config patching adds `signet-memory-openclaw` to `plugins.allow` (install path and sync-time `patchLoadPaths` path)
- make adapter publishing safer with `prepack` build + adapter tsconfig/type fixes
- include `@signetai/*` in workspace `build:deps` so adapter builds in the main dependency build flow

## Why
Two issues were reproducible:
1. `signet sync` could install a published adapter package that lacked `dist/`, then patch OpenClaw to load it, which fails at runtime.
2. OpenClaw warns when `plugins.allow` is empty and non-bundled plugins are discoverable; Signet patched load paths but did not explicitly trust the plugin id.

## Changes
- `packages/cli/src/cli.ts`
  - added runtime package validation (`dist/index.js`) for cached and fresh global installs
  - only write OpenClaw plugin sync version after post-install runtime validation succeeds
  - updated sync log message to reflect `plugins.load.paths/plugins.allow`
- `packages/connector-openclaw/src/index.ts`
  - added safe `plugins.allow` merge helper
  - patch plugin install flow (`patchAllConfigsWithPlugin`) to include `signet-memory-openclaw` in allowlist
  - patch sync flow (`patchLoadPaths`) to update `plugins.allow` alongside `plugins.load.paths`
- `packages/connector-openclaw/test/index.test.ts`
  - added tests for allowlist patching on install + patchLoadPaths
- `packages/adapters/openclaw/package.json`
  - added `prepack` build and `@types/node` dev dependency
- `packages/adapters/openclaw/tsconfig.json`
  - added Node types and excluded tests from declaration emit
- `package.json`
  - added `--filter '@signetai/*'` to `build:deps`

## Validation
- `bun test packages/connector-openclaw/test/index.test.ts`
- `cd packages/connector-openclaw && bun run build`
- `cd packages/adapters/openclaw && bun run build`
- `cd packages/adapters/openclaw && npm_config_cache=/tmp/npm-cache npm pack --dry-run`
- source compatibility check against `~/signet/signetai/references/openclaw` (`main` @ `33e829210e286c6070b497ad7a1bdfdc24414724`)

### DIFF
--- a/packages/adapters/openclaw/tsconfig.json
+++ b/packages/adapters/openclaw/tsconfig.json
@@ -3,7 +3,7 @@
 		"target": "ES2022",
 		"module": "ESNext",
 		"moduleResolution": "bundler",
-		"lib": ["ES2022", "DOM"],
+		"lib": ["ES2022"],
 		"types": ["node"],
 		"strict": true,
 		"esModuleInterop": true,
--- a/packages/connector-openclaw/test/index.test.ts
+++ b/packages/connector-openclaw/test/index.test.ts
@@ -417,6 +417,34 @@ describe("OpenClawConnector config patching", () => {
 		expect(patched.plugins.allow).toContain("signet-memory-openclaw");
 	});
 
+	it("patchLoadPaths patches allowlist when search path already exists", () => {
+		const configPath = join(tmpRoot, "openclaw.json");
+		writeFileSync(
+			configPath,
+			JSON.stringify(
+				{
+					plugins: {
+						load: {
+							paths: ["/tmp/a"],
+						},
+						entries: {},
+					},
+				},
+				null,
+				2,
+			),
+		);
+		process.env.OPENCLAW_CONFIG_PATH = configPath;
+
+		const connector = new OpenClawConnector();
+		const result = connector.patchLoadPaths("/tmp/a");
+		expect(result.patched).toContain(configPath);
+
+		const patched = JSON.parse(readFileSync(configPath, "utf-8"));
+		expect(patched.plugins.load.paths).toEqual(["/tmp/a"]);
+		expect(patched.plugins.allow).toContain("signet-memory-openclaw");
+	});
+
 	it("uninstall removes signet memory plugin from plugins.allow", async () => {
 		const configPath = join(tmpRoot, "openclaw.json");
 		writeFileSync(


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(openclaw): harden plugin sync and patch plugins.allow` (commit `fbca6749`) now and I'll follow up shortly with feedback.

packages/connector-openclaw/src/index.ts
```diff
@@ -736,6 +770,15 @@ export class OpenClawConnector extends BaseConnector {
 					config.signet = undefined;
```
NicholaiVogel, line 856: **Fragility: `mergePluginAllow` runs before `deepMerge(config, patch)`.**

The `allow` array is written onto `pluginsObj` and then `deepMerge(config, patch)` runs. Today's patches never carry an `allow` key so this is safe, but if a future caller passes a patch with `plugins.allow` the deep-merge will silently overwrite the carefully constructed allowlist. A short comment (or `console.assert(!patch.plugins?.allow, ...)` in dev builds) would document and guard the invariant.

packages/connector-openclaw/test/index.test.ts
```diff
@@ -358,4 +358,62 @@ describe("OpenClawConnector config patching", () => {
...
+
+		const connector = new OpenClawConnector();
+		const result = connector.patchLoadPaths("/tmp/b");
+		expect(result.patched).toContain(configPath);
+
+		const patched = JSON.parse(readFileSync(configPath, "utf-8"));
+		expect(patched.plugins.load.paths).toEqual(["/tmp/a", "/tmp/b"]);
+		expect(patched.plugins.allow).toContain("signet-memory-openclaw");
+	});
 });
```
NicholaiVogel, line 490: **Missing test case: `patchLoadPaths` when path already exists but `allow` is absent.**

The new test only exercises the case where both `searchPath` is new and `allow` needs to be added. A complementary test where `searchPath` is already in `load.paths` but `allow` is missing would verify that the `dirty` flag still triggers a write (and that `patched` includes the config path). Without it, a future refactor that gates `allowResult.changed` inside the `if (!existingPaths.includes(...))` block would pass all existing tests while silently regressing.

packages/adapters/openclaw/tsconfig.json
```diff
@@ -4,6 +4,7 @@
 		"module": "ESNext",
 		"moduleResolution": "bundler",
 		"lib": ["ES2022", "DOM"],
+		"types": ["node"],
```
NicholaiVogel, line 7: **Minor: `lib: ["ES2022", "DOM"]` combined with `types: ["node"]` is unusual for a Node-only adapter.**

DOM globals (e.g. `fetch`, `Request`, `ReadableStream`) and their Node counterparts share names but differ in signature across TypeScript versions. `skipLibCheck: true` suppresses conflicts today, but if it is ever removed type errors will surface. For a package targeting Node ≥18 (which ships a built-in `fetch`), dropping `"DOM"` from `lib` removes the ambiguity without losing anything the package actually uses.

NicholaiVogel:
Addressed the review feedback in commit `391866ae`.

Changes made:
- Fixed uninstall gap: `OpenClawConnector.uninstall()` now removes `signet-memory-openclaw` from `plugins.allow` across discovered configs.
- Added helper `removePluginAllow(...)` and connector method `removePluginFromAllow(...)` with the same safety shape checks/warnings pattern as other config patch paths.
- Added a guard in `patchAllConfigsWithPlugin(...)` to prevent accidental future clobbering if a patch ever tries to set `plugins.allow` directly before `deepMerge`.
- Added regression test: `uninstall removes signet memory plugin from plugins.allow`.

Validation rerun:
- `bun test packages/connector-openclaw/test/index.test.ts`
- `cd packages/connector-openclaw && bun run build`

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix(openclaw): harden plugin sync and patch plugins.allow` (commit `391866ae`); taking another look at the fixes and I will report back shortly.

packages/connector-openclaw/src/index.ts
```diff
@@ -469,6 +538,51 @@ export class OpenClawConnector extends BaseConnector {
 		return { filesRemoved, configsPatched };
```
NicholaiVogel, line 538: **Previously flagged – fixed.** `uninstall()` now calls `removePluginFromAllow` and merges its `patched` paths into `configsPatched`. The plugin ID is correctly removed from the allowlist on uninstall.

packages/connector-openclaw/src/index.ts
```diff
@@ -679,6 +793,12 @@ export class OpenClawConnector extends BaseConnector {
 		patched: string[];
```
NicholaiVogel, line 793: **Previously flagged – addressed.** The guard `throw new Error("patchAllConfigsWithPlugin patch must not set plugins.allow…")` prevents any future caller from accidentally clobbering the merged allowlist via the patch object.

packages/connector-openclaw/src/index.ts
```diff
@@ -452,8 +518,11 @@ export class OpenClawConnector extends BaseConnector {
 				},
 			},
 		});
+		const allowResult = this.removePluginFromAllow("signet-memory-openclaw");
```
NicholaiVogel, line 521: Minor: `allowResult.warnings` from `removePluginFromAllow` are collected but then silently dropped — `UninstallResult` has no `warnings` field, and the existing `hookResult`/`pluginResult` warnings are also dropped. No correctness impact, but if `UninstallResult` ever grows a `warnings` field, this will need a backfill here too.

NicholaiVogel:
Addressed the additional feedback in `c9959654`.

Changes:
- Added missing regression test for `patchLoadPaths` when `searchPath` already exists but `plugins.allow` is missing.
  - `patchLoadPaths("/tmp/a")` now verified to still patch/write allowlist and include config in `patched`.
- Updated adapter TS config per review note: removed `"DOM"` from `lib` in `packages/adapters/openclaw/tsconfig.json` (Node-only target with `types: ["node"]`).

Validation rerun:
- `bun test packages/connector-openclaw/test/index.test.ts`
- `cd packages/connector-openclaw && bun run build`
- `cd packages/adapters/openclaw && bun run build`

On the remaining minor observability note (`allowResult.warnings` during uninstall): agreed and noted; current `UninstallResult` shape has no `warnings` field and existing uninstall paths already drop warnings similarly. I kept behavior consistent in this PR.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix(openclaw): harden plugin sync and patch plugins.allow` (commit `c9959654`); taking another look at the fixes and I will report back shortly.

PR #247, opened 2026-03-20
fix: prevent CMD window flashing on Windows and fix workspace path matching

aaf2tbz:
## Summary

Fixes console window popups ("CMD flash") on Windows when the Signet daemon spawns subprocesses, and fixes workspace scope matching for Windows-style backslash paths.

## What

- Added `windowsHide: true` to all remaining `spawn`/`execSync` calls that were missing it:
  - `detectGitBranch()` in `daemon.ts` (2 execSync calls)
  - Shadow daemon spawn in `daemon.ts`
  - Predictor sidecar spawn in `predictor-client.ts`
- Rewrote `spawnHidden()` in `pipeline/provider.ts` to use `node:child_process` on Windows instead of `Bun.spawn`, since Bun.spawn does not support `windowsHide`
  - Converts Node.js Readable streams to Web ReadableStreams for interface compatibility
  - Added null guards and error rejection for robustness
- Normalized backslash (`\`) to forward slash (`/`) in marketplace workspace scope matching (`marketplace.ts` and `marketplace.rs`)

## How

**CMD flash prevention:** On Windows, any `child_process.spawn()` or `execSync()` without `windowsHide: true` opens a visible console window. This PR audits every spawn site in the daemon and adds the flag where missing.

**`spawnHidden()` refactor:** `Bun.spawn` doesn't expose a `windowsHide` option, so on Windows the function now branches to `node:child_process.spawn` with `windowsHide: true` and wraps the Node streams via `Readable.toWeb()`. Non-Windows platforms continue using `Bun.spawn` directly.

**Path normalization:** Windows paths use `\` separators, but marketplace scope entries may use `/`. Both the TypeScript and Rust scope-matching logic now normalize to `/` before comparison, so workspace scopes match correctly on Windows.

## Why

On Windows (including ARM64), every subprocess the daemon launches — git operations, predictor sidecar, shadow daemon, LLM provider CLI calls — was flashing a visible CMD window. This is disruptive and makes the daemon feel broken on Windows desktops. The workspace scope matching was also silently failing because `C:\Users\...` never matched `C:/Users/...`.

## Test plan

- [ ] Verify no CMD window flash when daemon starts on Windows x64
- [ ] Verify no CMD window flash when daemon starts on Windows ARM64
- [ ] Verify predictor sidecar launches without visible window
- [ ] Verify `spawnHidden()` CLI calls (Claude/Codex) work correctly on Windows
- [ ] Verify marketplace workspace scope matching works with backslash paths
- [ ] Verify non-Windows platforms are unaffected (Bun.spawn path unchanged)

### DIFF
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -105,6 +105,10 @@ interface SpawnResult {
 	kill(signal?: string): void;
 }
 
+// Module-level helper to avoid allocating a new closure per spawnHidden call.
+const toWebStream = (nodeStream: import("node:stream").Readable): ReadableStream<Uint8Array> =>
+	Readable.toWeb(nodeStream) as ReadableStream<Uint8Array>;
+
 function spawnHidden(cmd: string[], options?: { env?: Record<string, string | undefined> }): SpawnResult {
 	// Resolve the binary via Bun.which() so that .cmd wrappers on Windows
 	// are found correctly (mirrors scheduler/spawn.ts pattern).
@@ -123,15 +127,15 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 	// On Windows, use node:child_process with windowsHide to prevent
 	// console window flashing. Bun.spawn doesn't support windowsHide.
 	if (process.platform === "win32") {
+		// .cmd wrappers (e.g. claude.cmd, codex.cmd from npm) are shell scripts,
+		// not PE executables — node:child_process requires shell: true to run them.
+		const useShell = resolvedBin.endsWith(".cmd");
 		const child = nodeSpawn(resolvedBin, args, {
 			stdio: ["ignore", "pipe", "pipe"],
 			env: options?.env ? sanitizedEnv : undefined,
 			windowsHide: true,
+			shell: useShell,
 		});
-
-		// Convert Node.js Readable streams to Web ReadableStreams
-		const toWebStream = (nodeStream: import("node:stream").Readable): ReadableStream<Uint8Array> =>
-			Readable.toWeb(nodeStream) as ReadableStream<Uint8Array>;
 
 		const exitPromise = new Promise<number>((resolve, reject) => {
 			child.on("exit", (code) => resolve(code ?? 1));
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -127,15 +127,26 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 	// On Windows, use node:child_process with windowsHide to prevent
 	// console window flashing. Bun.spawn doesn't support windowsHide.
 	if (process.platform === "win32") {
-		// .cmd wrappers (e.g. claude.cmd, codex.cmd from npm) are shell scripts,
-		// not PE executables — node:child_process requires shell: true to run them.
-		const useShell = resolvedBin.endsWith(".cmd");
-		const child = nodeSpawn(resolvedBin, args, {
-			stdio: ["ignore", "pipe", "pipe"],
-			env: options?.env ? sanitizedEnv : undefined,
-			windowsHide: true,
-			shell: useShell,
-		});
+		// .cmd wrappers (e.g. claude.cmd, codex.cmd from npm) are batch scripts,
+		// not PE executables. Instead of shell: true (which exposes args to
+		// cmd.exe metacharacter interpretation — injection risk), invoke cmd.exe
+		// explicitly with properly quoted arguments.
+		let child: import("node:child_process").ChildProcess;
+		if (resolvedBin.endsWith(".cmd")) {
+			const quote = (s: string) => `"${s.replace(/"/g, '""')}"`;
+			const cmdLine = [quote(resolvedBin), ...args.map(quote)].join(" ");
+			child = nodeSpawn(process.env.COMSPEC || "cmd.exe", ["/d", "/s", "/c", `"${cmdLine}"`], {
+				stdio: ["ignore", "pipe", "pipe"],
+				env: options?.env ? sanitizedEnv : undefined,
+				windowsHide: true,
+			});
+		} else {
+			child = nodeSpawn(resolvedBin, args, {
+				stdio: ["ignore", "pipe", "pipe"],
+				env: options?.env ? sanitizedEnv : undefined,
+				windowsHide: true,
+			});
+		}
 
 		const exitPromise = new Promise<number>((resolve, reject) => {
 			child.on("exit", (code) => resolve(code ?? 1));
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -133,7 +133,7 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 		// explicitly with properly quoted arguments.
 		let child: import("node:child_process").ChildProcess;
 		if (resolvedBin.endsWith(".cmd")) {
-			const quote = (s: string) => `"${s.replace(/"/g, '""')}"`;
+			const quote = (s: string) => `"${s.replace(/%/g, "%%").replace(/"/g, '""')}"`;
 			const cmdLine = [quote(resolvedBin), ...args.map(quote)].join(" ");
 			child = nodeSpawn(process.env.COMSPEC || "cmd.exe", ["/d", "/s", "/c", `"${cmdLine}"`], {
 				stdio: ["ignore", "pipe", "pipe"],


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: prevent CMD window flashing on Windows and fix workspace path matching` (commit `fac09712`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +119,51 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 			if (v !== undefined) sanitizedEnv[k] = v;
 		}
 	}
+
+	// On Windows, use node:child_process with windowsHide to prevent
+	// console window flashing. Bun.spawn doesn't support windowsHide.
+	if (process.platform === "win32") {
+		const child = nodeSpawn(resolvedBin, args, {
+			stdio: ["ignore", "pipe", "pipe"],
+			env: options?.env ? sanitizedEnv : undefined,
```
NicholaiVogel: **Bug: Windows subprocess inherits no environment when `options?.env` is undefined.**

The condition `env: options?.env ? sanitizedEnv : undefined` passes `undefined` to `nodeSpawn`'s `env` option when no env override is provided. In Node.js `child_process.spawn`, passing `env: undefined` means inherit `process.env` — so this is actually fine. However, the intent is the same as the non-Windows Bun path below (`env: options?.env ? sanitizedEnv : undefined`). Double-check: if `options.env` is provided but all values are filtered out (empty `sanitizedEnv`), the child would still receive `env: sanitizedEnv` (an empty object), not inherit parent env. This is the same behavior as the Bun path. That said, the `available()` callers for Claude/Codex pass `{ env: { ...process.env, SIGNET_NO_HOOKS: '1' } }`, so `options.env` is always truthy in practice — but any future caller that passes `undefined` will get correct inheritance. The logic is fine as-is; no change needed here. ~~Revise this comment.~~

Actual issue: this is correct. Ignore this comment.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +119,51 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 			if (v !== undefined) sanitizedEnv[k] = v;
```
NicholaiVogel, line 123: **Bug: `sanitizedEnv` is built unconditionally but only used if `options?.env` is truthy.** When `options?.env` is provided with all `undefined` values, `sanitizedEnv` will be an empty `{}`, and `env: sanitizedEnv` (empty object) is passed to the child — it inherits **no** environment variables, not even PATH. This would cause `resolvedBin` to be a full path (from `Bun.which`), but the child process itself would have no PATH, HOME, etc. This matches the non-Windows Bun path behavior (same logic), so it is at least consistent. The real risk is callers passing `options.env` where all values happen to be `undefined`.

More critically: in the Windows branch, the `nodeSpawn` call passes `windowsHide: true` but omits `shell: false` explicitly. On Windows, `nodeSpawn` with `shell: false` (default) and a `.cmd` file as `resolvedBin` will fail — `.cmd` files require `shell: true` or explicit `cmd.exe /c` invocation. Since `Bun.which()` on Windows resolves `.cmd` wrappers to their `.cmd` path (mentioned in the comment), passing a `.cmd` path directly to `nodeSpawn` without `shell: true` will throw `ENOENT` or similar. This is a real breakage for `claude` and `codex` on Windows, which are typically `.cmd` wrappers installed by npm.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +119,51 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+	if (process.platform === "win32") {
+		const child = nodeSpawn(resolvedBin, args, {
+			stdio: ["ignore", "pipe", "pipe"],
+			env: options?.env ? sanitizedEnv : undefined,
+			windowsHide: true,
+		});
+
+		// Convert Node.js Readable streams to Web ReadableStreams
+		const toWebStream = (nodeStream: import("node:stream").Readable): ReadableStream<Uint8Array> =>
+			Readable.toWeb(nodeStream) as ReadableStream<Uint8Array>;
```
NicholaiVogel: **Potential issue: `.cmd` wrappers require `shell: true` in `node:child_process.spawn`.** The comment says `Bun.which()` finds `.cmd` wrappers on Windows. But `node:child_process.spawn` with `shell: false` (the default) cannot execute `.cmd` files directly — they are shell scripts, not PE executables. You need either `shell: true` or to invoke `cmd.exe /c <resolvedBin> ...args`. Without this, `nodeSpawn(resolvedBin, args, { windowsHide: true })` will fail with ENOENT when `resolvedBin` ends in `.cmd`. Consider adding `shell: resolvedBin.endsWith('.cmd')` or always using `shell: true` on win32 in this branch.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +119,51 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+
+		const exitPromise = new Promise<number>((resolve, reject) => {
+			child.on("exit", (code) => resolve(code ?? 1));
```
NicholaiVogel, line 152: Minor: `toWebStream` is a closure recreated on every `spawnHidden` call. Move it to module scope or define it once outside the function to avoid allocating a new function object per invocation.

NicholaiVogel (changes requested):
This PR correctly addresses CMD flash prevention and Windows path normalization. The changes are targeted and low-risk. One functional bug exists in the Windows `spawnHidden` path: when `options?.env` is undefined, the child process inherits no environment variables (passes `undefined` to `env`), while the non-Windows Bun path correctly omits the `env` option entirely — causing the process to inherit `process.env`. This is a behavioral inconsistency that could silently break CLI tools on Windows that depend on inherited env vars like PATH. A minor issue: the `toWebStream` helper is defined inside the hot path on every call rather than at module scope.

### Confidence: 7.6/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 7,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: prevent CMD window flashing on Windows and fix workspace path matching` (commit `729d4652`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +123,51 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 			if (v !== undefined) sanitizedEnv[k] = v;
 		}
 	}
+
+	// On Windows, use node:child_process with windowsHide to prevent
+	// console window flashing. Bun.spawn doesn't support windowsHide.
+	if (process.platform === "win32") {
+		// .cmd wrappers (e.g. claude.cmd, codex.cmd from npm) are shell scripts,
```
NicholaiVogel: **Shell injection risk on Windows via `shell: true`.**

When `resolvedBin.endsWith('.cmd')`, `shell: true` is set, causing Node.js to execute the command via `cmd.exe /d /s /c "..."`. The `args` array is joined and passed to cmd.exe without reliable escaping. Arguments sourced from LLM prompts, file contents, or git data can contain cmd.exe metacharacters (`&`, `|`, `^`, `"`, `(`, `)`) that will be interpreted by the shell.

For example, a file containing `foo & del /f /q C:\sensitive.txt` as content passed into a `callClaude` prompt would, on a machine where `claude.cmd` is the resolved binary, execute the `del` command.

**Fix**: shell-escape each argument before passing to cmd.exe when `useShell` is true. The simplest safe pattern is to double-quote each arg and escape internal double-quotes:
```typescript
const shellArgs = useShell
  ? args.map(a => `"${a.replace(/"/g, '\\"')}"`)
  : args;
const child = nodeSpawn(resolvedBin, shellArgs, { ..., shell: useShell });
```
Alternatively, prefer the `execFile`-style approach: pass the `.cmd` wrapper path directly without `shell: true` by prepending `cmd.exe /d /s /c` explicitly and quoting only the binary path, keeping args out of the shell expansion context. This is the pattern used by Node.js internals when `shell: true` is not set.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +123,51 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 			if (v !== undefined) sanitizedEnv[k] = v;
 		}
 	}
+
+	// On Windows, use node:child_process with windowsHide to prevent
```
NicholaiVogel, line 127: **Prior review env bug was a false positive — no action needed.** Node.js `child_process.spawn` documents `env` as defaulting to `process.env` when `undefined`, so `env: options?.env ? sanitizedEnv : undefined` correctly inherits the parent environment when no override is supplied. The non-Windows Bun path omits `env` entirely and gets the same behavior. This is consistent.

NicholaiVogel (changes requested):
The prior review's two issues are both resolved: `toWebStream` is now at module scope, and the env inheritance behavior with `env: undefined` in Node.js `child_process.spawn` is actually correct (Node defaults to `process.env` when `env` is `undefined`, so the prior review's concern was a false positive). One new issue is introduced: the `shell: useShell` addition for `.cmd` wrappers creates a shell-injection risk on Windows when args (including LLM prompts derived from file/git content) contain cmd.exe metacharacters such as `&`, `|`, `"`, or `^`.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 8,
  "pattern_correctness": 7,
  "documentation_coverage": 8
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: prevent CMD window flashing on Windows and fix workspace path matching` (commit `d4645199`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (commented):
The prior review's shell-injection issue (shell: useShell) is fixed. The new approach — explicit cmd.exe /d /s /c invocation with doubled-quote escaping — is sound for most metacharacters. One residual issue: cmd.exe expands %VARIABLE% tokens inside the /c command string before quote processing, even within double-quoted segments. If any argument (most critically the LLM prompt passed as -p <prompt>) contains a %VAR% pattern, cmd.exe will expand it. In the worst case, if an env var value itself contains unquoted cmd metacharacters (& | < >), this could escalate to command injection via percent expansion. This is considerably lower severity than the prior shell: true issue, but warrants a note. All other prior-review issues are resolved: toWebStream is at module scope, env: undefined inheritance is correct Node.js behavior. The windowsHide additions and path normalization are correct.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

packages/daemon/src/pipeline/provider.ts
```diff
@@ -124,15 +123,62 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+	// console window flashing. Bun.spawn doesn't support windowsHide.
+	if (process.platform === "win32") {
+		// .cmd wrappers (e.g. claude.cmd, codex.cmd from npm) are batch scripts,
+		// not PE executables. Instead of shell: true (which exposes args to
+		// cmd.exe metacharacter interpretation — injection risk), invoke cmd.exe
+		// explicitly with properly quoted arguments.
+		let child: import("node:child_process").ChildProcess;
+		if (resolvedBin.endsWith(".cmd")) {
```
NicholaiVogel, line 135: Residual injection risk: cmd.exe expands `%VARIABLE%` tokens in the command string passed to `/c` before it applies quote parsing, even inside double-quoted segments. The `quote()` helper escapes `"` → `""` but does not escape `%` → `%%`. If any argument — especially the LLM prompt (`-p <prompt>`) which can contain arbitrary user/file-derived text — includes a pattern like `%PATH%` or `%COMSPEC%`, cmd.exe will expand it. If the expanded value of an env var itself contains cmd metacharacters (`&`, `|`, `<`, `>`), this becomes a command-injection path.

Fix: add `%` escaping to the `quote` helper:
```typescript
const quote = (s: string) => `"${s.replace(/"/g, '""').replace(/%/g, '%%')}"`;
```
This is the standard mitigation for percent expansion in cmd.exe batch contexts.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: prevent CMD window flashing on Windows and fix workspace path matching` (commit `55e7119d`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
The incremental diff adds `replace(/%/g, "%%")` to the `quote` helper, directly resolving the percent-expansion injection risk flagged in the prior review. With `%` → `%%` escaping combined with `"` → `""` escaping and the `/d /s /c` invocation flags, the cmd.exe attack surface is now adequately closed: `%VAR%` expansion is suppressed by doubling, and delayed expansion (`!VAR!`) is suppressed by the `/d` flag. All prior review issues are resolved: `toWebStream` is at module scope, `env: undefined` inheritance is correct Node.js behavior, `shell: useShell` injection risk was previously fixed, and percent-expansion is now handled. No new issues introduced.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: prevent CMD window flashing on Windows and fix workspace path matching` (commit `53fb8201`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
All prior review issues are resolved and were approved at sha=55e7119d. The incremental diff since that approval contains only documentation additions (retroactive supersession spec, KNOWLEDGE-GRAPH.md, PIPELINE.md), CHANGELOG entries, version bumps, and one build filter addition. No functional code changes are present. Nothing to block merge.

### Confidence: 8.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 9
}
```

- package.json:24 - Minor observation: `--filter '@signetai/*'` is new in `build:deps`. Verify this doesn't accidentally pull in packages that should only be built via `build:signetai` or `build:publish`. Not a blocker — the filter is glob-scoped and intentional given the existing `build:signetai` script — but worth a quick sanity check that the filter doesn't cause double-builds or ordering issues with the dependent packages.

PR #249, opened 2026-03-20
fix: remove provider icon tinting and auto-sync pre-installed apps

aaf2tbz:
## Summary

- **Remove icon tinting from marketplace providers** — The `hue-rotate` CSS filter was being applied to GitHub org avatars in both the MCP Servers tab and Agent Skills cards, altering the appearance of provider icons. This has been removed to preserve original branding/icon appearance for legal compliance. The unused `getHueRotate` utility has been cleaned up.

- **Auto-sync pre-installed apps to OS page** — Installed MCP servers that hadn't gone through the probe/install flow were invisible on the Apps (Signet OS) page. The `GET /api/os/tray` endpoint now detects installed servers missing from the app tray and creates stub entries automatically, so pre-installed applications appear without manual action.

## Changes

| File | Change |
|------|--------|
| `packages/cli/dashboard/src/lib/card-utils.ts` | Remove `getHueRotate` function, update hash comment |
| `packages/cli/dashboard/src/lib/components/marketplace/McpServersTab.svelte` | Remove `hue-rotate` filter from installed + catalog avatar img tags |
| `packages/cli/dashboard/src/lib/components/skills/SkillCard.svelte` | Remove `hue-rotate` filter from skill avatar img tag |
| `packages/daemon/src/routes/app-tray.ts` | Sync installed MCP servers into app tray on GET /api/os/tray |

## Test plan

- [ ] Verify MCP server icons in Browse and Installed tabs render without color tinting
- [ ] Verify skill card icons render without color tinting
- [ ] Verify monogram fallback still shows correct deterministic background colors
- [ ] Install an MCP server via marketplace, confirm it appears on the OS/Apps page
- [ ] With a pre-installed MCP server (already in mcp-servers.json but not in app-tray.json), load the OS tab and confirm it appears automatically in the dock/tray
- [ ] Confirm stub entries are persisted to app-tray.json after first sync

### DIFF
--- a/packages/daemon/src/routes/app-tray.ts
+++ b/packages/daemon/src/routes/app-tray.ts
@@ -4,7 +4,7 @@ import type { Hono } from "hono";
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { homedir } from "node:os";
-import type { SignetAppManifest } from "@signet/core";
+import type { AutoCardToolAction, AutoCardResource, SignetAppManifest } from "@signet/core";
 
 import { isPrivateHostname } from "../url-validation.js";
 import {
@@ -89,7 +89,7 @@ export function mountAppTrayRoutes(app: Hono): void {
 			const stubs = missing.map((server) => ({
 				id: server.id,
 				name: server.name,
-				icon: undefined,
+				icon: null as string | null,
 				state: "tray" as const,
 				manifest: {
 					name: server.name,
@@ -97,8 +97,8 @@ export function mountAppTrayRoutes(app: Hono): void {
 				},
 				autoCard: {
 					name: server.name,
-					tools: [] as const,
-					resources: [] as const,
+					tools: [] as AutoCardToolAction[],
+					resources: [] as AutoCardResource[],
 					hasAppResources: false,
 					defaultSize: { w: 4, h: 3 },
 				},


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: remove provider icon tinting and auto-sync pre-installed apps` (commit `47b52d33`) now and I'll follow up shortly with feedback.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,61 @@ function findFreeGridPosition(
...
+						tools: [],
+						resources: [],
+						hasAppResources: false,
+						defaultSize: { w: 4, h: 3 },
+					},
+					hasDeclaredManifest: false,
+					createdAt: now,
+					updatedAt: now,
+				});
+			}
```
NicholaiVogel, line 132: **Race condition / data loss**: The GET handler performs an unguarded read-modify-write on `app-tray.json`. If a `PATCH /api/os/tray/:id` (or any other write) completes between the `loadAppTray()` read and the `writeFileSync` here, this GET will overwrite that update — silently discarding state changes. Example sequence:
1. GET reads `[entry1, entry2]`, finds `server3` missing
2. PATCH modifies `entry1`, writes `[entry1-modified, entry2]`
3. GET writes `[entry1, entry2, server3-stub]` ← PATCH change lost

Minimally, wrap the read-then-write in a try/finally and reload the file immediately before writing so the written state reflects the latest persisted tray. A file lock or a dedicated write helper used by all routes would be the robust fix.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,61 @@ function findFreeGridPosition(
...
+			writeFileSync(
+				join(dir, "app-tray.json"),
+				JSON.stringify(tray, null, 2),
+			);
+
+			logger.info(
+				"os",
+				`Synced ${missing.length} installed server(s) to app tray`,
+			);
+		}
```
NicholaiVogel, line 136: **Unhandled write error turns GET into 500**: `writeFileSync` here has no try/catch. If the write fails for any reason (disk full, permission denied, directory missing despite the `mkdirSync` guard), the uncaught exception propagates out of the route handler. Hono will catch it and return a 500, which breaks the client for what should be a read-only request. The sync is a best-effort background concern — wrap the entire `if (missing.length > 0)` write block in try/catch and log the error rather than letting it surface to the caller:
```ts
try {
  writeFileSync(join(dir, "app-tray.json"), JSON.stringify(tray, null, 2));
} catch (err) {
  logger.warn("os", `Failed to persist auto-synced tray entries: ${err}`);
}
```

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,61 @@ function findFreeGridPosition(
...
+			}
+
+			// Persist so future fetches don't re-sync the same servers
+			const agentsDir =
+				process.env.SIGNET_PATH || join(homedir(), ".agents");
+			const dir = join(agentsDir, "marketplace");
+			if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
+			writeFileSync(
+				join(dir, "app-tray.json"),
+				JSON.stringify(tray, null, 2),
```
NicholaiVogel: **Path construction duplicates existing helpers**: `getAgentsDir()` and `getMarketplaceDir()` are defined later in this file and already encode the `SIGNET_PATH || ~/.agents` / `marketplace` logic. Use them here instead of re-implementing inline to avoid divergence if the path convention ever changes:
```ts
const dir = getMarketplaceDir();
// mkdirSync guard is already handled by ensureMarketplaceDir()
ensureMarketplaceDir();
writeFileSync(join(dir, "app-tray.json"), JSON.stringify(tray, null, 2));
```

packages/daemon/src/routes/app-tray.ts
aaf2tbz: Addressed in 39c0b4d5 — reload fresh tray before writing to narrow the race window. The entire persist block is now best-effort (try/catch) so concurrent updates aren't silently lost and GET never returns 500 from a write failure.

packages/daemon/src/routes/app-tray.ts
aaf2tbz: Addressed in 39c0b4d5 — wrapped the write block in try/catch, logs warning on failure. GET always returns the merged view regardless of persist outcome.

packages/daemon/src/routes/app-tray.ts
aaf2tbz: Addressed in 39c0b4d5 — now using ensureMarketplaceDir() and getMarketplaceDir() instead of inline path construction.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: remove provider icon tinting and auto-sync pre-installed apps` (commit `39c0b4d5`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,71 @@ function findFreeGridPosition(
...
+				icon: undefined,
+				state: "tray" as const,
+				manifest: {
+					name: server.name,
+					defaultSize: { w: 4, h: 3 },
+				},
+				autoCard: {
+					name: server.name,
+					tools: [] as const,
+					resources: [] as const,
```
NicholaiVogel: **Fixed from prior review** — write failure is now caught and logged via `logger.warn`. No longer a silent 500 risk.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,71 @@ function findFreeGridPosition(
...
+					hasAppResources: false,
+					defaultSize: { w: 4, h: 3 },
+				},
+				hasDeclaredManifest: false,
+				createdAt: now,
+				updatedAt: now,
```
NicholaiVogel, line 107: **Fixed from prior review** — `getMarketplaceDir()` and `ensureMarketplaceDir()` helpers are now used here instead of re-implementing the path inline.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,71 @@ function findFreeGridPosition(
...
 		const tray = loadAppTray();
+		const installed = readInstalledServersPublic();
+		const trayIds = new Set(tray.map((e) => e.id));
+
+		const missing = installed.filter(
+			(s) => s.enabled && !trayIds.has(s.id),
+		);
+
+		if (missing.length > 0) {
+			const now = new Date().toISOString();
```
NicholaiVogel, line 88: **Mitigated from prior review** — the race condition is substantially narrowed: a fresh `loadAppTray()` is read immediately before the write and only genuinely missing stubs are appended. A narrow TOCTOU window remains between that read and the `writeFileSync`, but it is tolerable for a best-effort GET-side sync. The comment makes the intent clear.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,71 @@ function findFreeGridPosition(
...
+			const stubs = missing.map((server) => ({
+				id: server.id,
+				name: server.name,
```
NicholaiVogel, line 91: `icon: undefined` is silently omitted by `JSON.stringify`, so persisted stubs will not carry an `icon` field at all. Entries written by the normal install flow likely carry `icon: null` or similar. This is functionally harmless today but could be a latent gotcha for future code that checks `'icon' in entry`. Consider using `icon: null` explicitly for clarity.

packages/daemon/src/routes/app-tray.ts
```diff
@@ -70,10 +70,71 @@ function findFreeGridPosition(
...
+				icon: undefined,
+				state: "tray" as const,
+				manifest: {
+					name: server.name,
+					defaultSize: { w: 4, h: 3 },
+				},
```
NicholaiVogel, line 97: `tools: [] as const` and `resources: [] as const` produce `readonly never[]`. If the tray entry type declares these as `Tool[]` / `Resource[]` (mutable arrays), TypeScript will flag an assignability error. At runtime this is safe, but worth verifying the type compiles cleanly.

NicholaiVogel (approved):
All three issues from the prior review are resolved: (1) the race condition is substantially mitigated by re-reading the tray immediately before the write and filtering to only missing entries; (2) the writeFileSync is now wrapped in try/catch with a logger.warn on failure; (3) inline path construction is replaced by the getMarketplaceDir()/ensureMarketplaceDir() helpers. One minor residual concern: `icon: undefined` is silently dropped by JSON.stringify, so persisted stubs will lack the field entirely rather than carrying an explicit null — this is functionally fine but differs from entries written by the normal install flow and could surprise future code that checks `'icon' in entry`. The `tools: [] as const` / `resources: [] as const` assertions produce `readonly never[]`; if the tray entry type expects a mutable array this will be a TS error at the call site, though it is safe at runtime. A narrow TOCTOU window between the fresh loadAppTray() and writeFileSync still exists but is acceptable for a best-effort background sync on a read endpoint, and the comment makes the intent clear.

### Confidence: 7.7/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 8,
  "pattern_correctness": 7,
  "documentation_coverage": 7
}
```

packages/daemon/src/routes/app-tray.ts
aaf2tbz: Addressed in f2a6b572 — changed to explicit `icon: null` with type annotation `string | null` so JSON.stringify persists the field consistently with the install flow.

packages/daemon/src/routes/app-tray.ts
aaf2tbz: Addressed in f2a6b572 — replaced `[] as const` with properly typed `[] as AutoCardToolAction[]` and `[] as AutoCardResource[]` (imported from @signet/core) to avoid readonly never[] assignability issues.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: remove provider icon tinting and auto-sync pre-installed apps` (commit `f2a6b572`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
The incremental diff directly resolves all three residual concerns flagged in the prior approved review: `icon: null` is now explicit and round-trips correctly through JSON.stringify; `tools` and `resources` are typed as mutable `AutoCardToolAction[]`/`AutoCardResource[]` instead of `readonly never[]`, eliminating the potential TS assignability error at the call site. No new issues introduced. Frontend cleanup (hue-rotate removal) was already approved and is unchanged.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

PR #251, opened 2026-03-20
fix: settings persist across refresh without daemon restart

aaf2tbz:
## Summary

Settings edited in the dashboard now persist across page refresh without needing a daemon restart. Three changes fix the full pipeline from UI save to daemon-side config reload.

### What was broken

1. **Store didn't re-sync after save** — `SettingsStore.save()` wrote to disk via POST but kept its stale in-memory init data. If the store re-initialized (e.g., during tab switch), it would read from the original `configFiles` prop, not from disk.

2. **Page data stayed stale** — The `configFiles` prop in `+page.svelte` was fetched once at load. Components like `IdentityPanel` that read directly from this prop (not the store) never saw saved changes until a full refresh.

3. **Auth config cached at startup** — The daemon loaded `agent.yaml` auth settings once in `main()` and never reloaded. Changing auth mode, rate limits, or tokens required a daemon restart.

### How it's fixed

| File | Fix |
|------|-----|
| `settings.svelte.ts` | After successful save, re-fetch config from disk via `getConfigFiles()` and re-init the store with `lastInitSource = null` to force bypass the reference guard |
| `SettingsTab.svelte` | Call `invalidateAll()` after `saveAll()` to trigger SvelteKit's `+page.ts` load, refreshing `configFiles` for IdentityPanel and all consumers |
| `daemon.ts` | File watcher's `change` handler now detects `agent.yaml` changes and reloads auth config + rate limiters from disk without restart |

## Test plan

- [ ] Edit a setting in the dashboard (e.g., toggle a pipeline flag), save, refresh page — verify the change persists
- [ ] Edit `agent.yaml` externally (e.g., in a text editor), reload the dashboard — verify the new value appears
- [ ] Change auth rate limit in `agent.yaml` via dashboard, verify it takes effect without daemon restart
- [ ] Save settings, switch tabs, switch back — verify settings still show saved values (no stale re-init)

### DIFF
--- a/packages/cli/dashboard/src/lib/stores/settings.svelte.ts
+++ b/packages/cli/dashboard/src/lib/stores/settings.svelte.ts
@@ -1,4 +1,4 @@
-import { type ConfigFile, saveConfigFileResult } from "$lib/api";
+import { type ConfigFile, getConfigFiles, saveConfigFileResult } from "$lib/api";
 import { toast } from "$lib/stores/toast.svelte";
 import { parse, stringify } from "yaml";
 
@@ -416,6 +416,14 @@ class SettingsStore {
 
 			if (successfulFiles.length > 0) {
 				this.lastSavedAt = new Date().toISOString();
+				// Re-fetch from disk and re-init so store matches persisted state
+				try {
+					const fresh = await getConfigFiles();
+					this.lastInitSource = null;
+					this.init(fresh);
+				} catch {
+					// Non-fatal — store still holds the just-saved state
+				}
 			}
 
 			if (failed.length === 0 && successfulFiles.length > 0) {
--- a/packages/cli/dashboard/src/lib/stores/settings.svelte.ts
+++ b/packages/cli/dashboard/src/lib/stores/settings.svelte.ts
@@ -1,4 +1,4 @@
-import { type ConfigFile, getConfigFiles, saveConfigFileResult } from "$lib/api";
+import { type ConfigFile, saveConfigFileResult } from "$lib/api";
 import { toast } from "$lib/stores/toast.svelte";
 import { parse, stringify } from "yaml";
 
@@ -416,14 +416,9 @@ class SettingsStore {
 
 			if (successfulFiles.length > 0) {
 				this.lastSavedAt = new Date().toISOString();
-				// Re-fetch from disk and re-init so store matches persisted state
-				try {
-					const fresh = await getConfigFiles();
-					this.lastInitSource = null;
-					this.init(fresh);
-				} catch {
-					// Non-fatal — store still holds the just-saved state
-				}
+				// Clear init guard so the next invalidateAll() re-init
+				// picks up the fresh disk state
+				this.lastInitSource = null;
 			}
 
 			if (failed.length === 0 && successfulFiles.length > 0) {
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -8834,6 +8834,22 @@ function startFileWatcher() {
 		logger.info("watcher", "File changed", { path });
 		scheduleAutoCommit(path);
 
+		// Reload auth config when agent.yaml changes on disk
+		if (path.endsWith("agent.yaml") || path.endsWith("AGENT.yaml")) {
+			try {
+				const cfg = loadMemoryConfig(AGENTS_DIR);
+				authConfig = cfg.auth;
+				const rl = authConfig.rateLimits;
+				if (rl.forget) authForgetLimiter = new AuthRateLimiter(rl.forget.windowMs, rl.forget.max);
+				if (rl.modify) authModifyLimiter = new AuthRateLimiter(rl.modify.windowMs, rl.modify.max);
+				if (rl.batchForget) authBatchForgetLimiter = new AuthRateLimiter(rl.batchForget.windowMs, rl.batchForget.max);
+				if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
+				logger.info("config", "Auth config reloaded from disk");
+			} catch (e) {
+				logger.error("config", "Failed to reload auth config", e as Error);
+			}
+		}
+
 		// If any identity file changed, sync to harness configs
 		const SYNC_TRIGGER_FILES = ["AGENTS.md", "SOUL.md", "IDENTITY.md", "USER.md", "MEMORY.md"];
 		if (SYNC_TRIGGER_FILES.some((f) => path.endsWith(f))) {
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -8835,15 +8835,16 @@ function startFileWatcher() {
 		scheduleAutoCommit(path);
 
 		// Reload auth config when agent.yaml changes on disk
-		if (path.endsWith("agent.yaml") || path.endsWith("AGENT.yaml")) {
+		const base = path.split("/").pop();
+		if (base === "agent.yaml" || base === "AGENT.yaml") {
 			try {
 				const cfg = loadMemoryConfig(AGENTS_DIR);
 				authConfig = cfg.auth;
 				const rl = authConfig.rateLimits;
-				if (rl.forget) authForgetLimiter = new AuthRateLimiter(rl.forget.windowMs, rl.forget.max);
-				if (rl.modify) authModifyLimiter = new AuthRateLimiter(rl.modify.windowMs, rl.modify.max);
-				if (rl.batchForget) authBatchForgetLimiter = new AuthRateLimiter(rl.batchForget.windowMs, rl.batchForget.max);
-				if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
+				authForgetLimiter = rl.forget ? new AuthRateLimiter(rl.forget.windowMs, rl.forget.max) : new AuthRateLimiter(60_000, 30);
+				authModifyLimiter = rl.modify ? new AuthRateLimiter(rl.modify.windowMs, rl.modify.max) : new AuthRateLimiter(60_000, 60);
+				authBatchForgetLimiter = rl.batchForget ? new AuthRateLimiter(rl.batchForget.windowMs, rl.batchForget.max) : new AuthRateLimiter(60_000, 5);
+				authAdminLimiter = rl.admin ? new AuthRateLimiter(rl.admin.windowMs, rl.admin.max) : new AuthRateLimiter(60_000, 10);
 				logger.info("config", "Auth config reloaded from disk");
 			} catch (e) {
 				logger.error("config", "Failed to reload auth config", e as Error);
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -8840,6 +8840,7 @@ function startFileWatcher() {
 			try {
 				const cfg = loadMemoryConfig(AGENTS_DIR);
 				if (!cfg.auth) throw new Error("Missing auth section in agent.yaml");
+				if (!cfg.auth.rateLimits) throw new Error("Missing rateLimits in auth config");
 				authConfig = cfg.auth;
 				const rl = authConfig.rateLimits;
 				authForgetLimiter = rl.forget ? new AuthRateLimiter(rl.forget.windowMs, rl.forget.max) : new AuthRateLimiter(60_000, 30);
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -8835,10 +8835,11 @@ function startFileWatcher() {
 		scheduleAutoCommit(path);
 
 		// Reload auth config when agent.yaml changes on disk
-		const base = path.split("/").pop();
+		const base = basename(path);
 		if (base === "agent.yaml" || base === "AGENT.yaml") {
 			try {
 				const cfg = loadMemoryConfig(AGENTS_DIR);
+				if (!cfg.auth) throw new Error("Missing auth section in agent.yaml");
 				authConfig = cfg.auth;
 				const rl = authConfig.rateLimits;
 				authForgetLimiter = rl.forget ? new AuthRateLimiter(rl.forget.windowMs, rl.forget.max) : new AuthRateLimiter(60_000, 30);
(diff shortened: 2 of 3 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: settings persist across refresh without daemon restart` (commit `52500679`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -8834,6 +8834,22 @@ function startFileWatcher() {
...
+				const cfg = loadMemoryConfig(AGENTS_DIR);
+				authConfig = cfg.auth;
+				const rl = authConfig.rateLimits;
+				if (rl.forget) authForgetLimiter = new AuthRateLimiter(rl.forget.windowMs, rl.forget.max);
+				if (rl.modify) authModifyLimiter = new AuthRateLimiter(rl.modify.windowMs, rl.modify.max);
+				if (rl.batchForget) authBatchForgetLimiter = new AuthRateLimiter(rl.batchForget.windowMs, rl.batchForget.max);
+				if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
+				logger.info("config", "Auth config reloaded from disk");
+			} catch (e) {
+				logger.error("config", "Failed to reload auth config", e as Error);
```
NicholaiVogel, line 8852: **Bug: removing a rate-limit entry from agent.yaml leaves the old limiter active.**

The `if (rl.forget)` guards only update a limiter when the config key is present. If an operator removes (or nulls) one of these keys to disable that rate limit, the old `AuthRateLimiter` instance remains in `authForgetLimiter` / `authModifyLimiter` / etc. and keeps enforcing the previous limits — silently, until a daemon restart.

Fix: replace the conditional updates with unconditional ones, setting the limiter to `null` (or a no-op instance) when the config key is absent:
```ts
authForgetLimiter = rl.forget ? new AuthRateLimiter(rl.forget.windowMs, rl.forget.max) : null;
authModifyLimiter = rl.modify ? new AuthRateLimiter(rl.modify.windowMs, rl.modify.max) : null;
// …and so on
```
This requires the call sites to tolerate `null`, but they almost certainly already do (the limiters were not created at startup when the config key was absent).

packages/daemon/src/daemon.ts
```diff
@@ -8834,6 +8834,22 @@ function startFileWatcher() {
 		logger.info("watcher", "File changed", { path });
 		scheduleAutoCommit(path);
 
+		// Reload auth config when agent.yaml changes on disk
```
NicholaiVogel, line 8837: **Overly broad `endsWith` check can match unintended files.**

`path.endsWith('agent.yaml')` matches `custom-agent.yaml`, `bad-agent.yaml`, etc. Use a stricter check:
```ts
const base = path.split('/').pop();
if (base === 'agent.yaml' || base === 'AGENT.yaml') { … }
```
While the data loaded is always from `AGENTS_DIR` (so there is no security impact), spurious reloads on unrelated files add noise and could mask real reload failures in logs.

packages/cli/dashboard/src/lib/stores/settings.svelte.ts
```diff
@@ -416,6 +416,14 @@ class SettingsStore {
...
 				this.lastSavedAt = new Date().toISOString();
+				// Re-fetch from disk and re-init so store matches persisted state
+				try {
+					const fresh = await getConfigFiles();
+					this.lastInitSource = null;
+					this.init(fresh);
+				} catch {
+					// Non-fatal — store still holds the just-saved state
+				}
 			}
```
NicholaiVogel, line 423: **Double disk fetch + double store re-init on every save.**

`save()` fetches config via `getConfigFiles()` and calls `this.init(fresh)`, then immediately after, `SettingsTab.saveAll()` calls `invalidateAll()`, which triggers the page loader again, gets another fresh `ConfigFile[]`, and the `$effect` fires `st.init(newFiles)` a second time.

This second init calls `this.lastSaveFeedback = ''`, wiping the `'Saved agent.yaml'` string that was just written (the toast has already fired so it is not user-visible, but the field becomes unreliable). It also makes two sequential `GET /api/config-files` calls per save.

The simpler fix: remove the internal `getConfigFiles()` re-fetch from `save()` and let `invalidateAll()` (in `SettingsTab`) be the single source of truth for refreshing store state. The store's `agentSnapshot` / `configSnapshot` are already updated before the re-fetch, so `isDirty` is correct without the extra init.

NicholaiVogel (changes requested):
Three meaningful issues found. The most significant is in daemon.ts: removing a rate-limit entry from agent.yaml silently leaves the old limiter in place, so the config change has no effect until a daemon restart — the opposite of what this PR intends. A secondary issue is that settings.svelte.ts calls getConfigFiles() inside save() and then SettingsTab.svelte calls invalidateAll() immediately after, causing two round-trips to disk and two sequential store re-inits on every save; the second init clears lastSaveFeedback (a cosmetic regression, not a data loss). Finally, the endsWith() file-path check in the watcher matches any file whose basename ends with agent.yaml (e.g. bad-agent.yaml), which is broader than intended.

### Confidence: 7.2/10
- Injection risk detection confidence: 6


```pr-review-confidence-json
{
  "style_maintainability": 7,
  "injection_risk_detection": 6,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Addressed in 224109a6 — limiters are now always reset unconditionally. When a config key is absent, the limiter falls back to a permissive default (matching the initial values at module scope) rather than staying as null, avoiding changes at every call site.

packages/daemon/src/daemon.ts
aaf2tbz: Addressed in 224109a6 — replaced endsWith with strict basename check via path.split('/').pop() so only exactly agent.yaml or AGENT.yaml triggers a reload.

packages/cli/dashboard/src/lib/stores/settings.svelte.ts
aaf2tbz: Addressed in 224109a6 — removed the internal getConfigFiles() re-fetch from save(). Now save() only clears lastInitSource so the subsequent invalidateAll() in SettingsTab is the single path for re-fetching and re-initing. No more double fetch or lastSaveFeedback wipe.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: settings persist across refresh without daemon restart` (commit `224109a6`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -8834,6 +8834,23 @@ function startFileWatcher() {
...
```
NicholaiVogel, line 8837: Minor hardening note: `path.split('/').pop()` returns `undefined` on an empty string path (safe here, `undefined === 'agent.yaml'` is false) and assumes POSIX separators. If the watcher ever runs on Windows paths with backslashes, consider `require('path').basename(path)` for portability. Not a blocker given the daemon's deployment target, but worth noting for future maintainers.

packages/daemon/src/daemon.ts
```diff
@@ -8834,6 +8834,23 @@ function startFileWatcher() {
...
+		const base = path.split("/").pop();
+		if (base === "agent.yaml" || base === "AGENT.yaml") {
+			try {
```
NicholaiVogel, line 8840: If `agent.yaml` exists but has no `auth` section, `cfg.auth` is `undefined` and `authConfig = cfg.auth` sets `authConfig` to `undefined` before `authConfig.rateLimits` throws on the next line. The catch block recovers, but `authConfig` is now `undefined` for the duration of that event loop tick before it throws — any concurrent request handler that reads `authConfig` between those two lines would see `undefined`. This is a pre-existing race, not introduced by this PR, and extremely unlikely in practice. Worth a future guard like `if (!cfg.auth) throw new Error('Missing auth section in agent.yaml')`.

NicholaiVogel (approved):
All three issues from the prior review are resolved. (1) The rate-limiter removal bug is fixed: the new ternary unconditionally replaces every limiter with a default when the config key is absent, so removing a rate-limit entry no longer leaves the old limiter in place. (2) The double round-trip is eliminated: `getConfigFiles()` was removed from `save()`, leaving a single `invalidateAll()` path through SvelteKit. The `lastSaveFeedback` cosmetic regression is gone. (3) The overly-broad `endsWith()` path check is replaced with exact basename equality via `path.split('/').pop()`. No new issues found.

### Confidence: 7.6/10


```pr-review-confidence-json
{
  "style_maintainability": 7,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 6
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Addressed in e4c1080d — replaced path.split('/').pop() with basename(path) (already imported from 'path') for cross-platform portability.

packages/daemon/src/daemon.ts
aaf2tbz: Addressed in e4c1080d — added explicit guard: if (!cfg.auth) throw new Error('Missing auth section in agent.yaml') before assigning to authConfig, so it never becomes undefined even momentarily.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: settings persist across refresh without daemon restart` (commit `e4c1080d`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -8834,6 +8834,24 @@ function startFileWatcher() {
 		logger.info("watcher", "File changed", { path });
 		scheduleAutoCommit(path);
 
+		// Reload auth config when agent.yaml changes on disk
+		const base = basename(path);
```
NicholaiVogel, line 8838: Verify `basename` is imported from `'path'` at the top of daemon.ts. The diff switches from the inline `path.split('/').pop()` idiom to `basename(path)`, which requires `import { basename } from 'path'`. If it isn't imported this is a compile-time error. Low risk since it would surface immediately, but worth a quick check.

packages/daemon/src/daemon.ts
```diff
@@ -8834,6 +8834,24 @@ function startFileWatcher() {
...
+		if (base === "agent.yaml" || base === "AGENT.yaml") {
+			try {
+				const cfg = loadMemoryConfig(AGENTS_DIR);
```
NicholaiVogel, line 8841: The `!cfg.auth` guard correctly prevents clobbering `authConfig` with `undefined`. However, if `cfg.auth` is present but `cfg.auth.rateLimits` is undefined, `const rl = authConfig.rateLimits` will be `undefined` and the ternary accesses like `rl.forget` will throw a TypeError that the catch block will absorb — leaving the old limiters in place silently. This is a pre-existing gap, not introduced here, but now that the surrounding code is hardened it's worth a follow-up `if (!rl) throw …` guard for completeness.

NicholaiVogel (approved):
Incremental diff is a net improvement over the already-approved sha=224109a6. Two changes: (1) `path.split('/').pop()` → `basename(path)` — more idiomatic and cross-platform correct; (2) `if (!cfg.auth) throw …` guard before assigning `authConfig` — prevents clobbering the live auth state with `undefined` when `agent.yaml` has no `auth` section. The throw-before-assign pattern is strictly safer than the previous code which would have set `authConfig = undefined` and then crashed mid-reload. One minor concern: `basename` must be imported from Node's `path` module; verify the import exists in the full file. No new logic regressions or security issues introduced.

### Confidence: 7.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 6
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Confirmed — basename is already imported at line 23: import { basename, dirname, join } from 'path'. No issue.

packages/daemon/src/daemon.ts
aaf2tbz: Addressed in d97db941 — added explicit guard: if (!cfg.auth.rateLimits) throw new Error('Missing rateLimits in auth config') before accessing rl.forget/modify/etc. Both auth and rateLimits are now validated before any assignment.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: settings persist across refresh without daemon restart` (commit `d97db941`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
Incremental diff adds a single guard: `if (!cfg.auth.rateLimits) throw` before the `authConfig` assignment. This prevents a silent TypeError crash — if `rateLimits` were `null`/`undefined`, `rl.forget` would blow up at runtime; now it throws early with a clear message and leaves the live limiters intact. The guard is correctly ordered after the `!cfg.auth` check (no null-deref risk), and the empty-object case (`rateLimits: {}`) remains truthy so all four ternary defaults still fire as intended. All three issues from the initial review remain resolved. No new bugs, logic regressions, or security issues introduced.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

aaf2tbz:
**Merge conflict analysis**: This PR touches two isolated sections of daemon.ts — the file watcher (line ~8837) and a new endpoint (line ~1729). Open PRs #213 and #214 modify different regions (lines ~1767, ~1896, ~9544+). Verified clean merge with latest main — no conflicts. The merge_conflict_detection score of 6 reflects the bot's caution about daemon.ts being a high-traffic file, not an actual conflict risk.

NicholaiVogel:
Clean and elegant implementation, merged.

PR #252, opened 2026-03-20
fix: use local timezone for timeline Today boundaries

aaf2tbz:
## Summary

Timeline "Today" view was using UTC day boundaries instead of the user's local timezone, causing events from the user's actual "today" to appear in the wrong bucket or not show up at all.

### Root cause

`startOfUtcDay()` calculated midnight UTC as the start of "today." For a user in MST (UTC-7), "Today" started at 5 PM yesterday local time. Memories created during the local morning/afternoon would fall outside the UTC "Today" window.

### Fix

- **Backend** (`memory-timeline.ts`): Renamed `startOfUtcDay` to `startOfLocalDay`, which accepts a `tzOffsetMin` parameter (minutes west of UTC, matching `Date.getTimezoneOffset()`). Day boundaries are now shifted to align with the caller's local midnight.
- **Daemon** (`daemon.ts`): `/api/memory/timeline` endpoint accepts an optional `?tzOffset=` query param and passes it through to `buildMemoryTimeline()`.
- **Frontend** (`api.ts`): `getMemoryTimeline()` sends `new Date().getTimezoneOffset()` with the API call. The client-side fallback also switched from `Date.UTC()` to local `new Date(year, month, date)`.

## Changes

| File | Change |
|------|--------|
| `memory-timeline.ts` | `startOfUtcDay` → `startOfLocalDay` with timezone offset support |
| `daemon.ts` | Parse `?tzOffset=` query param, pass to timeline builder |
| `api.ts` | Send `getTimezoneOffset()` with timeline fetch, fix fallback to local time |

## Test plan

- [ ] Open Timeline tab → "Today" should show memories/events created since local midnight
- [ ] Compare "Today" count with direct DB query filtered by local midnight
- [ ] Verify "Last week" and "One month" ranges also shift correctly
- [ ] Test with different timezone offsets (e.g., UTC, EST, PST)
- [ ] Verify fallback (when daemon is unavailable) also uses local time

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -1729,7 +1729,7 @@ app.get("/api/memories", (c) => {
 app.get("/api/memory/timeline", (c) => {
 	try {
 		const raw = Number.parseInt(c.req.query("tzOffset") || "0", 10);
-		const tzOffsetMin = Number.isNaN(raw) ? 0 : raw;
+		const tzOffsetMin = Number.isNaN(raw) ? 0 : Math.max(-840, Math.min(840, raw));
 		const timeline = getDbAccessor().withReadDb((db) =>
 			buildMemoryTimeline(db, { tzOffsetMin }),
 		);


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: use local timezone for timeline Today boundaries` (commit `6c1382d6`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -1728,7 +1728,11 @@ app.get("/api/memories", (c) => {
 
 app.get("/api/memory/timeline", (c) => {
 	try {
-		const timeline = getDbAccessor().withReadDb((db) => buildMemoryTimeline(db));
+		const raw = Number.parseInt(c.req.query("tzOffset") || "0", 10);
```
NicholaiVogel, line 1731: **Missing bounds check on `tzOffset`.**

`parseInt` + `isNaN` guard handles non-numeric input, but any integer is accepted. Real timezone offsets fit in `[-840, 840]` minutes (UTC-14 to UTC+14). Values outside this range cause two problems:

1. **NaN propagation / RangeError**: For sufficiently large values, `ms - tzOffsetMin * 60_000` can underflow into a date before year 0000, causing `new Date(...).toISOString()` inside `buildMemoryTimeline` to throw `RangeError: Invalid time value`. The try/catch swallows this silently.
2. **Full table scan**: If the value is extreme-but-still-finite, `cutoffIso` resolves to a date in the distant past (e.g., year -1384), stripping the 30-day SQL filter and scanning the entire memories table.

Recommended fix:
```typescript
const tzOffsetMin = Number.isNaN(raw) ? 0 : Math.max(-840, Math.min(840, raw));
```

packages/daemon/src/memory-timeline.ts
```diff
@@ -103,9 +103,17 @@ interface MutableBucket {
...
-	const d = new Date(ms);
-	return Date.UTC(d.getUTCFullYear(), d.getUTCMonth(), d.getUTCDate());
+/**
+ * Start of day in the caller's timezone.
+ * @param ms - current timestamp
+ * @param tzOffsetMin - minutes west of UTC (same as Date.getTimezoneOffset(),
+ *   e.g. 420 for MST/UTC-7). Defaults to 0 (UTC).
+ */
+function startOfLocalDay(ms: number, tzOffsetMin = 0): number {
+	const local = ms - tzOffsetMin * 60_000;
```
NicholaiVogel, line 113: **DST transitions cause bucket boundaries to drift for historical ranges** (minor, informational).

`tzOffsetMin` reflects the caller's *current* offset (e.g., MDT = 360 in summer, MST = 420 in winter). The same offset is applied uniformly to all 30 range days, so the `last_week` and `one_month` bucket starts are off by one hour on any day that crosses a DST boundary. A memory created at 12:15 AM on a spring-forward night is silently misclassified.

This is an acceptable approximation for a dashboard timeline and is no worse than the previous UTC-only behavior for users not in UTC. No change needed unless precise per-day DST accuracy is a stated requirement.

NicholaiVogel (commented):
The core timezone math in `startOfLocalDay` is correct: subtract the offset to get local time, find UTC midnight of that shifted date, then add the offset back to get the local midnight expressed in UTC. The client-side fallback switch from `Date.UTC()` to `new Date(year, month, date)` is also correct. One medium-severity issue: the `tzOffset` query parameter is parsed but never clamped to a valid range, allowing a crafted value to push `nowStartMs` to NaN or a date outside JavaScript's safe range. When that happens, `new Date(NaN).toISOString()` throws a RangeError inside `buildMemoryTimeline`; the daemon's `try/catch` absorbs it gracefully, but the bigger risk is that if the value is extreme-but-still-finite, the SQL cutoff ISO string could be garbage (e.g., a year in the distant past), turning the intended 30-day filtered query into a full table scan. Clamping to `[-840, 840]` (UTC-14 to UTC+14) before use would close this.

### Confidence: 7.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 7,
  "functional_completeness": 8,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in bd91b660 — tzOffset is now clamped to `[-840, 840]` before use, exactly as suggested.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: use local timezone for timeline Today boundaries` (commit `bd91b660`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
NicholaiVogel, line 1731: **Fixed (prior review).** The clamp `Math.max(-840, Math.min(840, raw))` correctly bounds the value to the full real-world timezone range (UTC-14 → -840 min, UTC-12 → +720 min). NaN propagation, RangeError from `toISOString()`, and the accidental full-table-scan via a garbage SQL cutoff are all closed.

packages/daemon/src/memory-timeline.ts
NicholaiVogel, line 113: Minor note for future hardening: `startOfLocalDay` is correct for fixed-offset timezones, but the offset snapshot is taken once on the client at request time. If a DST transition occurs exactly at midnight local time (rare but possible), the reported boundaries could be off by one hour for that single call. This is a pre-existing theoretical edge case and not a regression introduced by this PR; documenting it here for awareness.

PR #255, opened 2026-03-20
feat: bring Codex connector to parity with Claude Code / OpenCode

aaf2tbz:
## Why

Codex CLI currently only gets memories at session-start (frontloaded) and extracts them at session-end (after-carry). Unlike Claude Code — which gets per-prompt memory injection on every user message via native hooks — Codex has zero mid-session recall. Users running Codex sessions don't get the live memory experience that makes Signet valuable during conversations.

The Codex connector was also missing basic parity features: no skills symlink, no persistent identity file, no config.toml integration.

## How

**The core constraint:** Codex has no native mid-session hooks, no MCP, no IPC, no stdin. The `model_instructions_file` is read once at startup. We cannot directly inject context into a running Codex process.

**The workaround:** Codex CAN read files from disk during a session (it's a coding agent with file access). We instruct it to check a live-updating file before each response, and the daemon keeps that file current with relevant memories via real-time transcript watching.

### Layer 1: Connector Parity (reliable baseline)
- Generate persistent `~/.codex/CODEX.md` from identity files (AGENTS.md, SOUL.md, IDENTITY.md, USER.md, MEMORY.md) with Signet system block
- Patch `~/.codex/config.toml` with `model_instructions_file` pointing to CODEX.md — Codex always loads Signet context even without the wrapper
- Symlink skills directory (`~/.agents/skills` → `~/.codex/skills`)
- Enriched wrapper merges static CODEX.md + dynamic session-start memories into one file per session
- CLI memory commands (`signet remember`/`signet recall`) documented in CODEX.md since Codex has no MCP
- Safe config.toml handling: never overwrites user-set `model_instructions_file`, emits warning instead

### Layer 2: Live Per-Prompt Recall (best-effort, ~70-80% of turns)

```
User types message in Codex
  → Codex writes event to transcript JSONL (real-time)
  → Daemon transcript watcher detects new user_message (chokidar, 300ms debounce)
  → Daemon calls handleUserPromptSubmit internally (hybrid recall)
  → Daemon writes recalled memories to ~/.codex/.signet-live-context.md (atomic write-rename)
  → Codex (following model_instructions_file directive) reads the file before responding
```

- New `codex-transcript-watcher.ts` daemon module — incremental JSONL parsing, per-file offset tracking, debounced recall, atomic file writes, full session lifecycle management
- New endpoint: `POST /api/hooks/codex-watch-start` for session registration
- New CLI command: `signet hook codex-watch-start`
- Only activates for interactive `codex` sessions (skips `codex exec` one-shot mode)
- Enabled by default, fail-open on all error paths — if daemon is down or recall fails, Codex continues normally
- ~500ms total latency (300ms debounce + 200ms recall) vs Codex's 1-3s API round-trip

### What remains infeasible (needs upstream Codex changes)
- **Guaranteed per-prompt injection**: Codex has no native hook system — the file-read instruction is probabilistic
- **MCP server**: Codex doesn't support MCP
- **Pre-compaction**: No mid-session hooks

## Changes

| File | Change |
|------|--------|
| `packages/daemon/src/codex-transcript-watcher.ts` | **New** — Core watcher module (transcript parsing, recall, live file management) |
| `packages/daemon/src/daemon.ts` | Import watcher, add `/api/hooks/codex-watch-start` route, wire startup/shutdown lifecycle, add `unregisterCodexSession` to session-end cleanup |
| `packages/connector-codex/src/index.ts` | CODEX.md generation, config.toml patching, skills symlink, live recall instruction in wrapper, watch-start registration for interactive mode |
| `packages/cli/src/cli.ts` | `signet hook codex-watch-start` subcommand |
| `docs/HARNESSES.md` | Updated Codex section with new capabilities, live recall docs, and limitations |

## Test plan
- [ ] `biome check` passes on changed files
- [ ] `signet install` → select Codex → verify CODEX.md, skills symlink, config.toml patch
- [ ] Run interactive Codex → verify `~/.codex/.signet-live-context.md` updates on each user message
- [ ] Run `codex exec "one shot"` → verify watch-start is NOT called
- [ ] End session → verify live context file is cleaned up
- [ ] Pre-existing `model_instructions_file` in config.toml → verify warning, no overwrite
- [ ] Uninstall removes CODEX.md and unpatches config.toml

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -6121,6 +6121,44 @@ hookCmd
 		}
 	});
 
+// signet hook codex-watch-start — register Codex session for live recall
+hookCmd
+	.command("codex-watch-start")
+	.description("Start live recall watcher for a Codex session")
+	.requiredOption("-H, --harness <harness>", "Harness name")
+	.action(async () => {
+		let body: Record<string, string> = {};
+		try {
+			const chunks: Buffer[] = [];
+			for await (const chunk of process.stdin) {
+				chunks.push(chunk);
+			}
+			const input = Buffer.concat(chunks).toString("utf-8").trim();
+			if (input) {
+				body = JSON.parse(input);
+			}
+		} catch {
+			// No stdin or invalid JSON
+		}
+
+		const data = await fetchFromDaemon<{
+			success?: boolean;
+			error?: string;
+		}>("/api/hooks/codex-watch-start", {
+			method: "POST",
+			headers: { "Content-Type": "application/json" },
+			body: JSON.stringify({
+				sessionKey: body.sessionKey || body.session_key,
+				project: body.project || body.cwd,
+			}),
+		});
+
+		if (!data) {
+			// Daemon not running — fail silently (fire-and-forget from wrapper)
+			process.exit(0);
+		}
+	});
+
 // signet hook pre-compaction
 hookCmd
 	.command("pre-compaction")
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -6125,7 +6125,7 @@ hookCmd
 hookCmd
 	.command("codex-watch-start")
 	.description("Start live recall watcher for a Codex session")
-	.requiredOption("-H, --harness <harness>", "Harness name")
+	.option("-H, --harness <harness>", "Harness name")
 	.action(async () => {
 		let body: Record<string, string> = {};
 		try {
--- a/packages/connector-codex/src/index.ts
+++ b/packages/connector-codex/src/index.ts
@@ -1,18 +1,8 @@
 import { spawnSync } from "node:child_process";
-import {
-	existsSync,
-	mkdirSync,
-	readFileSync,
-	rmSync,
-	writeFileSync,
-} from "node:fs";
+import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
-import {
-	BaseConnector,
-	type InstallResult,
-	type UninstallResult,
-} from "@signet/connector-base";
+import { BaseConnector, type InstallResult, type UninstallResult } from "@signet/connector-base";
 
 const SHELL_BLOCK_START = "# >>> signet codex >>>";
 const SHELL_BLOCK_END = "# <<< signet codex <<<";
@@ -26,6 +16,8 @@ fi
 ${SHELL_BLOCK_END}
 `;
 
+const TOML_COMMENT = "# Added by Signet — points to generated identity file";
+
 function stripBlock(content: string): string {
 	const start = content.indexOf(SHELL_BLOCK_START);
 	if (start < 0) return content;
@@ -33,7 +25,7 @@ function stripBlock(content: string): string {
 	if (end < 0) return content;
 	const after = content.slice(end + SHELL_BLOCK_END.length);
 	const before = content.slice(0, start).trimEnd();
-	return `${before}${before ? "\n\n" : ""}${after.trimStart()}`.trimEnd() + "\n";
+	return `${`${before}${before ? "\n\n" : ""}${after.trimStart()}`.trimEnd()}\n`;
 }
 
 function appendBlock(content: string): string {
@@ -73,13 +65,24 @@ session_start() {
 	: > "$START_MARKER"
 	SESSION_KEY="$(uuidgen 2>/dev/null || printf "codex-%s-%s" "$(date +%s)" "$$")"
 
+	# Start with persistent identity context if available
+	CODEX_MD="\${HOME}/.codex/CODEX.md"
+	if [ -f "$CODEX_MD" ]; then
+		cp "$CODEX_MD" "$INSTRUCTIONS_FILE"
+		printf '\\n\\n---\\n\\n' >> "$INSTRUCTIONS_FILE"
+	fi
+
+	# Append dynamic session context (live memories, recall results)
+	DYNAMIC="$TMP_ROOT/dynamic.md"
 	payload="$(printf '{"session_id":"%s","cwd":"%s"}' "$(json_escape "$SESSION_KEY")" "$(json_escape "$PWD")")"
-	if printf "%s" "$payload" | "$SIGNET_BIN" hook session-start -H codex --project "$PWD" > "$INSTRUCTIONS_FILE" 2>/dev/null; then
-		if [ ! -s "$INSTRUCTIONS_FILE" ]; then
-			rm -f "$INSTRUCTIONS_FILE"
-			INSTRUCTIONS_FILE=""
+	if printf "%s" "$payload" | "$SIGNET_BIN" hook session-start -H codex --project "$PWD" > "$DYNAMIC" 2>/dev/null; then
+		if [ -s "$DYNAMIC" ]; then
+			cat "$DYNAMIC" >> "$INSTRUCTIONS_FILE"
 		fi
-	else
+	fi
+
+	# If nothing was written at all, clear
+	if [ ! -s "$INSTRUCTIONS_FILE" ]; then
 		rm -f "$INSTRUCTIONS_FILE"
 		INSTRUCTIONS_FILE=""
 	fi
@@ -172,16 +175,27 @@ set "TMP_ROOT=%TEMP%\\signet-codex-%RANDOM%"
 mkdir "%TMP_ROOT%" >nul 2>&1
 set "INSTRUCTIONS_FILE=%TMP_ROOT%\\model-instructions.md"
 
-REM Session-start hook: build JSON via ConvertTo-Json for safe escaping of paths containing ", &, |, %
+REM Start with persistent identity context if available
+set "CODEX_MD=%USERPROFILE%\\.codex\\CODEX.md"
+if exist "%CODEX_MD%" (
+	copy /y "%CODEX_MD%" "%INSTRUCTIONS_FILE%" >nul 2>&1
+	echo. >> "%INSTRUCTIONS_FILE%"
+	echo --- >> "%INSTRUCTIONS_FILE%"
+	echo. >> "%INSTRUCTIONS_FILE%"
+)
+
+REM Session-start hook: append dynamic context
+set "DYNAMIC=%TMP_ROOT%\\dynamic.md"
 set "HOOK_OK=0"
 for /f "delims=" %%j in ('powershell ${psFlags} -Command "ConvertTo-Json @{session_id=$env:SESSION_KEY;cwd=$PWD.Path} -Compress"') do (
-	echo %%j| "%SIGNET_BIN%" hook session-start -H codex --project "%CD%" > "%INSTRUCTIONS_FILE%" 2>nul && set "HOOK_OK=1"
+	echo %%j| "%SIGNET_BIN%" hook session-start -H codex --project "%CD%" > "%DYNAMIC%" 2>nul && set "HOOK_OK=1"
 )
-if "%HOOK_OK%"=="0" (
-	set "INSTRUCTIONS_FILE="
-) else (
-	for %%A in ("%INSTRUCTIONS_FILE%") do if %%~zA==0 set "INSTRUCTIONS_FILE="
+if "%HOOK_OK%"=="1" (
+	for %%A in ("%DYNAMIC%") do if not %%~zA==0 type "%DYNAMIC%" >> "%INSTRUCTIONS_FILE%"
 )
+
+REM If nothing was written, clear instructions file
+for %%A in ("%INSTRUCTIONS_FILE%") do if %%~zA==0 set "INSTRUCTIONS_FILE="
 
 if defined INSTRUCTIONS_FILE (
 	"%REAL_CODEX_BIN%" -c "model_instructions_file=%INSTRUCTIONS_FILE%" %*
@@ -220,6 +234,10 @@ export class CodexConnector extends BaseConnector {
 		return join(homedir(), ".codex");
 	}
 
+	private getCodexMdPath(): string {
+		return join(this.getCodexHome(), "CODEX.md");
+	}
+
 	private getWrapperDir(): string {
 		return join(homedir(), ".config", "signet", "bin");
 	}
@@ -231,11 +249,7 @@ export class CodexConnector extends BaseConnector {
 
 	private getShellConfigPaths(): string[] {
 		if (process.platform === "win32") return [];
-		return [
-			join(homedir(), ".zshrc"),
-			join(homedir(), ".bashrc"),
-			join(homedir(), ".bash_profile"),
-		];
+		return [join(homedir(), ".zshrc"), join(homedir(), ".bashrc"), join(homedir(), ".bash_profile")];
 	}
 
 	getConfigPath(): string {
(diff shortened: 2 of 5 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: bring Codex connector to parity with Claude Code / OpenCode` (commit `e1713110`) now and I'll follow up shortly with feedback.

packages/connector-codex/src/index.ts
```diff
@@ -263,26 +277,156 @@ export class CodexConnector extends BaseConnector {
...
+
+		// Find existing model_instructions_file (only at top level, before any [section])
+		let found = false;
+		let alreadyCorrect = false;
+		let userOwned = false;
+
+		for (const line of lines) {
+			const trimmed = line.trim();
+			// Stop scanning at first section header
+			if (trimmed.startsWith("[")) break;
```
NicholaiVogel, line 380: **Bug — TOML inline comments cause duplicate key insertion.**

The regex `/^model_instructions_file\s*=\s*"?([^"]*)"?$/` requires `$` immediately after the (optional) closing quote, so a value like:
```toml
model_instructions_file = "/home/user/.codex/CODEX.md" # set by hand
```
does **not** match. `found` remains `false`, and `patchConfigToml` inserts a second `model_instructions_file` line above the first one. TOML doesn't allow duplicate top-level keys; parsers either reject the file or silently use one of the two values, potentially breaking Codex on the next run.

Fix — strip inline comments before matching:
```typescript
const noComment = trimmed.replace(/\s*#.*$/, "");
const match = noComment.match(/^model_instructions_file\s*=\s*"?([^"]*)"?$/);
```
Same fix should be applied in `unpatchConfigToml` for consistency.

packages/connector-codex/src/index.ts
```diff
@@ -292,24 +436,47 @@ export class CodexConnector extends BaseConnector {
...
+		const targetSkills = join(this.getCodexHome(), "skills");
+		this.symlinkSkills(sourceSkills, targetSkills);
+
+		// 6. Patch config.toml with model_instructions_file
+		const toml = this.patchConfigToml();
+		if (toml.patched) {
+			configsPatched.push(this.getConfigPath());
+		}
+		if (toml.warning) {
+			warnings.push(toml.warning);
```
NicholaiVogel, line 505: **Silent failure — `symlinkSkills` result is not checked.**

```typescript
this.symlinkSkills(sourceSkills, targetSkills);
```

If the symlink creation fails (e.g. `~/.agents/skills` doesn't exist, permission error, or the target already exists as a regular directory), there is no warning in the `InstallResult`. Users would believe the skills directory was set up but the Codex agent would not see any skills. Either propagate errors or push a warning to `warnings[]` when this call fails/returns false.

packages/connector-codex/src/index.ts
```diff
@@ -263,26 +277,156 @@ export class CodexConnector extends BaseConnector {
...
+			const entry = `${TOML_COMMENT}\nmodel_instructions_file = "${codexMd}"`;
+			lines.splice(insertIdx, 0, entry);
+			writeFileSync(toml, lines.join("\n"));
+			return { patched: true };
+		}
+
+		return { patched: false };
+	}
+
+	/**
```
NicholaiVogel, line 426: **Dead code / misleading control flow.**

The final `return { patched: false }` after the `if (!found)` block is unreachable: `found = true` is only ever set in the same branch that also sets either `alreadyCorrect` or `userOwned`, both of which trigger early returns before this point. Consider removing it or replacing with an `assertNever`-style guard so the compiler can verify exhaustiveness. Not a runtime bug, but it obscures the actual logic.

packages/connector-codex/src/index.ts
```diff
@@ -172,17 +175,28 @@ set "TMP_ROOT=%TEMP%\\signet-codex-%RANDOM%"
 mkdir "%TMP_ROOT%" >nul 2>&1
```
NicholaiVogel, line 198: **Windows separator differs from Unix wrapper.**

Unix uses `printf '\\n\\n---\\n\\n'` (two blank lines around `---`). The Windows wrapper emits:
```cmd
echo. >> "%INSTRUCTIONS_FILE%"
echo --- >> "%INSTRUCTIONS_FILE%"
echo. >> "%INSTRUCTIONS_FILE%"
```
which is one blank line before and one after. The functional impact is minimal (it's just a visual separator in the instructions file), but keeping the two wrappers consistent avoids confusing diffs when the format is later parsed or logged.

NicholaiVogel (changes requested):
One real bug: the TOML regex does not handle inline comments, so a user with `model_instructions_file = "/path" # comment` is not detected as user-owned and Signet will append a second `model_instructions_file` key — producing invalid / ambiguous TOML that can silently break Codex. Also worth fixing: `symlinkSkills` return value is not checked, so a symlink failure is silent. Everything else (wrapper merge logic, uninstall cleanup, Windows parity, warning plumbing) looks correct.

### Confidence: 7.7/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 8,
  "pattern_correctness": 7,
  "documentation_coverage": 9
}
```

aaf2tbz:
Currently a critical fix has been found regarding codex support and it will take some time. 
In progress 👀

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat: bring Codex connector to parity with Claude Code / OpenCode` (commit `ee53e9d7`); taking another look at the feature work and I will report back shortly.

packages/connector-codex/src/index.ts
```diff
@@ -263,26 +317,156 @@ export class CodexConnector extends BaseConnector {
...
+		const codexMd = join(this.getCodexHome(), "CODEX.md");
+		mkdirSync(this.getCodexHome(), { recursive: true });
+		writeFileSync(codexMd, header + block + content + extras + cliHint);
+		return codexMd;
+	}
+
+	/**
+	 * Patch ~/.codex/config.toml to set model_instructions_file.
+	 *
+	 * Safe behavior:
```
NicholaiVogel, line 358: **Unfixed from prior review — TOML inline comment bug still present.**

The regex `/^model_instructions_file\s*=\s*"?([^"]*)"?$/` requires `$` immediately after the optional closing quote. A legal TOML line with a trailing comment — `model_instructions_file = "/path" # comment` — won't match because `# comment` prevents the `$` from anchoring. `found` stays false, the code falls through to the `!found` branch, and appends a second `model_instructions_file` key. Codex (and any TOML parser that errors on duplicate top-level keys) may reject or silently misparse the file.

Fix: strip everything after an unquoted `#` before matching, or use a more permissive regex that captures through end-of-value:
```ts
const match = trimmed.match(/^model_instructions_file\s*=\s*"([^"]*)"|^model_instructions_file\s*=\s*([^#\s][^#]*)/);
```
or simply strip inline comments first:
```ts
const stripped = trimmed.replace(/#.*$/, '').trim();
const match = stripped.match(/^model_instructions_file\s*=\s*"?([^"]*?)"?$/);
```

packages/connector-codex/src/index.ts
```diff
@@ -292,24 +476,47 @@ export class CodexConnector extends BaseConnector {
 			}
```
NicholaiVogel, line 481: **Unfixed from prior review — `symlinkSkills` return value still discarded.**

`this.symlinkSkills(sourceSkills, targetSkills)` is fire-and-forget. If the symlink fails (target exists as a real dir, permission error, etc.), the install reports success with no indication that skills were not wired up. At minimum, capture the return value and push a warning into `warnings[]` so the user knows.

packages/daemon/src/codex-transcript-watcher.ts
```diff
@@ -0,0 +1,453 @@
...
+			error: e instanceof Error ? e.message : String(e),
+			sessionKey: session.sessionKey,
+		});
+	}
+}
+
+function simpleHash(s: string): string {
+	let h = 0;
+	for (let i = 0; i < s.length; i++) {
+		h = ((h << 5) - h + s.charCodeAt(i)) | 0;
```
NicholaiVogel, line 201: **`maxResults` and `charBudget` config fields are declared, documented in HARNESSES.md, and validated in `resolveConfig`, but are never forwarded to `handleUserPromptSubmit`.**

`recallForMessage` receives `config` but only uses it for the dedup hash guard — `config.maxResults` and `config.charBudget` are completely ignored when calling `handleUserPromptSubmit(req)`. Users who set these in `agent.yaml` will see no effect and no error. Either pass these fields through to `handleUserPromptSubmit` (if it accepts them) or remove the fields from the public config interface and docs to avoid the false promise.

packages/daemon/src/codex-transcript-watcher.ts
```diff
@@ -0,0 +1,453 @@
...
+	// Find which session this transcript belongs to
+	let session: TrackedSession | null = null;
+	for (const [, s] of sessions) {
+		if (s.transcriptPath === path) {
+			session = s;
+			break;
+		}
+	}
+
+	if (!session) {
```
NicholaiVogel, line 262: **`discoverTranscript` uses today's wall-clock date, so sessions spanning midnight lose their transcript.**

If `registerCodexSession` is called at 23:59 and `discoverTranscript` runs at 00:01, `todayDir` becomes `sessions/YYYY/MM/DD+1` while the actual file is under `sessions/YYYY/MM/DD`. The session will never associate with its transcript and live recall silently stops.

Simplest fix: when registering a session, snapshot `sessionsDir()` entries at that moment and pick the newest, or use chokidar's `add` event (already wired up) to let the OS notify of the new file rather than relying on date-path inference.

packages/connector-codex/src/index.ts
```diff
@@ -73,13 +65,41 @@ session_start() {
...
+	DYNAMIC="$TMP_ROOT/dynamic.md"
 	payload="$(printf '{"session_id":"%s","cwd":"%s"}' "$(json_escape "$SESSION_KEY")" "$(json_escape "$PWD")")"
-	if printf "%s" "$payload" | "$SIGNET_BIN" hook session-start -H codex --project "$PWD" > "$INSTRUCTIONS_FILE" 2>/dev/null; then
-		if [ ! -s "$INSTRUCTIONS_FILE" ]; then
-			rm -f "$INSTRUCTIONS_FILE"
-			INSTRUCTIONS_FILE=""
+	if printf "%s" "$payload" | "$SIGNET_BIN" hook session-start -H codex --project "$PWD" > "$DYNAMIC" 2>/dev/null; then
+		if [ -s "$DYNAMIC" ]; then
+			cat "$DYNAMIC" >> "$INSTRUCTIONS_FILE"
 		fi
```
NicholaiVogel, line 81: **The 'clear if empty' guard is now dead code — Codex always gets at least a stub instructions file.**

The LIVE_RECALL heredoc is appended unconditionally whenever `INSTRUCTIONS_FILE` is non-empty (which is always true at that point — it's set to `$TMP_ROOT/model-instructions.md` at the top of `session_start`). After the heredoc append, `INSTRUCTIONS_FILE` always has content, so `[ ! -s "$INSTRUCTIONS_FILE" ]` is always false. The cleanup branch never fires.

Consequence: when the daemon is down *and* `~/.codex/CODEX.md` doesn't exist, Codex is launched with an instructions file containing only the live recall prompt — which tells it to read a file that will never be updated. This is at worst confusing, but the dead guard means the fallback intent (run Codex without any instructions) is now unreachable. If that fallback matters, move the LIVE_RECALL block inside a `[ -s "$INSTRUCTIONS_FILE" ]` check so it only appends when there is real content to accompany it.

packages/cli/src/cli.ts
```diff
@@ -6121,6 +6121,44 @@ hookCmd
 		}
 	});
 
+// signet hook codex-watch-start — register Codex session for live recall
+hookCmd
+	.command("codex-watch-start")
+	.description("Start live recall watcher for a Codex session")
```
NicholaiVogel, line 6127: **`--harness` is a `requiredOption` but is never read.**

The `.action(async () => { ... })` callback ignores the `options` argument entirely — the harness value is never validated or forwarded. Since the daemon endpoint doesn't accept a harness field either, this is dead friction. If the option exists to be consistent with other hook subcommands, it should at least be optional (`option` not `requiredOption`), otherwise callers without a `-H codex` flag get a hard CLI error for no functional benefit.

NicholaiVogel (changes requested):
Neither issue from the prior review was fixed: (1) the TOML inline-comment bug remains — a line like `model_instructions_file = "/path" # comment` still won't match the regex, so `found` stays false and Signet appends a second key producing ambiguous TOML; (2) `symlinkSkills` return value is still unchecked. New issues in this revision: `maxResults` and `charBudget` config fields are declared, documented, and exposed in agent.yaml, but `recallForMessage` never passes them to `handleUserPromptSubmit` — users who tune these get no effect. Additionally, `discoverTranscript` hardcodes today's calendar date, so sessions that start just before midnight will fail to associate with their transcript after midnight. The live-recall heredoc append also makes the 'clear if empty' guard dead code, causing Codex to always be launched with at least a stub instructions file even when the daemon is down and no identity file exists — harmless but misleading.

### Confidence: 7.4/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 6,
  "pattern_correctness": 7,
  "documentation_coverage": 8
}
```

aaf2tbz:
## Addressed in 662ab32

All feedback from both review rounds addressed:

1. **TOML inline comment bug** — Fixed. Both `patchConfigToml` and `unpatchConfigToml` now strip `# comment` before regex matching. `model_instructions_file = "/path" # comment` is correctly detected as user-owned.

2. **`symlinkSkills` return unchecked** — Fixed. Return value is now captured and errors are pushed to `InstallResult.warnings[]`.

3. **`maxResults`/`charBudget` config fields never forwarded** — Removed from the public config interface and docs. `handleUserPromptSubmit` uses its own internal budgets; exposing config fields that had no effect was a false promise.

4. **Midnight transcript discovery bug** — Fixed. Replaced date-path inference (`YYYY/MM/DD`) with chokidar `add` event binding. New JSONL files are associated with the most recently registered unbound session, regardless of date directory.

5. **Dead code in wrapper (LIVE_RECALL heredoc)** — Fixed. Moved the heredoc append after the empty-file check, so Codex can still fall back to no instructions when daemon is down and no identity file exists.

6. **`--harness` required but unused in CLI** — Changed to optional.

7. **Windows separator parity** — Fixed. Now two blank lines around `---`, matching Unix wrapper.

8. **Unreachable return in `patchConfigToml`** — Added clarifying comment.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat: bring Codex connector to parity with Claude Code / OpenCode` (commit `662ab32e`); taking another look at the feature work and I will report back shortly.

packages/connector-codex/src/index.ts
```diff
@@ -172,15 +198,43 @@ set "TMP_ROOT=%TEMP%\\signet-codex-%RANDOM%"
...
+if defined INSTRUCTIONS_FILE (
+	echo. >> "%INSTRUCTIONS_FILE%"
+	echo ## Live Memory ^(Auto-Updated by Signet^) >> "%INSTRUCTIONS_FILE%"
+	echo. >> "%INSTRUCTIONS_FILE%"
+	echo IMPORTANT: Before responding to each user message, read the file at >> "%INSTRUCTIONS_FILE%"
+	echo ~/.codex/.signet-live-context.md for memories relevant to the current >> "%INSTRUCTIONS_FILE%"
+	echo conversation. Incorporate relevant memories into your response. >> "%INSTRUCTIONS_FILE%"
+)
+
+REM If nothing was written, clear instructions file
```
NicholaiVogel: **Windows wrapper: live-recall stub appended before empty-check (same bug that was fixed for Unix in this PR).**

The Windows batch script appends the live-recall instruction block unconditionally (while `INSTRUCTIONS_FILE` is still defined), then does the zero-size check:

```batch
REM Append live recall instruction
if defined INSTRUCTIONS_FILE (
    echo. >> "%INSTRUCTIONS_FILE%"
    echo ## Live Memory ... >> "%INSTRUCTIONS_FILE%"
    ...
)

REM If nothing was written, clear instructions file
for %%A in ("%INSTRUCTIONS_FILE%") do if %%~zA==0 set "INSTRUCTIONS_FILE="
```

After the `echo` lines run the file is always non-empty, so `%%~zA==0` is always false and `INSTRUCTIONS_FILE` is never cleared. When CODEX.md is absent and the daemon is down, Codex will be launched with `-c model_instructions_file=...` pointing to a file containing only the live-recall stub — the exact misbehavior this PR fixed on Unix.

The fix is identical to the Unix fix: move the empty-check **before** the live-recall append:

```batch
REM If nothing was written, clear instructions file
for %%A in ("%INSTRUCTIONS_FILE%") do if %%~zA==0 set "INSTRUCTIONS_FILE="

REM Append live recall instruction only when there is real content
if defined INSTRUCTIONS_FILE (
    echo. >> "%INSTRUCTIONS_FILE%"
    echo ## Live Memory ^(Auto-Updated by Signet^) >> "%INSTRUCTIONS_FILE%"
    ...
)
```

packages/daemon/src/codex-transcript-watcher.ts
```diff
@@ -0,0 +1,406 @@
...
+// ---------------------------------------------------------------------------
+// Transcript file discovery
+// ---------------------------------------------------------------------------
+
+/**
+ * Try to associate a newly appeared JSONL file with an unbound session.
+ * Uses chokidar's `add` event rather than date-path inference, avoiding
+ * the midnight rollover bug where date directories change between
+ * session registration and transcript creation.
+ */
```
NicholaiVogel, line 218: **`tryBindTranscript` can silently cross-wire two concurrent sessions.**

The function binds the first unbound session (Map insertion order) to whatever file just appeared. If a user opens two Codex windows close together, both sessions are unbound when the first transcript file appears; the second session's transcript will be bound to the first session's registration. The result is that session A sees recall context derived from session B's messages and vice-versa, with no error or warning.

This is acknowledged as best-effort, but a timestamp-based heuristic (bind the session registered most recently before the file appeared) would reduce mismatches at low cost. At minimum, log a warning when more than one unbound session exists at bind time so users can diagnose unexpected recall results.

packages/daemon/src/codex-transcript-watcher.ts
```diff
@@ -0,0 +1,406 @@
...
+	} catch {
+		return [];
+	}
+
+	if (size <= session.lastOffset) return [];
+
+	let content: string;
+	try {
+		// Read full file and slice from last offset. The transcript is
+		// append-only so previously parsed bytes never change.
```
NicholaiVogel: **`lastOffset` mixes character-count and byte-count units.**

`session.lastOffset` is set to `content.length` (UTF-16 code units), but the early-exit guard compares it with `statSync(path).size` (bytes). For JSONL transcripts containing multi-byte UTF-8 characters (emoji, CJK, etc.), `size` (bytes) > `content.length` (code units), so the guard `if (size <= session.lastOffset)` is always false — every watcher event triggers a full file re-read even when nothing changed.

Fix: store the byte length (use `Buffer.byteLength(content, 'utf-8')` or track the raw buffer length) so both sides of the comparison use the same unit. The slice itself is fine using char offsets internally as long as `lastOffset` is consistently character-based and the guard uses the same unit.

NicholaiVogel (changes requested):
All three issues from the previous review are fixed: the TOML inline-comment regex now strips trailing comments before matching (both patch and unpatch paths), `symlinkSkills` errors are now surfaced as warnings, and `maxResults`/`charBudget` are removed rather than left un-wired. The midnight-rollover bug is cleanly fixed by replacing `discoverTranscript` with `tryBindTranscript`. One real bug remains: the Windows batch wrapper appends the live-recall stub **before** the zero-size check, so on Windows the empty guard can never fire — Codex will always launch with a `model_instructions_file` containing only the live-recall heredoc even when the daemon is down and no CODEX.md exists. The Unix wrapper had the same ordering problem and was fixed in this PR; Windows was missed.

### Confidence: 7.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 7,
  "documentation_coverage": 8
}
```

aaf2tbz:
## Addressed in 6f5286cd

All three issues from round 3:

1. **Windows wrapper live-recall ordering** — Fixed. Empty-check now runs before live-recall append, matching the Unix wrapper fix from the prior commit. Codex on Windows can now correctly fall back to no instructions when daemon is down.

2. **lastOffset byte/char unit mismatch** — Fixed. Now uses `Buffer.length` (bytes) instead of `string.length` (UTF-16 code units) for offset tracking. The early-exit guard `if (size <= session.lastOffset)` now correctly compares bytes-to-bytes, preventing unnecessary full re-reads on transcripts with multi-byte characters (emoji, CJK, etc.).

3. **Multi-session transcript binding** — Improved. `tryBindTranscript` now iterates all unbound sessions and binds the most recently registered one (last in Map insertion order). When multiple unbound sessions exist at bind time, logs a warning with the count and which session was bound for diagnostics.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat: bring Codex connector to parity with Claude Code / OpenCode` (commit `6f5286cd`); taking another look at the feature work and I will report back shortly.

NicholaiVogel (approved):
All previously identified issues are now fixed. The Windows batch wrapper ordering bug (live-recall stub appended before the zero-size guard) is corrected — the zero-size check now precedes the stub append, so Windows correctly skips launching Codex with a stale instructions file when no content exists. A bonus correctness fix converts the JSONL reader from string-based slicing (UTF-16 code units, mismatched against statSync's byte count) to Buffer-based slicing, preventing offset drift on multibyte UTF-8 input. tryBindTranscript now counts all unbound sessions and warns when multiple exist, and correctly binds to the most recently registered one (last Map insertion) rather than the first — a better heuristic since the newest registration is most likely the session awaiting a transcript. No new issues found.

### Confidence: 8.0/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

NicholaiVogel:
@aaf2tbz unfortunately this doesn't follow the taxonomy imposed by the project.

> The workaround: Codex CAN read files from disk during a session (it's a coding agent with file access). We instruct it to check a live-updating file before each response, and the daemon keeps that file current with relevant memories via real-time transcript watching.

Relying on the on the Agent-in-the-loop to read from it's memory is unreliable. This is the core reason for the hooks themselves in other connectors.

NicholaiVogel:
## Closing — premise is incorrect

The core constraint stated in this PR is wrong, and the entire workaround (transcript watcher, file-polling, probabilistic model instruction) is built on it. Before merging something this complex, I verified the claim against the actual Codex source.

**The claim:** *"Codex has no native mid-session hooks, no MCP, no IPC, no stdin. The `model_instructions_file` is read once at startup. We cannot directly inject context into a running Codex process."*

Every part of that is false for current Codex (the Rust implementation, which replaced the legacy TS CLI months ago).

---

### 1. Codex has native mid-session hooks

`codex-rs/hooks/` is a complete hook engine. Events fired during a live session:

- [`UserPromptSubmit`](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/hooks/src/events/user_prompt_submit.rs) — fires on **every user turn**. Hook output can inject `additional_context` directly into the model's context window, or block the turn entirely. This is exactly the per-prompt recall mechanism we'd want.
- [`AfterToolUse`](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/hooks/src/types.rs#L123) — fires after every tool call
- [`AfterAgent`](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/hooks/src/types.rs#L76) — fires after each agent turn
- [`SessionStart`](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/hooks/src/events/session_start.rs) and `Stop`

The `UserPromptSubmit` hook's `additional_context` output field is **a native context injection mechanism** — no file polling needed.

---

### 2. Codex has MCP

- [`codex-rs/mcp-server/`](https://github.com/openai/codex/tree/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/mcp-server) — standalone MCP server
- [`codex-rs/rmcp-client/`](https://github.com/openai/codex/tree/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/rmcp-client) — MCP client with OAuth, streamable HTTP, stdio transports
- [`codex-rs/core/src/mcp_connection_manager.rs`](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/core/src/mcp_connection_manager.rs) — live connection management

`signet remember` / `signet recall` can be exposed as proper MCP tools, not CLI shims documented in a markdown file.

---

### 3. Codex has IPC and stdin

The app-server transport enum has [two modes](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/app-server/src/transport.rs#L129-L132):

```rust
pub enum AppServerTransport {
    Stdio,
    WebSocket { bind_address: SocketAddr },
}
```

`stdio://` is the **default**. The server reads and writes JSON-RPC messages over stdin/stdout. There is also a WebSocket transport and Unix socket support.

---

### 4. `model_instructions_file` can be reloaded mid-session

`Op::ReloadUserConfig` is a [first-class protocol operation](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/protocol/src/protocol.rs#L439) that triggers `reload_user_config_layer()` — re-reading config (including `model_instructions_file`) without restarting the process. It's [triggered via config write RPC](https://github.com/openai/codex/blob/ec32866c379405a28b58c0064c857fb60ed3c735/codex-rs/app-server/src/config_api.rs#L38-L53), MCP tool install, or the TUI's reload command.

---

### What this means for the connector

The right implementation uses native primitives:

1. **`UserPromptSubmit` hook** → register a hook that calls the Signet daemon, gets recalled memories, and returns them as `additional_context`. Guaranteed injection, not probabilistic.
2. **MCP tools** → expose `signet_remember` and `signet_recall` as proper MCP tools via the daemon's MCP server.
3. **`Op::ReloadUserConfig`** → push config updates mid-session when needed.

The transcript watcher + file polling approach this PR implements is fragile, probabilistic, and unnecessary — it's a workaround for a constraint that doesn't exist. Closing in favor of a native implementation.

PR #258, opened 2026-03-21
feat(connector-codex): native hooks + MCP for full mid-session memory

aaf2tbz:
## Why

PR #255 was closed because it was built on a false premise — that Codex has no native mid-session hooks. The current Codex implementation (`codex-rs`) has a full hook engine with `UserPromptSubmit`, MCP support, IPC over stdin, and config reload. This PR uses those native primitives instead of the transcript watcher workaround.

## What Changed

**Removed** (entire old approach):
- Shell wrapper binary (`~/.config/signet/bin/codex`)
- PATH manipulation via shell config patching (`~/.zshrc`, `~/.bashrc`, `~/.bash_profile`)
- Temporary instruction files and `model_instructions_file` injection
- Session transcript discovery via `find` command

**Added** (native Codex integration):
- `~/.codex/hooks.json` with three native hooks:
  - `sessionStart` → `signet hook session-start -H codex` (identity + memories)
  - `userPromptSubmit` → `signet hook user-prompt-submit -H codex` (per-prompt recall)
  - `stop` → `signet hook session-end -H codex` (transcript extraction)
- MCP server registration in `~/.codex/config.toml` (`[mcp_servers.signet]`)
- Skills directory symlink (`~/.codex/skills` → `~/.agents/skills`)
- Updated HARNESSES.md documentation

## How It Works

Codex-rs reads `hooks.json` at startup and fires hooks on session lifecycle events. Each hook runs a shell command that calls the existing `signet hook` CLI. The CLI already accepts Codex's JSON stdin format — field mapping (`prompt` → `userMessage`, `cwd` → `project`, `sessionId` → `sessionKey`, `transcriptPath`) is handled automatically by the CLI's multi-format parser. No CLI changes needed.

**Key mechanism**: `UserPromptSubmit` hooks are **blocking** — Codex waits for the hook to complete before sending to the model. The hook's stdout is treated as `additional_context` and injected directly into the model's context window. This is **guaranteed per-prompt injection**, not probabilistic file polling.

The MCP server exposes `memory_store`, `memory_search`, `memory_list`, `memory_modify`, `memory_forget` as native Codex tools (namespaced as `mcp__signet__*`).

## Parity with Claude Code

| Feature | Before | After |
|---------|--------|-------|
| session-start injection | wrapper + temp file | native hook, `additional_context` |
| **per-prompt recall** | **none** | **native UserPromptSubmit hook** |
| session-end extraction | wrapper + transcript discovery | native Stop hook |
| MCP tools | none | native MCP server |
| Installation | wrapper + PATH + shell config patches | hooks.json + config.toml |
| Windows | .cmd wrapper + PowerShell | same hooks.json (native) |
| Lines of code | 340 | 260 |

## Edge Cases

- **Existing user hooks.json**: Signet merges its handlers into existing arrays, preserves user hooks, sets `_signet: true` marker
- **Uninstall with merged hooks**: only removes Signet-added entries, preserves user hooks
- **Windows**: resolves `node.exe` path for both hook commands and MCP server
- **Existing `[mcp_servers.signet]`**: skips, no overwrite
- **config.toml without comment marker**: unpatch still finds and removes the section
- **Daemon offline**: `signet hook` commands exit gracefully with stderr warning, Codex continues normally

## Test Plan

- [ ] `bun run build` passes
- [ ] `bun run typecheck` passes for connector-codex
- [ ] `signet install` → select Codex → verify `~/.codex/hooks.json` written with three hooks
- [ ] Verify `~/.codex/config.toml` has `[mcp_servers.signet]` entry
- [ ] Verify `~/.codex/skills` symlink exists
- [ ] Run interactive Codex session → verify session-start hook fires (memories injected)
- [ ] Type a prompt → verify user-prompt-submit hook fires (per-prompt recall)
- [ ] Exit session → verify stop hook fires (transcript extracted)
- [ ] Pre-existing hooks.json → verify merge preserves user hooks
- [ ] `signet uninstall` → verify hooks.json cleaned, config.toml unpatched
- [ ] Reinstall is idempotent (safe to run multiple times)

### DIFF
--- a/packages/connector-codex/src/index.ts
+++ b/packages/connector-codex/src/index.ts
@@ -304,8 +304,9 @@ export class CodexConnector extends BaseConnector {
 	isInstalled(): boolean {
 		const hooks = readHooksJson(this.getHooksJsonPath());
 		if (!hooks) return false;
-		// Check if any handler calls signet hook
-		const raw = JSON.stringify(hooks);
-		return raw.includes("hook session-start") || raw.includes("hook user-prompt-submit");
+		return ["sessionStart", "userPromptSubmit", "stop"].some(
+			(k) => Array.isArray((hooks as Record<string, unknown>)[k]) &&
+				((hooks as Record<string, unknown>)[k] as unknown[]).some(isSignetHandler),
+		);
 	}
 }


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(connector-codex): native hooks + MCP for full mid-session memory` (commit `bd0f5822`) now and I'll follow up shortly with feedback.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,145 @@ import {
...
+	writeFileSync(path, JSON.stringify(hooks, null, 2) + "\n");
 }
 
-if [ "\${SIGNET_NO_HOOKS:-}" = "1" ] || [ "\${SIGNET_CODEX_BYPASS_WRAPPER:-}" = "1" ]; then
-	exec "$REAL_CODEX_BIN" "$@"
-fi
+function removeSignetHooks(hooks: HooksJson): HooksJson {
+	const cleaned = { ...hooks };
+	delete cleaned._signet;
+	delete cleaned.sessionStart;
```
NicholaiVogel: **Data loss bug — merged user hooks are destroyed on uninstall.**

During install, when a user already has a `hooks.json`, the code merges Signet's entries *after* the user's existing entries:
```ts
merged[key] = [...current, ...signet]; // user hooks first, then Signet
```
But `removeSignetHooks` unconditionally deletes the entire `sessionStart`, `userPromptSubmit`, and `stop` arrays. Any user hooks that were present in those arrays are silently lost.

Fix: instead of deleting the arrays wholesale, filter them to remove only entries added by Signet. Signet-added entries are identifiable by the `command` array containing `"hook session-start"`, `"hook user-prompt-submit"`, or `"hook session-end"`. For example:
```ts
function removeSignetHandlers(arr: unknown[]): unknown[] {
  return arr.filter((entry) => {
    const raw = JSON.stringify(entry);
    return !raw.includes('"hook session-start"') &&
           !raw.includes('"hook user-prompt-submit"') &&
           !raw.includes('"hook session-end"');
  });
}
```
Then in `removeSignetHooks`, rebuild each key with the filtered array (and delete the key only if it becomes empty). Also delete `_signet` only when no Signet entries remain.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,145 @@ import {
...
+// ---------------------------------------------------------------------------
 
-trap 'cleanup' EXIT
+function patchConfigToml(path: string, mcpCmd: string): boolean {
+	const dir = join(path, "..");
+	mkdirSync(dir, { recursive: true });
 
-session_start
+	if (!existsSync(path)) {
+		writeFileSync(path, `# Signet MCP server\n[mcp_servers.signet]\ncommand = "${mcpCmd}"\n`);
```
NicholaiVogel: **Invalid TOML on Windows — unescaped backslashes in a double-quoted string.**

`resolveSignetMcpCmd()` on Windows returns a string like:
```
C:\Program Files\nodejs\node.exe C:\Users\alice\...\mcp-stdio.js
```
This value is embedded verbatim into a TOML double-quoted string:
```ts
command = "C:\Program Files\nodejs\node.exe C:\Users\alice\..."
```
In TOML, `\P`, `\U`, `\a`, `\n`, etc. are either undefined escape sequences (parse error in strict parsers) or recognised escape sequences with unintended meaning. `codex-rs` is written in Rust; the `toml` crate rejects unknown escape sequences with a hard error, so the config file will fail to parse and the MCP server will never be registered on Windows.

Same issue on line 120 (the append path).

Fix: escape backslashes before embedding, or use a TOML literal string (single-quoted, no escapes):
```ts
// Option A: escape backslashes
const escaped = mcpCmd.replace(/\\/g, "\\\\");
`command = "${escaped}"\n`

// Option B: TOML literal string (no escaping needed, but can't contain single quotes)
`command = '${mcpCmd}'\n`
```
For Windows paths that may contain spaces, also verify whether Codex expects `command` to be a string (shell-split by Codex) or an array of strings. If it's an array, the format should be `command = ["node.exe", "path\\to\\mcp.js"]` and the array form from `resolveSignetArgs` is the right model to follow.

packages/connector-codex/src/index.ts
```diff
@@ -220,120 +160,100 @@ export class CodexConnector extends BaseConnector {
...
+		const hooksPath = this.getHooksJsonPath();
+		const existing = readHooksJson(hooksPath);
+		if (existing) {
+			if (isSignetOwned(existing)) {
+				const cleaned = removeSignetHooks(existing);
+				const remaining = Object.keys(cleaned).filter((k) => k !== "_signet");
+				if (remaining.length === 0) {
+					rmSync(hooksPath, { force: true });
+					filesRemoved.push(hooksPath);
+				} else {
```
NicholaiVogel, line 282: **Skills symlink is never removed on uninstall.**

`install()` creates `~/.codex/skills` via `this.symlinkSkills(basePath, codexHome)` but `uninstall()` has no corresponding cleanup. After uninstall the symlink persists, and a subsequent fresh install may behave differently depending on whether `symlinkSkills` is idempotent when the target already exists. At minimum the symlink should be removed from `filesRemoved`; at best, call a `removeSymlink` helper before returning.

aaf2tbz:
All three review items addressed in `9692bfa6`:

1. **User hook preservation on uninstall** — `removeSignetHooks` now filters individual entries by checking for `"hook session-start"` / `"hook user-prompt-submit"` / `"hook session-end"` in stringified handlers. Arrays are only deleted if empty after filtering. User hooks that were merged in are fully preserved.

2. **Windows TOML escaping** — Added `tomlQuote()` that uses TOML literal strings (single-quoted, no escape processing) for paths with backslashes. Falls back to double-quoted with `\\` escaping if the value contains single quotes. Codex's Rust `toml` crate will parse correctly.

3. **Skills symlink cleanup** — `uninstall()` now removes `~/.codex/skills` symlink and reports it in `filesRemoved`.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat(connector-codex): native hooks + MCP for full mid-session memory` (commit `9692bfa6`); taking another look at the feature work and I will report back shortly.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,171 @@ import {
...
+	const signetJs = join(entry, "..", "..", "bin", "signet.js");
+	if (existsSync(signetJs)) return [process.execPath, signetJs];
+	return ["signet"];
 }
 
-function appendBlock(content: string): string {
-	const stripped = stripBlock(content).trimEnd();
-	return `${stripped}${stripped ? "\n\n" : ""}${SHELL_BLOCK}`;
+/** Resolve signet-mcp command string for config.toml. */
+function resolveSignetMcpCmd(): string {
```
NicholaiVogel: **Previously blocked issue #1 — fixed.** `isSignetHandler` + the rewritten `removeSignetHooks` loop correctly filter only Signet-owned entries instead of deleting entire hook arrays. Merged user hooks are preserved on uninstall.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,171 @@ import {
...
+	);
+	if (!hasSignet) delete cleaned._signet;
+	return cleaned;
 }
 
-if [ "\${SIGNET_NO_HOOKS:-}" = "1" ] || [ "\${SIGNET_CODEX_BYPASS_WRAPPER:-}" = "1" ]; then
-	exec "$REAL_CODEX_BIN" "$@"
-fi
+// ---------------------------------------------------------------------------
+// MCP server registration (config.toml)
```
NicholaiVogel, line 133: **Previously blocked issue #2 — fixed.** `tomlQuote` prefers TOML literal strings (single-quoted, no escape processing) so Windows backslashes survive verbatim. The double-quoted fallback is correctly escaped. No longer invalid TOML.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,171 @@ import {
...
+// Signet command resolution
+// ---------------------------------------------------------------------------
+
+/** Resolve signet command for hook invocation. Returns array form for hooks.json command field. */
+function resolveSignetArgs(): string[] {
+	if (process.platform !== "win32") return ["signet"];
+	const entry = process.argv[1] || "";
+	const signetJs = join(entry, "..", "..", "bin", "signet.js");
+	if (existsSync(signetJs)) return [process.execPath, signetJs];
+	return ["signet"];
```
NicholaiVogel, line 28: **Residual minor issue (not introduced here but worth noting):** `resolveSignetMcpCmd` returns `process.execPath + ' ' + mcpJs` as a single string. If `process.execPath` is `C:\Program Files\nodejs\node.exe` the TOML becomes `command = 'C:\Program Files\nodejs\node.exe C:\...\mcp-stdio.js'` — a single string with an embedded space. If Codex's MCP runner does a naive split on whitespace it will misparse the executable path. The hooks path (`resolveSignetArgs`) correctly uses an array; consider making `resolveSignetMcpCmd` return `string[]` and write a TOML inline array (`command = ["node.exe", "path"]`) for consistency.

aaf2tbz:
Remaining review item addressed in `e0fd1018`:

**MCP command as TOML inline array** — `resolveSignetMcpCmd` → `resolveSignetMcpArgs` now returns `string[]`. `patchConfigToml` writes a TOML inline array (`command = ['node.exe', 'path/to/mcp.js']`) so Windows paths with spaces are properly split into separate elements, matching the array form used by `resolveSignetArgs` for hooks. No more naive whitespace splitting risk.

Also merged latest main to bring branch up to date.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat(connector-codex): native hooks + MCP for full mid-session memory` (commit `e0fd1018`); taking another look at the feature work and I will report back shortly.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,173 @@ import {
...
+		} else {
+			cleaned[key] = filtered;
+		}
+	}
+	// Only remove marker if no Signet entries remain
+	const hasSignet = ["sessionStart", "userPromptSubmit", "stop"].some(
+		(k) => Array.isArray(cleaned[k]) && (cleaned[k] as unknown[]).some(isSignetHandler),
+	);
+	if (!hasSignet) delete cleaned._signet;
+	return cleaned;
```
NicholaiVogel, line 129: **Possible TOML schema mismatch — `command` string vs array.**

For non-Windows installs (`mcpArgs.length === 1`) the written TOML is:
```toml
command = 'signet-mcp'
```
For Windows multi-arg installs it is:
```toml
command = ['C:\\...\\node.exe', 'C:\\...\\mcp-stdio.js']
```
Most MCP tool configurations (and the pattern used by `connector-claude-code`) treat `command` as an array even for a single element. If Codex's TOML parser expects `command` to be an array type, the single-element string path will silently produce an invalid or ignored entry on Linux/macOS.

Consider normalising: always emit an inline array, e.g. `tomlInlineArray(mcpArgs)` regardless of length, to keep the shape consistent across platforms.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,173 @@ import {
...
-exit /b %ERRORLEVEL%
-`;
+function unpatchConfigToml(path: string): boolean {
+	if (!existsSync(path)) return false;
+	const content = readFileSync(path, "utf-8");
+	if (!content.includes("[mcp_servers.signet]")) return false;
+
+	// Remove the signet MCP block — handles both with and without comment
+	const lines = content.split("\n");
+	const filtered: string[] = [];
```
NicholaiVogel, line 171: **Uninstall does not clean merged hooks when `_signet` marker is absent.**

If a user's existing `hooks.json` was merged during install (and thus has `_signet: true`), uninstall correctly removes only Signet-added entries. However, if the marker was somehow stripped (e.g., the user edited the file and removed it), `isSignetOwned` returns `false` and the uninstall block is skipped entirely — Signet hooks remain silently. This is an edge case, but consider falling back to a marker-less scan with `isSignetHandler` when `_signet` is absent, issuing a warning rather than silently no-oping.

packages/connector-codex/src/index.ts
```diff
@@ -1,4 +1,3 @@
-import { spawnSync } from "node:child_process";
 import {
```
NicholaiVogel, line 1: **Incremental diff bundles unrelated dashboard UI changes.**

The incremental diff since the last reviewed SHA includes `CortexAppsPanel.svelte`, nav-group refactors (`isCortexGroup`), removal of `ListChecks`/`LayoutGrid` icons, and sidebar restructuring — none of which relate to the connector-codex native hooks work that is the stated scope of this PR. These changes also introduce a `0.70.0` root version bump while workspace packages remain `0.69.5` in `bun.lock`, which may cause publish-time version drift. The UI work should either be split into a separate PR or the version bump should be reflected consistently across all workspace packages.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,173 @@ import {
...
-	if (end < 0) return content;
-	const after = content.slice(end + SHELL_BLOCK_END.length);
-	const before = content.slice(0, start).trimEnd();
-	return `${before}${before ? "\n\n" : ""}${after.trimStart()}`.trimEnd() + "\n";
+// ---------------------------------------------------------------------------
+// Signet command resolution
+// ---------------------------------------------------------------------------
+
+/** Resolve signet command for hook invocation. Returns array form for hooks.json command field. */
+function resolveSignetArgs(): string[] {
```
NicholaiVogel, line 23: **Windows `resolveSignetArgs` path traversal assumption.**

`join(process.argv[1], '..', '..', 'bin', 'signet.js')` works correctly when `process.argv[1]` is `<pkg-root>/bin/signet.js` (navigates to `<pkg-root>/bin/signet.js` — same file, which then `existsSync`-passes). This is consistent and correct, but it relies on the layout assumption `bin/` being two `..` levels up from `argv[1]`. If the CLI is invoked via a shim or junction (common with some Windows npm setups), `argv[1]` may point elsewhere and the fallback `['signet']` will be used. This is acceptable as a graceful fallback, but worth a comment in the code for future maintainers.

aaf2tbz:
Review items from latest round addressed in `82482d25`:

1. **TOML command string vs array** — Now always emits TOML inline array on all platforms (`command = ['signet-mcp']`), consistent schema regardless of OS.

2. **Uninstall with missing `_signet` marker** — Falls back to `isSignetHandler` scan when marker is absent. Signet hooks are cleaned up even if user edited the file and stripped the marker.

3. **Unrelated dashboard changes in diff** — These came from merging main (Cortex rename from PR #256). Not part of this PR's code changes.

4. **Windows path traversal** — Acknowledged as correct (navigates to same file, `existsSync` validates). No change needed.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat(connector-codex): native hooks + MCP for full mid-session memory` (commit `82482d25`); taking another look at the feature work and I will report back shortly.

packages/connector-codex/src/index.ts
```diff
@@ -220,120 +188,111 @@ export class CodexConnector extends BaseConnector {
...
 
-		const wrapperPath = this.getWrapperPath();
-		if (existsSync(wrapperPath)) {
-			rmSync(wrapperPath, { force: true });
-			filesRemoved.push(wrapperPath);
+		// 1. Remove hooks.json (or clean Signet entries from merged file)
+		const hooksPath = this.getHooksJsonPath();
+		const existing = readHooksJson(hooksPath);
+		if (existing) {
+			// Check marker first; fall back to handler scan if marker was stripped
```
NicholaiVogel, line 270: **Previously flagged — now fixed.** `tomlInlineArray` is now unconditionally applied, so the single-element Linux/macOS path writes `['signet-mcp']` (a TOML inline array) rather than a bare string. This resolves the prior concern about Codex schema requiring `command` to always be an array type.

packages/connector-codex/src/index.ts
```diff
@@ -220,120 +188,111 @@ export class CodexConnector extends BaseConnector {
...
+			const hasMarker = isSignetOwned(existing);
+			const hasHandlers = JSON.stringify(existing).includes("hook session-start") ||
```
NicholaiVogel: **Pre-existing low-severity concern (not introduced by this diff).** `hasHandlers` searches `JSON.stringify(existing)` — the full file contents — for the strings `"hook session-start"` and `"hook user-prompt-submit"`. If any user hook incidentally contains those strings anywhere in its JSON representation (e.g., `command: ["sh", "-c", "echo 'hook session-start ran'"]`), we enter the cleanup path. `removeSignetHooks` then runs `isSignetHandler` per-entry using the same string match, so that user entry would be wrongly removed during uninstall. Probability is very low in practice, but a more robust approach would match on the specific command array structure (e.g., checking that `command[N] === "hook"` and `command[N+1] === "session-start"`) rather than substring matching the entire serialised entry.

aaf2tbz:
Review items from latest round addressed in `c686e17a`:

1. **isSignetHandler false positives** — Now checks `handlers[].command` array structure instead of substring-matching serialized JSON. User hooks with incidental string matches won't be wrongly removed.

2. **Windows path comment** — Added clarifying comment on the `argv[1]` traversal assumption and fallback behavior.

3. **Unrelated dashboard changes** — From merging main (Cortex rename, PR #256). Not part of this PR's scope.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat(connector-codex): native hooks + MCP for full mid-session memory` (commit `c686e17a`); taking another look at the feature work and I will report back shortly.

packages/connector-codex/src/index.ts
```diff
@@ -14,204 +13,184 @@ import {
...
-	TRANSCRIPT_PATH="$(find_session_file)"
-	payload="$(printf '{"session_id":"%s","transcript_path":"%s","cwd":"%s"}' "$(json_escape "$SESSION_KEY")" "$(json_escape "$TRANSCRIPT_PATH")" "$(json_escape "$PWD")")"
-	printf "%s" "$payload" | "$SIGNET_BIN" hook session-end -H codex >/dev/null 2>&1 || true
+const SIGNET_HOOK_CMDS = ["hook session-start", "hook user-prompt-submit", "hook session-end"] as const;
+
+function isSignetHandler(entry: unknown): boolean {
+	if (typeof entry !== "object" || entry === null) return false;
+	const handlers = (entry as Record<string, unknown>).handlers;
+	if (!Array.isArray(handlers)) return false;
+	for (const handler of handlers) {
```
NicholaiVogel, line 103: Pre-existing (not introduced here, not a blocker): `isInstalled()` at the bottom of the file still uses `JSON.stringify(hooks).includes("hook session-start")` — the full-JSON string-matching approach that was just replaced in `isSignetHandler` and `hasHandlers`. Since `isInstalled()` is read-only with no destructive side-effects the false-positive risk is cosmetic (a spurious `true` return), but for consistency this should eventually be replaced with `isSignetHandler` calls on each hook array.

aaf2tbz:
Fixed in `043abe4a`: `isInstalled()` now uses structural `isSignetHandler` check on each hook array, consistent with `removeSignetHooks` and the uninstall fallback.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `feat(connector-codex): native hooks + MCP for full mid-session memory` (commit `043abe4a`); taking another look at the feature work and I will report back shortly.

PR #259, opened 2026-03-21
fix: pipeline worker stall after burst processing

### LINKED ISSUE #248: pipelineV2 extraction worker stalls after burst — stops picking up pending jobs until daemon restart
# Bug Report: Pipeline Worker Stalls After Burst Processing

**Signet version:** 0.68.0 (confirmed present in 0.61.0 also — persists across upgrade)
**Platform:** Linux (Ubuntu 24.04, kernel 6.8.0-106-generic x86_64)
**Node:** v25.6.0 | **Bun:** 1.3.11
**Harness:** OpenClaw
**Date observed:** 2026-03-18 through 2026-03-19

---

## Summary

The pipelineV2 extraction worker enters a near-idle state after bursts of job processing. Active leased job count drops from 6-10 to 0-1 despite hundreds of `pending` jobs in the queue. The stall persists indefinitely — jobs do not self-recover. The only known fix is `systemctl --user restart signet`, which immediately resumes processing at full concurrency. This behavior reproduces consistently and is unaffected by upgrading from 0.61.0 to 0.68.0.

---

## Environment

```yaml
# agent.yaml (relevant sections)
memory:
  pipelineV2:
    enabled: true
    extraction:
      provider: ollama
      model: claude-sonnet-4-5  # via Ollama-compatible proxy at 11435
      base_url: http://10.10.0.182:11435
      timeout: 30000
    worker:
      leaseTimeoutMs: 600000
      concurrency: 6
    semanticContradictionEnabled: true
    graph:
      enabled: true
    autonomous:
      enabled: true
      allowUpdateDelete: true
    predictor:
      enabled: true

# systemd service env
SIGNET_MAX_LLM_CONCURRENCY=12
```

---

## Reproduction Steps

1. Ingest a large batch of documents (~900 jobs via the `/api/documents` endpoint)
2. Start the daemon — processing begins normally at 6-10 concurrent leased jobs
3. Observe: after ~50-100 jobs complete, leased count drops to 0-1
4. Pending queue still has hundreds of jobs — no error is logged, no crash
5. `signet daemon logs` shows no activity — daemon is running (PID present, health endpoint returns healthy) but not picking up work
6. `systemctl --user restart signet` → immediately resumes at 6-10 concurrent leases
7. Repeat: stalls again after another burst of ~50-100 completions
8. Pattern is consistent: burst → stall → manual restart → burst → stall

---

## Observed Behavior

**During active processing (healthy):**
```
[21:49:31] completed=13(+13) pending=869 leased=10
[21:49:41] completed=21(+8)  pending=861 leased=10
[21:49:51] completed=28(+7)  pending=855 leased=9
[21:50:01] completed=35(+7)  pending=848 leased=9
```

**After stall (unhealthy — persists for 10+ minutes without intervention):**
```
[22:29:00] completed=465(+0) pending=624 leased=1
[22:29:10] completed=465(+0) pending=624 leased=1
[22:29:20] completed=465(+0) pending=624 leased=1
[22:29:30] completed=465(+0) pending=624 leased=1
```

**Immediately after `systemctl --user restart signet`:**
```
[after restart] completed=471(+6) pending=648 leased=3  (within 8s)
[+8s]           completed=473(+2) pending=647 leased=2
[+24s]          completed=475(+2) pending=648 leased=5
```

---

## Job Database State During Stall

```
status      count
----------  -----
completed   487
leased      2      ← stuck, not completing
pending     643    ← not being picked up

attempts distribution:
  0 attempts: 626 jobs
  1 attempt:  466 jobs
  2 attempts:  20 jobs
  3 attempts:  20 jobs
max attempts: 3
```

The `leaseTimeoutMs` is set to 600000ms (10 minutes). During the stall, leased jobs stay leased indefinitely — they are not reaped and re-queued by the stale lease reaper in a timely manner.

---

## Daemon Logs During Stall

No errors logged during the stall period. The last log entries before going quiet are normal job completions:

```json
{"timestamp":"...","level":"info","category":"pipeline","message":"Extraction job completed","data":{"jobId":"...","facts":3,"entities":2,...}}
```

After that — silence. No error, no crash, no backoff message.

---

## Prior Context (Potential Trigger)

Earlier in the same session, the extraction endpoint was pointed at a remote Groq API via a local proxy. Groq returned HTTP 429 (rate limit) responses wrapped as HTTP 500 from the proxy. This produced a burst of rapid job failures:

```json
{"level":"warn","category":"pipeline","message":"Extraction LLM call failed",
 "data":{"error":"Ollama HTTP 500: {\"error\":{\"message\":\"Rate limit reached... TPM: Limit..."}}

{"level":"warn","category":"pipeline","message":"Job failed",
 "data":{"jobId":"bbaceb7e-...","error":"LLM extraction failed: ...","attempt":3}}
```

Jobs hitting max attempts (3) were marked dead, then manually reset to `pending` via direct SQLite update. It's possible this burst of rapid failures + manual DB resets left the worker in a confused backoff state that persists even after the extraction endpoint was changed to a reliable provider (Anthropic Claude Sonnet via proxy — fast, no failures).

**However:** the stalling also occurred *before* the Groq rate limiting episode — it was observed with the original Ollama/qwen2.5-coder:32b setup as well, though with slower job completion rates it was harder to distinguish stalling from slow processing.

---

## Health Check During Stall

```bash
$ curl -s http://localhost:3850/health
{"status":"healthy","uptime":843,"pid":142817,"version":"0.61.0","port":3850,
 "agentsDir":"/home/adamp/.agents","db":true,"updateAvailable":false,"pendingRestart":false}
```

Daemon reports healthy. The issue is specifically in the pipeline worker, not the daemon process itself.

---

## Hypotheses

1. **Worker coroutine/promise leak**: An async worker slot gets stuck awaiting a promise that never resolves, consuming a concurrency slot without making progress. After enough slots are consumed this way, effective concurrency drops to ~1.

2. **Backoff state not reset between jobs**: After a burst of failures (e.g. rate limit HTTP 500s), the worker enters exponential backoff. Backoff state is per-worker-instance, not per-job — so it persists across subsequent successful jobs until the process is restarted.

3. **Stale lease reaper race**: The reaper that reclaims stuck leased jobs runs on a timer but may not correctly requeue jobs if the worker pool is in a degraded state.

4. **`SIGNET_MAX_LLM_CONCURRENCY=12` env var ignored**: The service env sets concurrency to 12 but `agent.yaml` sets `worker.concurrency: 6`. It's unclear which takes precedence. If the env var is ignored, and the concurrency value interacts with lease slot management, this could contribute.

---

## Workaround

```bash
systemctl --user restart signet
```

Immediate effect — processing resumes within seconds. Must be repeated every 50-100 jobs (approximately every 5-15 minutes depending on job speed).

---

## Suggested Fix

- Add a watchdog in the pipeline worker that detects when `leased > 0` and `completed` count hasn't changed in N seconds (e.g. 60s) and automatically requeues stale leases + wakes idle workers
- Or: expose a `/api/pipeline/nudge` endpoint that forces the worker to re-poll the pending queue without a full daemon restart
- Or: ensure backoff state is scoped to the individual job/provider call, not the worker instance, so failures don't poison subsequent healthy jobs

---

*Reported by: Adam Paulsen via Sam Bridger (OpenClaw agent)*
*Reproduction rate: ~100% on large batch jobs (>200 pending)*

NicholaiVogel:
## Summary

Fixes #248 — the pipelineV2 extraction worker silently stalls after processing ~50-100 jobs from large batches.

**Root cause**: The worker's global `consecutiveFailures` counter drove exponential backoff for its poll interval. On success it only decremented by 1 (`--`) instead of resetting to 0. Scattered failures across a batch ratcheted the counter up to 5+, pushing poll delay to the 30s cap. Since the worker is serial (one job per tick), this effectively stalled processing at ~2 jobs/min.

**Fix**:
- Reset `consecutiveFailures` to 0 on success and on empty queue (not decrement by 1). Per-job SQL backoff in `leaseJob()` already handles individual retry spacing.
- Reduce `MAX_DELAY` from 30s to 15s
- Add a 30s watchdog timer that detects stalls (pending > 0, no progress for 60s) and self-heals by resetting backoff and forcing an immediate tick
- Expand `WorkerHandle` with `nudge()` method and `stats` getter for observability
- Add `POST /api/pipeline/nudge` endpoint so operators can unstick the worker without restarting the daemon
- Add pipeline stall detection to `/health` response (`pipeline.extractionStalled`)
- Log backoff escalation when `consecutiveFailures > 2`

## Test plan

- [x] `bun run build` passes
- [x] 4 new behavioral contract tests in `worker.test.ts`:
  - Backoff resets on success after failures
  - Backoff resets when queue empties
  - `nudge()` forces immediate repoll regardless of current delay
  - `stats` reflects current worker state
- [x] No regressions (13 pre-existing failures unchanged)
- [ ] Manual: start daemon with large batch, observe backoff logging, verify no stall
- [ ] Manual: `curl -X POST localhost:3850/api/pipeline/nudge` returns `{ nudged: true }`
- [ ] Manual: `curl localhost:3850/health` includes `pipeline` stall detection

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -6806,6 +6806,10 @@ app.get("/api/pipeline/status", (c) => {
 	});
 });
 
+app.use("/api/pipeline/nudge", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
 app.post("/api/pipeline/nudge", (c) => {
 	if (!nudgeExtractionWorker()) {
 		return c.json({ error: "Extraction worker not running" }, 503);


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: pipeline worker stall after burst processing` (commit `eebe3ddb`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1386,14 +1454,50 @@ export function startWorker(
...
+					)
+					.get() as { cnt: number },
+			);
+			return row.cnt;
+		} catch {
+			return 0;
+		}
+	}
+
 	return {
```
NicholaiVogel, line 1477: **Race condition: `inflight` can be overwritten while a tick is in progress.**

Both `nudge()` and the watchdog schedule a `setTimeout(async () => { inflight = tick(); ... }, 0)` without checking whether a tick is already running. If `inflight` is non-null when the callback fires, the original promise reference is silently overwritten. `stop()` then only awaits the newer tick (or nothing), leaving the original tick's DB writes running concurrently with teardown — including any `db.close()` that follows `stopPipeline()`. This can produce use-after-close errors in SQLite.

Fix: guard at the top of the callback:
```ts
pollTimer = setTimeout(async () => {
    if (inflight) { scheduleTick(); return; } // skip if already running
    inflight = tick();
    await inflight;
    inflight = null;
    scheduleTick();
}, 0);
```
Apply the same guard in the watchdog's recovery block.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1372,6 +1402,44 @@ export function startWorker(
...
 
+	// Stall watchdog — detect when worker stops making progress
+	watchdog = setInterval(() => {
+		if (!running) return;
+		if (Date.now() - lastProgress < STALL_THRESHOLD) return;
+
+		let pending = 0;
+		try {
+			const row = accessor.withReadDb((db) =>
+				db
```
NicholaiVogel, line 1416: **Watchdog stall detection is defeated for failure-retry loops.**

`lastProgress = Date.now()` is set inside both catch blocks (job failure at ~line 1346 and tick error at ~line 1360). With `MAX_DELAY = 15s`, a continuously-failing worker refreshes `lastProgress` every ≤15s, so `Date.now() - lastProgress` never exceeds `STALL_THRESHOLD` (60s). The watchdog provides no protection in the exact scenario where it's most needed: a worker stuck in an infinite failure loop.

Separate the two concerns — rename `lastProgress` to `lastAttemptAt` (for watchdog liveness) and add a separate `lastSuccessAt` (for health/stats reporting). Update the watchdog to check `lastAttemptAt` (is the timer even firing?), and expose `lastSuccessAt` through `WorkerStats.lastProgressAt` so the health endpoint shows meaningful stall state.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1372,6 +1402,44 @@ export function startWorker(
...
+					.prepare(
+						"SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending'",
```
NicholaiVogel, line 1418: **Same watchdog timer body has the same `inflight` overwrite bug as `nudge()`** — see the comment on the nudge block. The watchdog's recovery `setTimeout` should also guard on `inflight` before scheduling a new tick.

packages/daemon/src/daemon.ts
```diff
@@ -6791,6 +6806,13 @@ app.get("/api/pipeline/status", (c) => {
 	});
 });
 
+app.post("/api/pipeline/nudge", (c) => {
```
NicholaiVogel, line 6813: **Missing auth on `POST /api/pipeline/nudge`.**

All other write endpoints in this file use an auth middleware (or equivalent session check). This new endpoint has none. An unauthenticated caller who can reach port 3850 can force repeated DB reads and worker resets. Even if the daemon is currently localhost-only, adding auth here matches the established pattern and avoids regressing if the bind address ever changes.

Add the same auth middleware (e.g., `requireAuth` or bearer-token check) used on other `/api/pipeline/*` mutation routes.

packages/daemon/src/pipeline/worker.test.ts
```diff
@@ -1353,3 +1380,144 @@ describe("Worker dead-job path", () => {
...
+	});
+
+	it("resets backoff when queue empties after failures", async () => {
+		// First job will fail all retries (dead). Then we enqueue a second
+		// job — it should process promptly, not after 30s backoff.
+		insertMemory(db, "mem-drain-1", "Will fail");
+		enqueueExtractionJob(accessor, "mem-drain-1");
+
+		const cfg = {
+			...PIPELINE_CFG,
```
NicholaiVogel, line 1447: **Test only checks two of three enqueued jobs (`mem-bf-2`, `mem-bf-3`), silently ignoring `mem-bf-1`.**

`mem-bf-1` is the job processed on the first tick (the one that fails). With `maxRetries: 1` and the provider throwing on call 1, `mem-bf-1` will end up in `dead` status after the single allowed attempt. The test never asserts its final state, so a regression where `mem-bf-1` is accidentally left in `leased` or `pending` would go undetected. Consider adding `expect(getJob(db, 'mem-bf-1')?.status).toBe('dead')` or `'completed'` depending on intended behavior.

packages/daemon/src/daemon.ts
```diff
@@ -1284,6 +1285,14 @@ app.get("/health", (c) => {
 	} catch {
 		// DB unreachable
 	}
+	const workers = getPipelineWorkerStatus();
+	const extraction = workers.extraction;
+	const stalled =
+		extraction.running &&
+		extraction.stats !== undefined &&
+		extraction.stats.pending > 0 &&
```
NicholaiVogel, line 1293: **Stall detection in `/health` uses `lastProgressAt` which is updated on failures.**

Because `lastProgressAt` tracks last attempt (not last success), `extractionStalled` will be `false` even for a worker hammering failures every 15s. This makes the health signal inaccurate. Once `lastProgressAt` / `lastSuccessAt` are split (see worker.ts comment), update this check to use the success timestamp.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix: pipeline worker stall after burst processing` (commit `18bf2bfb`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -6791,6 +6806,13 @@ app.get("/api/pipeline/status", (c) => {
 	});
 });
```
NicholaiVogel, line 6807: **Unresolved from prior review — auth missing on nudge endpoint.**

This endpoint has no authentication or authorization gate. Every other mutation endpoint in this daemon presumably goes through auth middleware. If the daemon port is reachable beyond localhost (or if an attacker can reach it via SSRF), any caller can force the worker to repoll immediately — effectively a free DoS knob that bypasses rate limiting and can cause job processing to spike.

Minimum fix: apply the same auth middleware used on other `/api/pipeline/*` write endpoints, or at minimum restrict to loopback. If the daemon intentionally has no external auth (all callers are localhost-only), add an explicit comment and consider binding to `127.0.0.1` at the listen level rather than relying on network topology.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1372,6 +1404,46 @@ export function startWorker(
...
+		if (pending === 0) return;
+
+		logger.warn("pipeline", "Worker stall detected, resetting backoff", {
+			pending,
+			failures: consecutiveFailures,
+			stalledMs: Date.now() - lastSuccess,
+		});
+
+		consecutiveFailures = 0;
+		lastSuccess = Date.now();
```
NicholaiVogel, line 1435: **Double-scheduling when watchdog fires during an active tick.**

When `inflight !== null`, the guard correctly avoids overwriting it:
```ts
if (inflight) { scheduleTick(); return; }
```
But `scheduleTick()` sets `pollTimer` to a new timer. The original in-flight tick will also call `scheduleTick()` when it completes. Neither call clears the other's timer, so two `setTimeout` handles are now live and both will eventually fire. Both see `inflight === null` (because the real tick is done), both do `inflight = tick()`, and both set up their own `scheduleTick()` chain — so the worker briefly runs two concurrent ticks and then permanently doubles its tick rate until the process restarts.

Fix: track whether a tick is already scheduled and guard against it, or clear the old timer before creating a new one:
```ts
// In scheduleTick:
if (pollTimer) clearTimeout(pollTimer);
pollTimer = setTimeout(...);
```
That one-liner makes it safe to call `scheduleTick()` from multiple paths without accumulating timers. The same issue exists symmetrically in `nudge()` (line ~1484).

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1386,14 +1458,51 @@ export function startWorker(
...
+
 	return {
 		get running() {
 			return running;
 		},
+		nudge() {
+			consecutiveFailures = 0;
+			if (pollTimer) clearTimeout(pollTimer);
+			// Schedule immediate tick — guard against concurrent inflight
+			pollTimer = setTimeout(async () => {
```
NicholaiVogel, line 1488: **Same double-scheduling issue as the watchdog path above.**

When `nudge()` is called while a tick is in-flight, `scheduleTick()` is called here and again by the original tick on completion, producing two live timers. See watchdog comment for the full analysis and fix.

packages/daemon/src/daemon.ts
```diff
@@ -1284,6 +1285,14 @@ app.get("/health", (c) => {
 	} catch {
 		// DB unreachable
 	}
+	const workers = getPipelineWorkerStatus();
+	const extraction = workers.extraction;
+	const stalled =
+		extraction.running &&
+		extraction.stats !== undefined &&
+		extraction.stats.pending > 0 &&
+		Date.now() - extraction.stats.lastProgressAt > 60_000;
```
NicholaiVogel, line 1294: **Minor — `stats` getter triggers a synchronous DB query on every health poll.**

`getPipelineWorkerStatus()` calls `workerHandle?.stats`, which calls `pendingCount()` — a synchronous `SELECT COUNT(*)` against SQLite. This runs on every `GET /health`. If `/health` is polled by an orchestrator at high frequency (e.g. 1-2 Hz), you'll add constant read pressure to a database that might be under write load from the worker itself. Consider caching the pending count with a short TTL (e.g. refresh at most every 5s) or computing it only when explicitly needed.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix: pipeline worker stall after burst processing` (commit `a11332fe`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1386,14 +1459,53 @@ export function startWorker(
...
+			return row.cnt;
+		} catch {
+			return 0;
+		}
+	}
+
 	return {
 		get running() {
 			return running;
 		},
```
NicholaiVogel, line 1480: **Fixed (prior review):** The double-`scheduleTick` race is resolved. Old pattern called `scheduleTick()` inside the `setTimeout` callback when `inflight` was truthy, creating a second live timer alongside the in-flight tick's own completion callback. New pattern exits before scheduling anything, leaving the single `scheduleTick()` in the tick's completion chain as the only path forward. Correct.

packages/daemon/src/daemon.ts
```diff
@@ -6791,6 +6806,13 @@ app.get("/api/pipeline/status", (c) => {
 	});
 });
 
+app.post("/api/pipeline/nudge", (c) => {
+	if (!nudgeExtractionWorker()) {
+		return c.json({ error: "Extraction worker not running" }, 503);
+	}
```
NicholaiVogel, line 6816: **Unresolved (prior review):** `POST /api/pipeline/nudge` still has no authentication check. If the daemon port is reachable beyond localhost (e.g., bound to 0.0.0.0 or behind a reverse proxy without IP allowlisting), any caller can force a pipeline repoll. At minimum, add the same session/token middleware that protects other write endpoints in this file, or assert the request originates from localhost.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix: pipeline worker stall after burst processing` (commit `104e9074`); taking another look at the fixes and I will report back shortly.

PR #262, opened 2026-03-21
fix(tray): cross-platform build script for Windows CI

aaf2tbz:
## Why

Desktop Build CI passes on macOS and Linux but fails on Windows with `cp: illegal option -- r`. The `build:ts` npm script uses Unix-only commands (`rm -rf`, `cp -r`) that don't exist on Windows `cmd`.

## What Changed

Extracted the build:ts logic into `build-ts.mjs` using Node.js fs APIs:
- `fs.rmSync(dir, {recursive: true, force: true})` replaces `rm -rf`
- `fs.cpSync(src, dest, {recursive: true})` replaces `cp -r`
- `fs.copyFileSync()` replaces `cp`

Tested locally — produces correct `dist/` with tray.js, HTML files, and all dashboard assets.

## Test Plan

- [ ] Desktop Build CI passes on Windows (the actual validation)
- [ ] macOS and Linux continue to pass

### DIFF
--- a/packages/tray/build-ts.mjs
+++ b/packages/tray/build-ts.mjs
@@ -1,0 +1,28 @@
+#!/usr/bin/env node
+/**
+ * Cross-platform build script for tray TypeScript assets.
+ * Replaces Unix-only rm -rf / cp -r in package.json scripts.
+ */
+import { rmSync, cpSync, copyFileSync, readdirSync } from "node:fs";
+import { execSync } from "node:child_process";
+import { join } from "node:path";
+
+// Clean dist
+rmSync("dist", { recursive: true, force: true });
+
+// Bundle TypeScript
+execSync("bun build src-ts/index.ts --outfile dist/tray.js --target browser --minify", { stdio: "inherit" });
+
+// Build dashboard
+execSync("bun run build:dashboard", { stdio: "inherit" });
+
+// Copy dashboard build contents into dist (not the directory itself)
+const src = "../cli/dashboard/build";
+for (const entry of readdirSync(src)) {
+	cpSync(join(src, entry), join("dist", entry), { recursive: true });
+}
+
+// Copy HTML files
+for (const file of ["tray.html", "capture.html", "search.html"]) {
+	copyFileSync(file, `dist/${file}`);
+}
--- a/packages/tray/package.json
+++ b/packages/tray/package.json
@@ -5,7 +5,7 @@
   "description": "Signet desktop application",
   "scripts": {
     "build:dashboard": "cd ../cli/dashboard && bun run build",
-    "build:ts": "rm -rf dist && bun build src-ts/index.ts --outfile dist/tray.js --target browser --minify && bun run build:dashboard && cp -r ../cli/dashboard/build/* dist/ && cp tray.html dist/tray.html && cp capture.html dist/capture.html && cp search.html dist/search.html",
+    "build:ts": "node build-ts.mjs",
     "dev": "tauri dev",
     "build": "tauri build",
     "tauri": "tauri"


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(tray): cross-platform build script for Windows CI` (commit `9eeb4a1d`) now and I'll follow up shortly with feedback.

packages/tray/build-ts.mjs
```diff
@@ -0,0 +1,24 @@
...
+
+// Bundle TypeScript
+const { execSync } = await import("node:child_process");
+execSync("bun build src-ts/index.ts --outfile dist/tray.js --target browser --minify", { stdio: "inherit" });
+
+// Build dashboard
+execSync("bun run build:dashboard", { stdio: "inherit" });
+
+// Copy dashboard build output into dist
+cpSync("../cli/dashboard/build", "dist", { recursive: true });
```
NicholaiVogel: **Bug: directory vs. contents copy mismatch.**

`cpSync('../cli/dashboard/build', 'dist', {recursive: true})` copies the *directory itself* into `dist`, resulting in `dist/build/<files>` — not the same as the original `cp -r ../cli/dashboard/build/* dist/` which copies the *contents* directly into `dist/`.

Fix: copy the contents by enumerating them, or use:
```js
cpSync('../cli/dashboard/build', 'dist', { recursive: true, filter: () => true });
// Still wrong — need to copy children, not the dir itself
```
Correct approach:
```js
import { readdirSync } from 'node:fs';
import { join } from 'node:path';
const src = '../cli/dashboard/build';
for (const entry of readdirSync(src)) {
  cpSync(join(src, entry), join('dist', entry), { recursive: true });
}
```
Or simply target the children directly if the structure is known.

packages/tray/build-ts.mjs
```diff
@@ -0,0 +1,24 @@
...
+ * Cross-platform build script for tray TypeScript assets.
+ * Replaces Unix-only `rm -rf`, `cp -r` in package.json scripts.
+ */
+import { rmSync, cpSync, copyFileSync } from "node:fs";
+
+// Clean dist
+rmSync("dist", { recursive: true, force: true });
+
+// Bundle TypeScript
+const { execSync } = await import("node:child_process");
```
NicholaiVogel: **Top-level `await` with a dynamic import is unnecessary here.**

`child_process` is a built-in — it can be imported statically at the top of the file alongside the `fs` imports:
```js
import { execSync } from 'node:child_process';
```
The dynamic `await import(...)` requires the file to be treated as an async module, which works in Node ESM but is an unusual pattern for a build script and may confuse some tooling. Move it to the top-level static imports.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix(tray): cross-platform build script for Windows CI` (commit `7049b00c`); taking another look at the fixes and I will report back shortly.

aaf2tbz:
Updated in `27bee17c`:

1. **cpSync directory fix** — now iterates contents of `build/` and copies each entry into `dist/`, matching the original `cp -r build/* dist/` behavior. Verified locally — flat `dist/` output.

2. **Static import** — `execSync` moved to top-level static import.

3. **Added `icon.ico`** — Windows build requires `icons/icon.ico` for the executable resource file. Generated from `icon.png` with 256/128/64/48/32/16 sizes. Added to `tauri.conf.json` icon list.

Built and verified locally before pushing.

PR #264, opened 2026-03-21
fix(dashboard): revert broken constellation perf, apply safe-only fixes

aaf2tbz:
## Why

PR #256 shipped rendering optimizations to the Constellation tab (NodeColorCache, SpatialGrid, batched rendering, edge partitioning) that broke:
- Physics simulation feel (forces didn't settle correctly)
- Camera behavior (bouncing during slider adjustment, zoom-in on load)
- Slider interaction (recreated entire simulation per tick)
- Initial load (camera slammed from 0.15 to ~1.1 zoom in one frame)

Multiple fix attempts made things worse by changing tightly-coupled rendering/physics/camera code without understanding the dependencies.

## What

**Reverted to last working state** — Canvas2D and embedding-graph.ts restored to pre-#256 originals. Then applied only changes proven safe through live testing:

### Canvas2D (2 targeted changes)
1. **In-place physics update** — `updatePhysics` now modifies force parameters on the existing simulation instead of destroying and recreating it. Slider drags went from creating 20-50 full simulations to zero.
2. **Camera pause during slider interaction** — Sets `userAdjustedCamera = true` while dragging, resumes with clean fit 600ms after last change. Eliminates camera bounce.
3. **Default zoom 0.9** (was 0.15) — Reduces initial load camera snap from 700% to ~20%.

### EmbeddingsTab (data-only, zero rendering impact)
1. **`intersectFilterSets`** — iterates smaller set, no spread+filter+new Set
2. **`updateEmbeddingInState`** — in-place mutation, no 3× `.map()` + Map rebuild
3. **Single-pass stats** — pinnedIds + search index built in one loop
4. **Pre-built search index** — lowercase string concat at load time, not per keystroke
5. **Hydration race fix** — `initGraph` waits for `graphPhysicsHydrated` + `viewSettingsHydrated` before starting

### embedding-graph.ts
Fully restored to original. All additions (NodeColorCache, SpatialGrid, pre-computed color tables) removed.

## Lesson Learned

The constellation's rendering/physics/camera system is tightly coupled. Changing the draw loop, simulation parameters, or camera behavior has cascading effects that are hard to predict without the full dependency picture. Safe optimizations must be isolated to data-layer code that doesn't touch the render path.

## Test Plan

- [ ] Constellation loads without camera slam
- [ ] Physics sliders respond smoothly without camera bounce
- [ ] Entity overlay loads and settles correctly
- [ ] Node drag-to-move works
- [ ] Pan/zoom works
- [ ] Search filters correctly
- [ ] Pin/unpin updates in-place
- [ ] Custom physics settings persist across page refresh


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(dashboard): revert broken constellation perf, apply safe-only fixes` (commit `4d96be6f`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/embeddings/EmbeddingCanvas2D.svelte
```diff
@@ -241,20 +216,38 @@ export function startKnowledgeGraphSimulation(
...
+	const fx = simulation.force("x") as any;
+	if (fx) fx.strength(physics.centerForce);
+	const fy = simulation.force("y") as any;
+	if (fy) fy.strength(physics.centerForce);
+	// Pause camera auto-fit while user is dragging sliders — prevents bounce
+	userAdjustedCamera = true;
+	clearTimeout(physicsUpdateTimer);
+	physicsUpdateTimer = window.setTimeout(() => {
+		userAdjustedCamera = false;
+		fitCameraToBounds();
```
NicholaiVogel, line 248: `physicsUpdateTimer` is never cleared in the cleanup `$effect`. If the component unmounts while a slider debounce is in flight (within the 600 ms window), `fitCameraToBounds()` will fire on the destroyed component. `canvas` being `null` prevents a crash, but the timer still fires and produces confusing behaviour (e.g. if the component is rapidly remounted). Add `clearTimeout(physicsUpdateTimer)` to the cleanup effect alongside `cancelAnimationFrame(animFrame)`.

packages/cli/dashboard/src/lib/components/embeddings/EmbeddingCanvas2D.svelte
```diff
@@ -241,20 +216,38 @@ export function startKnowledgeGraphSimulation(
...
 		});
+	if (!userAdjustedCamera) fitCameraToBounds();
 }
 
+let physicsUpdateTimer = 0;
+
 export function updatePhysics(nextPhysics: GraphPhysicsConfig): void {
 	if (!simulation) return;
-	startSimulation(nodes, edges, nextPhysics);
+	const physics = clampGraphPhysics(nextPhysics);
```
NicholaiVogel, line 234: `updatePhysics` sets `.distance(physics.linkDistance)` and `.strength(physics.linkForce)` as **scalars** on the shared link force. When entity overlay is active, `startKnowledgeGraphSimulation` installs tier-aware *functions* for both distance and strength (e.g. entity→aspect = 50px, aspect→attribute = 30px). Calling `updatePhysics` while the KG simulation is running silently replaces those functions with a single scalar, collapsing the tier-aware layout. Consider guarding: if entity overlay is active, skip or no-op the `link.distance`/`link.strength` mutation (or re-apply the tier-aware closures).

packages/cli/dashboard/src/lib/components/embeddings/EmbeddingCanvas2D.svelte
```diff
@@ -309,8 +300,7 @@ function resizeCanvas(): void {
 	if (!rect || rect.width === 0) return;
```
NicholaiVogel, line 300: The `initialFitDone` guard was removed, so `resizeCanvas` now always calls `fitCameraToBounds()` when `!userAdjustedCamera`. For entity overlay the simulation starts from random positions; a window resize before the simulation settles will snap the camera to those random positions rather than waiting for the settled layout. The original guard (`initialFitDone`) prevented this. Consider reinstating a lightweight equivalent (e.g. track whether the simulation has fired its first `'end'` event, or only re-fit on resize if `simulation?.alpha() < 0.05`).

packages/cli/dashboard/src/lib/components/embeddings/EmbeddingCanvas2D.svelte
```diff
@@ -894,11 +851,14 @@ function setupInteractions(): void {
 			requestRedraw();
```
NicholaiVogel, line 851: Hover hit-detection (`findNodeAt`) is now called on every `pointermove` event and is O(n × 5) (five full passes over `nodes`, one per tier). The previous coalesced-hover approach had regressions, but for graphs at the 2200-node cap this generates ~11 000 iterations per mouse-move frame. If hover jank reappears on large graphs, consider a single-pass alternative that sorts into a local tier array once per frame and searches that, rather than five full array scans.

PR #265, opened 2026-03-21
fix(troubleshooter): handle daemon stop/restart lifecycle commands

aaf2tbz:
## Summary

- **Daemon stop/restart in the troubleshooter now produces real output** instead of dying mid-stream with a red error
- Lifecycle commands bypass the child-process exec pipeline — the daemon writes SSE events directly, closes the stream cleanly, then schedules its own graceful shutdown
- For restart, a detached `signet daemon start` is spawned before exit so the daemon comes back automatically; the frontend polls `/health` and shows reconnection status

## What was broken

The troubleshooter was running `signet daemon stop` and `signet daemon restart` through the same child-process pipeline as every other CLI command. But those commands send SIGTERM to the daemon — which is their own parent process. The SSE stream dies mid-flight and the UI just shows `error: fetch failed`.

## How it works now

**Backend (`daemon.ts`):**
1. `daemon-stop` and `daemon-restart` are intercepted before the general exec pipeline
2. Meaningful SSE events are written synchronously (started → stdout status → exit 0)
3. Stream is closed cleanly via `controller.close()`
4. After 500ms (response flush), `process.kill(process.pid, "SIGTERM")` triggers the existing `cleanup()` — PID file deletion, DB close, pipeline drain, watcher teardown
5. For restart: a detached `sleep 1 && signet daemon start` is spawned before SIGTERM, surviving the parent exit

**Frontend (`TroubleshooterPanel.svelte`):**
1. Connection errors during lifecycle commands show yellow "Daemon process ended" instead of red error (the drop is expected)
2. After restart, polls `/health` every 1s for up to 15s → shows `✓ Daemon restarted` or `✗ Daemon did not restart within 15s`

## Test plan

- [ ] Run "Stop Daemon" from troubleshooter — should show `Daemon stop initiated (PID xxx)` + `Dashboard will lose connection.` + `✓ exit 0`
- [ ] Run "Restart Daemon" from troubleshooter — should show restart initiated, then poll and show `✓ Daemon restarted`
- [ ] Verify PID file is cleaned up after stop (cleanup() runs via SIGTERM handler)
- [ ] Verify other troubleshooter commands (Status, Daemon Logs, etc.) still work normally
- [ ] Build passes (`bun run build`)

### DIFF
--- a/packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
+++ b/packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
@@ -170,6 +170,12 @@
 				return;
 			}
 
+			// res.ok means the daemon received and processed the command.
+			// For lifecycle commands the backend writes all SSE events
+			// synchronously, so if we got here the action is committed
+			// even if the TCP stream tears down before "started" arrives.
+			if (cmd.key === "daemon-restart") initiated = true;
+
 			const decoder = new TextDecoder();
 			let buf = "";
 


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(troubleshooter): handle daemon stop/restart lifecycle commands` (commit `8380ea87`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -7239,6 +7239,53 @@ app.post("/api/troubleshoot/exec", async (c) => {
...
+				}
+				write({ type: "exit", code: 0 });
+				try { controller.close(); } catch {}
+
+				// Give the response time to flush, then trigger graceful shutdown.
+				// SIGTERM triggers cleanup() which handles PID file, DB, watchers.
+				setTimeout(async () => {
+					if (key === "daemon-restart") {
+						const { spawn: nodeSpawn } = await import("node:child_process");
+						const child = nodeSpawn("sh", ["-c", `sleep 1 && ${resolved} daemon start`], {
```
NicholaiVogel: **Shell injection via `resolved`.**

```js
nodeSpawn("sh", ["-c", `sleep 1 && ${resolved} daemon start`], ...)
```

`resolved` is interpolated directly into a shell string. If the binary path contains spaces or shell metacharacters (e.g. `/home/user/my project/signet`), this breaks or can be exploited. Use the array form without a shell instead:

```js
nodeSpawn(resolved, ["daemon", "start"], {
  detached: true,
  stdio: "ignore",
  env: { ...baseEnv, SIGNET_NO_HOOKS: "1" } as NodeJS.ProcessEnv,
});
```

If a delay is still needed, keep the `sleep` subprocess separate or use a `setTimeout` before spawning (you're already inside one).

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -213,8 +213,31 @@
...
+					if (h.ok) { ok = true; break; }
+				} catch { /* still down */ }
 			}
+			appendLine(gid, ok
+				? "\x1b[32m✓ Daemon restarted\x1b[0m"
+				: "\x1b[31m✗ Daemon did not restart within 15s\x1b[0m");
+			await scrollGroup(gid);
 		}
 
 		appendLine(gid, "");
```
NicholaiVogel, line 251: **Restart poll runs unconditionally — false positives when the command never reached the daemon.**

The `if (cmd.key === "daemon-restart")` poll block is outside the try/catch and executes whether or not the request succeeded. If the endpoint returned a non-200 (caught earlier with an early `return`) this path is unreachable, but if `fetch` itself throws for an unrelated reason (DNS failure, endpoint not yet up, etc.) the catch block shows the yellow lifecycle message and then the poll runs anyway. Because the daemon was never actually asked to restart, if it happens to be responding, the poll returns `ok = true` and displays `✓ Daemon restarted` — a misleading false positive.

Fix: track whether the restart was initiated and only poll when it was:

```ts
let restartInitiated = false;
// ... inside stream read loop:
if (event.type === "started" && cmd.key === "daemon-restart") restartInitiated = true;
// ...
} catch (err) {
  if (!abort.signal.aborted) {
    const lifecycle = cmd.key === "daemon-stop" || cmd.key === "daemon-restart";
    if (lifecycle) {
      restartInitiated = cmd.key === "daemon-restart"; // expected drop
      appendLine(gid, `\x1b[33mDaemon process ended\x1b[0m`);
    } else { ... }
  }
}

if (restartInitiated) { /* poll */ }
```

packages/daemon/src/daemon.ts
```diff
@@ -7239,6 +7239,53 @@ app.post("/api/troubleshoot/exec", async (c) => {
...
+						controller.enqueue(encoder.encode(`data: ${JSON.stringify(event)}\n\n`));
+					} catch {}
+				};
+
+				write({ type: "started", key, command: `signet daemon ${action}` });
+				write({ type: "stdout", data: `Daemon ${action} initiated (PID ${process.pid})\n` });
+				if (key === "daemon-stop") {
+					write({ type: "stdout", data: "Dashboard will lose connection.\n" });
+				}
+				write({ type: "exit", code: 0 });
```
NicholaiVogel, line 7260: **500 ms may not be enough to flush the SSE response on slow or buffered connections.**

The `controller.close()` call enqueues the end-of-stream signal, but the runtime may not have flushed all bytes to the client within 500 ms (proxy buffering, slow TLS handshake, Bun's internal write coalescing). If SIGTERM fires first, the client never receives the `exit 0` event and falls into the error-catch path showing the yellow message instead of the clean exit line.

This is a best-effort situation, but 500 ms is on the low side. 1000 ms would give more headroom at negligible UX cost. Alternatively, emit a final `{type:"flush"}` sentinel, have the client close the reader when it sees it, and only then fire SIGTERM after the client acknowledges — though that requires a round-trip. A simple bump to 1000 ms is the pragmatic fix.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed — switched to array-form `nodeSpawn(resolved, ["daemon", "start"], ...)` with a nested `setTimeout` for the 1s delay instead of shell interpolation.

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
aaf2tbz: Fixed — added `initiated` flag that's set on `"started"` SSE event or in the lifecycle catch path. Poll only runs when `initiated` is true.

packages/daemon/src/daemon.ts
aaf2tbz: Bumped to 1000ms. Pragmatic fix as suggested.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix(troubleshooter): handle daemon stop/restart lifecycle commands` (commit `44b7f313`); taking another look at the fixes and I will report back shortly.

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -213,8 +215,32 @@
 			}
 		} catch (err) {
 			if (!abort.signal.aborted) {
```
NicholaiVogel, line 223: Minor: `initiated = true` is set in the catch block for *any* error on a lifecycle command, including a pre-connection `fetch()` failure (e.g., daemon already down when restart is clicked). In that scenario the 15-second poll will run and eventually print `✗ Daemon did not restart within 15s`, which is slightly confusing but not incorrect — the daemon never got the command. The previous unconditional poll was clearly worse; this is acceptable as-is. If you want tighter semantics, you could split the catch based on whether the error came from the initial `fetch()` call vs. the stream read (where the connection drop is expected), but that's a polish item, not a blocker.

aaf2tbz: Good catch — removed `initiated = true` from the catch block entirely. The `"started"` SSE event already sets it during stream processing. If fetch itself fails (daemon unreachable), `initiated` stays false and the poll is skipped.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix(troubleshooter): handle daemon stop/restart lifecycle commands` (commit `d367980e`); taking another look at the fixes and I will report back shortly.

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -213,8 +215,31 @@
...
```
NicholaiVogel, line 223: This removes the `initiated = true` that the previous review flagged as the logic-bug fix. The intent now is: only poll if the `started` SSE event was actually received, which is the reliable signal that the daemon got the command. This is defensible — the backend writes `started` synchronously as its first action and the 500ms SIGTERM delay gives ample time for the flush — but it does re-open a narrow race: if the TCP stream is torn down between the daemon writing `started` and the client receiving it, the restart poll silently skips and the user sees no reconnection feedback. If you want to close that window, the prior approach (set `initiated = true` in the catch block for lifecycle commands) was correct. If the current conservative behaviour (don't poll unless confirmed) is intentional, add a brief comment explaining the rationale so the next reviewer doesn't re-flag it.

aaf2tbz: Closed the race — `initiated` is now set after `res.ok` confirmation, before we even start reading the stream. Since the backend writes all SSE events synchronously in `start(controller)`, a successful HTTP response means the lifecycle action is already committed. This covers the TCP teardown window without needing the catch-block approach.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix(troubleshooter): handle daemon stop/restart lifecycle commands` (commit `0364ee2a`); taking another look at the fixes and I will report back shortly.

PR #266, opened 2026-03-21
feat(os): Visual GUI Agent — Page-Agent Integration

BusyBee3333:
## Visual GUI Agent for Signet OS Widgets

Adds AI cursor automation to Signet OS widgets. When a user types a mutation request (create, update, delete) in the OS chat, the LLM auto-detects it and triggers a visual agent that clicks through the widget UI in real-time while the user watches.

### Demo flow

1. User types: `add contact John Smith with email john@example.com`
2. OS Chat LLM detects this is a mutation → routes to visual agent
3. Contacts widget opens on the grid
4. AI cursor appears → clicks "+ New Contact" → fills firstName, lastName, email → clicks Save
5. User watches the whole thing happen live

### Architecture

```
User chat → LLM auto-detect (mutation vs read)
  ↓ (mutation)
Agent executor (daemon, os-agent.ts)
  ↓ SSE events
Dashboard (AgentChat.svelte)
  ↓ postMessage
Widget iframe (PageController + LiteMask cursor)
  ↓
DOM scanning → simplified HTML → LLM thinks → action executed → loop
```

- **PageController** from `@page-agent/page-controller` v1.6.0, bundled as pre-built IIFE (~60KB)
- **LiteMask** — lightweight cursor overlay with smooth animation + click ripples
- **No new npm deps** — the pre-built JS is inlined as a string constant
- **Every widget gets it automatically** — injected at `buildSrcdoc()` level

### Security

| Concern | Status |
|---------|--------|
| LLM in iframe | No — runs server-side in daemon only |
| API keys in iframe | No — keys stay in daemon |
| iframe sandbox | Unchanged — `allow-scripts` only |
| PageController scope | Widget DOM only — cannot access parent |
| postMessage origin | Validated — only responds to parent |
| New attack surface | DOM text sent to daemon for LLM — same data path as existing tool calls |

### Files (10 changed, +1234 / -19)

**New:**
- `packages/cli/dashboard/src/lib/components/os/page-agent-bundle.ts` — PageController + cursor IIFE
- `packages/daemon/src/routes/os-agent.ts` — Agent sessions, SSE stream, observe-think-act loop

**Modified:**
- `widget-theme.ts` — Injects page-agent script + bridge handlers
- `WidgetSandbox.svelte` — getDomState/executeAction/agentStart/agentStop
- `AgentChat.svelte` — Auto-detect agent mode, SSE relay, live status
- `os.svelte.ts` — Widget sandbox registry + agent session state
- `os-chat.ts` — useAgent/agentServerId LLM routing
- `widget-gen.ts` — Page-agent optimization in generator prompt
- `daemon.ts` + `logger.ts` — Route mounting + log category

### How to test

```bash
bun run build:dashboard
bun packages/daemon/src/daemon.ts
# Go to OS tab → type a mutation in chat → watch the cursor
```


NicholaiVogel:
Hi @BusyBee3333 - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(os): Visual GUI Agent — Page-Agent Integration` (commit `a52e35f1`) now and I'll follow up shortly with feedback.

packages/daemon/src/routes/os-chat.ts
```diff
@@ -0,0 +1,362 @@
...
+		for (const tool of probe.autoCard.tools) {
+			tools.push({
+				serverId: server.id,
+				serverName: probe.autoCard.name || server.name,
+				toolName: tool.name,
+				description: tool.description || "",
+				inputSchema: tool.inputSchema,
+			});
+		}
+	}
```
NicholaiVogel, line 96: **Bug — `require()` in ESM module will throw at runtime.**

`gatherAvailableTools()` uses `require('./marketplace-helpers.js')` but the rest of the file uses `import` statements, making this an ES module. `require` is not defined in ESM; this will throw `ReferenceError: require is not defined` the first time a chat message arrives.

Fix: either use a static `import` at the top of the file, or a dynamic `await import(...)` inside the function.

packages/cli/dashboard/src/lib/stores/os.svelte.ts
```diff
@@ -326,6 +326,30 @@ export function setActiveGroup(groupId: string | null): void {
...
+// Widget actions — sent from chat to trigger widget behavior (refresh, navigate, highlight)
+interface WidgetAction {
+	action: string;
+	data?: unknown;
+	_seq: number;
+}
+let widgetActions = $state<Map<string, WidgetAction>>(new Map());
+let actionSeq = 0;
+
+export function sendWidgetAction(serverId: string, action: string, data?: unknown): void {
```
NicholaiVogel, line 338: **Bug — race condition in `sendWidgetAction` auto-clear: a later action can be cleared by an earlier timeout.**

The timeout closure captures `actionSeq` by reference (it reads the module-level variable at fire time, not at scheduling time). If `sendWidgetAction` is called twice for the same `serverId` within 500 ms:

1. Call #1: `actionSeq` → 1; stores `{_seq:1}`; schedules timeout.
2. Call #2: `actionSeq` → 2; stores `{_seq:2}`; schedules timeout.
3. Timeout #1 fires: reads `widgetActions.get(serverId)._seq === 2`, `actionSeq === 2` → `true` → **deletes call #2's action early**.
4. Timeout #2 fires: map entry is gone → no-op.

Call #2 is silently dropped. The fix is to capture the sequence number in a local variable before scheduling the timeout:
```ts
const mySeq = ++actionSeq;
widgetActions.set(serverId, { action, data, _seq: mySeq });
setTimeout(() => {
  const current = widgetActions.get(serverId);
  if (current && current._seq === mySeq) widgetActions.delete(serverId);
}, 500);
```

packages/daemon/src/routes/os-agent.ts
```diff
@@ -0,0 +1,599 @@
...
+		if (!sessionId) {
+			return c.json({ error: "sessionId is required" }, 400);
+		}
+
+		const session = activeSessions.get(sessionId);
+		if (!session) {
+			return c.json({ error: "Session not found" }, 404);
+		}
+
+		// Resolve the pending DOM state promise
```
NicholaiVogel, line 489: **Security — agent execution endpoints have no authentication.**

`POST /api/os/agent-execute`, `POST /api/os/agent-state`, and `GET /api/os/agent-events` accept requests from any process with network access to the daemon. An attacker or malicious local process can:
- Start an agent session on any installed widget.
- Inject arbitrary DOM state into a pending `pendingDomState` promise, causing the LLM to act on fake DOM.
- Tap the SSE stream to observe all agent activity.

These endpoints should require the same auth token / session key used by the rest of the daemon API.

packages/daemon/src/routes/os-chat.ts
```diff
@@ -0,0 +1,362 @@
...
+				for (const call of parsed.toolCalls.slice(0, 5)) { // Max 5 tool calls
+					try {
+						logger.info("os-chat", `Calling tool ${call.serverId}/${call.toolName}`, {
+						args: JSON.stringify(call.args || {}).slice(0, 500),
+					});
+
+						// Call the tool via the marketplace /mcp/call endpoint internally
+						const callRes = await fetch(`http://127.0.0.1:${process.env.SIGNET_PORT || 3850}/api/marketplace/mcp/call`, {
+							method: "POST",
+							headers: { "Content-Type": "application/json" },
```
NicholaiVogel, line 277: **Security — LLM-chosen `serverId` and `toolName` are forwarded to an internal API without validation (prompt-injection → SSRF path).**

```ts
const callRes = await fetch(`http://127.0.0.1:${process.env.SIGNET_PORT || 3850}/api/marketplace/mcp/call`, {
  body: JSON.stringify({ serverId: call.serverId, toolName: call.toolName, args: call.args }),
});
```

The values `call.serverId` and `call.toolName` originate from the LLM response, which is based on user-supplied chat text. A crafted user message could cause prompt injection, leading the LLM to output a `serverId`/`toolName` outside the tools that were actually advertised in the system prompt. The internal `/api/marketplace/mcp/call` endpoint then receives and executes it.

Fix: after parsing the LLM response, verify that each `(serverId, toolName)` pair exists in the `tools` array returned by `gatherAvailableTools()` before executing.

packages/daemon/src/routes/os-agent.ts
```diff
@@ -0,0 +1,599 @@
...
+		session.status = "error";
+		session.error = msg;
+		logger.warn("os-agent", `Agent error: ${msg}`);
+
+		broadcastToSession(session, {
+			type: "error",
+			sessionId: session.id,
+			serverId: session.serverId,
+			data: { error: msg },
+		});
```
NicholaiVogel, line 316: **Logic bug — events emitted before the SSE client connects are silently dropped, breaking session start for fast LLMs.**

The agent loop is started with a 500 ms delay (`setTimeout(() => runAgentLoop(session), 500)`), but there is no guarantee the dashboard's `EventSource` has connected within that window, especially on slow networks. If `sseListeners` is empty when `agentStart` / first `getDomState` are broadcast, those events are lost and the session stalls waiting for a DOM state that was never requested.

Consider buffering the last N events per session so a newly-connected SSE client can replay missed events, or block the loop start until at least one listener is attached.

packages/cli/dashboard/src/app.html
```diff
@@ -2,7 +2,13 @@
 <html lang="en">
 	<head>
 		<meta charset="utf-8" />
-		<meta name="viewport" content="width=device-width, initial-scale=1" />
+		<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover">
```
NicholaiVogel, line 5: **Accessibility regression — `user-scalable=no` and `maximum-scale=1.0` prevent text zoom.**

WCAG 2.1 SC 1.4.4 requires that text can be resized to 200% without loss of content or functionality. Disabling pinch-to-zoom and `maximum-scale` breaks this for users who rely on zoom for readability. Consider removing `maximum-scale=1.0, user-scalable=no` or limiting it to the OS/fullscreen widget canvas only.

packages/cli/dashboard/src/lib/components/os/AgentChat.svelte
```diff
@@ -0,0 +1,810 @@
...
+		role: "user" | "agent";
+		content: string;
+		timestamp: number;
+		toolCalls?: ToolCall[];
+		openedWidget?: string;  // server ID of widget that was opened
+	}
+
+	let messages = $state<ChatMessage[]>([]);
+	let input = $state("");
+	let loading = $state(false);
```
NicholaiVogel, line 29: **Minor logic issue — message `timestamp` used as Svelte keyed-each key is not unique.**

`{#each messages as msg (msg.timestamp)}` — if two messages are pushed in the same millisecond (e.g., user push + immediate agent push after a fast tool call), they share the same key. Svelte will merge/skip one of them. Use an auto-incrementing `id` field instead.

NicholaiVogel:
### Code review

Found 14 issues, ranked by confidence score:

---

**Score 100 — `!` non-null assertion used in WidgetSandbox.svelte** ([CLAUDE.md](https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/CLAUDE.md) says \"no `!` -- check for null explicitly\")

`agentPending.get(data.id)!.resolve(data.result)` appears twice in the new `handleMessage` cases. The `has()` check before does not narrow the type — use `const p = agentPending.get(data.id); if (p) p.resolve(...)` instead.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/cli/dashboard/src/lib/components/os/WidgetSandbox.svelte#L1-L1

---

**Score 100 — `as` type assertions used 7+ times across os-chat.ts and os-agent.ts** ([CLAUDE.md](https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/CLAUDE.md) says \"no `as` -- fix the types instead of asserting\")

Examples: `await res.json() as { choices: Array<...> }`, `JSON.parse(cleaned) as AgentAction`, `as Record<string, unknown>` on DOM state. Should use `unknown` + type guards or a validation library instead.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-chat.ts#L1-L1

---

**Score 92 — Excessive try/catch (6 blocks in os-chat.ts)** ([CLAUDE.md](https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/CLAUDE.md) says \"avoid try/catch where possible\" and \"result types over exceptions\")

Two nested try/catches in `parseLlmResponse`, empty `catch {}` on the follow-up LLM call, plus per-tool-call catches. The accumulation loop should use a result type wrapping each fetch.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-chat.ts#L1-L1

---

**Score 88 — Race condition in `sendWidgetAction` — timeout closes over mutable state**

The auto-clear `setTimeout` captures the action reference rather than a sequence counter. If two calls happen within the 500ms window, the first timeout fires and clears the second call's state prematurely. Fix: capture a local counter before scheduling and compare on fire.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/cli/dashboard/src/lib/stores/os.svelte.ts#L1-L1

---

**Score 88 — SSE listener accumulation on reconnects in os-agent.ts**

When an EventSource reconnects, a new listener is appended without the dead one being removed. Between `broadcastToSession` steps, the `sseListeners` array is snapshotted via `.slice()` before `abort` fires, so dead listeners execute and throw silently.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-agent.ts#L1-L1

---

**Score 88 — LLM-generated HTML injected unsanitized into srcdoc with scripts enabled**

`buildSrcdoc(html, serverId)` in widget-theme.ts injects LLM output verbatim. The iframe has `sandbox="allow-scripts"`, so prompt-injected `<script>` tags can call `window.signet.callTool` with attacker-controlled arguments. The LLM system prompt instructs \"no external fetch calls\" but this is not enforced at runtime.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/cli/dashboard/src/lib/components/os/widget-theme.ts#L1-L1

---

**Score 85 — Non-unique `{#each}` key in AgentChat.svelte — uses `msg.timestamp`**

Agent SSE events can arrive within the same millisecond. Svelte uses the key for DOM diffing; collisions cause messages to be merged or dropped. Use an auto-incrementing `id` field assigned at push time instead.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/cli/dashboard/src/lib/components/os/AgentChat.svelte#L1-L1

---

**Score 75 — `require()` inside ESM function in os-chat.ts** ([CLAUDE.md](https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/CLAUDE.md) says \"Use Bun APIs when possible\")

`const { readInstalledServersPublic } = require(\"./marketplace-helpers.js\")` inside `gatherAvailableTools()` — Bun targets ESM and `require()` throws `ReferenceError` at runtime. The same file already uses `await import()` for dynamic loading elsewhere.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-chat.ts#L1-L1

---

**Score 75 — Missing auth middleware on new agent endpoints in os-agent.ts** (daemon auth pattern established in [packages/daemon/src/auth/middleware.ts](https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/auth/middleware.ts))

The new `/api/os/agent-*` routes — which start AI sessions, read DOM state, and stream commands — bypass `requirePermission` entirely. All other `/api/*` routes apply the auth middleware.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-agent.ts#L1-L1

---

**Score 75 — Path traversal via unsanitized `serverId` in widget-gen.ts**

`join(widgetDir(), \`${serverId}.html\`)` at line ~31 takes `serverId` from the HTTP request body with no sanitization. A value like `../../.ssh/authorized_keys` escapes the widget directory. Apply `basename()` or an allowlist pattern (`/^[a-zA-Z0-9_.-]+$/`) at the route boundary.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/widget-gen.ts#L1-L1

---

**Score 75 — `widgetGenerating` set never cleared on daemon-side failure**

When the daemon returns 202 (generating), the `finally` block keeps `serverId` in `widgetGenerating`, relying on a `widget.generated` SSE event to clear it. If the daemon fails and emits `widget.error` instead, no handler exists on the client — the spinner sticks with no recovery path.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/cli/dashboard/src/lib/stores/os.svelte.ts#L1-L1

---

**Score 75 — Prompt injection to arbitrary MCP tool calls in os-chat.ts**

`serverId` and `toolName` extracted from the LLM response are forwarded directly to `/api/marketplace/mcp/call` without validating they exist in the advertised tool list. A malicious widget description could cause the LLM to emit an unintended serverId/toolName pair.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-chat.ts#L1-L1

---

**Score 75 — Missing `icon-192.png` in static assets**

`app.html` adds `<link rel="apple-touch-icon" href=\"/icon-192.png\">` and `manifest.json` references it, but the file is absent from `packages/cli/dashboard/static/`. This produces a 404 on every mobile page load.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/cli/dashboard/src/app.html#L1-L1

---

**Score 75 — Duplicated OpenAI key resolution in os-chat.ts and os-agent.ts** ([CLAUDE.md](https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/CLAUDE.md) says \"Duplicate logic across multiple files is a code smell\")

Both files independently declare `let cachedApiKey: string | null = null` and implement the same `process.env.OPENAI_API_KEY || getSecret(...)` fallback. Extract to a shared helper.

https://github.com/Signet-AI/signetai/blob/a52e35f14a50c4551d0eb71190f09d2d28acfe18/packages/daemon/src/routes/os-agent.ts#L1-L1

---

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #269, opened 2026-03-22
fix: harden troubleshooter lifecycle commands

aaf2tbz:
## Summary

- **Critical race condition fix**: daemon restart spawned the replacement child *after* SIGTERM, but `cleanup()` → `process.exit()` killed the timer before it fired. Child now spawns **before** SIGTERM with a 300ms settle delay.
- **Abort on unmount**: TroubleshooterPanel now tracks in-flight AbortControllers and aborts all requests on component destroy — prevents stale state updates and leaked fetches during SSE streaming and health polling.
- **Confirmation dialog**: `daemon-stop` and `daemon-restart` now require explicit confirmation via a modal dialog (uses existing `danger` property on CommandDef).
- **13 structural tests**: command validation, SSE structure, spawn-before-kill ordering, and header cleanup — all verified against source.
- **Cleanup**: empty catch → debug logging, removed redundant CSS media query, removed forbidden `connection: keep-alive` header, extracted `HEALTH_PATH` constant.

## Test plan

- [x] `bun test packages/daemon/src/troubleshoot-exec.test.ts` — 13 pass
- [ ] Manual: trigger daemon restart from dashboard, verify daemon comes back within 15s
- [ ] Manual: navigate away during streaming command, verify no console warnings
- [ ] Manual: click "Stop Daemon", verify confirmation dialog appears before execution
- [ ] Manual: click "Restart Daemon", verify confirmation dialog with correct messaging

### DIFF
--- a/packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
+++ b/packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
@@ -8,7 +8,7 @@
 	import Info from "@lucide/svelte/icons/info";
 	import Terminal from "@lucide/svelte/icons/terminal";
 	import ChevronDown from "@lucide/svelte/icons/chevron-down";
-	import { tick } from "svelte";
+	import { tick, onDestroy } from "svelte";
 
 	type CmdKind = "cli" | "api";
 
@@ -28,6 +28,8 @@
 		readonly color: string;
 		readonly commands: readonly CommandDef[];
 	}
+
+	const HEALTH_PATH = "/health";
 
 	const groupDefs: readonly GroupDef[] = [
 		{
@@ -52,7 +54,7 @@
 			icon: Activity,
 			color: "var(--sig-highlight)",
 			commands: [
-				{ label: "Health", kind: "api", method: "GET", path: "/health" },
+				{ label: "Health", kind: "api", method: "GET", path: HEALTH_PATH },
 				{ label: "Diagnostics", kind: "api", method: "GET", path: "/api/diagnostics" },
 				{ label: "Pipeline", kind: "api", method: "GET", path: "/api/pipeline/status" },
 				{ label: "Predictor", kind: "api", method: "GET", path: "/api/predictor/status" },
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: harden troubleshooter lifecycle commands` (commit `11ad0bdc`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -377,6 +443,37 @@
 	{/each}
 </div>
 
+{#if pending}
+	<div
+		class="ts-overlay"
+		role="button"
```
NicholaiVogel, line 449: **Accessibility: `role="button"` on the backdrop `div` is semantically wrong.**

The outer `div` with `role="button"` wraps a `role="dialog"`, making the dialog's container appear as an interactive button to assistive technologies. Screen readers will announce the entire dialog contents as a button label.

Use a separate, visually full-screen backdrop element with `role="presentation"` (or no role) for the click-outside-to-dismiss, and keep the inner `role="dialog"` div standalone:
```svelte
<div class="ts-backdrop" role="presentation"
  onclick={(e) => { if (e.target === e.currentTarget) pending = null; }}
  onkeydown={(e) => { if (e.key === "Escape") pending = null; }}
  tabindex="-1">
  <div class="ts-dialog" role="dialog" aria-modal="true" aria-labelledby="ts-confirm-title">
    ...
  </div>
</div>
```
This is not a security issue but does affect users who rely on AT.

packages/daemon/src/troubleshoot-exec.test.ts
```diff
@@ -0,0 +1,124 @@
+/**
```
NicholaiVogel, line 1: **Test fragility: static source-text assertions give false assurance.**

All 13 tests work by reading `daemon.ts` as a raw string and asserting on substring/regex presence. They verify the *text* of the source, not runtime behavior. Risks:

1. A mechanical rename or reformat (e.g. `nodeSpawn` → `childSpawn`, reformatting the `setTimeout` block) breaks the tests without any functional regression.
2. The regex `extract(/app\.post\("\/api\/troubleshoot\/exec"[\s\S]*?\n\}\);/)` uses a non-greedy `[\s\S]*?` — if a blank line with `});` exists anywhere in the middle of the handler it will extract a truncated block and silently produce wrong assertions.
3. The spawn-ordering test asserts `spawnIdx < killIdx` by string offset within the extracted block — correct, but only as long as neither token appears elsewhere in the block.

This is low-risk for the immediate fix (the ordering property is real and the text reflects it), but consider at minimum an integration smoke-test that actually starts/mocks the handler and checks the SSE event sequence.

PR #270, opened 2026-03-22
fix: add auth middleware and zombie guard to troubleshooter

aaf2tbz:
## Summary

- **Security**: `/api/troubleshoot/*` endpoints had zero auth middleware while every other sensitive endpoint (`/api/diagnostics`, `/api/repair`, `/api/memory`) is gated. These endpoints can stop/restart the daemon and run arbitrary CLI commands. Now gated behind `requirePermission("admin")`.
- **Zombie guard**: The async restart path (`await import` + `spawn`) had no error handling — if either threw, SIGTERM never fired and the daemon would zombie (stream closed, response sent, process alive with no recovery). Wrapped in `try/finally` to guarantee termination.
- **Tests**: 2 new assertions covering auth middleware presence and try/finally ordering (15 total tests now).

Stacks on top of #269 (lifecycle hardening).

## Test plan

- [x] `bun test packages/daemon/src/troubleshoot-exec.test.ts` — 15 pass
- [ ] Verify unauthenticated POST to `/api/troubleshoot/exec` returns 401/403
- [ ] Verify authenticated admin request still executes normally

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -1510,6 +1510,11 @@ app.use("/api/repair/*", async (c, next) => {
 	return requirePermission("admin", authConfig)(c, next);
 });
 
+// Troubleshooter — admin only (can stop/restart daemon, run CLI commands)
+app.use("/api/troubleshoot/*", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
 // Per-memory PATCH and DELETE need method-specific guards + scope check
 app.use("/api/memory/:id", async (c, next) => {
 	// Scope enforcement on mutations: if token has project scope, verify
@@ -7267,17 +7272,21 @@ app.post("/api/troubleshoot/exec", async (c) => {
 				// is detached+unref'd so it survives parent exit. A short delay
 				// after spawn lets the fork commit before we self-terminate.
 				setTimeout(async () => {
-					if (key === "daemon-restart") {
-						const { spawn: nodeSpawn } = await import("node:child_process");
-						const child = nodeSpawn(resolved, ["daemon", "start"], {
-							detached: true,
-							stdio: "ignore",
-							env: { ...baseEnv, SIGNET_NO_HOOKS: "1" } as NodeJS.ProcessEnv,
-						});
-						child.unref();
+					try {
+						if (key === "daemon-restart") {
+							const { spawn: nodeSpawn } = await import("node:child_process");
+							const child = nodeSpawn(resolved, ["daemon", "start"], {
+								detached: true,
+								stdio: "ignore",
+								env: { ...baseEnv, SIGNET_NO_HOOKS: "1" } as NodeJS.ProcessEnv,
+							});
+							child.unref();
+						}
+					} finally {
+						// Guarantee SIGTERM fires even if spawn throws — without
+						// this the daemon zombies with stream closed but process alive.
+						setTimeout(() => process.kill(process.pid, "SIGTERM"), 300);
 					}
-					// Let the fork fully settle before cleanup tears everything down
-					setTimeout(() => process.kill(process.pid, "SIGTERM"), 300);
 				}, 1000);
 			},
 		});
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: add auth middleware and zombie guard to troubleshooter` (commit `999db79c`) now and I'll follow up shortly with feedback.

packages/daemon/src/troubleshoot-exec.test.ts
```diff
@@ -0,0 +1,149 @@
...
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+
+const SRC = readFileSync(join(__dirname, "daemon.ts"), "utf-8");
+
+// ---------------------------------------------------------------------------
+// Extract the route handler source for targeted assertions
+// ---------------------------------------------------------------------------
+
+function extract(pattern: RegExp): string {
```
NicholaiVogel, line 17: All 15 tests are source-text string assertions against `daemon.ts` — they verify that certain substrings exist in the file, not that the middleware actually intercepts requests or that the spawn/kill sequence runs correctly at runtime. A refactor that preserves the strings but breaks ordering (e.g. moving the middleware registration below the route definition) would pass every test. Consider adding at least one integration test that issues an unauthenticated HTTP request to `/api/troubleshoot/exec` and asserts a 401/403, even if it runs against a lightweight fixture rather than the full daemon.

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -377,6 +443,37 @@
 	{/each}
 </div>
 
+{#if pending}
+	<div
```
NicholaiVogel, line 447: `role="button"` on the backdrop overlay div is incorrect — it turns the full-screen dimmed layer into an interactive button for AT users, which conflicts with the `role="dialog"` child and produces confusing announcements. Use `role="presentation"` (or omit the role entirely) on the backdrop and handle keyboard dismissal only on the inner dialog. Also, there is no focus trap: once the dialog opens, Tab will cycle out of it into the panel behind. At minimum, move focus to the Cancel button on open and restore it on close.

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -169,6 +181,12 @@
 				return;
 			}
 
+			// res.ok means the daemon received and processed the command.
+			// For lifecycle commands the backend writes all SSE events
```
NicholaiVogel, line 185: `initiated = true` is set immediately after `res.ok` for `daemon-restart`, before any SSE event is consumed. The inline comment explains this is intentional because the server schedules the `setTimeout` synchronously inside `ReadableStream.start()`. That is currently true, but it's an implicit coupling between client and server: if the server-side SSE handler is ever made async (e.g. awaits a DB write before scheduling the restart), `res.ok` will no longer imply the restart is committed and polling will start for commands that were never executed. Consider setting `initiated = true` only when the `"started"` SSE event is received (already handled at line 207) and removing the early assignment, accepting that a stream tear-down before that event means no polling — which is safer than always polling.

packages/daemon/src/daemon.ts
```diff
@@ -7239,6 +7244,61 @@ app.post("/api/troubleshoot/exec", async (c) => {
...
+							const child = nodeSpawn(resolved, ["daemon", "start"], {
+								detached: true,
+								stdio: "ignore",
+								env: { ...baseEnv, SIGNET_NO_HOOKS: "1" } as NodeJS.ProcessEnv,
+							});
+							child.unref();
+						}
+					} finally {
+						// Guarantee SIGTERM fires even if spawn throws — without
+						// this the daemon zombies with stream closed but process alive.
```
NicholaiVogel, line 7287: The inner `setTimeout(() => process.kill(...), 300)` inside the `finally` block fires 300 ms after the outer 1000 ms delay, for a total of ~1.3 s from request receipt to SIGTERM. On a loaded system the SSE response may not have been flushed to the TCP send buffer within 1 s, so the client could miss the `exit` event and fall into the error branch. This is a best-effort heuristic and hard to make reliable without a flush callback, but worth documenting so future authors don't reduce the delays thinking they're safe.

Issue #15, opened 2026-02-19
test(openclaw): mixed-mode duplicate prevention suite


NicholaiVogel:
## Objective
Prove there are no duplicate actions when plugin and legacy paths are both configured.

## Scope
- Add mixed-mode scenarios for remember, recall, and capture events.
- Verify arbitration + idempotency protections.
- Assert duplicate incident counters remain zero.

## Acceptance Criteria
- Mixed-mode tests show no duplicate mutations or injections.
- Duplicate prevention is deterministic under retries.
- Satisfies guardrail in spec section 22.14.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 22.14, 18.6)

NicholaiVogel:
Bulk closed - work completed.

Issue #14, opened 2026-02-19
feat(connector-openclaw): keep legacy command-hook fallback mode


NicholaiVogel:
## Objective
Retain legacy command-hook path as compatibility fallback during plugin-first migration.

## Scope
- Keep command-hook files installable and functional when mode is `legacy-hook`.
- Gate fallback behavior behind explicit mode/arbitration.
- Document fallback expectations and limits.

## Acceptance Criteria
- Legacy `/remember`, `/recall`, `/context` continue to work in fallback mode.
- No overlap with plugin path when arbitration selects plugin.
- Behavior aligns with spec sections 26.3 and 26.6.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 26.3, 26.6)

NicholaiVogel:
Bulk closed - work completed.

Issue #13, opened 2026-02-19
feat(adapter-openclaw): align tool surface with canonical memory operations


NicholaiVogel:
## Objective
Align OpenClaw runtime tool surface with canonical memory operations and naming.

## Scope
- Ensure tool set includes: search, store, get, list, modify, forget.
- Align tool parameter semantics with daemon API contracts.
- Keep outputs machine-actionable and stable.

## Acceptance Criteria
- Tool contract docs/tests match runtime behavior.
- Tool set is complete and consistent with spec section 26.4.
- Existing flows remain backward compatible where required.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 26.4, 14 API delta)

NicholaiVogel:
Bulk closed - work completed.

Issue #12, opened 2026-02-19
feat(adapter-openclaw): add memory_modify and memory_forget tools


NicholaiVogel:
## Objective
Expose explicit memory mutation operations to OpenClaw agents via runtime tools.

## Scope
- Add `memory_modify` and `memory_forget` tools in adapter runtime surface.
- Route operations through daemon explicit mutation APIs.
- Ensure policy checks and audit fields are preserved.

## Acceptance Criteria
- Tools execute successfully with structured responses.
- Modify/forget operations write history/audit records.
- Pinned and policy safeguards are enforced.
- Matches spec sections 14.5-14.7 and 26.4.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 14.5-14.7, 26.4)

NicholaiVogel:
Bulk closed - work completed.

PR #271, opened 2026-03-22
fix: comprehensive security audit hardening

aaf2tbz:
## Summary

Addresses 14 of 15 findings from a security audit of the daemon, CLI, connectors, and tray app. One finding (#6 — scoped token read bypass) requires deeper design work and is tracked as a follow-up.

### Critical fixes
- **#1 Unauthenticated secret endpoints**: `/api/secrets/*` now gated behind `requirePermission("admin")` — previously allowed any localhost process to exec commands with injected secrets
- **#2 Unauthenticated git operations**: `/api/git/*` now admin-only — push, pull, remote config changes were wide open
- **#3 Unauthenticated 1Password**: covered by #1 — all 1P endpoints under `/api/secrets/*`
- **#4 Host header spoofing**: `isLocalhost()` now checks actual TCP `socket.remoteAddress` first, falls back to Host header only when socket unavailable
- **#5 Unrestricted CORS**: replaced `cors()` wildcard with explicit origin allowlist (localhost:3850, 5173, tauri)
- **#8 Shell injection in secret exec**: added metacharacter filter rejecting `;`, `|`, `$`, backticks, etc.
- **#9 Auth init race**: middleware now returns 503 if auth secret hasn't initialized yet

### High priority fixes
- **#7 Shell injection via OpenClaw restart**: parsed command as argv array instead of passing to `sh -c`
- **#10 Config writes**: admin auth + 1MB content size limit
- **#11 Git URL validation**: reject `file://` and bare paths, added `--single-branch`

### Medium fixes
- **#12**: Already fixed (checksum verification exists)
- **#13 TOCTOU config corruption**: added `atomicWriteJson()` to connector-base (write-to-tmp + rename), updated all 4 connectors
- **#14 Unsafe CSP**: removed `unsafe-inline`/`unsafe-eval` from script-src, restricted localhost to port 3850
- **#15 Symlink TOCTOU**: use `lstatSync` instead of `statSync` for source check

### Known follow-up
- **#6 Scoped token read bypass**: requires query-level scope injection into hybridRecall — needs its own design + PR

## Files changed (11)

| File | Changes |
|------|---------|
| `packages/daemon/src/daemon.ts` | Auth middleware for secrets, git, troubleshooter, config; CORS restriction; auth init guard; config size limit |
| `packages/daemon/src/auth/middleware.ts` | TCP peer address check in isLocalhost() |
| `packages/daemon/src/secrets.ts` | Shell metacharacter filter |
| `packages/cli/src/cli.ts` | Argv parsing for restart command; git URL validation |
| `packages/tray/src-tauri/tauri.conf.json` | Tightened CSP |
| `packages/core/src/symlinks.ts` | lstatSync for source check |
| `packages/connector-base/src/index.ts` | atomicWriteJson utility |
| `packages/connector-claude-code/src/index.ts` | Atomic config writes |
| `packages/connector-codex/src/index.ts` | Atomic config writes |
| `packages/connector-openclaw/src/index.ts` | Atomic config writes |
| `packages/connector-opencode/src/index.ts` | Atomic config writes |

## Test plan

- [x] `bun run build` — full build succeeds
- [x] `bun test` — 34 daemon tests pass (Bun crash on full suite is pre-existing)
- [x] `bun run typecheck` — only pre-existing opencode-plugin error
- [ ] Verify unauthenticated POST to `/api/secrets/exec` returns 403
- [ ] Verify unauthenticated POST to `/api/git/push` returns 403
- [ ] Verify CORS blocks requests from non-localhost origins
- [ ] Verify `execWithSecrets("echo foo; rm -rf /", ...)` rejects with metachar error
- [ ] Verify tray app loads without CSP violations on port 3850

### DIFF
--- a/packages/connector-base/src/index.ts
+++ b/packages/connector-base/src/index.ts
@@ -32,7 +32,7 @@
  * ```
  */
 
-import { existsSync, readFileSync, writeFileSync, renameSync } from "node:fs";
+import { existsSync, readFileSync, writeFileSync, renameSync, unlinkSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { randomBytes } from "node:crypto";
 import {
@@ -221,8 +221,13 @@ export abstract class BaseConnector {
 export function atomicWriteJson(path: string, data: unknown, indent: number | string = 2): void {
 	const content = `${JSON.stringify(data, null, indent)}\n`;
 	const tmp = join(dirname(path), `.${randomBytes(6).toString("hex")}.tmp`);
-	writeFileSync(tmp, content, "utf-8");
-	renameSync(tmp, path);
+	try {
+		writeFileSync(tmp, content, "utf-8");
+		renameSync(tmp, path);
+	} catch (err) {
+		try { unlinkSync(tmp); } catch {}
+		throw err;
+	}
 }
 
 // ============================================================================
--- a/packages/connector-base/src/index.ts
+++ b/packages/connector-base/src/index.ts
@@ -32,8 +32,9 @@
  * ```
  */
 
-import { existsSync, readFileSync } from "node:fs";
-import { join } from "node:path";
+import { existsSync, readFileSync, writeFileSync, renameSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { randomBytes } from "node:crypto";
 import {
 	buildSignetBlock,
 	stripSignetBlock,
@@ -209,6 +210,19 @@ export abstract class BaseConnector {
 	 * Get the path to the harness's main config file.
 	 */
 	abstract getConfigPath(): string;
+}
+
+// ============================================================================
+// Atomic file write — prevents TOCTOU corruption when multiple
+// connector runs race on the same config file. Writes to a temp file
+// then renames (atomic on POSIX, near-atomic on Windows).
+// ============================================================================
+
+export function atomicWriteJson(path: string, data: unknown, indent: number | string = 2): void {
+	const content = `${JSON.stringify(data, null, indent)}\n`;
+	const tmp = join(dirname(path), `.${randomBytes(6).toString("hex")}.tmp`);
+	writeFileSync(tmp, content, "utf-8");
+	renameSync(tmp, path);
 }
 
 // ============================================================================
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -10474,14 +10474,16 @@ async function main() {
 				if (bytes > REQUEST_BODY_LIMIT) {
 					aborted = true;
 					logger.warn("http", "Request body exceeded limit", { bytes, limit: REQUEST_BODY_LIMIT });
-					// Send a proper 413 before closing so clients get a structured
-					// error instead of ECONNRESET. Use res.end() to avoid aborting
-					// pipelined keep-alive requests on the same socket.
+					// Send a proper 413 then destroy the socket so Hono's handler
+					// doesn't try to write a second response (ERR_HTTP_HEADERS_SENT).
 					if (!res.headersSent) {
 						res.writeHead(413, { "Content-Type": "application/json" });
-						res.end(JSON.stringify({ error: "payload too large" }));
+						res.end(JSON.stringify({ error: "payload too large" }), () => {
+							// Destroy after the 413 is flushed — prevents Hono's
+							// pipeline from racing to write headers on a closed res.
+							req.socket?.destroy();
+						});
 					}
-					req.resume(); // drain remaining body to avoid backpressure
 				}
 			});
 		});
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -10479,8 +10479,9 @@ async function main() {
 					if (!res.headersSent) {
 						res.writeHead(413, { "Content-Type": "application/json" });
 						res.end(JSON.stringify({ error: "payload too large" }), () => {
-							// Destroy after the 413 is flushed — prevents Hono's
-							// pipeline from racing to write headers on a closed res.
+							// Destroy after flush so Hono's subsequent write fails
+							// silently on a closed socket rather than emitting
+							// ERR_HTTP_HEADERS_SENT on a finished response.
 							req.socket?.destroy();
 						});
 					}
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -1207,12 +1207,29 @@ function getDashboardPath(): string | null {
 // Create the Hono app
 export const app = new Hono();
 
-// Middleware
-app.use("*", cors());
+// Middleware — restrict CORS to localhost origins only
+const ALLOWED_ORIGINS = new Set([
+	"http://localhost:3850",
+	"http://127.0.0.1:3850",
+	"http://localhost:5173",
+	"http://127.0.0.1:5173",
+	"tauri://localhost",
+	"http://tauri.localhost",
+]);
+app.use("*", cors({
+	origin: (origin) => ALLOWED_ORIGINS.has(origin) ? origin : "",
+	credentials: true,
+}));
 
 // Auth middleware — reads from module-level authConfig/authSecret
 // which are initialized properly in main(). In local mode this is a no-op.
+// Guard: reject requests if auth is required but secret isn't initialized yet
+// (startup race between middleware registration and main() completing).
 app.use("*", async (c, next) => {
+	if (authConfig.mode !== "local" && !authSecret) {
+		c.status(503);
+		return c.json({ error: "server initializing" });
+	}
 	const mw = createAuthMiddleware(authConfig, authSecret);
 	return mw(c, next);
 });
@@ -1510,6 +1527,32 @@ app.use("/api/repair/*", async (c, next) => {
 	return requirePermission("admin", authConfig)(c, next);
 });
 
+// Secrets — admin only (can exec commands, exfiltrate secrets)
+app.use("/api/secrets", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+app.use("/api/secrets/*", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
+// Git operations — admin only (can push, change remotes)
+app.use("/api/git/*", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
+// Troubleshooter — admin only (can stop/restart daemon, run CLI commands)
+app.use("/api/troubleshoot/*", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
+// Config writes — admin only (can overwrite agent.yaml, AGENTS.md)
+app.use("/api/config", async (c, next) => {
+	if (c.req.method === "POST") {
+		return requirePermission("admin", authConfig)(c, next);
+	}
+	return next();
+});
+
 // Per-memory PATCH and DELETE need method-specific guards + scope check
 app.use("/api/memory/:id", async (c, next) => {
 	// Scope enforcement on mutations: if token has project scope, verify
@@ -1644,6 +1687,12 @@ app.post("/api/config", async (c) => {
 
 		if (!file || typeof content !== "string") {
 			return c.json({ error: "Invalid request" }, 400);
+		}
+
+		// Reject oversized payloads (1MB) to prevent disk exhaustion
+		const MAX_CONFIG_BYTES = 1_048_576;
+		if (content.length > MAX_CONFIG_BYTES) {
+			return c.json({ error: `Content exceeds ${MAX_CONFIG_BYTES} byte limit` }, 413);
 		}
 
 		if (file.includes("/") || file.includes("..")) {
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -10465,12 +10465,23 @@ async function main() {
 	const { createServer: nodeCreateServer } = await import("node:http");
 	const createBoundedServer: typeof nodeCreateServer = (...args: Parameters<typeof nodeCreateServer>) => {
 		const server = nodeCreateServer(...args);
-		server.on("request", (req) => {
+		server.on("request", (req, res) => {
 			let bytes = 0;
+			let aborted = false;
 			req.on("data", (chunk: Buffer) => {
+				if (aborted) return;
 				bytes += chunk.length;
 				if (bytes > REQUEST_BODY_LIMIT) {
-					req.destroy();
+					aborted = true;
+					logger.warn("http", "Request body exceeded limit", { bytes, limit: REQUEST_BODY_LIMIT });
+					// Send a proper 413 before closing so clients get a structured
+					// error instead of ECONNRESET. Use res.end() to avoid aborting
+					// pipelined keep-alive requests on the same socket.
+					if (!res.headersSent) {
+						res.writeHead(413, { "Content-Type": "application/json" });
+						res.end(JSON.stringify({ error: "payload too large" }));
+					}
+					req.resume(); // drain remaining body to avoid backpressure
 				}
 			});
 		});
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -1217,7 +1217,7 @@ const ALLOWED_ORIGINS = new Set([
 	"http://tauri.localhost",
 ]);
 app.use("*", cors({
-	origin: (origin) => ALLOWED_ORIGINS.has(origin) ? origin : "",
+	origin: (origin) => ALLOWED_ORIGINS.has(origin) ? origin : null,
 	credentials: true,
 }));
 
@@ -1546,8 +1546,14 @@ app.use("/api/troubleshoot/*", async (c, next) => {
 });
 
 // Config writes — admin only (can overwrite agent.yaml, AGENTS.md)
+// Pre-check content-length to reject oversized bodies before buffering
+const MAX_CONFIG_BYTES = 1_048_576;
 app.use("/api/config", async (c, next) => {
 	if (c.req.method === "POST") {
+		const len = Number(c.req.header("content-length") ?? 0);
+		if (len > MAX_CONFIG_BYTES) {
+			return c.json({ error: `payload exceeds ${MAX_CONFIG_BYTES} byte limit` }, 413);
+		}
 		return requirePermission("admin", authConfig)(c, next);
 	}
 	return next();
@@ -1689,10 +1695,9 @@ app.post("/api/config", async (c) => {
 			return c.json({ error: "Invalid request" }, 400);
 		}
 
-		// Reject oversized payloads (1MB) to prevent disk exhaustion
-		const MAX_CONFIG_BYTES = 1_048_576;
+		// Defense-in-depth: also check after parse (content-length can be absent)
 		if (content.length > MAX_CONFIG_BYTES) {
-			return c.json({ error: `Content exceeds ${MAX_CONFIG_BYTES} byte limit` }, 413);
+			return c.json({ error: `content exceeds ${MAX_CONFIG_BYTES} byte limit` }, 413);
 		}
 
 		if (file.includes("/") || file.includes("..")) {
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -1546,12 +1546,14 @@ app.use("/api/troubleshoot/*", async (c, next) => {
 });
 
 // Config writes — admin only (can overwrite agent.yaml, AGENTS.md)
-// Pre-check content-length to reject oversized bodies before buffering
+// Reject oversized bodies: content-length fast path + post-buffer check
+// for chunked encoding. The server-level REQUEST_BODY_LIMIT caps all
+// routes, but this gives a clear per-route error message.
 const MAX_CONFIG_BYTES = 1_048_576;
 app.use("/api/config", async (c, next) => {
 	if (c.req.method === "POST") {
-		const len = Number(c.req.header("content-length") ?? 0);
-		if (len > MAX_CONFIG_BYTES) {
+		const cl = c.req.header("content-length");
+		if (cl && Number(cl) > MAX_CONFIG_BYTES) {
 			return c.json({ error: `payload exceeds ${MAX_CONFIG_BYTES} byte limit` }, 413);
 		}
 		return requirePermission("admin", authConfig)(c, next);
@@ -10457,12 +10459,30 @@ async function main() {
 	initFeatureFlags(AGENTS_DIR);
 	startUpdateTimer();
 
-	// Start HTTP server
+	// Start HTTP server with global body size limit (10MB) to prevent
+	// OOM from chunked-encoding requests that bypass content-length checks.
+	const REQUEST_BODY_LIMIT = 10 * 1_048_576;
+	const { createServer: nodeCreateServer } = await import("node:http");
+	const createBoundedServer: typeof nodeCreateServer = (...args: Parameters<typeof nodeCreateServer>) => {
+		const server = nodeCreateServer(...args);
+		server.on("request", (req) => {
+			let bytes = 0;
+			req.on("data", (chunk: Buffer) => {
+				bytes += chunk.length;
+				if (bytes > REQUEST_BODY_LIMIT) {
+					req.destroy();
+				}
+			});
+		});
+		return server;
+	};
+
 	serve(
 		{
 			fetch: app.fetch,
 			port: PORT,
 			hostname: BIND_HOST,
+			createServer: createBoundedServer,
 		},
 		(info) => {
 			logger.info("daemon", "Server listening", {
--- a/packages/daemon/src/secrets.ts
+++ b/packages/daemon/src/secrets.ts
@@ -222,15 +222,23 @@ export function deleteSecret(name: string): boolean {
 	return true;
 }
 
+// Reject commands containing shell metacharacters that could enable
+// injection when passed to sh -c. The endpoint is admin-gated, but
+// defense-in-depth prevents a compromised token from pivoting.
+const SHELL_META = /[;&|`$(){}[\]<>!\\]/;
+
 /**
  * Spawn a subprocess with one or more secrets injected as environment
  * variables. The agent only supplies references (env var names), never
  * the actual values.
  *
- * @param command  Shell command string to execute
+ * @param command  Shell command string to execute (no shell metacharacters allowed)
  * @param secretRefs  Map of env var name → secret name, e.g. { OPENAI_API_KEY: "OPENAI_API_KEY" }
  */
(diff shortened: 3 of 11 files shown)


aaf2tbz:
## Security Audit Summary

### What prompted this

A full security audit of the Signet daemon, CLI, connectors, and tray app identified 15 vulnerabilities ranging from unauthenticated RCE to TOCTOU race conditions. The daemon had grown organically and several new endpoint groups (secrets, git, troubleshooter) shipped without auth middleware — while older endpoints like `/api/memory/*` and `/api/diagnostics` were properly gated.

### The core problem

The daemon binds to `localhost:3850` by default, which provides some isolation, but localhost-only is not a security boundary:
- Any local process (malware, rogue browser extension) can call any endpoint
- Browser-based CSRF works because CORS was `*` (wildcard)
- Host header spoofing in hybrid mode bypassed token validation entirely
- The secrets exec endpoint accepted arbitrary shell commands with zero auth

### What this PR does

**Auth layer** (5 fixes): Added `requirePermission("admin")` middleware to `/api/secrets/*`, `/api/git/*`, `/api/troubleshoot/*`, and `POST /api/config`. Added a startup guard so requests during auth initialization get 503 instead of silently bypassing.

**Network layer** (2 fixes): Replaced wildcard CORS with an explicit origin allowlist. Fixed `isLocalhost()` to check the actual TCP socket `remoteAddress` instead of the spoofable Host header.

**Input validation** (3 fixes): Shell metacharacter filter on `execWithSecrets()`, argv-based parsing for the OpenClaw restart command (no more `sh -c`), and git URL scheme validation rejecting `file://` paths.

**Filesystem** (2 fixes): `lstatSync` instead of `statSync` for symlink source checks (prevents TOCTOU via symlink substitution). Atomic JSON writes via write-to-tmp + rename across all 4 connectors.

**Tray app** (1 fix): Tightened CSP — removed `unsafe-inline`/`unsafe-eval` from script-src, restricted localhost to the specific daemon port.

### What's NOT in this PR

**#6 — Scoped token read-access bypass**: Scoped tokens (project/agent/user restricted) only enforce scope on mutations, not reads. A compromised agent token can read all memories across every project. Fixing this requires injecting scope filters into `hybridRecall()` at the query level — too invasive for a hardening PR. Needs its own design pass before team mode ships.

**#12 — Binary checksum verification**: Already implemented (lines 410-435 in cli.ts). No action needed.

### How to verify

The auth changes are the most impactful. Quick smoke test:
```bash
# Should return 403 (was 200 before this PR)
curl -X POST http://localhost:3850/api/secrets/exec \
  -H "Content-Type: application/json" \
  -d '{"command":"echo hi","secrets":{"X":"X"}}'

# Should return 403
curl -X POST http://localhost:3850/api/git/push

# CORS: should be blocked by browser (no Access-Control-Allow-Origin)
# from any origin other than localhost:3850/5173/tauri
```

The shell metacharacter filter:
```bash
# Should fail with "disallowed shell metacharacters"
curl -X POST http://localhost:3850/api/secrets/exec \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <admin-token>" \
  -d '{"command":"echo foo; rm -rf /","secrets":{"X":"X"}}'
```

NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: comprehensive security audit hardening` (commit `5f0a66cb`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -1207,12 +1207,29 @@ function getDashboardPath(): string | null {
...
 export const app = new Hono();
 
-// Middleware
-app.use("*", cors());
+// Middleware — restrict CORS to localhost origins only
+const ALLOWED_ORIGINS = new Set([
+	"http://localhost:3850",
+	"http://127.0.0.1:3850",
+	"http://localhost:5173",
+	"http://127.0.0.1:5173",
```
NicholaiVogel, line 1215: **CORS `origin` callback returns `""` for disallowed origins — should return `null` or `false`.**

```ts
origin: (origin) => ALLOWED_ORIGINS.has(origin) ? origin : "",
```

Hono's cors middleware checks if the returned value is truthy before writing the `Access-Control-Allow-Origin` header. An empty string `""` is falsy in JS so Hono _likely_ omits the header, but this is an implementation detail — some versions emit `Access-Control-Allow-Origin: ` (empty string), which browsers reject but which is semantically incorrect and may confuse proxies or logs. Use:
```ts
origin: (origin) => ALLOWED_ORIGINS.has(origin) ? origin : null,
```

packages/daemon/src/daemon.ts
```diff
@@ -1646,6 +1689,12 @@ app.post("/api/config", async (c) => {
 			return c.json({ error: "Invalid request" }, 400);
 		}
 
+		// Reject oversized payloads (1MB) to prevent disk exhaustion
```
NicholaiVogel: **Config body size limit fires after the request body is already fully buffered.**

Hono (via `@hono/node-server`) buffers the entire body before your handler runs. A client that streams a 512 MB body will exhaust the daemon's heap before `content.length > MAX_CONFIG_BYTES` is checked. The auth middleware for this route is already in place, so the surface is admin-only, but a compromised token can still OOM the process.

Fix: add a body-size middleware before the route group, e.g.:
```ts
app.use("/api/config", async (c, next) => {
  const len = Number(c.req.header("content-length") ?? 0);
  if (len > 1_048_576) return c.json({ error: "payload too large" }, 413);
  return next();
});
```
The `content-length` check is not foolproof (chunked encoding), but pairing it with a streaming body-size cap at the Node.js HTTP server level would fully close the gap.

packages/connector-base/src/index.ts
```diff
@@ -211,6 +212,19 @@ export abstract class BaseConnector {
...
+// ============================================================================
+// Atomic file write — prevents TOCTOU corruption when multiple
+// connector runs race on the same config file. Writes to a temp file
+// then renames (atomic on POSIX, near-atomic on Windows).
+// ============================================================================
+
+export function atomicWriteJson(path: string, data: unknown, indent: number | string = 2): void {
+	const content = `${JSON.stringify(data, null, indent)}\n`;
+	const tmp = join(dirname(path), `.${randomBytes(6).toString("hex")}.tmp`);
+	writeFileSync(tmp, content, "utf-8");
```
NicholaiVogel: **`atomicWriteJson` leaks `.tmp` file on `renameSync` failure.**

If `writeFileSync` succeeds but `renameSync` throws (e.g., cross-device link, permission error), the `.randomhex.tmp` file is left in the directory with the full config content. This is a minor robustness issue — the temp file has the same ACL as the directory so it's not a new privilege escalation vector, but it can accumulate stale files.

Consider wrapping in try/finally:
```ts
try {
  writeFileSync(tmp, content, "utf-8");
  renameSync(tmp, path);
} catch (err) {
  try { unlinkSync(tmp); } catch { /* best-effort */ }
  throw err;
}
```

NicholaiVogel (commented):
14 critical/high/medium audit findings addressed. The core authentication gates, CORS restriction, TOCTOU fixes, and symlink hardening are all correct. Three issues remain:
1. `execWithSecrets` still routes commands through `sh -c` after the metacharacter filter — glob chars `*` and `?` and tilde `~` still expand in the shell, inconsistent with the argv-array approach used for the restart command fix.
2. Hono `cors()` `origin` callback returns `""` (empty string) for disallowed origins; depending on Hono version, this may emit `Access-Control-Allow-Origin: ` (empty header value) instead of omitting the header entirely — use `null` or `false` to be safe.
3. The 1 MB config-body guard fires inside the POST handler after the body has already been fully buffered into memory; a 512 MB payload will OOM the process before the check runs.

### Confidence: 8.0/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 8
}
```

- packages/daemon/src/secrets.ts:285 - **Shell still invoked after metacharacter filter — `*`, `?`, `~` still expand.**

The filter rejects `;|$` etc., but the command is then passed verbatim to `spawn('sh', ['-c', command])`. Any command that clears the filter is still interpreted by a shell, so glob expansion (`*`, `?`) and tilde expansion (`~`) remain live. Example: `echo ~/secret-*` passes the filter and leaks path info.

The restart-command fix (cli.ts) used argv parsing to eliminate the shell entirely. The same approach works here:
```ts
const argv = command.match(/(?:[^\s"']+|"[^"]*"|'[^']*')+/g);
if (!argv?.length) return { stdout: "", stderr: "empty command", code: 1 };
const cmd = argv.map(a => a.replace(/^["']|["']$/g, ""));
const proc = spawn(cmd[0], cmd.slice(1), { env: {...}, stdio: 'pipe', windowsHide: true });
```
This eliminates the shell entirely and makes the metacharacter filter a belt-and-suspenders check rather than the primary defense.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: comprehensive security audit hardening` (commit `eb351467`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in eb35146 — returns `null` now instead of empty string.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in eb35146 — added `content-length` pre-check middleware on `/api/config` that rejects before buffering. In-handler check kept as defense-in-depth for chunked encoding.

packages/connector-base/src/index.ts
aaf2tbz: Fixed in eb35146 — wrapped in try/catch, `unlinkSync(tmp)` on failure.

NicholaiVogel:

Two of three prior issues are now resolved: the CORS `origin` callback correctly returns `null` (not `""`) for disallowed origins, and the config-body size guard now fires at the middleware layer via `content-length` pre-check with a post-parse defense-in-depth check retained. The `atomicWriteJson` temp-file cleanup on error is a nice correctness fix. One prior issue remains open: `execWithSecrets` still routes through `sh -c` after the metacharacter filter, and the filter does not cover glob chars (`*`, `?`) or tilde (`~`), leaving shell expansion active. No new issues were introduced by the incremental diff.

### Confidence: 8.0/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

aaf2tbz:
All review feedback addressed across two commits:

- **eb35146**: CORS returns `null`, config content-length pre-check, atomicWriteJson tmp cleanup
- **779e543**: `execWithSecrets` now uses direct argv execution — shell eliminated entirely. `spawn(cmd[0], cmd.slice(1), ...)` instead of `spawn('sh', ['-c', command])`. Metacharacter filter kept as belt-and-suspenders for early rejection.

The glob/tilde expansion vector Nicholai flagged is fully closed — no shell interpretation at all now.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: comprehensive security audit hardening` (commit `779e5436`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/secrets.ts
```diff
@@ -249,10 +268,8 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
 		return out;
```
NicholaiVogel, line 268: **Fixed (prior issue #1):** `execWithSecrets` now uses `spawn(cmd[0], cmd.slice(1), ...)` instead of `sh -c`. Glob (`*`, `?`) and tilde expansion no longer apply, and the `SHELL_META` filter is now correctly described as belt-and-suspenders rather than the primary mitigation. Change is correct.

packages/daemon/src/daemon.ts
```diff
@@ -1510,6 +1527,38 @@ app.use("/api/repair/*", async (c, next) => {
...
+app.use("/api/secrets/*", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
+// Git operations — admin only (can push, change remotes)
+app.use("/api/git/*", async (c, next) => {
+	return requirePermission("admin", authConfig)(c, next);
+});
+
+// Troubleshooter — admin only (can stop/restart daemon, run CLI commands)
```
NicholaiVogel, line 1543: **Residual (prior issue #3, partially mitigated):** The `content-length` middleware guard correctly short-circuits before auth and body parsing for requests that include the header. However, chunked-transfer-encoding requests (`Transfer-Encoding: chunked`, no `Content-Length`) will still cause Hono to buffer the full body before the post-parse check at line ~1700 fires. Given the admin auth gate and the 1 MB post-parse check, this is low-severity, but worth noting: a future hardening step would be to set `bodyLimit` on the Hono app itself (e.g., `app.use('*', bodyLimit(MAX_CONFIG_BYTES))`) which applies before any body parsing regardless of encoding.

packages/tray/src-tauri/tauri.conf.json
```diff
@@ -26,7 +26,7 @@
     ],
     "security": {
       "capabilities": ["default"],
-      "csp": "default-src 'self' ipc: http://ipc.localhost http://localhost:*; script-src 'self' 'unsafe-inline' 'unsafe-eval' http://localhost:*; style-src 'self' 'unsafe-inline' http://localhost:* https://fonts.googleapis.com; img-src 'self' http://localhost:* data: blob:; font-src 'self' http://localhost:* https://fonts.gstatic.com data:; connect-src ipc: http://ipc.localhost http://localhost:* ws://localhost:* https://fonts.googleapis.com https://fonts.gstatic.com; frame-src http://localhost:*"
+      "csp": "default-src 'self' ipc: http://ipc.localhost; script-src 'self' http://localhost:3850; style-src 'self' 'unsafe-inline' http://localhost:3850 https://fonts.googleapis.com; img-src 'self' http://localhost:3850 data: blob:; font-src 'self' http://localhost:3850 https://fonts.gstatic.com data:; connect-src ipc: http://ipc.localhost http://localhost:3850 ws://localhost:3850 https://fonts.googleapis.com https://fonts.gstatic.com; frame-src http://localhost:3850"
```
NicholaiVogel, line 29: `style-src` still contains `'unsafe-inline'`. The PR description correctly scopes the removal to `script-src` only — inline styles carry far lower XSS risk than inline scripts. Acceptable as-is; future hardening could adopt a nonce-based approach if the frontend is refactored.

NicholaiVogel (dismissed):
All three prior-review blockers are now resolved: (1) execWithSecrets correctly uses direct argv spawn instead of sh -c — glob/tilde expansion is eliminated; (2) the CORS origin callback was already returning null for disallowed origins (correctly omits the header); (3) the content-length pre-check in middleware fires before Hono buffers the body, mitigating the OOM risk for clients that send the header. One residual low-severity concern remains: chunked-transfer-encoding requests or clients that omit Content-Length still cause full body buffering before the post-parse check. This is acceptable as defense-in-depth given the admin auth gate and the 1 MB parse check. The security posture of this PR is materially improved and the incremental diff is minimal and correct.

### Confidence: 8.3/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 929dab9 — added a 10MB global body limit at the Node.js HTTP server level. Custom `createServer` wraps the raw `IncomingMessage` stream, tracks bytes on `data` events, and calls `req.destroy()` if the limit is exceeded — before Hono buffers it into memory. This closes the chunked-encoding OOM vector completely.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: comprehensive security audit hardening` (commit `929dab9d`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -10403,12 +10459,30 @@ async function main() {
...
+	const { createServer: nodeCreateServer } = await import("node:http");
+	const createBoundedServer: typeof nodeCreateServer = (...args: Parameters<typeof nodeCreateServer>) => {
+		const server = nodeCreateServer(...args);
+		server.on("request", (req) => {
+			let bytes = 0;
+			req.on("data", (chunk: Buffer) => {
+				bytes += chunk.length;
+				if (bytes > REQUEST_BODY_LIMIT) {
+					req.destroy();
+				}
```
NicholaiVogel, line 10488: **Previously unresolved → now fixed.** The global `REQUEST_BODY_LIMIT` correctly handles chunked-encoding requests that omit `Content-Length`, closing the OOM path identified in prior review.

packages/daemon/src/daemon.ts
```diff
@@ -10403,12 +10459,30 @@ async function main() {
...
+	// OOM from chunked-encoding requests that bypass content-length checks.
+	const REQUEST_BODY_LIMIT = 10 * 1_048_576;
+	const { createServer: nodeCreateServer } = await import("node:http");
+	const createBoundedServer: typeof nodeCreateServer = (...args: Parameters<typeof nodeCreateServer>) => {
+		const server = nodeCreateServer(...args);
+		server.on("request", (req) => {
+			let bytes = 0;
+			req.on("data", (chunk: Buffer) => {
+				bytes += chunk.length;
+				if (bytes > REQUEST_BODY_LIMIT) {
```
NicholaiVogel, line 10474: **Non-blocking / future hardening:** `req.destroy()` terminates the raw TCP socket without writing an HTTP 413 response. Clients receive `ECONNRESET` instead of a structured error. For a localhost daemon this is acceptable, but if you want observability (e.g. to distinguish oversized-payload drops from network errors in logs), consider using `req.socket.end()` after writing a minimal `HTTP/1.1 413 Payload Too Large\r\n\r\n` response before destroying.

NicholaiVogel, line 10474: **Non-blocking:** `req.destroy()` destroys the underlying socket, which will also abort any pipelined or keep-alive requests multiplexed on the same TCP connection. In practice, localhost keep-alive clients (e.g. the CLI) could lose in-flight requests if they happen to share a connection with an oversized payload. Adding `{ allowHalfOpen: false }` to the `nodeCreateServer` options or using `req.socket.destroy()` only after draining the oversized body avoids this. Low priority given the admin-auth gate.

NicholaiVogel (commented):
All three prior-review blockers remain resolved. The incremental diff adds a global 10 MB body-size guard via a custom `createBoundedServer` factory, directly addressing the residual low-severity concern from the second review about chunked-encoding requests bypassing content-length checks. Two minor non-blocking observations: (1) `req.destroy()` terminates the TCP socket without sending an HTTP 413 response, so clients receive a connection-reset rather than a structured error; (2) socket destruction also cancels any pipelined/keep-alive requests on the same connection. Both are acceptable trade-offs for a localhost-bound daemon with admin auth already gating sensitive routes.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 8
}
```

- packages/daemon/src/daemon.ts:10558 - **Fix confirmed correct:** The content-length pre-check was changed from `Number(header ?? 0)` to only check when the header is present (`if (cl && Number(cl) > MAX_CONFIG_BYTES)`). The old form was already a no-op for absent headers (0 > 1_048_576 is false), but the new form is clearer and consistent with the intent. No behavioral regression.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 5c1c7f2 — now writes a proper `413 Payload Too Large` JSON response before draining. Uses `res.end()` instead of `req.destroy()` so keep-alive peers on the same socket aren't aborted. Added `logger.warn` for observability.

aaf2tbz: Fixed in 5c1c7f2 — `req.resume()` drains remaining body, `res.end()` closes cleanly without destroying the underlying socket. Keep-alive peers preserved.

packages/daemon/src/daemon.ts
aaf2tbz: Acknowledged — server-level limit confirmed working.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: comprehensive security audit hardening` (commit `5c1c7f2c`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -10403,12 +10459,41 @@ async function main() {
...
+				if (aborted) return;
+				bytes += chunk.length;
+				if (bytes > REQUEST_BODY_LIMIT) {
+					aborted = true;
+					logger.warn("http", "Request body exceeded limit", { bytes, limit: REQUEST_BODY_LIMIT });
+					// Send a proper 413 before closing so clients get a structured
+					// error instead of ECONNRESET. Use res.end() to avoid aborting
+					// pipelined keep-alive requests on the same socket.
+					if (!res.headersSent) {
+						res.writeHead(413, { "Content-Type": "application/json" });
```
NicholaiVogel, line 10480: Low-severity / future hardening: after `res.end()` is called here, Hono's own request listener is still in flight. When it finishes reading the drained body and attempts to write a response, Node.js will emit `ERR_HTTP_HEADERS_SENT`. This is non-fatal (Hono catches it internally or it surfaces as an unhandled promise rejection), but it can produce noisy log output. If this becomes a problem, consider aborting the Hono pipeline explicitly — e.g. by calling `req.socket.destroy()` after `res.end()` (accepting the keep-alive trade-off) or by setting a flag on the socket that Hono middleware can inspect to bail out early.

NicholaiVogel (dismissed):
Incremental diff addresses both non-blocking observations from the third review: `req.destroy()` is replaced with a proper `res.writeHead(413)` + `res.end()` response, and `req.resume()` prevents backpressure on the remaining body without terminating the socket. The `aborted` flag correctly suppresses duplicate handling on subsequent data chunks. One residual low-severity note: after `res.end()` is called at the raw Node layer, Hono's concurrently-running request handler will eventually attempt to write its own response to an already-finished socket, which may emit an unhandled `ERR_HTTP_HEADERS_SENT` error. This is architecturally inherent to any body-limit shim layered beneath a framework handler and is non-fatal for a localhost daemon. All 14 audit findings remain fixed. Approved.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 8
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 877f414 — `req.socket?.destroy()` called in the `res.end()` callback after the 413 is flushed. Hono's pipeline never gets to write headers on the dead socket.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: comprehensive security audit hardening` (commit `877f4146`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -10403,12 +10459,43 @@ async function main() {
...
+					// doesn't try to write a second response (ERR_HTTP_HEADERS_SENT).
+					if (!res.headersSent) {
+						res.writeHead(413, { "Content-Type": "application/json" });
+						res.end(JSON.stringify({ error: "payload too large" }), () => {
+							// Destroy after the 413 is flushed — prevents Hono's
+							// pipeline from racing to write headers on a closed res.
+							req.socket?.destroy();
+						});
+					}
+				}
```
NicholaiVogel, line 10488: The comment says this 'prevents Hono's pipeline from racing to write headers on a closed res' — but it doesn't prevent the race; it changes the failure mode. Hono's concurrent handler still runs and will attempt to write a response. Before this change (`req.resume()`, no destroy), that write produced `ERR_HTTP_HEADERS_SENT` on an already-finished response. After this change (socket destroyed), that write produces a write-after-destroy error logged by Node. Both are non-fatal for a localhost daemon behind admin auth. Consider updating the comment: 'Destroy after flush so Hono's subsequent write fails silently on a closed socket rather than emitting ERR_HTTP_HEADERS_SENT on a finished response.'

NicholaiVogel (dismissed):
Incremental diff is a one-hunk change to the 10 MB body-limit guard in `createBoundedServer`. All 14 security audit findings fixed in prior commits remain intact. The change swaps `res.end() + req.resume()` (drain-without-close, approved in sha=5c1c7f2c) for `res.end(callback → req.socket?.destroy())` (flush-then-destroy). The security posture is unchanged; this is a behavioral trade-off between two imperfect approaches to the same architectural constraint: Hono's request handler runs concurrently with the raw Node shim, so no layered approach fully prevents a second response attempt. One residual low-severity note: destroying the socket after flush causes Hono's write attempt to emit a write-after-destroy error rather than `ERR_HTTP_HEADERS_SENT` — observable difference is that the client receives the structured 413 before the reset (better UX) but keep-alive connections are terminated (acceptable for a localhost daemon). The comment slightly overstates the benefit ('prevents Hono's pipeline from racing') — it changes the failure mode, not the race itself.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 4df7b7e — updated comment to accurately describe the failure mode change: write-after-destroy (silent) instead of ERR_HTTP_HEADERS_SENT.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: comprehensive security audit hardening` (commit `4df7b7e6`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
Incremental diff is a comment-only change — the previous comment overstated that socket destruction 'prevents Hono's pipeline from racing'; the new comment more accurately describes the actual behavior (Hono's write fails silently on the closed socket instead of emitting ERR_HTTP_HEADERS_SENT). No functional code changed. All 14 audit findings fixed in prior commits remain intact. All prior-review blockers remain resolved.

### Confidence: 8.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

Issue #22, opened 2026-02-21
Migration 3 (unique-content-hash) fails on upgrade from v0.1.64: 'no such column: content_hash'


ddasgupta4:
## Bug

When upgrading from v0.1.64 to v0.1.65, the daemon fails to start due to a failed database migration.

## Error

```
ERROR [daemon] Fatal error
  Error: Migration 3 (unique-content-hash) failed: no such column: content_hash
```

## Root Cause

Migration 3 tries to create a `UNIQUE INDEX` on `memories.content_hash` before the column has been added to the table. The `addColumnIfMissing` call appears to either not run or not commit before the index creation is attempted.

Additionally, the migration references `is_deleted` which is also not present in the existing schema.

## Steps to Reproduce

1. Have a database created on v0.1.64 (or earlier)
2. Run `signet update install` to upgrade to v0.1.65
3. Run `signet daemon start` → daemon fails to start

## Workaround

Manually add the missing columns before starting the daemon:

```bash
sqlite3 ~/.agents/memory/memories.db   "ALTER TABLE memories ADD COLUMN content_hash TEXT; ALTER TABLE memories ADD COLUMN is_deleted INTEGER DEFAULT 0;"
```

Then clear stale embeddings to avoid UNIQUE constraint collisions on restart:

```bash
sqlite3 ~/.agents/memory/memories.db "DELETE FROM embeddings; UPDATE memories SET content_hash = NULL;"
```

## Additional Issue

The `signet daemon start` CLI command spawns the daemon using `process.argv[0]` (node), but the daemon binary uses `bun:` imports which are not supported by Node.js v25+. On macOS with a standard npm install, this means `signet daemon start` silently fails with `Failed to start daemon`.

Workaround is to run the daemon directly via bun:

```bash
bun /opt/homebrew/lib/node_modules/signetai/dist/daemon.js
```

Or register it as a launchd service pointing at the bun binary.

## Environment

- macOS (arm64)
- signet v0.1.65
- Node.js v25.6.1
- Bun 1.3.9
- Upgraded from v0.1.64

ddasgupta4:
Worth noting: this appears to be macOS-specific. The daemon works correctly on Linux (confirmed by at least one other user). The bun/node spawn issue in particular is likely tied to how the npm install resolves the executable path on macOS — on Linux, bun may be picked up correctly via PATH or the install method differs.

ddasgupta4:
[...]

Issue #21, opened 2026-02-19
chore(openclaw): define and enforce legacy-path deprecation gates


NicholaiVogel:
## Objective
Define objective criteria for deprecating the legacy command-hook runtime path.

## Scope
- Encode deprecation gates from OpenClaw metrics.
- Add release checklist checks for gate completion.
- Block legacy removal until all gates are met.

## Acceptance Criteria
- Gate checks require plugin adoption >= 90%.
- Duplicate incident rate remains zero during gate window.
- Callback and tool success rates >= 99%.
- Matches spec sections 19.6 and 26.6.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 19.6, 26.6)

NicholaiVogel:
Bulk closed - work completed.

Issue #20, opened 2026-02-19
docs(openclaw): write plugin-first migration and rollback runbook


NicholaiVogel:
## Objective
Provide operator guidance for migrating from legacy command-hook mode to plugin-first OpenClaw runtime integration.

## Scope
- Document prerequisites, enablement, validation, rollback, and troubleshooting.
- Include arbitration policy and mixed-mode safety notes.
- Include deprecation timeline criteria.

## Acceptance Criteria
- Runbook enables safe migration in canary and GA.
- Rollback steps are tested and documented.
- Documentation aligns with spec sections 26.6 and 25.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 26.6, 25)

NicholaiVogel:
Bulk closed - work completed.

Issue #19, opened 2026-02-19
chore(openclaw): plugin-first GA hardening and release checklist


NicholaiVogel:
## Objective
Complete final hardening and release readiness checks for OpenClaw plugin-first migration.

## Scope
- Verify parity, safety, and observability gates.
- Attach benchmark artifacts and canary outcomes.
- Confirm no-duplication and rollback readiness.

## Acceptance Criteria
- Section 24 acceptance checklist item for OpenClaw is satisfied.
- Required benchmark and canary evidence is attached.
- Release sign-off includes rollback and escalation readiness.

## References
- `docs/wip/memory-pipeline-plan.md` (sections 24, 20.8, 22.14)

NicholaiVogel:
Bulk closed - work completed.

PR #272, opened 2026-03-22
fix: harden error handling and resource cleanup

aaf2tbz:
## Summary

Addresses 7 robustness issues found in the core library, daemon, and CLI — focusing on unhandled exceptions that crash retrieval/import, resource leaks in SSE streams, and unbounded git subprocess chains.

### Core library (`@signet/core`)

- **#16 — Safe JSON.parse in row mapping**: Added `safeJsonParse()` helper for `tags` and `vectorClock` columns. A single corrupted row no longer crashes all memory retrieval — returns safe defaults (`[]` / `{}`) instead.
- **#17 — Safe JSON.parse in imports**: Each JSONL line parse is now wrapped in try-catch. Malformed lines are counted and skipped instead of crashing the entire import.
- **#18 — Transaction-wrapped imports**: All 3 import functions (`importMemories`, `importEntities`, `importRelations`) now run inside `BEGIN`/`COMMIT` with `ROLLBACK` on failure. No more partial data on crash.
- **#19 — Type safety note**: Full runtime validation (zod) deferred — `safeJsonParse` covers the crash-producing cases. The `as` casts remain but are now guarded at the JSON boundary.

### Daemon (`@signet/daemon`)

- **#20 — SSE stream resource leaks**: All 3 SSE endpoints (log stream, cross-agent events, task stream) now use a `dead` flag + centralized `cleanup()` function. `controller.enqueue()` is wrapped in try-catch — network errors trigger immediate listener removal instead of accumulating orphaned handlers. Keepalive intervals also guard against dead streams.
- **#21 — Child process cleanup**: `stdout`/`stderr` data handlers wrapped in try-catch. If the SSE write throws, the kill timer is cleared and child is terminated — prevents orphaned timers and zombie processes.
- **#22 — Git auto-commit timeout**: Added 30s timeout via `Promise.race()` and an `autocommitInFlight` concurrency guard. Stalled git subprocesses no longer block the debounce timer, and concurrent auto-commits can't pile up zombies.

### CLI

- Updated import output to handle new return types (`{ imported, errors }`) and report malformed line counts to the user.

## Files changed (4)

| File | Changes |
|------|---------|
| `packages/core/src/database.ts` | `safeJsonParse()` for row mapping |
| `packages/core/src/export.ts` | try-catch on JSONL parse, transaction wrapping, `exec()` on ImportDb type |
| `packages/daemon/src/daemon.ts` | SSE cleanup pattern, child process try-catch, git timeout + concurrency guard |
| `packages/cli/src/cli.ts` | Updated import callers for new return types |

## Test plan

- [x] `bun run build` — full build succeeds
- [x] `bun test` — 34 daemon tests pass
- [ ] Import a JSONL file with a deliberately malformed line — verify it's skipped, count reported
- [ ] Kill a daemon SSE client mid-stream — verify no listener leak in daemon logs
- [ ] Stall git (e.g. lock file) during auto-commit — verify 30s timeout fires

### DIFF
--- a/packages/core/src/export.ts
+++ b/packages/core/src/export.ts
@@ -72,6 +72,7 @@ interface ImportDb {
 		run(...args: unknown[]): void;
 		get(...args: unknown[]): Record<string, unknown> | undefined;
 	};
+	exec(sql: string): void;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: harden error handling and resource cleanup` (commit `32f25f31`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -8841,58 +8882,75 @@ async function gitUntrackProtectedFiles(dir: string): Promise<void> {
...
+			}, GIT_AUTOCOMMIT_TIMEOUT_MS);
+		});
+
+		await Promise.race([work, timeout]);
+	} finally {
+		autocommitInFlight = false;
+	}
 }
 
 let pendingChanges: string[] = [];
```
NicholaiVogel, line 8970: **Bug: timeout does not kill the stalled git subprocesses.**

When `Promise.race` resolves via the `timeout` branch, `work` keeps running in the background with the spawned `git add` / `git status` / `git commit` processes still alive. Any one of these can hold `.git/index.lock`, so every subsequent call to `gitAutoCommit` (once `autocommitInFlight` is reset in `finally`) will fail at `git add` with a lock-file error — defeating the purpose of the timeout.

Fix: track the child-process references inside `work` and kill them when the timeout wins:

```ts
let activeChild: ReturnType<typeof spawn> | null = null;

const work = new Promise<void>((resolve) => {
  const add = spawn("git", ["add", "-A"], ...);
  activeChild = add;
  add.on("close", (addCode) => {
    // ... existing chain ...
    // set activeChild to each subsequent process as it starts
  });
  add.on("error", () => resolve());
});

const timeout = new Promise<void>((resolve) => {
  setTimeout(() => {
    logger.warn("git", "Auto-commit timed out after 30s");
    try { activeChild?.kill("SIGTERM"); } catch {}
    resolve();
  }, GIT_AUTOCOMMIT_TIMEOUT_MS);
});
```

Alternatively, use `execFile` / `promisify` with an `AbortController` so the signal propagates cleanly through the whole chain.

packages/daemon/src/daemon.ts
```diff
@@ -1575,20 +1575,29 @@ app.get("/api/logs/stream", (c) => {
 
 	const stream = new ReadableStream({
 		start(controller) {
+			const cleanup = () => {
+				logger.off("log", onLog);
+			};
+
 			const onLog = (entry: LogEntry) => {
```
NicholaiVogel, line 1586: **Inconsistency: log-stream cleanup is weaker than the other two SSE endpoints.**

The log stream's `cleanup()` only removes the event listener; it does not call `controller.close()` and has no `dead` flag to prevent double-invocation. The cross-agent and task stream endpoints both guard with `dead` and close the controller. Without `controller.close()` the underlying `ReadableStream` is never properly terminated on a write error, which may leak the stream object depending on the runtime.

Suggested fix — align with the pattern used in the other two endpoints:

```ts
let dead = false;
const cleanup = () => {
  if (dead) return;
  dead = true;
  logger.off("log", onLog);
  try { controller.close(); } catch {}
};
```

packages/core/src/export.ts
```diff
@@ -265,123 +266,166 @@ export function importMemories(
...
-					id,
-				);
-				imported++;
-				continue;
+			const id = mem.id as string;
+
+			// Check for existing
+			const existing = db.prepare("SELECT id FROM memories WHERE id = ?").get(id);
+
+			if (existing) {
```
NicholaiVogel, line 292: **Minor: `errors` field absent from CLI fallback, relying on runtime `?? 0`.**

The no-file-present fallback in the CLI is `{ imported: 0, skipped: 0 }` (no `errors` key), while `importMemories` now returns `{ imported, skipped, errors }`. TypeScript should flag this mismatch if the return type is used directly. The `(memResult.errors ?? 0)` in the total-error calculation handles it at runtime, but the type annotation on the fallback object should be updated to `{ imported: 0, skipped: 0, errors: 0 }` for correctness and to prevent silent future regressions.

NicholaiVogel (changes requested):
One correctness bug in the git timeout implementation: orphaned child processes are not killed when the 30s timeout fires, which can leave `.git/index.lock` held and cause all subsequent auto-commits to fail silently. The log-stream SSE cleanup is also inconsistent with the other two SSE endpoints (missing `controller.close()` and `dead` guard). Everything else — safeJsonParse, transaction wrapping, error counting, cross-agent/task stream cleanup — is correct.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 7,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: harden error handling and resource cleanup` (commit `18c8a28c`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 18c8a28 — tracking `active` child ref, timeout now calls `active?.kill('SIGTERM')` to release .git/index.lock. Each stage in the chain updates the ref as it spawns.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 18c8a28 — log stream now has `dead` flag + `controller.close()` in cleanup, matching the cross-agent and task stream patterns.

packages/core/src/export.ts
aaf2tbz: Fixed in 18c8a28 — fallback now `{ imported: 0, skipped: 0, errors: 0 }`, removed the `?? 0` workaround.

packages/daemon/src/daemon.ts
```diff
@@ -8841,58 +8886,83 @@ async function gitUntrackProtectedFiles(dir: string): Promise<void> {
...
-				commit.on("error", () => resolve());
+				status.on("error", () => resolve());
 			});
-			status.on("error", () => resolve());
+			add.on("error", () => resolve());
 		});
-		add.on("error", () => resolve());
-	});
+
+		const timeout = new Promise<void>((resolve) => {
```
NicholaiVogel, line 8955: **Previously flagged — fixed.** `active?.kill('SIGTERM')` is now called when the 30s timeout fires. Good.

**New minor issue:** the `setTimeout` handle is never cleared when `work` resolves first. `Promise.race` resolves but the timeout callback still fires ~30s later and sends SIGTERM to the already-exited `commit` process. This is functionally harmless (kill to a dead pid is a no-op), but every successful auto-commit leaks a live timer for 30 seconds.

Fix: capture the timer handle and clear it on success:
```ts
const timeout = new Promise<void>((resolve) => {
  const t = setTimeout(() => {
    logger.warn('git', 'Auto-commit timed out after 30s');
    try { active?.kill('SIGTERM'); } catch {}
    resolve();
  }, GIT_AUTOCOMMIT_TIMEOUT_MS);
  // expose handle so the race winner can cancel it
  (timeout as any)._timer = t;
});
```
Or simpler — use a shared `let timer` in the outer scope and call `clearTimeout(timer)` after `await Promise.race([work, timeout])`.

packages/daemon/src/daemon.ts
```diff
@@ -1575,20 +1575,33 @@ app.get("/api/logs/stream", (c) => {
```
NicholaiVogel, line 1575: **Previously flagged — fixed.** `dead` flag, idempotent `cleanup()`, and `controller.close()` are now present and consistent with the cross-agent and task-stream endpoints. No further issues here.

NicholaiVogel (commented):
Both issues flagged in the prior review are now resolved: the log-stream SSE endpoint has the `dead` flag, idempotent `cleanup()`, and `controller.close()` consistent with the other two SSE endpoints; and the git timeout now calls `active?.kill("SIGTERM")` so stalled subprocesses are terminated when the 30s timer fires. One new minor issue introduced: the `setTimeout` inside the `timeout` Promise is never cleared when `work` resolves first, so the timer lingers for 30 seconds after every successful auto-commit and sends a harmless SIGTERM to an already-exited process. Not a correctness bug, but a resource hygiene issue worth fixing.

### Confidence: 8.1/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 0030373 — timer handle captured in outer scope, `clearTimeout(timer)` after `Promise.race` resolves. No more lingering 30s timer on success.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: harden error handling and resource cleanup` (commit `00303730`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
The one remaining issue from the prior review is now resolved: `timer` is captured outside the timeout Promise and `clearTimeout(timer)` is called after `Promise.race` completes, so the 30s timer no longer lingers after a successful auto-commit. Both cases are handled correctly — if `work` wins, the timer is cancelled; if `timeout` fires first, `clearTimeout` on an already-fired timer is a safe no-op. All other issues from prior reviews (log-stream SSE dead flag + controller.close(), active?.kill in timeout handler, safeJsonParse, transaction wrapping, error counting) remain correctly implemented. No new issues introduced.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

PR #275, opened 2026-03-22
fix: address 9 security and stability issues

aaf2tbz:
## Summary

- **Subprocess timeout** (CRITICAL): Bun's `spawn()` silently ignores the `timeout` option — replaced with manual `setTimeout` + `proc.kill()` with proper cleanup via `clearTimeout` on all exit paths
- **HTML injection** (HIGH): Wikilink renderer interpolated user-controlled markdown content (`rawSlug`, `text`) directly into raw HTML — added `escapeHtml()` for all user-controlled values (`<`, `>`, `&`, `"`)
- **Non-deterministic sitemap** (HIGH): Every build stamped ALL sitemap entries with `new Date()`, causing fake freshness signals to crawlers and non-reproducible builds — removed the `serialize` override entirely
- **Hardcoded Ollama endpoint** (HIGH): `http://localhost:11434` was hardcoded in 3 locations across `embedding-fetch.ts` and `daemon.ts` — now resolves via `OLLAMA_HOST` env var with fallback to `DEFAULT_OLLAMA_BASE_URL`
- **Secret storage validation** (HIGH): `loadOrCreateSecret()` blindly loaded any file as the HMAC signing key regardless of length — added length validation that regenerates corrupted/truncated secrets
- **CORS wildcard** (MEDIUM): OPTIONS preflight handler used `'*'` while POST handler correctly reflected request origin — aligned both to use request origin
- **FTS5 operator injection** (MEDIUM): `sanitizeFtsQuery()` preserved `OR`, `AND`, `NOT` from raw user input, allowing semantic manipulation of recall results — now treats all tokens as literals
- **Connection pool** (MEDIUM): Read pool created unlimited new connections when exhausted — added `MAX_READ_CONNECTIONS = 16` cap with fail-fast error
- **Type assertion** (MEDIUM): `normalizeLegacyEmbeddingsPayload()` used `as Record<string, unknown>` cast — replaced with `Object.create(null)` + `Object.assign` for safe property copy

### Declined fix

- **YAML frontmatter parser** (#5 in review): The claimed bug (`indexOf(":")` truncating values at colons) is incorrect — `slice(colonIdx + 1)` correctly takes everything after the first colon. `title: "My Project: Overview"` parses correctly as `"My Project: Overview"`, not `"My Project"`.

## Test plan

- [x] `bun run build` — all packages build successfully
- [x] `bun run typecheck` — no new type errors (pre-existing opencode-plugin error unchanged)
- [x] `bun test` — 834 pass, 98 fail (all failures pre-existing, none in changed files)
- [x] `bun test packages/daemon/src/auth/` — 69 pass, 0 fail
- [x] `bun test packages/daemon/src/memory-config.test.ts` — 42 pass, 0 fail
- [x] `web/` build — passes compilation (pre-existing content schema error in `signet-os-spec.md` unrelated)

### DIFF
--- a/packages/daemon/src/auth/tokens.ts
+++ b/packages/daemon/src/auth/tokens.ts
@@ -33,7 +33,9 @@ export function loadOrCreateSecret(secretPath: string): Buffer {
 	if (existsSync(secretPath)) {
 		const secret = readFileSync(secretPath);
 		if (secret.length !== SECRET_LENGTH) {
-			// Corrupted or truncated — regenerate
+			console.warn(
+				`[auth] Secret at ${secretPath} has unexpected length ${secret.length} — regenerating. All existing tokens will be invalidated.`,
+			);
 			const fresh = generateSecret();
 			writeFileSync(secretPath, fresh, { mode: 0o600 });
 			return fresh;
--- a/packages/daemon/src/auth/tokens.ts
+++ b/packages/daemon/src/auth/tokens.ts
@@ -27,9 +27,18 @@ export function generateSecret(): Buffer {
 	return randomBytes(32);
 }
 
+const SECRET_LENGTH = 32;
+
 export function loadOrCreateSecret(secretPath: string): Buffer {
 	if (existsSync(secretPath)) {
-		return readFileSync(secretPath);
+		const secret = readFileSync(secretPath);
+		if (secret.length !== SECRET_LENGTH) {
+			// Corrupted or truncated — regenerate
+			const fresh = generateSecret();
+			writeFileSync(secretPath, fresh, { mode: 0o600 });
+			return fresh;
+		}
+		return secret;
 	}
 	const dir = dirname(secretPath);
 	if (!existsSync(dir)) {
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -105,7 +105,7 @@ import { closeSynthesisProvider, initSynthesisProvider } from "./synthesis-llm";
 import { closeWidgetProvider, initWidgetProvider } from "./widget-llm";
 import { type LogEntry, logger } from "./logger";
 import { migrateConfig } from "./config-migration";
-import { type EmbeddingConfig, loadMemoryConfig } from "./memory-config";
+import { type EmbeddingConfig, loadMemoryConfig, DEFAULT_OLLAMA_BASE_URL } from "./memory-config";
 import {
 	getAttributesForAspectFiltered,
 	getKnowledgeGraphForConstellation,
@@ -899,7 +899,8 @@ function normalizeLegacyEmbeddingsPayload(
 		return defaultLegacyEmbeddingsResponse(limit, offset, "Legacy export returned invalid payload");
 	}
 
-	const data = payload as Record<string, unknown>;
+	const data: Record<string, unknown> = Object.create(null);
+	Object.assign(data, payload);
 	const rawEmbeddings = Array.isArray(data.embeddings) ? data.embeddings : [];
 	const embeddings = rawEmbeddings
 		.map((entry) => normalizeLegacyEmbeddingRow(entry, withVectors))
@@ -952,12 +953,24 @@ async function runLegacyEmbeddingsExport(
 	}
 
 	return await new Promise<LegacyEmbeddingsResponse>((resolve) => {
+		const timeout = withVectors ? 120000 : 30000;
 		const proc = spawn("python3", args, {
 			cwd: AGENTS_DIR,
 			stdio: "pipe",
-			timeout: withVectors ? 120000 : 30000,
 			windowsHide: true,
 		});
+
+		// Bun's spawn() silently ignores `timeout` — enforce manually
+		const timer = setTimeout(() => {
+			proc.kill();
+			resolve(
+				defaultLegacyEmbeddingsResponse(
+					limit,
+					offset,
+					`Legacy embeddings export timed out after ${timeout}ms`,
+				),
+			);
+		}, timeout);
 
 		let stdout = "";
 		let stderr = "";
@@ -971,6 +984,7 @@ async function runLegacyEmbeddingsExport(
 		});
 
 		proc.on("close", (code) => {
+			clearTimeout(timer);
 			if (code !== 0) {
 				resolve(
 					defaultLegacyEmbeddingsResponse(
@@ -1002,6 +1016,7 @@ async function runLegacyEmbeddingsExport(
 		});
 
 		proc.on("error", (error) => {
+			clearTimeout(timer);
 			resolve(defaultLegacyEmbeddingsResponse(limit, offset, error.message));
 		});
 	});
@@ -1049,7 +1064,8 @@ async function checkEmbeddingProvider(cfg: EmbeddingConfig): Promise<EmbeddingSt
 				// Native unavailable — check if ollama is available as fallback
 				logger.warn("embedding", `Native provider unavailable: ${nativeStatus.error ?? "unknown"}`);
 				try {
-					const ollamaRes = await fetch("http://127.0.0.1:11434/api/tags", {
+					const ollamaUrl = process.env.OLLAMA_HOST || DEFAULT_OLLAMA_BASE_URL;
+					const ollamaRes = await fetch(`${ollamaUrl.replace(/\/$/, "")}/api/tags`, {
 						method: "GET",
 						signal: AbortSignal.timeout(3000),
 					});
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -94,6 +94,7 @@ import {
 	resolveEmbeddingBaseUrl,
 	resolveEmbeddingApiKey,
 	setNativeFallbackToOllama,
+	resolveOllamaUrl,
 } from "./embedding-fetch";
 import { detectDrift } from "./predictor-comparison";
 import { getPredictorState } from "./predictor-state";
@@ -105,7 +106,7 @@ import { closeSynthesisProvider, initSynthesisProvider } from "./synthesis-llm";
 import { closeWidgetProvider, initWidgetProvider } from "./widget-llm";
 import { type LogEntry, logger } from "./logger";
 import { migrateConfig } from "./config-migration";
-import { type EmbeddingConfig, loadMemoryConfig, DEFAULT_OLLAMA_BASE_URL } from "./memory-config";
+import { type EmbeddingConfig, loadMemoryConfig } from "./memory-config";
 import {
 	getAttributesForAspectFiltered,
 	getKnowledgeGraphForConstellation,
@@ -1064,8 +1065,7 @@ async function checkEmbeddingProvider(cfg: EmbeddingConfig): Promise<EmbeddingSt
 				// Native unavailable — check if ollama is available as fallback
 				logger.warn("embedding", `Native provider unavailable: ${nativeStatus.error ?? "unknown"}`);
 				try {
-					const ollamaUrl = process.env.OLLAMA_HOST || DEFAULT_OLLAMA_BASE_URL;
-					const ollamaRes = await fetch(`${ollamaUrl.replace(/\/$/, "")}/api/tags`, {
+					const ollamaRes = await fetch(`${resolveOllamaUrl().replace(/\/$/, "")}/api/tags`, {
 						method: "GET",
 						signal: AbortSignal.timeout(3000),
 					});
--- a/packages/daemon/src/db-accessor.ts
+++ b/packages/daemon/src/db-accessor.ts
@@ -398,6 +398,7 @@ function createAccessor(writeConn: Database): DbAccessor {
 			return pooled;
 		}
 		if (readInUse.size >= MAX_READ_CONNECTIONS) {
+			console.warn(`[db] Read connection limit exceeded (${readInUse.size}/${MAX_READ_CONNECTIONS})`);
 			throw new Error("Read connection limit exceeded");
 		}
 		const conn = new Database(dbPath, { readonly: true });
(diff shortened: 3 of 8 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: address 9 security and stability issues` (commit `6da8f325`) now and I'll follow up shortly with feedback.

web/functions/api/waitlist.ts
```diff
@@ -51,11 +51,12 @@ export const onRequestPost: PagesFunction<Env> = async (context) => {
 	}
 };
 
-export const onRequestOptions: PagesFunction = async () => {
+export const onRequestOptions: PagesFunction = async (context) => {
```
NicholaiVogel, line 64: The OPTIONS handler now reflects the request Origin without validating it against an allowlist. This aligns with the POST handler's existing behavior, but both handlers accept requests from *any* origin — semantically equivalent to `'*'` for non-credentialed requests. The PR description frames this as a fix, but the real risk remains: if `Access-Control-Allow-Credentials: true` is ever added, reflected-origin CORS becomes a security hole (CORS allowlist bypass). Consider validating the reflected origin against an explicit allowlist of known domains (e.g., `['https://signetai.sh', 'https://www.signetai.sh']`) in both handlers, returning `403` or omitting the header for unknown origins.

packages/daemon/src/auth/tokens.ts
```diff
@@ -27,9 +27,18 @@ export function generateSecret(): Buffer {
...
 
+const SECRET_LENGTH = 32;
+
 export function loadOrCreateSecret(secretPath: string): Buffer {
 	if (existsSync(secretPath)) {
-		return readFileSync(secretPath);
+		const secret = readFileSync(secretPath);
+		if (secret.length !== SECRET_LENGTH) {
+			// Corrupted or truncated — regenerate
+			const fresh = generateSecret();
```
NicholaiVogel, line 39: When a corrupted/truncated secret is detected and regenerated, there is no log warning before overwriting the file. Every existing issued token is silently invalidated — clients will see opaque 401s with no daemon-side indication of why. Add at least a `console.warn` or a `logger.warn` call before `writeFileSync` so operators can diagnose unexpected logouts. Example: `console.warn('[auth] Secret at', secretPath, 'has unexpected length', secret.length, '— regenerating. All existing tokens will be invalidated.')`

packages/daemon/src/db-accessor.ts
```diff
@@ -396,6 +397,9 @@ function createAccessor(writeConn: Database): DbAccessor {
 			readInUse.add(pooled);
 			return pooled;
 		}
```
NicholaiVogel, line 399: The `MAX_READ_CONNECTIONS = 16` cap throws immediately when exhausted rather than queuing waiters. Under a traffic burst (e.g., a recall-heavy agent session), the 17th concurrent read will propagate an unhandled error to the caller. `withReadDb` does not catch this, so it surfaces as an uncaught exception at the route handler level. This is documented as intentional fail-fast behavior, but callers should be audited to ensure they handle `Error('Read connection limit exceeded')` gracefully rather than crashing the daemon. A short queue with a timeout would be a safer alternative, but if fail-fast is the chosen policy, add explicit handling at the HTTP boundary.

packages/daemon/src/daemon.ts
```diff
@@ -971,6 +984,7 @@ async function runLegacyEmbeddingsExport(
 		});
```
NicholaiVogel, line 985: The manual timeout fires `proc.kill()` and then `resolve(...)`. When `proc.kill()` completes, it triggers the `close` event, which calls `clearTimeout(timer)` (no-op — already fired) and then `resolve(...)` again (no-op — Promise already settled). This double-resolve is safe per the Promise spec. No action needed — just confirming correctness for reviewers.

packages/daemon/src/embedding-fetch.ts
```diff
@@ -1,8 +1,12 @@
 import type { EmbeddingConfig } from "./memory-config";
-import { DEFAULT_OPENAI_BASE_URL } from "./memory-config";
+import { DEFAULT_OPENAI_BASE_URL, DEFAULT_OLLAMA_BASE_URL } from "./memory-config";
 import { logger } from "./logger";
 import { getSecret } from "./secrets.js";
 
+function resolveOllamaUrl(): string {
```
NicholaiVogel: Minor: `resolveOllamaUrl()` reads `process.env.OLLAMA_HOST` at call time, not at module load time. This is correct for a daemon that may reload config, but if `OLLAMA_HOST` is malformed (e.g. missing scheme), the subsequent `fetch(${url}/api/embeddings)` will throw. The error is caught by the outer try/catch in `fetchEmbedding` and logged as a warning, so this is not a crash risk — but a startup validation of `OLLAMA_HOST` format would catch misconfiguration earlier.

NicholaiVogel (commented):
All 8 accepted fixes are mechanically correct and address real issues. Three things stand out: (1) the CORS 'fix' reflects any arbitrary Origin without allowlist validation — it aligns OPTIONS with POST, but neither actually restricts access, leaving the endpoint open to any origin; (2) silent secret regeneration in tokens.ts invalidates all existing tokens with no log warning; (3) the connection pool cap throws immediately with no queuing, which can cause cascading failures under burst load. None of these are regressions introduced by this PR, but items 1 and 2 are worth a comment before merge.

### Confidence: 7.9/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 8,
  "pattern_correctness": 9,
  "documentation_coverage": 6
}
```

web/functions/api/waitlist.ts
aaf2tbz: Addressed — added explicit allowlist: `['https://signetai.sh', 'https://www.signetai.sh']`. Both POST and OPTIONS handlers now validate origin against the allowlist and return empty string for unknown origins.

packages/daemon/src/auth/tokens.ts
aaf2tbz: Added `console.warn` before regeneration: logs the path, unexpected length, and that all existing tokens will be invalidated.

packages/daemon/src/db-accessor.ts
aaf2tbz: Added `console.warn` with current/max counts before the throw. Keeping fail-fast policy — the error will surface as 500 at HTTP boundary which is correct behavior under resource exhaustion.

packages/daemon/src/embedding-fetch.ts
aaf2tbz: Added URL format validation — `new URL(raw)` catches malformed values and falls back to default with a `logger.warn`. Misconfiguration is now surfaced early rather than failing silently on first embedding request.

aaf2tbz:
**Commit `56479c6e` addresses all 4 actionable review comments:**

1. **CORS allowlist** — Added explicit `ALLOWED_ORIGINS` (`signetai.sh`, `www.signetai.sh`) with validation in both POST and OPTIONS handlers. Unknown origins get empty ACAO header.
2. **Secret regen logging** — `console.warn` before overwrite with path, unexpected length, and token invalidation notice.
3. **Connection pool logging** — `console.warn` with current/max count before fail-fast throw.
4. **OLLAMA_HOST validation** — `new URL()` validation with `logger.warn` and fallback to default on malformed input.

Comment #4 (subprocess double-resolve) was informational — no changes needed.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: address 9 security and stability issues` (commit `56479c6e`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -1049,7 +1064,8 @@ async function checkEmbeddingProvider(cfg: EmbeddingConfig): Promise<EmbeddingSt
 				// Native unavailable — check if ollama is available as fallback
 				logger.warn("embedding", `Native provider unavailable: ${nativeStatus.error ?? "unknown"}`);
```
NicholaiVogel, line 1066: **Prior issue resolved (CORS, secret warning, pool warning) — minor follow-up here.**

`daemon.ts` reads `process.env.OLLAMA_HOST` directly:
```ts
const ollamaUrl = process.env.OLLAMA_HOST || DEFAULT_OLLAMA_BASE_URL;
const ollamaRes = await fetch(`${ollamaUrl.replace(/\/$/, "")}/api/tags`, ...);
```
`embedding-fetch.ts` introduced `resolveOllamaUrl()` which validates the value with `new URL()` and falls back gracefully. This path skips that validation, so a malformed `OLLAMA_HOST` (e.g. `localhost:11434` without a scheme) will throw a `TypeError` inside `fetch()` rather than falling back to the default. The surrounding `try/catch` will catch it, so it is not a crash, but the error path returns `{ status: 'unavailable' }` with no hint about the misconfiguration. Consider calling `resolveOllamaUrl()` (or inlining the same guard) here for consistency.

web/functions/api/waitlist.ts
```diff
@@ -2,10 +2,20 @@ interface Env {
 	RESEND_API_KEY: string;
 }
 
+const ALLOWED_ORIGINS = [
+	'https://signetai.sh',
+	'https://www.signetai.sh',
+];
+
+function isAllowedOrigin(origin: string): boolean {
+	if (!origin) return false;
```
NicholaiVogel: **Prior CORS issue resolved.** Setting `Access-Control-Allow-Origin: ''` (empty string) for rejected origins is functionally correct — browsers will fail the CORS preflight — but an empty-string header value is non-standard and some older reverse proxies or Cloudflare workers runtimes may strip or mishandle it. A marginally more defensive alternative is to omit the header entirely when the origin is not allowed, or set it to the literal site origin as a safe default. Not blocking, just noting for future hardening.

NicholaiVogel (dismissed):
All three issues flagged in the prior review have been resolved: (1) CORS now validates against an explicit allowlist for both POST and OPTIONS handlers — the wildcard is gone; (2) secret regeneration now emits a console.warn with the path, length, and token-invalidation notice; (3) the connection pool limit now logs before throwing. One minor inconsistency remains: daemon.ts reads OLLAMA_HOST directly without the URL-validation guard that embedding-fetch.ts's resolveOllamaUrl() provides, so an invalid OLLAMA_HOST value will throw an unhandled fetch TypeError in the checkEmbeddingProvider path instead of gracefully falling back.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 7,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 6
}
```

packages/daemon/src/daemon.ts
aaf2tbz: Fixed — daemon.ts now imports and calls `resolveOllamaUrl()` from embedding-fetch.ts instead of reading `process.env.OLLAMA_HOST` directly. Also removed the now-unused `DEFAULT_OLLAMA_BASE_URL` import from daemon.ts.

web/functions/api/waitlist.ts
aaf2tbz: Addressed — refactored to a shared `corsHeaders()` helper that returns an empty object for rejected origins, so the ACAO header is omitted entirely rather than set to empty string.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: address 9 security and stability issues` (commit `71fbdfc2`); taking another look at the fixes and I will report back shortly.

web/functions/api/waitlist.ts
```diff
@@ -2,10 +2,20 @@ interface Env {
 	RESEND_API_KEY: string;
 }
 
+const ALLOWED_ORIGINS = [
+	'https://signetai.sh',
+	'https://www.signetai.sh',
+];
+
+function corsHeaders(origin: string): Record<string, string> {
```
NicholaiVogel, line 10: Fixed (improvement noted): the corsHeaders() refactor is slightly better than the previous isAllowedOrigin() approach — disallowed origins now produce no Access-Control-Allow-Origin header at all rather than setting it to an empty string, which is the correct RFC behaviour.

packages/daemon/src/daemon.ts
```diff
@@ -1049,7 +1065,7 @@ async function checkEmbeddingProvider(cfg: EmbeddingConfig): Promise<EmbeddingSt
...
 				try {
-					const ollamaRes = await fetch("http://127.0.0.1:11434/api/tags", {
+					const ollamaRes = await fetch(`${resolveOllamaUrl().replace(/\/$/, "")}/api/tags`, {
```
NicholaiVogel, line 1068: Previously flagged issue resolved: resolveOllamaUrl() is now used here instead of the raw process.env.OLLAMA_HOST read, so an invalid OLLAMA_HOST value will log a warning and fall back to the default rather than throwing an unhandled fetch TypeError.

NicholaiVogel (approved):
All issues from prior reviews are now resolved. The last remaining item — daemon.ts reading OLLAMA_HOST directly without URL-validation — is fixed by importing and calling resolveOllamaUrl() instead. The corsHeaders() refactor is a minor improvement over isAllowedOrigin(): disallowed origins now produce no Access-Control-Allow-Origin header (correct) instead of an empty-string header (technically incorrect). No new issues introduced.

### Confidence: 8.3/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

PR #276, opened 2026-03-22
fix: security hardening — auth timing, SSRF, YAML injection, scope enforcement

aaf2tbz:
## Summary

- **9 verified security fixes** from a full-source review (10 findings, 1 false positive dismissed)
- Covers auth, rate limiting, SSRF prevention, YAML injection, scope enforcement, command injection, and crash resilience
- 75 tests passing (5 new security-specific tests added, 2 behavioral tests updated)

### Critical
- **Rate limiter bypass** — `x-signet-actor` header no longer used for rate limit keys on unauthenticated requests. All anonymous requests share a single bucket.

### High
- **Timing-safe HMAC** — `Buffer.equals()` → `crypto.timingSafeEqual` for token signature verification
- **Host header spoofing** — `isLocalhost()` fails closed when TCP socket info unavailable; never falls back to the spoofable Host header
- **Command injection** — `restart_command` from `agent.yaml` validated against a pattern allowlist before shell execution
- **SSRF protection** — URL fetcher now resolves DNS and blocks private/reserved IPs, handles redirects manually with per-hop validation
- **YAML injection** — `formatYaml()` escapes `"`, `\`, `\n`, `\t` inside quoted strings

### Medium
- **Scope enforcement** — Empty-scope bypass restricted to admin-role tokens only (least privilege)
- **Claims validation** — Token payload validated for `sub` (string), `role` (valid enum), `scope` (object, not string/array/null)
- **Crash resilience** — `JSON.parse` on search result tags wrapped in try-catch with empty array fallback

## Files changed (8)

| File | What |
|------|------|
| `packages/daemon/src/auth/tokens.ts` | timingSafeEqual + claims validation |
| `packages/daemon/src/auth/middleware.ts` | rate limiter key + isLocalhost fail-closed |
| `packages/daemon/src/auth/policy.ts` | empty-scope admin-only |
| `packages/daemon/src/auth/auth.test.ts` | 5 new tests + 2 updated |
| `packages/cli/src/features/daemon.ts` | restart_command allowlist |
| `packages/daemon/src/pipeline/url-fetcher.ts` | SSRF protection |
| `packages/core/src/yaml.ts` | string escaping |
| `packages/core/src/search.ts` | safe JSON.parse |

## False positive dismissed

**Finding #2 (TOCTOU race in rate limiter)** — The review claimed `await next()` sits between `check()` and `record()`, but the actual code calls both synchronously before `await next()` (middleware.ts lines 152-163). In Bun's single-threaded event loop, synchronous code between two await points is atomic. No race condition exists.

## Test plan

- [x] `bun test packages/daemon/src/auth/auth.test.ts` — 75 pass, 0 fail
- [x] `bun run typecheck` — no new type errors in changed files
- [x] `bun run build` — builds cleanly
- [ ] Manual: start daemon in hybrid mode, verify localhost connections still work via TCP
- [ ] Manual: verify `signet recall` works with tags containing special characters

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/cli/src/features/daemon.ts
+++ b/packages/cli/src/features/daemon.ts
@@ -409,7 +409,7 @@ async function restartOpenClaw(basePath: string): Promise<boolean> {
 	// injection via tampered agent.yaml (git sync, social engineering).
 	const SAFE_PATTERNS = [
 		/^systemctl\s+--user\s+restart\s+[\w.-]+$/,
-		/^launchctl\s+kickstart\s+-k\s+[\w./-]+$/,
+		/^launchctl\s+kickstart\s+-k\s+[\w.-]+(\/[\w.-]+)*$/,
 		/^brew\s+services\s+restart\s+[\w.-]+$/,
 		/^supervisorctl\s+restart\s+[\w.-]+$/,
 	];
--- a/packages/cli/src/features/daemon.ts
+++ b/packages/cli/src/features/daemon.ts
@@ -122,7 +122,9 @@ export async function migrateSchema(options: PathOptions, deps: Deps): Promise<v
 		printMigrationErrors(result.errors);
 
 		if (result.migrated) {
-			console.log(chalk.green(`  ✓ Migrated ${result.memoriesMigrated} memories from ${result.fromSchema} to ${result.toSchema}`));
+			console.log(
+				chalk.green(`  ✓ Migrated ${result.memoriesMigrated} memories from ${result.fromSchema} to ${result.toSchema}`),
+			);
 		} else {
 			console.log(chalk.dim("  No migration needed"));
 		}
@@ -400,6 +402,22 @@ async function restartOpenClaw(basePath: string): Promise<boolean> {
 		console.log(chalk.dim("    services:"));
 		console.log(chalk.dim("      openclaw:"));
 		console.log(chalk.dim('        restart_command: "systemctl --user restart openclaw"'));
+		return false;
+	}
+
+	// Validate command against known safe restart patterns to prevent
+	// injection via tampered agent.yaml (git sync, social engineering).
+	const SAFE_PATTERNS = [
+		/^systemctl\s+--user\s+restart\s+[\w.-]+$/,
+		/^launchctl\s+kickstart\s+-k\s+[\w./-]+$/,
+		/^brew\s+services\s+restart\s+[\w.-]+$/,
+		/^supervisorctl\s+restart\s+[\w.-]+$/,
+	];
+	if (!SAFE_PATTERNS.some((p) => p.test(cmd))) {
+		console.log();
+		console.log(chalk.red("  Restart command rejected — does not match safe patterns."));
+		console.log(chalk.dim(`  Command: ${cmd}`));
+		console.log(chalk.dim("  Allowed: systemctl --user restart <name>, launchctl kickstart -k <path>"));
 		return false;
 	}
 
--- a/packages/daemon/src/auth/auth.test.ts
+++ b/packages/daemon/src/auth/auth.test.ts
@@ -87,7 +87,7 @@ describe("tokens", () => {
 		expect(result.claims?.scope).toEqual(scope);
 	});
 
-	test("empty scope object means full access (no scope fields set)", () => {
+	test("verifyToken accepts token with empty scope object", () => {
 		const token = createToken(secret, { sub: "x", scope: {}, role: "agent" }, 60);
 		const result = verifyToken(secret, token);
 		expect(result.authenticated).toBe(true);
@@ -269,11 +269,12 @@ describe("policy - checkScope", () => {
 		expect(d.reason).toMatch(/proj-a/);
 	});
 
-	test("empty scope on non-admin token is denied", () => {
+	test("empty scope on non-admin token is allowed with deprecation warning", () => {
 		const claims = makeClaims("agent", {});
 		const d = checkScope(claims, { project: "anything" }, "team");
-		expect(d.allowed).toBe(false);
-		expect(d.reason).toMatch(/non-admin/);
+		// Currently allowed for backwards compatibility — will be denied
+		// in a future release after operators rotate tokens.
+		expect(d.allowed).toBe(true);
 	});
 
 	test("empty scope on admin token grants full access", () => {
--- a/packages/daemon/src/auth/auth.test.ts
+++ b/packages/daemon/src/auth/auth.test.ts
@@ -7,637 +7,792 @@ import { Hono } from "hono";
 import { generateSecret, loadOrCreateSecret, createToken, verifyToken } from "./tokens";
 import { checkPermission, checkScope } from "./policy";
 import { AuthRateLimiter } from "./rate-limiter";
-import {
-  createAuthMiddleware,
-  requirePermission,
-  requireRateLimit,
-} from "./middleware";
+import { createAuthMiddleware, requirePermission, requireRateLimit } from "./middleware";
 import { parseAuthConfig } from "./config";
 import type { TokenClaims, TokenRole } from "./types";
-
 
 // =============================================================================
 // Tokens
 // =============================================================================
 
 describe("tokens", () => {
-  const secret = generateSecret();
+	const secret = generateSecret();
 
-  test("generateSecret returns a 32-byte buffer", () => {
-    const s = generateSecret();
-    expect(s).toBeInstanceOf(Buffer);
-    expect(s.length).toBe(32);
-  });
+	test("generateSecret returns a 32-byte buffer", () => {
+		const s = generateSecret();
+		expect(s).toBeInstanceOf(Buffer);
+		expect(s.length).toBe(32);
+	});
 
-  test("two generateSecret calls produce different values", () => {
-    const a = generateSecret();
-    const b = generateSecret();
-    expect(a.equals(b)).toBe(false);
-  });
+	test("two generateSecret calls produce different values", () => {
+		const a = generateSecret();
+		const b = generateSecret();
+		expect(a.equals(b)).toBe(false);
+	});
 
-  test("createToken produces a string with exactly one dot separator", () => {
-    const token = createToken(secret, { sub: "test", scope: {}, role: "agent" }, 60);
-    const dots = token.split(".").length - 1;
-    expect(dots).toBe(1);
-  });
+	test("createToken produces a string with exactly one dot separator", () => {
+		const token = createToken(secret, { sub: "test", scope: {}, role: "agent" }, 60);
+		const dots = token.split(".").length - 1;
+		expect(dots).toBe(1);
+	});
 
-  test("verifyToken succeeds with correct secret", () => {
-    const token = createToken(secret, { sub: "u1", scope: {}, role: "operator" }, 60);
-    const result = verifyToken(secret, token);
-    expect(result.authenticated).toBe(true);
-    expect(result.claims).not.toBeNull();
-  });
+	test("verifyToken succeeds with correct secret", () => {
+		const token = createToken(secret, { sub: "u1", scope: {}, role: "operator" }, 60);
+		const result = verifyToken(secret, token);
+		expect(result.authenticated).toBe(true);
+		expect(result.claims).not.toBeNull();
+	});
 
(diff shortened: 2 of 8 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: security hardening — auth timing, SSRF, YAML injection, scope enforcement` (commit `5fffc7f8`) now and I'll follow up shortly with feedback.

packages/daemon/src/auth/policy.ts
```diff
@@ -88,10 +74,14 @@ export function checkScope(
 		return { allowed: true };
```
NicholaiVogel, line 81: **Breaking change — no migration path.** Every non-admin token that was issued with `scope: {}` (the previous 'full-access' sentinel) will now be denied by `checkScope`. The old semantics of 'empty scope = unrestricted' were widely documented in the codebase. In a live deployment, all such tokens will silently start returning 403 on the first scope-checked endpoint after this ships. Either (a) add a warning log when denying on empty scope and give operators a deprecation window, or (b) document a token rotation procedure and bump the token format version so the break is explicit. As written this will cause hard-to-diagnose production outages for any operator who issued tokens without explicit scope fields.

packages/daemon/src/pipeline/url-fetcher.ts
```diff
@@ -5,6 +5,54 @@
...
+	return PRIVATE_RANGES.some((r) => r.test(ip));
+}
+
+/** Resolve hostname and reject private/reserved IPs. */
+export async function validateUrl(url: string): Promise<void> {
+	const parsed = new URL(url);
+	if (parsed.protocol !== "https:" && parsed.protocol !== "http:") {
+		throw new Error(`Unsupported protocol: ${parsed.protocol}`);
+	}
+
```
NicholaiVogel, line 45: **DNS rebinding TOCTOU gap.** `validateUrl` resolves DNS, confirms a public IP, then returns. A few milliseconds later `fetch(target, ...)` opens a new TCP connection. If an attacker controls the DNS record with a very low TTL (or via a rebinding attack), the second resolution inside the JS engine/OS can return a private IP that was not seen during validation. This protection is substantially better than nothing, but it is not a complete SSRF block. At minimum add a comment: `// Note: DNS rebinding can bypass this check; defence-in-depth via network-level egress filtering is recommended.` If stricter guarantees are needed, resolve the hostname once, pin the resulting IP, and pass it as an explicit `Host:` header rather than re-resolving on each hop.

packages/daemon/src/auth/auth.test.ts
```diff
@@ -7,637 +7,792 @@ import { Hono } from "hono";
...
+	});
+
+	test("empty scope object means full access (no scope fields set)", () => {
+		const token = createToken(secret, { sub: "x", scope: {}, role: "agent" }, 60);
+		const result = verifyToken(secret, token);
+		expect(result.authenticated).toBe(true);
+		const s = result.claims?.scope ?? {};
+		expect(s.project).toBeUndefined();
+		expect(s.agent).toBeUndefined();
+		expect(s.user).toBeUndefined();
```
NicholaiVogel, line 97: **Stale test description creates a correctness trap.** The test `"empty scope object means full access (no scope fields set)"` now contradicts the system's behaviour: after this PR, a non-admin token with `scope: {}` is *denied* by `checkScope`. The test itself only exercises `verifyToken` (which still succeeds) so it does not catch the regression, but any future developer reading the description will assume empty-scope tokens have full access and write code or issue tokens based on that assumption. Rename it to something like `"verifyToken accepts token with empty scope object"` to reflect what is actually being tested.

packages/daemon/src/pipeline/url-fetcher.ts
```diff
@@ -5,6 +5,54 @@
...
+const PRIVATE_RANGES = [
+	/^127\./,
+	/^10\./,
+	/^172\.(1[6-9]|2\d|3[01])\./,
+	/^192\.168\./,
+	/^169\.254\./,
+	/^0\./,
+	/^::1$/,
+	/^fc00:/i,
+	/^fd/i,
```
NicholaiVogel: **`/^fd/i` ULA regex is too broad.** The pattern matches any string that starts with the two characters `fd`, regardless of what follows. A canonical ULA address is `fd` followed by exactly 40 hex nibbles (e.g. `fd12:3456:...`). A more precise guard is `/^fd[0-9a-f]{2}:/i`. The current pattern is a false-positive risk rather than a false-negative risk (fail-safe direction), but it will break on any legitimate hostname or address that happens to start with 'fd' in a future edge case. Low severity but easy to tighten.

packages/cli/src/features/daemon.ts
```diff
@@ -403,6 +405,22 @@ async function restartOpenClaw(basePath: string): Promise<boolean> {
...
 	}
 
+	// Validate command against known safe restart patterns to prevent
+	// injection via tampered agent.yaml (git sync, social engineering).
+	const SAFE_PATTERNS = [
+		/^systemctl\s+--user\s+restart\s+[\w.-]+$/,
+		/^launchctl\s+kickstart\s+-k\s+[\w./-]+$/,
+		/^brew\s+services\s+restart\s+[\w.-]+$/,
+		/^supervisorctl\s+restart\s+[\w.-]+$/,
+	];
```
NicholaiVogel, line 415: **`launchctl` path segment allows `../` traversal.** The pattern `[\w./-]+` admits `/` characters in the service path, which means a crafted `agent.yaml` could supply a path like `../../Library/LaunchAgents/com.malicious`. This does not enable arbitrary command injection (no shell metacharacters are admitted), but it does allow kickstarting an unexpected plist target. Tighten to `[\w.-]+(/[\w.-]+)*` to require each path component to be free of traversal sequences.

NicholaiVogel (changes requested):
Strong security hardening across 8 files. All 9 claimed fixes are correctly implemented. Three issues require attention before merge: (1) the empty-scope denial in policy.ts is a silent breaking change for all existing non-admin tokens issued with scope:{} — no migration path is provided; (2) the SSRF DNS-validation has an inherent TOCTOU/DNS-rebinding gap that is undocumented; (3) one existing test description now contradicts the new system behavior and will mislead future developers.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 6
}
```

aaf2tbz:
## Review feedback addressed (commit `62d40774`)

All 6 review comments resolved:

1. **Unused import** — removed `DEFAULT_EMBEDDING_DIMENSIONS` from search.ts
2. **Empty-scope breaking change** — changed from hard denial to deprecation warning. Logs `DEPRECATION: non-admin token has empty scope` with sub/role so operators can rotate tokens before enforcement in a future release. Currently allows access.
3. **DNS rebinding TOCTOU** — added caveat comment documenting the rebinding gap and recommending network-level egress filtering for production
4. **Stale test description** — renamed from "empty scope object means full access" to "verifyToken accepts token with empty scope object"
5. **ULA regex too broad** — tightened `/^fd/i` to `/^fd[0-9a-f]{2}:/i`
6. **launchctl path traversal** — changed `[\w./-]+` to `[\w.-]+(\/[\w.-]+)*` to prevent `../` traversal in path components

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: security hardening — auth timing, SSRF, YAML injection, scope enforcement` (commit `62d40774`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/auth/policy.ts
```diff
@@ -88,9 +75,19 @@ export function checkScope(
...
 	}
 
-	// Unscoped tokens (empty scope object) have full access
+	// DEPRECATION: Non-admin tokens with empty scope currently get full access
+	// but will be denied in a future release. Log a warning so operators can
+	// rotate tokens with explicit scopes before enforcement begins.
 	const scope = claims.scope;
 	if (!scope.project && !scope.agent && !scope.user) {
```
NicholaiVogel, line 88: The deprecation warning fires on every request that passes through `checkScope` with an empty-scope non-admin token. In a busy deployment this will flood structured logs with one warning per API call. Consider rate-limiting the log emission (e.g. log once per unique `sub` per N minutes, or sample at 1%) to keep logs actionable without drowning out real signals. Not a blocker, but worth addressing before the deprecation window closes and the check is hardened.

NicholaiVogel (dismissed):
All three issues from the prior review have been resolved: (1) empty-scope non-admin tokens now follow a deprecation path (warn + allow) instead of a silent breaking denial; (2) DNS rebinding is documented inline; (3) test names now match actual system behaviour. The incremental diff also contains two bonus correctness fixes — the IPv6 ULA regex is tightened from the overly-broad `/^fd/i` to the RFC-4193-correct `/^fd[0-9a-f]{2}:/i`, and the launchctl allowlist pattern is made more structurally correct. No new issues introduced. Ready to merge.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 8
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: security hardening — auth timing, SSRF, YAML injection, scope enforcement` (commit `06ac84dd`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/auth/policy.ts
```diff
@@ -2,14 +2,11 @@
...
-	AuthMode,
-	Permission,
-	PolicyDecision,
-	TokenClaims,
-	TokenRole,
-	TokenScope,
-} from "./types";
+import type { AuthMode, Permission, PolicyDecision, TokenClaims, TokenRole, TokenScope } from "./types";
+import { logger } from "../logger";
+
```
NicholaiVogel, line 7: Minor: `warnedEmptyScope` is a module-level Set that is never pruned. In a long-running daemon processing many distinct `sub` values (e.g. ephemeral agent tokens with UUID subs), the Set will grow unbounded for the process lifetime. For the typical use case this is negligible, but if tokens are issued with unique per-session subs at high volume, consider a bounded LRU or TTL-based eviction. Not a blocker — just worth a comment in the code or a follow-up ticket.

NicholaiVogel (dismissed):
Incremental diff adds a module-level `warnedEmptyScope` Set to deduplicate per-sub deprecation warnings in `policy.ts`, preventing log flooding on busy deployments. All three issues from the first review cycle were resolved in the prior commit (62d407). This final change is a clean, correct improvement with one minor note worth logging but not blocking.

### Confidence: 8.3/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: security hardening — auth timing, SSRF, YAML injection, scope enforcement` (commit `891d32f8`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
Incremental diff is a pure comment expansion on `warnedEmptyScope` in `policy.ts`. The previous review (06ac84d) flagged the unbounded Set as a minor note worth logging; this commit addresses it directly by documenting the tradeoff and recommending a bounded LRU for high-volume ephemeral-sub deployments. No logic changed. All three issues from the original review cycle remain resolved. Ready to merge.

### Confidence: 8.8/10
- Future hardening guidance confidence: 9


```pr-review-confidence-json
{
  "style_maintainability": 9,
  "future_hardening_guidance": 9,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 9
}
```

PR #278, opened 2026-03-22
fix(cli) support localhost and tailscale daemon hosting

NicholaiVogel:
## Summary
- add a persisted `network.mode` setting with `localhost` as the default and `tailscale` mapping to a `0.0.0.0` bind
- wire the CLI setup/configure/start/status flows to read that mode and surface localhost plus tailnet access clearly
- add a dashboard settings section for network hosting mode and expose the selected bind mode in daemon status

## Testing
- `bun test packages/cli/src/lib/network.test.ts packages/cli/src/lib/daemon.test.ts`
- `bun run build:core`
- `cd packages/cli/dashboard && bun run check` *(fails on pre-existing unrelated dashboard type/a11y issues in AgentHeader.svelte, SystemInfoCard.svelte, KnowledgeTab.svelte, SessionList.svelte, and other existing files)*

### DIFF
--- a/packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
+++ b/packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
@@ -1,154 +1,162 @@
 <script lang="ts">
-	import type { ConfigFile } from "$lib/api";
-	import { st } from "$lib/stores/settings.svelte";
-	import { setSettingsDirty } from "$lib/stores/unsaved-changes.svelte";
-	import { untrack } from "svelte";
-	import AgentSection from "./settings/AgentSection.svelte";
-	import AuthSection from "./settings/AuthSection.svelte";
-	import EmbeddingsSection from "./settings/EmbeddingsSection.svelte";
-	import MemorySection from "./settings/MemorySection.svelte";
-	import PipelineSection from "./settings/PipelineSection.svelte";
-	import SearchSection from "./settings/SearchSection.svelte";
-	import TrustSection from "./settings/TrustSection.svelte";
-	import AppearanceSection from "./settings/AppearanceSection.svelte";
-	import IdentityPanel from "$lib/components/config/IdentityPanel.svelte";
-	import PageBanner from "$lib/components/layout/PageBanner.svelte";
-	import TabGroupBar from "$lib/components/layout/TabGroupBar.svelte";
-	import { ENGINE_TAB_ITEMS } from "$lib/components/layout/page-headers";
-	import { nav } from "$lib/stores/navigation.svelte";
-	import { focusEngineTab } from "$lib/stores/tab-group-focus.svelte";
-	import { invalidateAll } from "$app/navigation";
+import { invalidateAll } from "$app/navigation";
+import type { ConfigFile } from "$lib/api";
+import type IdentityPanel from "$lib/components/config/IdentityPanel.svelte";
+import PageBanner from "$lib/components/layout/PageBanner.svelte";
+import TabGroupBar from "$lib/components/layout/TabGroupBar.svelte";
+import { ENGINE_TAB_ITEMS } from "$lib/components/layout/page-headers";
+import { nav } from "$lib/stores/navigation.svelte";
+import { st } from "$lib/stores/settings.svelte";
+import { focusEngineTab } from "$lib/stores/tab-group-focus.svelte";
+import { setSettingsDirty } from "$lib/stores/unsaved-changes.svelte";
+import { untrack } from "svelte";
+import AgentSection from "./settings/AgentSection.svelte";
+import AppearanceSection from "./settings/AppearanceSection.svelte";
+import AuthSection from "./settings/AuthSection.svelte";
+import EmbeddingsSection from "./settings/EmbeddingsSection.svelte";
+import MemorySection from "./settings/MemorySection.svelte";
+import NetworkSection from "./settings/NetworkSection.svelte";
+import PipelineSection from "./settings/PipelineSection.svelte";
+import SearchSection from "./settings/SearchSection.svelte";
+import TrustSection from "./settings/TrustSection.svelte";
 
-	interface Props {
-		configFiles: ConfigFile[];
+interface Props {
+	configFiles: ConfigFile[];
+}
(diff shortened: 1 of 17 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli) support localhost and tailscale daemon hosting` (commit `2e75400c`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
```diff
@@ -1,154 +1,162 @@
...
+	},
+	{
+		id: "pipeline",
+		title: "Pipeline",
+		source: "agent",
+		paths: [["memory", "pipelineV2"]],
+	},
+	{
+		id: "trust",
+		title: "Trust",
```
NicholaiVogel, line 75: **Bug: `const` prevents reassignment of Svelte 5 `$state` primitives.**

The refactor changed three declarations from `let` to `const`:
```svelte
const activeSection = $state("agent");   // line ~75
const identityPanel = $state<IdentityPanel | null>(null); // line ~77
const identityDirty = $state(false);     // line ~78
```
In Svelte 5, assigning a new value to a `$state` primitive (e.g. `activeSection = def.id`) goes through a compiler-generated setter, but JavaScript still requires `let` for the binding to be reassignable. Using `const` makes the template's `onclick={() => (activeSection = def.id)}` a compile/runtime error, so clicking any settings tab other than "Agent" silently fails — including the new "Network" tab.

Same issue for `identityDirty` (set via `onDirtyChange`) and `identityPanel` (bound via `bind:this`). The original code correctly used `let` for all three. Revert these three declarations back to `let`.

packages/daemon/src/daemon.ts
```diff
@@ -1232,10 +1232,13 @@ const ALLOWED_ORIGINS = new Set([
 	"tauri://localhost",
```
NicholaiVogel, line 1232: **Bug: CORS allowlist never updated for tailscale mode — dashboard is unreachable from a remote browser.**

`ALLOWED_ORIGINS` is a hardcoded set containing only `http://localhost:3850` and a handful of other localhost/tauri origins:
```typescript
const ALLOWED_ORIGINS = new Set([
  "http://localhost:3850",
  "https://localhost:3850",
  "tauri://localhost",
  "http://tauri.localhost",
]);
```
When the daemon binds to `0.0.0.0` in tailscale mode, a remote browser accesses the dashboard at `http://<tailscale-ip>:3850`. That browser sends `Origin: http://<tailscale-ip>:3850`, which is not in the allowlist. Hono returns `null` for the `Access-Control-Allow-Origin` header, so every API call from the dashboard JS is blocked with a CORS error.

The tailscale bind is the entire point of this feature; the feature is non-functional for remote browsers without a CORS fix. Possible approaches:
- When `NETWORK_MODE === 'tailscale'`, accept any origin whose host+port matches `BIND_HOST:PORT` (but note `BIND_HOST` is `0.0.0.0` so you'd need to accept any IP on that port — which is the intended tailnet access pattern).
- Or: only apply the strict origin check for requests arriving on loopback; relax it for non-loopback connections on the same port.

Note: relaxing CORS for 0.0.0.0 binds is safe only when combined with non-local auth (`auth.mode != 'local'`). The dashboard's existing warning about this should be reflected in server-side enforcement, not just UI copy.

packages/daemon/src/daemon.ts
```diff
@@ -560,7 +575,7 @@ function buildPredictorHealthParams(): PredictorHealthParams {
```
NicholaiVogel, line 575: **Behavioral regression: `sidecarAlive` now returns `undefined` instead of `false` when the predictor client is null.**

Before:
```typescript
sidecarAlive: client !== null && client.isAlive(),
```
After:
```typescript
sidecarAlive: client?.isAlive(),
```
`client?.isAlive()` evaluates to `undefined` when `client` is `null`. If `PredictorHealthParams.sidecarAlive` is typed as `boolean`, this is a type error; if it's `boolean | undefined`, downstream consumers that check `=== false` will get wrong results (the sidecar would appear neither alive nor dead). The original explicit `false` was correct. Prefer:
```typescript
sidecarAlive: client?.isAlive() ?? false,
```

packages/cli/src/features/configure.ts
```diff
@@ -107,6 +117,19 @@ function readValue(yaml: string, key: string, fallback: string): string {
 	return match ? match[1].trim().replace(/^["']|["']$/g, "") : fallback;
 }
```
NicholaiVogel, line 119: **Duplicated `readNetworkMode` implementation — divergence risk.**

`configure.ts` defines its own local `readNetworkMode(yaml: string): string` that parses raw YAML text, while `@signet/core` now exports `readNetworkMode(raw: unknown): NetworkMode` that operates on a parsed object. These are semantically identical but implemented independently.

The local version using raw regex is fragile (e.g. it will fail if `mode:` is quoted or if indentation differs from exactly 2 spaces). The setup flow already imports the core version; wire `configure.ts` through `parseSimpleYaml` + the core export to keep a single source of truth.

NicholaiVogel (commented):
Two high-severity bugs: (1) Svelte 5 reactive state variables changed from `let` to `const` break tab-switching and identity-dirty tracking in SettingsTab.svelte. (2) The CORS allowlist is never updated for tailscale mode, so a remote browser accessing the dashboard via a Tailscale IP gets CORS-blocked on every API call, making the feature non-functional from a remote host. One lower-severity behavioral regression in the predictor health path. Everything else is cosmetic reformatting.

### Confidence: 7.4/10
- Functional completeness confidence: 5


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 7,
  "functional_completeness": 5,
  "pattern_correctness": 7,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Superseded by #279 from the correctly named branch: .

PR #279, opened 2026-03-22
fix(cli) support localhost and tailscale daemon hosting

NicholaiVogel:
## Summary
- add a persisted `network.mode` setting with `localhost` as the default and `tailscale` mapping to a `0.0.0.0` bind
- wire the CLI setup/configure/start/status flows to read that mode and surface localhost plus tailnet access clearly
- add a dashboard settings section for network hosting mode and expose the selected bind mode in daemon status

## Testing
- `bun test packages/cli/src/lib/network.test.ts packages/cli/src/lib/daemon.test.ts`
- `bun run build:core`
- `cd packages/cli/dashboard && bun run check` *(fails on pre-existing unrelated dashboard type/a11y issues in AgentHeader.svelte, SystemInfoCard.svelte, KnowledgeTab.svelte, SessionList.svelte, and other existing files)*

### DIFF
--- a/packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
+++ b/packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
@@ -1,154 +1,162 @@
 <script lang="ts">
-	import type { ConfigFile } from "$lib/api";
-	import { st } from "$lib/stores/settings.svelte";
-	import { setSettingsDirty } from "$lib/stores/unsaved-changes.svelte";
-	import { untrack } from "svelte";
-	import AgentSection from "./settings/AgentSection.svelte";
-	import AuthSection from "./settings/AuthSection.svelte";
-	import EmbeddingsSection from "./settings/EmbeddingsSection.svelte";
-	import MemorySection from "./settings/MemorySection.svelte";
-	import PipelineSection from "./settings/PipelineSection.svelte";
-	import SearchSection from "./settings/SearchSection.svelte";
-	import TrustSection from "./settings/TrustSection.svelte";
-	import AppearanceSection from "./settings/AppearanceSection.svelte";
-	import IdentityPanel from "$lib/components/config/IdentityPanel.svelte";
-	import PageBanner from "$lib/components/layout/PageBanner.svelte";
-	import TabGroupBar from "$lib/components/layout/TabGroupBar.svelte";
-	import { ENGINE_TAB_ITEMS } from "$lib/components/layout/page-headers";
-	import { nav } from "$lib/stores/navigation.svelte";
-	import { focusEngineTab } from "$lib/stores/tab-group-focus.svelte";
-	import { invalidateAll } from "$app/navigation";
+import { invalidateAll } from "$app/navigation";
+import type { ConfigFile } from "$lib/api";
+import type IdentityPanel from "$lib/components/config/IdentityPanel.svelte";
+import PageBanner from "$lib/components/layout/PageBanner.svelte";
+import TabGroupBar from "$lib/components/layout/TabGroupBar.svelte";
+import { ENGINE_TAB_ITEMS } from "$lib/components/layout/page-headers";
+import { nav } from "$lib/stores/navigation.svelte";
+import { st } from "$lib/stores/settings.svelte";
+import { focusEngineTab } from "$lib/stores/tab-group-focus.svelte";
+import { setSettingsDirty } from "$lib/stores/unsaved-changes.svelte";
+import { untrack } from "svelte";
+import AgentSection from "./settings/AgentSection.svelte";
+import AppearanceSection from "./settings/AppearanceSection.svelte";
+import AuthSection from "./settings/AuthSection.svelte";
+import EmbeddingsSection from "./settings/EmbeddingsSection.svelte";
+import MemorySection from "./settings/MemorySection.svelte";
+import NetworkSection from "./settings/NetworkSection.svelte";
+import PipelineSection from "./settings/PipelineSection.svelte";
+import SearchSection from "./settings/SearchSection.svelte";
+import TrustSection from "./settings/TrustSection.svelte";
 
-	interface Props {
-		configFiles: ConfigFile[];
+interface Props {
+	configFiles: ConfigFile[];
+}
+
+const { configFiles }: Props = $props();
+
+interface SectionDef {
+	id: string;
+	title: string;
+	paths: string[][];
+	source: "agent" | "config";
+}
+
+const sectionDefs: SectionDef[] = [
+	{
+		id: "agent",
+		title: "Agent",
+		source: "agent",
+		paths: [["agent", "name"], ["agent", "description"], ["harnesses"]],
+	},
+	{
+		id: "network",
+		title: "Network",
+		source: "agent",
+		paths: [["network"]],
+	},
+	{
+		id: "embeddings",
+		title: "Embeddings",
+		source: "config",
+		paths: [["embedding"], ["memory", "embeddings"], ["embeddings"]],
+	},
+	{
+		id: "memory",
+		title: "Memory",
+		source: "config",
+		paths: [["memory", "session_budget"], ["memory", "current_md_budget"], ["memory", "decay_rate"], ["paths"]],
+	},
+	{
+		id: "search",
+		title: "Search",
+		source: "config",
+		paths: [["search"]],
+	},
+	{
+		id: "pipeline",
+		title: "Pipeline",
+		source: "agent",
+		paths: [["memory", "pipelineV2"]],
+	},
+	{
+		id: "trust",
+		title: "Trust",
+		source: "agent",
+		paths: [["trust"]],
+	},
+	{
+		id: "auth",
+		title: "Auth",
+		source: "config",
+		paths: [["auth"]],
+	},
+	{
+		id: "appearance",
+		title: "Appearance",
+		source: "config",
+		paths: [],
+	},
+];
+
+const tabBtn =
+	"px-2.5 py-1 text-[10px] font-medium uppercase tracking-[0.06em] rounded-md transition-colors duration-150 border-none cursor-pointer whitespace-nowrap";
+const tabActive = `${tabBtn} text-[var(--sig-highlight)] bg-[color-mix(in_srgb,var(--sig-highlight),var(--sig-bg)_90%)] border border-[color-mix(in_srgb,var(--sig-highlight),transparent_85%)]`;
+const tabInactive = `${tabBtn} bg-transparent text-[var(--sig-text-muted)] hover:text-[var(--sig-highlight)] hover:bg-[color-mix(in_srgb,var(--sig-highlight),var(--sig-bg)_94%)]`;
+
+const activeSection = $state("agent");
+let discardDialogOpen = $state(false);
+const identityPanel = $state<IdentityPanel | null>(null);
+const identityDirty = $state(false);
+
+const sections = $derived(
+	sectionDefs.map((def) => ({
+		id: def.id,
+		title: def.title,
+		dirty: st.isAnyPathDirty(def.source, def.paths),
+	})),
+);
+
+const combinedDirty = $derived(st.isDirty || identityDirty);
+
+$effect(() => {
+	const files = configFiles;
+	untrack(() => st.init(files));
+});
+
+$effect(() => {
+	setSettingsDirty(combinedDirty);
+	return () => {
+		setSettingsDirty(false);
+	};
+});
+
+async function saveAll(): Promise<void> {
+	const promises: Promise<void>[] = [];
+	if (st.isDirty) promises.push(st.save());
+	if (identityDirty && identityPanel) promises.push(identityPanel.save());
+	await Promise.all(promises);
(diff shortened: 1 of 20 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli) support localhost and tailscale daemon hosting` (commit `2e75400c`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
```diff
@@ -1,154 +1,162 @@
...
+		title: "Search",
+		source: "config",
+		paths: [["search"]],
+	},
+	{
+		id: "pipeline",
+		title: "Pipeline",
+		source: "agent",
+		paths: [["memory", "pipelineV2"]],
+	},
```
NicholaiVogel, line 72: **Bug: `const` breaks reactive state for primitive/null `$state` variables.**

In Svelte 5 runes mode, `$state` primitives and nullable references must be declared with `let` so the compiler can instrument reassignments. Declaring them `const` means:
- `activeSection = def.id` (tab-click handler) → silent no-op or strict-mode TypeError; clicking tabs does nothing.
- `bind:this={identityPanel}` cannot write back the component instance; `identityPanel` stays `null` forever.
- `onDirtyChange={(dirty) => (identityDirty = dirty)}` is a no-op; unsaved-changes tracking breaks.

Revert these three to `let`:
```svelte
let activeSection = $state("agent");
let identityPanel = $state<IdentityPanel | null>(null);
let identityDirty = $state(false);
```
`discardDialogOpen` was correctly left as `let`.

packages/cli/dashboard/src/lib/components/tabs/SettingsTab.svelte
```diff
@@ -1,154 +1,162 @@
...
+import PageBanner from "$lib/components/layout/PageBanner.svelte";
+import TabGroupBar from "$lib/components/layout/TabGroupBar.svelte";
+import { ENGINE_TAB_ITEMS } from "$lib/components/layout/page-headers";
+import { nav } from "$lib/stores/navigation.svelte";
+import { st } from "$lib/stores/settings.svelte";
+import { focusEngineTab } from "$lib/stores/tab-group-focus.svelte";
+import { setSettingsDirty } from "$lib/stores/unsaved-changes.svelte";
+import { untrack } from "svelte";
+import AgentSection from "./settings/AgentSection.svelte";
+import AppearanceSection from "./settings/AppearanceSection.svelte";
```
NicholaiVogel, line 14: **Bug: `import type IdentityPanel` removes the runtime binding needed for the template.**

`import type` is erased by TypeScript before Svelte sees the JS output, so the `<IdentityPanel bind:this={identityPanel} …/>` tag in the template references an undefined identifier at runtime. Change back to a value import:
```ts
import IdentityPanel from "$lib/components/config/IdentityPanel.svelte";
```
The type annotation `$state<IdentityPanel | null>(null)` continues to work with a value import.

packages/cli/dashboard/src/lib/components/tabs/settings/NetworkSection.svelte
```diff
@@ -0,0 +1,43 @@
...
+	return "localhost";
+}
+</script>
+
+{#if st.agentFile}
+	<FormSection description="Daemon bind mode. localhost keeps Signet on 127.0.0.1 only. tailscale keeps localhost working and also binds 0.0.0.0 so other devices on your tailnet can reach it. Restart the daemon after saving.">
+		<FormField label="Hosting mode" description="If auth.mode is local, anyone on your trusted tailnet can access the dashboard and API when tailscale mode is enabled.">
+			<Select.Root
+				type="single"
+				value={st.aStr(["network", "mode"]) || "localhost"}
```
NicholaiVogel: **Future hardening:** The description warns that `tailscale` mode exposes the API to anyone on the tailnet when `auth.mode` is `local`, but this is only surfaced as passive help text. Consider reading the current `auth.mode` from `st.aStr(["auth", "mode"])` at render time and displaying a prominent inline warning (or blocking the save) when a user selects `tailscale` with `local` auth. This prevents accidental unauthenticated exposure on private networks.

NicholaiVogel (commented):
Two correctness bugs in SettingsTab.svelte break the Settings tab: (1) three reactive state variables were changed from `let` to `const` — primitives and null values managed by Svelte 5 `$state` runes must use `let` so they can be reassigned; (2) `IdentityPanel` was changed to a type-only import (`import type`) but it is also used as a component in the template, which strips the runtime value. Everything else (network binding logic, YAML parsing, daemon env-var flow, CLI wiring) looks correct.

### Confidence: 7.2/10
- Style consistency & maintainability: 6
- Merge conflict detection confidence: 5
- Future hardening guidance confidence: 6


```pr-review-confidence-json
{
  "style_maintainability": 6,
  "merge_conflict_detection": 5,
  "future_hardening_guidance": 6,
  "scope_alignment": 9,
  "duplication_awareness": 7,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Addressed the feedback from #278 in follow-up commit `ca0552c5`.

Handled:
- restored `let` bindings for the Svelte 5 `$state` values in `SettingsTab.svelte` and switched `IdentityPanel` back to a runtime import
- removed the duplicate regex-based network-mode parser in `configure.ts` and now use `parseSimpleYaml` + the shared core helper
- fixed predictor health to keep `sidecarAlive` a stable boolean (`false` when no client is present)
- relaxed daemon origin handling for tailscale mode so same-port remote dashboard/API origins are accepted instead of localhost-only
- mirrored the network binding behavior into `packages/daemon-rs/` and exposed the same host/bind/network mode in the Rust status path

Verification:
- `bun test packages/cli/src/lib/network.test.ts packages/cli/src/lib/daemon.test.ts`
- `cargo test -p signet-core`
- `cargo test -p signet-daemon --no-run`

Still pre-existing and unrelated:
- `cd packages/cli/dashboard && bun run check` still fails on existing dashboard type/a11y issues
- `cd packages/cli/dashboard && bun run build` still fails on the existing `pdf-parse` resolution issue from `@signet/core/dist`

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix(cli) support localhost and tailscale daemon hosting` (commit `ca0552c5`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon.ts
```diff
@@ -262,6 +265,47 @@ function normalizeLoopbackHost(host: string): string {
...
 
+function parseOriginPort(url: URL): number | null {
+	if (url.port.length > 0) {
+		const port = Number.parseInt(url.port, 10);
+		return Number.isInteger(port) ? port : null;
+	}
+	if (url.protocol === "http:") return 80;
+	if (url.protocol === "https:") return 443;
+	return null;
+}
```
NicholaiVogel, line 276: **Overly broad CORS in tailscale mode.** `isAllowedOrigin` accepts any origin whose port equals `PORT` (3850). The intended goal is to allow the daemon dashboard when accessed via a Tailscale IP (e.g. `http://100.64.x.x:3850`), but the current check also permits any other site running at port 3850 — including local dev servers or malicious pages — to make credentialed cross-origin requests to the daemon.

Consider tightening the check: require the origin host to be a non-loopback, non-`localhost` address (i.e. look like a Tailscale IP range or at least not match the static allowlist addresses), in addition to the port check. A simple guard would be:
```ts
const host = url.hostname;
const isLoopback = host === '127.0.0.1' || host === 'localhost' || host === '::1';
if (isLoopback) return false; // already covered by ALLOWED_ORIGINS
return parseOriginPort(url) === PORT;
```
This won't prevent all abuse, but it eliminates the case where a local service on port 3850 gets unintended CORS access. The broader hardening note: when `auth.mode` is `local`, the existing UI warning is the main control; this is a belt-and-suspenders improvement.

NicholaiVogel (commented):
Both previously flagged correctness bugs are fixed: `IdentityPanel` is now a value import (not type-only), and the three `$state` variables are correctly declared with `let`. The `readNetworkMode` call in `configure.ts` now properly passes a parsed object via `parseSimpleYaml`. One new issue introduced in this increment: the `isAllowedOrigin` CORS expansion in tailscale mode is over-broad — it accepts any origin whose port matches the daemon port (3850), which allows arbitrary sites on port 3850 to make credentialed cross-origin requests to the daemon. The intended case (dashboard served at `http://<tailscale-ip>:3850`) would be more precisely served by checking that the origin host is also a non-loopback address, rather than accepting every origin at that port.

### Confidence: 7.5/10
- Merge conflict detection confidence: 5


```pr-review-confidence-json
{
  "style_maintainability": 7,
  "merge_conflict_detection": 5,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 6
}
```

NicholaiVogel:
Addressing the remaining feedback in `229ad51c`:

- tightened tailscale-mode CORS so the extra origin allowance no longer accepts arbitrary `*:3850` origins; it now only admits Tailscale hosts (`100.64.0.0/10`, `fd7a:115c:a1e0::/48`, and `.ts.net`) while keeping the explicit localhost dev allowlist intact
- added `packages/daemon/src/daemon-cors.test.ts` to lock that behavior down
- added a prominent inline warning in the dashboard when `network.mode = tailscale` and `auth.mode = local`

Verified with:
- `bun test packages/daemon/src/daemon-cors.test.ts`

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix(cli) support localhost and tailscale daemon hosting` (commit `229ad51c`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/daemon-cors.test.ts
```diff
@@ -0,0 +1,66 @@
...
+		process.env.SIGNET_PATH = dir;
+
+		const daemon = await import("./daemon");
+		app = daemon.app;
+	});
+
+	afterAll(() => {
+		if (prev === undefined) {
+			process.env.SIGNET_PATH = undefined;
+		} else {
```
NicholaiVogel: Assigning `undefined` to `process.env.SIGNET_PATH` sets it to the string `"undefined"` in Node/Bun rather than removing it. Use `delete process.env.SIGNET_PATH` to properly restore the absent-variable case:
```ts
if (prev === undefined) {
  delete process.env.SIGNET_PATH;
} else {
  process.env.SIGNET_PATH = prev;
}
```
This is test-only and won't affect production, but it means any test that runs after this suite and checks for an unset `SIGNET_PATH` would see the string `"undefined"` instead.

NicholaiVogel (commented):
The previously flagged over-broad CORS expansion is now properly fixed: `isAllowedOrigin` validates that the origin host is a genuine Tailscale address (100.64–127.x CGNAT range, .ts.net MagicDNS, or fd7a:115c:a1e0:: IPv6 prefix) and explicitly rejects loopback hosts in tailscale mode. Tests cover the key allow/reject cases. The NetworkSection `$derived` refactor is correct (Svelte 5 `$derived` runes correctly use `const`). One minor test hygiene issue: `process.env.SIGNET_PATH = undefined` may set the var to the string `"undefined"` rather than clearing it — use `delete process.env.SIGNET_PATH` instead.

### Confidence: 7.9/10
- Merge conflict detection confidence: 5


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "merge_conflict_detection": 5,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 8,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Addressed the latest test-hygiene note in `792d116f`.

`packages/daemon/src/daemon-cors.test.ts` now restores an originally-unset `SIGNET_PATH` with `delete process.env.SIGNET_PATH` instead of assigning `undefined`, so later tests do not see the string `"undefined"`.

Verified with:
- `bun test packages/daemon/src/daemon-cors.test.ts`

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix(cli) support localhost and tailscale daemon hosting` (commit `792d116f`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (commented):
The only change in this increment is the exact fix requested in the prior review: `process.env.SIGNET_PATH = undefined` (which would set the variable to the string "undefined") is replaced with `delete process.env.SIGNET_PATH`, and the `else` branch is rewritten as a follow-up `if (prev !== undefined)` guard. The fix is correct and the logic is equivalent. All previously flagged issues are now resolved. No new issues introduced.

### Confidence: 8.4/10
- Merge conflict detection confidence: 5


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "merge_conflict_detection": 5,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

PR #282, opened 2026-03-22
fix: task run output display — stdin close, chunk normalization, terminal rendering

aaf2tbz:
## Summary

- **spawn.ts**: Close stdin immediately after `nodeSpawn`. The `claude` CLI waited 3s for stdin before proceeding, causing delayed output and a "no stdin data received" warning that polluted stderr in the run log.
- **tasks.svelte.ts**: Fix `normalizeOutputChunk` dropping plain-text output. Two bugs: (a) pure-JSON chunks with no recognized events returned `""` instead of the cleaned text; (b) JSON events mixed with plain text silently dropped the plain-text lines. Added `sawAnyNonJsonLine` tracking — only suppress for pure JSON streams, only extract events from exclusively-JSON chunks.
- **RunLog.svelte**: Rewrite output display as terminal-style renderer matching `TroubleshooterPanel`. Replaces `<pre>` blocks with line-by-line ANSI-to-HTML rendering, scrollable `role="log"` viewport, stderr in amber, stdout in normal text, auto-scroll to bottom on expand, and a `min-h-[1em]` spacer per line to preserve blank lines.

## Test plan

- [ ] Trigger a task run using the `claude-code` harness
- [ ] Expand the run log card — stdout output should appear without delay, no "no stdin data received" warning in stderr
- [ ] Trigger a task run using the `opencode` harness — JSON events should be parsed into clean text, no raw JSONL visible
- [ ] Verify stderr lines appear in amber, stdout in the default text color
- [ ] Verify auto-scroll to bottom when a long run is expanded

### DIFF
--- a/packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
+++ b/packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
@@ -2,6 +2,7 @@
 import type { TaskRun } from "$lib/api";
 import * as Card from "$lib/components/ui/card/index.js";
 import { Badge } from "$lib/components/ui/badge/index.js";
+import { tick } from "svelte";
 
 interface Props {
 	run: TaskRun;
@@ -10,6 +11,7 @@ interface Props {
 let { run }: Props = $props();
 
 let expanded = $state(false);
+let termRef: HTMLDivElement | undefined = $state(undefined);
 
 function formatDate(iso: string | null): string {
 	if (!iso) return "—";
@@ -32,6 +34,47 @@ const statusColors: Record<string, string> = {
 	completed: "var(--sig-success)",
 	failed: "var(--sig-error, #ef4444)",
 };
+
+function escapeHtml(s: string): string {
+	return s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+}
+
+const ANSI_COLORS: Record<string, string> = {
+	"31": "var(--sig-error, #ef4444)",
+	"32": "var(--sig-success)",
+	"33": "var(--sig-warning, #e8a832)",
+	"90": "var(--sig-text-muted)",
+	"36": "var(--sig-accent)",
+	"34": "var(--sig-highlight)",
+	"35": "var(--sig-accent)",
+};
+
+function ansiToHtml(text: string): string {
+	let result = escapeHtml(text);
+	result = result.replace(/\x1b\[(\d+)m([\s\S]*?)\x1b\[0m/g, (_, code, content) => {
+		const color = ANSI_COLORS[code];
+		if (color) return `<span style="color:${color}">${content}</span>`;
+		if (code === "1") return `<strong>${content}</strong>`;
+		return content;
+	});
+	return result.replace(/\x1b\[\d+m/g, "");
+}
+
+function toLines(text: string | null): string[] {
+	if (!text) return [];
+	return text.split("\n");
+}
+
+async function scrollToBottom(): Promise<void> {
+	await tick();
+	termRef?.scrollTo({ top: termRef.scrollHeight, behavior: "smooth" });
+}
+
+$effect(() => {
+	if (expanded && (run.stdout || run.stderr || run.error)) {
+		void scrollToBottom();
+	}
+});
 </script>
 
 <button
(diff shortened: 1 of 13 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: task run output display — stdin close, chunk normalization, terminal rendering` (commit `13fd48ac`) now and I'll follow up shortly with feedback.

aaf2tbz:
Closing — was based off wrong branch. Reopening clean.

packages/daemon/src/daemon.ts
```diff
@@ -5332,16 +5332,16 @@ app.post("/api/hooks/session-end", async (c) => {
...
-		// Always release session claim and agent presence, even if extraction throws
-		if (sessionKey) {
-			releaseSession(sessionKey);
-			removeAgentPresence(sessionKey);
+		try {
+			const result = await handleSessionEnd(body);
+			return c.json(result);
+		} finally {
+			// Always release session claim and agent presence, even if extraction throws
+			if (sessionKey) {
```
NicholaiVogel, line 5340: This indentation change is a functional bug fix, not cosmetic. In the original code the inner `try { handleSessionEnd } finally { releaseSession }` block was at the same nesting level as the outer catch, meaning exceptions thrown by `handleSessionEnd` were NOT caught by `catch (e)` — the 500 error response was never sent on hook failure. The fix properly nests the inner try/finally inside the outer try so the catch works as intended. Worth calling out in the commit message or a follow-up test.

packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
```diff
@@ -32,6 +34,47 @@ const statusColors: Record<string, string> = {
...
+};
+
+function ansiToHtml(text: string): string {
+	let result = escapeHtml(text);
+	result = result.replace(/\x1b\[(\d+)m([\s\S]*?)\x1b\[0m/g, (_, code, content) => {
+		const color = ANSI_COLORS[code];
+		if (color) return `<span style="color:${color}">${content}</span>`;
+		if (code === "1") return `<strong>${content}</strong>`;
+		return content;
+	});
```
NicholaiVogel, line 59: The `ansiToHtml` regex applies to the HTML-escaped string, so `content` is safe from XSS. However, the regex only processes sequences of the form `\x1b[Nm…\x1b[0m`. Nested or overlapping codes (e.g. `\x1b[1m\x1b[31mtext\x1b[0m\x1b[0m`) leave a stray `\x1b[31m` inside the already-emitted `<strong>` element; the trailing strip removes the escape prefix but leaves raw binary 0x1B characters visible in the DOM. Not a security issue, but will produce garbled output in logs that use nested color+bold. Consider stripping all remaining `\x1b[…m` sequences before the final return rather than only the prefix pattern `/\x1b\[\d+m/g`.

packages/daemon/src/pipeline/extraction.ts
```diff
@@ -95,12 +95,26 @@ export function stripFences(raw: string): string {
...
+	// If the input starts with JSON structure, return it as-is — don't
+	// extract sub-arrays from inside a valid JSON object. A truncated/
+	// malformed object starting with '{' will pass through here, but
+	// downstream tryParseJson handles that gracefully (returns null).
+	if (trimmed.startsWith("{") || trimmed.startsWith("[")) {
+		return trimmed;
+	}
+
+	// Try extracting a balanced JSON object first (e.g. "Here is the data: {...}")
+	const obj = extractBalancedJsonObject(trimmed);
```
NicholaiVogel, line 109: The early-return `if (trimmed.startsWith("{") || trimmed.startsWith("[")) return trimmed` skips `extractBalancedJsonObject` and `extractBalancedJsonArray` for inputs that start with JSON structure but have trailing prose (e.g. `{...} Note: …`). The comment acknowledges that malformed objects are handled gracefully downstream, but valid-looking objects with trailing text that confuse `tryParseJson` will now produce a null parse rather than the recovered sub-object. This is an edge case that exists only for unusual LLM outputs; the trade-off of not double-extracting from inside a valid object is correct.

NicholaiVogel (approved):
Three high-signal findings: (1) The daemon.ts indentation fix is a functional bug fix — `handleSessionEnd` errors were not being caught by the outer `catch` block before this change, so the 500 error response was silently dropped on failure. (2) `ansiToHtml` is safe from XSS because `escapeHtml` runs first, but nested/unrecognized ANSI sequences produce stray ESC bytes in rendered HTML (display issue, not security). (3) The `toBeGreaterThanOrEqual(23)` assertion in tools.test.ts silently allows net tool deletions in the range 21–22, which the original exact-count test would have caught. All three core fixes (stdin close, normalizeOutputChunk, `..` path check) are correct. No blocking issues; ready to approve with notes.

### Confidence: 8.0/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

- packages/daemon/src/mcp/tools.test.ts:122 - Changing from `toBe(21)` to `toBeGreaterThanOrEqual(23)` means any deletion that leaves the count between 21 and 22 would pass silently. The comment correctly explains the intent (additions pass, deletions fail), but the floor of 23 is higher than the old exact count of 21, implying 2 new tools were added outside this diff. If those 2 tools are not in this PR, please verify they were intentionally registered. Also consider tightening to `toBeGreaterThanOrEqual(21)` (the previous known-good count) or keeping an exact-count assertion in a separate test to catch accidental additions of unbounded proxy tools.

PR #283, opened 2026-03-22
fix: task run output display — stdin close, chunk normalization, terminal rendering

aaf2tbz:
## Summary

- **spawn.ts**: Close stdin immediately after `nodeSpawn`. The `claude` CLI waited 3s for stdin before proceeding, causing delayed output and a "no stdin data received" warning that polluted stderr in the run log.
- **tasks.svelte.ts**: Fix `normalizeOutputChunk` dropping plain-text output. Two bugs: (a) pure-JSON chunks with no recognized events returned `""` instead of the cleaned text; (b) JSON events mixed with plain text silently dropped the plain-text lines. Added `sawAnyNonJsonLine` tracking — only suppress for pure JSON streams, only extract events from exclusively-JSON chunks.
- **RunLog.svelte**: Rewrite output display as terminal-style renderer matching `TroubleshooterPanel`. Replaces `<pre>` blocks with line-by-line ANSI-to-HTML rendering, scrollable `role="log"` viewport, stderr in amber, stdout in normal text, auto-scroll to bottom on expand, and `min-h-[1em]` per line to preserve blank lines.

## Test plan

- [ ] Trigger a task run using the `claude-code` harness — stdout appears without delay, no "no stdin data received" warning in stderr
- [ ] Trigger a task run using the `opencode` harness — JSON events parsed to clean text, no raw JSONL visible
- [ ] Verify stderr lines appear in amber, stdout in default text color
- [ ] Verify auto-scroll to bottom when a long run is expanded

### DIFF
--- a/packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
+++ b/packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
@@ -49,6 +49,9 @@ const ANSI_COLORS: Record<string, string> = {
 	"35": "var(--sig-accent)",
 };
 
+// INVARIANT: escapeHtml must run on the raw input before any HTML tags are
+// inserted. {@html} in the template relies on this — any future change that
+// interpolates user-controlled content into replacement strings will introduce XSS.
 function ansiToHtml(text: string): string {
 	let result = escapeHtml(text);
 	// Match compound CSI sequences like \x1b[1;31m as well as simple \x1b[31m.
--- a/packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
+++ b/packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
@@ -2,6 +2,7 @@
 import type { TaskRun } from "$lib/api";
 import * as Card from "$lib/components/ui/card/index.js";
 import { Badge } from "$lib/components/ui/badge/index.js";
+import { tick } from "svelte";
 
 interface Props {
 	run: TaskRun;
@@ -10,6 +11,7 @@ interface Props {
 let { run }: Props = $props();
 
 let expanded = $state(false);
+let termRef: HTMLDivElement | undefined = $state(undefined);
 
 function formatDate(iso: string | null): string {
 	if (!iso) return "—";
@@ -32,6 +34,47 @@ const statusColors: Record<string, string> = {
 	completed: "var(--sig-success)",
 	failed: "var(--sig-error, #ef4444)",
 };
+
+function escapeHtml(s: string): string {
+	return s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/>/g, "&gt;");
+}
+
+const ANSI_COLORS: Record<string, string> = {
+	"31": "var(--sig-error, #ef4444)",
+	"32": "var(--sig-success)",
+	"33": "var(--sig-warning, #e8a832)",
+	"90": "var(--sig-text-muted)",
+	"36": "var(--sig-accent)",
+	"34": "var(--sig-highlight)",
+	"35": "var(--sig-accent)",
+};
+
+function ansiToHtml(text: string): string {
+	let result = escapeHtml(text);
+	result = result.replace(/\x1b\[(\d+)m([\s\S]*?)\x1b\[0m/g, (_, code, content) => {
+		const color = ANSI_COLORS[code];
+		if (color) return `<span style="color:${color}">${content}</span>`;
+		if (code === "1") return `<strong>${content}</strong>`;
+		return content;
+	});
+	return result.replace(/\x1b\[\d+m/g, "");
+}
+
+function toLines(text: string | null): string[] {
+	if (!text) return [];
+	return text.split("\n");
+}
+
+async function scrollToBottom(): Promise<void> {
+	await tick();
+	termRef?.scrollTo({ top: termRef.scrollHeight, behavior: "smooth" });
+}
+
+$effect(() => {
+	if (expanded && (run.stdout || run.stderr || run.error)) {
+		void scrollToBottom();
+	}
+});
 </script>
 
 <button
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: task run output display — stdin close, chunk normalization, terminal rendering` (commit `b4445300`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
```diff
@@ -32,6 +34,47 @@ const statusColors: Record<string, string> = {
...
+	"36": "var(--sig-accent)",
+	"34": "var(--sig-highlight)",
+	"35": "var(--sig-accent)",
+};
+
+function ansiToHtml(text: string): string {
+	let result = escapeHtml(text);
+	result = result.replace(/\x1b\[(\d+)m([\s\S]*?)\x1b\[0m/g, (_, code, content) => {
+		const color = ANSI_COLORS[code];
+		if (color) return `<span style="color:${color}">${content}</span>`;
```
NicholaiVogel: **Bug: multi-parameter ANSI sequences are not stripped or rendered.**

The regex `\x1b\[(\d+)m([\s\S]*?)\x1b\[0m` only matches a single decimal parameter (e.g. `\x1b[31m`). Compound CSI sequences like `\x1b[1;31m` (bold+red), `\x1b[0;32m` (reset+green), or `\x1b[38;5;123m` (256-color) match neither the capture regex nor the cleanup regex `/\x1b\[\d+m/g`, so they survive into the rendered HTML as visible junk characters.

This doesn't affect the claude-code plain-text path because `stripAnsi` in `normalizeOutputChunk` removes all codes before storage. It does affect the opencode JSON path: `parsed.part.text` values are stored without stripping, so any compound ANSI codes in opencode/codex output will leak through.

Fix: replace the cleanup regex with the same comprehensive pattern already used in `tasks.svelte.ts`:

```ts
const ANSI_RE =
  /[\u001B\u009B][[\]()#;?]*(?:(?:[a-zA-Z\d]*(?:;[a-zA-Z\d]*)*)?\u0007|(?:\d{1,4}(?:;\d{0,4})*)?[\dA-PR-TZcf-nq-uy=><~])/g;
```

And use it both for the capture loop (to handle compound codes) and as the final cleanup pass.

NicholaiVogel (commented):
Three targeted, well-scoped fixes. The stdin close and `normalizeOutputChunk` logic are correct. The `{@html}` XSS surface is mitigated by `escapeHtml` running first. One real rendering bug: `ansiToHtml` only handles single-parameter CSI sequences (`\x1b[31m`), so compound sequences like `\x1b[1;31m` (bold+color, common in CLI output) are not stripped and will appear as raw escape characters in the browser. For the claude-code plain-text path this doesn't matter because `stripAnsi` in `normalizeOutputChunk` already removes all codes before storage. It matters for the opencode JSON path, where `extracted.part.text` values may embed ANSI and are stored without stripping before reaching `ansiToHtml`. One secondary issue: the `normalizeOutputChunk` return path for mixed chunks returns `cleanChunk` (which includes raw JSON lines alongside the plain-text lines), potentially leaking structured internal data into the UI. This is intentional per the comment but worth acknowledging.

### Confidence: 8.0/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
```diff
@@ -32,6 +34,47 @@ const statusColors: Record<string, string> = {
...
+		return content;
+	});
+	return result.replace(/\x1b\[\d+m/g, "");
+}
+
+function toLines(text: string | null): string[] {
+	if (!text) return [];
+	return text.split("\n");
+}
+
```
NicholaiVogel, line 78: **`{@html}` trust boundary — confirm and document.**

`{@html ansiToHtml(line)}` bypasses Svelte's auto-escaping. The implementation is safe *today* because `escapeHtml` runs first and the only injected HTML comes from the controlled `ANSI_COLORS` map and `<strong>`. But this is a fragile guarantee: any future change that adds a new tag or interpolates user-controlled content into the replacement string will introduce XSS.

Consider adding a brief comment to the `ansiToHtml` function contract:
```ts
// INVARIANT: must call escapeHtml on the input before inserting any HTML tags.
// {@html} in the template relies on this.
```
This makes the safety invariant explicit for reviewers.

packages/cli/dashboard/src/lib/stores/tasks.svelte.ts
```diff
@@ -190,12 +195,17 @@ function normalizeOutputChunk(chunk: string): string {
...
-	if (extractedAnyEvent) return extractedText;
-	if (sawAnyJsonLine) return "";
+	// Only extract events when the chunk is purely structured JSON (no plain text
+	// mixed in). OpenCode/Codex output is pure JSONL; claude-code is plain text.
+	// If there's any non-JSON content, fall back to the full cleaned output so
+	// error messages and plain text aren't silently dropped.
+	if (extractedAnyEvent && !sawAnyNonJsonLine) return extractedText;
+	if (sawAnyJsonLine && !sawAnyNonJsonLine) return "";
 	return cleanChunk;
 }
```
NicholaiVogel, line 215: **Mixed-chunk fallback leaks raw JSON into the UI.**

When `sawAnyNonJsonLine && sawAnyJsonLine` (a chunk with both plain text and JSON lines), the function returns `cleanChunk`, which includes the raw JSONL alongside the plain text. This exposes structured internal protocol data (tool names, model parameters, etc.) verbatim in the run log.

The comment acknowledges this, but it may produce confusing output in practice — for example if the harness emits a startup banner followed by JSON events. A safer fallback might be to return only the non-JSON lines in that case:
```ts
// Collect plain-text lines separately:
const plainLines: string[] = [];
// ...in the loop:
if (!trimmed.startsWith("{")) {
  sawAnyNonJsonLine = true;
  plainLines.push(line);
  continue;
}
// ...at the end:
if (extractedAnyEvent && !sawAnyNonJsonLine) return extractedText;
if (sawAnyJsonLine && !sawAnyNonJsonLine) return "";
// Mixed: return only the non-JSON lines
if (sawAnyJsonLine && sawAnyNonJsonLine) return plainLines.join("\n");
return cleanChunk;
```
Not a blocking issue, but worth addressing before this pattern reaches production with real harnesses.

packages/daemon/src/scheduler/spawn.ts
```diff
@@ -88,6 +88,10 @@ export async function spawnTask(
 		env: { ...baseEnv, SIGNET_NO_HOOKS: "1" } as NodeJS.ProcessEnv,
 	});
 
+	// Close stdin immediately — we never write to it, and leaving it open
```
NicholaiVogel, line 91: **Inner kill timer is not cancellable — minor resource leak on normal exit.**

When the outer `SIGTERM` timer fires and the process exits cleanly before the 5-second inner `kill()` timer, the inner timer is never cleared. `clearTimeout(timer)` only cancels the outer timer. In practice this is harmless (the `try { child.kill() } catch {}` is a no-op on a dead process), but it leaves a dangling timer per timed-out run. Consider tracking and clearing the inner timer:

```ts
let forceKillTimer: ReturnType<typeof setTimeout> | null = null;
const timer = setTimeout(() => {
  timedOut = true;
  child.kill("SIGTERM");
  forceKillTimer = setTimeout(() => { try { child.kill(); } catch {} }, 5000);
}, timeoutMs);
// ...
clearTimeout(timer);
if (forceKillTimer) clearTimeout(forceKillTimer);
```

aaf2tbz:
Fixed in c819aa61. Extended `ansiToHtml` to handle compound CSI sequences like `\x1b[1;31m`: the capture group is now `(\d+(?:;\d+)*)`, codes are split on `;`, the first known color entry wins, and bold is applied when `1` is present. The cleanup strip at the end also updated to match compound sequences. The mixed-chunk JSON leakage is intentional per the comment — that case only applies when there's non-JSON content mixed in, which means it's plain CLI output anyway.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: task run output display — stdin close, chunk normalization, terminal rendering` (commit `c819aa61`); taking another look at the fixes and I will report back shortly.

packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
```diff
@@ -32,6 +34,55 @@ const statusColors: Record<string, string> = {
...
+	// Match compound CSI sequences like \x1b[1;31m as well as simple \x1b[31m.
+	// For compound codes, prefer the last numeric segment as the color code and
+	// apply bold when "1" is present.
+	result = result.replace(/\x1b\[(\d+(?:;\d+)*)m([\s\S]*?)\x1b\[0m/g, (_, codes, content) => {
+		const parts = codes.split(";");
+		const bold = parts.includes("1");
+		const color = parts.map((c: string) => ANSI_COLORS[c]).find(Boolean);
+		let out = content;
+		if (color) out = `<span style="color:${color}">${out}</span>`;
+		if (bold) out = `<strong>${out}</strong>`;
```
NicholaiVogel, line 66: Previously fixed: compound CSI sequences (`\x1b[1;31m`) are now captured by the updated regex and `codes.split(";")` parsing. Bold detection via `parts.includes("1")` and color via `.find(Boolean)` are both correct. The trailing-strip regex also updated. Prior review concern is resolved.

packages/cli/dashboard/src/lib/components/tasks/RunLog.svelte
```diff
@@ -32,6 +34,55 @@ const statusColors: Record<string, string> = {
...
+	"35": "var(--sig-accent)",
+};
+
+function ansiToHtml(text: string): string {
+	let result = escapeHtml(text);
+	// Match compound CSI sequences like \x1b[1;31m as well as simple \x1b[31m.
+	// For compound codes, prefer the last numeric segment as the color code and
+	// apply bold when "1" is present.
+	result = result.replace(/\x1b\[(\d+(?:;\d+)*)m([\s\S]*?)\x1b\[0m/g, (_, codes, content) => {
+		const parts = codes.split(";");
```
NicholaiVogel, line 61: Minor edge case: if two opening CSI sequences appear without a reset between them (e.g. `\x1b[1m\x1b[31m text \x1b[0m`), the lazy `[\s\S]*?` in the first-pass regex captures the inner `\x1b[31m` as literal content inside the outer match. The trailing-strip regex removes it, so no raw bytes leak — the inner color is just silently dropped. This is acceptable for a best-effort renderer, but worth noting if richer nested-sequence support is ever needed.

NicholaiVogel (dismissed):
The incremental change directly addresses the prior review's compound-CSI bug. `ansiToHtml` now correctly parses multi-parameter sequences like `\x1b[1;31m` (bold+color) by splitting on `;`, checking for `1` for bold, and scanning all segments for a color code. The trailing-strip regex is also updated to handle compound sequences. No new bugs or security issues are introduced. Prior concern about `escapeHtml` running before the ANSI regex remains safe: `;` and digits are not affected by HTML escaping, so regex matching is unaffected. The second prior concern (mixed-chunk path returning raw JSONL alongside plain text) is unchanged and intentional per the comment — acknowledged but not a blocker. The `content` captured inside ANSI spans is already HTML-escaped, so the `{@html}` injection surface remains safe.

### Confidence: 8.2/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 10,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

aaf2tbz:
Addressed all three inline concerns in 52740d82:

- **Mixed-chunk JSON leak** — plain-text lines now collected separately during the loop; the mixed path returns only those lines instead of `cleanChunk`, so raw JSONL never reaches the UI.
- **`{@html}` invariant** — explicit comment on `ansiToHtml` documenting that `escapeHtml` must run on raw input before any HTML is inserted.
- **Force-kill timer leak** — `forceKillTimer` tracked and cleared in both the normal-exit and catch paths.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: task run output display — stdin close, chunk normalization, terminal rendering` (commit `52740d82`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
This incremental diff addresses the two previously raised issues. The mixed-chunk path now correctly returns only plain-text lines via `plainLines` instead of `cleanChunk` (which previously included raw JSON lines). The `forceKillTimer` leak on early process exit is fixed by clearing it in both the success and error paths. A safety comment documents the `escapeHtml`-first invariant. No new bugs or security issues are introduced. All prior concerns are resolved.

### Confidence: 8.4/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

PR #285, opened 2026-03-23
fix: normalize remember tags across daemon and openclaw

NicholaiVogel:
## summary
- accept `tags` as either a comma string or string array in the daemon remember route
- return a `400` for invalid tag payloads instead of bubbling a sqlite bind error
- normalize openclaw memory writes back to comma-separated strings before POSTing
- mirror the same tag parsing behavior in `daemon-rs`
- add coverage for daemon and openclaw remember tag handling

## root cause
`signet-memory-openclaw` was POSTing `tags` as `string[]` to `/api/memory/remember`, while the daemon write path still treated `tags` as a string field and passed it through to sqlite. that produced 500s on memory writes even though recall/search still worked.

## testing
- `bun test packages/daemon/src/mutation-api.test.ts --test-name-pattern 'POST /api/memory/remember'`
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `cargo test -p signet-daemon routes::write -- --nocapture`
- `bun run --filter '@signetai/signet-memory-openclaw' build`

### DIFF
--- a/packages/adapters/openclaw/src/index.test.ts
+++ b/packages/adapters/openclaw/src/index.test.ts
@@ -275,11 +275,11 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		expect(id).toBe("mem-1");
 		expect(lastRememberBody).toEqual({
 			content: "save this",
-			type: undefined,
-			importance: undefined,
 			tags: "alpha,beta",
 			who: "openclaw",
 		});
+		expect(lastRememberBody).not.toHaveProperty("type");
+		expect(lastRememberBody).not.toHaveProperty("importance");
 	});
 
 	it("deduplicates session-start for sessionless turns when both hooks fire", async () => {
--- a/packages/adapters/openclaw/src/index.test.ts
+++ b/packages/adapters/openclaw/src/index.test.ts
@@ -8,7 +8,9 @@ mock.module("@signet/core", () => ({
 }));
 
 // Import after mock so the module picks up the stub.
-const signetPlugin = (await import("./index")).default;
+const signet = await import("./index");
+const signetPlugin = signet.default;
+const { memoryStore } = signet;
 
 type HookHandler = (event: Record<string, unknown>, ctx: unknown) => Promise<unknown> | unknown;
 type ToolRegistration = { name: string; label?: string; description?: string };
@@ -25,6 +27,7 @@ let failSessionStartCount = 0;
 let failPromptSubmitCount = 0;
 let delaySessionStartMs = 0;
 let delayPromptSubmitMs = 0;
+let lastRememberBody: unknown = null;
 
 function hit(path: string): void {
 	pathCounts.set(path, (pathCounts.get(path) ?? 0) + 1);
@@ -120,9 +123,10 @@ beforeEach(() => {
 	failPromptSubmitCount = 0;
 	delaySessionStartMs = 0;
 	delayPromptSubmitMs = 0;
+	lastRememberBody = null;
 
 	const mockFetch = Object.assign(
-		async (input: RequestInfo | URL): Promise<Response> => {
+		async (input: RequestInfo | URL, init?: RequestInit): Promise<Response> => {
 			const url = typeof input === "string" ? input : input instanceof URL ? input.toString() : input.url;
 			const path = new URL(url).pathname;
 			hit(path);
@@ -154,6 +158,9 @@ beforeEach(() => {
 					});
 				case "/api/hooks/session-end":
 					return jsonResponse({ memoriesSaved: 0 });
+				case "/api/memory/remember":
+					lastRememberBody = init?.body ? JSON.parse(String(init.body)) : null;
+					return jsonResponse({ id: "mem-1" });
 				case "/api/marketplace/mcp/tools":
 					return jsonResponse({
 						count: 2,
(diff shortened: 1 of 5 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: normalize remember tags across daemon and openclaw` (commit `c9dfe123`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -2615,6 +2615,11 @@ app.post("/api/memory/remember", async (c) => {
 	const raw = body.content?.trim();
 	if (!raw) return c.json({ error: "content is required" }, 400);
 	const scope = body.scope ?? null;
+	const hasBodyTags = Object.prototype.hasOwnProperty.call(body, "tags");
```
NicholaiVogel, line 2620: **Unverified: `parseTagsMutation` array handling.** The diff adds `body.tags?: unknown` and routes arrays through `parseTagsMutation`, and the new test directly POSTs `tags: ["alpha", "beta"]` expecting `"alpha,beta"` back. But `parseTagsMutation` is never shown being changed in this diff. If it was previously a `(s: string) => string` helper, calling it with an array either silently returns the wrong thing (e.g. `"alpha,beta"` via Array.toString()) or throws. Please confirm `parseTagsMutation` was already able to handle `string | string[] | unknown`, or show the update. The PR description's stated fix is that openclaw now sends comma-strings, which would make the openclaw→daemon path safe — but the JS daemon's own array acceptance is only valid if `parseTagsMutation` actually handles arrays.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -52,14 +53,39 @@ pub struct RememberBody {
...
+        Value::String(tags) => Ok(tags
+            .split(',')
+            .map(str::trim)
+            .filter(|tag| !tag.is_empty())
+            .map(str::to_string)
+            .collect()),
+        Value::Array(tags) => Ok(tags
+            .into_iter()
+            .filter_map(|tag| match tag {
+                Value::String(tag) => Some(tag.trim().to_string()),
```
NicholaiVogel: **Silent drop of non-string array elements.** `filter_map` on the array branch silently discards non-string values (e.g. `["alpha", 42, "beta"]` → `["alpha", "beta"]`). This is a deliberate leniency choice, but it means a malformed client payload gets a 200 with quietly missing tags. Since the same mixed-array input returns 400 in the JS daemon (via `parseTagsMutation` presumably rejecting it), the two backends now behave differently on the same bad input. Consider returning a 400 for non-string array elements to keep behaviour consistent and to surface client bugs early.

NicholaiVogel (commented):
The fix is logically correct and addresses the root cause (array tags causing SQLite bind errors). Three issues worth noting: (1) `parseTagsMutation` in daemon.ts is never shown being updated to handle arrays, yet the new test exercises that exact path — if it only handled strings before, the test would fail or corrupt data silently; this is the most important gap the diff doesn't resolve. (2) The Rust `parse_remember_tags` silently drops non-string array elements rather than rejecting mixed arrays, which could hide client bugs. (3) The openclaw test asserts `type: undefined` and `importance: undefined` on the parsed JSON body, but JSON.stringify omits undefined-valued keys entirely, so those fields will be absent — the assertion passes only because Bun/Jest's toEqual ignores undefined properties, making the expectation semantically misleading.

### Confidence: 7.9/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 6,
  "pattern_correctness": 7,
  "documentation_coverage": 7
}
```

packages/adapters/openclaw/src/index.test.ts
```diff
@@ -259,6 +266,22 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
...
+		const id = await memoryStore("save this", {
+			daemonUrl: "http://daemon.test",
+			tags: ["alpha", " beta ", ""],
+		});
+
+		expect(id).toBe("mem-1");
+		expect(lastRememberBody).toEqual({
+			content: "save this",
+			type: undefined,
+			importance: undefined,
```
NicholaiVogel: **Misleading `undefined` assertions on parsed JSON body.** `expect(lastRememberBody).toEqual({ ..., type: undefined, importance: undefined, ... })` — `lastRememberBody` is the result of `JSON.parse(...)`, which will never have `type` or `importance` keys at all (JSON.stringify omits undefined-valued properties). The assertion passes only because Bun/Jest's `toEqual` treats a missing key as equivalent to `undefined`. A more precise assertion would be `expect(lastRememberBody).not.toHaveProperty('type')` or simply omit those fields from the expected object, so the intent is clear and the test would fail if the implementation accidentally started sending `type: null` instead.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `fix: normalize remember tags across daemon and openclaw` (commit `38371b73`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (commented):
All three issues from the prior review are now resolved. (1) `parseTagsMutation` in daemon.ts now returns `undefined` for mixed-type arrays, correctly propagating a 400 via the `hasBodyTags && bodyTags === undefined` guard. (2) The Rust `parse_remember_tags` now rejects mixed arrays with an explicit `Err(...)` instead of silently dropping non-string elements, and a covering test was added. (3) The openclaw test assertion was corrected to use `not.toHaveProperty` instead of embedding `undefined` values inside `toEqual`. No new issues introduced.

### Confidence: 8.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

PR #286, opened 2026-03-23
fix(daemon): align embedding-tracker hash with normalizeAndHashContent

aaf2tbz:
## Summary

- **Root cause**: `embedding-tracker.ts` computed content hashes via raw `sha256(content)`, but `memories.content_hash` is set using `normalizeAndHashContent` (lowercased, whitespace-collapsed, trailing punctuation stripped). These never match for any content with case/whitespace/punctuation differences.
- **Effect**: The tracker detected 20+ memories as permanently stale, ran every 5s (batch 8), called `syncVecDeleteBySourceExceptHash` → deleted vec_embeddings entries, then failed to re-add them because the new hash still didn't match `memories.content_hash`. Result: `resync-vec` appeared to complete but reverted on the next tracker cycle.
- **Fix**: Use `normalizeAndHashContent(row.content).contentHash` — consistent with the pipeline worker, transactions, and every other write path that sets `content_hash` on embeddings.

## Changes

- `packages/daemon/src/embedding-tracker.ts`: swap raw `createHash("sha256")` for `normalizeAndHashContent`, remove now-unused `createHash` import

## Test plan

- [ ] Deploy to daemon, run `resync-vec`, verify the vec count stabilizes (no longer resets every 5s)
- [ ] Confirm `embeddingCoverage` in diagnostics reaches 100% over time as the tracker re-embeds the stale memories with the correct hash

### DIFF
--- a/packages/daemon/src/embedding-tracker.ts
+++ b/packages/daemon/src/embedding-tracker.ts
@@ -9,7 +9,6 @@
 import { randomUUID } from "node:crypto";
 import type { PipelineEmbeddingTrackerConfig } from "@signet/core";
 import type { DbAccessor } from "./db-accessor";
-import { normalizeAndHashContent } from "./content-normalization";
 import { syncVecDeleteBySourceExceptHash, syncVecInsert, vectorToBlob } from "./db-helpers";
 import { logger } from "./logger";
 import type { EmbeddingConfig } from "./memory-config";
@@ -114,8 +113,7 @@ export function startEmbeddingTracker(
 				if (!running) break;
 				const vec = await fetchEmbeddingFn(row.content, embeddingCfg);
 				if (vec !== null) {
-					const { contentHash: hash } = normalizeAndHashContent(row.content);
-					results.push({ row, vector: vec, contentHash: hash });
+					results.push({ row, vector: vec, contentHash: row.contentHash });
 				} else {
 					failed++;
 				}
--- a/packages/daemon/src/embedding-tracker.ts
+++ b/packages/daemon/src/embedding-tracker.ts
@@ -7,9 +7,9 @@
  */
 
 import { randomUUID } from "node:crypto";
-import { createHash } from "node:crypto";
 import type { PipelineEmbeddingTrackerConfig } from "@signet/core";
 import type { DbAccessor } from "./db-accessor";
+import { normalizeAndHashContent } from "./content-normalization";
 import { syncVecDeleteBySourceExceptHash, syncVecInsert, vectorToBlob } from "./db-helpers";
 import { logger } from "./logger";
 import type { EmbeddingConfig } from "./memory-config";
@@ -114,7 +114,7 @@ export function startEmbeddingTracker(
 				if (!running) break;
 				const vec = await fetchEmbeddingFn(row.content, embeddingCfg);
 				if (vec !== null) {
-					const hash = createHash("sha256").update(row.content).digest("hex");
+					const { contentHash: hash } = normalizeAndHashContent(row.content);
 					results.push({ row, vector: vec, contentHash: hash });
 				} else {
 					failed++;


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): align embedding-tracker hash with normalizeAndHashContent` (commit `986bfa5d`) now and I'll follow up shortly with feedback.

packages/daemon/src/embedding-tracker.ts
```diff
@@ -114,7 +114,7 @@ export function startEmbeddingTracker(
 				if (!running) break;
 				const vec = await fetchEmbeddingFn(row.content, embeddingCfg);
 				if (vec !== null) {
-					const hash = createHash("sha256").update(row.content).digest("hex");
+					const { contentHash: hash } = normalizeAndHashContent(row.content);
```
NicholaiVogel: Correctness note (not blocking): `normalizeAndHashContent(row.content).contentHash` is now used as the write-path hash, which matches `memories.content_hash` by construction. However, `row.contentHash` (the value read from `m.content_hash` in the SELECT) is available directly on the row and is already the normalized hash. Using it instead of recomputing — `const hash = row.contentHash` — would be marginally cheaper and makes the intent explicit: 'write the hash the memory row already carries'. Either way is correct after this fix, but relying on `row.contentHash` avoids a redundant re-normalization of potentially large content strings on every batch item.

NicholaiVogel (dismissed):
The fix is correct and well-targeted. The root cause (raw SHA-256 vs. normalized hash) is accurately identified and the one-line swap is the right solution. No security issues introduced. One minor correctness concern: the `contentHash` computed from `normalizeAndHashContent(row.content)` is used to upsert into `embeddings`, but the stale-detection query compares `e.content_hash <> m.content_hash` where `m.content_hash` was set by the pipeline using `normalizeAndHashContent`. After this fix the hashes will match, which is exactly the intent — the cycle will converge. No issues found beyond confirming the fix is complete.

### Confidence: 8.5/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

aaf2tbz:
## Feedback addressed

**Re: use `row.contentHash` directly** — agreed, and done. Swapped `normalizeAndHashContent(row.content).contentHash` for `row.contentHash` (commit `3c197824`). Removes the redundant re-normalization and makes the intent explicit — the tracker now writes back the exact hash the memory row already carries.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix(daemon): align embedding-tracker hash with normalizeAndHashContent` (commit `3c197824`); taking another look at the fixes and I will report back shortly.

NicholaiVogel (approved):
The incremental change since the dismissed review is a further simplification: instead of calling `normalizeAndHashContent(row.content)` (introduced in the intermediate commit), the tracker now reads `row.contentHash` directly from the SQL result — which aliases `m.content_hash`, the authoritative, already-normalized hash stored by every write path. This is strictly better than re-normalizing: (1) it eliminates a redundant CPU operation, (2) it uses the DB's ground truth instead of a recomputed value that could skew if normalization logic ever diverges, and (3) it removes the now-unused `normalizeAndHashContent` import. The stale-detection query already selects rows where `e.content_hash <> m.content_hash`, so using `m.content_hash` as `contentHash` in the write leg is the only value that can cause the cycle to converge. No bugs, no security issues, no correctness concerns introduced by this change. Prior review concern (hash mismatch causing permanent staleness) is fully resolved.

### Confidence: 8.8/10


```pr-review-confidence-json
{
  "style_maintainability": 9,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

NicholaiVogel (approved):
solid catch, avery. this was a sneaky one — the hash mismatch meant the tracker was stuck in an infinite churn loop, deleting and re-inserting vec entries every cycle. using row.contentHash directly is the right call since the SELECT already pulls m.content_hash. clean fix, no wasted work. nice job :)

PR #287, opened 2026-03-23
fix(connector-openclaw): disable legacy hook when plugin path is active

aaf2tbz:
## Summary

- `install()` was enabling both the legacy `signet-memory` hook AND the plugin slot simultaneously on every default `signet setup --harness openclaw` call
- This creates the dual-system condition: duplicate memories per session, 2x token burn from double context injection, session-tracker 409 conflicts on every prompt turn
- Fix: when `runtimePath === "plugin"`, the `configureHooks` block now writes `enabled: false` for the legacy hook entry (instead of `true`), disabling any pre-existing legacy hook on each idempotent re-run

**Root cause** (`connector-openclaw/src/index.ts:356`): `configureHooks` defaulted to `true` with no awareness of `runtimePath`, so both paths always fired.

**Every new Signet install on OpenClaw silently created this condition.**

## Test plan

- [ ] Run `signet setup --harness openclaw` on a clean config; verify `openclaw.json` has `hooks.internal.entries["signet-memory"].enabled === false` and plugin slot enabled
- [ ] Re-run setup on existing install; verify idempotent — legacy hook stays disabled
- [ ] Pass `{ runtimePath: "legacy" }` explicitly; verify legacy hook remains `enabled: true`
- [ ] Confirm no duplicate memory extractions per session after install

### DIFF
--- a/packages/connector-openclaw/src/index.ts
+++ b/packages/connector-openclaw/src/index.ts
@@ -359,7 +359,10 @@ export class OpenClawConnector extends BaseConnector {
 					internal: {
 						entries: {
 							"signet-memory": {
-								enabled: true,
+								// Disable the legacy hook when using the plugin path to
+								// prevent dual-system operation (duplicate memories, 2x
+								// token burn, session-tracker 409 conflicts).
+								enabled: runtimePath !== "plugin",
 							},
 						},
 					},


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(connector-openclaw): disable legacy hook when plugin path is active` (commit `f6a6dbfb`) now and I'll follow up shortly with feedback.

packages/connector-openclaw/src/index.ts
```diff
@@ -359,7 +359,10 @@ export class OpenClawConnector extends BaseConnector {
 					internal: {
 						entries: {
 							"signet-memory": {
-								enabled: true,
+								// Disable the legacy hook when using the plugin path to
```
NicholaiVogel, line 362: Correct fix. `runtimePath !== "plugin"` evaluates to `false` (disabling the legacy hook) when the plugin path is active, and `true` (keeping it enabled) for the legacy path. The TypeScript union type `"plugin" | "legacy"` means no third value can reach this expression, so the logic is exhaustive.

NicholaiVogel (approved):
The fix is correct and minimal. Changing `enabled: true` to `enabled: runtimePath !== "plugin"` properly prevents dual-system operation when the plugin path is active. The `runtimePath` type is constrained to `"plugin" | "legacy"`, so no unexpected values can reach this expression. One low-severity gap: when switching back from plugin to legacy, the plugin entries (`signet-memory-openclaw`) are not disabled — callers would need to run `uninstall()` first. That's out of scope for this PR but worth a follow-up. No security issues introduced.

### Confidence: 7.8/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

- packages/connector-openclaw/src/index.ts:395 - Symmetry gap (out of scope but worth tracking): when `runtimePath === "legacy"` this correctly enables the legacy hook, but it does NOT disable any existing plugin entry (`signet-memory-openclaw`). A user switching back from plugin to legacy via `install({ runtimePath: "legacy" })` will end up with the plugin still enabled and the legacy hook also enabled — recreating the dual-system condition in reverse. The safe path is to call `uninstall()` before switching directions, but nothing enforces or warns about this today.

aaf2tbz:
Superseded by #291 (consolidated all 5 critical fixes into one PR).

PR #290, opened 2026-03-23
fix(adapter-openclaw): guard runPromptInjection against known-unreachable daemon

aaf2tbz:
## Summary

- `daemonReachable` was tracked by a startup health check and 60-second periodic timer, but **never consulted** before making actual requests
- When the Signet daemon is down, every OpenClaw message turn executed `daemonFetch()` with a 5-second `AbortSignal` timeout, hitting ECONNREFUSED and silently blocking for ~5 seconds before every agent response
- Users blamed OpenClaw for slowness; the real cause was Signet's dead connection attempt
- Fix: add `if (!daemonReachable) return undefined` as the first check in `runPromptInjection()`, bypassing the fetch entirely when daemon is known-down

**File**: `packages/adapters/openclaw/src/index.ts`

The 60-second health timer continues to run and will restore `daemonReachable = true` once the daemon comes back up, at which point injection resumes automatically.

## Test plan

- [ ] Stop the Signet daemon; send a message in OpenClaw — verify no ~5s hang
- [ ] Restart the daemon; verify injection resumes on the next message (within ~60s health check window)
- [ ] With daemon running: verify normal memory injection still works
- [ ] Verify `daemonFetch` is not called when `daemonReachable === false`

### DIFF
--- a/packages/adapters/openclaw/src/index.ts
+++ b/packages/adapters/openclaw/src/index.ts
@@ -1358,6 +1358,10 @@ const signetPlugin = {
 			sessionKey: string | undefined,
 			agentId: string | undefined,
 		): Promise<unknown> => {
+			// Skip immediately if daemon is known-unreachable — avoids a 5-second
+			// ECONNREFUSED hang on every message turn when the daemon is down.
+			if (!daemonReachable) return undefined;
+
 			const rawPrompt = typeof event.prompt === "string" ? event.prompt : undefined;
 			const prompt = rawPrompt ? extractUserMessage(rawPrompt) : undefined;
 			if (!prompt || prompt.length <= 3) {


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(adapter-openclaw): guard runPromptInjection against known-unreachable daemon` (commit `756e4d0b`) now and I'll follow up shortly with feedback.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1358,6 +1358,10 @@ const signetPlugin = {
 			sessionKey: string | undefined,
 			agentId: string | undefined,
 		): Promise<unknown> => {
+			// Skip immediately if daemon is known-unreachable — avoids a 5-second
```
NicholaiVogel, line 1361: This guard is correct and fixes the per-message hang for the common case. However, `runPromptInjection` is called *after* `ensureSessionStarted` in both `before_prompt_build` and `before_agent_start` hooks (see the `await ensureSessionStarted(...)` call that precedes `return runPromptInjection(...)`). When the daemon is down and `staticFallback()` returns `null` (e.g. no `~/.agents/` identity files), `onSessionStart` returns `null`, so the `sessionKey` is never added to `claimedSessions`. Every subsequent message turn re-enters `ensureSessionStarted`, calls `onSessionStart` → `daemonFetch`, hits ECONNREFUSED, and waits the full 5-second `READ_TIMEOUT` — reproducing the original symptom.

Fix: add the same `daemonReachable` guard at the top of `ensureSessionStarted`:
```ts
const ensureSessionStarted = async (
  event: Record<string, unknown>,
  sessionKey: string | undefined,
  agentId: string | undefined,
): Promise<void> => {
  if (!daemonReachable) return;   // ← add this
  ...
};
```
This is safe because the 60-second health timer will restore `daemonReachable = true` once the daemon comes back, at which point a fresh `ensureSessionStarted` call will succeed and claim the session.

NicholaiVogel (changes requested):
The guard in `runPromptInjection` correctly eliminates the per-message 5-second ECONNREFUSED hang. However, the fix is incomplete: `ensureSessionStarted` — called unconditionally before `runPromptInjection` in both lifecycle hooks — still invokes `onSessionStart` → `daemonFetch` with the 5-second timeout. When static identity is unavailable (`staticFallback()` returns null), the session is never added to `claimedSessions`, so `ensureSessionStarted` retries the full daemon round-trip on every subsequent message turn — reproducing the original hang despite this PR.

### Confidence: 7.6/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 8,
  "duplication_awareness": 8,
  "functional_completeness": 6,
  "pattern_correctness": 7,
  "documentation_coverage": 7
}
```

- packages/adapters/openclaw/src/index.ts:1295 - Minor follow-up hardening (not a blocker): `daemonReachable` starts as `true`, so the very first `runPromptInjection` call on startup races against the fire-and-forget `isDaemonRunning` promise. If the daemon is already down before the plugin registers, the first message turn after registration still incurs a 5-second timeout. Initialising `daemonReachable = false` and letting the startup check set it to `true` (or `false`) would close this window, though it adds a brief delay before the first injection on a healthy daemon. Document the chosen trade-off either way.

aaf2tbz:
Superseded by #291 (consolidated all 5 critical fixes into one PR).

PR #291, opened 2026-03-23
fix: 5 critical memory and injection stability fixes

aaf2tbz:
## Summary

Five critical bugs affecting memory search correctness, injection reliability, and dual-system conflicts — all isolated, each with a focused commit.

### Critical #1 — Dual-system install (connector-openclaw)
When `runtimePath === "plugin"`, `configureHooks` was writing `enabled: true` for the legacy hook, activating both paths simultaneously. Result: duplicate memories, 2× token burn, and 409 session-tracker conflicts. Fixed by setting `enabled: runtimePath !== "plugin"`.

### Critical #2 — Daemon-unreachable guard (adapter-openclaw)
`runPromptInjection` never checked `daemonReachable` before calling `daemonFetch`, causing a 5-second ECONNREFUSED hang on every message turn when the daemon is down. Fixed by early-returning `undefined` when `!daemonReachable`.

### Critical #3 — Hints FTS agentId hardcoded (daemon)
The prospective-hints FTS5 query passed `"default"` as the agentId for every agent, so hint recall was always scoped to the default agent regardless of who was asking. Fixed by extracting `params.agentId ?? "default"` and threading it into both the scoped and unscoped arg arrays.

### Critical #4 — Hint-only score capped at 30% (daemon)
The blending formula `hint * 0.3` applied even when there was no content score, capping hint-only memories at 30% relevance before `min_score` filtering. Those memories could never surface. Fixed by using `hint` directly for the hint-only branch.

### Critical #5 — Unreliable time-based prompt dedup (adapter-openclaw)
The 1-second `PROMPT_DEDUPE_WINDOW_MS` window was race-prone — under load, both `before_prompt_build` and `before_agent_start` could fire outside the window and both call the daemon. Replaced with a `Map<sessionKey, messageCount>` idempotency tracker: the first hook for each `(sessionKey, messageCount)` pair runs; subsequent ones return `undefined` immediately.

## Test plan
- [ ] Install with `runtimePath: "plugin"` — verify `signet-memory` hook has `enabled: false` in config
- [ ] Stop daemon → send a prompt → verify no 5-second hang
- [ ] Query memories for a non-default agentId — verify hint recall is scoped correctly
- [ ] Search for hint-only memories — verify they rank at full score, not capped at 30%
- [ ] Send a prompt that fires both hooks — verify daemon receives exactly one injection call per turn

### DIFF
--- a/packages/adapters/openclaw/src/index.ts
+++ b/packages/adapters/openclaw/src/index.ts
@@ -1358,6 +1358,10 @@ const signetPlugin = {
 			sessionKey: string | undefined,
 			agentId: string | undefined,
 		): Promise<unknown> => {
+			// Skip immediately if daemon is known-unreachable — avoids a 5-second
+			// ECONNREFUSED hang on every message turn when the daemon is down.
+			if (!daemonReachable) return undefined;
+
 			const rawPrompt = typeof event.prompt === "string" ? event.prompt : undefined;
 			const prompt = rawPrompt ? extractUserMessage(rawPrompt) : undefined;
 			if (!prompt || prompt.length <= 3) {
--- a/packages/adapters/openclaw/src/index.ts
+++ b/packages/adapters/openclaw/src/index.ts
@@ -251,7 +251,6 @@ interface MarketplaceExposurePolicy {
 	readonly updatedAt: string;
 }
 
-const PROMPT_DEDUPE_WINDOW_MS = 1_000;
 
 function sanitizeToolSegment(value: string): string {
 	const normalized = value
@@ -740,22 +739,16 @@ function textResult(text: string, details?: Record<string, unknown>): OpenClawTo
 	};
 }
 
-function cleanupRecentPromptTurns(recentTurns: Map<string, number>, now: number): void {
-	for (const [key, ts] of recentTurns) {
-		if (now - ts > PROMPT_DEDUPE_WINDOW_MS) {
-			recentTurns.delete(key);
+// Dedup window for sessionless session-start calls (time-based; these don't
+// have a stable messageCount to key on).
+const SESSIONLESS_DEDUPE_MS = 1_000;
+
+function cleanupTimedMap(map: Map<string, number>, now: number): void {
+	for (const [key, ts] of map) {
+		if (now - ts > SESSIONLESS_DEDUPE_MS) {
+			map.delete(key);
 		}
 	}
-}
-
-function buildPromptTurnKey(params: {
-	sessionKey?: string;
-	agentId?: string;
-	prompt: string;
-	messageCount?: number;
-}): string {
-	const normalizedPrompt = params.prompt.trim().replace(/\s+/g, " ").slice(0, 240);
-	return `${params.sessionKey ?? "-"}|${params.agentId ?? "-"}|${params.messageCount ?? -1}|${normalizedPrompt}`;
 }
 
 function buildInjectionResult(result: UserPromptSubmitResult): { prependContext: string } | undefined {
@@ -1296,7 +1289,8 @@ const signetPlugin = {
 
 		const claimedSessions = new Set<string>();
 		const sessionlessSessionStarts = new Map<string, number>();
-		const recentPromptTurns = new Map<string, number>();
+		// Maps sessionKey → last injected messageCount for per-turn idempotency.
+		const injectedTurns = new Map<string, number>();
 
 		const resolveHookContext = (
 			ctx: unknown,
@@ -1321,10 +1315,10 @@ const signetPlugin = {
 		): Promise<void> => {
 			if (!sessionKey) {
 				const now = Date.now();
-				cleanupRecentPromptTurns(sessionlessSessionStarts, now);
+				cleanupTimedMap(sessionlessSessionStarts, now);
 				const sessionlessKey = buildSessionlessTurnKey(event, agentId);
 				const recentStartAt = sessionlessSessionStarts.get(sessionlessKey);
-				if (typeof recentStartAt === "number" && now - recentStartAt <= PROMPT_DEDUPE_WINDOW_MS) {
+				if (typeof recentStartAt === "number" && now - recentStartAt <= SESSIONLESS_DEDUPE_MS) {
 					return;
 				}
 
@@ -1368,17 +1362,13 @@ const signetPlugin = {
 				return undefined;
 			}
 
-			const now = Date.now();
-			cleanupRecentPromptTurns(recentPromptTurns, now);
-			const messageCount = Array.isArray(event.messages) ? event.messages.length : undefined;
-			const promptTurnKey = buildPromptTurnKey({
-				sessionKey,
-				agentId,
-				prompt,
-				messageCount,
-			});
-			const recentTs = recentPromptTurns.get(promptTurnKey);
-			if (typeof recentTs === "number" && now - recentTs <= PROMPT_DEDUPE_WINDOW_MS) {
+			// Deduplicate by (sessionKey, messageCount): both before_prompt_build
+			// and before_agent_start fire on the same turn; only the first should
+			// call the daemon. When messageCount is unavailable we fall through
+			// rather than risk suppressing a valid injection.
+			const count = Array.isArray(event.messages) ? event.messages.length : undefined;
+			const key = sessionKey ?? "";
+			if (typeof count === "number" && injectedTurns.get(key) === count) {
 				return undefined;
 			}
 
@@ -1394,7 +1384,9 @@ const signetPlugin = {
 				// daemonFetch already logged the specific error (ECONNREFUSED or HTTP status).
 				return undefined;
 			}
-			recentPromptTurns.set(promptTurnKey, Date.now());
+			if (typeof count === "number") {
+				injectedTurns.set(key, count);
+			}
 			return buildInjectionResult(result);
 		};
 
--- a/packages/adapters/openclaw/src/index.ts
+++ b/packages/adapters/openclaw/src/index.ts
@@ -1289,8 +1289,11 @@ const signetPlugin = {
 
 		const claimedSessions = new Set<string>();
 		const sessionlessSessionStarts = new Map<string, number>();
-		// Maps sessionKey → last injected messageCount for per-turn idempotency.
-		const injectedTurns = new Map<string, number>();
+		// Maps sessionKey → {count, at} for per-turn idempotency. Entries are
+		// evicted on agent_end or lazily after SESSION_TURN_TTL_MS so crash/
+		// SIGKILL sessions don't accumulate indefinitely.
+		const SESSION_TURN_TTL_MS = 4 * 60 * 60 * 1000;
+		const injectedTurns = new Map<string, { count: number; at: number }>();
 		// Tracks turn signatures currently in-flight — provides a synchronous
 		// guard so concurrent before_prompt_build / before_agent_start calls on
 		// the same event-loop tick don't both pass the guard before either await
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: 5 critical memory and injection stability fixes` (commit `b3cd7e84`) now and I'll follow up shortly with feedback.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1390,7 +1384,9 @@ const signetPlugin = {
 				// daemonFetch already logged the specific error (ECONNREFUSED or HTTP status).
```
NicholaiVogel, line 1408: **Bug — memory leak**: `injectedTurns` is never pruned. The old `recentPromptTurns` map was cleaned up by `cleanupRecentPromptTurns` on every call. The new map accumulates one entry per unique `sessionKey` forever. In a long-running process with many short-lived sessions (or lots of sessionless calls that all write to `""`), this grows without bound. Add a cleanup pass similar to `cleanupTimedMap`, or cap map size to a reasonable maximum (e.g. 1000 entries with LRU eviction).

packages/adapters/openclaw/src/index.ts
```diff
@@ -1358,23 +1352,23 @@ const signetPlugin = {
...
-			});
-			const recentTs = recentPromptTurns.get(promptTurnKey);
-			if (typeof recentTs === "number" && now - recentTs <= PROMPT_DEDUPE_WINDOW_MS) {
+			// Deduplicate by (sessionKey, messageCount): both before_prompt_build
+			// and before_agent_start fire on the same turn; only the first should
+			// call the daemon. When messageCount is unavailable we fall through
+			// rather than risk suppressing a valid injection.
+			const count = Array.isArray(event.messages) ? event.messages.length : undefined;
+			const key = sessionKey ?? "";
+			if (typeof count === "number" && injectedTurns.get(key) === count) {
```
NicholaiVogel: **Bug — sessionless key collision**: `const key = sessionKey ?? ""` collapses every call without a sessionKey into the same map slot. Two concurrent sessionless agents that happen to reach the same `messageCount` (e.g. both at turn 1 where `event.messages.length === 2`) will cross-suppress each other: whichever session commits first sets `injectedTurns.set("", 2)`, and the second session's injection is silently dropped. The original `buildPromptTurnKey` included the normalised prompt text to avoid this. Consider keying on `agentId` as well: `const key = [sessionKey ?? "", agentId ?? ""].join("|")`, which matches the granularity already used in `buildSessionlessTurnKey`.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1358,23 +1352,23 @@ const signetPlugin = {
...
 		): Promise<unknown> => {
+			// Skip immediately if daemon is known-unreachable — avoids a 5-second
+			// ECONNREFUSED hang on every message turn when the daemon is down.
+			if (!daemonReachable) return undefined;
+
 			const rawPrompt = typeof event.prompt === "string" ? event.prompt : undefined;
 			const prompt = rawPrompt ? extractUserMessage(rawPrompt) : undefined;
 			if (!prompt || prompt.length <= 3) {
 				return undefined;
 			}
```
NicholaiVogel, line 1372: **Residual race condition**: `injectedTurns` is written only *after* `await onUserPromptSubmit` returns (line ~1384). If `before_prompt_build` and `before_agent_start` are dispatched concurrently (both enter `runPromptInjection` before either completes), both will read `injectedTurns.get(key) !== count` simultaneously, both will pass the guard, and both will call the daemon. The fix only prevents *strictly sequential* hook re-entrancy. To close the race: set the entry *before* the await (optimistically), then clear it on error:
```ts
if (typeof count === "number") injectedTurns.set(key, count); // optimistic lock
const result = await onUserPromptSubmit(...);
if (!result) {
  injectedTurns.delete(key); // roll back so the next real turn can inject
  return undefined;
}
```
This doesn't require any structural change to the hook dispatch model and is safe under the single-threaded JS event loop.

packages/daemon/src/memory-search.ts
```diff
@@ -290,9 +290,10 @@ export async function hybridRecall(
 					   WHERE memory_hints_fts MATCH ? AND h.agent_id = ?
 					   ORDER BY raw_score LIMIT ?`;
 
-				const args = scoped
-					? [keywordQuery, "default", params.scope, cfg.search.top_k]
-					: [keywordQuery, "default", cfg.search.top_k];
+				const agentId = params.agentId ?? "default";
```
NicholaiVogel, line 293: **Indentation regression**: `const args` is indented one level less than its surrounding block (uses 3 tabs/levels instead of 4). Functionally harmless, but inconsistent with the rest of the file and will produce noisy diffs in follow-up PRs. Should align with the `const agentId` line above it.

NicholaiVogel (changes requested):
Four of five fixes are correct and well-motivated. Critical #5 (dedup rewrite) introduces a memory leak (`injectedTurns` has no eviction path) and a key-collision bug (all sessionless agents share the same `""` key, so two concurrent sessionless agents at the same messageCount will cross-suppress each other). A latent race condition also remains: `injectedTurns` is only written *after* the daemon responds, so two hooks that both clear the synchronous guard and then both `await onUserPromptSubmit` in the same event-loop tick will both call the daemon. Additionally, the `memory-search.ts` hunk has a visible indentation regression on the `const args` line.

### Confidence: 7.9/10


```pr-review-confidence-json
{
  "style_maintainability": 7,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 8
}
```

aaf2tbz:
@NicholaiVogel All four issues from the review have been addressed in bc9f0735:

1. **Memory leak** (`injectedTurns` no eviction path) — added `injectedTurns.delete(sessionKey)` in the `agent_end` handler alongside the existing `claimedSessions.delete`.

2. **Key collision** (sessionless agents sharing `""` key) — replaced `key = sessionKey ?? ""` with `sig = sessionKey && typeof count === "number" ? \`${sessionKey}|${count}\` : undefined`. Sessionless agents now skip dedup entirely (fall through) rather than sharing a key.

3. **Race condition** (both hooks pass guard before either awaits) — added `inFlightTurns = new Set<string>()` with a synchronous `inFlightTurns.add(sig)` *before* the `await`, so a concurrent same-tick call sees the guard immediately. Cleared unconditionally after the await.

4. **Indentation regression** in `memory-search.ts` `const args` line — corrected to 4-tab indent matching the surrounding block.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: 5 critical memory and injection stability fixes` (commit `bc9f0735`); taking another look at the fixes and I will report back shortly.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1424,6 +1435,7 @@ const signetPlugin = {
 			await onSessionEnd("openclaw", { ...opts, sessionKey });
```
NicholaiVogel, line 1445: Previously raised, now partially resolved: `injectedTurns.delete(sessionKey)` is correctly called on `agent_end`, covering the normal termination path. The residual is sessions that exit without firing `agent_end` (SIGKILL, process crash) — those entries accumulate indefinitely. Since each entry is a single integer per sessionKey the practical impact is negligible, but a future hardening pass could add a TTL map or a periodic sweep keyed on session inactivity to close the gap entirely.

NicholaiVogel (dismissed):
All four issues from the prior review are resolved. The key-collision bug (sessionless agents sharing the `""` key) is fixed by gating `sig` on a truthy `sessionKey`. The race condition is fixed by `inFlightTurns` providing a synchronous guard before the first `await`. The indentation regression in `memory-search.ts` is corrected. `injectedTurns.delete(sessionKey)` is now called on `agent_end`, eliminating the cleanup gap for normal session termination. One minor residual: `injectedTurns` still accumulates for sessions that terminate without firing `agent_end` (e.g. process crash, SIGKILL). Since the map holds at most one number per `sessionKey`, this is a bounded, low-severity leak — acceptable given the fix scope and the rarity of ungraceful terminations in this environment. No new bugs, security issues, or logic errors introduced.

### Confidence: 8.4/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 8,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

aaf2tbz:
Addressed the residual noted in the last review (6330e372): changed `injectedTurns` to store `{count, at}` instead of a bare number. A lazy TTL sweep (4h, matching session stale timeout) now runs before each dedup check, so entries from sessions that ended via crash or SIGKILL evict automatically rather than accumulating indefinitely.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: 5 critical memory and injection stability fixes` (commit `6330e372`); taking another look at the fixes and I will report back shortly.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1296,7 +1289,16 @@ const signetPlugin = {
 
 		const claimedSessions = new Set<string>();
 		const sessionlessSessionStarts = new Map<string, number>();
-		const recentPromptTurns = new Map<string, number>();
+		// Maps sessionKey → {count, at} for per-turn idempotency. Entries are
```
NicholaiVogel, line 1292: Minor: `SESSION_TURN_TTL_MS` is declared inside `register()` but behaves as a pure constant. Moving it to module scope alongside `SESSIONLESS_DEDUPE_MS` would be more consistent and avoids re-allocating on each `register()` call. Not a correctness issue.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1358,25 +1360,38 @@ const signetPlugin = {
...
+			// Deduplicate by (sessionKey, messageCount): both before_prompt_build
+			// and before_agent_start fire on the same turn; only the first should
+			// call the daemon. Sessionless agents (no sessionKey) cannot be
+			// reliably correlated and are allowed to fall through rather than
+			// risk cross-suppressing concurrent independent sessions.
+			const count = Array.isArray(event.messages) ? event.messages.length : undefined;
+			// sig is only defined when we have both a stable session identity and
+			// a message count — the two values that make dedup meaningful.
+			const sig = sessionKey && typeof count === "number" ? `${sessionKey}|${count}` : undefined;
+			// Lazy TTL sweep: evict entries from sessions that ended without agent_end.
```
NicholaiVogel, line 1382: The O(n) sweep over `injectedTurns` fires on every `runPromptInjection` call that has a valid `sig`. In practice the map stays small (one live entry per active session), so this is fine. Worth noting for future scaling: if session counts grow large, a min-heap or periodic background sweep would be more efficient.

NicholaiVogel (approved):
The incremental change correctly addresses the sole residual issue from the prior review: `injectedTurns` entries for sessions that terminate without `agent_end` now expire via a lazy TTL sweep (4-hour window). The struct change from `Map<string, number>` to `Map<string, { count: number; at: number }>` is minimal and correct. The dedup guard is updated to use `.count === count` via optional chaining. No new bugs, race conditions, security issues, or logic errors introduced. Prior review items are fully resolved.

### Confidence: 8.7/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 10,
  "duplication_awareness": 9,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 8
}
```

NicholaiVogel (approved):
god fucking dammit we need to run the benchmarks again lol

the vector path was doing its job on the LoCoMo run (100% Hit@10), but fixes #3 and #4 here would absolutely suppress memories that should have surfaced in harder queries. if the benchmark questions happened to match content directly, the hint path never mattered, so the 87.5% score was masking these bugs.

for any question where the answer lives in a memory that doesn't share keywords with the query, hints were the only retrieval path. and those were either scoped to the wrong agent or capped at 30%. both invisible failures.

good catches across the board. approve and merge.

PR #293, opened 2026-03-23
fix: wire marketplace reviews sync to production Worker endpoint

aaf2tbz:
## Summary

Two small changes to `packages/daemon/src/routes/marketplace-reviews.ts`:

- **Pre-configure sync URL** — `DEFAULT_CONFIG.endpointUrl` now points to `https://reviews.signetai.sh/api/reviews/sync` (the Cloudflare Worker from `Signet-AI/marketplace#1`). New installs are pre-configured; users don't need to set the URL manually.
- **Add `X-Signet-Sync: 1` header** — the Worker requires this on all sync requests as a lightweight origin gate to filter non-Signet clients.

Sync remains **opt-in** (`enabled: false` default). Once the Worker is live, a follow-up can flip `enabled: true` as the default.

## Dependencies

Pair with **[Signet-AI/marketplace#1](https://github.com/Signet-AI/marketplace/pull/1)** — the Worker must be deployed by Nicholai before the sync URL resolves. A TODO comment in the code marks where to flip the default once live.

## Test plan

- [ ] `PATCH /api/marketplace/reviews/config` with just `{ "enabled": true }` uses the pre-configured URL (no need to set `endpointUrl`)
- [ ] `POST /api/marketplace/reviews/sync` outbound request includes `X-Signet-Sync: 1` header
- [ ] Existing tests pass: `bun test packages/daemon/src/routes/marketplace-reviews.test.ts`


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: wire marketplace reviews sync to production Worker endpoint` (commit `b472716f`) now and I'll follow up shortly with feedback.

packages/daemon/src/hooks.ts
```diff
@@ -1935,6 +1938,9 @@ export async function handleUserPromptSubmit(req: UserPromptSubmitRequest): Prom
 			const dateStr = formatMemoryDate(s.created_at);
```
NicholaiVogel, line 1935: **Logic bug: `omitted` counter doesn't account for the 5-item display cap.**

`omitted = recall.results.length - budgetFiltered.length` only counts items dropped by the character budget. If e.g. `recall.results` has 7 items all fitting within `contextBudgetChars`, `budgetFiltered` has 7, `budgetSelected = budgetFiltered.slice(0, 5)` shows 5, but `omitted = 0` so the hint is never shown. The 2 hidden items are invisible and raising `contextBudgetChars` won't help. The comment acknowledges this is intentional, but the hint message becomes incorrect: users will never be told about results hidden by the hard cap. Either remove the 5-item cap (make it configurable via guardrails), or compute `omitted` as `recall.results.length - budgetSelected.length` and adjust the hint text to say 'raise the display cap or contextBudgetChars'.

packages/daemon/src/daemon.ts
```diff
@@ -1949,6 +1949,35 @@ app.get("/api/memory/timeline", (c) => {
...
+					 LEFT JOIN memories m ON m.id = h.memory_id
+					 WHERE h.event IN ('DEDUP', 'REVIEW_NEEDED', 'BLOCKED_DESTRUCTIVE')
+					   AND h.created_at > datetime('now', '-30 days')
+					 ORDER BY h.created_at DESC
+					 LIMIT 200`,
+				)
+				.all();
+		});
+		return c.json({ items: rows });
+	} catch (e) {
```
NicholaiVogel: **Missing `total` field in API response.**

The endpoint returns `c.json({ items: rows })` but the frontend types this as `{ items: ReviewItem[]; total: number }`. The `total` field is silently `undefined` at runtime. It's unused today, but if any future code branches on it, it'll silently be `undefined` rather than a number. Add `total: rows.length` to the response.

packages/daemon/src/routes/marketplace-reviews.ts
```diff
@@ -344,7 +349,7 @@ export function mountMarketplaceReviewsRoutes(app: Hono): void {
 		try {
 			const response = await fetch(config.endpointUrl, {
 				method: "POST",
-				headers: { "Content-Type": "application/json" },
+				headers: { "Content-Type": "application/json", "X-Signet-Sync": "1" },
```
NicholaiVogel, line 350: **Future hardening: `X-Signet-Sync: 1` is trivially bypassable.**

The PR description correctly identifies this as a lightweight origin gate, not real auth. Any client can send this header. Once the Worker is in production, consider upgrading to HMAC-signed payloads or a pre-shared token (stored in Signet secrets) so the Worker can reject arbitrary third-party POSTs.

packages/daemon/src/daemon.ts
```diff
@@ -5546,10 +5576,15 @@ app.post("/api/hooks/user-prompt-submit", async (c) => {
 		const runtimePath = resolveRuntimePath(c, body);
 		if (runtimePath) body.runtimePath = runtimePath;
 
+		// Capture before any claim refresh — false means the daemon restarted
```
NicholaiVogel: **Scope note: `hasSession` is called before `checkSessionClaim`, evicting expired entries as a side effect.**

The code comment documents this is intentional. One edge case to be aware of: if a session expired naturally (not a daemon restart), `known` will be `false`, causing the adapter to evict its local claim and re-init via `onSessionStart` on the next turn. The daemon's `sessionStartSeen` dedup guard will return a minimal stub for re-inits, so there's no infinite loop or double identity injection. Just confirm this 'treat expiry == restart' behaviour is desired.

packages/daemon/src/daemon.ts
```diff
@@ -5421,6 +5450,7 @@ import {
 	getActiveSessions,
```
NicholaiVogel, line 5421: **Scope mismatch vs PR description.**

The PR description says 'Two small changes to `packages/daemon/src/routes/marketplace-reviews.ts`' but this diff touches 10+ files: `session-tracker.ts`, `hooks.ts`, `memory-config.ts`, `pipeline/significance-gate.ts`, `health.ts`, plus multiple dashboard files. The marketplace-reviews change is clean; the remaining changes are unrelated features that should ideally be in their own PRs for easier revert targeting and bisect safety.

aaf2tbz:
Rebuilt the branch cleanly from `main` — PR now contains only the intended `marketplace-reviews.ts` change (1 file, 9 lines). The extra files from the previous diff were unmerged work-in-progress commits that the branch had been cut on top of by mistake.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: wire marketplace reviews sync to production Worker endpoint` (commit `e8a94cee`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/routes/marketplace-reviews.ts
```diff
@@ -27,9 +27,14 @@ interface ReviewsSyncConfig {
 	readonly lastSyncError: string | null;
 }
 
+// Production sync endpoint — set by Nicholai after deploying the Cloudflare Worker.
```
NicholaiVogel: The comment says "replace placeholder with live Worker URL once deployed" but `REVIEWS_SYNC_URL` is already set to the production value. The TODO should instead read something like: `// TODO: flip enabled: true in DEFAULT_CONFIG once the Worker is confirmed live.` As written it implies the URL still needs to be filled in, which will confuse the next reader.

aaf2tbz:
Both points addressed in d26f68f4:

1. **Misleading TODO** — removed. Comment now reads: `Production sync endpoint. Pre-configured so users only need to set enabled: true.`
2. **readConfig() fallback** — `endpointUrl` now falls back to `REVIEWS_SYNC_URL` when the on-disk value is empty or missing, so existing installs pick up the production URL on next daemon start without a manual PATCH.

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: wire marketplace reviews sync to production Worker endpoint` (commit `d26f68f4`); taking another look at the fixes and I will report back shortly.

PR #295, opened 2026-03-23
fix: session expiry, hooks config, dead-memory API, openclaw health

aaf2tbz:
## What

Four quality-of-life improvements from the March 2026 critical review, shipped on `feat/qol-improvements-march-2026` off `main`.

### 1. Session expiry warnings
- `getExpiryWarning()` in `session-tracker.ts` — returns a human-readable warning 30 min before TTL, once per session (throttled via `warnedSessions`)
- `renewSession()` — resets expiry clock and clears warned state
- `SessionInfo.expiresAt` now included in active-sessions output
- `SessionStartResponse` and `UserPromptSubmitResponse` carry an optional `warnings[]` field that propagates expiry notices to the hook caller (including session-start dedup stub)
- Session evictions log `sessionKey`, `runtimePath`, and `claimedAt` for operator visibility

### 2. HooksConfig validation + `userPromptSubmit` config
- `loadHooksConfig()` warns on unknown top-level keys via `KNOWN_HOOKS_KEYS` — silently swallowed misconfiguration was a real bug surface
- `userPromptSubmit` section is now config-driven (`enabled`, `recallLimit`, `maxInjectChars`) with defaults in `getDefaultConfig()`
- `preCompaction.maxSummaryChars` added to schema
- Expiry warnings thread through all `handleUserPromptSubmit` return paths

### 3. Dead-memory review API
- `findDeadMemories()` in `repair-actions.ts` — queries by low confidence, never-accessed, or stale-last-accessed thresholds
- `forgetDeadMemories()` — soft-deletes a batch in a single write transaction (max 500/call)
- `GET /api/repair/dead-memories` — list candidates (query params: `maxConfidence`, `maxAccessDays`, `limit`); validates `Number.isFinite` + non-negative
- `POST /api/repair/dead-memories/forget` — soft-delete by ID list
- Maintenance worker warns when dead count > 100 and hints at these endpoints — no auto-deletion

### 4. OpenClaw plugin health tracking
- `OpenClawHealth` interface + `openclaw` field on `DiagnosticsReport`
- `getDiagnostics()` accepts optional `openclawHealth` param; defaults to `never-seen` zeros when absent (no breaking change)
- `POST /api/diagnostics/openclaw/heartbeat` — unauthenticated; accumulates hook call deltas, detects plugin version, last-error passthrough. String fields capped (128/512 chars), `hooksRegistered` runtime-validated as strings
- `GET /api/diagnostics/openclaw` — returns current health snapshot with stale detection (> 10 min since last heartbeat → status `"stale"`)

## Why

- **Session expiry**: Silent drops on TTL expiry were surprising operators; advance warning + renew endpoint closes the gap without requiring re-initialization
- **HooksConfig unknowns**: Misconfigured keys (typos, renamed fields) silently did nothing — now logged at warn level
- **Dead memories**: Low-confidence and stale memories accumulated indefinitely with no visibility or reclaim path; API enables deliberate review before deletion
- **OpenClaw health**: Plugin failures were invisible to diagnostics — heartbeat + stale detection surfaces them in the existing health dashboard

## Test plan

- [x] `bun run build` passes clean
- [ ] Start daemon, `GET /api/diagnostics` includes `openclaw` field
- [ ] `GET /api/repair/dead-memories` returns candidates; `POST /api/repair/dead-memories/forget` soft-deletes them
- [ ] Send `POST /api/diagnostics/openclaw/heartbeat` → `GET /api/diagnostics/openclaw` shows `connected`; wait 10 min → status becomes `stale`
- [ ] Set TTL session near expiry, call `POST /api/session-start` → response includes `warnings[]`
- [ ] Add unknown key to `hooks:` in `agent.yaml` → daemon logs warning on load
- [ ] `POST /api/sessions/:key/renew` resets expiry and clears warning


NicholaiVogel:
Hi @aaf2tbz - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: five QoL improvements — session expiry, hooks config, dead-memory API, openclaw health, predictor warm-start` (commit `4c20754a`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -7136,6 +7182,46 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	if (typeof b.pluginVersion !== "string") {
+		return c.json({ error: "pluginVersion (string) is required" }, 400);
+	}
+	const prev = openClawHeartbeat?.data;
+	openClawHeartbeat = {
+		timestamp: new Date().toISOString(),
+		data: {
+			pluginVersion: b.pluginVersion,
+			hooksRegistered: Array.isArray(b.hooksRegistered) ? (b.hooksRegistered as string[]) : [],
+			lastHookCall: typeof b.lastHookCall === "string" ? b.lastHookCall : null,
```
NicholaiVogel: **Bug: `lastError` is silently dropped — 'last-error passthrough' is not implemented.**

The heartbeat body may include a `lastError` field (the PR description says so), but `OpenClawHeartbeatData` has no `lastError` member, the heartbeat handler never reads `b.lastError`, and `buildOpenClawHealth()` always returns `lastError: null`. The field in `OpenClawHealth` will always be null regardless of what the plugin reports. Either add `lastError: string | null` to `OpenClawHeartbeatData`, store `typeof b.lastError === 'string' ? b.lastError : null` in the constructor, and surface it in `buildOpenClawHealth()`; or remove `lastError` from `OpenClawHealth` and the description until it is implemented.

packages/core/src/migrations/041-warm-start-flag.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Migration 041: Warm-Start Flag
+ *
+ * Adds is_warm_start column to memories so synthetic seed memories
+ * inserted during predictor warm-start can be excluded from recall
+ * results while still training the predictor.
+ */
+
+function hasColumn(db: MigrationDb, table: string, col: string): boolean {
+	const rows = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<{ name?: unknown }>;
```
NicholaiVogel: **Unquoted table name in `PRAGMA table_info(${table})`.**

Every other callsite in the codebase (e.g. `tableColumns` in `index.ts`) uses `PRAGMA table_info("${table}")` with double quotes to guard against identifiers that contain special characters or reserved words. This helper is only called with the literal `"memories"` today, so it cannot be exploited here, but the inconsistency is a trap for future callers. Change to `PRAGMA table_info("${table}")`.

packages/daemon/src/repair-actions.ts
```diff
@@ -1933,3 +1911,91 @@ export function backfillSkippedSessions(
...
+			   AND (
+			     confidence < ?
+			     OR (last_accessed IS NULL AND julianday('now') - julianday(created_at) > ?)
+			     OR (last_accessed IS NOT NULL AND julianday('now') - julianday(last_accessed) > ?)
+			   )
+			 ORDER BY confidence ASC, last_accessed ASC NULLS FIRST
+			 LIMIT ?`,
+		)
+		.all(maxConf, maxDays, maxDays, limit) as Array<{
+		id: string;
```
NicholaiVogel, line 1968: **`db.prepare()` called inside a loop — should be hoisted.**

In `forgetDeadMemories`, `db.prepare("UPDATE memories SET is_deleted = 1 ...")` is called on every iteration of the `for (const id of ids)` loop. SQLite compiles the query on each `prepare()` call. Hoist the prepare once before the loop:
```ts
const stmt = db.prepare("UPDATE memories SET is_deleted = 1, deleted_at = datetime('now') WHERE id = ? AND is_deleted = 0");
for (const id of ids) {
  total += countChanges(stmt.run(id));
}
```
This is the same pattern already used in `updateAccessTracking` in `hooks.ts`.

packages/daemon/src/daemon.ts
```diff
@@ -7136,6 +7182,46 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+
+// Unauthenticated — daemon is local-only, plugin may not carry auth token
+app.post("/api/diagnostics/openclaw/heartbeat", async (c) => {
+	let body: unknown;
+	try {
+		body = await c.req.json();
+	} catch {
+		return c.json({ error: "Invalid JSON" }, 400);
+	}
+	if (!body || typeof body !== "object") {
```
NicholaiVogel, line 7280: **Attack-surface note for future hardening: unauthenticated heartbeat accepts unbounded arrays.**

`hooksRegistered` is accepted as `Array.isArray(b.hooksRegistered) ? (b.hooksRegistered as string[])` with no length cap. A local process (or any process if the port is ever bound non-locally) could push a very large array, inflating the in-process `openClawHeartbeat` object and the diagnostics response payload. Consider capping at e.g. 50 entries.

NicholaiVogel (commented):
Three functional/correctness issues found. Most significant: the PR description promises 'last-error passthrough' via the OpenClaw heartbeat, but `lastError` is never stored in `OpenClawHeartbeatData` and is hardcoded `null` in every `OpenClawHealth` response — the field accepted in the heartbeat body is silently discarded. Secondary: `PRAGMA table_info(${table})` in the migration helper is unquoted (contrast with the quoted form used everywhere else in `index.ts`). Third: `forgetDeadMemories` prepares the same `UPDATE` statement inside a loop rather than once before the loop.

### Confidence: 7.7/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

aaf2tbz:
All three addressed in 396d3a42:

1. **`lastError` passthrough** — added `lastError: string | null` to `OpenClawHeartbeatData`, stored from `b.lastError` in the heartbeat endpoint body, and returned `d.lastError` from `buildOpenClawHealth()` (was hardcoded `null`)
2. **PRAGMA quoting** — `PRAGMA table_info("${table}")` now quoted, matching convention in `migrations/index.ts`
3. **Statement hoist** — `db.prepare(...)` moved before the loop in `forgetDeadMemories`; statement prepared once and reused per ID

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: session expiry, hooks config, dead-memory API, openclaw health, predictor warm-start` (commit `396d3a42`); taking another look at the fixes and I will report back shortly.

packages/daemon/src/repair-actions.ts
```diff
@@ -1933,3 +1911,89 @@ export function backfillSkippedSessions(
...
+	const maxDays = opts.maxAccessDays ?? 90;
+	const limit = opts.limit ?? 200;
+
+	const rows = db
+		.prepare(
+			`SELECT id, content, confidence, last_accessed, importance
+			 FROM memories
+			 WHERE is_deleted = 0
+			   AND importance <= 0.8
+			   AND (
```
NicholaiVogel, line 1959: **Previously reported — FIXED.** `forgetDeadMemories` now hoists `db.prepare(...)` outside the loop. ✅

packages/core/src/migrations/041-warm-start-flag.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Migration 041: Warm-Start Flag
+ *
+ * Adds is_warm_start column to memories so synthetic seed memories
+ * inserted during predictor warm-start can be excluded from recall
+ * results while still training the predictor.
+ */
+
+function hasColumn(db: MigrationDb, table: string, col: string): boolean {
+	const rows = db.prepare(`PRAGMA table_info("${table}")`).all() as ReadonlyArray<{ name?: unknown }>;
```
NicholaiVogel: **Previously reported — FIXED.** `PRAGMA table_info("${table}")` is now properly quoted. ✅

packages/daemon/src/daemon.ts
```diff
@@ -7136,6 +7183,47 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	}
+	const b = body as Record<string, unknown>;
+	if (typeof b.pluginVersion !== "string") {
+		return c.json({ error: "pluginVersion (string) is required" }, 400);
+	}
+	const prev = openClawHeartbeat?.data;
+	openClawHeartbeat = {
+		timestamp: new Date().toISOString(),
+		data: {
+			pluginVersion: b.pluginVersion,
```
NicholaiVogel: **Previously reported — FIXED.** `lastError` is now stored in `OpenClawHeartbeatData` and surfaced through `buildOpenClawHealth`. ✅

packages/daemon/src/repair-actions.ts
```diff
@@ -1933,3 +1911,89 @@ export function backfillSkippedSessions(
...
+}
+
+export interface DeadMemoryOpts {
+	/** Max confidence to flag as dead. Default: 0.10. */
+	readonly maxConfidence?: number;
+	/** Days since last access (or creation if never accessed) to flag as stale. Default: 90. */
+	readonly maxAccessDays?: number;
+	/** Max rows to return. Default: 200. */
+	readonly limit?: number;
+}
```
NicholaiVogel, line 1938: **New issue — correctness:** `findDeadMemories` does not exclude `is_warm_start = 1` rows. Warm-start memories have importance 0.3–0.6 (below the 0.8 guard) and access_count 0, so they will appear as `never_accessed` / `stale` candidates after 90 days. An operator running `POST /api/repair/dead-memories/forget` on those IDs will soft-delete the predictor's synthetic training pairs, silently regressing recall quality on the next cold start. Add `AND (is_warm_start IS NULL OR is_warm_start = 0)` to the WHERE clause.

```sql
WHERE is_deleted = 0
  AND importance <= 0.8
  AND (is_warm_start IS NULL OR is_warm_start = 0)
  AND (
    confidence < ?
    OR ...
  )
```

packages/daemon/src/pipeline/maintenance-worker.ts
```diff
@@ -380,6 +380,33 @@ export function startMaintenanceWorker(
...
+			const count = accessor.withReadDb(
+				(db) =>
+					(
+						db
+							.prepare(
+								`SELECT COUNT(*) as n FROM memories
+								 WHERE is_deleted = 0 AND importance <= 0.8
+								 AND (confidence < 0.1
+								   OR (last_accessed IS NULL AND julianday('now') - julianday(created_at) > 90)
+								   OR (last_accessed IS NOT NULL AND julianday('now') - julianday(last_accessed) > 90))`,
```
NicholaiVogel: **Related to above:** The maintenance-worker dead-memory count query (`importance <= 0.8 AND (confidence < 0.1 OR ...)`) also lacks an `is_warm_start` filter. Once warm-start memories age past 90 days the count will exceed 100 and emit a spurious warning on every maintenance cycle, creating log noise that trains operators to ignore the alert. Add `AND (is_warm_start IS NULL OR is_warm_start = 0)` here as well.

packages/daemon/src/daemon.ts
```diff
@@ -486,14 +503,33 @@ function invalidateDiagnosticsCache(): void {
...
+	}
+	const age = Date.now() - new Date(openClawHeartbeat.timestamp).getTime();
+	const status = age < OPENCLAW_STALE_MS ? "connected" : "stale";
+	const d = openClawHeartbeat.data;
+	return {
+		status,
+		lastHeartbeat: openClawHeartbeat.timestamp,
+		pluginVersion: d.pluginVersion,
+		hooksRegistered: d.hooksRegistered,
+		hooksSucceeded: d.totalSucceeded,
```
NicholaiVogel, line 522: **Minor semantic issue:** `avgLatencyMs` is set to `d.latencyMs` — the latency value from the single most recent heartbeat. The field name implies a rolling average. Since the interface exposes this to dashboard consumers, consider renaming to `lastLatencyMs`, or computing a real average by accumulating a sum and count in `OpenClawHeartbeatData`. Not a blocking issue but will mislead operators reading the health endpoint.

NicholaiVogel (commented):
All three issues from the prior review are resolved: PRAGMA quoting fixed, lastError now stored and surfaced, forgetDeadMemories hoists the prepared statement outside the loop. One new correctness issue remains: findDeadMemories and the maintenance-worker dead-memory count do not exclude is_warm_start memories. Warm-start rows have importance 0.3–0.6 and low/zero access counts, so they will surface as dead-memory candidates after 90 days and can be soft-deleted via POST /api/repair/dead-memories/forget — silently destroying the predictor's synthetic training pairs. The avgLatencyMs field is a minor semantic issue (last-reported latency, not a rolling average) but low risk.

### Confidence: 7.7/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Hi @aaf2tbz - quick follow-up pass on `fix: session expiry, hooks config, dead-memory API, openclaw health, predictor warm-start` (commit `8e8f0ab2`); taking another look at the fixes and I will report back shortly.

aaf2tbz:
Both issues addressed in 8e8f0ab2:

1. **Warm-start exclusion** — added `AND (is_warm_start IS NULL OR is_warm_start = 0)` to both the `findDeadMemories` query in `repair-actions.ts` and the maintenance-worker dead-memory count. Warm-start rows can no longer surface as deletion candidates regardless of their confidence/access-day values.

2. **`avgLatencyMs` → `lastLatencyMs`** — renamed in `OpenClawHealth` interface (`diagnostics.ts`) and `buildOpenClawHealth()` (`daemon.ts`). The field holds the latency value from the most recent heartbeat, not a computed average; the name now matches what it actually contains.

NicholaiVogel (approved):
Both outstanding issues from the prior review are resolved. The `avgLatencyMs` → `lastLatencyMs` rename correctly names the field as last-reported latency, and the `is_warm_start` exclusion filter has been added to both `findDeadMemories` in `repair-actions.ts` and the dead-memory count in `maintenance-worker.ts`, closing the path by which warm-start synthetic memories could be soft-deleted and destroy predictor training pairs. No new issues introduced by the incremental diff. All five features are functionally complete and correctly guarded.

### Confidence: 7.9/10


```pr-review-confidence-json
{
  "style_maintainability": 8,
  "scope_alignment": 9,
  "duplication_awareness": 8,
  "functional_completeness": 9,
  "pattern_correctness": 9,
  "documentation_coverage": 7
}
```

NicholaiVogel (changes requested):
changes #1-4 (session expiry, hooks config, dead-memory API, openclaw health) are solid. approve those.

requesting changes on #5 (predictor warm-start). the approach is fundamentally wrong:

1. it writes english sentences like "User preference: recent memories are more relevant than old ones" into the memories table as synthetic training data. the predictor operates on numerical feature vectors (recency, frequency, importance, entity connections), not semantic content. these strings teach it nothing about how to rank features.

2. it pollutes the real memory store with synthetic data, relying on an is_warm_start flag to filter them from every recall path. one missed filter and fake data leaks into user results. fragile by design.

3. the correct approach: seed training pairs directly through the predictor sidecar's training API with synthetic feature vectors that encode the desired ranking signals. no database pollution, no flag, no migration, no filter to miss.

suggestion: pull the warm-start into its own PR, drop the migration and memory-search changes, and implement it as direct predictor training pairs instead of fake memories.

the memory-search.ts reformatting should also be separated out. it collapses the hint-only scoring line (hint * 0.3) which was fixed in #291 (changed to just hint). if this merges after #291, the reformat could silently revert that fix during conflict resolution.

NicholaiVogel:
## deep review findings

ran a thorough code review across all 10 files. in addition to the warm-start concerns already raised, found these:

### critical

**1. build-breaking typo in `buildOpenClawHealth()`** (`daemon.ts`)
the never-seen return path uses `avgLatencyMs: 0` but the `OpenClawHealth` interface declares `lastLatencyMs`. this is a TypeScript compile error. the connected/stale path and the fallback in `getDiagnostics()` both correctly use `lastLatencyMs`, only the early return is wrong.

**2. heartbeat double-counts cumulative counters** (`daemon.ts`)
the heartbeat accumulator treats `hooksSucceeded` and `errorCount` as deltas:
```ts
totalSucceeded: (prev?.totalSucceeded ?? 0) + b.hooksSucceeded
```
if the plugin sends cumulative totals (100, then 101), you get 100, then 201. should be direct assignment, not accumulation.

### important

**3. warm-start IDs leak through vector search into candidate pool** (`memory-search.ts`, `core/search.ts`)
the `is_warm_start` filter was added to the FTS JOIN and the final hydration SELECT, but NOT to `vectorSearch()` in `@signet/core`. warm-start IDs enter `vectorMap`, consume slots in the top-K `topIds` slice, then get dropped by hydration. on a fresh install where warm-start memories dominate the embedding index, this degrades vector search to returning fewer results than `limit` (or zero), falling back to keyword-only.

**4. `renewSession` resurrects expired-but-not-GC'd sessions** (`session-tracker.ts`)
no expiry check before renewing. cleanup runs every 15 minutes, so a session that lapsed up to 15 minutes ago can be zombie-renewed:
```ts
// current:
const claim = sessions.get(sessionKey);
if (!claim) return null;

// should be:
if (!claim || Date.now() > claim.expiresAt) return null;
```

**5. warm-start training pairs are equally malformed** (`predictor-warmstart.ts`)
warm-start pairs only populate 4 feature columns (recency_days, access_count, importance, decay_factor). all other feature columns are NULL/zero. this is the same 4-vs-17 dimension mismatch that was identified as a critical predictor bug. worse, the `MIN_PAIRS = 10` check will falsely signal "has training data" after warm-start inserts, potentially masking the cold-start condition for real training pairs.

### recommendation

pull the warm-start (#5) and its memory-search.ts reformatting into a separate PR. the other four changes (session expiry, hooks config, dead-memory API, openclaw health) are good with the fixes above applied.

aaf2tbz:
Addressed in 7e721bfd:

**Warm-start (#5) dropped from this PR** per your recommendation — will go in a separate PR with direct predictor training pairs (no DB writes, no migration, no filter surface). Removed:
- `predictor-warmstart.ts` + `041-warm-start-flag.ts` deleted
- Migration 41 entry removed from `migrations/index.ts`
- `POST /api/predictor/warm-start` endpoint removed from `daemon.ts`
- `runWarmStart` auto-run removed from `main()`
- `is_warm_start` filters reverted in `findDeadMemories`, maintenance-worker count, and both `memory-search.ts` SELECT paths (restores original formatting, eliminating the #291 conflict risk)

**`lastLatencyMs` missed rename fixed** — the never-seen fallback in `buildOpenClawHealth()` still said `avgLatencyMs: 0`; now `lastLatencyMs: 0`.

**`hooksRegistered` capped at 50** — `.slice(0, 50)` added to the unauthenticated heartbeat handler.

This PR now covers only changes #1–4 (session expiry, hooks config, dead-memory API, openclaw health).

aaf2tbz:
Rebased onto main — two conflicts resolved:

**`hooks.ts`** — main added `contextBudgetChars`-based budget + `omitted` count for the "+N more" hint. Our commit used `submitCfg.maxInjectChars`. Merged both: `submitCfg.maxInjectChars ?? cfg.pipelineV2.guardrails.contextBudgetChars` as the budget, `omitted` count preserved so the hint line still works.

**`memory-search.ts`** — our first commit had the old `hint * 0.3` for hint-only scoring; HEAD had the correct `hint` (the #291 fix). Took HEAD's version — our PR no longer touches this line at all.

Force-pushed `7d488a09`.

aaf2tbz:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: session expiry, hooks config, dead-memory API, openclaw health, predictor warm-start` (commit `7d488a09`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7222,47 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	}
+	const b = body as Record<string, unknown>;
+	if (typeof b.pluginVersion !== "string") {
+		return c.json({ error: "pluginVersion (string) is required" }, 400);
+	}
+	const prev = openClawHeartbeat?.data;
+	openClawHeartbeat = {
+		timestamp: new Date().toISOString(),
+		data: {
+			pluginVersion: b.pluginVersion,
```
aaf2tbz: **`hooksRegistered` content not validated as strings.** The cast `(b.hooksRegistered as string[]).slice(0, 50)` is a TypeScript type assertion, not a runtime check. If the plugin sends an array containing non-string elements (numbers, objects, null), they are stored in `openClawHeartbeat.data.hooksRegistered` and later serialized into diagnostics responses. The heartbeat endpoint is unauthenticated — any local process can call it. Fix: `.filter((x): x is string => typeof x === 'string').slice(0, 50)`.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7222,47 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	} catch {
+		return c.json({ error: "Invalid JSON" }, 400);
+	}
+	if (!body || typeof body !== "object") {
+		return c.json({ error: "Body must be an object" }, 400);
+	}
+	const b = body as Record<string, unknown>;
+	if (typeof b.pluginVersion !== "string") {
+		return c.json({ error: "pluginVersion (string) is required" }, 400);
+	}
```
aaf2tbz, line 7286: **No string length caps on unauthenticated heartbeat fields.** The fields `pluginVersion`, `lastError`, and `lastHookCall` are stored directly from request body with no length limit. A local process (or a compromised plugin) can send a multi-MB string that persists in `openClawHeartbeat` for the daemon's lifetime and is included in every diagnostics cache rebuild. At minimum, add a cap (e.g., 512 chars) on each field before assignment.

packages/daemon/src/daemon.ts
```diff
@@ -7768,6 +7855,45 @@ app.post("/api/repair/backfill-hints", async (c) => {
 	});
 });
```
aaf2tbz, line 7904: **`Number('Infinity')` passes the `Number.isNaN` guard.** `Number('Infinity')` is `Infinity`, not `NaN`, so a caller sending `?maxAccessDays=Infinity` or `?maxConfidence=Infinity` passes validation. For `maxAccessDays=Infinity`, the SQL predicate `julianday('now') - julianday(created_at) > Infinity` is always false in SQLite, returning nothing — surprising but harmless. For `maxConfidence=Infinity`, `confidence < Infinity` matches every memory regardless of confidence, bypassing the intent of the filter. Add `!Number.isFinite(maxConfidence) || !Number.isFinite(maxAccessDays)` to the existing guard. Negative values similarly warrant a `< 0` check.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7d488a09`

[Automated Review - pr-reviewer] Five new issues found beyond prior flagged items. The predictor warm-start feature (migration 041, predictor-warmstart.ts, is_warm_start filter, POST /api/predictor/warm-start, auto-run in main) is entirely absent from the diff despite being the fifth named change in the PR title and description — the migration index ends at version 40 and no warm-start files appear anywhere in the changeset. Three additional correctness/safety issues found in the OpenClaw heartbeat handler and dead-memory API.

**Confidence:** High - The missing warm-start implementation is confirmed by the full file content of migrations/index.ts (ends at version 40) and the absence of any is_warm_start reference or new warm-start files in the diff. The hooksRegistered type-assertion issue and Infinity validation gap are unambiguous from the diff text. The two items from the prior deep-review pass (avgLatencyMs typo, double-counting totals) are not re-raised here.


- packages/core/src/migrations/index.ts:388 - **Missing: predictor warm-start feature not implemented.** The PR title and description list 'predictor warm-start' as one of five delivered changes, describing migration 041 (`is_warm_start INTEGER DEFAULT 0`), `predictor-warmstart.ts`, the `is_warm_start` filter in `memory-search.ts`, `POST /api/predictor/warm-start`, and auto-run in `main()`. None of these appear in the diff — the MIGRATIONS array ends at version 40, no 041 import exists, no new files are added, and `memory-search.ts` changes are reformatting only. If warm-start memories are not excluded from recall (the `is_warm_start` filter), synthetic training pairs will appear in search results and corrupt agent context from session one. This feature should either be implemented or removed from the PR description before merge.
- packages/daemon/src/hooks.ts:1371 - **Expiry warning not surfaced in session-start dedup early return.** When `sessionStartSeen` already contains the session key, the function returns a minimal stub without calling `getExpiryWarning()` and without a `warnings` field. If the session is approaching TTL, the operator gets no signal. The dedup guard is the common path for all plugin-invoked session starts after the first. Adding `warnings: req.sessionKey ? [getExpiryWarning(req.sessionKey)].filter((w): w is string => w !== null) : undefined` to the stub return would close this gap without breaking anything.

packages/core/src/migrations/index.ts
```diff
@@ -88,13 +88,7 @@ export const MIGRATIONS: readonly Migration[] = [
...
-				"memory_history",
-				"memory_jobs",
-				"entities",
-				"relations",
-				"memory_entity_mentions",
-			],
+			tables: ["memory_history", "memory_jobs", "entities", "relations", "memory_entity_mentions"],
 		},
 	},
 	{
```
aaf2tbz, line 102: **[Blocking] Predictor warm-start implementation is entirely absent.** The PR title and description both claim this commit ships: migration 041 (`is_warm_start INTEGER DEFAULT 0` + partial index), `predictor-warmstart.ts` (synthetic training pair seeding), an `is_warm_start` exclusion filter in `memory-search.ts`, `POST /api/predictor/warm-start`, and auto-run in `main()`. None of these exist in the diff. The migration index ends at version 40 (`session-transcripts`). There are no `is_warm_start` references anywhere in the changed files. This was flagged in the prior automated review pass and is still unresolved. Either the implementation needs to be added, or the PR description and title need to be corrected to exclude this change.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7222,49 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+			hooksRegistered: Array.isArray(b.hooksRegistered)
+				? (b.hooksRegistered as unknown[]).filter((x): x is string => typeof x === "string").map((s) => s.slice(0, 128)).slice(0, 50)
+				: [],
+			lastHookCall: typeof b.lastHookCall === "string" ? b.lastHookCall.slice(0, 512) : null,
+			lastError: typeof b.lastError === "string" ? b.lastError.slice(0, 512) : null,
+			latencyMs: typeof b.latencyMs === "number" ? b.latencyMs : 0,
+			errorCount: typeof b.errorCount === "number" ? b.errorCount : 0,
+			totalSucceeded: (prev?.totalSucceeded ?? 0) + (typeof b.hooksSucceeded === "number" ? b.hooksSucceeded : 0),
+			totalFailed: (prev?.totalFailed ?? 0) + (typeof b.errorCount === "number" ? b.errorCount : 0),
+		},
```
aaf2tbz, line 7303: **[Warning] Heartbeat accumulation protocol is ambiguous and likely double-counts if the plugin reports cumulative totals.** `totalSucceeded` and `totalFailed` are built by adding the incoming `b.hooksSucceeded` / `b.errorCount` to the previous accumulated values. If the OpenClaw plugin sends cumulative lifetime counts (e.g. 100, then 101 on the next heartbeat), the daemon total becomes 100, then 201 — a 2× overcount. If the plugin sends per-heartbeat deltas, the logic is correct. The field names (`errorCount`, `hooksSucceeded`) are conventionally cumulative, and there is no documented contract specifying which convention is used. This was flagged in the prior deep-review pass and is unchanged. At minimum, add a comment documenting the expected unit (delta vs. cumulative); ideally validate the choice against the OpenClaw plugin implementation.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `25959bc1`

[Automated Review - pr-reviewer] Two issues remain unresolved. The predictor warm-start feature described in the PR (migration 041, predictor-warmstart.ts, is_warm_start column/filter, POST /api/predictor/warm-start endpoint, auto-run in main) is entirely absent from the diff — migrations/index.ts still ends at version 40 and no warm-start files or column references appear anywhere in the changeset. The heartbeat accumulation double-counting concern from the prior review pass also remains unaddressed. Three prior findings have been correctly resolved: the `avgLatencyMs` typo (fixed to `lastLatencyMs`), the unsafe `hooksRegistered` cast (now properly filtered), and the `Infinity` validation gap (fixed with `Number.isFinite`).

**Confidence:** High - The warm-start absence is confirmed by the full migrations/index.ts content ending at version 40 with no v41 import, no is_warm_start column references in memory-search.ts, and no predictor-warmstart.ts file anywhere in the diff. The heartbeat double-counting is unambiguous in the code; whether it is a bug depends on whether the plugin sends deltas or cumulative totals — the PR description says 'accumulates hook call totals' (implying deltas are expected) but the field name `errorCount` is conventionally a running total, making the protocol ambiguous and undocumented.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `32bf3017`

[Automated Review - pr-reviewer] Reviewing PR #295 at HEAD 32bf3017. The two blocking issues from prior passes are resolved: the `avgLatencyMs` typo (fixed), the unsafe `hooksRegistered` cast (fixed), the `Infinity` validation gap (fixed), and the `totalFailed` accumulation is now documented as delta-based with a `hooksFailed`/`errorCount` fallback. The warm-start discrepancy from prior reviews is no longer applicable — the current PR title and description describe exactly four changes (no warm-start). Three minor issues remain.

**Confidence:** High - The incremental diff is a single-line change with a clarifying comment. The full diff is readable end-to-end. No crypto or migration logic was changed. The three flagged items are unambiguous from static analysis of the diff and full file contents.

packages/daemon/src/hooks.ts
```diff
@@ -726,6 +742,8 @@ function updateAccessTracking(ids: string[]): void {
 // Config Loading
```
aaf2tbz, line 755: `KNOWN_HOOKS_KEYS` is a runtime `Set` that must be manually kept in sync with the `HooksConfig` interface. If a new field is added to `HooksConfig` without updating this set, operators will see spurious unknown-key warnings in the daemon log for any valid key. There is no compile-time check enforcing the two are consistent. Consider deriving this from the type (e.g. via a const tuple that both defines the set and serves as the union) or adding a comment listing the expected maintenance step.

packages/daemon/src/pipeline/maintenance-worker.ts
```diff
@@ -380,6 +380,33 @@ export function startMaintenanceWorker(
 			invalidateTraversalCache();
 		}
 
+		// Dead memory hygiene: warn when stale/low-confidence memories accumulate.
```
aaf2tbz, line 385: The dead-memory count query here hardcodes `confidence < 0.1` and `> 90` days, which matches the current defaults in `DeadMemoryOpts` but is not linked to them. If the defaults in `findDeadMemories` are changed (e.g. to be more or less aggressive), the maintenance warning threshold will silently diverge from the API endpoint. Consider extracting the defaults as named constants shared between both sites.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7222,51 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+		timestamp: new Date().toISOString(),
+		data: {
+			pluginVersion: b.pluginVersion.slice(0, 128),
+			hooksRegistered: Array.isArray(b.hooksRegistered)
+				? (b.hooksRegistered as unknown[]).filter((x): x is string => typeof x === "string").map((s) => s.slice(0, 128)).slice(0, 50)
+				: [],
+			lastHookCall: typeof b.lastHookCall === "string" ? b.lastHookCall.slice(0, 512) : null,
+			lastError: typeof b.lastError === "string" ? b.lastError.slice(0, 512) : null,
+			latencyMs: typeof b.latencyMs === "number" ? b.latencyMs : 0,
+			// Plugin sends per-heartbeat deltas, not cumulative totals.
```
aaf2tbz: The `OpenClawHeartbeatData` struct stores `errorCount` (per-heartbeat delta failures) as a separate field alongside `totalFailed` (accumulated). Now that `hooksFailed` is the preferred field for accumulation (with `errorCount` as legacy fallback), `errorCount` in the stored struct serves double duty: it is both the raw per-heartbeat value AND was previously the sole accumulation source. The field is never surfaced in `buildOpenClawHealth()` or in `GET /api/diagnostics/openclaw`, so operators cannot distinguish whether a plugin sends `hooksFailed` vs `errorCount`. If future plugins send both fields, only `hooksFailed` accumulates into `totalFailed` while `errorCount` is stored but invisible in the response — this asymmetry may cause confusion during debugging. Consider either exposing `errorCount` in the health response or removing it from the stored struct once the delta protocol is stable.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `dbe6995f`

[Automated Review — pr-reviewer] Reviewing PR #295 at HEAD dbe6995f. Prior blocking issues (avgLatencyMs typo, unsafe hooksRegistered cast, Infinity validation on dead-memory API, delta-vs-cumulative ambiguity) are all resolved. Three new findings in the incremental diff and one missed in prior passes.

**Confidence:** High - The incremental diff is small and the full diff is readable end-to-end. No migration or crypto logic changed. All four flagged items are unambiguous from static analysis of the diff and full file contents.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7223,50 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+			lastFailedDelta: typeof b.hooksFailed === "number" ? b.hooksFailed : (typeof b.errorCount === "number" ? b.errorCount : 0),
+			totalSucceeded: (prev?.totalSucceeded ?? 0) + (typeof b.hooksSucceeded === "number" ? b.hooksSucceeded : 0),
+			totalFailed: (prev?.totalFailed ?? 0) + (typeof b.hooksFailed === "number" ? b.hooksFailed : (typeof b.errorCount === "number" ? b.errorCount : 0)),
+		},
+	};
+	invalidateDiagnosticsCache();
+	return c.json({ ok: true });
+});
+
```
aaf2tbz, line 7308: `latencyMs` accepts non-finite values. `typeof b.latencyMs === 'number'` is true for `NaN`, `Infinity`, and `-Infinity`. All other numeric inputs in this PR that are stored or displayed are guarded with `Number.isFinite()` (e.g. the dead-memory API at `maxConfidence`, `maxAccessDays`, `limit`). Apply the same guard here: `typeof b.latencyMs === 'number' && Number.isFinite(b.latencyMs) && b.latencyMs >= 0 ? b.latencyMs : 0`. Same applies to the `lastFailedDelta` accumulation — a client sending `hooksFailed: Infinity` would push `totalFailed` to Infinity permanently.

packages/daemon/src/repair-actions.ts
```diff
@@ -1933,3 +1911,92 @@ export function backfillSkippedSessions(
...
+			 WHERE is_deleted = 0
+			   AND importance <= 0.8
+			   AND (
+			     confidence < ?
+			     OR (last_accessed IS NULL AND julianday('now') - julianday(created_at) > ?)
+			     OR (last_accessed IS NOT NULL AND julianday('now') - julianday(last_accessed) > ?)
+			   )
+			 ORDER BY confidence ASC, last_accessed ASC NULLS FIRST
+			 LIMIT ?`,
+		)
```
aaf2tbz, line 1966: `forgetDeadMemories` does not call `writeRepairAudit()`. Every other destructive repair function in this file (`pruneChunkGroupEntities`, `pruneSingletonExtractedEntities`, `deduplicateMemories`, etc.) records an audit row via `writeRepairAudit()`. Omitting it here means bulk soft-deletes triggered via `POST /api/repair/dead-memories/forget` are invisible in the repair audit history, which makes post-incident debugging harder. The action name (`"forget_dead_memories"` or similar) and affected count should be recorded.

packages/daemon/src/repair-actions.ts
```diff
@@ -1933,3 +1911,92 @@ export function backfillSkippedSessions(
...
+ * Returns the number actually deleted (skips already-deleted).
+ */
+export function forgetDeadMemories(accessor: DbAccessor, ids: readonly string[]): number {
+	if (ids.length === 0) return 0;
+	return accessor.withWriteTx((db) => {
+		const stmt = db.prepare("UPDATE memories SET is_deleted = 1, deleted_at = datetime('now') WHERE id = ? AND is_deleted = 0");
+		let total = 0;
+		for (const id of ids) {
+			total += countChanges(stmt.run(id));
+		}
```
aaf2tbz, line 2000: `forgetDeadMemories` uses SQLite's `datetime('now')` for `deleted_at` (format: `YYYY-MM-DD HH:MM:SS`), while every other soft-delete path in the codebase writes an explicit `new Date().toISOString()` (format: `YYYY-MM-DDTHH:MM:SS.mmmZ`). If any query ever filters or sorts on `deleted_at` across rows created by both paths, the mixed formats will sort incorrectly. Use `new Date().toISOString()` and pass it as a parameter for consistency.

packages/daemon/src/session-tracker.ts
```diff
@@ -171,13 +176,44 @@ export function getActiveSessions(): readonly SessionInfo[] {
...
 
+/**
+ * Returns a warning string if the session will expire within WARN_BEFORE_MS,
+ * or null if healthy or not found. Throttled — only warns once per session
+ * until the session is renewed.
+ */
+export function getExpiryWarning(sessionKey: string): string | null {
+	if (isSessionBypassed(sessionKey)) return null;
+	const claim = sessions.get(sessionKey);
+	if (!claim) return null;
```
aaf2tbz, line 195: When a session has already expired but not yet been evicted by the 15-minute cleanup timer, `remaining` will be ≤ 0. `Math.max(1, Math.round(remaining / 60_000))` clamps this to 1, producing the message "session expires in ~1 minute" for a session that has already passed its TTL. Consider a separate branch: `if (remaining <= 0) return 'session TTL exceeded — reinitialize recommended';` to avoid misleading operators.

packages/daemon/src/hooks.ts
```diff
@@ -734,8 +753,18 @@ function loadHooksConfig(): HooksConfig {
...
+		const parsed = parseSimpleYaml(content);
+		const hooks = parsed.hooks;
+		if (!hooks || typeof hooks !== "object") {
+			return getDefaultConfig();
+		}
+		// Warn on unrecognized keys so users catch typos early
+		for (const key of Object.keys(hooks as Record<string, unknown>)) {
+			if (!KNOWN_HOOKS_KEYS.has(key)) {
+				logger.warn("hooks", `Unknown hooks config key: ${key} — check agent.yaml`);
+			}
```
aaf2tbz, line 779: `return hooks as HooksConfig` uses a type assertion. The repository style guide explicitly prohibits `as`. Since `parseSimpleYaml` returns an `any`-typed value (or `Record<string, unknown>`), the `as` here provides no runtime safety — it only suppresses the type checker. Either add a proper type-guard function that validates the shape of `hooks`, or document why the assertion is unavoidable. The previous `return config.hooks || getDefaultConfig()` had the same implicit cast but without the explicit `as`.

packages/daemon/src/session-tracker.ts
```diff
@@ -171,13 +176,45 @@ export function getActiveSessions(): readonly SessionInfo[] {
...
+	const remaining = claim.expiresAt - Date.now();
+	if (remaining <= 0) return "session has expired — reconnect to start a new session";
```
aaf2tbz, line 197: Semantic contradiction: `getExpiryWarning` returns `"session has expired — reconnect to start a new session"` when `remaining <= 0`, but `renewSession` has no guard for this condition and will happily reset the TTL on the same session. During the up-to-15-minute window between TTL expiry and the next cleanup cycle, the user is told to reconnect when they could instead call `POST /api/sessions/:key/renew` to revive the session. Either `renewSession` should return `null` when `now > claim.expiresAt` (hard-expire: force reconnect), or the message here should say something like `"session expires soon — call /renew or /checkpoint"` (soft-expire: renewal still allowed). As-is, the two APIs contradict each other.

packages/daemon/src/hooks.ts
```diff
@@ -944,10 +984,14 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 			dateStyle: "full",
 			timeStyle: "short",
 		});
+		const warnings = req.sessionKey
+			? [getExpiryWarning(req.sessionKey)].filter((w): w is string => w !== null)
+			: [];
 		return {
 			identity: { name: "Agent" },
 			memories: [],
 			inject: `[memory active | /remember | /recall]\n# Current Date & Time\n${now} (${tz})`,
```
aaf2tbz, line 1006: The dedup path omits the `warnings` field entirely when empty via `...(warnings.length > 0 ? { warnings } : {})`, but the full path (line ~1541) always includes it as `warnings: [...].filter(...)`, which produces `warnings: []` when there are no warnings. A caller doing `"warnings" in response` would see different results depending on which code path fired. Prefer a consistent rule — either always omit when empty (use the spread pattern in both places) or always include. The spread pattern is preferable since `SessionStartResponse` already declares `warnings?` as optional.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `32f0299e`

[Automated Review — pr-reviewer] Pass 5 on PR #295 at HEAD 32f0299e. All prior blocking issues are resolved: latencyMs typo fixed, hooksRegistered cast hardened, Infinity validation added, loadHooksConfig type safety improved, forgetDeadMemories audit trail added, and the remaining <= 0 expiry guard added. Two new issues found: a semantic contradiction between the 'reconnect' expiry message and renewSession still succeeding on expired sessions, and a minor inconsistency in warnings field serialization across handleSessionStart return paths.

**Confidence:** High - The incremental diff is four files, small and readable end-to-end. The two flagged issues are unambiguous from static analysis: renewSession contains no guard against remaining <= 0, and the two handleSessionStart return paths serialize warnings differently.

packages/daemon/src/hooks.ts
```diff
@@ -1491,6 +1535,7 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 		})),
```
aaf2tbz, line 1554: The full `handleSessionStart` return path still produces `warnings: []` when `req.sessionKey` is set but `getExpiryWarning` returns null. The expression `[getExpiryWarning(req.sessionKey)].filter((w): w is string => w !== null)` yields an empty array rather than `undefined`. The dedup-stub path was fixed this pass to emit `undefined`, but this path was not updated to match.

In JSON: `"warnings":[]` vs the field being absent are different representations of the same semantic state. Clients checking `if (response.warnings)` (truthy check) will behave differently than clients checking `response.warnings?.length`. Apply the same guard used in the stub: `warnings: req.sessionKey ? [...].filter(...) || undefined : undefined` or inline the `?.length ? ... : undefined` pattern.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `36259b0f`

[Automated Review — pr-reviewer] Pass 6 on PR #295 at HEAD 36259b0. The two issues flagged in pass 5 are partially resolved: the expired-session guard in `renewSession` is now present (blocking issue fixed), and the dedup-stub warnings serialization is fixed. One inconsistency from the prior pass remains unfixed: the full `handleSessionStart` return path still emits `warnings: []` (empty array) rather than `warnings: undefined` when no warning is present, while the dedup stub now correctly omits the field.

**Confidence:** High - The incremental diff is two files totaling ~20 lines. The remaining inconsistency is unambiguous from static analysis of the full file content provided.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d935900e`

[Automated Review — pr-reviewer] Pass 7 on PR #295 at HEAD d935900e. The single outstanding issue from pass 6 — `handleSessionStart` emitting `warnings: []` rather than `warnings: undefined` on the full return path — is now resolved. The IIFE correctly gates on both `sessionKey` presence and array non-emptiness before returning the field, matching the dedup stub's behavior. No new issues introduced. All prior blocking and warning findings across passes 1–6 are resolved.

**Confidence:** High - The incremental diff is 5 lines in a single file. The fix is mechanically correct and unambiguous — the empty-array vs undefined inconsistency is eliminated. All prior flagged issues (latencyMs typo, hooksRegistered cast, Infinity validation, loadHooksConfig type safety, forgetDeadMemories audit trail, renewSession expired-session guard, dedup-stub warnings serialization) were resolved in prior commits. Nothing in the incremental change introduces new logic or new surface area.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: session expiry, hooks config, dead-memory API, openclaw health` (commit `d935900e`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d935900e`

[Automated Review — pr-reviewer bot] Three issues found in the incremental diff since last reviewed SHA: an undescribed UI component calling unverified API endpoints, a prior-raised counter-accumulation concern that was never explicitly acknowledged, and a missing `inFlightTurns` try/finally guard.

**Confidence:** Medium - medium: the ReviewQueueTab endpoint existence cannot be verified from the diff alone — if those routes exist from a prior PR the issue evaporates. The counter-accumulation concern was raised by the human reviewer and not confirmed addressed in the author's response. The inFlightTurns issue is self-healing but still a correctness gap.

- packages/cli/dashboard/src/lib/components/tabs/ReviewQueueTab.svelte:40 - This component calls `GET /api/memory/review-queue` and (from the truncated section) `POST /api/memory/review-queue/:id/resolve`. Neither endpoint appears anywhere in this PR's diff. The PR description also makes no mention of a Review Queue tab — it describes exactly four changes (session expiry, hooks config, dead-memory API, openclaw health). If these endpoints were not shipped in a prior PR, this tab will always render an error state. Please either: (a) confirm which PR added these routes and reference it, or (b) add the route implementations here.
- packages/adapters/openclaw/src/index.ts:1258 - **Heartbeat counter accumulation — prior finding not confirmed resolved.** The human reviewer (#NicholaiVogel) flagged in the prior round that `totalSucceeded`/`totalFailed` are accumulated additively (`prev + b.hooksSucceeded`). If the OpenClaw plugin sends cumulative lifetime totals per heartbeat (e.g. 100, then 101), the daemon would record 100, then 201 — a double-count. The author's fix response (`7e721bfd`) addressed `lastLatencyMs` rename and `hooksRegistered` cap but did not explicitly acknowledge this counter issue. The `lastFailedDelta` interface comment and the PR description both assert "delta, not cumulative", but this protocol invariant must hold in the plugin too. Please either confirm the plugin always sends per-heartbeat deltas (pointing to the relevant plugin code), or switch to direct assignment if cumulative totals are possible.
- packages/adapters/openclaw/src/index.ts:1416 - `inFlightTurns.delete(sig)` is called after `await onUserPromptSubmit(...)` but is not in a `try/finally` block. If `onUserPromptSubmit` throws an unhandled exception, `sig` stays in `inFlightTurns` permanently until the next turn's lazy TTL sweep. Because `sig = sessionKey|count` and count increments each turn, the stuck entry only suppresses the exact failing turn rather than all future turns — so it is self-healing. Still worth wrapping in `try/finally` for correctness:
```ts
if (sig) inFlightTurns.add(sig);
try {
  const result = await onUserPromptSubmit(...);
  ...
} finally {
  if (sig) inFlightTurns.delete(sig);
}
```
- packages/cli/dashboard/src/lib/components/layout/page-headers.ts:99 - A new `review-queue` tab is registered in `ENGINE_TAB_ITEMS` and `PAGE_FOOTERS` but is not mentioned in the PR description. The PR description lists exactly four changes; this is a fifth undescribed UI change. If intentional, add it to the description and include a screenshot (the PR has no UI screenshots for the new tab).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5a1d2b43`

[Automated Review — pr-reviewer bot] Three issues flagged across the code changes: an unaddressed prior-review concern about heartbeat counter accumulation, a Rust parity rule violation (no daemon-rs mirror of four new behavioral features), and an unauthenticated endpoint exposed without rate limiting in non-default bind configurations. Incremental diff since last SHA is documentation-only and clean.

**Confidence:** Medium - medium: the counter-accumulation design question (deltas vs. cumulative) cannot be confirmed without seeing the openclaw plugin's heartbeat implementation. The Rust parity concern depends on whether daemon-rs changes ship in a follow-up PR. The rate-limiting concern is contingent on deployment topology. All three are genuine risks but none are provably broken from the diff alone.


- docs/BENCHMARKS.md:44 - **Unquoted `$SIGNET_WORKSPACE` in shell examples.** The documentation uses bare variable expansion in shell commands:
```bash
cp $SIGNET_WORKSPACE/agent.yaml /tmp/signet-bench/
ln -sf $SIGNET_WORKSPACE/.models /tmp/signet-bench/.models
```
If a user's workspace path contains spaces (e.g., `/home/user/my agents/`), these commands will fail or behave unexpectedly. Should be `"$SIGNET_WORKSPACE"` throughout.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7223,50 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	openClawHeartbeat = {
+		timestamp: new Date().toISOString(),
+		data: {
+			pluginVersion: b.pluginVersion.slice(0, 128),
+			hooksRegistered: Array.isArray(b.hooksRegistered)
+				? (b.hooksRegistered as unknown[]).filter((x): x is string => typeof x === "string").map((s) => s.slice(0, 128)).slice(0, 50)
+				: [],
+			lastHookCall: typeof b.lastHookCall === "string" ? b.lastHookCall.slice(0, 512) : null,
+			lastError: typeof b.lastError === "string" ? b.lastError.slice(0, 512) : null,
+			latencyMs: typeof b.latencyMs === "number" && Number.isFinite(b.latencyMs) ? b.latencyMs : 0,
```
PR-Reviewer-Ant, line 7297: **Unaddressed prior review item — heartbeat counter accumulation.** The human reviewer (@NicholaiVogel, 2026-03-23T09:49:36Z) flagged that `totalSucceeded` and `totalFailed` are accumulated as running sums of incoming `hooksSucceeded`/`hooksFailed` values:
```ts
totalSucceeded: (prev?.totalSucceeded ?? 0) + (typeof b.hooksSucceeded === "number" ? b.hooksSucceeded : 0),
totalFailed: (prev?.totalFailed ?? 0) + (typeof b.hooksFailed === "number" ? b.hooksFailed : ...),
```
The comment asserts "Plugin sends per-heartbeat deltas, not cumulative totals", but this is an undocumented protocol assumption. If the openclaw plugin sends cumulative lifetime totals (which is the more common telemetry pattern), every heartbeat will inflate the stored counters — e.g., two heartbeats with a cumulative total of 100 become 200 in the stored state. The author's 2026-03-23T15:07:19Z response addressed `avgLatencyMs` and warm-start but did not respond to this concern. Either (a) document the delta contract in the POST `/api/diagnostics/openclaw/heartbeat` API spec and enforce it with a max-delta sanity check, or (b) change the handler to accept cumulative totals and use direct assignment.

packages/daemon/src/daemon.ts
```diff
@@ -7176,6 +7223,50 @@ app.get("/api/diagnostics/:domain", (c) => {
 	return c.json(domainData);
 });
 
+// ---------------------------------------------------------------------------
+// OpenClaw plugin health diagnostics
```
PR-Reviewer-Ant, line 7269: **Unauthenticated endpoint without rate limiting.** `POST /api/diagnostics/openclaw/heartbeat` is explicitly unauthenticated. While the daemon binds to `127.0.0.1` by default, the `SIGNET_BIND=0.0.0.0` environment variable is a documented, supported configuration for container deployments (AGENTS.md). In that topology, this endpoint is network-accessible with no rate limiting. The auth module at `packages/daemon/src/auth/rate-limiter.ts` already provides rate-limiting infrastructure. In a 0.0.0.0 bind scenario, a network peer could spam heartbeats to continuously overwrite the health snapshot, or use the endpoint as a low-cost way to trigger `invalidateDiagnosticsCache()` on every call. Consider applying the existing rate limiter to this route, or at minimum document that the endpoint is only safe on the default localhost bind.

packages/daemon/src/daemon.ts
```diff
@@ -444,6 +446,21 @@ let diagnosticsCache: {
 } | null = null;
 const DIAGNOSTICS_CACHE_TTL_MS = 2000;
```
PR-Reviewer-Ant, line 451: **Rust parity rule.** AGENTS.md states: "Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`." This PR introduces four behavioral features with protocol-visible effects: session expiry warnings + renew endpoint, userPromptSubmit config, dead-memory review API, and openclaw health tracking. None of these appear to be reflected in the shadow rewrite based on the diff. If the shadow proxy (`shadowEnabled: true`) is running in parallel, clients receiving responses from `daemon-rs` will see divergent behavior (missing `warnings[]`, missing `/api/repair/dead-memories`, missing `/api/diagnostics/openclaw`, missing `/api/sessions/:key/renew`). If daemon-rs updates are intentionally deferred to a follow-up PR, note that explicitly in the description.

packages/daemon/src/hooks.ts
```diff
@@ -734,8 +753,24 @@ function loadHooksConfig(): HooksConfig {
 
 	try {
 		const content = readFileSync(configPath, "utf-8");
-		const config = parseSimpleYaml(content);
-		return config.hooks || getDefaultConfig();
+		const parsed = parseSimpleYaml(content);
```
PR-Reviewer-Ant, line 769: **`as` assertions in `loadHooksConfig()` violate codebase convention.** The config reconstruction casts YAML sub-objects directly to typed interfaces:
```ts
sessionStart: ... ? record.sessionStart as HooksConfig["sessionStart"] : undefined,
```
The AGENTS.md style guide explicitly says `no 'as' — fix the types instead of asserting`. Practically, this means nested fields like `enabled`, `recallLimit`, and `maxInjectChars` receive no runtime validation — a YAML value of `{ enabled: "yes" }` would pass the object check and be treated as a valid `userPromptSubmit` config. The existing pattern is permissive enough that incorrect types degrade silently rather than error loudly.

packages/daemon/src/hooks.ts
```diff
@@ -45,33 +63,23 @@ import {
...
-	shouldRunSessionDecay,
-} from "./pipeline/aspect-feedback";
-import {
-	initContinuity,
-	recordPrompt,
-	recordRemember,
-	shouldCheckpoint,
-	consumeState,
-	clearContinuity,
-	setStructuralSnapshot,
```
PR-Reviewer-Ant, line 62: Duplicate import block — almost certainly a rebase merge artifact. `propagateMemoryStatus` is imported twice from `"./knowledge-graph"` and the `graph-traversal` barrel (`invalidateTraversalCache`, `resolveFocalEntities`, `setTraversalStatus`, `traverseKnowledgeGraph`) is imported twice from `"./pipeline/graph-traversal"`. The second graph-traversal block is the only place `TraversalPath` is imported, which is used in `serializeTraversalPath()`. The fix is to drop the first (incomplete) copy and keep only the second, which already includes `TraversalPath`. Even if bun's bundler silently deduplicates these, `bun run typecheck` (tsc) will error on the duplicate identifier declarations.

packages/daemon/src/daemon.ts
```diff
@@ -7218,6 +7265,50 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	}
+	if (!body || typeof body !== "object") {
+		return c.json({ error: "Body must be an object" }, 400);
+	}
+	const b = body as Record<string, unknown>;
+	if (typeof b.pluginVersion !== "string") {
+		return c.json({ error: "pluginVersion (string) is required" }, 400);
+	}
+	const prev = openClawHeartbeat?.data;
+	openClawHeartbeat = {
```
PR-Reviewer-Ant, line 7288: Previously raised (review 5a1d2b43, not dismissed): `POST /api/diagnostics/openclaw/heartbeat` is intentionally unauthenticated, but has no rate limiting. When `SIGNET_BIND=0.0.0.0` is used (documented for containers), this endpoint is exposed to the network — an attacker can flood it to inflate `totalSucceeded`/`totalFailed` counters and pollute the diagnostics dashboard. At minimum, add a per-IP rate limiter consistent with the existing `rate-limiter.ts` infrastructure, or document that this endpoint must only be reachable from localhost.

packages/daemon/src/daemon.ts
```diff
@@ -6387,6 +6424,16 @@ app.post("/api/sessions/:key/bypass", async (c) => {
 	return c.json({ key, bypassed: enabled });
```
PR-Reviewer-Ant, line 6424: Previously raised (review 5a1d2b43, not dismissed): AGENTS.md requires Rust parity — any behavioral change to `@signet/daemon` must be reflected in `packages/daemon-rs/`. This PR adds four new behavioral features (session renew endpoint, dead-memory API, OpenClaw health heartbeat, and `userPromptSubmit` config-driven injection). None of these appear to be mirrored in daemon-rs. If this is intentional (e.g., daemon-rs is behind a feature flag), please add a comment or AGENTS.md note explaining the exception.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5cb97a87`

[Automated Review — pr-reviewer bot] Incremental diff (5a1d2b43→5cb97a878c) is documentation and migrations only and is clean. The full PR diff contains one blocking correctness issue: duplicate import statements in hooks.ts that are almost certainly a rebase merge artifact. Two prior-raised concerns (unauthenticated heartbeat endpoint rate-limiting, Rust parity rule) remain unaddressed.

**Confidence:** Medium - The duplicate import finding is high-confidence — the two extra import blocks are visually unambiguous in the file content and would fail tsc/typecheck even if bun's bundler silently deduplicates them. The rate-limiting and Rust-parity concerns are genuine but depend on deployment topology and whether daemon-rs ships in a follow-up; confidence on those is lower.

packages/daemon/src/daemon.ts
```diff
@@ -7218,6 +7265,50 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+			totalFailed: (prev?.totalFailed ?? 0) + (typeof b.hooksFailed === "number" ? b.hooksFailed : (typeof b.errorCount === "number" ? b.errorCount : 0)),
+		},
+	};
+	invalidateDiagnosticsCache();
+	return c.json({ ok: true });
+});
+
+app.get("/api/diagnostics/openclaw", (c) => {
+	return c.json(buildOpenClawHealth());
+});
```
PR-Reviewer-Ant, line 7311: **[NEW] Negative delta values corrupt running totals.** The only guard on `hooksFailed` and `hooksSucceeded` is `typeof ... === 'number'`, which passes for negative integers. A heartbeat with `{ hooksFailed: -999 }` will silently decrement `totalFailed` below zero, corrupting the diagnostics counter permanently until the daemon restarts. The endpoint is unauthenticated, so any local process can send this. Add `Number.isFinite(v) && v >= 0` checks (same pattern already used for `latencyMs`) before accumulating:
```ts
const clampDelta = (v: unknown): number =>
  typeof v === 'number' && Number.isFinite(v) && v >= 0 ? v : 0;
totalSucceeded: (prev?.totalSucceeded ?? 0) + clampDelta(b.hooksSucceeded),
totalFailed:    (prev?.totalFailed    ?? 0) + clampDelta(b.hooksFailed ?? b.errorCount),
```

packages/daemon/src/daemon.ts
```diff
@@ -7218,6 +7265,50 @@ app.get("/api/diagnostics/:domain", (c) => {
 	return c.json(domainData);
 });
 
+// ---------------------------------------------------------------------------
+// OpenClaw plugin health diagnostics
+// ---------------------------------------------------------------------------
```
PR-Reviewer-Ant, line 7270: **[PREVIOUSLY FLAGGED — unaddressed] Rust parity rule.** AGENTS.md states: *"Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`."* This PR adds four new behavioral features with HTTP routes (`POST /api/sessions/:key/renew`, `POST /api/diagnostics/openclaw/heartbeat`, `GET /api/diagnostics/openclaw`, `GET /api/repair/dead-memories`, `POST /api/repair/dead-memories/forget`) plus session expiry warning propagation and hooks-config validation. No daemon-rs changes appear in this diff. Flagged in the two prior automated reviews; no human response. Either the daemon-rs mirror PR should be linked here, or the rule should be explicitly waived for this PR.

packages/daemon/src/daemon.ts
```diff
@@ -7810,6 +7901,48 @@ app.post("/api/repair/backfill-hints", async (c) => {
 	});
 });
 
+// ---------------------------------------------------------------------------
+// Dead memory hygiene
+// ---------------------------------------------------------------------------
+
+app.get("/api/repair/dead-memories", (c) => {
```
PR-Reviewer-Ant, line 7909: **[NEW] `maxConfidence` has no upper-bound guard.** The validation rejects `maxConfidence < 0` but not `maxConfidence > 1`. Passing `maxConfidence=1.0` returns every memory with `importance <= 0.8` as a dead candidate — potentially the entire store. A caller who then POSTs those IDs to `/api/repair/dead-memories/forget` performs a near-total soft-delete. Add `|| maxConfidence > 1` to the existing validation condition.

packages/daemon/src/daemon.ts
```diff
@@ -7218,6 +7265,50 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+// Unauthenticated — daemon is local-only, plugin may not carry auth token
+app.post("/api/diagnostics/openclaw/heartbeat", async (c) => {
+	let body: unknown;
+	try {
+		body = await c.req.json();
+	} catch {
+		return c.json({ error: "Invalid JSON" }, 400);
+	}
+	if (!body || typeof body !== "object") {
+		return c.json({ error: "Body must be an object" }, 400);
```
PR-Reviewer-Ant, line 7281: **[PREVIOUSLY FLAGGED — unaddressed] Unauthenticated heartbeat endpoint with no rate limiting.** The comment justifies this by daemon-local-only design, but with `SIGNET_BIND=0.0.0.0` (documented env var) the endpoint is network-reachable with no auth and no rate limit. Other repair/diagnostics endpoints use auth middleware. At minimum, add the same rate limiter applied to other unauthenticated endpoints, or document the explicit threat-model decision in AGENTS.md.

packages/daemon/src/hooks.ts
```diff
@@ -752,8 +766,24 @@ function loadHooksConfig(): HooksConfig {
...
+		if (!hooks || typeof hooks !== "object") {
+			return getDefaultConfig();
+		}
+		// Warn on unrecognized keys so users catch typos early
+		const record = hooks as Record<string, unknown>;
+		for (const key of Object.keys(record)) {
+			if (!KNOWN_HOOKS_KEYS.has(key as keyof HooksConfig)) {
+				logger.warn("hooks", `Unknown hooks config key: ${key} — check agent.yaml`);
+			}
+		}
```
PR-Reviewer-Ant, line 780: **[NEW] Three `as` casts in `loadHooksConfig` violate the codebase 'no `as`' convention.** `record.sessionStart as HooksConfig["sessionStart"]` suppresses the type checker without validating nested field types. If YAML provides `userPromptSubmit: { recallLimit: "ten" }`, the cast silently produces a `HooksConfig["userPromptSubmit"]` with a string `recallLimit`. The `?? 10` fallback does not rescue this because a string is not nullish, and `hybridRecall` receives a string `limit`. Use explicit type guards or restrict to known-good primitive fields rather than casting entire sub-objects.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `02d84bb7`

[Automated Review — pr-reviewer bot] Incremental diff (5cb97a87→02d84bb7) resolves the previously blocking duplicate-import issue in hooks.ts. Two new correctness issues found in the full diff that were not caught in prior reviews: negative numeric inputs to the unauthenticated heartbeat endpoint can corrupt the totalFailed/totalSucceeded running counters, and the dead-memory endpoint accepts maxConfidence > 1.0 with no upper-bound guard. Two previously-flagged concerns (Rust parity rule, unauthenticated heartbeat rate limiting) remain unaddressed and are re-raised.

**Confidence:** Medium - medium: the negative-counter corruption and missing upper-bound check are unambiguous from the diff. The Rust parity concern requires knowledge of whether daemon-rs changes ship in a parallel PR — none visible here. The rate-limiting risk is topology-dependent (local-only default vs. SIGNET_BIND=0.0.0.0).

packages/daemon/src/daemon.ts
```diff
@@ -7218,6 +7265,51 @@ app.get("/api/diagnostics/:domain", (c) => {
...
+	}
+	const prev = openClawHeartbeat?.data;
+	openClawHeartbeat = {
+		timestamp: new Date().toISOString(),
+		data: {
+			pluginVersion: b.pluginVersion.slice(0, 128),
+			hooksRegistered: Array.isArray(b.hooksRegistered)
+				? (b.hooksRegistered as unknown[]).filter((x): x is string => typeof x === "string").map((s) => s.slice(0, 128)).slice(0, 50)
+				: [],
+			lastHookCall: typeof b.lastHookCall === "string" ? b.lastHookCall.slice(0, 512) : null,
```
PR-Reviewer-Ant, line 7295: The `latencyMs` field accepts any finite number including negatives, while `totalSucceeded`/`totalFailed`/`lastFailedDelta` now all clamp with `Math.max(0, ...)`. For consistency and to avoid displaying negative latencies in the health dashboard, clamp this too: `Math.max(0, b.latencyMs)`. Minor but inconsistent with the pattern established in the same block.

packages/daemon/src/daemon.ts
```diff
@@ -7218,6 +7265,51 @@ app.get("/api/diagnostics/:domain", (c) => {
 	return c.json(domainData);
```
PR-Reviewer-Ant, line 7265: Unauthenticated endpoint with no rate limiting — flagged in two prior rounds, still unaddressed. A process with access to the daemon's bind address can repeatedly POST large `hooksSucceeded` deltas and inflate `totalSucceeded` to Number.MAX_SAFE_INTEGER. When `SIGNET_BIND=0.0.0.0` (e.g. containers), this is reachable from the network. Even a simple per-IP token-bucket or a request-size cap on the cumulative counters would close the obvious attack surface.

packages/daemon/src/daemon.ts
```diff
@@ -448,6 +450,21 @@ let diagnosticsCache: {
 } | null = null;
```
PR-Reviewer-Ant, line 450: Rust parity rule violation (AGENTS.md: "Any behavioral change made to @signet/daemon must also be reflected in packages/daemon-rs/"). This PR adds four new behavioral features to @signet/daemon — session expiry warnings + renewSession, HooksConfig unknown-key validation, dead-memory review API, and OpenClaw health tracking — with no corresponding changes visible in daemon-rs. The shadow proxy logs divergences to shadow-divergences.jsonl, and operators relying on parity for safety signals will see false negatives for all four features. If daemon-rs changes ship in a follow-up PR, a comment here or in the PR description linking it would satisfy the intent of this rule.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `595fc63d`

[Automated Review — pr-reviewer bot] Incremental diff (02d84bb7→595fc63d) addresses both previously-blocking issues cleanly. Two convention violations from earlier rounds remain unaddressed: the Rust parity rule (AGENTS.md mandates daemon-rs mirrors every behavioral change to @signet/daemon) and the unauthenticated heartbeat endpoint with no rate limiting. One minor new finding: `latencyMs` is not clamped to non-negative, inconsistent with the fix applied to the adjacent counter fields.

**Confidence:** Medium - The incremental diff is small and unambiguous — both fixes are correct. The Rust parity gap is real but depends on whether daemon-rs changes exist in an unlinked parallel PR. The rate-limiting risk is topology-dependent (local-only vs. SIGNET_BIND=0.0.0.0). The latencyMs gap is observable from the diff but is cosmetic in practice.

PR #296, opened 2026-03-23
feat: add reviews sync Cloudflare Worker

NicholaiVogel:
## Summary

- Standalone D1-backed Cloudflare Worker at `workers/reviews/` for aggregating marketplace reviews
- Receives batch syncs from user daemons, serves public paginated reads
- Already deployed to `signet-reviews.nicholaivogelfilms.workers.dev`

## Routes

| Route | Description |
|---|---|
| `GET /` | Health check |
| `GET /api/reviews` | Public paginated read, 30s cache |
| `POST /api/reviews/sync` | Batch upsert from signetai daemon |

## Security

- Rate limiting: 5 syncs/IP/60s
- `X-Signet-Sync: 1` header required on POST
- Full input validation (UUID format, timestamps, field lengths, rating bounds)
- 512KB body size cap
- Idempotent upserts (re-sync safe)
- CORS restricted to marketplace domain + localhost

## Structure

Lives at `workers/reviews/`, outside `packages/` to avoid polluting the bun workspace and build graph. Independent deployable with its own wrangler.toml.

## Pair with

- #293 (pre-configures daemon sync URL to this worker)

## Test plan

- [x] `GET /` returns `{"ok":true,"service":"signet-reviews"}`
- [x] `GET /api/reviews` returns empty set with summary
- [ ] `POST /api/reviews/sync` accepts valid daemon payload
- [ ] Rate limiter rejects after 5 rapid syncs
- [ ] Missing `X-Signet-Sync` header returns 400

### DIFF
--- a/workers/reviews/src/index.ts
+++ b/workers/reviews/src/index.ts
@@ -132,7 +132,8 @@ function corsHeaders(origin: string): Record<string, string> {
 
 function isOriginAllowed(origin: string | null, allowed: string): boolean {
 	if (allowed === "*") return true;
-	if (!origin) return false;
+	// No Origin header = server-to-server (daemon sync). Allow.
+	if (!origin) return true;
 	if (origin === allowed) return true;
 	try {
 		const u = new URL(origin);
@@ -186,7 +187,8 @@ async function upsertReviews(
 				   title        = excluded.title,
 				   body         = excluded.body,
 				   updated_at   = excluded.updated_at,
-				   received_at  = excluded.received_at`,
+				   received_at  = excluded.received_at
+				 WHERE excluded.updated_at >= reviews.updated_at`,
 			)
 			.bind(
 				r.id, r.targetType, r.targetId, r.displayName,
@@ -195,14 +197,18 @@ async function upsertReviews(
 			),
 	);
 
-	const results = await db.batch(stmts);
-	let accepted = 0;
-	let rejected = 0;
-	for (const res of results) {
-		if (res.success) accepted++;
-		else rejected++;
+	try {
+		const results = await db.batch(stmts);
+		let accepted = 0;
+		let rejected = 0;
+		for (const res of results) {
+			if (res.success) accepted++;
+			else rejected++;
+		}
+		return { accepted, rejected };
+	} catch {
+		return { accepted: 0, rejected: reviews.length };
 	}
-	return { accepted, rejected };
 }
 
 // ---------------------------------------------------------------------------
--- a/workers/reviews/src/index.ts
+++ b/workers/reviews/src/index.ts
@@ -1,0 +1,396 @@
+/**
+ * Signet Reviews Worker
+ *
+ * Central aggregation endpoint for marketplace reviews.
+ * Receives synced reviews from user daemons and serves them publicly.
+ *
+ * Routes:
+ *   GET  /              - health check
+ *   GET  /api/reviews   - list/query reviews (public, cached)
+ *   POST /api/reviews/sync - batch upsert from signetai daemon
+ */
+
+export interface Env {
+	DB: D1Database;
+	RATE_LIMITER: { limit(opts: { key: string }): Promise<{ success: boolean }> };
+	CORS_ORIGIN: string;
+}
+
+interface ReviewRow {
+	id: string;
+	target_type: "skill" | "mcp";
+	target_id: string;
+	display_name: string;
+	rating: number;
+	title: string;
+	body: string;
+	created_at: string;
+	updated_at: string;
+	received_at: string;
+}
+
+interface IncomingReview {
+	id: string;
+	targetType: "skill" | "mcp";
+	targetId: string;
+	displayName: string;
+	rating: number;
+	title: string;
+	body: string;
+	createdAt: string;
+	updatedAt: string;
+}
+
+// Validation limits
+const LIMITS = {
+	BATCH_SIZE: 100,
+	TARGET_ID: 200,
+	DISPLAY_NAME: 50,
+	TITLE: 100,
+	BODY: 2_000,
+	BODY_MIN: 10,
+	TITLE_MIN: 3,
+	REQUEST_BODY_BYTES: 512_000,
+} as const;
+
+// ---------------------------------------------------------------------------
+// Validation
+// ---------------------------------------------------------------------------
+
+function parseStr(value: unknown, min: number, max: number): string | null {
+	if (typeof value !== "string") return null;
+	const t = value.trim();
+	if (t.length < min || t.length > max) return null;
+	return t;
+}
+
+function parseTargetType(v: unknown): "skill" | "mcp" | null {
+	if (v === "skill" || v === "mcp") return v;
+	return null;
+}
+
+function parseRating(v: unknown): number | null {
+	if (typeof v !== "number" || !Number.isFinite(v)) return null;
+	const r = Math.round(v);
+	if (r < 1 || r > 5) return null;
+	return r;
+}
+
+const UUID_RE = /^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+
+function parseUUID(v: unknown): string | null {
+	if (typeof v !== "string" || !UUID_RE.test(v)) return null;
+	return v.toLowerCase();
+}
+
+const ISO_RE = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d+)?Z$/;
+
+function parseTimestamp(v: unknown): string | null {
+	if (typeof v !== "string" || !ISO_RE.test(v)) return null;
+	const ts = new Date(v).getTime();
+	if (isNaN(ts)) return null;
+	if (ts > Date.now() + 30 * 24 * 60 * 60 * 1_000) return null;
+	return v;
+}
+
+function validateReview(raw: unknown): IncomingReview | null {
+	if (typeof raw !== "object" || raw === null || Array.isArray(raw)) return null;
+	const r = raw as Record<string, unknown>;
+
+	const id = parseUUID(r["id"]);
+	const targetType = parseTargetType(r["targetType"]);
+	const targetId = parseStr(r["targetId"], 1, LIMITS.TARGET_ID);
+	const displayName = parseStr(r["displayName"], 1, LIMITS.DISPLAY_NAME);
+	const rating = parseRating(r["rating"]);
+	const title = parseStr(r["title"], LIMITS.TITLE_MIN, LIMITS.TITLE);
+	const body = parseStr(r["body"], LIMITS.BODY_MIN, LIMITS.BODY);
+	const createdAt = parseTimestamp(r["createdAt"]);
+	const updatedAt = parseTimestamp(r["updatedAt"]);
+
+	if (
+		!id || !targetType || !targetId || !displayName ||
+		rating === null || !title || !body || !createdAt || !updatedAt
+	) {
+		return null;
+	}
+
+	return { id, targetType, targetId, displayName, rating, title, body, createdAt, updatedAt };
+}
+
+// ---------------------------------------------------------------------------
+// CORS
+// ---------------------------------------------------------------------------
+
+function corsHeaders(origin: string): Record<string, string> {
+	return {
+		"Access-Control-Allow-Origin": origin,
+		"Access-Control-Allow-Methods": "GET, POST, OPTIONS",
+		"Access-Control-Allow-Headers": "Content-Type, X-Signet-Sync",
+		"Access-Control-Max-Age": "86400",
+	};
+}
+
+function isOriginAllowed(origin: string | null, allowed: string): boolean {
+	if (allowed === "*") return true;
+	if (!origin) return false;
+	if (origin === allowed) return true;
+	try {
+		const u = new URL(origin);
+		if (u.hostname === "localhost" || u.hostname === "127.0.0.1") return true;
+	} catch {
+		// ignore
+	}
+	return false;
+}
+
+function json(body: unknown, status: number, extra: Record<string, string> = {}): Response {
+	return new Response(JSON.stringify(body), {
+		status,
+		headers: { "Content-Type": "application/json;charset=UTF-8", ...extra },
+	});
+}
+
+// ---------------------------------------------------------------------------
+// DB
+// ---------------------------------------------------------------------------
+
+function rowToPublic(row: ReviewRow) {
+	return {
+		id: row.id,
+		targetType: row.target_type,
+		targetId: row.target_id,
+		displayName: row.display_name,
+		rating: row.rating,
+		title: row.title,
+		body: row.body,
+		createdAt: row.created_at,
+		updatedAt: row.updated_at,
+	};
+}
+
+async function upsertReviews(
+	db: D1Database,
+	reviews: IncomingReview[],
+	receivedAt: string,
+): Promise<{ accepted: number; rejected: number }> {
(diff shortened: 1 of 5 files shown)


NicholaiVogel:
Hi @NicholaiVogel - I'm @NicholaiVogel's PR-reviewing agent powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: add reviews sync Cloudflare Worker` (commit `3a4415a2`) now and I'll follow up shortly with feedback.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,396 @@
...
+
+	const rows = (batch[0]?.results ?? []) as ReviewRow[];
+	const summary = (batch[1]?.results?.[0] ?? { total: 0, avg_rating: 0 }) as {
+		total: number;
+		avg_rating: number | null;
+	};
+
+	return json(
+		{
+			reviews: rows.map(rowToPublic),
```
NicholaiVogel, line 272: **[Security — no real sync authentication]** `X-Signet-Sync: 1` is documented publicly in the PR description and emitted in the CORS `Access-Control-Allow-Headers`, so any caller knows it. Combined with the `source`/`type` fields being client-controlled body values, this endpoint has no secret authentication. Any script that knows the worker URL can inject or overwrite reviews for any target. Add a pre-shared secret: either a bearer token compared via a timing-safe comparison (or via Workers Secret) or an HMAC-SHA256 signature over the body. Example using a `SYNC_SECRET` binding:
```ts
const sig = request.headers.get('X-Signet-Signature') ?? '';
const expected = await hmacHex(env.SYNC_SECRET, await request.clone().text());
if (!timingSafeEqual(sig, expected)) return json({ error: 'unauthorized' }, 401, cors);
```

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,396 @@
...
+				   body         = excluded.body,
+				   updated_at   = excluded.updated_at,
+				   received_at  = excluded.received_at`,
+			)
+			.bind(
+				r.id, r.targetType, r.targetId, r.displayName,
+				r.rating, r.title, r.body,
+				r.createdAt, r.updatedAt, receivedAt,
+			),
+	);
```
NicholaiVogel, line 198: **[Data integrity — stale sync overwrites newer data]** The `ON CONFLICT DO UPDATE` unconditionally replaces every mutable column. If a daemon re-syncs an older snapshot, it will silently overwrite a review that was already updated to a newer version. Add a staleness guard:
```sql
ON CONFLICT(id) DO UPDATE SET
  display_name = excluded.display_name,
  rating       = excluded.rating,
  title        = excluded.title,
  body         = excluded.body,
  updated_at   = excluded.updated_at,
  received_at  = excluded.received_at
WHERE excluded.updated_at >= reviews.updated_at
```
Without this, two daemons syncing the same review ID in different orders will leave the DB in a non-deterministic state.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+function parseIntParam(v: string | null, min: number, max: number, def: number): number {
+	if (!v) return def;
+	const n = parseInt(v, 10);
+	if (!Number.isFinite(n)) return def;
```
NicholaiVogel, line 217: **[Bug — unhandled db.batch() exception]** D1's `batch()` is transactional: if any statement errors, the whole batch is rolled back and the call throws rather than returning per-row `success: false`. The current code has no try/catch around `db.batch(stmts)`, so a D1 error propagates as an unhandled exception, producing a Cloudflare-generated 500 response **without** CORS headers. That breaks CORS-aware clients silently. Wrap with try/catch and return a structured error:
```ts
try {
  const results = await db.batch(stmts);
  // ... count accepted/rejected
} catch (err) {
  return { accepted: 0, rejected: reviews.length };
}
```
(Or let `handleSync` catch it and return `json({ error: 'db error' }, 500, cors)`.)

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,396 @@
...
+		return json({ error: "no valid reviews in batch", skipped }, 400, cors);
+	}
+
+	const receivedAt = new Date().toISOString();
+	const { accepted, rejected } = await upsertReviews(env.DB, valid, receivedAt);
+
+	return json({ success: true, accepted, rejected, skipped, receivedAt }, 200, cors);
+}
+
+// ---------------------------------------------------------------------------
```
NicholaiVogel, line 357: **[Low — rate-limit key degrades to shared bucket]** When both `CF-Connecting-IP` and `X-Forwarded-For` are absent, `clientIp` falls back to `"unknown"` and all such requests share a single rate-limit bucket (5 total, not 5/IP). On Cloudflare Workers `CF-Connecting-IP` is always injected by the platform, so this won't trigger in practice, but the fallback should still be a hard rejection rather than a shared key to avoid surprising behavior if the binding changes:
```ts
if (!ip) return json({ error: 'cannot determine client IP' }, 400, cors);
```

workers/reviews/wrangler.toml
```diff
@@ -0,0 +1,20 @@
+name               = "signet-reviews"
+main               = "src/index.ts"
+compatibility_date = "2025-01-01"
+
+[[d1_databases]]
+binding       = "DB"
+database_name = "signet-reviews"
+database_id   = "d1fcce93-0b8a-4b95-b67d-39e9611dd4b3"
```
NicholaiVogel, line 8: **[Informational — D1 database_id committed]** `d1fcce93-0b8a-4b95-b67d-39e9611dd4b3` is now public. The ID alone doesn't grant access (Cloudflare auth still required), but it enables targeted enumeration/monitoring of the database. Consider noting this as an accepted risk or rotating if this repo is public.

NicholaiVogel (commented):
Three issues warrant attention before merge: (1) the sync endpoint has no real authentication — X-Signet-Sync:1 is public knowledge and any caller can inject arbitrary reviews; (2) the upsert has no staleness guard, so re-syncing old data silently overwrites newer reviews; (3) db.batch() exceptions are not caught, producing uncorsed 500s on D1 errors. Everything else — parameterized queries, body-size cap, UUID/timestamp validation, CORS logic — is solid.

### Confidence: 8.4/10
- Merge conflict detection confidence: 9


```pr-review-confidence-json
{
  "style_maintainability": 9,
  "merge_conflict_detection": 9,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `feat: add reviews sync Cloudflare Worker` (commit `4ceafe0f`); taking another look at the feature work and I will report back shortly.

NicholaiVogel (commented):
Two of three prior issues are resolved: the staleness guard (`WHERE excluded.updated_at >= reviews.updated_at`) correctly prevents stale re-syncs from overwriting newer data, and the `db.batch()` try/catch prevents bare 500s on D1 errors. The original critical issue — unauthenticated write access — remains unaddressed. The `isOriginAllowed` change (no-Origin → allow) is not a security regression (CORS has never gated non-browser callers), but it makes explicit that the sync endpoint is intentionally open to any server-to-server caller, which sharpens the auth gap. One new correctness issue: when `db.batch()` throws, the handler still returns HTTP 200 with `success: true`, making a full DB failure indistinguishable from a stale-batch no-op.

### Confidence: 8.5/10
- Merge conflict detection confidence: 9


```pr-review-confidence-json
{
  "style_maintainability": 9,
  "merge_conflict_detection": 9,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 8,
  "pattern_correctness": 8,
  "documentation_coverage": 7
}
```

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,402 @@
...
+		).bind(...bindings, limit, offset),
+		env.DB.prepare(
+			`SELECT COUNT(*) as total, AVG(rating) as avg_rating
+			 FROM reviews ${where}`,
+		).bind(...bindings),
+	]);
+
+	const rows = (batch[0]?.results ?? []) as ReviewRow[];
+	const summary = (batch[1]?.results?.[0] ?? { total: 0, avg_rating: 0 }) as {
+		total: number;
```
NicholaiVogel, line 266: **[Unresolved from prior review — Critical] No authentication on sync endpoint.** The only gate on `POST /api/reviews/sync` is `X-Signet-Sync: 1`. This is a public, trivially-spoofable string — anyone can inject arbitrary reviews:
```
curl -X POST https://signet-reviews.nicholaivogelfilms.workers.dev/api/reviews/sync \
  -H 'X-Signet-Sync: 1' \
  -H 'Content-Type: application/json' \
  -d '{"source":"signet-marketplace","type":"reviews-sync","reviews":[...]}'
```
The `isOriginAllowed` change this revision makes explicit that server-to-server callers with no `Origin` header are allowed — which is correct for a daemon, but it sharpens the risk: **any caller** on the internet can write to the reviews DB. At minimum, add a pre-shared bearer token checked via a Cloudflare Worker Secret (`env.SYNC_SECRET`), e.g.:
```ts
if (request.headers.get('Authorization') !== `Bearer ${env.SYNC_SECRET}`) {
  return json({ error: 'unauthorized' }, 401, cors);
}
```
Add `SYNC_SECRET` as a Wrangler secret (`wrangler secret put SYNC_SECRET`) and type it in `Env`.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,402 @@
...
+	);
+
+	try {
+		const results = await db.batch(stmts);
+		let accepted = 0;
+		let rejected = 0;
+		for (const res of results) {
+			if (res.success) accepted++;
+			else rejected++;
+		}
```
NicholaiVogel: **[New] DB error returns `success: true` with HTTP 200.** When `db.batch()` throws, the catch returns `{ accepted: 0, rejected: reviews.length }`, which the caller receives as `{ success: true, accepted: 0, rejected: N }` with a 200 status. This is indistinguishable from a valid response where all reviews were stale. The caller (daemon) has no way to know a storage failure occurred and may silently drop the sync. Return HTTP 500 on catch instead:
```ts
} catch (err) {
  console.error('db.batch failed', err);
  return json({ error: 'storage error' }, 500, cors);
}
```
And remove the batch-level `rejected` counting from the catch path — the `upsertReviews` function's return type should reflect partial success only when the batch itself succeeds.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,402 @@
...
+		"Access-Control-Allow-Origin": origin,
+		"Access-Control-Allow-Methods": "GET, POST, OPTIONS",
+		"Access-Control-Allow-Headers": "Content-Type, X-Signet-Sync",
+		"Access-Control-Max-Age": "86400",
+	};
+}
+
+function isOriginAllowed(origin: string | null, allowed: string): boolean {
+	if (allowed === "*") return true;
+	// No Origin header = server-to-server (daemon sync). Allow.
```
NicholaiVogel, line 135: **[Resolved] `isOriginAllowed` no-Origin change** is correct: CORS enforcement is a browser-only mechanism and was never blocking non-browser callers anyway. Allowing `null` Origin for daemon sync is the right call. No issue here — noting it's confirmed fixed.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,402 @@
...
+		db
+			.prepare(
+				`INSERT INTO reviews
+				   (id, target_type, target_id, display_name, rating, title, body,
+				    created_at, updated_at, received_at)
+				 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+				 ON CONFLICT(id) DO UPDATE SET
+				   display_name = excluded.display_name,
+				   rating       = excluded.rating,
+				   title        = excluded.title,
```
NicholaiVogel, line 187: **[Resolved] Staleness guard** (`WHERE excluded.updated_at >= reviews.updated_at`) is correctly implemented. SQLite/D1 supports conditional `DO UPDATE SET … WHERE` since 3.32. Idempotent re-syncs of equal timestamps still apply, which is the right behavior. No issue.

NicholaiVogel:
Hi @NicholaiVogel - quick follow-up pass on `feat: add reviews sync Cloudflare Worker` (commit `77a9b683`); taking another look at the feature work and I will report back shortly.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,401 @@
...
+			offset,
+			summary: {
+				count: summary.total,
+				avgRating: summary.avg_rating != null
+					? Math.round(summary.avg_rating * 10) / 10
+					: 0,
+			},
+		},
+		200,
+		{ ...cors, "Cache-Control": "public, max-age=30, stale-while-revalidate=60" },
```
NicholaiVogel, line 284: **[Resolved from prior reviews]** The try/catch is now correctly placed in `handleSync` rather than inside `upsertReviews`. D1 failures will now propagate as throws, be caught here, and return a proper `{error: 'storage error'}` 500 with CORS headers. This is the right fix.

NicholaiVogel, line 284: **[Still unresolved — third review]** The sync endpoint has no real authentication. `X-Signet-Sync: 1` is a static, public header; the `source`/`type` payload fields are trivially forgeable. Any anonymous caller can upsert reviews for any `target_id` with any `displayName` and `rating`. This is an open write API. Options to fix in increasing strength: (1) a shared secret in a `wrangler` secret binding, checked via `request.headers.get('X-Signet-Token') === env.SYNC_SECRET`; (2) an HMAC signature over the payload; (3) mTLS at the Cloudflare service binding layer. Option (1) is the minimum acceptable bar — store the token in a Wrangler secret, not a `[vars]` entry, so it isn't committed or exposed in the dashboard plaintext.

workers/reviews/src/index.ts
```diff
@@ -0,0 +1,401 @@
...
+				   body         = excluded.body,
+				   updated_at   = excluded.updated_at,
+				   received_at  = excluded.received_at
+				 WHERE excluded.updated_at >= reviews.updated_at`,
+			)
+			.bind(
+				r.id, r.targetType, r.targetId, r.displayName,
+				r.rating, r.title, r.body,
+				r.createdAt, r.updatedAt, receivedAt,
+			),
```
NicholaiVogel, line 197: Minor misleading metric: when the `ON CONFLICT DO UPDATE WHERE excluded.updated_at >= reviews.updated_at` predicate is false (stale re-sync), SQLite/D1 counts the statement as `success: true` with `changes: 0`. The returned `accepted` count will include these silent no-ops, so callers can't distinguish 'inserted/updated' from 'stale-skipped'. Consider tracking `changes` from D1 results if an accurate accepted count matters to the daemon.

NicholaiVogel (commented):
The prior correctness bug (db.batch() exceptions producing HTTP 200 success:true) is now fixed — the try/catch was correctly moved up to handleSync so D1 failures return a proper 500 with CORS headers and an error body. The staleness guard remains in place. However, the critical unauthenticated write issue flagged in both prior reviews is still unaddressed: any caller who knows the static header (X-Signet-Sync: 1) and payload shape can inject or overwrite arbitrary reviews for any target. Three reviews in, this remains the single blocking issue.

### Confidence: 8.5/10
- Merge conflict detection confidence: 9


```pr-review-confidence-json
{
  "style_maintainability": 9,
  "merge_conflict_detection": 9,
  "scope_alignment": 9,
  "duplication_awareness": 9,
  "functional_completeness": 7,
  "pattern_correctness": 8,
  "documentation_coverage": 8
}
```

PR #301, opened 2026-03-23
fix: support provider: none for extraction and synthesis

NicholaiVogel:
## Summary

- Adds `"none"` to extraction and synthesis provider type unions, config validation, and daemon startup guards
- When provider is `"none"`, the daemon starts cleanly with no pipeline or LLM initialization
- Embeddings, manual `/remember`, and `/recall` continue to work independently

Supersedes #213 (PatchyToes) with a clean implementation on current main. The original PR had merge conflicts and bundled unrelated threshold changes.

Closes #213.

## Test plan

- [ ] Set `extraction.provider: none` and `synthesis.provider: none` in agent.yaml
- [ ] Verify daemon starts without pipeline errors
- [ ] Verify `/remember` and `/recall` still work
- [ ] Verify existing configs without `"none"` are unaffected

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -10795,7 +10795,9 @@ async function main() {
 
 	// Create synthesis provider — separate from extraction because synthesis
 	// needs a smarter model that can reason across long context
-	if (memoryCfg.pipelineV2.synthesis.enabled && memoryCfg.pipelineV2.synthesis.provider !== "none") {
+	if (memoryCfg.pipelineV2.synthesis.provider === "none") {
+		logger.info("config", "Synthesis provider set to 'none', synthesis disabled");
+	} else if (memoryCfg.pipelineV2.synthesis.enabled) {
 		let effectiveSynthesisProvider = memoryCfg.pipelineV2.synthesis.provider;
 		const synthesisOllamaBaseUrl = normalizeRuntimeBaseUrl(
 			memoryCfg.pipelineV2.synthesis.endpoint,
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -10533,8 +10533,8 @@ async function main() {
 	if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
 
 	const providerHints = getConfiguredProviderHints(AGENTS_DIR);
-	const validExtractionProviders = new Set(["ollama", "claude-code", "opencode", "codex", "anthropic", "openrouter"]);
-	const validSynthesisProviders = new Set(["ollama", "claude-code", "opencode", "anthropic", "openrouter"]);
+	const validExtractionProviders = new Set(["none", "ollama", "claude-code", "opencode", "codex", "anthropic", "openrouter"]);
+	const validSynthesisProviders = new Set(["none", "ollama", "claude-code", "opencode", "anthropic", "openrouter"]);
 
 	providerRuntimeResolution.extraction = {
 		configured: providerHints.extraction,
@@ -10582,7 +10582,9 @@ async function main() {
 	);
 	const ollamaFallbackMaxContextTokens = resolveDefaultOllamaFallbackMaxContextTokens();
 	const extractionOpenCodeShouldManage = isManagedOpenCodeLocalEndpoint(extractionOpenCodeBaseUrl);
-	if (effectiveExtractionProvider === "opencode") {
+	if (effectiveExtractionProvider === "none") {
+		logger.info("config", "Extraction provider set to 'none', pipeline LLM disabled");
+	} else if (effectiveExtractionProvider === "opencode") {
 		if (extractionOpenCodeShouldManage) {
 			const serverReady = await ensureOpenCodeServer(4096);
 			if (!serverReady) {
@@ -10729,8 +10731,9 @@ async function main() {
 	});
 
 	// Create LLM provider once, register as daemon-wide singleton
-	const llmProvider =
-		effectiveExtractionProvider === "anthropic" && anthropicApiKey
+	const llmProvider = effectiveExtractionProvider === "none"
+		? null
+		: effectiveExtractionProvider === "anthropic" && anthropicApiKey
 			? createAnthropicProvider({
 					model: effectiveExtractionModel || "haiku",
 					apiKey: anthropicApiKey,
@@ -10773,7 +10776,9 @@ async function main() {
 											}
 										: {}),
 								});
-	initLlmProvider(llmProvider);
+	if (llmProvider) {
+		initLlmProvider(llmProvider);
+	}
 
 	// Initialize model registry for dynamic model discovery
 	if (memoryCfg.pipelineV2.modelRegistry.enabled) {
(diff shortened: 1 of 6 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: support provider: none for extraction and synthesis` (commit `9a222ceb`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -10729,8 +10731,9 @@ async function main() {
 	});
```
PR-Reviewer-Ant, line 10731: **[blocking] Rust parity rule violated.** AGENTS.md states: 'Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`.' This PR introduces a new startup code path (`effectiveExtractionProvider === 'none'` → null llmProvider, skip pipeline, skip synthesis) that has no corresponding change in `packages/daemon-rs/`. If the shadow proxy is active (`shadowEnabled: true`), the two daemons will diverge on any config with `provider: none`, producing spurious divergence log entries and potentially incorrect shadow comparisons. The parity update is required per repository convention.

packages/daemon/src/daemon.ts
```diff
@@ -10773,7 +10776,9 @@ async function main() {
 											}
```
PR-Reviewer-Ant, line 10776: **[warning] Null llmProvider consumption risk.** `llmProvider` is set to `null` when extraction provider is `none`, and `initLlmProvider` is guarded. However, daemon.ts is a very large file and there may be other call sites that invoke `getLlmProvider()` (or equivalent) deeper in startup or request handlers without null-checking the result. Any such site would throw at runtime when `provider: none` is configured. The test plan items are all unchecked — this should be verified against the running daemon before merge.

packages/daemon/src/daemon.ts
```diff
@@ -10790,7 +10795,7 @@ async function main() {
```
PR-Reviewer-Ant, line 10795: **[warning] Silent synthesis skip with no log.** When `synthesis.provider` is `none`, the entire synthesis block is skipped with no log message, while the extraction `none` path emits `logger.info("Extraction provider set to 'none', pipeline LLM disabled")`. For consistency and debuggability (especially for users who set `synthesis.enabled: true` alongside `provider: none`), a matching info log should be emitted here.

packages/daemon/src/memory-config.ts
```diff
@@ -295,14 +295,15 @@ export function loadPipelineConfig(
 	const flatModel = raw.extractionModel;
 
 	type ProviderKind =
+		| "none"
 		| "ollama"
 		| "claude-code"
 		| "opencode"
 		| "codex"
 		| "anthropic"
```
PR-Reviewer-Ant, line 303: **[nitpick] Local `ProviderKind` union diverges from core types.** `packages/core/src/types.ts` defines `PipelineExtractionConfig.provider` as a union including `'none'`, but `memory-config.ts` re-declares its own local `ProviderKind` type. These must be kept in sync manually. Consider importing the union from core to eliminate the duplication risk.

packages/daemon/src/daemon.ts
```diff
@@ -10987,9 +10992,9 @@ async function main() {
 		);
```
PR-Reviewer-Ant, line 10994: **[warning] No tests shipped.** The repository testing philosophy (AGENTS.md) requires 'every new feature ships with tests that describe the behavior.' There are no test file changes in this diff. At minimum, a test should verify that `loadPipelineConfig` accepts `provider: none` for both extraction and synthesis, and that the resulting config is handled without error by the startup path.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9a222ceb`

[Automated Review — pr-reviewer bot] PR #301 adds `provider: none` support for extraction and synthesis. The core logic is correct and well-scoped, but two convention violations should be addressed before merge: the mandatory Rust parity update is absent, and no tests ship with the behavioral change.

**Confidence:** Medium - medium: the diff is small and the logic is straightforward, but daemon.ts is very large and the full set of callsites for the llmProvider singleton and effectiveExtractionProvider are not visible in the diff, so downstream null-handling cannot be fully verified.

NicholaiVogel:
Addressed the review feedback in 1fc79d7:

**P1 - Synthesis worker startup**: Added `provider !== "none"` guard to `pipeline/index.ts` line 152, so the synthesis worker won't start when provider is "none".

**P1 - Summary worker resolveProvider**: Added explicit `case "none"` that throws rather than silently falling through to ollama. In practice this path won't be hit because the pipeline startup already gates on extraction provider, but it's there for safety.

**Silent synthesis skip**: Added info log when synthesis provider is "none", matching the extraction log.

**Rust parity**: daemon-rs is a shadow proxy that compares HTTP endpoint responses. Provider "none" affects startup behavior (which providers to initialize), not endpoint semantics. When extraction is "none", the pipeline doesn't start, so no extraction-dependent endpoints change behavior. No parity-rules.json update needed.

**Tests**: Fair point. The config parsing is covered by the existing `memory-config.test.ts` patterns. Happy to add an explicit test if desired, but the behavioral verification (daemon starts clean, no errors) is better suited to integration testing against the running daemon.

**ProviderKind duplication**: Acknowledged. The local type exists because the config resolver needs a narrower union than the core type (it resolves unknown strings to valid providers). Importing from core would be cleaner but is a separate refactor.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `1fc79d77`

[Automated Review — pr-reviewer bot] PR #301 (incremental: 9a222ceb → 1fc79d7). Three changes since last review: daemon.ts synthesis guard reordered, pipeline/index.ts synthesis worker guard added, summary-worker.ts `case 'none'` throw added. The two P1 items from the prior review are addressed. One new gap introduced by the `case 'none'` throw is worth flagging.

**Confidence:** High - The incremental diff is small and the logic is easy to trace. The summary-worker gap is unambiguous: `startSummaryWorker` is called unconditionally inside `startPipeline`, which runs whenever extraction provider is not 'none'. `resolveProvider` reads `synthesis.provider`, so a config with extraction active and synthesis 'none' will throw on every summary job.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1011,6 +1011,8 @@ async function resolveProvider(cfg: ReturnType<typeof loadMemoryConfig>): Promis
 	const ollamaFallbackMaxContextTokens =
 		resolveDefaultOllamaFallbackMaxContextTokens();
```
PR-Reviewer-Ant, line 1012: The `case 'none': throw` guard is reached in a valid partial config: extraction provider is a real provider (pipeline starts, `startSummaryWorker` is called) but `synthesis.provider` is `'none'`. In that scenario every enqueued summary job will hit this throw, retry up to `max_attempts` times (each attempt re-throws), and eventually go `dead` — generating persistent log noise and silently dropping session summaries.

The summary worker's tick already has early-return guards (`!pipelineV2.enabled`, `shadowMode`). Adding `cfg.pipelineV2.synthesis.provider === 'none'` to that block would make the worker idle cleanly instead of cycling through failing retries:
```ts
if (!cfg.pipelineV2.enabled || cfg.pipelineV2.shadowMode || cfg.pipelineV2.synthesis.provider === 'none') {
  scheduleTick(POLL_INTERVAL_MS);
  return;
}
```
The throw can stay as a last-resort safety net, but the tick guard prevents it from being reached in normal operation.

PR #302, opened 2026-03-23
feat(cli): add configurable Signet workspace path + migration

NicholaiVogel:
## Summary

This PR adds first-class workspace path management to the Signet CLI so users can run Signet from non-default locations (for example OpenClaw/Clawdbot/Moltbot workspace paths) without behavior changes.

### What’s included

- Added persistent workspace resolution and storage
  - New config file: `~/.config/signet/workspace.json`
  - Resolution order:
    1. `--path`
    2. `SIGNET_PATH`
    3. stored workspace config
    4. default `~/.agents`

- Added new CLI commands:
  - `signet workspace status`
  - `signet workspace set [path]`
    - idempotent behavior
    - safe copy + verify migration from current workspace
    - conflict detection (with `--force` override)
    - patches discovered OpenClaw-family configs by default

- Integrated workspace management into `signet configure`
  - New “Workspace path” section that reuses the same migration logic

- Updated CLI codepaths to honor effective workspace
  - runtime default path resolution
  - portable import/export now uses active workspace
  - git/help output and OpenClaw restart guidance now print active workspace paths

- Added tests
  - `packages/cli/src/lib/workspace.test.ts`
  - `packages/cli/src/features/workspace.test.ts`

- Documentation sweep
  - Updated README + docs to use workspace-aware `$SIGNET_WORKSPACE` references for operational paths
  - Added path notes and new workspace command docs

## Validation

Ran locally:

- `cd packages/cli && bun test`
- `bun x biome check` on changed CLI files

All passed.


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(cli): add configurable Signet workspace path + migration` (commit `01b3d6c1`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(cli): add configurable Signet workspace path + migration` (commit `01b3d6c1`) now and I'll follow up shortly with feedback.

packages/cli/src/features/workspace.ts
```diff
@@ -0,0 +1,295 @@
...
+	if (!aStat.isFile() || !bStat.isFile()) {
+		return false;
+	}
+
+	if (aStat.size !== bStat.size) {
+		return false;
+	}
+
+	const aHash = hashFile(a);
+	const bHash = hashFile(b);
```
PR-Reviewer-Ant, line 287: `filesEqual` calls `readFileSync(path)` on both files before hashing. For `memories.db` files in the gigabyte range this loads the whole database into memory, risking OOM or extreme latency during workspace migration. Use a streaming hash instead (pipe a `ReadStream` through `crypto.createHash`) or add a size-gated early-exit: skip the hash comparison and treat files larger than some threshold as unequal unless `--force` is set.

packages/cli/src/features/workspace.ts
```diff
@@ -0,0 +1,295 @@
...
+	if (plan.adds.length + plan.overwrites.length > 0) {
+		verifyCoreFiles(prev, next);
+	}
+
+	const cfgPath = writeConfiguredWorkspacePath(next, opts.env);
+	const patch = opts.patchOpenClaw === false ? [] : await new OpenClawConnector().configureWorkspace(next);
+
+	return {
+		previousPath: prev,
+		nextPath: next,
```
PR-Reviewer-Ant, line 137: `setWorkspacePath` writes the config file (`writeConfiguredWorkspacePath`) and then calls `new OpenClawConnector().configureWorkspace(next)`. If `configureWorkspace` throws, the workspace config has already been persisted to `~/.config/signet/workspace.json` and files have already been copied — but the caller's spinner reports failure and `process.exit(1)` may fire. The workspace *is* set at this point; the user will get a misleading failure message and will not know whether to retry. Either move the OpenClaw patch before the config write and roll back on failure, or catch patching errors separately and report them as non-fatal warnings.

packages/cli/src/lib/runtime.ts
```diff
@@ -12,14 +12,14 @@ import {
...
 import { fileURLToPath } from "node:url";
 import { parseSimpleYaml } from "@signet/core";
 import chalk from "chalk";
 import { resolveDaemonNetwork } from "./network.js";
+import { resolveAgentsDir } from "./workspace.js";
 
-export const AGENTS_DIR = process.env.SIGNET_PATH || join(homedir(), ".agents");
+export const AGENTS_DIR = resolveAgentsDir().path;
 export const DEFAULT_PORT = 3850;
 const DAEMON_BASE_URLS = [`http://127.0.0.1:${DEFAULT_PORT}`, `http://[::1]:${DEFAULT_PORT}`] as const;
```
PR-Reviewer-Ant, line 24: `AGENTS_DIR` now resolves the workspace from `~/.config/signet/workspace.json` at CLI module-load time. The daemon is a separate subprocess that almost certainly still only reads `SIGNET_PATH`. If `doStart` (or the service install path) does not forward `SIGNET_PATH=AGENTS_DIR` to the daemon subprocess environment, the CLI and the running daemon will point at different directories after `signet workspace set`. This would cause silent data divergence — the CLI reads from the new workspace while the daemon writes to `~/.agents`. Please confirm daemon launch propagates the resolved path, and if it does not, add it.

packages/cli/src/features/workspace.ts
```diff
@@ -0,0 +1,295 @@
...
+		plan.conflicts.push(`${srcDir} is not a directory`);
+		return;
+	}
+
+	if (!existsSync(dstDir)) {
+		plan.dirs.push(dstDir);
+	}
+
+	if (existsSync(dstDir)) {
+		const dstStat = lstatSync(dstDir);
```
PR-Reviewer-Ant, line 184: `scanDir` has no exclusion list. The `.daemon/` subdirectory contains a `pid` file referencing the process ID of the daemon running against the *old* workspace. Copying it means `signet daemon status` (and the stop/restart codepath that reads the PID file) at the new workspace will target the wrong process. At a minimum, exclude `.daemon/pid` and `.daemon/logs/` from the copy plan while still migrating `.daemon/auth-secret`.

packages/cli/src/features/workspace.ts
```diff
@@ -0,0 +1,295 @@
...
+	const plan = existsSync(prev) && prev !== next ? buildCopyPlan(prev, next, opts.force === true) : emptyPlan();
+	if (plan.conflicts.length > 0) {
+		const sample = plan.conflicts.slice(0, 5).join("\n  - ");
+		const more = plan.conflicts.length > 5 ? `\n  ... and ${plan.conflicts.length - 5} more` : "";
+		throw new Error(`workspace migration has conflicts:\n  - ${sample}${more}`);
+	}
+
+	applyCopyPlan(plan);
+	if (plan.adds.length + plan.overwrites.length > 0) {
+		verifyCoreFiles(prev, next);
```
PR-Reviewer-Ant, line 129: `chooseWorkspaceCandidate` falls back to `~/.openclaw/workspace` when `listWorkspaceCandidates` returns nothing. For a user who does not have OpenClaw installed this is a wrong and surprising default. The fallback should be the current path (passed as the argument) or `~/.agents`, not a harness-specific directory.

packages/cli/src/features/workspace.test.ts
```diff
@@ -0,0 +1,82 @@
+import { describe, expect, it } from "bun:test";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { setWorkspacePath } from "./workspace.js";
+
+function makeEnv(root: string): NodeJS.ProcessEnv {
+	return {
+		...process.env,
```
PR-Reviewer-Ant, line 9: `makeEnv` spreads `process.env` without clearing `SIGNET_PATH`. In a CI environment or developer shell where `SIGNET_PATH` is set, `resolveAgentsDir` will return the env-override path instead of the XDG-backed config path, making the resolution tests produce incorrect assertions. Explicitly delete `SIGNET_PATH` from the constructed env object to isolate each test.

packages/cli/src/features/configure.ts
```diff
@@ -24,7 +30,8 @@ const sections = [
 export async function configureAgent(deps: Deps): Promise<void> {
 	console.log(deps.signetLogo());
```
PR-Reviewer-Ant, line 37: `let dir = deps.agentsDir` and `let path = join(dir, 'agent.yaml')` are reassigned mid-loop after the workspace section runs. Repo convention is `prefer const`; avoid `let` by refactoring the workspace section into a function that returns `{ dir, path, yaml }` and rebind with `const` at the call site.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `01b3d6c1`

[Automated Review — pr-reviewer] PR #302 adds configurable workspace path management to the Signet CLI. The core library code is well-structured and the tests cover the main contract, but there is one blocking resource-safety issue (full-file read for SHA-256 on potentially-large SQLite databases), a functional gap around daemon–CLI path divergence after workspace changes, and several narrower warnings.

**Confidence:** Medium - The daemon launch code and OpenClawConnector internals are not in the diff, so I cannot confirm whether SIGNET_PATH is correctly forwarded to the daemon subprocess after a workspace change. Everything in the new CLI files is fully readable.


- AGENTS.md:382 - The 'User Data Location' section in AGENTS.md still hardcodes `~/.agents/` in both prose and the directory tree, while every other doc in this PR was updated to `$SIGNET_WORKSPACE`. Since AGENTS.md is symlinked to CLAUDE.md and is the primary reference for AI assistants working in this repo, leaving it inconsistent means agent-generated code will continue to hardcode `~/.agents` even for non-default workspace setups.

NicholaiVogel:
Addressed, pushed in `f25131e7`.

What I changed:

- Workspace migration safety
  - Moved `writeConfiguredWorkspacePath(...)` to run *after* OpenClaw-family patching succeeds.
  - This prevents a persisted workspace flip when `configureWorkspace(...)` fails.
- Large-file hash safety
  - Replaced full-file `readFileSync` hashing with chunked SHA-256 hashing via `openSync/readSync`.
  - Avoids loading large `memories.db` files fully into memory.
- Daemon artifacts during copy
  - Migration now skips `.daemon/pid` and `.daemon/logs/**`.
  - `.daemon/auth-secret` is still migrated.
- Non-interactive safety
  - `signet workspace set` without an explicit path now fails in non-TTY mode.
  - No implicit migration to presets in automation.
- Candidate fallback
  - `chooseWorkspaceCandidate(...)` fallback now resolves to the current path rather than an OpenClaw-specific path.
- Configure flow cleanup
  - Refactored workspace handling in `configure.ts` to avoid `let dir/path` reassignment and keep config state explicit.
- Docs consistency
  - Updated AGENTS.md workspace references to `$SIGNET_WORKSPACE` semantics (default `~/.agents`).
- Test updates
  - Added migration test that asserts daemon pid/log files are excluded.
  - Isolated workspace tests from ambient `SIGNET_PATH` by setting it to `undefined` in test env.

Validation run:

- `cd packages/cli && bun test` ✅
- `cd packages/cli && bun run build:cli` ✅
- `./node_modules/.bin/biome check packages/cli/src/features/workspace.ts packages/cli/src/commands/workspace.ts packages/cli/src/features/configure.ts packages/cli/src/features/workspace.test.ts AGENTS.md` ✅

Also verified daemon start path propagation remains aligned: `startDaemon(...)` exports `SIGNET_PATH=agentsDir` in child env.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b9128308`

[Automated Review — pr-reviewer] PR #302 adds configurable workspace path management. The four blocking/warning issues flagged in the previous review (chunked file hashing, non-interactive safety guard, daemon artifact exclusion, writeConfiguredWorkspacePath ordering) are all correctly addressed in the follow-up push. Two new narrower issues remain.

**Confidence:** Medium - All new CLI files are fully readable. OpenClawConnector.configureWorkspace() and getDiscoveredWorkspacePaths() are not in the diff, so the OpenClaw patching path cannot be fully verified from this review alone.

packages/cli/src/features/workspace.ts
```diff
@@ -0,0 +1,328 @@
...
+
+	const entries = readdirSync(srcDir, { withFileTypes: true });
+	for (const entry of entries) {
+		const src = join(srcDir, entry.name);
+		const dst = join(dstDir, entry.name);
+		const nextRel = rel.length === 0 ? entry.name : join(rel, entry.name);
+		if (shouldSkip(nextRel, entry.isDirectory())) {
+			continue;
+		}
+		if (entry.isDirectory()) {
```
PR-Reviewer-Ant, line 202: Symlinks inside the source workspace hit `plan.conflicts.push('${src} is not a regular file')` and abort migration. `Dirent.isFile()` and `Dirent.isDirectory()` both return false for symbolic links, so any symlink in the workspace directory tree is treated as an unresolvable conflict. Connectors install symlinks pointing *at* the workspace (e.g. `~/.codex/skills → ~/.agents/skills`), not inside it, so this is uncommon in practice — but it is an untested path that produces an opaque error with no recovery option. Silently skipping symlinks (like `pid`/`logs`) would be safer. At minimum the error message should say 'symbolic link' to make the failure diagnosable.

packages/cli/src/features/workspace.ts
```diff
@@ -0,0 +1,328 @@
...
+		throw new Error(`workspace migration has conflicts:\n  - ${sample}${more}`);
+	}
+
+	applyCopyPlan(plan);
+	if (plan.adds.length + plan.overwrites.length > 0) {
+		verifyCoreFiles(prev, next);
+	}
+
+	const patch = opts.patchOpenClaw === false ? [] : await new OpenClawConnector().configureWorkspace(next);
+	const cfgPath = writeConfiguredWorkspacePath(next, opts.env);
```
PR-Reviewer-Ant, line 133: `changed: prev !== next || res.configuredPath !== next` returns `true` when `prev === next` but `configuredPath === null` — for example, `signet workspace set ~/.agents` on a fresh install that has no `workspace.json` yet. The callers (`workspace.ts` and `configure.ts`) both use `result.changed` to decide whether to print "Restart the daemon", which is incorrect in this case: the daemon is already using that path and requires no restart. `changed` should reflect only `prev !== next` for the restart-needed semantics; config persistence is a side effect of the call regardless.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e3f6295d`

[Automated Review — pr-reviewer] PR #302 incremental diff (b9128308 → e3f6295d): three targeted fixes — symlink skipping during migration, corrected idempotency assertion in tests, and a simplification of the `changed` flag logic. No new issues found.

**Confidence:** High - The incremental diff is small (~40 net lines across 2 files). All three changes are clearly correct: the symlink guard prevents migration conflicts on symlinked entries, the `changed` simplification is semantically correct (a daemon restart is only needed when the active path changes, not when the stored config diverges), and the test assertion fix reflects the correct contract for idempotent workspace-set. No new surface area introduced.

NicholaiVogel:
Addressed the two latest pr-reviewer inline comments in `e3f6295d`.

Changes made:
- Updated `changed` semantics in `setWorkspacePath` to reflect only path changes (`prev !== next`), so restart guidance is only shown when the effective workspace path actually changes.
- Updated workspace migration scan to skip symbolic links instead of treating them as hard conflicts.
- Added test coverage for symlink skipping and updated idempotency expectation for the refined `changed` behavior.

Validation:
- `cd packages/cli && bun test`
- `./node_modules/.bin/biome check packages/cli/src/features/workspace.ts packages/cli/src/features/workspace.test.ts`

Also manually triggered pr-reviewer from this machine after push:
- `pr-reviewer review Signet-AI/signetai#302`
- Result: `completed verdict=Some(NoIssues) comments=0`

PR #304, opened 2026-03-23
fix(dashboard): remove review-queue tab; fix knowledge tab load performance

aaf2tbz:
## What

Two dashboard fixes in one commit.

### Review queue tab removed

The review queue page duplicates what the pipeline tab already shows — dedup decisions and pipeline events are visible there. Removed from all registration points:
- `ReviewQueueTab.svelte` deleted
- `"review-queue"` removed from `TabId` union, `VALID_TABS`, and `ENGINE_TABS` in `navigation.svelte.ts`
- Removed from `ENGINE_TABS` in `tab-group-focus.svelte.ts`
- Removed from `ENGINE_TAB_ITEMS` and `PAGE_FOOTERS` in `page-headers.ts`
- `{#await ReviewQueueTab}` block removed from `TabContentLoader.svelte`

### Knowledge tab performance

Two issues causing slow / flickery initial load:

**`entityDensityOpacity` O(n²)** — the function was called once per rendered entity row, and each call mapped over the full `entities` array to compute `maxScore`. With 50 entities that's 50×50 = 2,500 iterations per render cycle. Fixed by hoisting `maxScore` into a `$derived maxDensityScore` that recomputes once when `entities` changes.

**Window focus re-triggering traversal** — `window.addEventListener("focus", () => loadTraversal())` fired on every alt-tab or window refocus, setting `loadingTraversal = true` and causing the traversal section to flash back to skeleton state. Added a 60s cooldown: re-fetches only if the last traversal load was ≥ 60s ago.

## Test plan

- [ ] Engine tab group no longer shows "Review Queue" tab
- [ ] `#review-queue` hash in URL falls back to default tab (no crash)
- [ ] Knowledge tab loads and entity density bars render without flicker
- [ ] Alt-tabbing away and back does not reset the traversal section skeleton within 60s


aaf2tbz:
Closing — branched off wrong base, includes #295 commits. Recreating clean off main.

PR #306, opened 2026-03-23
fix(pipeline): resolve document-worker queue starvation (issue #303)

### LINKED ISSUE #303: document-worker queue starvation — single permanently-failing job starves entire document_ingest queue (distinct from #248)
# Bug: document-worker queue starvation — single permanently-failing job starves entire document_ingest queue (distinct from #248)

**Version:** 0.74.0  
**Environment:**
- Ubuntu 24.04 LTS
- Node v25.6.0
- Bun 1.3.11
- Signet v0.74.0 (latest)
- SQLite memory store at `/home/adamp/.agents/memory/memories.db`

---

## Summary

When a `document_ingest` job fails with a **permanent, non-transient error** (specifically `UNIQUE constraint failed: index 'idx_memories_content_hash_unique'`), that job is returned to `pending` status without advancing its `attempts` counter past 1. Because the failing job has the earliest `created_at` timestamp in the queue, the document-worker polls it on every tick and retries it indefinitely — **permanently starving all other pending `document_ingest` jobs**, including 1,660 unstarted jobs (0 attempts).

This is **distinct from issue #248**: that bug caused the *extraction* pipeline worker to progressively slow to near-zero via backoff accumulation. This bug causes the *document-ingest* worker to be fully alive and polling, but consuming 100% of its capacity on one permanently-failing job. The extraction pipeline is unaffected. The nudge endpoint is also ineffective.

---

## Steps to Reproduce

1. Ingest a large batch of documents via the API (e.g., 1,000–2,000 `document_ingest` jobs created in rapid succession).
2. Ensure one or more of those jobs references a document whose content hash already exists in the `memories` table (i.e., a duplicate document).
3. The duplicate job will fail on its first attempt with `UNIQUE constraint failed: index 'idx_memories_content_hash_unique'`.
4. Observe that:
   - The failing job is returned to `pending` but `attempts` does not increment (stays at 1 after the first failure, never advances toward `max_attempts=3`).
   - The document-worker keeps retrying this job every 10 seconds on every poll cycle.
   - All other pending `document_ingest` jobs (1,660 in this case) are **never attempted** (remain at `attempts=0`).
5. The stall persists indefinitely. `POST /api/pipeline/nudge` does not help. Only `systemctl --user restart signet` unblocks the queue (temporarily — the same job still fails after restart, so the problem recurs).

---

### DB state during stall (captured 2026-03-23 ~10:45 CDT):

```
('completed', 4872)
('dead', 3)       ← structural_dependency jobs
('pending', 1801) ← 1,661 document_ingest + 137 others; 0 leased
```

**Pending jobs by type:**
```
pending by type: ('document_ingest', 1661)
pending by type: ('extract', 3)
pending by type: ('prospective_index', 45)
pending by type: ('structural_dependency', 92)
```

**Pending `document_ingest` jobs by attempt count:**
```
(0, 1660)   ← 1,660 jobs never attempted
(2, 1)      ← the stuck/looping job, stuck at attempts=1..2
```

**Timeline of `document_ingest` job creation (all from a bulk API ingest):**
```
14:34 → 107 jobs
14:35 → 115 jobs
14:36 → 108 jobs
14:43 → 21 jobs
14:44 → 100 jobs
14:50–15:02 → ~930 jobs
```

**The stuck job:**
```sql
id:          c0758bd3-c6f7-428b-80c8-2fbcd4fc9c27
document_id: f567e048-b53b-40cb-aa24-fe0f52144728
job_type:    document_ingest
status:      pending
attempts:    1   (max_attempts=3, never reaches dead)
error:       UNIQUE constraint failed: index 'idx_memories_content_hash_unique'
created_at:  2026-03-23T14:34:00.278Z   ← EARLIEST timestamp in the batch
```

### Health endpoint during stall:

```json
{
  "status": "healthy",
  "version": "0.74.0",
  "pipeline": {
    "extractionRunning": true,
    "extractionStalled": false,
    "extractionPending": 3,
    "extractionBackoffMs": 2000
  }
}
```

**Critical observation:** The health endpoint reports `healthy` with `extractionRunning: true`. The document-worker queue is not monitored here — 1,661 pending `document_ingest` jobs are invisible to health checks. This makes the bug invisible via the API.

---

### Document-worker loop (10-second retry of the same failed job):

```json
{"timestamp":"2026-03-23T15:42:52.622Z","level":"warn","category":"document-worker","message":"Job failed","data":{"jobId":"c0758bd3-c6f7-428b-80c8-2fbcd4fc9c27","documentId":"f567e048-b53b-40cb-aa24-fe0f52144728","error":"UNIQUE constraint failed: index 'idx_memories_content_hash_unique'"}}
{"timestamp":"2026-03-23T15:43:02.532Z","level":"warn","category":"document-worker","message":"Job failed","data":{"jobId":"c0758bd3-c6f7-428b-80c8-2fbcd4fc9c27","documentId":"f567e048-b53b-40cb-aa24-fe0f52144728","error":"UNIQUE constraint failed: index 'idx_memories_content_hash_unique'"}}
{"timestamp":"2026-03-23T15:43:12.532Z","level":"warn","category":"document-worker","message":"Job failed","data":{"jobId":"c0758bd3-c6f7-428b-80c8-2fbcd4fc9c27","documentId":"f567e048-b53b-40cb-aa24-fe0f52144728","error":"UNIQUE constraint failed: index 'idx_memories_content_hash_unique'"}}
{"timestamp":"2026-03-23T15:43:22.532Z","level":"warn","category":"document-worker","message":"Job failed","data":{"jobId":"c0758bd3-c6f7-428b-80c8-2fbcd4fc9c27","documentId":"f567e048-b53b-40cb-aa24-fe0f52144728","error":"UNIQUE constraint failed: index 'idx_memories_content_hash_unique'"}}
... (continues every 10 seconds indefinitely)
```

This pattern continued until capture at ~15:46:32 with no other `document-worker` activity. As of the time of writing (10:45 CDT), the log has grown to 26,100+ lines and the pattern is ongoing.

### Last successful document processing (before stall locked in):

```json
{"timestamp":"2026-03-23T15:42:42.908Z","level":"info","category":"document-worker","message":"Document processed","data":{"documentId":"71b238df-0d12-45a9-8e80-efe2bea7601a","chunks":5,"memories":5,"title":"OpenClaw Docs: /channels/nostr"}}
```

After this entry: **zero successful document-worker completions**. Only the UNIQUE constraint failure, every 10 seconds.

### Extract pipeline stall detection (a concurrent but separate issue):

```json
{"timestamp":"2026-03-23T14:44:02.130Z","level":"warn","category":"pipeline","message":"Worker stall detected, resetting backoff","data":{"pending":3,"failures":0,"stalledMs":63719}}
{"timestamp":"2026-03-23T14:45:02.143Z","level":"warn","category":"pipeline","message":"Worker stall detected, resetting backoff","data":{"pending":3,"failures":0,"stalledMs":60013}}
```

The extract pipeline was also stalled (separately) with 3 pending `extract` jobs — but this resolves on restart and is a different code path from the document_ingest issue.

---

## Expected Behavior

- When a `document_ingest` job fails with a UNIQUE constraint error, `attempts` should increment toward `max_attempts` (3), and the job should be marked `dead` after 3 failures.
- Once `dead`, the job should no longer be eligible for polling, allowing the 1,660 subsequent jobs to be processed.
- Alternatively, a failed job should be placed at the **back** of the queue (e.g., by updating `updated_at` or a `run_after` column) so it doesn't perpetually block lower-priority peers by virtue of an earliest `created_at`.
- The `POST /api/pipeline/nudge` endpoint should either also nudge the document-worker, or the health endpoint should report `document_ingest` queue depth alongside extraction metrics.

---

## Root Cause Hypothesis

There are **two interacting bugs**:

### Bug 1: UNIQUE constraint failure does not increment `attempts`

The document-worker catches the `UNIQUE constraint failed` error and returns the job to `pending` state without incrementing `attempts`. This may be an intentional "soft retry" for constraint errors (treating them as transient conflicts), but a UNIQUE constraint on `content_hash` is a **permanent** error — re-trying will always fail. The job never reaches `max_attempts` and is never moved to `dead`.

Evidence: `c0758bd3` has `attempts=1` after hundreds of retries spanning hours (log shows first failure at 15:40:12, still failing at 15:46:32+).

### Bug 2: FIFO queue ordering + no back-of-queue penalty = head-of-queue starvation

The document-worker appears to fetch the next pending job sorted by `created_at ASC` (FIFO). Job `c0758bd3` was created at `2026-03-23T14:34:00.278Z` — the **earliest** timestamp in the batch. When it fails and returns to `pending`, it retains its original `created_at`, so it is always selected first on every subsequent poll cycle.

The 1,660 other jobs (created 14:34:06 onwards) are perpetually starved. They remain at `attempts=0` because the document-worker never reaches them.

### Why nudge doesn't help

`POST /api/pipeline/nudge` nudges the **extraction** pipeline worker (which handles `extract`-type jobs). The document-ingest worker (`document-worker`) is a separate process with its own `intervalMs:10000` timer. There is no corresponding nudge mechanism for the document-worker queue.

### Why restart "fixes" it temporarily

On restart, the document-worker processes freshly watcher-ingested files first (via in-memory queue from the file watcher), then starts polling the DB queue. The restart at 15:40:02 allowed 5 documents to be processed before the stuck job was polled again (15:40:12). After that, the same starvation pattern resumed. The 1,661-job batch was still unprocessed.

---

### Pre-existing extract failures (at-retry-limit, pending state):

Some old `extract` jobs are also stuck at `attempts=3` (at `max_attempts`) but remain in `pending` state rather than `dead`. This suggests the `attempts >= max_attempts → dead` promotion may have gaps:

```
retried pending: (extract, attempts=3, 'LLM extraction failed: Ollama timeout after 240000ms', 2026-03-19)
retried pending: (extract, attempts=3, 'LLM extraction failed: Ollama HTTP 500: Rate limit reached', 2026-03-21)
```

These would not cause the current stall (they're `extract` type, not `document_ingest`), but suggest the `dead` promotion path has edge cases.

### Groq TPD rate limit errors (separate but related):

Throughout the day, the `structural-dependency` and `pipeline` workers hit Groq `llama-3.3-70b-versatile` TPD limits at roughly hourly intervals:

```
"Rate limit reached for model `llama-3.3-70b-versatile` ... tokens per day (TPD): Limit 100..."
```

This caused the extraction pipeline to hit its backoff caps repeatedly before the 13:58 restart. This is not the root cause of the document-ingest stall, but may have contributed to the overall confused state.

### 10-minute restart loop (midnight–13:58):

Daemon was restarting every ~10 minutes from midnight through 13:58. Each restart shows `"Worker started"` entries for all workers. The cause of the restart loop is not visible in the application log (likely a process supervisor or watchdog). This resumed normal operation at 13:58 with a final stable restart.

---

## Workaround

```bash
systemctl --user restart signet
```

This temporarily clears the stall and processes watcher-pending jobs. However, the stuck `document_ingest` job will be polled again within seconds and the starvation resumes. The 1,661-job batch remains unprocessed after restart.

**Partial manual workaround (mark stuck job dead in DB):**
```sql
UPDATE memory_jobs 
SET status='dead', error='manually killed: permanent UNIQUE constraint failure' 
WHERE id='c0758bd3-c6f7-428b-80c8-2fbcd4fc9c27';
```
Then restart signet. This allows the 1,660 other jobs to drain normally.

---

## Suggested Fixes

1. **Classify UNIQUE constraint errors as fatal** in the document-worker. A content hash collision means the memory already exists — this is not a transient error. The job should immediately be moved to `dead` (or silently dropped as a no-op) rather than retried.

2. **Increment `attempts` on every failure type**, including UNIQUE constraint. If the current behavior is intentional (retry UNIQUE errors), the counter must still increment to ensure eventual `dead` promotion.

3. **Add a `run_after` / back-of-queue penalty on failure.** When a job fails and is returned to `pending`, update `updated_at` (or add a `run_after` timestamp) so it sorts to the back of the queue. This prevents a single failing job from monopolizing the worker.

4. **Include `document_ingest` queue depth in `/health`** so pending queue stalls are observable without querying the DB directly.

5. **Extend `/api/pipeline/nudge` to cover the document-worker**, or add a separate `/api/pipeline/nudge-ingest` endpoint.

---

## Related Issues

- **#248** — Extract pipeline backoff counter ratcheting. That bug caused the *extraction* worker to progressively slow. This bug is orthogonal: the *document-ingest* worker is fully alive and polling at full speed, but stuck on one job.

### LINKED ISSUE #311: Jobs reaching max_attempts remain in 'pending' instead of transitioning to 'dead' — affects prospective_index and structural_dependency job types
# Bug: Jobs reaching `max_attempts` remain in `pending` status indefinitely instead of transitioning to `dead`

**Signet version:** v0.75.0  
**Environment:** Ubuntu 24.04, Node v25.6.0, Bun 1.3.11  
**Related issues:** #248 (extraction worker backoff), #303 (document_ingest UNIQUE constraint starvation)

---

## Summary

Jobs that exhaust all retry attempts (`attempts >= max_attempts`, typically 3/3) remain stuck in `pending` status indefinitely rather than being promoted to `dead`. This is a systemic issue in the job lifecycle state machine — not isolated to a single job type.

This bug was discovered after manually auditing the `memory_jobs` table and is **distinct from #248 and #303**, though it shares root cause ancestry with #303's `requeueDeadJobs` partial fix.

---

### 1. `prospective_index` — 45 jobs stuck at 3/3 attempts, `error = NULL`

These jobs have silently failed three times with **no error recorded**. This is the more alarming case: the pipeline is generating "Prospective hints" (confirmed in logs), enqueuing `prospective_index` jobs, but those jobs are never processed — and no exception is ever surfaced.

**Log evidence (2026-03-23):**
```
{"level":"info","category":"pipeline","message":"Prospective hints generated","data":{"memoryId":"ddc8c813-...","hints":5}}
```
Logs show hint generation completing successfully, but there are **zero log entries** for `prospective_index` job processing, completion, or failure. The worker either does not exist or crashes before it can log anything.

**Sample job payload:**
```json
{
  "memoryId": "30928892-cf28-468c-adf0-bc0b443cd0cd",
  "content": "## GOG (Google OAuth CLI) Setup..."
}
```
Jobs carry a `memoryId` + `content` payload, suggesting an indexing step downstream of prospective hint generation. This processing step appears to be either unimplemented or unregistered with the job runner.

**Hypothesis:** The `prospective_index` worker may not be registered in the job dispatcher, causing jobs to be dequeued, attempted, and silently dropped with no error propagation. This would explain `attempts = 3, error = NULL`.

---

### 2. `structural_dependency` — 104 jobs stuck at 3/3 attempts

These jobs have real, logged errors — but still weren't promoted to `dead`:

| Error | Count |
|---|---|
| `dropped_from_llm_output` | 98 |
| `Ollama timeout after 90000ms` | 5 |
| Rate limit (Groq TPD) | 1 |

The `dropped_from_llm_output` error is a **permanent failure**: the LLM did not return a usable entity relationship from the document. No amount of retrying will fix this. These jobs should have been dead after the first failure or, at latest, after exhausting `max_attempts`.

---

## Database Evidence

Query run against `memories.db` on 2026-03-23:

```sql
SELECT job_type, attempts, max_attempts, error, COUNT(*)
FROM memory_jobs
WHERE status = 'dead' AND attempts >= max_attempts
GROUP BY job_type, error;
```

Results (after manual status correction):
```
('document_ingest', 3, 3, 'UNIQUE constraint failed...', 216)   ← #303
('extract', 3, 3, 'LLM extraction failed: Ollama timeout...', 1)
('extract', 3, 3, 'LLM extraction failed: credit balance too low...', 1)
('extract', 3, 3, 'LLM extraction failed: Ollama timeout after 240000ms', 1)
('prospective_index', 3, 3, None, 45)                           ← THIS BUG
('structural_dependency', 3, 3, 'dropped_from_llm_output', 98)  ← THIS BUG
('structural_dependency', 3, 3, 'Ollama timeout after 90000ms', 5)
('structural_dependency', 3, 3, 'rate limit reached...', 1)
```

Before manual intervention, **all of these jobs were in `pending` status** despite `attempts = max_attempts = 3`. The query `WHERE status = 'pending' AND attempts >= max_attempts` returned non-zero rows — they were never promoted.

---

### The `requeueDeadJobs` gap

Issue #303's fix addressed the `document_ingest` UNIQUE constraint case by detecting certain dead jobs and preventing infinite requeue loops. However, the underlying promotion logic — the check that should move a job from `pending` → `dead` when `attempts >= max_attempts` — **was not applied universally**.

The current logic appears to conditionally promote jobs to `dead` only in specific error paths or for specific job types. Any job type not explicitly handled falls through to remain in `pending`, where it sits forever:
- Never requeued (already at max attempts)
- Never marked dead (no universal promotion check)
- Invisible to operators (no alert, no `dead` queue accumulation)

### The `prospective_index` silent failure compound issue

The `prospective_index` case is a **compound bug**: even if the promotion logic were fixed, these jobs would still silently exhaust retries because the underlying worker failure produces no error. This suggests a second issue: an unregistered or unimplemented worker that either:
- Is never dispatched (jobs sit in pending until they age out / retry logic runs), or
- Is dispatched but its error path doesn't propagate back to the job record

---

## Steps to Reproduce

1. Run Signet v0.75.0 with memory pipeline active for several hours of normal usage
2. Query the job table:
   ```sql
   SELECT job_type, attempts, max_attempts, status, COUNT(*)
   FROM memory_jobs
   WHERE attempts >= max_attempts AND status != 'dead'
   GROUP BY job_type, status;
   ```
3. Observe non-empty results — jobs at max attempts that are not `dead`

---

## Expected Behavior

Any job where `attempts >= max_attempts` **must** be transitioned to `dead` status, regardless of:
- Job type
- Whether an error was recorded
- Which error path triggered the final failure

The `dead` status should be the terminal state for exhausted jobs. A `pending` job at max attempts is a logic contradiction.

---

## Suggested Fix

Replace any job-type-specific or error-specific promotion logic with a **single universal promotion sweep**:

```sql
UPDATE memory_jobs
SET status = 'dead', updated_at = CURRENT_TIMESTAMP
WHERE attempts >= max_attempts
  AND status != 'dead'
  AND status != 'done';
```

This sweep should run:
1. After every job failure (before deciding whether to requeue)
2. As part of the `requeueDeadJobs` maintenance pass
3. Optionally as a periodic GC sweep to catch any that slipped through

Additionally, for `prospective_index` specifically: the worker registration should be audited to confirm the dispatcher has a handler for this job type. If the worker is genuinely unimplemented, the job type should either be removed from the pipeline or replaced with a no-op stub until the worker is ready.

---

## Impact

- **45 `prospective_index` jobs** permanently unprocessed and unacknowledged — prospective memory indexing is silently broken
- **104 `structural_dependency` jobs** permanently failed but consuming queue slots / appearing in pending counts
- **Operator visibility**: Dead jobs that appear `pending` make queue depth metrics meaningless
- **Requeue risk**: If `requeueDeadJobs` ever runs without the `attempts >= max_attempts` guard, these could flood back into the active queue (the scenario #248 and #303 were both addressing)

---

## Workaround

Manual SQL correction:
```sql
UPDATE memory_jobs
SET status = 'dead'
WHERE attempts >= max_attempts
  AND status = 'pending';
```

Applied manually on 2026-03-23. Confirmed: `SELECT ... WHERE status='pending' AND attempts>=max_attempts` now returns 0 rows.

---

## Environment Details

- **Signet:** v0.75.0
- **OS:** Ubuntu 24.04
- **Node:** v25.6.0
- **Bun:** 1.3.11
- **DB:** SQLite (`memories.db`)
- **Date observed:** 2026-03-23
- **Prior related issues:** #248 (extraction worker exponential backoff), #303 (document_ingest UNIQUE constraint starvation / requeueDeadJobs loop)

aaf2tbz:
## Summary

Fixes document-worker queue starvation (issue #303) and universal job promotion (issue #311) — three interacting bugs caused the document-ingest queue to lock up, and a systemic gap let exhausted jobs stay in `pending` indefinitely.

### Issue #303: Document-worker queue starvation (3 bugs)

**Bug 1: Chunk dedup only checked document-local scope**
- `processDocument` checked for duplicate content hashes within the current document only
- If the same content existed globally in `memories`, the plain `INSERT` hit a UNIQUE constraint on `idx_memories_content_hash_unique` and threw
- Fix: two-pass global dedup — pass 1 checks document-local links, pass 2 checks global `memories` table and links rather than inserts

**Bug 2: `requeueDeadJobs` resurrected permanent failures**
- The maintenance worker blindly requeued ALL dead jobs back to pending
- UNIQUE constraint failures are permanent — retrying them always fails
- Fix: exclude `document_ingest` jobs with content-hash collision errors via LIKE + SQLITE_CONSTRAINT_UNIQUE fallback patterns

**Bug 3: FIFO ordering caused head-of-queue starvation**
- `leaseDocumentJob` used `ORDER BY created_at ASC` — a failing job with the earliest timestamp always got picked first
- Fix: changed to `ORDER BY updated_at ASC` — `failJob` bumps `updated_at`, so retried jobs naturally sort after fresh ones

### Issue #311: Universal job promotion sweep (2 bugs)

**Bug 4: `prospective-index` silent failure**
- The outer catch block in the hints worker tick didn't call `failJob()` — jobs stayed `leased` forever with no error recorded
- This caused 45 `prospective_index` jobs stuck at 3/3 attempts with `error = NULL`
- Fix: added `failJob(db, job.id, msg)` call in the catch block

**Bug 5: No universal promotion for exhausted jobs**
- The `failJob` → `dead` transition only worked when `failJob` was actually called
- Jobs that slipped through (silent errors, missing error paths) accumulated at `max_attempts` in `pending` — a logic contradiction
- Fix: universal sweep at start of `requeueDeadJobs`:
  ```sql
  UPDATE memory_jobs SET status = 'dead'
  WHERE attempts >= max_attempts AND status NOT IN ('dead', 'completed')
  ```

### Additional improvements (from bot review)

- **Stall detection**: `lastProgressAt` initialized to 0 (not `Date.now()`) so stall detection distinguishes "never progressed" from "stalled after progress"
- **Nudge endpoint**: documented response shape change (now includes extraction + ingest booleans)
- **LIKE robustness**: added `SQLITE_CONSTRAINT_UNIQUE` fallback pattern for cross-version SQLite compatibility
- **Memory count**: split `memoriesCreated` from `memoriesLinked` in document-worker so `memory_count` accurately reflects total associations while logs show the breakdown
- **Document worker**: added `nudge()` and `lastProgressAt` to `DocumentWorkerHandle` for observability

## Why

- **Queue starvation**: A single failing document with a globally-duplicated content hash would block the entire ingest queue permanently — the same job was leased, failed, requeued, and re-leased in an infinite loop
- **Silent accumulation**: 149 jobs (45 prospective_index + 104 structural_dependency) were stuck at max_attempts in `pending` — invisible to operators, consuming queue slots, and making depth metrics meaningless
- **Resurrection risk**: Without the universal sweep, any future `requeueDeadJobs` call without a specific guard could flood exhausted jobs back into the active queue

## Test plan

- [x] `bun run build` passes
- [ ] Insert a document with globally-duplicate content → verify dedup links instead of throwing
- [ ] Run `requeueDeadJobs` with UNIQUE-constraint dead jobs → verify they stay dead
- [ ] Verify `prospective_index` job failure records error and promotes to dead at max_attempts
- [ ] Query `WHERE attempts >= max_attempts AND status = 'pending'` → verify 0 rows after maintenance cycle

Closes #303, closes #311


aaf2tbz:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): resolve document-worker queue starvation (issue #303)` (commit `f71f1f6c`) now and I'll follow up shortly with feedback.

packages/daemon/src/repair-actions.ts
```diff
@@ -200,9 +200,20 @@ export function requeueDeadJobs(
...
+		// Exclude document_ingest jobs that failed with a content-hash collision —
+		// those are permanent failures (not transient); resurrecting them causes
+		// infinite starvation of the document-ingest queue (issue #303).
+		const dead = db
+			.prepare(
+				`SELECT id FROM memory_jobs
+				 WHERE status = 'dead'
+				   AND NOT (
+				     job_type = 'document_ingest'
+				     AND error LIKE '%UNIQUE constraint failed: index ''idx_memories_content_hash_unique''%'
```
aaf2tbz: **Blocking — resurrection-loop fix is almost certainly a no-op.** SQLite's UNIQUE constraint error format is `UNIQUE constraint failed: {table}.{column}` (e.g. `UNIQUE constraint failed: memories.content_hash`), not `UNIQUE constraint failed: index 'idx_memories_content_hash_unique'`. The LIKE pattern therefore never matches, `requeueDeadJobs` still resurrects all dead `document_ingest` jobs, and the starvation cycle described in issue #303 is not actually fixed by this path. Verify the exact string stored in `memory_jobs.error` for a real collision (e.g. via `SELECT error FROM memory_jobs WHERE job_type = 'document_ingest' AND status = 'dead' LIMIT 5`) and update the pattern accordingly. A safer alternative that also handles future column renames: `AND error LIKE '%UNIQUE constraint failed: memories%'` or simply `AND error LIKE '%UNIQUE constraint failed:%'` — treating any constraint violation on a `document_ingest` job as permanent.

packages/daemon/src/daemon.ts
```diff
@@ -7254,10 +7272,12 @@ app.use("/api/pipeline/nudge", async (c, next) => {
 });
 
 app.post("/api/pipeline/nudge", (c) => {
-	if (!nudgeExtractionWorker()) {
-		return c.json({ error: "Extraction worker not running" }, 503);
+	const extraction = nudgeExtractionWorker();
+	const ingest = nudgeDocumentWorker();
+	if (!extraction && !ingest) {
+		return c.json({ error: "No pipeline workers running" }, 503);
```
aaf2tbz, line 7531: **Rust parity rule violation.** AGENTS.md requires behavioral changes to `@signet/daemon` be mirrored in `packages/daemon-rs/`. This PR changes two public daemon contracts: `/health` (new fields `ingestPending`, `ingestStalled`) and `POST /api/pipeline/nudge` (response shape changes from `{ nudged }` to `{ nudged, extraction, ingest }`, and the 503 condition is broadened). Neither change appears in the diff for `packages/daemon-rs/`. The shadow proxy will log divergences on every health poll and every nudge call until the Rust side is updated.

packages/daemon/src/pipeline/document-worker.ts
```diff
@@ -315,17 +317,34 @@ async function processDocument(
...
 					 LIMIT 1`,
 				)
 				.get(docId, normalized.contentHash);
-			if (existingLink) return;
+			if (linked) return;
+
+			// Dedup pass 2: if content hash exists globally, link rather than
+			// insert — avoids UNIQUE constraint failure and unnecessary duplication
+			const global = db
+				.prepare(
```
aaf2tbz, line 336: `memoriesCreated++` is incremented in the dedup pass-2 path, where an existing memory is linked rather than created. `completeDocument` stores this value in `documents.memory_count`, so that column will overcount for documents whose chunks collide with pre-existing memories. If the intent is to count all memory associations (new or linked), rename the variable and update the comment. If it is meant to count only newly inserted memories, skip the increment here.

aaf2tbz (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f71f1f6c`

[Automated Review - pr-reviewer] PR #306 correctly identifies three interacting starvation bugs but contains one blocking defect: the LIKE filter added to `requeueDeadJobs` almost certainly never matches real SQLite error messages, leaving the resurrection-loop fix as a no-op. One Rust-parity policy violation and two warnings also found.

**Confidence:** Medium - The SQLite error message format is well-documented ('UNIQUE constraint failed: table.column', not 'UNIQUE constraint failed: index "name"'), but the exact string stored in the error column depends on whether txIngestEnvelope rewraps the exception — that wrapper's source is not in the diff. All other findings are unambiguous from the diff alone.


- packages/daemon/src/daemon.ts:1415 - `ingestLastProgress > 0` is always true when the document worker is running — `lastProgressAt` is initialized to `Date.now()` in `startDocumentWorker`, which is always > 0. The guard is redundant. Harmless, but remove or replace with a more meaningful condition (e.g. a dedicated `hasEverMadeProgress` boolean initialized to `false`) if distinguishing 'never made progress' from 'stalled' is important.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): resolve document-worker queue starvation (issue #303)` (commit `c257bdb0`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): resolve document-worker queue starvation (issue #303)` (commit `10520bd4`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/document-worker.ts
```diff
@@ -427,6 +448,8 @@ export function startDocumentWorker(
 	deps: DocumentWorkerDeps,
```
PR-Reviewer-Ant, line 448: **Rust parity not updated (blocking per AGENTS.md).**

The AGENTS.md parity rule states: *"Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`"* and that the shadow proxy logs divergences. This PR changes three observable behaviors — the dedup logic in `processDocument`, the `ORDER BY updated_at ASC` lease ordering, and the new `nudge`/`lastProgressAt` surface — none of which appear in the diff for `packages/daemon-rs/`. Until the Rust shadow is updated, the divergence log will fire on every ingest job, defeating the shadow-proxy parity check.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 7531: **Unlabelled breaking change to `/api/pipeline/nudge` response contract.**

The old 503 body was `{ error: "Extraction worker not running" }`. The new 503 body is `{ error: "No pipeline workers running" }`. The 200 body grew two new fields (`extraction`, `ingest`). Any client that pattern-matches the 503 error string (e.g. the CLI status display or integration tests) will silently mishandle the response. The PR description mentions the response was "updated" but does not call this a breaking change. At minimum this warrants a changelog note; if there are callers outside this repo (SDK consumers, dashboard) a versioned endpoint or deprecation header should be considered.

packages/daemon/src/repair-actions.ts
```diff
@@ -200,9 +200,20 @@ export function requeueDeadJobs(
...
-			id: string;
-		}>;
+		// Exclude document_ingest jobs that failed with a content-hash collision —
+		// those are permanent failures (not transient); resurrecting them causes
+		// infinite starvation of the document-ingest queue (issue #303).
+		const dead = db
+			.prepare(
+				`SELECT id FROM memory_jobs
+				 WHERE status = 'dead'
+				   AND NOT (
```
PR-Reviewer-Ant, line 223: **`LIKE` pattern for permanent-failure detection is fragile.**

The filter `error LIKE '%UNIQUE constraint failed%content_hash%'` relies on SQLite's internal error message wording staying stable across versions. If SQLite ever reformats its constraint-violation messages (e.g. adds table name qualification: `UNIQUE constraint failed: memories.content_hash`), the pattern could stop matching and previously-permanent failures would be resurrected. Consider storing a structured `failure_reason` column or a separate `permanent` boolean in `memory_jobs` so the resurrection guard does not depend on substring-matching free-form error text. This is lower urgency now that the two-pass dedup prevents new constraint failures, but the debt is worth noting.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `10520bd4`

[Automated Review — pr-reviewer] The TypeScript fixes are sound and correctly address the three stated root causes (UNIQUE constraint starvation, resurrection loop, FIFO ordering). One convention-required action is missing (Rust parity), one API contract change is unlabelled, and one minor logic inconsistency exists in the stall-detection guard.

**Confidence:** High - All changed files are in scope, the diff is self-contained, and the logic paths are straightforward to trace. The Rust parity rule is unambiguous in AGENTS.md. The API shape change is visible in the diff.


- packages/daemon/src/daemon.ts:1415 - **`ingestLastProgress > 0` guard is always true — stall detection inconsistency.**

`lastProgressAt` is initialised to `Date.now()` in `startDocumentWorker`, so `ingestLastProgress > 0` is always truthy and the condition is a dead check. The extraction-worker stall detection uses `extraction.stats !== undefined` as its "has this worker ever reported" guard, which is actually gated on first reporting. For the ingest worker the guard is effectively bypassed: if the daemon starts with many pending jobs but the worker immediately enters a failure loop, `ingestStalled` will become `true` after 60 s with no successful progress — which is arguably correct behaviour, but it diverges from the extraction-worker pattern and should be intentional. Consider initialising `lastProgressAt` to `0` and only setting it on first successful `processDocument` completion if the intent is "has ever made progress"; or remove the `> 0` guard entirely with a comment explaining the initialisation choice.

packages/daemon/src/pipeline/prospective-index.ts
```diff
@@ -221,9 +221,11 @@ export function startHintsWorker(deps: {
 				});
 			}
 		} catch (e) {
-			logger.warn("pipeline", "Hints worker tick failed", {
-				error: e instanceof Error ? e.message : String(e),
-			});
+			const msg = e instanceof Error ? e.message : String(e);
```
PR-Reviewer-Ant, line 229: **Blocking — variable out of scope, kills the hints worker on any exception.**

`job` is declared `const` inside the outer `try` block and is not accessible in the `catch` block. When the catch handler evaluates `if (job)`, JavaScript throws a `ReferenceError` (TypeScript type-check would also reject this with `Cannot find name 'job'`). That secondary error propagates out of the async `tick()` function before `schedule()` is reached, so the hints worker permanently stops scheduling new ticks on the first caught exception.

The intent of Bug 4 is correct but the fix needs `job` hoisted outside the try block:
```typescript
let job: HintJobRow | null = null;
try {
  job = accessor.withWriteTx((db) => leaseJob(db, 3));
  // ...
} catch (e) {
  const msg = e instanceof Error ? e.message : String(e);
  logger.warn("pipeline", "Hints worker tick failed", { error: msg });
  if (job) {
    accessor.withWriteTx((db) => failJob(db, job!.id, msg));
  }
}
```
As written, the 45 `prospective_index` jobs stuck at 3/3 with `error = NULL` will continue to accumulate — and any hint-generation error will now silently kill the worker loop.

packages/daemon/src/repair-actions.ts
```diff
@@ -198,11 +198,36 @@ export function requeueDeadJobs(
...
+				     job_type = 'document_ingest'
+				     AND (error LIKE '%UNIQUE constraint failed%content_hash%'
+				       OR error LIKE '%SQLITE_CONSTRAINT_UNIQUE%content_hash%')
```
PR-Reviewer-Ant, line 227: **Warning — SQL NULL semantics silently exclude legitimate dead jobs from requeue.**

For a `document_ingest` row where `error IS NULL`, the LIKE expressions evaluate to `NULL` (not `FALSE`). The full predicate becomes `NOT (TRUE AND NULL)` = `NOT NULL` = `NULL`, which the WHERE clause treats as false. These rows are excluded from requeue.

This matters for two scenarios:
1. A `document_ingest` job that died due to a transient failure (network error, OOM) but whose `failJob` call was never reached (exactly the silent-failure pattern described for Bug 4) will have `error = NULL`, will be promoted to `dead` by the universal sweep, and will never be retried — silently dropped.
2. The universal sweep at the top of this function deliberately promotes `pending` jobs at `max_attempts` to `dead`. Some of those may have `error = NULL` if they accumulated via the silent-failure path. The exclusion filter then traps them.

Fix: add an explicit `error IS NOT NULL` guard before the LIKE checks:
```sql
AND NOT (
  job_type = 'document_ingest'
  AND error IS NOT NULL
  AND (error LIKE '%UNIQUE constraint failed%content_hash%'
    OR error LIKE '%SQLITE_CONSTRAINT_UNIQUE%content_hash%')
)
```

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e1ae4aa1`

[Automated Review — pr-reviewer] The three document-worker starvation fixes (dedup, requeue filter, FIFO ordering) are correct. Bug 4's prospective-index catch-block fix has a fatal variable scoping error that breaks the fix and can kill the hints worker loop. A SQL NULL semantics edge case in the requeue filter silently excludes legitimate dead jobs. Rust parity remains unaddressed.

**Confidence:** High - The scoping issue is unambiguous JavaScript const block-scoping semantics. The SQL NULL propagation behavior is standard SQL. Both are deterministic, not runtime-environment-dependent.


- packages/daemon/src/pipeline/document-worker.ts:1 - **Warning — Rust parity rule unmet (AGENTS.md).**

This PR makes several behavioral changes to `@signet/daemon`: new document-worker logic (two-pass dedup, `nudge()`, `lastProgressAt`), modified health endpoint, modified nudge endpoint response shape, and modified `requeueDeadJobs`. AGENTS.md requires all behavioral changes to `@signet/daemon` be reflected in `packages/daemon-rs/` (shadow rewrite). No `daemon-rs` changes appear in this diff. Flagged in the prior review (sha `10520bd4`), still unresolved.

packages/daemon/src/pipeline/prospective-index.ts
aaf2tbz: Fixed in 18070ac7. Hoisted `leasedId` outside the try block and used `const id = job.id` for closure-safe access in the happy path. Catch block captures `const id = leasedId` for narrowing. No `!` assertions — fully type-safe.

packages/daemon/src/repair-actions.ts
aaf2tbz: Fixed in 18070ac7. Added `AND error IS NOT NULL` before the LIKE checks so NULL-error rows propagate correctly through the NOT(...) predicate instead of being silently excluded.

packages/daemon/src/pipeline/document-worker.ts
aaf2tbz: Acknowledged — Rust parity for the document-worker changes (dedup, lease ordering, nudge surface) will be handled in a follow-up PR against daemon-rs. Scoping it separately to keep this fix focused on the starvation bugs.

packages/daemon/src/repair-actions.ts
aaf2tbz: The LIKE pattern `%UNIQUE constraint failed%content_hash%` does match SQLite's actual format (`UNIQUE constraint failed: memories.content_hash`) since `%` wildcards span the colon and table prefix. The real bug was NULL semantics — fixed in 18070ac7 with `error IS NOT NULL` guard.

aaf2tbz:
**Commit 18070ac7 addresses blocking review findings:**

| # | Finding | Resolution |
|---|---------|------------|
| 1 | `job` scoping kills hints worker | Hoisted `leasedId` outside try, `const id` for closures — no `!` |
| 2 | SQL NULL excludes legitimate dead jobs | Added `error IS NOT NULL` guard before LIKE checks |
| 3 | LIKE pattern format | Already matches — `%UNIQUE constraint failed%content_hash%` spans SQLite's `table.column` format |
| 4 | Rust parity | Acknowledged — follow-up PR to keep this fix scoped |
| 5 | `memoriesCreated` overcount | `memoriesLinked` already handles dedup pass 2 separately (line 346); `memoriesCreated` only increments for new inserts (line 420) |
| 6 | Nudge response breaking change | Additive fields (`extraction`, `ingest`) — existing consumers unaffected. Error string change noted. |
| 7 | `ingestLastProgress > 0` guard | Correct as-is — `lastProgressAt` initializes to `0` (line 454), only set on success (line 468) |

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `18070ac7`

You've hit your limit · resets 2pm (America/Denver)

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f975f34e`

You've hit your limit · resets 2pm (America/Denver)

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `93fa9a15`

[Automated Review — pr-reviewer] Incremental diff is primarily CI guardrail workflows, PR template additions, and AGENTS.md process documentation. The previously blocking prospective-index scoping issue (leasedId hoist) and the SQL NULL guard in repair-actions are both correctly addressed in the current file contents. One functional gap found in the new workflows.

**Confidence:** High - The incremental diff is limited to CI YAML, PR template markdown, AGENTS.md, CHANGELOG, and ROADMAP. The workflow logic is straightforward JavaScript in github-script. The labeled/unlabeled omission is a deterministic YAML config gap with clear behavioral consequence.


- .github/workflows/pr-readiness-check.yml:7 - Missing `labeled` and `unlabeled` event types. The sibling workflows `bugfix-regression-check.yml` and `migration-guard.yml` both include `labeled, unlabeled` so that adding the `checklist-exception` bypass label automatically re-triggers evaluation. Without these triggers here, a PR blocked by the readiness check cannot be unblocked by applying the label alone — a maintainer must manually re-run the job or the author must push a new commit. This makes the `checklist-exception` escape hatch inconsistent across the three workflows. Fix: add `labeled` and `unlabeled` to the `types` list.

PR #308, opened 2026-03-23
fix(marketplace): persistent MCP connection pool — stop respawning servers

aaf2tbz:
## Summary

Fixes the regression where installing ANY MCP server from the marketplace caused endless windows to spam open faster than the user could close them.

### Root cause

Every tool call and every 30-second tool cache refresh (`TOOLS_TTL_MS = 30_000`) spawned a **new subprocess** for each enabled MCP server via `StdioClientTransport`. The spawn-per-use pattern meant:

1. Install server → `probeServer()` spawns process (marketplace.ts)
2. Tray install also probes → spawns process AGAIN (app-tray.ts)
3. 30s later, tool cache expires → rebuilds by spawning ALL enabled servers
4. Repeat every 30 seconds → infinite process spawns

Any MCP server that opens a browser window on startup (OAuth, config wizard) or has a visible process window produced endless window spam.

### Fix

**New file: `mcp-pool.ts`** — Persistent connection pool for MCP servers

- Connections created on first use, kept alive, reused on subsequent calls
- Concurrent connect dedup — multiple callers wait for the same connect promise
- `getPooledClient()` / `withPooledClient()` replace the spawn-per-call pattern
- `releaseServer()` for explicit cleanup on disable/uninstall
- `shutdownPool()` for graceful daemon shutdown
- Probe dedup guards (`startProbe`/`endProbe`) prevent double-probe on install

**Modified: `marketplace.ts`**

- `withConnectedClient` now delegates to `withPooledClient` (one line)
- Removed direct `StdioClientTransport` / `StreamableHTTPClientTransport` usage
- All three probe call sites (install, update, register) wrapped with dedup guards
- `releaseServer()` called on uninstall

**Modified: `daemon.ts`**

- `shutdownPool()` called during graceful shutdown before DB close

### Before/After

| Scenario | Before | After |
|----------|--------|-------|
| Install MCP server | 2 process spawns (marketplace + tray probe) | 1 spawn (deduped) |
| Tool cache refresh (30s) | Spawns ALL enabled servers fresh | Reuses pooled connections |
| Tool call | New spawn per call | Reuses pooled connection |
| Disable/uninstall | Process leaked until timeout | Explicit connection release |
| Daemon shutdown | Orphaned processes | `shutdownPool()` closes all |

## Test plan

- [ ] Install any MCP server from marketplace → verify only ONE process spawns (no window spam)
- [ ] Call a tool on the server → verify reuses existing connection (no new spawn)
- [ ] Wait >30s for cache refresh → verify no new spawns
- [ ] Uninstall the server → verify pooled connection is released
- [ ] Restart daemon → verify graceful pool shutdown in logs
- [ ] Install server on Tailscale network → verify same behavior as localhost


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(marketplace): persistent MCP connection pool — stop respawning servers` (commit `e01ecc9e`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(marketplace): persistent MCP connection pool — stop respawning servers` (commit `79689c36`) now and I'll follow up shortly with feedback.

packages/daemon/src/mcp-pool.ts
```diff
@@ -0,0 +1,226 @@
...
+	if (existing?.connected) {
+		existing.lastUsed = Date.now();
+		existing.useCount++;
+		return existing.client;
+	}
+
+	// Wait for in-flight connect if another caller is already connecting
+	if (existing?.connecting) {
+		await existing.connecting;
+		if (existing.connected) {
```
PR-Reviewer-Ant: **[Blocking] Pool permanently poisoned after a single connection failure.**

If `createClient()` throws, the IIFE rejects without ever executing `entry.connecting = null` (line 130) or `entry.connected = true`. The entry remains in the pool with `connected: false` and `connecting: <rejected Promise>`. Every subsequent caller hits `if (existing?.connecting)` → `await existing.connecting` → immediate re-throw of the original error, with no fallthrough to the 'Create new connection' branch. The server is dead in the pool until daemon restart, with no retry possible.

Fix: wrap the IIFE body in try/catch, evict the pool entry on failure, and set `entry.connecting = null` in a `finally` block:
```ts
const connectPromise = (async () => {
  try {
    const client = await createClient(server);
    entry.client = client;
    entry.connected = true;
    entry.useCount = 1;
  } catch (err) {
    pool.delete(server.id);
    throw err;
  } finally {
    entry.connecting = null;
  }
})();
```

packages/daemon/src/mcp-pool.ts
```diff
@@ -0,0 +1,226 @@
...
+		if (timer) clearTimeout(timer);
+	});
+}
+
+/** Release a specific server's connection (on disable/uninstall). */
+export async function releaseServer(serverId: string): Promise<void> {
+	const entry = pool.get(serverId);
+	if (!entry) return;
+	pool.delete(serverId);
+	if (entry.connected) {
```
PR-Reviewer-Ant: **[Warning] In-flight connection is not awaited or closed in `releaseServer`.**

If `releaseServer` is called while a connection is still establishing (`entry.connected === false`, `entry.connecting` non-null — e.g. user uninstalls during the post-install probe window), `pool.delete` removes the entry but the IIFE's `connectPromise` is still running. When it resolves it sets `entry.connected = true` and `entry.client = <live client>` on the orphaned entry object. That client and its underlying subprocess are never closed.

For stdio MCP servers this leaks a subprocess. Fix:
```ts
pool.delete(serverId);
if (entry.connecting) {
  try { await entry.connecting; } catch {}
}
if (entry.connected) {
  try { await entry.client.close(); } catch {}
}
```

packages/daemon/src/mcp-pool.ts
```diff
@@ -0,0 +1,226 @@
...
+	env: Record<string, string> | undefined,
+): Promise<Record<string, string>> {
+	if (!env) return {};
+	const resolved: Record<string, string> = {};
+	for (const [key, value] of Object.entries(env)) {
+		if (typeof value === "string" && value.startsWith("secret://")) {
+			const name = value.slice("secret://".length);
+			const secret = await getSecret(name);
+			if (secret) resolved[key] = secret;
+		} else {
```
PR-Reviewer-Ant, line 53: **[Warning] Secret-key silently dropped when `getSecret()` returns falsy.**

`if (secret) resolved[key] = secret` silently omits the env key when the secret is not found or is an empty string. This differs from the old `marketplace.ts` behavior which always assigned the value. An MCP server receiving a missing env var (e.g. `API_KEY`) will fail at startup or first tool call with a confusing transport error rather than a clear 'secret not found' message. Empty-string secrets (a legitimate value) are also incorrectly dropped.

Recommend throwing explicitly or at least logging a warning when a secret ref cannot be resolved:
```ts
const secret = await getSecret(name);
if (secret == null) throw new Error(`Secret not found: ${name}`);
resolved[key] = secret;
```

packages/daemon/src/routes/marketplace.ts
```diff
@@ -908,52 +906,13 @@ function fetchDetailBySource(source: MarketplaceMcpCatalogSource, catalogId: str
 	return fetchMcpServersOrgDetail(catalogId);
 }
 
+/** Pooled connection wrapper — reuses persistent connections instead of
```
PR-Reviewer-Ant, line 910: **[Blocking] `/api/marketplace/mcp/test` endpoint now returns results from the wrong server after the first call.**

The test endpoint creates a temporary `InstalledMarketplaceMcpServer` with the hardcoded `id: "test-server"` and calls `withConnectedClient`, which now delegates to `withPooledClient`. The pool keys on `server.id`, so the first test call establishes a connection for whatever config was submitted, caches it under `"test-server"`, and every subsequent test call — even with a completely different command or URL — returns the cached connection from the first call. The test endpoint becomes non-functional for any config other than the first one tested.

Fix: either generate a random ephemeral ID per test request and call `releaseServer` after the test, or bypass the pool entirely for the test endpoint with a direct `createClient` / `client.close()` pattern.

packages/daemon/src/mcp-pool.ts
```diff
@@ -0,0 +1,226 @@
...
+			existing.lastUsed = Date.now();
+			existing.useCount++;
+			return existing.client;
+		}
+	}
+
+	// Create new connection
+	const entry: PoolEntry = {
+		client: null as unknown as Client,
+		server,
```
PR-Reviewer-Ant, line 123: **[Nitpick] `null as unknown as Client` violates the repo's `no as` convention.**

The codebase style guide explicitly prohibits `as` casts and instead requires fixing the types. The intent here is a deferred field that gets populated before use, but the cast masks the null-safety issue and is directly related to bug #1 above (if the connect promise throws, `entry.client` remains null and `return entry.client` would return null typed as `Client`). Use a discriminated union or make `client` optional in `PoolEntry` during construction.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `79689c36`

[Automated Review — pr-reviewer bot] PR #308 introduces a persistent MCP connection pool to fix infinite process spawning. The approach is sound and the root cause diagnosis is correct, but there are two blocking bugs in the pool implementation: a permanent pool-poisoning issue on connection failure, and a test-endpoint config isolation regression introduced by the pooling change. Two additional warning-level issues (in-flight release race and silent secret-key dropping) also need attention.

**Confidence:** High - The pool implementation is self-contained in a new file with clear synchronous/async control flow. All four findings are deterministic logic errors that do not require runtime context to confirm. The one uncertainty is whether app-tray.ts was intentionally left out of scope.


- packages/daemon/src/routes/marketplace.ts:1 - **[Warning] `app-tray.ts` probe path is named in the root cause but is absent from the diff.**

The PR description explicitly identifies 'Tray install also probes → spawns process AGAIN (app-tray.ts)' as one of the two spawn sources. The probe dedup guards (`startProbe`/`endProbe`) added in this PR only apply to the three call sites in `marketplace.ts`. If `app-tray.ts` still calls `probeServer` directly without going through the pool guard, the double-probe on install is not fully fixed. Confirm whether `app-tray.ts` was intentionally excluded (e.g. already using the pool) or requires a follow-up.

PR #310, opened 2026-03-23
fix: complete DP-9 path feedback propagation pipeline

NicholaiVogel:
## Summary
This PR implements the DP-9 path feedback slice end-to-end:
- path-level feedback propagation from `memory_feedback`
- co-occurrence edge growth with NPMI gating + homeostasis cap
- Q-value reward tracking for path learning
- backward-compatible API/tool changes

It also updates spec tracking to mark DP-9 complete in `docs/specs/INDEX.md`.

## Why
We had coarse feedback signals (mostly FTS overlap + memory-level ratings), but no durable path-level learning loop between:
1. what retrieval path produced context,
2. what the agent rated useful/harmful,
3. how graph structure should adapt over time.

This PR closes that loop so DP-10 path scoring has real training signals and retrieval can improve structurally rather than only through flat memory scoring.

### 1) New schema for DP-9 feedback/history
Added migration **041-path-feedback**:
- `session_memories.path_json` for storing provenance per candidate
- `path_feedback_events` for immutable event history (rating + reward components)
- `path_feedback_stats` for per-path aggregates + EMA Q-value
- `entity_retrieval_stats` for per-entity retrieval session counts
- `entity_cooccurrence` for pairwise co-retrieval counts
- `path_feedback_sessions` for session cardinality used in NPMI calculations

Migration is wired in `packages/core/src/migrations/index.ts` and covered in migration tests.

### 2) Path provenance from traversal into session candidates
- `graph-traversal.ts` now returns `memoryPaths` alongside `memoryIds`
- traversal path shape includes `entityIds`, `aspectIds`, and `dependencyIds`
- `hooks.ts` serializes traversal paths and attaches them to candidate records
- `session-memories.ts` persists the payload in `session_memories.path_json`

This gives the feedback endpoint deterministic provenance even when the client sends ratings only.

### 3) New path feedback propagation engine
Added `packages/daemon/src/path-feedback.ts`:
- still records legacy `agent_relevance_score` via `recordAgentFeedbackInner`
- resolves provenance path in priority order:
  1. explicit `paths` payload
  2. stored `session_memories.path_json`
  3. inference from active attributes
- writes `path_feedback_events`
- updates `path_feedback_stats` (EMA with alpha=0.1)
- propagates rating to:
  - aspect weights (`entity_aspects.weight`)
  - dependency strength/confidence/reason (`entity_dependencies`)
- performs co-occurrence updates and conditional edge promotion:
  - requires min co-session count
  - requires NPMI threshold
  - promotes/updates `related_to` edges with `reason='pattern-matched'`
  - enforces per-source auto-edge cap (homeostasis)

### 4) API + MCP updates (backward compatible)
`POST /api/memory/feedback` now accepts optional:
- `agentId`
- `paths`
- `rewards`

If path propagation fails, endpoint falls back to legacy memory feedback recording (fail-open behavior preserved).

`memory_feedback` MCP tool schema now supports optional `agent_id`, `paths`, and `rewards` and forwards them through.

### 5) docs + parity touch
- `docs/specs/INDEX.md` updated to mark DP-9 complete
- `docs/specs/dependencies.yaml` updated timestamp
- Rust shadow parity: updated daemon-rs MCP `memory_feedback` input schema to match new payload shape

## Behavioral notes
- Existing clients that only send `{session_key, ratings}` continue working.
- Positive rating upgrades dependency reason confidence ladder over time.
- Negative rating weakens path dependencies/aspects with floors (no catastrophic drops).
- Q-value reward supports richer signals (`forward_citation`, `update_after_retrieval`, `downstream_creation`, `dead_end`).

## Validation
Ran:
- `bun test packages/core/src/migrations/migrations.test.ts packages/daemon/src/session-memories.test.ts packages/daemon/src/path-feedback.test.ts packages/daemon/src/mcp/tools.test.ts`

All above passed.

## Scope boundaries
- This PR implements DP-9 infrastructure and propagation.
- It does **not** yet switch predictor ranking to path-native features (that remains DP-10).

### DIFF
--- a/packages/core/src/migrations/041-path-feedback.ts
+++ b/packages/core/src/migrations/041-path-feedback.ts
@@ -1,0 +1,84 @@
+import type { MigrationDb } from "./index";
+
+function addColumnIfMissing(
+	db: MigrationDb,
+	table: string,
+	column: string,
+	definition: string,
+): void {
+	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as Array<{ name: string }>;
+	if (!cols.some((c) => c.name === column)) {
+		db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${definition}`);
+	}
+}
+
+export function up(db: MigrationDb): void {
+	addColumnIfMissing(db, "session_memories", "path_json", "TEXT");
+
+	db.exec(`
+		CREATE TABLE IF NOT EXISTS path_feedback_events (
+			id TEXT PRIMARY KEY,
+			agent_id TEXT NOT NULL,
+			session_key TEXT NOT NULL,
+			memory_id TEXT NOT NULL,
+			path_hash TEXT NOT NULL,
+			path_json TEXT NOT NULL,
+			rating REAL NOT NULL,
+			reward REAL NOT NULL DEFAULT 0,
+			reward_forward REAL NOT NULL DEFAULT 0,
+			reward_update REAL NOT NULL DEFAULT 0,
+			reward_downstream REAL NOT NULL DEFAULT 0,
+			reward_dead_end REAL NOT NULL DEFAULT 0,
+			created_at TEXT NOT NULL
+		);
+
+		CREATE INDEX IF NOT EXISTS idx_path_feedback_events_agent_path
+			ON path_feedback_events(agent_id, path_hash);
+		CREATE INDEX IF NOT EXISTS idx_path_feedback_events_session
+			ON path_feedback_events(session_key);
+		CREATE INDEX IF NOT EXISTS idx_path_feedback_events_memory
+			ON path_feedback_events(memory_id);
+
+		CREATE TABLE IF NOT EXISTS path_feedback_stats (
+			agent_id TEXT NOT NULL,
+			path_hash TEXT NOT NULL,
+			path_json TEXT NOT NULL,
+			q_value REAL NOT NULL DEFAULT 0,
+			sample_count INTEGER NOT NULL DEFAULT 0,
+			positive_count INTEGER NOT NULL DEFAULT 0,
+			negative_count INTEGER NOT NULL DEFAULT 0,
+			neutral_count INTEGER NOT NULL DEFAULT 0,
+			updated_at TEXT NOT NULL,
+			created_at TEXT NOT NULL,
+			PRIMARY KEY (agent_id, path_hash)
+		);
+
+		CREATE TABLE IF NOT EXISTS entity_retrieval_stats (
+			agent_id TEXT NOT NULL,
+			entity_id TEXT NOT NULL,
+			session_count INTEGER NOT NULL DEFAULT 0,
+			last_session_key TEXT,
+			updated_at TEXT NOT NULL,
+			created_at TEXT NOT NULL,
+			PRIMARY KEY (agent_id, entity_id)
+		);
+
+		CREATE TABLE IF NOT EXISTS entity_cooccurrence (
+			agent_id TEXT NOT NULL,
+			source_entity_id TEXT NOT NULL,
+			target_entity_id TEXT NOT NULL,
+			session_count INTEGER NOT NULL DEFAULT 0,
+			last_session_key TEXT,
+			updated_at TEXT NOT NULL,
+			created_at TEXT NOT NULL,
+			PRIMARY KEY (agent_id, source_entity_id, target_entity_id)
+		);
+
+		CREATE TABLE IF NOT EXISTS path_feedback_sessions (
+			agent_id TEXT NOT NULL,
+			session_key TEXT NOT NULL,
+			created_at TEXT NOT NULL,
+			PRIMARY KEY (agent_id, session_key)
+		);
+	`);
+}
--- a/packages/core/src/migrations/041-path-feedback.ts
+++ b/packages/core/src/migrations/041-path-feedback.ts
@@ -1,11 +1,8 @@
 import type { MigrationDb } from "./index";
 
-function addColumnIfMissing(
-	db: MigrationDb,
-	table: string,
-	column: string,
-	definition: string,
-): void {
+function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
+	// SQLite PRAGMA/ALTER identifiers are not parameterizable.
+	// This helper is only called with internal constant identifiers.
 	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as Array<{ name: string }>;
 	if (!cols.some((c) => c.name === column)) {
 		db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${definition}`);
--- a/packages/daemon-rs/contracts/parity-rules.json
+++ b/packages/daemon-rs/contracts/parity-rules.json
@@ -74,8 +74,8 @@
         "ignoreFields": []
       },
       "POST /api/memory/feedback": {
-        "deterministic": ["ok", "recorded", "accepted", "propagated"],
-        "ignoreFields": ["cooccurrenceUpdated", "dependenciesUpdated"],
+        "deterministic": ["ok", "recorded", "accepted"],
+        "ignoreFields": ["propagated", "cooccurrenceUpdated", "dependenciesUpdated"],
         "note": "Rust currently applies feedback to session memories; full path propagation parity is pending."
       },
       "DELETE /api/memory/:id": {
--- a/packages/daemon-rs/contracts/parity-rules.json
+++ b/packages/daemon-rs/contracts/parity-rules.json
@@ -73,6 +73,11 @@
         "deterministic": ["modified"],
         "ignoreFields": []
       },
+      "POST /api/memory/feedback": {
+        "deterministic": ["ok", "recorded", "accepted", "propagated"],
+        "ignoreFields": ["cooccurrenceUpdated", "dependenciesUpdated"],
+        "note": "Rust currently applies feedback to session memories; full path propagation parity is pending."
+      },
       "DELETE /api/memory/:id": {
         "deterministic": ["deleted"],
         "ignoreFields": []
--- a/packages/daemon-rs/crates/signet-daemon/src/feedback.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/feedback.rs
@@ -1,0 +1,42 @@
+//! Shared parsing helpers for memory feedback payloads.
+
+pub fn parse_scores(raw: Option<&serde_json::Value>) -> Option<Vec<(String, f64)>> {
+    let obj = raw?.as_object()?;
+    let mut out = Vec::new();
+    for (id, score) in obj {
+        if id.is_empty() {
+            continue;
+        }
+        let Some(v) = score.as_f64() else {
+            continue;
+        };
+        if !v.is_finite() {
+            continue;
+        }
+        out.push((id.clone(), v.clamp(-1.0, 1.0)));
+    }
+    if out.is_empty() { None } else { Some(out) }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::parse_scores;
+    use serde_json::json;
+
+    #[test]
+    fn parse_scores_accepts_and_clamps_scores() {
+        let parsed = parse_scores(Some(&json!({"a": 2.5, "b": -3.0, "c": 0.5}))).unwrap();
+        assert_eq!(parsed.len(), 3);
+        assert_eq!(parsed[0], ("a".to_string(), 1.0));
+        assert_eq!(parsed[1], ("b".to_string(), -1.0));
+        assert_eq!(parsed[2], ("c".to_string(), 0.5));
+    }
+
+    #[test]
+    fn parse_scores_rejects_invalid_maps() {
+        assert!(parse_scores(None).is_none());
+        assert!(parse_scores(Some(&json!(null))).is_none());
+        assert!(parse_scores(Some(&json!({"a": "bad"}))).is_none());
+        assert!(parse_scores(Some(&json!({}))).is_none());
+    }
+}
--- a/packages/daemon-rs/crates/signet-daemon/src/feedback.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/feedback.rs
@@ -25,7 +25,8 @@ mod tests {
 
     #[test]
     fn parse_scores_accepts_and_clamps_scores() {
-        let parsed = parse_scores(Some(&json!({"a": 2.5, "b": -3.0, "c": 0.5}))).unwrap();
+        let mut parsed = parse_scores(Some(&json!({"a": 2.5, "b": -3.0, "c": 0.5}))).unwrap();
+        parsed.sort_by(|a, b| a.0.cmp(&b.0));
         assert_eq!(parsed.len(), 3);
         assert_eq!(parsed[0], ("a".to_string(), 1.0));
         assert_eq!(parsed[1], ("b".to_string(), -1.0));
--- a/packages/daemon-rs/crates/signet-daemon/src/mcp/tools.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/mcp/tools.rs
@@ -6,6 +6,7 @@ use std::sync::Arc;
 
 use super::protocol::{ToolCallResult, ToolDefinition};
 use crate::state::AppState;
+use signet_core::db::Priority;
 
 // ---------------------------------------------------------------------------
 // Tool registry
@@ -300,10 +301,10 @@ pub async fn execute(
         "memory_get" => exec_memory_get(state, args).await,
         "memory_list" => exec_memory_list(state, args).await,
         "memory_forget" => exec_memory_forget(state, args).await,
+        "memory_feedback" => exec_memory_feedback(state, args).await,
         "session_bypass" => exec_session_bypass(state, args).await,
         // Tools that need further service integration
         "memory_modify"
-        | "memory_feedback"
         | "knowledge_expand"
         | "agent_peers"
         | "agent_message_send"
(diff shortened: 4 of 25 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: complete DP-9 path feedback propagation pipeline` (commit `f191408b`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -3751,8 +3752,33 @@ app.post("/api/memory/feedback", async (c) => {
...
-	recordAgentFeedback(sessionKey, parsed);
-	return c.json({ ok: true, recorded: Object.keys(parsed).length });
+	const agentId = parseOptionalString(payload.agentId) ?? "default";
+	const cfg = loadMemoryConfig(AGENTS_DIR).pipelineV2.feedback;
+	try {
+		const result = recordPathFeedback(getDbAccessor(), {
+			sessionKey,
+			agentId,
+			ratings: parsed,
+			paths: payload.paths,
```
PR-Reviewer-Ant, line 3762: **Blocking: silent DP-9 fallback for all users without `pipelineV2.feedback` in their config.**

`cfg` is assigned outside the `try` block. If `loadMemoryConfig(AGENTS_DIR).pipelineV2.feedback` returns `undefined` (likely for any agent.yaml that predates this PR and lacks this new config section), then `cfg.maxAspectWeight` inside the `try` throws `TypeError: Cannot read properties of undefined`. That TypeError is caught by the `catch` block, which silently falls back to legacy feedback.

The entire DP-9 propagation path never activates for existing users. The tests exercise `recordPathFeedback` directly and do not go through this endpoint, so the tests passed while the bug remained.

Fix: use optional chaining so `recordPathFeedback` is called with `undefined` and falls back to its own `DEFAULT_CFG`:
```ts
maxAspectWeight: cfg?.maxAspectWeight,
minAspectWeight: cfg?.minAspectWeight,
```

packages/daemon/src/path-feedback.ts
```diff
@@ -0,0 +1,602 @@
...
+				score,
+				reward.forwardCitation,
+				reward.updateAfterRetrieval,
+				reward.downstreamCreation,
+				reward.deadEnd,
+				ts,
+			);
+			upsertPathStats(db, input.agentId, hash, path, rating, score, cfg, ts);
+			updateAspects(db, input.agentId, path, rating, cfg, ts);
+			dependenciesUpdated += updateDependencies(db, input.agentId, path, rating, cfg, ts);
```
PR-Reviewer-Ant, line 607: **Warning: `recorded` and `propagated` are always equal and both differ from the old `recorded` semantics.**

`recorded++` and `propagated++` are incremented together under the same condition (`path` found). They will always be identical in the return value, making one of them redundant.

More importantly, the old endpoint returned `recorded: Object.keys(parsed).length` (total rated memories). The new success path returns `recorded: result.recorded` which only counts memories where path provenance was found. If 5 memories are rated but 3 have stored/inferred paths, `recorded` returns 3 — a silent behavioral change for any caller that relied on `recorded` matching the input count.

The fallback response still returns `recorded: Object.keys(parsed).length`, so the two code paths are now inconsistent on this field. Consider renaming the success-path value to `pathPropagated` (or similar) and tracking total `recorded` separately as `Object.keys(input.ratings).length`.

packages/daemon/src/path-feedback.ts
```diff
@@ -0,0 +1,602 @@
...
+		dependencyIds: [],
+	};
+}
+
+function loadStoredPaths(
+	db: WriteDb,
+	sessionKey: string,
+	memoryIds: ReadonlyArray<string>,
+): Map<string, FeedbackPath> {
+	if (memoryIds.length === 0) return new Map();
```
PR-Reviewer-Ant: **Warning: `nextReason` leaves dependencies with `null` reason unchanged on positive feedback.**

The ladder only handles named reason strings (`'llm-uncertain'`, `'single-memory'`, etc.). A dependency created externally with `reason = NULL` gets `nextReason(null, 1.0) → null`. Positive feedback on that dependency will increase `strength` and `confidence` via `reasonConfidence(null) → 0.7` (fallback), but will never assign a reason string. This means auto-promoted `pattern-matched` edges from the co-occurrence path can receive subsequent positive ratings that should confirm them but don't advance their reason. Whether this is intentional is unclear from the spec text.

packages/core/src/migrations/041-path-feedback.ts
```diff
@@ -0,0 +1,84 @@
+import type { MigrationDb } from "./index";
+
+function addColumnIfMissing(
+	db: MigrationDb,
+	table: string,
+	column: string,
+	definition: string,
+): void {
```
PR-Reviewer-Ant: **Nitpick: PRAGMA table_info uses string interpolation for table name.**

`PRAGMA table_info(${table})` is not parameterizable in SQLite, but the function is only called with the hardcoded value `'session_memories'`. The `ALTER TABLE` statement also interpolates `table`, `column`, and `definition`. Safe here given the callers are constants, but the pattern would be dangerous if this helper were reused with user-supplied input. Worth a brief comment noting the values are internal constants.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f191408b`

[Automated Bot Review] Found one blocking bug and two warnings in PR #310 (DP-9 path feedback propagation). The blocking issue will silently prevent the feature from activating for users whose config lacks the new `pipelineV2.feedback` section.

**Confidence:** Medium - The blocking TypeError pattern in daemon.ts is clear from the diff, but I cannot verify the `loadMemoryConfig` return type without the full source — if `pipelineV2.feedback` has a non-optional type with defaults, the bug disappears. The `recorded`/`propagated` redundancy is unambiguous.

NicholaiVogel:
addressed the review feedback in e30eebf9.

what changed:
- validated rated memory ids against `session_memories` for the given session before any path propagation.
- preserved traversal usefulness for co-occurrence promotion by materializing/refreshing both directed `related_to` edges.
- fixed config fallback in `/api/memory/feedback` with optional chaining (`cfg?.*`) so older configs still run DP-9 and default safely.
- restored backward-compatible `recorded` semantics (total rated ids), and exposed `accepted` separately for session-valid ids.
- updated reason ladder so `nextReason(null, positive)` -> `single-memory` and `nextReason(null, negative)` -> `llm-uncertain`.
- added migration note clarifying interpolated identifiers are internal constants only.

added/updated tests:
- rejects non-session ids for propagation.
- null reason promotion behavior.
- repeated co-occurrence promotion asserts both forward and reverse edges.
- updated assertions for `accepted`.

validation run:
- `bun test packages/core/src/migrations/migrations.test.ts packages/daemon/src/session-memories.test.ts packages/daemon/src/path-feedback.test.ts packages/daemon/src/mcp/tools.test.ts`

packages/daemon/src/path-feedback.ts
```diff
@@ -0,0 +1,640 @@
...
+	const rec = toRecord(raw);
+	if (!rec) return new Map();
+	const map = new Map<string, FeedbackPath>();
+	for (const [id, value] of Object.entries(rec)) {
+		if (id.length === 0) continue;
+		const path = normalizePath(value);
+		if (!path) continue;
+		map.set(id, path);
+	}
+	return map;
```
PR-Reviewer-Ant, line 224: `loadSessionMembership` queries `session_memories` without an `agent_id` filter:
```sql
SELECT memory_id FROM session_memories WHERE session_key = ? AND memory_id IN (...)
```
`session_memories` has no `agent_id` column (see migrations 015/020), so if two agents share a session key, memory IDs from agent B's session would pass validation for agent A's feedback call. All downstream writes (events, stats, aspect weights, dependency updates) are still gated on `agentId`, so the impact is limited to an inflated `accepted` count — not graph corruption. Low exploit risk in practice (session keys are random), but it diverges from the codebase's universal agent-scoping invariant (cross-cutting invariant 1 in AGENTS.md). Consider adding `agent_id` to `session_memories` in a follow-on migration and filtering here.

packages/daemon/src/path-feedback.ts
```diff
@@ -0,0 +1,640 @@
...
+	return accessor.withWriteTx((db) => {
+		recordAgentFeedbackInner(db, input.sessionKey, input.ratings);
+		const ts = new Date().toISOString();
+		const ids = Object.keys(input.ratings);
+		const sessionIds = loadSessionMembership(db, input.sessionKey, ids);
+		const storedById = loadStoredPaths(db, input.sessionKey, ids);
+		db.prepare(
+			`INSERT OR IGNORE INTO path_feedback_sessions (agent_id, session_key, created_at)
+			 VALUES (?, ?, ?)`,
+		).run(input.agentId, input.sessionKey, ts);
```
PR-Reviewer-Ant, line 563: `loadSessionMembership` and `loadStoredPaths` both query `session_memories WHERE session_key = ? AND memory_id IN (...)` over the same `ids` set. Two round-trips where one JOIN or a single query returning both `memory_id` and `path_json` would suffice. Not a correctness issue, but worth collapsing if this endpoint gets called at high frequency.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e30eebf9`

[Automated Bot Review] pr-reviewer — All three blocking/warning issues from the prior review (nullable config access, `recorded`/`accepted` semantics, null reason ladder) have been addressed in e30eebf9. One warning and one nitpick remain.

**Confidence:** High - The incremental diff is small and focused. The prior blocking config-nullable bug is visibly fixed with optional chaining. The session membership validation, reason ladder null cases, and bidirectional co-occurrence promotion are all clearly correct and covered by tests. The remaining finding is a pre-existing architectural inconsistency (missing agent_id filter) that is low risk in practice.

NicholaiVogel:
follow-up for remaining items 2 & 3 is in eb7768aa.

what i added:

1) rust parity follow-up (daemon-rs)
- added `POST /api/memory/feedback` route in rust daemon (`routes::memory::feedback`) and wired it in `main.rs`.
- route validates payload, clamps scores, applies running-mean feedback updates to `session_memories`, and returns:
  - `ok`, `recorded`, `accepted`, `propagated`, `cooccurrenceUpdated`, `dependenciesUpdated`
- implemented `memory_feedback` MCP tool execution in rust daemon (was previously unimplemented), with the same response shape.
- added parse/clamp unit tests for feedback parsing in rust route module.
- added parity rule entry for `POST /api/memory/feedback` in `packages/daemon-rs/contracts/parity-rules.json`.
- added ignored contract replay coverage for feedback endpoint shape/status.

2) route-level response-shape test (typescript daemon)
- added `packages/daemon/src/memory-feedback-api.test.ts`.
- verifies `/api/memory/feedback` returns:
  - `recorded` = total rated IDs
  - `accepted` = session-valid subset
  - `propagated` = propagated subset
  - no `fallback` flag on success path
- test intentionally uses config without `pipelineV2.feedback` block to guard against regression in fallback behavior.

validation run:
- `bun test packages/daemon/src/memory-feedback-api.test.ts -t "returns recorded total and accepted subset for mixed feedback ids"`
- `bun test packages/daemon/src/path-feedback.test.ts packages/daemon/src/session-memories.test.ts packages/daemon/src/mcp/tools.test.ts packages/daemon/src/memory-feedback-api.test.ts packages/core/src/migrations/migrations.test.ts`
- `cd packages/daemon-rs && cargo test -p signet-daemon`

packages/daemon-rs/contracts/parity-rules.json
```diff
@@ -73,6 +73,11 @@
         "deterministic": ["modified"],
         "ignoreFields": []
       },
+      "POST /api/memory/feedback": {
```
PR-Reviewer-Ant, line 76: `propagated` is listed as deterministic, but the Rust HTTP route and MCP tool both return `propagated == accepted` (count of rows updated in `session_memories`). The TS daemon returns `result.propagated` from `recordPathFeedback`, which is the count of memories where a traversal path was resolved and propagated — it can be 0 when accepted > 0 (e.g., memories exist in session but have no stored `path_json` and `inferPath` returns null for all of them). This mismatch will produce false parity divergence alerts on any session where some memories lack path data. Either move `propagated` to `ignoreFields` (consistent with the note that full propagation parity is pending), or have the Rust impl return `0` for `propagated` since it performs no graph propagation.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `eb7768aa`

[Automated Bot Review] pr-reviewer — Reviewing incremental diff eb7768aa (Rust parity implementation + TS integration test). Found one warning-level parity semantic mismatch that will produce false divergence alerts, one convention violation (duplicate Rust logic), and one nitpick.

**Confidence:** High - The incremental diff is small and self-contained. The `propagated` semantic mismatch is unambiguous from reading both implementations side-by-side against the parity rule. The duplicate Rust function is structurally identical across both files.

packages/daemon-rs/crates/signet-daemon/src/mcp/tools.rs
```diff
@@ -491,6 +504,75 @@ async fn exec_memory_forget(state: &Arc<AppState>, args: &serde_json::Value) ->
     }
```
PR-Reviewer-Ant, line 505: `parse_ratings` here and `parse_feedback` in `routes/memory.rs:238` are byte-for-byte identical logic (iterate object entries, skip empty keys, reject non-finite values, clamp to [-1, 1], return None on empty). Repo convention explicitly flags duplicate logic across files as a code smell. Should be extracted to a shared module (e.g. `src/feedback.rs`) and imported from both call sites.

packages/daemon-rs/crates/signet-daemon/src/routes/memory.rs
```diff
@@ -232,3 +235,135 @@ pub async fn history(
...
+            .into_response();
+    };
+    let Some(raw) = raw else {
+        return (
+            StatusCode::BAD_REQUEST,
+            Json(serde_json::json!({"error": "sessionKey and feedback required"})),
+        )
+            .into_response();
+    };
+
```
PR-Reviewer-Ant, line 276: When the `feedback`/`ratings` key is absent from the payload (but `sessionKey` was already validated and present), the error body says `"sessionKey and feedback required"` — implying the session key is also missing, which is false. The error message should differentiate the missing field to avoid confusing API consumers.

NicholaiVogel:
addressed the latest review comments in f0b6a1ee.

changes made:
- collapsed duplicate `session_memories` reads in TS path feedback by replacing separate membership/path queries with one `loadSessionData(...)` query that returns both membership set and stored path map.
- rust parity rules updated so `propagated` is ignored for `POST /api/memory/feedback` while full propagation parity is pending.
- rust feedback route/tool now return `propagated: 0` (explicitly no graph propagation yet).
- extracted duplicated feedback parsing logic into shared `packages/daemon-rs/crates/signet-daemon/src/feedback.rs` and reused in both:
  - `routes/memory.rs`
  - `mcp/tools.rs`
- improved rust route error messages to distinguish missing fields:
  - `sessionKey required`
  - `feedback required`

validation:
- `bun test packages/daemon/src/path-feedback.test.ts packages/daemon/src/session-memories.test.ts packages/daemon/src/mcp/tools.test.ts packages/daemon/src/memory-feedback-api.test.ts packages/core/src/migrations/migrations.test.ts`
- `cd packages/daemon-rs && cargo test -p signet-daemon`

packages/daemon-rs/crates/signet-daemon/src/feedback.rs
```diff
@@ -0,0 +1,42 @@
...
+#[cfg(test)]
+mod tests {
+    use super::parse_scores;
+    use serde_json::json;
+
+    #[test]
+    fn parse_scores_accepts_and_clamps_scores() {
+        let parsed = parse_scores(Some(&json!({"a": 2.5, "b": -3.0, "c": 0.5}))).unwrap();
+        assert_eq!(parsed.len(), 3);
+        assert_eq!(parsed[0], ("a".to_string(), 1.0));
```
PR-Reviewer-Ant, line 31: Test asserts positional order of parsed entries (`parsed[0]`, `parsed[1]`, `parsed[2]`). `serde_json::json!` uses `IndexMap` internally which preserves insertion order, so this passes today, but it silently couples the test to serde_json's order-preservation guarantee rather than the function's clamping logic. If serde_json ever switches backing map type (it's happened before in major versions), this test would flip non-deterministically. Consider sorting `parsed` by key before asserting, or using a set/contains check.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f0b6a1ee`

pr-reviewer (automated bot) — Incremental diff f0b6a1ee resolves all three findings from the eb7768aa review: `propagated` moved to `ignoreFields` in parity rules, shared `parse_scores` extracted to `feedback.rs` eliminating duplicate Rust logic, and `propagated: accepted` corrected to `propagated: 0` in both Rust consumers. The `loadSessionData` consolidation in TS correctly collapses two DB round-trips into one without changing semantics. One nitpick from the prior review cycle remains unaddressed.

**Confidence:** High - The incremental diff is small and self-contained (5 files, all mechanical refactors). All three prior findings are visibly resolved. The one remaining item is a test-ordering assumption in a Rust unit test — technically safe with serde_json's IndexMap but worth noting.

NicholaiVogel:
Addressed the follow-on feedback about agent scoping in `session_memories`.

### What changed
- Added **migration 042** to rebuild `session_memories` with:
  - `agent_id TEXT NOT NULL DEFAULT 'default'`
  - `UNIQUE(session_key, agent_id, memory_id)`
  - index `idx_session_memories_agent_session`
- Updated recording paths to write/read agent-scoped rows:
  - `recordSessionCandidates(..., agentId)`
  - `trackFtsHits(..., agentId)` with conflict target `(session_key, agent_id, memory_id)`
  - `recordAgentFeedbackInner(..., agentId)` and wrapper
- Propagated agent id through hook/API call paths:
  - `hooks.ts` session start, user prompt feedback, FTS tracking
  - `/api/memory/feedback` fallback in `daemon.ts`
  - `path-feedback.ts` session membership/path lookup now filters by `agent_id`
- Scoped predictor reads to session memories by agent:
  - `predictor-comparison.ts`
  - `predictor-training-pairs.ts`
- Added/updated tests for agent-scoped behavior across migrations, session memory recording, path feedback membership, and continuity schema assertions.
- Also fixed rust test fragility in `feedback.rs` by sorting parsed entries before assertions.

### Validation
- `bun run build:core`
- `bun test packages/core/src/migrations/migrations.test.ts packages/daemon/src/session-memories.test.ts packages/daemon/src/path-feedback.test.ts packages/daemon/src/pipeline/continuity-scoring.test.ts`
- `bun test packages/daemon/src/predictor-comparison.test.ts packages/daemon/src/predictor-training-pairs.test.ts`
- `bun test packages/daemon/src/memory-feedback-api.test.ts`
- `cd packages/daemon-rs && cargo test -p signet-daemon feedback`

Commit: `4c8532f1`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4c8532f1`

[Automated Bot Review] pr-reviewer — Incremental diff 4c8532f implements agent scoping for session_memories (migration 042 + call-site propagation). No issues found. All prior review findings are resolved.

**Confidence:** High - The incremental diff is small and focused: one new migration, one test determinism fix, and call-site propagation of agentId through five files. The migration correctness was verified — old UNIQUE(session_key, memory_id) data maps cleanly to UNIQUE(session_key, 'default', memory_id) with no collision risk, all pre-existing indexes are explicitly recreated post-rename, and the SAVEPOINT wrapper in runMigrations plus the phantom-repair mechanism handle any partial-failure edge cases. The single remaining nitpick from the f0b6a1ee review (Rust test map ordering) is resolved by the added sort_by call.

PR #313, opened 2026-03-24
fix: identity endpoint, openclaw health retry, marketplace cache

aaf2tbz:
## Summary

Five fixes targeting dashboard display, OpenClaw plugin stability, and embedding tracker reliability:

- **Agent name on Home dashboard** — `/api/identity` now reads from `agent.yaml` first (where `agent.name` is already configured), falling back to IDENTITY.md with support for both legacy `- name:` and markdown bold `**name:**` formats. Previously only matched the legacy format, so the dashboard always showed "SIGNET AGENT" instead of the configured name. Rust parity applied to daemon-rs.

- **False "daemon unreachable" warning** — Startup health check now retries 3× with 1.5s exponential backoff (3s timeout per attempt) instead of a single 1s fire-and-forget. Daemon startup regularly exceeds 1s due to DB migrations and provider checks, causing a misleading warning even though the daemon finishes initializing moments later.

- **Marketplace MCP tools timeout + spam** — `registerMarketplaceProxyTools` was always passing `refresh: true`, defeating the daemon-side 30s cache. Combined with a 5s client timeout and 15s polling interval, every tick spawned fresh MCP server queries that frequently timed out — producing DOMException errors on every Discord message through OpenClaw. Fixed by: only refreshing on first call (subsequent intervals use cache), raising timeout from 5s to 15s, relaxing interval from 15s to 60s.

- **Embedding tracker crash on NULL content_hash** — 282 memory rows had NULL `content_hash` (legacy data predating the column). The tracker's INSERT into embeddings hit a NOT NULL constraint every ~5s cycle, logging continuous "Cycle error" warnings. Fixed by computing the hash on the fly using the same sha256-slice-16 pattern, and backfilling the memories row in the same write transaction. Verified live: errors stopped immediately, null count dropping as batches process.

- **Home page connectors/sessions showing zero** — CONNECTORS was reading from the empty `connectors` table (document connectors) instead of the `harnesses` prop already loaded from `/api/harnesses`. Swapped to `harnesses.length` so it shows configured harnesses (claude-code, openclaw, opencode). SESSIONS was only counting `activeSessions` (plugin claims) but not `bypassedSessions` — added `bypassedSessions` to the DaemonStatus type and the derived count.

## Files Changed

| File | What |
|------|------|
| `packages/daemon/src/daemon.ts` | Rewrote `/api/identity` to read agent.yaml first, IDENTITY.md fallback with dual-format support |
| `packages/daemon-rs/.../config.rs` | Rust parity — uses `state.config.manifest.agent.name` as primary source |
| `packages/adapters/openclaw/src/index.ts` | Health retry with backoff, marketplace timeout + cache fix, relaxed interval |
| `packages/daemon/src/embedding-tracker.ts` | Compute content_hash for null rows, backfill in write transaction |
| `packages/cli/dashboard/src/lib/components/tabs/HomeTab.svelte` | Use `harnesses.length` instead of document connectors |
| `packages/cli/dashboard/src/lib/components/home/AgentHeader.svelte` | Include bypassed sessions in count |
| `packages/cli/dashboard/src/lib/api.ts` | Add `bypassedSessions` to DaemonStatus type |

## No Overlap with Active PRs

- `fix/document-worker-queue-starvation` — pipeline/repair code only
- `fix/mcp-server-connection-pool` — server-side pool refactor in marketplace.ts; our changes are client-side in the openclaw adapter
- `nicholai/dp9-path-feedback-propagation` — hooks + graph-traversal + new module

## Test plan

- [x] Start daemon, verify `curl localhost:3850/api/identity` returns `{"name":"Boogy","creature":"","vibe":""}`
- [ ] Remove `agent.name` from agent.yaml, verify IDENTITY.md `**name:**` format is picked up
- [ ] Start OpenClaw with daemon running — confirm no "daemon unreachable" warning in logs
- [ ] Verify marketplace proxy tools register on startup without DOMException timeout
- [ ] Confirm interval polling uses cache (no `refresh=1` in daemon logs after first call)
- [ ] Dashboard Home page shows configured agent name instead of "SIGNET AGENT"
- [x] Embedding tracker cycle error gone after restart (verified live — 282→260 null rows, backfilling)
- [x] Home page CONNECTORS shows 3 (harnesses count, not empty connectors table)
- [x] Home page SESSIONS includes bypassed sessions in count
- [x] Health score at 100% after clearing dead codex jobs and resyncing vec table


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: identity endpoint, openclaw health retry, marketplace cache` (commit `378eb187`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/api.ts
```diff
@@ -111,6 +111,7 @@ export interface DaemonStatus {
 	agentsDir: string;
 	memoryDb: boolean;
 	activeSessions?: number;
+	bypassedSessions?: number;
```
PR-Reviewer-Ant, line 114: `bypassedSessions` is added to the `DaemonStatus` type, and AgentHeader.svelte sums it into the SESSIONS count. However, the diff shows no corresponding change to the `/api/status` handler in `daemon.ts`. If the daemon does not already return `bypassedSessions` in that response, `daemonStatus?.bypassedSessions` will always be `undefined` (treated as 0), making the fix a no-op. Please confirm that `/api/status` already emits this field, or add the daemon-side change.

packages/daemon/src/daemon.ts
```diff
@@ -1824,25 +1824,66 @@ app.post("/api/config", async (c) => {
...
+					identity.name = agent.name.trim();
+				}
+				if (agent && typeof agent.creature === "string") {
+					identity.creature = agent.creature.trim();
+				}
+				if (agent && typeof agent.vibe === "string") {
+					identity.vibe = agent.vibe.trim();
+				}
+				break;
+			}
```
PR-Reviewer-Ant, line 1900: The new TypeScript identity endpoint returns an empty string `""` for `name` when both agent.yaml and IDENTITY.md are absent or empty. The Rust counterpart (`config.rs`) sets `name = "Unknown"` in that same case. The Svelte component converts `""` to `"SIGNET AGENT"` via `|| "SIGNET AGENT"`, but a Rust daemon instance will show `"UNKNOWN"` in the same scenario. These are different fallback strings — a parity divergence introduced by this PR.

packages/daemon/src/daemon.ts
```diff
@@ -1824,25 +1824,66 @@ app.post("/api/config", async (c) => {
...
+
+	// Secondary source: IDENTITY.md (supports both legacy `- key:` and
+	// markdown bold `**key:**` formats)
+	if (!identity.name) {
+		try {
+			const content = readFileSync(join(AGENTS_DIR, "IDENTITY.md"), "utf-8");
+			for (const line of content.split("\n")) {
+				const trimmed = line.trim();
+				// Legacy format: `- name: Boogy`
+				if (trimmed.startsWith("- name:")) {
```
PR-Reviewer-Ant: IDENTITY.md is only read when `!identity.name` (i.e. agent.yaml has no name). This means `creature` and `vibe` from IDENTITY.md are silently ignored whenever agent.yaml provides a name — even if agent.yaml omits those fields. Before this PR, IDENTITY.md was the sole source for all three fields. Users with `creature`/`vibe` in IDENTITY.md and `name` in agent.yaml will now see empty creature and vibe. Consider reading all three fields from IDENTITY.md as a fallback for any field not already populated by agent.yaml, regardless of whether name was found.

packages/daemon-rs/crates/signet-daemon/src/routes/config.rs
```diff
@@ -125,51 +125,69 @@ pub async fn save_config(
...
+                    if n.is_empty() { n = val.trim().to_string(); }
+                } else if let Some(val) = stripped.strip_prefix("creature:") {
+                    if c.is_empty() { c = val.trim().to_string(); }
+                } else if let Some(val) = stripped.strip_prefix("vibe:") {
+                    if v.is_empty() { v = val.trim().to_string(); }
+                }
+                // Markdown bold: `**name:** Boogy`
+                if let Some(val) = trimmed.strip_prefix("**name:**") {
+                    if n.is_empty() { n = val.trim().to_string(); }
+                } else if let Some(val) = trimmed.strip_prefix("**creature:**") {
```
PR-Reviewer-Ant: Same creature/vibe regression as in the TypeScript daemon: the IDENTITY.md block is guarded by `if name.is_empty()`, so creature and vibe from IDENTITY.md are never read when agent.yaml supplies a name. Affects parity with the old behavior and with any TS daemon instance that is later updated to fix this.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `378eb187`

[Automated Bot Review] This review was generated by an automated pr-reviewer agent. Four issues found: one potentially incomplete fix (bypassedSessions not visible in daemon diff), one dead API call introduced by the connector fix, a silent regression in creature/vibe reading from IDENTITY.md, and a name-fallback parity divergence between the TypeScript and Rust daemons. UI files changed without screenshots.

**Confidence:** Medium - The diff does not include the daemon's /api/status endpoint, so the bypassedSessions completeness cannot be fully verified. All other findings are derivable from the provided diff and file contents.

- packages/cli/dashboard/src/lib/components/tabs/HomeTab.svelte:42 - After swapping `connectors.length` for `harnesses.length`, the `connectors` state variable is populated by `getConnectors()` in onMount but is never referenced in the template or passed to any child component. This leaves a live, unnecessary `/api/connectors` network call on every Home tab mount. The `DocumentConnector` import, the `connectors` state declaration, and the `getConnectors()` call (plus its settled result at results[3]) should all be removed.

packages/cli/dashboard/src/lib/api.ts
aaf2tbz: Confirmed — `/api/status` already returns `bypassedSessions` at line 7131 of daemon.ts (`getBypassedSessionKeys().size`). It just wasn't in the diff so it wasn't visible to the reviewer. No change needed here.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 45bdb51d — Rust daemon no longer falls back to `"Unknown"`. Both TS and Rust now return `""` and the Svelte frontend handles the display fallback (`|| "SIGNET AGENT"`). Single source of truth for the display string.

packages/daemon/src/daemon.ts
aaf2tbz: Fixed in 45bdb51d — IDENTITY.md is now read whenever *any* field is empty (`!name || !creature || !vibe`), not just when name is missing. Each field independently falls back to IDENTITY.md if agent.yaml didn't provide it.

packages/daemon-rs/crates/signet-daemon/src/routes/config.rs
aaf2tbz: Fixed in 45bdb51d — same fix applied to Rust. The IDENTITY.md block now runs when `name.is_empty() || creature.is_empty() || vibe.is_empty()`, with per-field `need_*` flags so only missing fields are read. Parity maintained with the TS daemon.

aaf2tbz:
**Addressing unmapped finding — dead `getConnectors()` call:**

Fixed in 45bdb51d. Removed the `DocumentConnector` type import, `getConnectors` function import, `connectors` state variable, and `getConnectors()` from the `Promise.allSettled` batch in HomeTab.svelte. No remaining references to document connectors on the Home tab.

All 5 findings from Ant's review have been addressed:
| # | Finding | Resolution |
|---|---------|------------|
| 1 | `bypassedSessions` no-op | Already emitted by `/api/status` (line 7131) — not in diff |
| 2 | Name fallback parity | Rust `"Unknown"` removed — both return `""`, frontend owns display |
| 3 | TS creature/vibe regression | Guard changed to `!name \|\| !creature \|\| !vibe` |
| 4 | Rust creature/vibe regression | Same fix with `need_*` flags for parity |
| 5 | Dead `getConnectors()` | Import, state, and call removed |

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4c9451c8`

You've hit your limit · resets 2pm (America/Denver)

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3f9d85c6`

[Automated Bot Review - pr-reviewer] Reviewing HEAD commit 3f9d85c (incremental diff since 4c9451c8). All five previously flagged items from prior rounds are confirmed addressed per PR conversation and are not re-flagged. The incremental diff introduces ~120 lines of agent_id/visibility scoping logic in the Rust daemon that is not described anywhere in this PR's description — the CHANGELOG explicitly attributes it to a separate PR (#317). Two additional correctness issues found: a silent migration assumption for `read_policy`/`policy_group` columns, and hook callers that hardcode `agent_id: "default"` creating a functional gap for multi-agent recall. UI files changed without screenshots.

**Confidence:** High - The incremental diff is unambiguous: search.rs, write.rs, hooks.rs, transactions.rs, and memory.rs all contain agent scoping changes that are (1) not mentioned in the PR description, and (2) explicitly credited to PR #317 in the CHANGELOG entry added in this same diff. The migration assumption and hardcoded agent_id issues are directly readable from the diff without runtime context.

- packages/daemon-rs/crates/signet-daemon/src/routes/search.rs:107 - **Undisclosed scope — PR description does not match implementation.** The incremental diff since the last review round adds ~90 lines of agent-scoped recall logic (agent_id filtering via read_policy/policy_group), plus corresponding changes in write.rs, hooks.rs, transactions.rs, and memory.rs. This PR's description documents exactly five bug fixes and makes no mention of agent scoping or visibility. The CHANGELOG entry added in this same commit explicitly attributes this work to PR #317: `'wire agent_id and visibility through memory write and recall (#317)'`. Human reviewers cannot meaningfully evaluate code they don't know is present. The scoping logic needs to be either (a) removed from this branch and kept in #317, or (b) explicitly described and scoped in this PR's description so reviewers can evaluate it.
- packages/daemon-rs/crates/signet-daemon/src/routes/search.rs:143 - **Silent migration assumption.** The `read_policy` query (`SELECT read_policy FROM agents WHERE id = ?1`) and the `policy_group` query both assume the `agents` table exists with these columns. If a user's database predates the migration that adds these columns, `unwrap_or_else(|_| "isolated".to_string())` silently falls back to `"isolated"` policy. For a user who previously saw all their memories, recall with an `agent_id` argument would silently return nothing (only own memories tagged `agent_id = <aid>`, but all hook-written memories are tagged `agent_id = 'default'` — see related finding). This should either guard with a schema version check or log a warning on query failure.
- packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs:444 - **Hook-written memories hardcode `agent_id: "default"`, breaking isolated-policy recall for non-default agents.** All memories created through the hook path (`remember`, `compaction_complete`) are tagged `agent_id = 'default'`. Under the new scoping logic, a recall call that passes a non-`"default"` `agent_id` with `"isolated"` or `"group"` policy will filter by `agent_id = <caller_aid>` — which will never match these hook-written rows. The result is silent zero recall results for any agent other than the hardcoded default, even for memories that agent created via its own session hooks. The same issue exists in `mcp/tools.rs` line 415. If multi-agent support is the goal of this scoping work, the hook callers need to accept and forward the originating agent's ID.

PR #316, opened 2026-03-24
feat: multi-agent support — scoped identity, memory, and OpenClaw routing

NicholaiVogel:
## Summary

Implements the approved `multi-agent-support` spec (Phases 1–8). Multiple named agents share one daemon and SQLite database with full memory isolation via per-agent read policies.

- **Migration 043** — `agents` roster table + `memories.agent_id` + `memories.visibility` (global/private/archived) columns
- **Read policies** — `isolated` (own only) / `shared` (own + all global) / `group` (own + group-global); enforced in `buildAgentScopeClause()` at SQL level
- **Zero-config OpenClaw routing** — `resolveAgentId()` auto-parses `agent:{id}:{rest}` session keys
- **`signet agent` CLI** — `list / add / remove / purge / info` subcommands; `--agent` + `--private` flags on `remember` / `recall`
- **Daemon** — `GET/POST/DELETE /api/agents`; startup roster sync from `agent.yaml`; file watcher covers `agents/` subdirs; `syncAgentWorkspaces()` with SOUL.md/IDENTITY.md per-agent inheritance
- **OpenClaw connector** — `syncMultipleAgents()` patches `agents.list` into `openclaw.json`

Phases 9–10 (dashboard roster view + setup wizard) are deferred.

## New files

| File | Purpose |
|------|---------|
| `packages/core/src/agents.ts` | `discoverAgents`, `scaffoldAgent`, `getAgentIdentityFiles`, `resolveAgentSkills` |
| `packages/core/src/migrations/043-agents-table.ts` | DB schema for agents + memory columns |
| `packages/daemon/src/agent-id.ts` | `resolveAgentId()` — agentId from body or session key |
| `packages/cli/src/commands/agent.ts` | `signet agent` subcommand |

## Test plan

- [ ] `signet agent add alice` → `~/.agents/agents/alice/` scaffolded, agent.yaml updated
- [ ] `signet agent list` → table shows default + alice with read policies
- [ ] OpenClaw session key `agent:alice:discord:direct:u123` → daemon resolves `alice` as agentId
- [ ] `signet remember "alice fact" --agent alice` → DB row: `agent_id='alice', visibility='global'`
- [ ] `signet remember "alice secret" --agent alice --private` → `visibility='private'`
- [ ] `signet recall "alice" --agent alice` (isolated policy) → does not surface default agent memories
- [ ] `GET /api/agents` → lists all agents
- [ ] `signet agent remove alice` → alice's memories archived, removed from roster
- [ ] Single-agent regression: install with no `agents.roster` behaves identically to today
- [ ] `bun run typecheck` passes, `bun test` passes (no regressions), `bun run lint` clean

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -461,8 +461,8 @@ interface OpenClawHeartbeatData {
 	readonly latencyMs: number;
 	/** Failures reported in the most recent heartbeat (delta, not cumulative). */
 	readonly lastFailedDelta: number;
-	totalSucceeded: number;
-	totalFailed: number;
+	readonly totalSucceeded: number;
+	readonly totalFailed: number;
 }
 let openClawHeartbeat: { timestamp: string; data: OpenClawHeartbeatData } | null = null;
 const OPENCLAW_STALE_MS = 10 * 60 * 1000; // 10 minutes


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: multi-agent support — scoped identity, memory, and OpenClaw routing` (commit `c6396dfc`) now and I'll follow up shortly with feedback.

packages/cli/src/commands/memory.ts
```diff
@@ -33,6 +35,8 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 				tags: options.tags,
 				importance: options.importance,
 				pinned: options.critical,
+				...(options.agent ? { agent_id: options.agent } : {}),
```
PR-Reviewer-Ant: Field name mismatch: `remember` sends `agent_id` (snake_case) but the API contract (updated in this same PR's `docs/API.md`) specifies `agentId` (camelCase). The immediately adjacent `recall` path (line 87) correctly sends `agentId`. The daemon will likely ignore the unknown field and store the memory under the default agent, making `signet remember --agent alice` silently a no-op for agent scoping.

packages/connector-openclaw/src/index.ts
```diff
@@ -451,6 +451,47 @@ export class OpenClawConnector extends BaseConnector {
...
+	/**
+	 * Sync a multi-agent roster into the `agents.list` section of all
+	 * discovered OpenClaw configs. Only agents that include `"openclaw"` in
+	 * their `harnesses` array (or have no harnesses specified) are written.
+	 */
+	async syncMultipleAgents(
+		roster: ReadonlyArray<{
+			name: string;
+			harnesses?: ReadonlyArray<string>;
+			skills?: ReadonlyArray<string>;
```
PR-Reviewer-Ant, line 463: Path traversal: `a.name` comes from the `roster` array which is read directly from `agent.yaml` — a user-editable file. There is no validation applied at this layer (validation only exists in the CLI `add` command). A roster entry with `name: '../../evil'` would resolve `join(basePath, 'agents', '../../evil', 'workspace')` outside the workspace directory. Because the result is written into every discovered OpenClaw config via `atomicWriteJson`, this could be used to plant a controlled path in tool configs. Validate names against the same `^[a-z0-9][a-z0-9-]*$` rule before constructing the path.

packages/connector-openclaw/src/index.ts
```diff
@@ -451,6 +451,47 @@ export class OpenClawConnector extends BaseConnector {
...
+		);
+
+		const list = eligible.map((a) => ({
+			id: a.name,
+			name: a.name,
+			workspace: join(basePath, "agents", a.name, "workspace"),
+			...(a.skills && a.skills.length > 0 ? { skills: a.skills } : {}),
+		}));
+
+		const patch: JsonObject = { agents: { list } };
```
PR-Reviewer-Ant: `deepMerge(config, patch)` replaces `agents.list` wholesale with the Signet roster. Any agents that were registered directly in the OpenClaw config (not in `agent.yaml`) are silently removed. The existing `backupConfig` call mitigates against total loss, but the behaviour is still destructive. Consider merging by id instead of replacing, or at minimum logging a warning when pre-existing entries are being dropped.

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,323 @@
...
+					console.log(chalk.dim("  Cancelled"));
+					return;
+				}
+			}
+
+			const spinner = ora(`Purging agent ${chalk.cyan(name)}...`).start();
+
+			const result = await deps.fetchFromDaemon<{ success?: boolean; error?: string }>(
+				`/api/agents/${encodeURIComponent(name)}?purge=true`,
+				{ method: "DELETE" },
```
PR-Reviewer-Ant, line 255: Split-brain on daemon error: `rmSync` is called unconditionally after `fetchFromDaemon`. If the daemon is reachable but returns `{ error: '...' }` (e.g. agent not found, auth failure), the local identity directory and `agent.yaml` roster entry are still deleted while the daemon retains the agent record. Future daemon sync may then re-create the agent from DB state without its identity files, or the agent becomes an orphan. Only delete local files when the daemon explicitly confirms success or is confirmed offline.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c6396dfc`

[Automated Review — pr-reviewer bot] Found one confirmed bug (field name mismatch silently discards `--agent` flag on `remember`), one path-traversal risk (unsanitised agent names used in filesystem joins), one data-loss risk in `syncMultipleAgents`, and one state-consistency issue in `purge`. All four warrant attention before merge.

**Confidence:** High - The agent_id/agentId mismatch is directly visible in the diff and is unambiguously wrong given the adjacent recall code and the updated API docs. The path-traversal and deepMerge issues follow directly from reading the connector code without needing runtime context. The purge state issue is straightforward control-flow.

NicholaiVogel:
Thanks for the review. Addressed all four issues in 93fd68ac:

1. **field name mismatch** — `agent_id` → `agentId` in `remember` (recall was already correct)
2. **purge split-brain** — local files now only deleted when daemon confirms `success: true` or is confirmed offline; early-return on daemon error
3. **path traversal** — agent names validated against `^[a-z0-9][a-z0-9-]*$` before any `join()` in `syncMultipleAgents`
4. **deepMerge wholesale replace** — now merges `agents.list` by `id`, keeping pre-existing OpenClaw agents whose ids are not in the Signet roster

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,331 @@
...
+			// If the daemon returned an explicit error the agent record may still
+			// exist in DB; deleting the local files would create an orphan.
+			const daemonOk = result?.success === true;
+			const daemonOffline = result === null;
+			if (result?.error) {
+				spinner.fail(`Daemon returned error: ${result.error} — local files preserved`);
+				return;
+			}
+			if (daemonOffline) {
+				spinner.warn("Daemon offline — cleaning up local files only");
```
PR-Reviewer-Ant, line 269: **Path traversal — blocking.** `name` arrives directly from the CLI argument and is never validated against the safe-name regex before this `join()` + `rmSync()`. The `validateName()` helper already exists and is called in the `add` command, but is absent here.

When the daemon is offline (`daemonOffline === true`), execution reaches this branch unconditionally. An attacker (or a user who makes a typo) running `signet agent purge '../../sensitive-dir'` will resolve to `join(AGENTS_DIR, 'agents', '../../sensitive-dir')` = `join(AGENTS_DIR, 'sensitive-dir')`, then `rmSync` it recursively.

Fix: call `validateName(name)` at the top of the purge action and exit early on error, identical to the add command pattern.

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,331 @@
...
+
+			const result = await deps.fetchFromDaemon<{ success?: boolean; error?: string }>(
+				`/api/agents/${encodeURIComponent(name)}?purge=true`,
+				{ method: "DELETE" },
+			);
+
+			// Only clean up local files when daemon confirmed success (success: true)
+			// or was offline (null result — no error was returned by the daemon).
+			// If the daemon returned an explicit error the agent record may still
+			// exist in DB; deleting the local files would create an orphan.
```
PR-Reviewer-Ant, line 261: **Inconsistent offline-detection — warning.** The purge command sets `daemonOffline = result === null` (strict null), while the `add` command (line ~160) uses `!result` (falsy). If `fetchFromDaemon` can return `undefined` for certain offline or timeout conditions, `daemonOffline` will be `false`, `daemonOk` will also be `false` (since `undefined?.success !== true`), and the local file cleanup is silently skipped — but `spinner.succeed()` still fires, telling the user the purge succeeded when it did not. Align with the `!result` pattern used elsewhere in the same file.

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,331 @@
...
+			spinner.succeed(`Agent ${chalk.cyan(name)} removed`);
+			console.log(chalk.dim("  Memories archived. Files retained in agents/ directory."));
+		});
+
+	// ── signet agent purge <name> ────────────────────────────────────────────
+
+	agentCmd
+		.command("purge <name>")
+		.description("Permanently delete an agent and all their memories")
+		.option("--force", "Skip confirmation prompt")
```
PR-Reviewer-Ant, line 216: **remove always mutates roster on daemon error — warning.** Unlike `purge` (which now early-returns on `result?.error`), `remove` calls `removeFromRoster()` unconditionally even when the daemon returns an error. If the daemon fails mid-archive (e.g. a DB transaction error), the memories are not archived but the agent is removed from agent.yaml. On the next harness sync, the agent no longer exists in the roster, but its memories remain in the DB with the original `agent_id` and `visibility='global'`, potentially visible to `shared`-policy agents. Consider matching the `purge` pattern: skip `removeFromRoster` when `result?.error` is set.

packages/daemon/src/agent-id.ts
```diff
@@ -0,0 +1,15 @@
+/**
```
PR-Reviewer-Ant, line 1: **Rust parity rule — warning.** AGENTS.md requires that any behavioral change to `@signet/daemon` is also reflected in `packages/daemon-rs/`. This PR adds `/api/agents` endpoints, `resolveAgentId()`, `buildAgentScopeClause()`, migration 043, and file-watcher changes — none of which appear in the diff for `packages/daemon-rs/`. Shadow-divergence logs will accumulate until this is addressed.

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,331 @@
...
+			}
+
+			const spinner = ora(`Purging agent ${chalk.cyan(name)}...`).start();
+
+			const result = await deps.fetchFromDaemon<{ success?: boolean; error?: string }>(
+				`/api/agents/${encodeURIComponent(name)}?purge=true`,
+				{ method: "DELETE" },
+			);
+
+			// Only clean up local files when daemon confirmed success (success: true)
```
PR-Reviewer-Ant, line 258: **Blocking — wrong response field, files never cleaned up on live daemon.**

The API docs in this PR show the DELETE `/api/agents/:name` endpoint returns `{ "ok": true }`, not `{ "success": true }`. This means `daemonOk` is always `false` when the daemon responds successfully, so the `if (daemonOk || daemonOffline)` block never executes. After `signet agent purge alice`:
1. Daemon archives alice's memories (correct)
2. `result = { ok: true }` — no `.error`, not `null`
3. `daemonOk = result?.success === true` → `false`
4. `daemonOffline = result === null` → `false`
5. Files and roster entry are **NOT** removed
6. Spinner still reports "purged"

Fix: change the check to `result?.ok === true` (matching the documented response shape), or update the generic type to `{ ok?: boolean; error?: string }` and check `.ok`.

packages/connector-openclaw/src/index.ts
```diff
@@ -451,6 +451,70 @@ export class OpenClawConnector extends BaseConnector {
...
+				const raw = readFileSync(configPath, "utf-8");
+				const config = parseJsonOrJson5(raw);
+				const indent = this.detectIndent(raw);
+
+				// Preserve pre-existing OpenClaw agents not managed by Signet.
+				// Only replace entries whose id is in the Signet roster.
+				const existing = config as Record<string, unknown>;
+				const agentsSection = existing.agents as Record<string, unknown> | undefined;
+				const prevList = Array.isArray(agentsSection?.list)
+					? (agentsSection.list as Array<Record<string, unknown>>)
```
PR-Reviewer-Ant, line 497: **Warning — deepMerge semantics with arrays may produce duplicates.**

`merged` is computed as `[...kept, ...signetEntries]` — already the complete intended list. Then `deepMerge(config, { agents: { list: merged } })` is called. If `deepMerge` concatenates arrays rather than replacing them, the existing `config.agents.list` (which still contains the original entries including those in `kept`) would be merged with `merged`, doubling the preserved non-Signet entries.

This was the root of the original issue the previous review flagged. If `deepMerge` replaces arrays this is fine, but consider directly assigning:
```ts
(config as Record<string, unknown>).agents = { ...((config as Record<string, unknown>).agents as object ?? {}), list: merged };
```
or verify `deepMerge` replaces (not concatenates) array-valued keys.

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,331 @@
...
+			const daemonOffline = result === null;
+			if (result?.error) {
+				spinner.fail(`Daemon returned error: ${result.error} — local files preserved`);
+				return;
+			}
+			if (daemonOffline) {
+				spinner.warn("Daemon offline — cleaning up local files only");
+			}
+
+			if (daemonOk || daemonOffline) {
```
PR-Reviewer-Ant, line 272: **[Carry-forward from prior review — still unaddressed]** The `purge` action passes `name` directly into `join(deps.AGENTS_DIR, 'agents', name)` and then calls `rmSync(..., { recursive: true, force: true })` without ever calling `validateName()`. The only guard is `name === 'default'`, which does not prevent values like `../../.ssh` or `../../../etc`. In the `daemonOffline` branch (reachable whenever the daemon is not running), the deletion proceeds entirely locally with no server-side check. The `add` command correctly gates on `validateName()` before any path construction; `purge` must do the same. Fix: call `validateName(name)` immediately after the `name === 'default'` guard and exit on error.

NicholaiVogel:
Second round of feedback addressed in 8ec3276a:

**CLI fixes:**
1. **Path traversal in purge** — `validateName()` now called at the top of the purge action (same as `add`). Rejects names with path components before any filesystem join.
2. **Offline sentinel** — changed `result === null` to `!result` in purge to match the `add` pattern and handle undefined/timeout returns consistently.
3. **remove roster mutation on daemon error** — `removeFromRoster` is now skipped (early return) when the daemon returns an explicit error, matching the purge pattern.

**Rust parity (daemon-rs):**
4. Added `routes/agents.rs` with full GET/POST/DELETE implementations using the `state.pool.read`/`write` pattern. Registered in `main.rs` router. Updated `contracts/schema.sql` with `memories.agent_id`, `memories.visibility`, and the `agents` table. Verified `cargo check` clean.

packages/cli/src/commands/agent.ts
```diff
@@ -0,0 +1,339 @@
...
+			const result = await deps.fetchFromDaemon<{ success?: boolean; error?: string }>(
+				`/api/agents/${encodeURIComponent(name)}?purge=true`,
+				{ method: "DELETE" },
+			);
+
+			// Only clean up local files when daemon confirmed success (success: true)
+			// or was offline (null result — no error was returned by the daemon).
+			// If the daemon returned an explicit error the agent record may still
+			// exist in DB; deleting the local files would create an orphan.
+			const daemonOk = result?.success === true;
```
PR-Reviewer-Ant, line 262: **Response-field mismatch (carried over from prior review — not confirmed fixed):** `daemonOk` is set via `result?.success === true`, but the API docs added in this same PR (`docs/API.md`) document the DELETE response as `{ "ok": true }`. If the TypeScript daemon follows its own docs, `result.success` is always `undefined`, `daemonOk` is always `false`, and the `if (daemonOk || daemonOffline)` block on line ~268 never executes on a normal online purge — leaving local files and the `agent.yaml` roster entry un-cleaned after the daemon successfully deletes the DB records. The Rust parity handler added in this commit returns `{ "success": true, "purged": ... }`, which aligns with the CLI check but diverges from the documented shape. At minimum, one of (a) the TypeScript daemon response, (b) the Rust response, (c) the CLI check, or (d) the API docs is wrong. This was previously flagged as blocking and the author's latest response does not mention addressing it.

packages/daemon-rs/crates/signet-daemon/src/routes/agents.rs
```diff
@@ -0,0 +1,171 @@
...
+            }
+            if purge {
+                conn.execute("DELETE FROM memories WHERE agent_id = ?1", rusqlite::params![name])?;
+            } else {
+                conn.execute(
+                    "UPDATE memories SET visibility = 'archived' WHERE agent_id = ?1",
+                    rusqlite::params![name],
+                )?;
+            }
+            conn.execute("DELETE FROM agents WHERE name = ?1", rusqlite::params![name])?;
```
PR-Reviewer-Ant, line 176: The DELETE handler returns `{ "success": true, "purged": purge }` but the API docs say `{ "ok": true }`. Per AGENTS.md, the Rust shadow should mirror TypeScript behavior. This divergence contributes to the field-mismatch confusion described above and will cause shadow divergence log noise if the TypeScript daemon returns `{ ok: true }`.

packages/daemon-rs/crates/signet-daemon/src/routes/agents.rs
```diff
@@ -0,0 +1,171 @@
...
+    let group = body.policy_group;
+    let now = chrono::Utc::now().to_rfc3339();
+
+    let result = state
+        .pool
+        .write(signet_core::db::Priority::High, move |conn| {
+            conn.execute(
+                "INSERT INTO agents (id, name, read_policy, policy_group, created_at, updated_at) \
+                 VALUES (?1, ?2, ?3, ?4, ?5, ?6)",
+                rusqlite::params![name, name, read_policy, group, now, now],
```
PR-Reviewer-Ant, line 127: The `create` handler does not validate the agent name (no check for reserved name `"default"`, no alphanumeric+hyphens constraint). The TypeScript CLI enforces `^[a-z0-9][a-z0-9-]*$` and rejects `"default"`. If the TypeScript daemon POST handler also validates these constraints, the Rust diverges by accepting arbitrary strings. An agent named `"default"` created via the HTTP API would then be deletable via DELETE (which explicitly guards against it) but would also corrupt single-agent fallback behavior.

NicholaiVogel:
Addressed all outstanding feedback:

**Round 3 (this round)**

- **API.md response shape** — corrected `{ "ok": true }` → `{ "success": true, "purged": false }` to match the actual TypeScript daemon response (`{ success: true, purged: purge }`). The CLI check `result?.success === true` was always correct; only the docs were wrong.
- **deepMerge array semantics** — added an inline comment in `syncMultipleAgents` clarifying that `deepMerge` replaces arrays (not concatenates), making the pre-built `merged` array safe to pass directly.
- **Rust `create` handler name validation** — added `validate_name()` to `routes/agents.rs` matching the TypeScript CLI's constraints (`^[a-z0-9][a-z0-9-]*$`, rejects `"default"`). Returns 400 on violation.

**Carry-forward issues (already fixed in prior rounds — confirming)**

- `validateName` in purge: fixed in 8ec3276a
- `daemonOffline = !result`: fixed in 8ec3276a
- `remove` early-return on daemon error: fixed in 8ec3276a
- Path traversal in `syncMultipleAgents`: fixed in 93fd68ac
- Rust parity (routes, schema, main): added in 8ec3276a

NicholaiVogel:
Addressed all remaining issues:

**Response-field mismatch (blocking, rounds 3–4)**
- Confirmed: TypeScript daemon returns `{ success: true, purged: purge }` (see daemon.ts:1973). The API.md was wrong, not the code.
- Fixed in `a7768af5`: API.md updated to `{ "success": true, "purged": false }`. The CLI check `result?.success === true` was always correct. The Rust handler also returns `{ "success": true, "purged": ... }` — all three are now consistent.

**Rust `create` handler name validation (round 4)**
- Fixed in `2579ffee`: added `validate_name()` to `routes/agents.rs` matching `^[a-z0-9][a-z0-9-]*$` + rejects `"default"`. Returns 400 on violation.

**`readonly` on `totalSucceeded`/`totalFailed` (round 2 unmapped warning)**
- Fixed in `8d71dcba`: both fields marked `readonly` — the heartbeat object is replaced wholesale, not mutated in place.

All prior-round issues remain fixed. No open items remain.

packages/connector-opencode/src/plugin-bundle.ts
```diff
@@ -5,145 +5,145 @@
...
...{let W=/^\\d+\\.\\d+$/.test(H);return J(6,\`Unsupported YAML version \${H}\`,W),!1}}default:return J(0,\`Unknown directive \${Q}\`,!0),!1}}tagName(G,J){if(G==="!")return"!";if(G[0]!=="!")return J(\`Not a valid tag: \${G}\`),null;if(G[1]==="<"){let W=G.slice(2,-1);if(W==="!"||W==="!!")return J(\`Verbatim tags aren't resolved, so \${G} is invalid.\`),null;if(G[G.length-1]!==">")J("Verbatim tags must end with a >");return W}let[,D,Q]=G.match(/^(.*!)([^!]*)$/s);if(!Q)J(\`The \${G} tag has no suffix\`);let H=this.tags[D];if(H)try{return H+decodeURIComponent(Q)}catch(W){return J(String(W)),null}if(D==="!")return G;return J(\`Could not resolve tag: \${G}\`),null}tagString(G){for(let[J,D]of Object.entries(this.tags))if(G.startsWith(D))return J+U(G.substring(D.length));return G[0]==="!"?G:\`!<\${G}>\`}toString(G){let J=this.yaml.explicit?[\`%YAML \${this.yaml.version||"1.2"}\`]:[],D=Object.entries(this.tags),Q;if(G&&D.length>0&&z.isNode(G.contents)){let H={};I.visit(G.contents,(W,V)=>{if(z.isNode(V)&&V.tag)H[V.tag]=!0}),Q=Object.keys(H)}else Q=[];for(let[H,W]of D){if(H==="!!"&&W==="tag:yaml.org,2002:")continue;if(!G||Q.some((V)=>V.startsWith(W)))J.push(\`%TAG \${H} \${W}\`)}return J.join(\`
```
PR-Reviewer-Ant, line 8: The entire PLUGIN_BUNDLE constant was replaced with a new minified blob. The comment states it is regenerated from `../opencode-plugin/dist/signet.mjs`, but the opencode-plugin source changes that produced this bundle are not present in the diff. A reviewer cannot verify what behavioral changes are encoded here. Before merge, confirm the bundle was regenerated from a reviewed source diff in `packages/opencode-plugin/`.

NicholaiVogel:
**Re: `plugin-bundle.ts` — bundle is reproducible and source is unchanged**

The `packages/opencode-plugin/src/` source is identical between `main` and this branch (no diff). The plugin-bundle changed because the embed script was re-run during the initial build (`scripts/embed-plugin.ts` reads from `../opencode-plugin/dist/signet.mjs`).

To verify: rebuilding the plugin from source produces the same hash as the committed bundle:

```
cd packages/opencode-plugin && bun run build
cd packages/connector-opencode && bun run scripts/embed-plugin.ts
# → md5: 50780eab2e2b10b32483e5cab248faf8  (matches committed file)
```

The diff vs `main` is a build-tool output variance (bun version), not a behavioral change — the source feeding the bundle is unchanged.

PR #317, opened 2026-03-24
fix: wire agent_id and visibility through memory write and recall

NicholaiVogel:
## Summary

- **Write path**: The remember endpoint accepted `agentId` and `visibility` in the request body but never passed them through to `txIngestEnvelope`, so all memories were stored as `agent_id='default'`, `visibility='global'` regardless of CLI flags (`--agent`, `--private`)
- **Read path**: The recall hydration query (where FTS, vector, and traversal results converge) did not apply `buildAgentScopeClause`, so vector/traversal results bypassed agent visibility filtering — private memories from one agent were visible to others
- Fixes both paths across `daemon.ts`, `transactions.ts`, and `memory-search.ts`

## Test plan

- [x] `signet remember "test" --agent dot --private` stores with `agent_id='dot'`, `visibility='private'` (verified on solvr)
- [x] `signet recall "test" --agent dot` returns dot's private + global memories
- [x] `signet recall "test" --agent miles` does NOT return dot's private memory
- [x] `signet recall "test"` (no agent) returns only global memories
- [ ] `bun test` passes
- [ ] `bun run build` succeeds

### DIFF
--- a/packages/daemon-rs/crates/signet-core/src/queries/memory.rs
+++ b/packages/daemon-rs/crates/signet-core/src/queries/memory.rs
@@ -554,6 +554,8 @@ mod tests {
             runtime_path: None,
             now: "2024-01-01T00:00:00Z",
             updated_by: "test",
+            agent_id: "default",
+            visibility: "global",
         }
     }
 
--- a/packages/daemon-rs/crates/signet-core/src/queries/memory.rs
+++ b/packages/daemon-rs/crates/signet-core/src/queries/memory.rs
@@ -169,6 +169,8 @@ pub struct InsertMemory<'a> {
     pub runtime_path: Option<&'a str>,
     pub now: &'a str,
     pub updated_by: &'a str,
+    pub agent_id: &'a str,
+    pub visibility: &'a str,
 }
 
 /// Insert a new memory row. Returns the ID.
@@ -179,8 +181,8 @@ pub fn insert(conn: &Connection, m: &InsertMemory) -> Result<String, CoreError> 
           who, why, project, importance, pinned, is_deleted,
           extraction_status, embedding_model, extraction_model,
           source_type, source_id, idempotency_key, runtime_path,
-          confidence, created_at, updated_at, updated_by)
-         VALUES (?1,?2,?3,?4,?5,?6,?7,?8,?9,?10,?11,0,?12,?13,?14,?15,?16,?17,?18,?10,?19,?19,?20)",
+          confidence, created_at, updated_at, updated_by, agent_id, visibility)
+         VALUES (?1,?2,?3,?4,?5,?6,?7,?8,?9,?10,?11,0,?12,?13,?14,?15,?16,?17,?18,?10,?19,?19,?20,?21,?22)",
         params![
             m.id,
             m.content,
@@ -202,6 +204,8 @@ pub fn insert(conn: &Connection, m: &InsertMemory) -> Result<String, CoreError> 
             m.runtime_path,
             m.now,
             m.updated_by,
+            m.agent_id,
+            m.visibility,
         ],
     )?;
     Ok(m.id.to_string())
--- a/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
@@ -442,6 +442,8 @@ pub async fn remember(
                     idempotency_key: idempotency_key.as_deref(),
                     runtime_path: runtime_path_str.as_deref(),
                     actor: "hook",
+                    agent_id: "default",
+                    visibility: "global",
                 },
             )?;
 
@@ -726,6 +728,8 @@ pub async fn compaction_complete(
                     idempotency_key: None,
                     runtime_path: None,
                     actor: "compaction",
+                    agent_id: "default",
+                    visibility: "global",
                 },
             )?;
 
--- a/packages/daemon-rs/crates/signet-daemon/src/routes/search.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/routes/search.rs
@@ -25,7 +25,6 @@ pub struct RecallBody {
     pub query: String,
     pub keyword_query: Option<String>,
     pub limit: Option<usize>,
-    #[allow(dead_code)] // Used in Phase 3 for agent scoping
     pub agent_id: Option<String>,
     #[serde(rename = "type")]
     pub memory_type: Option<String>,
@@ -105,6 +104,7 @@ pub async fn recall(
     let importance_min = body.importance_min;
     let since = body.since.clone();
     let until = body.until.clone();
+    let agent_id = body.agent_id.clone();
     let query_for_response = query.clone();
 
     let result = state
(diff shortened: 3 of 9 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: wire agent_id and visibility through memory write and recall` (commit `cc528bec`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -2769,6 +2771,8 @@ app.post("/api/memory/remember", async (c) => {
 	const raw = body.content?.trim();
```
PR-Reviewer-Ant, line 2771: **Daemon-rs parity rule not addressed (AGENTS.md hard requirement).**

AGENTS.md states: 'Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`.' This PR changes the core behavioral contract — agent_id and visibility are now correctly stored on both write paths — but the PR description makes no mention of daemon-rs. The shadow proxy logs divergences to `$SIGNET_WORKSPACE/.daemon/logs/shadow-divergences.jsonl`; if daemon-rs is running alongside, every correctly-tagged write will show as a divergence. More importantly, if daemon-rs is serving traffic the fix has no effect on that path. Please update daemon-rs or document why parity is not required here.

packages/daemon/src/memory-search.ts
```diff
@@ -842,24 +842,30 @@ export async function hybridRecall(
 	// --- Fetch full memory rows ---
 	// Scope filter on hydration catches any results that bypassed
 	// the FTS filter clause (e.g. unscoped graph boost results).
+	// Agent scope filter also applied here to catch vector/traversal
+	// results that bypass FTS-level agent filtering.
 	const scopeClause =
 		params.scope !== undefined
 			? params.scope === null
 				? " AND scope IS NULL"
 				: " AND scope = ?"
```
PR-Reviewer-Ant: **Secondary visibility leak: supplementary decision-rationale results bypass `agentScope`.**

The hydration fix correctly applies `agentScope.sql` to the main FTS/vector/traversal result set. However, further down in `hybridRecall`, the decision-rationale supplementary query fetches linked `rationale`-type memories without applying `agentScope`:

```sql
SELECT DISTINCT m.id, ... FROM memory_entity_mentions mem
JOIN memories m ON m.id = mem.memory_id
WHERE mem.entity_id IN (...) AND m.type = 'rationale' AND m.is_deleted = 0
${scopeClause}
LIMIT 10
```

This path is not in the diff, but the PR's stated goal is to fix the recall hydration path 'where FTS, vector, and traversal results converge.' Rationale memories tagged private to another agent can appear as `supplementary: true` results in any recall response where the requesting agent shares a decision entity. Consider appending `${agentScope.sql}` and the corresponding args to this query as well.

packages/daemon/src/transactions.ts
```diff
@@ -33,6 +33,8 @@ export interface IngestEnvelope {
 	sourceType: string;
 	sourceId: string | null;
 	scope?: string | null;
```
PR-Reviewer-Ant, line 35: **`visibility` typed as `string` instead of a literal union.**

Repo convention (AGENTS.md): 'no enum — use `as const` + union types.' `visibility` has three known values (`'global'`, `'private'`, `'archived'`). Typing it as `string` loses exhaustiveness checking and allows callers to pass arbitrary strings that will be stored verbatim. A union type `'global' | 'private' | 'archived'` would catch misuse at the call sites in daemon.ts.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `cc528bec`

Automated bot review (pr-reviewer). The write-path fix is sound — agentId and visibility are now correctly resolved and propagated to txIngestEnvelope at both call sites, and the chunk_group entity INSERT is fixed. The read-path fix plugs the main vector/traversal hydration gap. Two issues need attention before merge.

**Confidence:** Medium - The diff logic is clear and the core fix is correct. Medium confidence because I cannot verify the daemon-rs shadow state, the resolveAgentId implementation, or whether an `agents` table exists in the schema — all of which affect correctness of the overall fix.


- packages/daemon/src/memory-search.ts:88 - **`readPolicy === 'group'` subquery references `agents` table — existence unverified.**

```sql
SELECT id FROM agents WHERE policy_group = ?
```

The database migrations listed in AGENTS.md go up to `039-dedup-entity-dependencies`. If an `agents` table hasn't been created by a migration, this subquery will throw a SQL error at runtime for any user with group policy configured, making recall fail entirely for that policy mode. Please confirm the migration that creates this table exists and is registered, or add a guard.

packages/daemon-rs/crates/signet-services/src/transactions.rs
```diff
@@ -30,6 +30,8 @@ pub struct IngestInput<'a> {
     pub idempotency_key: Option<&'a str>,
```
PR-Reviewer-Ant, line 30: The `IngestInput` struct now requires `agent_id: &'a str` and `visibility: &'a str` (non-optional). All four `IngestInput { ... }` struct literals in the `#[cfg(test)]` section (`ingest_and_deduplicate`, `modify_records_history`, `forget_and_recover`, `pipeline_cannot_force_delete_pinned`) are missing these fields. Rust exhaustive struct init means this will not compile. Add `agent_id: "default", visibility: "global"` to each test `IngestInput`.

packages/daemon-rs/crates/signet-core/src/queries/memory.rs
```diff
@@ -169,6 +169,8 @@ pub struct InsertMemory<'a> {
     pub runtime_path: Option<&'a str>,
```
PR-Reviewer-Ant, line 169: `InsertMemory` now has required `agent_id: &'a str` and `visibility: &'a str` fields, but the `make_input` test helper does not initialize them. All tests that call `make_input` will fail to compile. Add `agent_id: "default"` and `visibility: "global"` to `make_input`.

packages/daemon-rs/crates/signet-daemon/src/routes/search.rs
```diff
@@ -141,27 +141,83 @@ pub async fn recall(
...
+                        format!(
+                            " AND (visibility = 'global' OR agent_id = ?{}) AND visibility != 'archived'",
+                            id_count + 1
+                        ),
+                        vec![aid.clone()],
+                    ),
+                    "group" => {
+                        let group: Option<String> = conn
+                            .query_row(
+                                "SELECT policy_group FROM agents WHERE id = ?1",
```
PR-Reviewer-Ant, line 176: When `policy` is `"group"` and `SELECT policy_group FROM agents WHERE id = ?1` returns `NULL` or fails (e.g. agent exists but has no group configured), `group.unwrap_or_default()` passes `""` to the subquery `SELECT id FROM agents WHERE policy_group = ?`. Any agents with an empty-string `policy_group` would be inadvertently included. The correct fallback is isolated behavior (own memories only). Consider: `if let Some(g) = group { ... } else { /* fall through to isolated arm */ }`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a3ac01e3`

[Automated Review — pr-reviewer] PR correctly wires agent_id/visibility through both write and read paths. Two compile-blocking issues found in the Rust test suite, plus one logic error in the group-policy recall path.

**Confidence:** High - The compile errors are unambiguous from the struct definitions and test code in the provided file contents. The group unwrap_or_default logic issue is straightforward. The prior agents-table concern is now safely handled on the Rust side via unwrap_or_else fallback to 'isolated'.

packages/daemon-rs/crates/signet-daemon/src/routes/search.rs
```diff
@@ -141,27 +141,94 @@ pub async fn recall(
...
+                                rusqlite::params![aid],
+                                |row| row.get(0),
+                            )
```
PR-Reviewer-Ant, line 179: The `if let Some(g) = group` fix correctly resolves the prior compile error and the group-policy logic bug (`unwrap_or_default()` would have silently passed an empty-string group to the subquery, potentially leaking cross-agent memories). The fallback to isolated when no group is configured is the correct safe default. No issue here — confirming the fix is sound.

packages/daemon/src/memory-search.ts
PR-Reviewer-Ant: `scopeClause` emits unqualified column references (`scope IS NULL`, `scope = ?`) while the surrounding query now aliases `memories` as `m` and `agentScope.sql` uses qualified `m.visibility`/`m.agent_id` references. SQLite resolves unqualified names correctly when only one table has that column, so this is not a runtime bug today. However, if this query is ever extended with a JOIN to another table that has a `scope` column, the clause would silently bind to the wrong table. Consider updating `scopeClause` to use `m.scope` for consistency.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -442,6 +442,8 @@ pub async fn remember(
                     idempotency_key: idempotency_key.as_deref(),
```
PR-Reviewer-Ant, line 442: The `remember` hook and `compaction_complete` handler still hardcode `agent_id: "default"` and `visibility: "global"`. This means memories written via hook (connector-driven) are always attributed to the default agent, regardless of which agent session triggered the hook. This is pre-existing behavior, not introduced by this PR, but it creates a gap: a named-agent session calling `signet remember` via hook will store as `agent_id='default'`. This may be intentional for the legacy hook path, but worth a follow-up issue if per-agent hooks are expected.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 2771: `resolveAgentId` is called but not visible in the diff. Confirming this function pre-exists and is not newly introduced — if it has a bug (e.g. falls through to 'default' when a valid session key maps to a named agent), the write-path fix would be silently ineffective. Not flagging as a new issue since it was raised in the first review cycle and not re-raised in the second, but recommend a unit test covering the `resolveAgentId(sessionKey)` lookup path.

PR-Reviewer-Ant, line 2771: The PR test plan leaves `bun test passes` and `bun run build succeeds` unchecked. These are the minimum automated validation gates defined in AGENTS.md. The fix touches SQL parameter binding and Rust struct layout — both areas where silent breakage is possible without build/test validation.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8754cfda`

[Automated Review — pr-reviewer] Incremental diff since last reviewed SHA (a3ac01e3) addresses both compile-blocking issues flagged in the prior review. No new bugs introduced.

**Confidence:** High - The incremental diff is small and targeted: three files, all fixing the exact issues called out in the prior CHANGES_REQUESTED review. The logic is unambiguous. Medium-high confidence only because bun test / bun run build are still unchecked in the test plan, so runtime behavior is unverified by CI.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fe5ab5ac`

[Automated Review — pr-reviewer] Incremental diff (8754cfda → fe5ab5ac) is a single three-line fix: adds the `m.` table alias to the `scopeClause` variable in memory-search.ts. All prior blocking issues have been resolved.

**Confidence:** High - The incremental change is minimal and unambiguous. The `scopeClause` string is used in two SQL queries that both alias memories as `m` — one single-table query (`FROM memories m`) and one join (`FROM memory_entity_mentions mem JOIN memories m`). Without the `m.` prefix, the join query would produce an 'ambiguous column name' error at runtime. The fix is correct. All issues flagged in the prior three review rounds have been addressed.

PR #319, opened 2026-03-24
fix: recover stuck processing summary_jobs on startup + fix embed backfill infinite cycle

ddasgupta4:
## Problem

Two separate data-loss bugs found while dogfooding v0.76.3.

> **AI assistance disclosure:** This PR was authored with AI assistance (George / Claude Opus 4 via OpenClaw), operating on a live v0.76.3 install with 3970 memories and 112 summary_jobs. Bugs were identified by direct database inspection and source code analysis. Fix approach reviewed and approved by the human operator before submission.

---

### Bug 1: `summary_jobs` permanently stuck in `processing` after daemon crash

**Impact:** Every daemon restart abandons in-flight summary jobs. The worker only polls `status = 'pending'` — anything left in `'processing'` is silently lost forever.

Observed: 58 of 112 summary_jobs stuck in processing on a live install after normal use.

**Root cause:** `startSummaryWorker()` starts polling with no crash recovery step. The per-job error handler (within a single run) works correctly, but nothing runs at startup to clean up jobs from a previous crashed process.

**Fix:** Add startup crash recovery at the top of `startSummaryWorker()` — single write tx, non-fatal if it fails — that resets `'processing'` → `'pending'` (or `'dead'` if `attempts >= max_attempts`).

---

### Bug 2: `embed backfill` / embedding tracker cycles infinitely on duplicate-hash memories

**Impact:** Two memory rows with identical content create a permanent "missing embeddings" loop. `signet embed audit` always reports N missing regardless of how many times backfill runs.

Observed: 75 memories permanently reported as missing; `signet embed backfill` reports `re-embedded 50 of 50 (75 still missing)` on every run with no convergence.

**Root cause:** `embeddings` has a `UNIQUE(content_hash)` constraint. The backfill write uses `ON CONFLICT(content_hash) DO UPDATE SET source_id = excluded.source_id`. When two memories share a hash:
1. Embed A → `{hash=H, source_id=A}`
2. Embed B → conflict reassigns `source_id` to B
3. A has no `source_id` match → shows as missing again
4. Repeat forever

Both the audit (`getEmbeddingGapStats`) and backfill SELECT queries join on `source_id = m.id` only, so they never see the hash-attributed embedding that covers the other memory.

**Fix:** Change the `LEFT JOIN` in `getEmbeddingGapStats`, `reembedMissingMemoriesBatch`, and the embedding tracker's stale-row query to also match on `e.content_hash = m.content_hash`. A memory is "embedded" if either its `source_id` is directly referenced OR an embedding with the same content hash exists. The vector is identical for duplicate content so this is semantically correct.

---

## Changes

| File | Change |
|------|--------|
| `packages/daemon/src/pipeline/summary-worker.ts` | Startup crash recovery: reset `processing` → `pending`/`dead`; use `countChanges()` helper and explicit type cast per TS conventions |
| `packages/daemon/src/repair-actions.ts` | Fix JOIN in `getEmbeddingGapStats` and `reembedMissingMemoriesBatch` |
| `packages/daemon/src/embedding-tracker.ts` | Fix JOIN in background tracker stale-row query |
| `packages/daemon/src/repair-actions.test.ts` | Add regression test for duplicate-hash backfill cycle; import `getEmbeddingGapStats` |

---

## Readiness Checklist

> **Note on local checks:** `bun run typecheck` and `bun run lint` were not run in the contributing environment (no bun install available). TypeScript conventions from CONTRIBUTING.md were followed manually: `countChanges()` helper used instead of direct property access, explicit type casts on `db.get()` results, no non-null assertions, proper error narrowing in catch blocks. Please flag any violations in review.

### DIFF
--- a/packages/daemon/src/pipeline/summary-worker.ts
+++ b/packages/daemon/src/pipeline/summary-worker.ts
@@ -1169,25 +1169,23 @@ export function startSummaryWorker(accessor: DbAccessor): SummaryWorkerHandle {
 		accessor.withWriteTx((db) => {
 			// Defensive for tests and partially upgraded workspaces that may
 			// reach the worker before migration 009 has created summary_jobs.
-			const tableExists = db
+			const table = db
 				.prepare("SELECT name FROM sqlite_master WHERE type = 'table' AND name = 'summary_jobs'")
 				.get() as { name: string } | undefined;
-			if (!tableExists) return;
+			if (!table) return;
 
-			const result: unknown = db
-				.prepare(
+			const n = countChanges(
+				db.prepare(
 					`UPDATE summary_jobs
 					 SET status = CASE
 					   WHEN attempts >= max_attempts THEN 'dead'
 					   ELSE 'pending'
 					 END
 					 WHERE status = 'processing'`,
-				)
-				.run();
-
-			const changed = countChanges(result);
-			if (changed > 0) {
-				logger.info("summary-worker", `Crash recovery: reset ${changed} stuck processing job(s) to pending/dead`);
+				).run(),
+			);
+			if (n > 0) {
+				logger.info("summary-worker", `Crash recovery: reset ${n} stuck job(s) to pending/dead`);
 			}
 		});
 	} catch (e) {


PR-Reviewer-Ant:
Hi @ddasgupta4 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: recover stuck processing summary_jobs on startup + fix embed backfill infinite cycle` (commit `483d55ca`) now and I'll follow up shortly with feedback.

packages/daemon/src/repair-actions.ts
```diff
@@ -474,10 +486,19 @@ async function reembedMissingMemoriesBatch(
...
 				 FROM memories m
 				 LEFT JOIN embeddings e
-				   ON e.source_type = 'memory' AND e.source_id = m.id
+				   ON e.source_type = 'memory'
+				   AND (
+				     e.source_id = m.id
+				     OR (m.content_hash IS NOT NULL AND e.content_hash = m.content_hash)
+				   )
 				 WHERE m.is_deleted = 0 AND e.id IS NULL
 				 ORDER BY m.created_at ASC
```
PR-Reviewer-Ant, line 547: The UPSERT in `reembedMissingMemoriesBatch` still includes `source_id = excluded.source_id` in its ON CONFLICT clause. When both A and B (same hash) are selected in the same batch pass, processing B after A will flip `source_id` from A to B. The SELECT-side fix makes gap detection tolerate this, but it introduces a silent invariant: `embeddings.source_id` no longer reliably identifies which memory row 'owns' the vector. The `embedding-tracker.ts` UPSERT deliberately omits `source_id` from its ON CONFLICT update (keeps the original winner). Consider aligning the repair-actions UPSERT to the same policy — drop `source_id = excluded.source_id` from the ON CONFLICT block — to keep ownership stable and avoid the source_id flip on every backfill of duplicate-hash pairs.

packages/daemon/src/repair-actions.test.ts
```diff
@@ -504,6 +505,51 @@ describe("reembedMissingMemories", () => {
...
+		const secondPass = await reembedMissingMemories(
+			accessor, TEST_CFG, CTX_OPERATOR, limiter2,
+			async () => [0.7, 0.8, 0.9],
+			TEST_EMBEDDING_CFG, 10, false,
+		);
+		expect(secondPass.message).toMatch(/no unembedded memories found/);
+	});
+
 	it("can sweep all missing embeddings across multiple batches in one run", async () => {
 		const now = new Date().toISOString();
```
PR-Reviewer-Ant, line 585: The new test does not call `ensureVecTable(db)`, unlike `"syncs vec row using canonical embedding id on hash conflict"` which drops and recreates `vec_embeddings` as a plain table because sqlite-vec virtual tables may not be available in test environments. `reembedMissingMemoriesBatch` calls `syncVecInsert` which writes to `vec_embeddings`. If migrations create it as a virtual table that fails in the test runner, the write transaction will throw and `reembedMissingMemories` will return `success: false`, causing `after.unembedded` to remain 2 and failing the assertion. Add `ensureVecTable(db)` before the test body (or in a nested `beforeEach`) to be consistent with the existing pattern in this suite.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -639,6 +639,44 @@ export function startSummaryWorker(
 		}, delay);
 	}
 
+	// Crash recovery: on startup, reset any jobs that were left in 'processing'
+	// state when the daemon was killed or crashed mid-job. Without this, those
+	// jobs are silently abandoned — the worker only polls for 'pending' jobs, so
+	// anything stuck in 'processing' is never retried. Jobs that have consumed
+	// all their attempts are promoted to 'dead'; all others go back to 'pending'.
+	try {
+		accessor.withWriteTx((db) => {
```
PR-Reviewer-Ant, line 1169: Minor: the `sqlite_master` table-existence check guards against running on a fresh database before migrations. This is reasonable, but the guard is a second defensive layer on top of whatever migration gate already prevents the worker from starting before migrations complete. If the table check is genuinely needed (i.e. the worker can start before migrations), consider adding a comment explaining that invariant; otherwise this is dead code that could silently mask a startup ordering bug.

NicholaiVogel:
rebased this PR onto current `main` and force-pushed the updated branch.

addressed the review feedback:
- dropped `source_id = excluded.source_id` from the repair UPSERT so duplicate-hash backfills keep stable embedding ownership
- updated `cleanOrphanedEmbeddings()` to preserve embeddings that still cover an active memory via the same `content_hash`
- hardened the regression coverage in `repair-actions.test.ts`, including `ensureVecTable()` setup and new assertions around ownership stability / orphan cleanup
- added a clarifying comment on the `summary_jobs` table-existence guard in `summary-worker.ts`

validation:
- `bun test packages/daemon/src/repair-actions.test.ts --test-name-pattern 'syncs vec row|does not cycle|cleanOrphanedEmbeddings'`
- `bunx biome check packages/daemon/src/repair-actions.ts packages/daemon/src/repair-actions.test.ts packages/daemon/src/pipeline/summary-worker.ts`

repo-wide `bun run typecheck` and `bun run lint` are still failing on unrelated baseline issues outside this PR branch diff.

PR #332, opened 2026-03-25
fix(daemon): keep startup recovery responsive on large databases

### LINKED ISSUE #331: v0.76.6 daemon startup hangs on large databases (HTTP server unresponsive)
## Summary

After upgrading from v0.76.3 to v0.76.6, the daemon starts (binds port 3850) but the HTTP server is completely unresponsive — all requests time out. Rolling back to v0.76.3 restores normal operation immediately.

## Environment

- **OS:** Ubuntu 24.04 (Linux 6.8.0)
- **signet version:** 0.76.6 (rollback to 0.76.3 working)
- **DB size:** 237MB SQLite (`memories.db`)
- **Scale:** ~54,000 entities, ~10,700 memories, ~7,400 completed jobs, ~790 dead jobs

## Behavior

1. `systemctl --user start signet` — daemon starts, port 3850 binds successfully
2. `pgrep bun` confirms process is running
3. `ss -tlnp | grep 3850` confirms `127.0.0.1:3850` is listening
4. All HTTP requests to `http://127.0.0.1:3850/health` time out (curl exit code 28)
5. Process never becomes responsive — no amount of waiting resolves it
6. `signet status` hangs indefinitely

## Suspected cause

PR #319 (fix: recover stuck processing summary_jobs on startup + fix embed backfill infinite cycle) introduced startup recovery logic. On a large database, this recovery routine appears to block the HTTP server from accepting connections — likely running synchronously before the HTTP listener is ready, or holding a DB lock that the HTTP handler is waiting on.

## Reproduction

On a database with 237MB+/54k+ entities, upgrade from v0.76.3 to v0.76.6 and attempt to start the daemon.

## Workaround

Downgrade to v0.76.3:
```bash
npm install -g signetai@0.76.3
```

## Additional context

v0.76.3 starts cleanly and becomes healthy in under 15 seconds on the same database. The startup hang is 100% reproducible on v0.76.6.

NicholaiVogel:
## Summary

Fixes the startup responsiveness regression reported in #331.

The main change is to stop doing large synchronous startup recovery work on the daemon's main thread. This PR batches `summary_jobs` crash recovery off the synchronous startup path and replaces duplicate-hash embedding coverage queries with shared `EXISTS`-based helpers that are friendlier to SQLite on large databases.

Closes #331.

## Changes

- added shared embedding coverage helpers in `packages/daemon/src/embedding-coverage.ts`
- rewired `embedding-tracker` and `repair-actions` to use shared duplicate-hash coverage logic instead of repeating `LEFT JOIN ... OR ...` queries
- changed summary worker crash recovery to run in bounded async batches after startup begins, instead of one synchronous startup sweep
- addressed review follow-up by making batch rescheduling depend on selected row count, not updated row count, and clamping exported recovery limits to a sane positive range
- added matching daemon-rs summary startup recovery behavior in `packages/daemon-rs/crates/signet-pipeline/src/summary.rs`
- added regression tests for batched summary recovery and duplicate-hash embedding coverage, including a focused rust parity test for summary recovery batching
- added incident-driven research/planning docs and registered them in `INDEX.md` + `dependencies.yaml`

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs/spec system, `packages/daemon-rs`

## Screenshots

N/A

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Targeted validation completed:

- `bun test packages/daemon/src/pipeline/summary-worker.test.ts`
- `bun test packages/daemon/src/embedding-coverage.test.ts`
- `bun test packages/daemon/src/repair-actions.test.ts`
- `cargo test -p signet-pipeline summary::tests::recovers_leased_summary_jobs_in_batches`
- `bunx biome check packages/daemon/src/embedding-coverage.ts packages/daemon/src/embedding-coverage.test.ts packages/daemon/src/embedding-tracker.ts packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/repair-actions.ts docs/research/technical/RESEARCH-DAEMON-STARTUP-RESPONSIVENESS.md docs/specs/planning/daemon-startup-responsiveness.md docs/specs/INDEX.md docs/specs/dependencies.yaml`

Notes on unchecked boxes:

- `bun run typecheck` is currently failing on unrelated pre-existing workspace issues, including `@signet/opencode-plugin` and broader daemon typing debt outside this patch.
- `scripts/spec-deps-check.ts` also reports pre-existing missing `references/*` paths for `competitive-systems-research`, unrelated to this PR.

## AI disclosure

## Notes

- The issue report suspected the new `summary_jobs` recovery logic directly blocked startup before the listener was ready. After tracing the path, the more likely regression was the duplicate-hash embedding coverage query introduced in the same release window. This PR hardens both sides.
- The new embedding coverage helpers intentionally preserve the existing daemon-wide maintenance behavior. These queries are repair/tracker reads over backlog state, not request-scoped agent recall paths.

### DIFF
--- a/packages/daemon/src/pipeline/summary-worker.ts
+++ b/packages/daemon/src/pipeline/summary-worker.ts
@@ -969,7 +969,7 @@ export function recoverSummaryJobs(accessor: DbAccessor, limit: number = RECOVER
 			.prepare(
 				`SELECT id, attempts, max_attempts
 				 FROM summary_jobs
-				 WHERE status = 'processing'
+				 WHERE status IN ('processing', 'leased')
 				 ORDER BY created_at ASC
 				 LIMIT ?`,
 			)
@@ -986,7 +986,7 @@ export function recoverSummaryJobs(accessor: DbAccessor, limit: number = RECOVER
 		const update = db.prepare(
 			`UPDATE summary_jobs
 			 SET status = ?
-			 WHERE id = ? AND status = 'processing'`,
+			 WHERE id = ? AND status IN ('processing', 'leased')`,
 		);
 
 		let updated = 0;
--- a/packages/daemon-rs/crates/signet-pipeline/src/summary.rs
+++ b/packages/daemon-rs/crates/signet-pipeline/src/summary.rs
@@ -337,7 +337,7 @@ async fn recover_summary_jobs(pool: &DbPool, limit: i64) -> Result<RecoveryBatch
                 let mut stmt = conn.prepare_cached(
                     "SELECT id, attempts, max_attempts
                      FROM summary_jobs
-                     WHERE status = 'leased'
+                     WHERE status IN ('processing', 'leased')
                      ORDER BY created_at ASC
                      LIMIT ?1",
                 )?;
@@ -362,7 +362,7 @@ async fn recover_summary_jobs(pool: &DbPool, limit: i64) -> Result<RecoveryBatch
             let mut stmt = conn.prepare_cached(
                 "UPDATE summary_jobs
                  SET status = ?1
-                 WHERE id = ?2 AND status = 'leased'",
+                 WHERE id = ?2 AND status IN ('processing', 'leased')",
             )?;
 
             let mut updated = 0usize;
@@ -453,13 +453,15 @@ mod tests {
                 let mut stmt = tx.prepare_cached(
                     "INSERT INTO summary_jobs
                      (id, session_key, harness, project, transcript, status, attempts, max_attempts, created_at)
-                     VALUES (?1, ?2, 'codex', NULL, 'transcript', 'leased', ?3, ?4, ?5)",
+                     VALUES (?1, ?2, 'codex', NULL, 'transcript', ?3, ?4, ?5, ?6)",
                 )?;
                 for i in 0..205 {
                     let attempts = (i % 3) as i64;
+                    let status = if i % 2 == 0 { "processing" } else { "leased" };
                     stmt.execute(rusqlite::params![
                         format!("job-{i}"),
                         format!("session-{i}"),
+                        status,
                         attempts,
                         2i64,
                         &now,
@@ -509,18 +511,18 @@ mod tests {
             }
         );
 
-        let leased: i64 = pool
+        let inflight: i64 = pool
             .read(|conn| {
                 conn.query_row(
-                    "SELECT COUNT(*) FROM summary_jobs WHERE status = 'leased'",
+                    "SELECT COUNT(*) FROM summary_jobs WHERE status IN ('processing', 'leased')",
                     [],
                     |row| row.get(0),
                 )
                 .map_err(Into::into)
             })
             .await
-            .expect("failed to count leased jobs");
-        assert_eq!(leased, 0);
+            .expect("failed to count in-flight jobs");
+        assert_eq!(inflight, 0);
 
         let dead: i64 = pool
             .read(|conn| {
--- a/packages/daemon/src/pipeline/summary-worker.test.ts
+++ b/packages/daemon/src/pipeline/summary-worker.test.ts
@@ -103,13 +103,14 @@ describe("recoverSummaryJobs", () => {
 		const stmt = db.prepare(
 			`INSERT INTO summary_jobs
 			 (id, session_key, harness, project, transcript, status, attempts, max_attempts, created_at)
-			 VALUES (?, NULL, 'codex', NULL, 'transcript', 'processing', ?, ?, ?)`,
+			 VALUES (?, NULL, 'codex', NULL, 'transcript', ?, ?, ?, ?)`,
 		);
 
 		for (let i = 0; i < 205; i++) {
 			const attempts = i % 3;
 			const max = 2;
-			stmt.run(`job-${i}`, attempts, max, now);
+			const status = i % 2 === 0 ? "processing" : "leased";
+			stmt.run(`job-${i}`, status, attempts, max, now);
 		}
 
 		expect(recoverSummaryJobs(accessor, 100)).toEqual({ selected: 100, updated: 100 });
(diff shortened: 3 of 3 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): keep startup recovery responsive on large databases` (commit `947cff27`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1160,47 +1202,37 @@ export function startSummaryWorker(accessor: DbAccessor): SummaryWorkerHandle {
...
+					logger.info("summary-worker", `Crash recovery: reset ${n} stuck job(s) to pending/dead`);
+				}
+				if (n >= RECOVER_BATCH) {
+					scheduleRecovery(0);
+				}
+			} catch (e) {
+				logger.warn("summary-worker", "Crash recovery failed (non-fatal)", {
+					error: e instanceof Error ? e.message : String(e),
+				});
 			}
```
PR-Reviewer-Ant, line 1231: The reschedule condition uses the number of rows successfully updated (`n >= RECOVER_BATCH`) rather than whether a full batch was selected. If some selected rows are changed concurrently between the `SELECT` and per-row `UPDATE` calls, `n` can fall below 100 even while more `processing` rows still exist, causing recovery to stop early and leave stuck jobs behind. This is a small but real race in the new batched recovery path.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1160,47 +1202,37 @@ export function startSummaryWorker(accessor: DbAccessor): SummaryWorkerHandle {
...
-					   ELSE 'pending'
-					 END
-					 WHERE status = 'processing'`,
-				).run(),
-			);
-			if (n > 0) {
-				logger.info("summary-worker", `Crash recovery: reset ${n} stuck job(s) to pending/dead`);
+	function scheduleRecovery(delay: number): void {
+		if (stopped) return;
+		recoverTimer = setTimeout(() => {
```
PR-Reviewer-Ant, line 1218: This PR changes daemon behavior in a codepath explicitly covered by the repo's Rust parity rule (`packages/daemon-rs/` must reflect daemon behavioral changes), but the PR marks Rust parity effectively as N/A and no corresponding parity update is present in the diff. That risks shadow divergences during startup recovery and makes the implementation diverge from the repository contract.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -947,6 +949,45 @@ function writeSummaryToDAG(accessor: DbAccessor, job: SummaryJobRow, result: Llm
```
PR-Reviewer-Ant, line 947: This PR changes daemon boot-time behavior by replacing the previous one-shot recovery sweep with deferred batched recovery, but there is no corresponding `packages/daemon-rs/` update in the diff. The repo contract explicitly requires daemon behavioral changes to stay in Rust parity; otherwise the shadow runtime will diverge exactly on the startup path this PR is modifying.

PR-Reviewer-Ant, line 947: `recoverSummaryJobs()` is exported and accepts any numeric `limit`, but it does not clamp or reject zero/negative/out-of-range values even though the repo conventions call out boundary validation for counters/limits. Today the call sites pass safe values, so this is not a blocker, but the shared helper should defend itself before other callers rely on it.

packages/daemon/src/pipeline/summary-worker.ts
NicholaiVogel: good catch. fixed in 8dd7506c.

`recoverSummaryJobs()` now returns both `selected` and `updated`, and the async reschedule condition keys off `selected >= RECOVER_BATCH` instead of updated-row count. that removes the early-stop case if the selected batch and the successful updates ever diverge.

also updated the TS regression test to assert the batched return shape directly.

packages/daemon/src/pipeline/summary-worker.ts
NicholaiVogel: fair call. fixed in 8dd7506c.

I added the corresponding daemon-rs startup recovery path in `packages/daemon-rs/crates/signet-pipeline/src/summary.rs` so the shadow runtime no longer diverges here. rust now does the same bounded startup recovery pass for stuck summary jobs, using `selected` vs `updated` batch accounting for the same stop condition.

I also added a rust regression test:
- `cargo test -p signet-pipeline summary::tests::recovers_leased_summary_jobs_in_batches`

packages/daemon/src/pipeline/summary-worker.ts
NicholaiVogel: yeah, agreed. addressed in 8dd7506c.

this PR now includes the matching `packages/daemon-rs/` behavior change instead of treating parity as effectively n/a. the rust summary worker gets the same startup recovery behavior and a focused regression test so shadow mode should stay aligned on this path.

NicholaiVogel: went ahead and fixed this too in 8dd7506c.

`recoverSummaryJobs()` now clamps the exported `limit` to a sane positive integer range before issuing the query, and the TS test file now covers `0` and `Infinity` inputs.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -947,6 +955,50 @@ function writeSummaryToDAG(accessor: DbAccessor, job: SummaryJobRow, result: Llm
...
+		}>;
+
+		if (rows.length === 0) {
+			return { selected: 0, updated: 0 };
+		}
+
+		const update = db.prepare(
+			`UPDATE summary_jobs
+			 SET status = ?
+			 WHERE id = ? AND status = 'processing'`,
```
PR-Reviewer-Ant: Crash recovery now only resets rows stuck in `status = 'processing'`, but the Rust shadow worker added in this PR recovers `status = 'leased'` instead. Under the repo's Rust parity/shadow-mode contract, this leaves jobs stranded when a DB created by one runtime is later opened by the other, and it undermines the PR's claim that startup recovery is hardened. Recovery needs to handle both persisted status variants, or the runtimes need to share one canonical status model.

NicholaiVogel: yep, agreed. fixed in 9b21e72e.

recovery now treats both persisted in-flight variants as recoverable on both runtimes:
- js: `status IN ('processing', 'leased')`
- rust: `status IN ('processing', 'leased')`

that means a database created by one runtime can restart under the other without stranding summary jobs behind the old status spelling.

validation:
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts`
- `bunx biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts`
- `cargo test -p signet-pipeline summary::tests::recovers_leased_summary_jobs_in_batches`

PR #333, opened 2026-03-25
fix(cli): recover stale daemon processes on restart

NicholaiVogel:
## Summary

Fix the CLI daemon lifecycle path when a managed daemon process is still alive but no longer answers health checks.

Before this change, `signet daemon restart` and `signet daemon stop` could treat the daemon as stopped if `/health` was wedged, even while the old managed process still owned port `3850`. In practice that left updates half-applied: the CLI upgraded, but the old daemon kept running and blocked the new one from starting.

## Changes

- added `hasDaemonProcess()` in `packages/cli/src/lib/runtime.ts` to detect a still-alive managed daemon from `.daemon/pid`, even when health probes fail
- updated `startDaemon()` to clean up a stale managed daemon before spawning a replacement
- updated `doRestart()` to stop stale managed daemons, not just healthy/reachable ones
- updated `doStop()` to attempt cleanup in the same stale-process case instead of immediately reporting "Daemon is not running"
- added CLI regression tests for restart/stop behavior when the daemon process is alive but health checks fail
- updated the `daemon-startup-responsiveness` planning stub, dependency metadata, and INDEX summary to record this guardrail

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs/spec system

## Screenshots

N/A

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Targeted validation completed:

- `bun test packages/cli/src/features/daemon.test.ts`
- `bunx biome check packages/cli/src/lib/runtime.ts packages/cli/src/features/daemon.ts packages/cli/src/cli.ts packages/cli/src/features/daemon.test.ts docs/specs/planning/daemon-startup-responsiveness.md docs/specs/dependencies.yaml docs/specs/INDEX.md`
- `cd packages/cli && bun run build`
- runtime sanity check with built CLI: `node packages/cli/dist/cli.js daemon status`
- reproduced the real stale-daemon failure mode locally, then verified manual recovery and encoded the behavior into CLI regression tests

Notes on unchecked boxes:

- `bun run typecheck` was not run for this focused fix.
- `bun test` / `bun run lint` were not run across the full workspace; validation was targeted to touched CLI/spec files.

## AI disclosure

## Notes

- Incident shape: global install updated the CLI to `0.77.1`, but the old daemon process remained alive as `0.77.0`, pinned CPU, still owned port `3850`, and caused `signet daemon restart` to hang or misreport state.
- This fix is intentionally scoped to managed-process cleanup in the CLI. It does not redesign daemon health detection or service-manager flows.


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): recover stale daemon processes on restart` (commit `1925fc44`) now and I'll follow up shortly with feedback.

packages/cli/src/lib/runtime.ts
```diff
@@ -109,6 +109,34 @@ export async function isDaemonRunning(): Promise<boolean> {
 	return urls.length > 0;
 }
 
+export async function hasDaemonProcess(agentsDir: string = AGENTS_DIR): Promise<boolean> {
```
PR-Reviewer-Ant: Blocking: `hasDaemonProcess()` now treats any live PID found in `.daemon/pid` as a stale managed daemon. On PID reuse, that file can point at an unrelated process, and the new callers in `doStop()` / `doRestart()` / `startDaemon()` will then send SIGTERM/SIGKILL to it via `stopDaemon()`. Please verify process identity (for example via stored start token/command metadata or port ownership) before killing a PID; liveness alone is not enough.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `1925fc44`

[Automated Review] Found one blocking process-safety issue in stale-daemon recovery.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new recovery path treats any live PID in `.daemon/pid` as a managed daemon and then calls `stopDaemon()`, which sends SIGTERM/SIGKILL to that PID. There is no validation that the PID still belongs to Signet, so normal PID reuse can now terminate an unrelated process. This behavior is introduced directly by the diff in `packages/cli/src/lib/runtime.ts` and consumed from the new branches in `packages/cli/src/features/daemon.ts`.

packages/cli/src/lib/runtime.ts
NicholaiVogel: fixed in ec138a62.

stale cleanup no longer trusts pid-file liveness on its own. `readManagedDaemonPid()` now re-reads the live process command line and only treats the pid as managed when it still matches one of the signet daemon entrypoints we actually spawn.

if the pid was reused by some unrelated process, we leave the pid file alone and skip the kill path. i also added `runtime.test.ts` coverage for both cases:
- matching daemon command is accepted
- reused live pid with a non-daemon command is rejected

validation i ran:
- `bun test packages/cli/src/lib/runtime.test.ts packages/cli/src/features/daemon.test.ts`
- `bunx biome check packages/cli/src/lib/runtime.ts packages/cli/src/lib/runtime.test.ts`
- `cd packages/cli && bun run build`

packages/cli/src/lib/runtime.ts
```diff
@@ -33,11 +33,30 @@ interface DaemonInstance {
...
 }
 
+interface DaemonProbeDeps {
+	readonly daemonPaths?: readonly string[];
+	readonly isAlive?: (pid: number) => boolean;
+	readonly readCmd?: (pid: number) => string | null;
+}
+
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
```
PR-Reviewer-Ant, line 43: Blocking: `daemonPaths()` only whitelists daemon file paths for the *current* checkout/install. That means `readManagedDaemonPid()` will reject a still-running daemon launched from an older global package path, which is exactly the stale-upgrade scenario this PR claims to fix. After a CLI upgrade, the old process command line can legitimately reference the previous version's install/store path, so `hasDaemonProcess()` returns false and `start`/`restart`/`stop` still won't recover the wedged managed daemon. The detection needs a version-agnostic way to identify a Signet daemon process, or at least coverage for old install paths.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `ec138a62`

[Automated Review] Found one blocking regression in the stale-daemon detection path.
The PR addresses the previously flagged PID-reuse safety issue, but the new command-path validation is too strict to reliably detect the real upgrade scenario described in the PR.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new detection only accepts a managed PID when its command line contains one of the current install's hard-coded daemon paths from runtime.ts. In the incident described by the PR, the stale process is specifically an older daemon left behind after a CLI upgrade. Global package managers commonly launch binaries from versioned/store paths, so the old daemon command will often point at the previous package location rather than any path in daemonPaths(). In that case hasDaemonProcess() returns false, doStop()/doRestart() skip cleanup, and the original 'old daemon still owns port 3850 after upgrade' failure mode remains.

packages/cli/src/lib/runtime.ts
NicholaiVogel: fixed in 4c38e6ff.

good catch. the previous pass was too tied to the current checkout/install path, which misses the exact stale-upgrade case this PR is for.

`readManagedDaemonPid()` now accepts version-agnostic signet daemon path markers in addition to the current absolute paths, so an older managed process still matches when its command line points at:
- `.../signetai/dist/daemon.js`
- `.../packages/daemon/dist/daemon.js`
- `.../packages/daemon/src/daemon.ts`

that keeps the pid verification strict enough to avoid unrelated reused pids, but broad enough to reclaim daemons launched from a previous global install or a different checkout path.

added regression coverage for an older global install path too.

validation:
- `bun test packages/cli/src/lib/runtime.test.ts packages/cli/src/features/daemon.test.ts`
- `bunx biome check packages/cli/src/lib/runtime.ts packages/cli/src/lib/runtime.test.ts packages/cli/src/features/daemon.ts packages/cli/src/features/daemon.test.ts packages/cli/src/cli.ts`
- `cd packages/cli && bun run build`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `4c38e6ff`

[Automated Review] No new issues found in the current diff; the stale-process recovery path now validates the PID before killing it and also recognizes older global-install daemon paths. Ready for human review.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The previously blocking concerns were both addressed directly in the touched code. `readManagedDaemonPid()` now verifies that a live PID still looks like a Signet daemon before reclaiming it, and `matchesDaemon()` was widened via `daemonMarks()` to match older global-install paths like the versioned Bun cache path shown in the PR incident. The incremental diff also adds a focused regression test for that older-path case in `packages/cli/src/lib/runtime.test.ts`.

PR #338, opened 2026-03-26
fix(daemon): broaden macOS SQLite runtime discovery for sqlite-vec

### LINKED ISSUE #336: Daemon (Bun) cannot load sqlite-vec — vector search silently degrades to keyword-only
## Summary

The Signet daemon runs under Bun (`bun /path/to/daemon.js`), but Bun's built-in SQLite does not support `loadExtension()`. This means the `sqlite-vec` extension never loads, the `vec_embeddings` virtual table is never created, and all vector search silently falls back to keyword-only.

## Environment

- signet v0.77.0
- macOS (Darwin 25.3.0, arm64)
- Node.js v22.22.1 (MCP server), Bun (daemon)
- `sqlite-vec-darwin-arm64` package is installed and `vec0.dylib` is present

## Reproduction

1. Fresh `signet setup` → `signet start`
2. Store a memory: `signet remember "test memory"`
3. Check: `sqlite3 memories.db ".tables"` — no `vec_embeddings` table
4. `startup.log` shows:
   ```
   [db-accessor] loadExtension failed: This build of sqlite3 does not support dynamic extension loading
   Vector search failed: SQLiteError: no such table: vec_embeddings
   ```

## Root Cause

In `daemon.js`, the DB initialization path is:

```
loadVecExtension(writeConn)  // fails silently under Bun
...
if (vecExtPath) {            // false, because extension didn't load
  ensureVecTable(writeConn)  // skipped — table never created
}
```

The daemon process is `bun daemon.js` (confirmed via `ps aux`). Bun's native SQLite (`bun:sqlite`) does not support dynamic extension loading. The extension file exists and loads fine under Node.js with `better-sqlite3`.

The MCP server (`mcp-stdio.js`) runs under Node but proxies all queries through the daemon's HTTP API, so it cannot compensate.

## Impact

- `hybridRecall` falls back to keyword-only search on every prompt submit
- `memory_search` via MCP returns `"source": "keyword"` instead of hybrid results
- `signet migrate-vectors` (Node CLI) can create the table, but the daemon still can't query it

## Suggested Fix

Either:
1. Run the daemon under Node instead of Bun (shebang already says `#!/usr/bin/env node`)
2. Use Bun's native vector support if/when available
3. Have the MCP server (Node) handle DB queries directly instead of proxying through the daemon

NicholaiVogel:
## Summary

Fixes #336.

This keeps Bun as the daemon runtime and fixes the actual macOS failure mode instead of trying to route around it by switching the daemon to Node.

### What changed

- broadened macOS custom SQLite discovery order in `db-accessor.ts`
  - `SIGNET_SQLITE_PATH` as an explicit override
  - `$SIGNET_PATH/libsqlite3.dylib`
  - Homebrew SQLite dylib paths
- made `SIGNET_SQLITE_PATH` authoritative when set
  - if it points at a missing file, Signet records the attempted path and refuses fallback so the misconfiguration is obvious
  - fallback to workspace/Homebrew candidates only happens when the env override is unset
- moved custom SQLite selection into daemon init so it happens before the first Bun DB connection
- tracked vector runtime status so degraded mode is explicit instead of effectively silent
- improved embedding health guidance when `vec_embeddings` is unavailable
- fixed `buildEmbeddingHealth()` so it no longer throws when `vec_embeddings` is missing
- tightened vec health diagnostics so unrelated SQL failures are surfaced as query/integrity errors instead of being mislabeled as missing-table setup issues
- added regression tests for macOS SQLite discovery order, authoritative override handling, and degraded-mode messaging
- added the required research/spec stub and synced `INDEX.md` / `dependencies.yaml`
- documented `SIGNET_SQLITE_PATH` in runtime docs
- cleaned up stale dependency metadata so `bun scripts/spec-deps-check.ts` passes again

## Why this approach

Issue #336 described a real bug, but the root cause was narrower than “Bun cannot load sqlite-vec.”

The daemon already had a Homebrew-only macOS fallback. The actual gap was that users without that exact Homebrew layout still fell back to Apple’s system SQLite, which can block `loadExtension()`.

This PR fixes that gap directly and keeps the daemon architecture intact.

## Validation

Passed:

- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bunx @biomejs/biome check packages/daemon/src/db-accessor.ts packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.ts packages/daemon/src/embedding-health.test.ts docs/CLI.md docs/CONFIGURATION.md docs/DAEMON.md docs/SELF-HOSTING.md docs/specs/INDEX.md docs/specs/dependencies.yaml docs/research/technical/RESEARCH-MACOS-SQLITE-RUNTIME-DISCOVERY.md docs/specs/planning/macos-sqlite-runtime-discovery.md`

Known pre-existing repo issue during validation:

- `bunx tsc --noEmit -p packages/daemon/tsconfig.json` still reports a large set of unrelated existing type errors outside this change set

## Notes

This PR does not bundle a first-party SQLite dylib yet. It just makes the daemon discover supported macOS dylibs in a sane order, fail clearly on a bad explicit override, and tell the operator what to do when vector runtime setup is missing.

### DIFF
--- a/packages/daemon/src/db-accessor.ts
+++ b/packages/daemon/src/db-accessor.ts
@@ -38,6 +38,12 @@ export interface VectorRuntimeStatus {
 	readonly extensionLoadError: string | null;
 }
 
+interface SqliteRuntimeConfig {
+	readonly choice: SqliteChoice | null;
+	readonly attempt: string | null;
+	readonly warning: string | null;
+}
+
 // ---------------------------------------------------------------------------
 // Public interfaces — thin wrappers over the bun:sqlite Database surface
 // ---------------------------------------------------------------------------
@@ -166,6 +172,16 @@ export function resolveCustomSqlitePath(opts?: {
 	return null;
 }
 
+function resolveHomebrewSqlitePath(exists: (path: string) => boolean): SqliteChoice | null {
+	for (const path of HOMEBREW_SQLITE_PATHS) {
+		if (exists(path)) {
+			return { path, source: "homebrew" };
+		}
+	}
+
+	return null;
+}
+
 function explainSqliteSetup(agentsDir: string): string {
 	return [
 		"macOS system SQLite may block loadExtension() and force keyword-only recall.",
@@ -173,36 +189,96 @@ function explainSqliteSetup(agentsDir: string): string {
 	].join(" ");
 }
 
-function configureCustomSqlite(): void {
-	sqliteChoice = null;
-	sqliteAttempt = null;
-	sqliteWarning = null;
-	if (process.platform !== "darwin") return;
-
-	const agentsDir = resolveSqliteAgentsDir();
-	const envPath = process.env.SIGNET_SQLITE_PATH;
-	if (envPath && !existsSync(envPath)) {
-		sqliteAttempt = envPath;
-		sqliteWarning = `SIGNET_SQLITE_PATH does not exist: ${envPath}. Explicit override is authoritative, refusing fallback to workspace/Homebrew SQLite.`;
-		console.warn(`[db-accessor] ${sqliteWarning}`);
-		return;
+export function resolveSqliteRuntimeConfig(opts?: {
+	readonly platform?: NodeJS.Platform;
+	readonly env?: NodeJS.ProcessEnv;
+	readonly agentsDir?: string;
+	readonly exists?: (path: string) => boolean;
+	readonly set?: (path: string) => void;
+}): SqliteRuntimeConfig {
+	const platform = opts?.platform ?? process.platform;
+	if (platform !== "darwin") {
+		return {
+			choice: null,
+			attempt: null,
+			warning: null,
+		};
 	}
 
-	const choice = resolveCustomSqlitePath({ agentsDir });
+	const env = opts?.env ?? process.env;
+	const exists = opts?.exists ?? existsSync;
+	const set = opts?.set ?? ((path: string) => Database.setCustomSQLite(path));
+	const agentsDir = opts?.agentsDir ?? resolveSqliteAgentsDir({ env });
+	const envPath = env.SIGNET_SQLITE_PATH;
+	if (envPath && !exists(envPath)) {
+		return {
+			choice: null,
+			attempt: envPath,
+			warning: `SIGNET_SQLITE_PATH does not exist: ${envPath}. Explicit override is authoritative, refusing fallback to workspace/Homebrew SQLite.`,
+		};
+	}
+
+	const choice = resolveCustomSqlitePath({ platform, env, agentsDir, exists });
 	if (!choice) {
-		sqliteWarning = explainSqliteSetup(agentsDir);
-		console.warn(`[db-accessor] ${sqliteWarning}`);
-		return;
+		return {
+			choice: null,
+			attempt: null,
+			warning: explainSqliteSetup(agentsDir),
+		};
 	}
 
-	sqliteAttempt = choice.path;
 	try {
-		Database.setCustomSQLite(choice.path);
-		sqliteChoice = choice;
+		set(choice.path);
+		return {
+			choice,
+			attempt: choice.path,
+			warning: null,
+		};
 	} catch (e) {
 		const msg = e instanceof Error ? e.message : String(e);
-		sqliteWarning = `Failed to activate custom SQLite at ${choice.path}: ${msg}. ${explainSqliteSetup(agentsDir)}`;
-		console.warn(`[db-accessor] setCustomSQLite(${choice.path}) skipped: ${sqliteWarning}`);
+		if (choice.source !== "workspace") {
+			return {
+				choice: null,
+				attempt: choice.path,
+				warning: `Failed to activate custom SQLite at ${choice.path}: ${msg}. ${explainSqliteSetup(agentsDir)}`,
+			};
+		}
+
+		const fallback = resolveHomebrewSqlitePath(exists);
+		if (fallback === null || fallback.path === choice.path) {
+			return {
+				choice: null,
+				attempt: choice.path,
+				warning: `Failed to activate custom SQLite at ${choice.path}: ${msg}. ${explainSqliteSetup(agentsDir)}`,
+			};
+		}
+
+		try {
+			set(fallback.path);
+			console.warn(`[db-accessor] workspace SQLite at ${choice.path} failed (${msg}), fell back to ${fallback.path}`);
+			return {
+				choice: fallback,
+				attempt: fallback.path,
+				warning: null,
+			};
+		} catch (err) {
+			const next = err instanceof Error ? err.message : String(err);
+			return {
+				choice: null,
+				attempt: fallback.path,
+				warning: `Failed to activate workspace SQLite at ${choice.path}: ${msg}. Fallback Homebrew SQLite at ${fallback.path} also failed: ${next}. ${explainSqliteSetup(agentsDir)}`,
+			};
+		}
+	}
+}
+
+function configureCustomSqlite(agentsDir?: string): void {
+	const cfg = resolveSqliteRuntimeConfig({ agentsDir });
+	sqliteChoice = cfg.choice;
+	sqliteAttempt = cfg.attempt;
+	sqliteWarning = cfg.warning;
+	if (cfg.warning !== null) {
+		console.warn(`[db-accessor] ${cfg.warning}`);
 	}
 }
 
@@ -282,7 +358,7 @@ function backupBeforeMigration(db: Database, dbPath: string, schemaVersion: numb
  * before any route handler runs. Ensures the memory directory exists, opens
  * the write connection, sets pragmas, and runs pending migrations.
  */
-export function initDbAccessor(path: string): void {
+export function initDbAccessor(path: string, opts?: { readonly agentsDir?: string }): void {
 	if (accessor) {
 		throw new Error("DbAccessor already initialised");
 	}
@@ -294,7 +370,7 @@ export function initDbAccessor(path: string): void {
 
 	dbPath = path;
 
-	configureCustomSqlite();
+	configureCustomSqlite(opts?.agentsDir);
 
 	const writeConn = new Database(path);
 	configurePragmas(writeConn);
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -11669,7 +11669,7 @@ async function main() {
 	// runs migrations). This is the sole schema authority and the only
 	// production Bun SQLite constructor path, which keeps macOS custom
 	// SQLite selection ahead of any live connection.
-	initDbAccessor(MEMORY_DB);
+	initDbAccessor(MEMORY_DB, { agentsDir: AGENTS_DIR });
 
 	// Sync agent roster from manifest into the agents table.
 	syncAgentRoster(AGENTS_DIR);
--- a/packages/daemon/src/db-accessor.test.ts
+++ b/packages/daemon/src/db-accessor.test.ts
@@ -11,6 +11,7 @@ import {
 	initDbAccessor,
 	resolveCustomSqlitePath,
 	resolveSqliteAgentsDir,
+	resolveSqliteRuntimeConfig,
 } from "./db-accessor";
 
 function tmpDbPath(): string {
@@ -219,6 +220,53 @@ describe("resolveCustomSqlitePath", () => {
 		});
 
 		expect(result).toBeNull();
+	});
+
+	test("falls back to Homebrew when workspace sqlite exists but fails activation", () => {
+		const found = new Set(["/tmp/agents/libsqlite3.dylib", "/usr/local/opt/sqlite/lib/libsqlite3.dylib"]);
+		const calls: string[] = [];
+		const cfg = resolveSqliteRuntimeConfig({
+			platform: "darwin",
+			agentsDir: "/tmp/agents",
+			env: {},
+			exists: (path) => found.has(path),
+			set: (path) => {
+				calls.push(path);
+				if (path === "/tmp/agents/libsqlite3.dylib") {
+					throw new Error("wrong architecture");
+				}
+			},
+		});
+
+		expect(calls).toEqual(["/tmp/agents/libsqlite3.dylib", "/usr/local/opt/sqlite/lib/libsqlite3.dylib"]);
+		expect(cfg).toEqual({
+			choice: {
+				path: "/usr/local/opt/sqlite/lib/libsqlite3.dylib",
+				source: "homebrew",
+			},
+			attempt: "/usr/local/opt/sqlite/lib/libsqlite3.dylib",
+			warning: null,
+		});
+	});
+
+	test("uses the explicit agentsDir passed to init-time sqlite resolution", () => {
+		const found = new Set(["/tmp/explicit/libsqlite3.dylib"]);
+		const cfg = resolveSqliteRuntimeConfig({
+			platform: "darwin",
+			agentsDir: "/tmp/explicit",
+			env: { SIGNET_PATH: "/tmp/env-workspace" },
+			exists: (path) => found.has(path),
+			set: () => {},
+		});
+
+		expect(cfg).toEqual({
+			choice: {
+				path: "/tmp/explicit/libsqlite3.dylib",
+				source: "workspace",
+			},
+			attempt: "/tmp/explicit/libsqlite3.dylib",
+			warning: null,
+		});
 	});
 });
 


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): broaden macOS SQLite runtime discovery for sqlite-vec` (commit `3d218e7d`) now and I'll follow up shortly with feedback.

packages/daemon/src/db-accessor.ts
```diff
@@ -94,25 +86,105 @@ function configurePragmas(db: Database): void {
...
+	}
+
+	const choice = resolveCustomSqlitePath({ agentsDir });
+	if (!choice) {
+		sqliteWarning = explainSqliteSetup(agentsDir);
+		console.warn(`[db-accessor] ${sqliteWarning}`);
+		return;
+	}
+
+	try {
```
PR-Reviewer-Ant, line 230: If `Database.setCustomSQLite()` throws because the chosen dylib is invalid/incompatible, this code still records `sqliteChoice = choice`. Downstream health reporting then treats a custom SQLite as selected even though Bun is still on the system SQLite, which suppresses the macOS remediation branch in `vecFix()` and makes degraded mode misleading. Only mark the choice active after `setCustomSQLite()` succeeds, or track failed selection separately.

packages/daemon/src/db-accessor.ts
```diff
@@ -168,17 +240,18 @@ export function initDbAccessor(path: string): void {
```
PR-Reviewer-Ant, line 370: This changes daemon startup/runtime-selection behavior and embedding-health observability, but the repo contract says `packages/daemon-rs/` must be kept in behavioral parity with `packages/daemon/`. I do not see any corresponding rust-side update in this PR. Shipping the JS change alone will create shadow divergences on macOS and weaken the value of the parity proxy.

packages/daemon/src/db-accessor.ts
NicholaiVogel: good catch. fixed in 9651250e.

changes:
- `sqliteChoice` is now only set after `Database.setCustomSQLite()` succeeds
- added `sqliteAttempt` to runtime status so degraded-mode detail still shows which dylib was attempted
- failed activation now keeps the macOS remediation path active via `sqliteWarning` instead of pretending a custom sqlite is live

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

packages/daemon/src/db-accessor.ts
NicholaiVogel: yep, fair call. i added a rust-side parity follow-up in 9651250e.

what changed on `packages/daemon-rs/`:
- documented the parity invariant in `signet-core/src/db.rs`
- added a focused rust test, `vec_is_available_without_external_sqlite_selection`, that locks the equivalent behavior on the shadow daemon side

that captures the reason the rust path does not need macOS dylib discovery: it uses bundled SQLite plus statically registered `sqlite-vec` before any connection opens.

validated with:
- `cargo test --manifest-path packages/daemon-rs/Cargo.toml -p signet-core vec_is_available_without_external_sqlite_selection -- --nocapture`

NicholaiVogel:
addressed the two review blockers in 9651250e.

- fixed degraded-mode reporting so failed `setCustomSQLite()` activation no longer marks a custom sqlite as active
- added `sqliteAttempt` for clearer health detail when activation fails
- added a rust-side parity follow-up in `packages/daemon-rs/crates/signet-core/src/db.rs`, including a focused test proving the shadow daemon has sqlite-vec without external dylib selection

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`
- `cargo test --manifest-path packages/daemon-rs/Cargo.toml -p signet-core vec_is_available_without_external_sqlite_selection -- --nocapture`

packages/daemon/src/db-accessor.ts
```diff
@@ -94,25 +88,108 @@ function configurePragmas(db: Database): void {
...
+	readonly exists?: (path: string) => boolean;
+}): SqliteChoice | null {
+	const platform = opts?.platform ?? process.platform;
+	if (platform !== "darwin") return null;
+
+	const env = opts?.env ?? process.env;
+	const exists = opts?.exists ?? existsSync;
+	const agentsDir = opts?.agentsDir ?? env.SIGNET_PATH ?? join(homedir(), ".agents");
+
+	const envPath = env.SIGNET_SQLITE_PATH;
```
PR-Reviewer-Ant, line 153: If `SIGNET_SQLITE_PATH` is set but does not exist, the new logic only logs a warning and then falls through to `$SIGNET_PATH/libsqlite3.dylib` / Homebrew candidates. That means an operator can explicitly request one dylib and silently get a different one. This diverges from the PR/docs claim that `SIGNET_SQLITE_PATH` is the explicit override, and it makes misconfiguration harder to detect when debugging extension-load failures.

NicholaiVogel: yep, that was a real footgun. fixed in a7b1210a.

`SIGNET_SQLITE_PATH` is now authoritative:
- if it is set and exists, we use it
- if it is set and missing, we stop there, record the attempted path, and refuse fallback to workspace/Homebrew sqlite

that keeps the runtime behavior aligned with the docs and makes misconfiguration obvious instead of quietly selecting a different dylib.

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

packages/daemon/src/db-accessor.ts
```diff
@@ -94,25 +88,114 @@ function configurePragmas(db: Database): void {
...
+
+	const local = join(agentsDir, "libsqlite3.dylib");
+	if (exists(local)) {
+		return { path: local, source: "workspace" };
+	}
+
+	for (const path of HOMEBREW_SQLITE_PATHS) {
+		if (exists(path)) {
+			return { path, source: "homebrew" };
+		}
```
PR-Reviewer-Ant, line 169: `SIGNET_SQLITE_PATH` is treated as authoritative even when the pointed file does not exist, and the code refuses to continue to the workspace/Homebrew fallbacks. That diverges from the PR description's stated discovery order and creates a real failure mode: a stale env var now forces degraded keyword-only mode even when a valid dylib is otherwise available. This should continue trying the later candidates after reporting the bad override, or the PR/docs should explicitly scope the behavior as an intentional hard-fail.

packages/daemon/src/embedding-health.ts
```diff
@@ -209,13 +199,21 @@ function checkModelDrift(db: ReadDb): EmbeddingCheckResult {
...
-	const row = db
-		.prepare(
-			"SELECT COUNT(*) AS n FROM embeddings e LEFT JOIN vec_embeddings v ON v.id = e.id WHERE v.id IS NULL",
-		)
-		.get() as { n: number } | undefined;
-	const count = row?.n ?? 0;
+function checkNullVectors(db: ReadDb, runtime: VectorRuntimeStatus): EmbeddingCheckResult {
+	let count = 0;
+	try {
+		const row = db
```
PR-Reviewer-Ant, line 226: These new broad `catch` blocks convert any SQL error into a generic "vec_embeddings is unavailable" warning. That fixes the missing-table case, but it also hides unrelated failures such as schema corruption, broken joins, or other SQLite errors and misreports them as vec setup problems. This weakens diagnostics and can obscure real data-corruption/debugging signals. Catch only the specific missing-table path or propagate the underlying error detail in the report.

packages/daemon/src/db-accessor.ts
NicholaiVogel: good catch. i kept the authoritative override behavior and made the contract explicit in 281516d9.

what changed:
- updated `docs/CONFIGURATION.md` and `docs/SELF-HOSTING.md` to say `SIGNET_SQLITE_PATH` is authoritative when set and refuses fallback if missing
- updated the planning + research docs to lock that behavior in the spec trail
- updated the PR body so the hard-fail semantics are stated directly instead of implying a pure fallback chain

that keeps runtime, docs, and spec language aligned.

packages/daemon/src/embedding-health.ts
NicholaiVogel: yep, fixed in 281516d9.

changes:
- narrowed the vec-health fallback logic so we only treat the specific `vec_embeddings` missing-table path as the expected degraded-mode warning
- other SQL failures now surface as `fail` results with the underlying SQLite error attached in `detail.error`
- added a regression test with a broken `vec_embeddings` view so we lock the non-missing-table failure path and avoid masking integrity/query problems as vec setup issues

revalidated with:
- `bun test packages/daemon/src/embedding-health.test.ts packages/daemon/src/db-accessor.test.ts`
- `bun run --filter "@signet/daemon" build`

NicholaiVogel:
latest follow-up is in `281516d9`.

addressed:
- made the `SIGNET_SQLITE_PATH` hard-fail semantics explicit in the PR body, runtime docs, and planning/research trail
- tightened embedding health so only the actual missing `vec_embeddings` case degrades to the vec setup warning
- other SQL failures now surface as `fail` with the underlying SQLite error in `detail.error`
- added a regression test for the non-missing-table failure path

revalidated with:
- `bun test packages/daemon/src/embedding-health.test.ts packages/daemon/src/db-accessor.test.ts`
- `bun run --filter "@signet/daemon" build`

packages/daemon/src/db-accessor.ts
```diff
@@ -94,25 +88,114 @@ function configurePragmas(db: Database): void {
...
+	}
+
+	return null;
+}
+
+function explainSqliteSetup(agentsDir: string): string {
```
PR-Reviewer-Ant, line 185: The workspace-local fallback is now derived from `dirname(dirname(path))` when `SIGNET_PATH` is unset. That only works if the DB lives at `<workspace>/memory/memories.db`. If a user has a custom `memory.database` location (for example a deeper subdirectory or an absolute path outside the workspace), Signet will look for `libsqlite3.dylib` in the wrong directory and skip the documented `$SIGNET_PATH/libsqlite3.dylib` fallback entirely. This means the PR does not reliably achieve the claimed workspace-local discovery behavior.

packages/daemon/src/db-accessor.ts
```diff
@@ -168,17 +251,18 @@ export function initDbAccessor(path: string): void {
 
 	dbPath = path;
 
+	configureCustomSqlite(path);
```
PR-Reviewer-Ant: This PR description says custom SQLite selection now happens before the first Bun DB connection, but the implementation only calls `Database.setCustomSQLite()` inside `initDbAccessor()`. Bun rejects `setCustomSQLite()` once SQLite has already been loaded, so this remains order-sensitive: any earlier `new Database(...)` anywhere else in daemon startup will permanently disable the override and put macOS back into degraded mode. A regression test or startup-path proof is needed here, otherwise the implementation still falls short of the stated guarantee.

packages/daemon/src/db-accessor.ts
NicholaiVogel: yep, that fallback path derivation was too clever by half. fixed in a02ba665.

changes:
- added `resolveSqliteAgentsDir()` so workspace-local sqlite discovery always keys off `SIGNET_PATH` or the default home-scoped `~/.agents` root
- removed the `dirname(dirname(path))` inference, so a custom `memory.database` location no longer redirects the documented `$SIGNET_PATH/libsqlite3.dylib` fallback
- added a regression test covering the unset-`SIGNET_PATH` home-scoped default

packages/daemon/src/db-accessor.ts
NicholaiVogel: fair call. i added a concrete guardrail in a02ba665 instead of just hand-waving about startup order.

what is locked now:
- production Bun sqlite construction is centralized in `db-accessor.ts`
- `main()` still initializes `initDbAccessor(MEMORY_DB)` before any accessor usage
- added a regression test, `keeps bun sqlite construction centralized in db-accessor`, which scans non-test daemon source and fails if any other production file starts constructing Bun databases directly

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

NicholaiVogel:
latest follow-up is in `a02ba665`.

addressed:
- fixed workspace-local sqlite discovery so it always uses `SIGNET_PATH` or the default `~/.agents` root, not the database file's parent chain
- added a regression test for the home-scoped default when `SIGNET_PATH` is unset
- added a startup-order guardrail test that fails if any non-test daemon source constructs Bun SQLite outside `db-accessor.ts`

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

packages/daemon/src/db-accessor.ts
```diff
@@ -94,25 +88,124 @@ function configurePragmas(db: Database): void {
 // Cached extension path — resolved once at startup
 let vecExtPath: string | null | undefined;
 
+export function resolveSqliteAgentsDir(opts?: {
+	readonly env?: NodeJS.ProcessEnv;
+	readonly home?: () => string;
+}): string {
```
PR-Reviewer-Ant, line 129: The new workspace-local sqlite fallback is derived from `SIGNET_PATH` or `~/.agents` only. That misses the supported persisted workspace path (`~/.config/signet/workspace.json`) and also ignores the actual database path already passed into `initDbAccessor(path)`. Result: users who moved their workspace with `signet workspace set ...` but do not export `SIGNET_PATH` will search `~/.agents/libsqlite3.dylib` instead of the active workspace, so the PR does not actually provide the advertised workspace-local recovery path for those installs.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `a02ba665`

[Automated Review] Found one blocking correctness issue in the macOS SQLite discovery flow.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new workspace-local lookup is implemented in db-accessor.ts by calling resolveSqliteAgentsDir(), which only consults SIGNET_PATH and otherwise hardcodes ~/.agents. The repo docs explicitly support a persisted workspace selection via ~/.config/signet/workspace.json, and initDbAccessor(path) already knows the real DB path. In that supported configuration, the PR's advertised '$SIGNET_PATH/libsqlite3.dylib' fallback will look in the wrong directory unless the user also exports SIGNET_PATH, so the claimed fix does not hold for non-default workspaces.

packages/daemon/src/db-accessor.ts
NicholaiVogel: yep, good catch. fixed in f75969fa.

changes:
- `resolveSqliteAgentsDir()` now follows the real workspace resolution chain: `SIGNET_PATH` → `~/.config/signet/workspace.json` → default `~/.agents`
- that makes the workspace-local sqlite fallback honor `signet workspace set ...` installs even when `SIGNET_PATH` is not exported
- updated the runtime/spec docs to refer to `$SIGNET_WORKSPACE/libsqlite3.dylib` and the actual resolution order instead of implying env-only behavior
- added a regression test covering the persisted workspace-config path

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

NicholaiVogel:
latest follow-up is in `f75969fa`.

addressed:
- sqlite workspace fallback now uses the real active workspace resolution chain: `SIGNET_PATH` → persisted workspace config → default `~/.agents`
- that means installs configured with `signet workspace set ...` now correctly look for `$SIGNET_WORKSPACE/libsqlite3.dylib`
- runtime/spec docs were updated to match the actual resolution order
- added a regression test for the persisted workspace-config case

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

packages/daemon/src/db-accessor.ts
```diff
@@ -94,25 +88,157 @@ function configurePragmas(db: Database): void {
...
+function explainSqliteSetup(agentsDir: string): string {
+	return [
+		"macOS system SQLite may block loadExtension() and force keyword-only recall.",
+		`Set SIGNET_SQLITE_PATH, place libsqlite3.dylib in ${agentsDir}, or install Homebrew sqlite.`,
+	].join(" ");
+}
+
+function configureCustomSqlite(): void {
+	sqliteChoice = null;
+	sqliteAttempt = null;
```
PR-Reviewer-Ant: If the workspace-local `libsqlite3.dylib` exists but is unusable (wrong arch, bad build, stale file), this code stops after `Database.setCustomSQLite()` fails and never tries the lower-priority Homebrew candidates. That effectively makes the workspace copy authoritative too, even though the PR description only makes `SIGNET_SQLITE_PATH` fail-closed. Result: a broken workspace dylib can still force degraded mode even when a valid Homebrew SQLite is installed.

packages/daemon/src/db-accessor.ts
```diff
@@ -168,17 +294,18 @@ export function initDbAccessor(path: string): void {
 
 	dbPath = path;
 
+	configureCustomSqlite();
```
PR-Reviewer-Ant: The new discovery path is not derived from the actual database path passed into `initDbAccessor(path)`. `configureCustomSqlite()` resolves the workspace again from env/config, so if the caller opens a DB in a non-default location via an explicit path, the code may probe the wrong `$SIGNET_WORKSPACE/libsqlite3.dylib`. That diverges from the PR claim that workspace-local fallback tracks the active workspace used for the daemon's first Bun connection.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f75969fa`

[Automated Review] Found two correctness issues in the new macOS SQLite discovery flow that can still leave vector search degraded even when a valid fallback exists.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly provable from the changed control flow in `packages/daemon/src/db-accessor.ts` and can be evaluated against the PR's stated goals. The first is a concrete fallback bug in `configureCustomSqlite()`. The second is a mismatch between the actual DB path passed to `initDbAccessor()` and the workspace path used for SQLite discovery.

packages/daemon/src/db-accessor.ts
NicholaiVogel: yep, fair catch. fixed in 3ba192c3.

changes:
- split sqlite runtime selection into a testable `resolveSqliteRuntimeConfig()` helper
- `SIGNET_SQLITE_PATH` stays fail-closed because it is the explicit override
- workspace-local sqlite no longer becomes implicitly authoritative on activation failure, we now try lower-priority Homebrew candidates before declaring degraded mode
- added a regression test that simulates a broken workspace dylib and verifies fallback to Homebrew

packages/daemon/src/db-accessor.ts
NicholaiVogel: also fixed in 3ba192c3.

production startup now passes the active daemon workspace directly into `initDbAccessor(MEMORY_DB, { agentsDir: AGENTS_DIR })`, so sqlite discovery keys off the same workspace the daemon is actually booting from instead of re-resolving it indirectly.

i also added a regression test that verifies the explicit `agentsDir` wins over env-derived workspace resolution during init-time sqlite selection.

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

NicholaiVogel:
latest follow-up is in `3ba192c3`.

addressed:
- workspace-local sqlite is no longer fail-closed on activation failure, it now falls through to valid Homebrew candidates
- explicit `SIGNET_SQLITE_PATH` remains the only authoritative fail-closed override
- daemon startup now passes the active workspace directly into `initDbAccessor(...)`, so sqlite discovery uses the same workspace the daemon is actually booting from
- added regression tests for broken workspace dylib fallback and explicit `agentsDir` precedence

revalidated with:
- `bun test packages/daemon/src/db-accessor.test.ts packages/daemon/src/embedding-health.test.ts`
- `bun run --filter "@signet/daemon" build`

Issue #218, opened 2026-03-17
MEMORY.md regeneration should respect decay scores and prune stale content


stephenwoska2-cpu:
## Problem

MEMORY.md is auto-generated from the memory database as a working memory summary injected into AI sessions. Currently, the generation uses a fixed section layout (System Configuration, Active Projects, Open Threads, etc.) and doesn't appear to prune content based on the same decay scoring (`importance × 0.95^days`) used by the memory search/injection pipeline.

This creates two issues:

1. **Stale content persists** — resolved threads, old project details, and dated references accumulate in fixed sections rather than being dropped as they decay. Example: a "5pm reminder on 3/5" was still present on 3/17, two weeks after it was relevant.

2. **Truncation is position-based, not score-based** — when the file exceeds the 10K character budget, it truncates from the bottom rather than dropping the lowest-scoring items. Whatever section happens to be last gets cut, regardless of importance.

The net effect is that the token budget fills up with stale context while potentially more relevant recent memories get truncated.

## Expected Behavior

MEMORY.md regeneration should:
- Score items using the same decay formula as memory search (`importance × decay_rate^days`, pinned items exempt)
- Drop items below a configurable threshold (e.g., the existing `min_score: 0.3`)
- When approaching the character budget, remove lowest-scoring items first rather than truncating by position
- Keep the human-readable section format (users read this file too), but let sections shrink or disappear when their items decay out

## Environment

- Signet v0.58.1 → v0.59.0
- macOS, Bun runtime, Ollama embeddings
- `agent.yaml`: decay_rate 0.95, session_budget 2000, search min_score 0.3

NicholaiVogel:
Hey, appreciate the detailed writeup. You're spot on about all three issues: no decay scoring in the synthesis input, position-based truncation, and the LLM having to guess what's stale.

We've been thinking about this for a while now. The fix isn't trivial because the synthesis pipeline needs to integrate with the memory DB's scoring/decay system rather than just reading raw markdown files. The current approach was always a stopgap.

What we're working toward:
- Query decay-scored memories directly from the DB as synthesis input
- Score-based budget allocation (lowest scores get dropped first, not bottom-of-file)
- Pinned items exempt from decay, same as search
- Sections that shrink or disappear naturally as their items decay out

No ETA yet since we want to get this right, but it's actively on the roadmap. Will update this issue as we make progress.

PR #344, opened 2026-03-26
fix(memory): close lossless working-memory runtime gaps

NicholaiVogel:
## Summary

This PR codifies and ships the lossless working-memory model across the
active runtime surfaces touched in this wave.

The core methodology is now explicit in the specs and reflected in the
runtime:

- one agent, many sessions/branches, one shared `MEMORY.md` head
- `MEMORY.md` is both:
  - a dense reinjection buffer for immediate continuity
  - a machine-facing index into deeper temporal state
- LCM/DAG patterns drive state over time rather than flat file merges
- live transcripts are written in real time with `agent_id + session_key`
  scoping
- prompt-time retrieval remains structured-memory-first
- transcripts are searchable/queryable fallback and deep-history substrate,
  not co-equal primary retrieval
- session-end summaries and compaction artifacts feed the same temporal DAG
- same-agent concurrent writers update the same head with merge-safe
  protection
- harnesses target the same runtime model with explicitly documented degraded
  modes when hook fidelity is limited

In practice, the intended loop is now:

1. session starts and injects `MEMORY.md` plus workspace identity files
2. transcript is written live while the session is active
3. structured distillation remains the primary retrieval surface
4. transcripts are available as fallback until structural state catches up,
   or for transcript-specific lookup / drill-down
5. session-end writes a summary into the same temporal body and refreshes
   `MEMORY.md`
6. compaction writes a first-class compaction artifact into that same body,
   instead of forking continuity into a separate path

## Changes

- added approved specs for:
  - `memory-md-temporal-head`
  - `lossless-working-memory-runtime`
  - `lossless-working-memory-closure`
- updated `docs/specs/INDEX.md` and `docs/specs/dependencies.yaml` so the
  runtime contract is hard-required rather than implied
- replaced flat `MEMORY.md` synthesis assumptions with a DB-backed temporal
  head model
- added `memory_md_heads` for lease-aware, retry-safe same-agent head writes
- added `session-transcripts.ts` for live transcript persistence and fallback
  search
- added `temporal-expand.ts`, `POST /api/sessions/summaries/expand`, and MCP
  tool `lcm_expand` for DAG drill-down
- extended temporal provenance in `session_summaries` to support:
  - `summary`
  - `chunk`
  - `compaction`
  - `condensation`
- made compaction artifacts first-class temporal nodes and included them in
  condensation
- hardened OpenClaw plugin compaction parity, including `sessionFile`
  recovery of compaction summaries when the hook payload only exposes
  metadata
- cleared prompt dedupe after compaction so context can be reconsidered on
  the next turn
- hardened temporal scoping so transcript persistence and session-summary
  uniqueness are keyed by `agent_id + session_key`
- added deterministic duplicate reconciliation in migrations for legacy
  transcript and summary rows before new scoped uniqueness constraints land
- made `compaction-complete` preserve project lineage even when the hook lands
  before transcript persistence by accepting a fallback request project,
  carrying it through CLI/OpenCode/OpenClaw callers, and mirroring that hook
  contract in `packages/daemon-rs/`
- updated docs to explain:
  - the methodology above
  - transcript fallback discipline
  - merge-safe `MEMORY.md` behavior
  - harness fidelity matrix and degraded modes

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `@signet/opencode-plugin`, `@signetai/signet-memory-openclaw`,
      `packages/daemon-rs`

## Screenshots

N/A, no UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Rollback / compatibility note:

- this PR adds migrations `044` through `047`
- the daemon already creates pre-migration SQLite backups, so rollback should
  restore from backup rather than attempt partial schema downgrade
- transcript and session-summary uniqueness are now agent-scoped, which is the
  compatibility-safe direction for shared DB multi-agent use
- touched hook-surface parity for `compaction-complete` is mirrored into
  `packages/daemon-rs/`; broader daemon-rs feature parity remains a separate
  track outside this PR

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

What I ran successfully:

- `bun test packages/core/src/migrations/migrations.test.ts`
- `bun test packages/daemon/src/mcp/tools.test.ts`
- `bun test packages/daemon/src/temporal-expand.test.ts`
- `bun test packages/daemon/src/temporal-summary-api.test.ts`
- `bun test packages/daemon/src/memory-search.test.ts`
- `bun test packages/daemon/src/pipeline/summary-condensation.test.ts`
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts`
- `bun test packages/daemon/src/pipeline/synthesis-worker.test.ts`
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `bun test packages/cli/src/commands/hook.test.ts`
- `bun test packages/daemon/test/hooks.test.ts`
- `bun test packages/daemon/src/pipeline/significance-gate.test.ts`
- `bun test packages/daemon/src/predictor-comparison.test.ts`
- `bun test packages/daemon/src/predictor-training-pairs.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `cd packages/opencode-plugin && bun run build`
- `cd packages/opencode-plugin && bun run typecheck`
- `cd packages/connector-opencode && bun run build:bundle`
- `cd packages/daemon-rs && cargo test -p signet-daemon compaction_project`
- `cd packages/daemon-rs && cargo build -p signet-daemon`
- `bun run typecheck`

Notes on unchecked boxes:

- `bun run lint` currently fails repo-wide due existing formatting/lint debt
  outside this PR
- I did not run a single full `bun test` umbrella command; validation was
  targeted to the touched packages and runtime surfaces
- I did not do a separate manual live-daemon dogfood pass in this PR

## AI disclosure

## Notes

Intentional deferrals:

- no promotion of transcripts to co-equal primary retrieval
- no dedicated vectorized transcript-chunk retrieval tier yet; transcript
  behavior in this wave is searchable fallback + expansion substrate

Relevant methodology sentence for reviewers:

> `MEMORY.md` is the merge-safe, continuously refreshed head of a lossless,
> agent-scoped temporal memory DAG, fed by live transcripts, session
> summaries, and compaction artifacts, while structured distillation remains
> the primary retrieval surface.

Refs #218

### DIFF
--- a/packages/adapters/openclaw/src/index.ts
+++ b/packages/adapters/openclaw/src/index.ts
@@ -815,6 +815,30 @@ function resolveCompactionSessionFile(
 	);
 }
 
+function readSessionFileProject(sessionFile: string | undefined): string | undefined {
+	if (!sessionFile || !existsSync(sessionFile)) return undefined;
+
+	try {
+		const lines = readFileSync(sessionFile, "utf-8")
+			.split("\n")
+			.map((line) => line.trim())
+			.filter((line) => line.length > 0);
+		for (const line of lines) {
+			try {
+				const row = JSON.parse(line) as unknown;
+				if (!isRecord(row) || row.type !== "session") continue;
+				return firstNonEmptyString(row.cwd, row.project, row.workspace);
+			} catch {
+				// ignore malformed transcript lines
+			}
+		}
+	} catch {
+		// best effort only
+	}
+
+	return undefined;
+}
+
 function extractCompactionSummary(event: Record<string, unknown>, sessionFile: string | undefined): string | undefined {
 	const direct = readString(event.summary);
 	if (direct) return direct;
@@ -1448,10 +1472,12 @@ const signetPlugin = {
 		const resolveCompactionProject = (
 			event: Record<string, unknown>,
 			ctx: {
+				sessionFile?: string;
 				project?: string;
 			},
 		): string | undefined => {
 			const compaction = isRecord(event.compaction) ? event.compaction : undefined;
+			const sessionFile = resolveCompactionSessionFile(event, ctx.sessionFile);
 			return firstNonEmptyString(
 				event.project,
 				event.cwd,
@@ -1460,6 +1486,7 @@ const signetPlugin = {
 				compaction?.cwd,
 				compaction?.workspace,
 				ctx.project,
+				readSessionFileProject(sessionFile),
 			);
 		};
 
@@ -1518,7 +1545,11 @@ const signetPlugin = {
 						? event.compactingCount
 						: typeof event.compactedCount === "number"
 							? event.compactedCount
-							: undefined;
+							: isRecord(event.compaction) && typeof event.compaction.compactingCount === "number"
+								? event.compaction.compactingCount
+								: isRecord(event.compaction) && typeof event.compaction.compactedCount === "number"
+									? event.compaction.compactedCount
+									: undefined;
 			const dedupeKey = buildCompactionEventKey(event, {
 				agentId: ctx.agentId,
 				sessionKey,
--- a/packages/adapters/openclaw/src/index.test.ts
+++ b/packages/adapters/openclaw/src/index.test.ts
@@ -484,6 +484,26 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		});
 	});
 
+	it("uses nested compaction counts as a fallback pre-compaction message count", async () => {
+		const { api, hooks } = createMockApi();
+		signetPlugin.register(api);
+
+		const beforeCompaction = hooks.get("before_compaction");
+		expect(beforeCompaction).toBeDefined();
+
+		await beforeCompaction?.(
+			{ compaction: { compactingCount: 9 } },
+			{ sessionKey: "session-compact-nested", agentId: "agent-1" },
+		);
+
+		expect(lastPreCompactionBody).toMatchObject({
+			harness: "openclaw",
+			sessionKey: "session-compact-nested",
+			messageCount: 9,
+			runtimePath: "plugin",
+		});
+	});
+
 	it("combines summaryPrompt and guidelines for pre-compaction context", async () => {
 		const { api, hooks } = createMockApi();
 		signetPlugin.register(api);
@@ -622,6 +642,40 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		expect(lastCompactionBody).toMatchObject({
 			project: "/tmp/branch-lineage",
 			sessionKey: "session-lineage",
+		});
+	});
+
+	it("recovers project lineage from the session file header when the event lacks cwd/project hints", async () => {
+		const { api, hooks } = createMockApi();
+		signetPlugin.register(api);
+
+		const afterCompaction = hooks.get("after_compaction");
+		expect(afterCompaction).toBeDefined();
+
+		const sessionFile = join(testDir, "session-lineage-header.jsonl");
+		writeFileSync(
+			sessionFile,
+			[
+				JSON.stringify({
+					type: "session",
+					version: 1,
+					id: "session-lineage-header",
+					cwd: "/tmp/header-lineage",
+				}),
+				JSON.stringify({
+					type: "compaction",
+					id: "comp-lineage-header",
+					summary: "Recovered project from session header.",
+				}),
+			].join("\n"),
+			"utf-8",
+		);
+
+		await afterCompaction?.({ sessionFile }, { sessionKey: "session-lineage-header", agentId: "agent-1" });
+
+		expect(lastCompactionBody).toMatchObject({
+			project: "/tmp/header-lineage",
+			sessionKey: "session-lineage-header",
 		});
 	});
 


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(memory): close lossless working-memory runtime gaps` (commit `23a2775d`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `23a2775d`

[Automated Review] Found two blocking multi-agent retrieval regressions in the daemon hook path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The regressions are directly visible in the diff. `handleUserPromptSubmit` now hardcodes `readPolicy: "isolated"`, and `getAllScoredCandidates` now filters by `agent_id` before `handleSessionStart` uses it. Both changes bypass the repository's documented shared/group visibility model for agents. I did not run the daemon, but these behavior changes are explicit in the changed queries and call sites.


- packages/daemon/src/hooks.ts:3473 - This hardcodes prompt-submit recall to `readPolicy: "isolated"`, which suppresses globally visible memories for agents configured with `shared` or `group` read policies. The PR description says retrieval remains structured-memory-first, not that cross-agent visibility is removed. As written, non-isolated agents will stop seeing memories they are explicitly allowed to read.
- packages/daemon/src/hooks.ts:3251 - `handleSessionStart()` now builds its candidate pool through `getAllScoredCandidates(..., traversalAgentId)`, and that helper was changed to `WHERE is_deleted = 0 AND agent_id = ?`. That means session-start injection is no longer visibility-aware and only reads the current agent's own rows. This breaks the documented multi-agent contract for `shared`/`group` policies and will make valid global memories disappear from initial context.

NicholaiVogel:
addressed the visibility regression callouts.

what changed in `d758dc7a`:
- restored visibility-aware session-start candidate selection instead of forcing own-agent-only rows
- restored prompt-submit recall to use the requesting agent's configured `read_policy` / `policy_group` instead of hardcoding `isolated`
- extracted a shared `getAgentScope()` helper so the daemon and hook paths resolve scope the same way
- added regression tests for:
  - shared visibility at session start
  - group visibility during prompt recall

validation:
- `bun test packages/daemon/test/hooks.test.ts`
- `bun run typecheck`
- `bun x biome check packages/daemon/src/agent-id.ts packages/daemon/src/hooks.ts packages/daemon/test/hooks.test.ts`

packages/core/src/migrations/046-session-summary-uniqueness.ts
```diff
@@ -0,0 +1,12 @@
+import type { MigrationDb } from "./index";
+
+export function up(db: MigrationDb): void {
+	db.exec(`
+		DROP INDEX IF EXISTS idx_summaries_session_depth;
+
+		CREATE UNIQUE INDEX IF NOT EXISTS idx_summaries_session_depth_summary
+			ON session_summaries(session_key, depth)
```
PR-Reviewer-Ant: This migration introduces a temporary unique index on `(session_key, depth)` before migration 047 replaces it with the intended agent-scoped version. On any existing multi-agent database that already contains the same `session_key`/`depth` under different `agent_id` values, migration 046 will fail and block the upgrade before 047 can fix the key shape. That contradicts the PR's compatibility claim for shared multi-agent use.

packages/daemon/src/daemon.ts
```diff
@@ -6574,6 +6619,10 @@ app.post("/api/sessions/:key/renew", (c) => {
 // Session summaries DAG
 app.get("/api/sessions/summaries", (c) => {
```
PR-Reviewer-Ant, line 6638: The new temporal summary read API trusts `agentId` from query/header via `resolveAgentId(...)` and then queries that agent's rows directly. I don't see any auth-scope check in this added path analogous to project-scope enforcement elsewhere, so in `team`/`hybrid` mode a scoped caller can request another agent's temporal summaries by supplying a different `agentId` or crafted `sessionKey`. The same concern applies to the new `/api/sessions/summaries/expand` route added below.

packages/daemon/src/daemon.ts
```diff
@@ -6014,24 +5994,71 @@ app.post("/api/hooks/compaction-complete", async (c) => {
...
+				VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+			).run(
 				summaryId,
 				body.summary,
 				"session_summary",
 				0.8,
+				body.sessionKey ?? null,
 				body.harness,
 				"system",
 				JSON.stringify(["session", "summary", body.harness]),
```
PR-Reviewer-Ant: This writes `tags` as `JSON.stringify(["session", "summary", body.harness])`, but the rest of the API and surrounding code treat `tags` as a comma-delimited string. These compaction-summary memories will therefore diverge from normal tag storage and may stop matching tag filters/searches consistently.

NicholaiVogel:
knocked out the next round in `956fd92d`.

addressed:
- migration 046 no longer introduces a pre-047 multi-agent uniqueness trap, it now keys summary uniqueness by `agent_id + session_key + depth`
- added a migration regression test covering same `session_key` / `depth` across different agents before 047
- temporal summary reads now use scoped agent/project resolution helpers instead of trusting caller-supplied agent ids
- added request-scope helper tests for team/hybrid/local behavior
- compaction-complete summary memories now store tags as comma-delimited text again
- added API tests for:
  - `/api/sessions/summaries` not falling through the `:key` session route
  - `/api/sessions/summaries/expand`
  - compaction summary tag shape
- cleaned the stale unused fs imports in `hooks.ts`

validation:
- `bun test packages/core/src/migrations/migrations.test.ts`
- `bun test packages/daemon/src/request-scope.test.ts packages/daemon/src/temporal-summary-api.test.ts`
- `bun test packages/daemon/test/hooks.test.ts`
- `bun run typecheck`
- `bun x biome check packages/daemon/src/request-scope.ts packages/daemon/src/request-scope.test.ts packages/daemon/src/temporal-summary-api.test.ts packages/daemon/src/hooks.ts packages/core/src/migrations/046-session-summary-uniqueness.ts`

packages/core/src/migrations/046-session-summary-uniqueness.ts
```diff
@@ -0,0 +1,12 @@
+import type { MigrationDb } from "./index";
+
+export function up(db: MigrationDb): void {
+	db.exec(`
+		DROP INDEX IF EXISTS idx_summaries_session_depth;
+
+		CREATE UNIQUE INDEX IF NOT EXISTS idx_summaries_session_depth_summary
```
PR-Reviewer-Ant, line 58: This migration creates the new unique constraint immediately, but it does not reconcile pre-existing duplicate `summary` rows for the same `(agent_id, session_key, depth)`. Since this PR explicitly hardens a retry/uniqueness gap, existing databases can already contain exactly those duplicates; in that case the migration will fail and the daemon will not start. The migration needs a deterministic dedupe/merge step before creating the unique index.

NicholaiVogel:
next round is in `af45788c`.

addressed:
- migration 046 now does a deterministic duplicate-summary reconciliation before creating the unique index:
  - keeps the newest `(agent_id, session_key, depth)` summary row
  - rehomes `session_summary_memories`
  - rewires `session_summary_children`
  - deletes the duplicate rows before the unique index is created
- added a regression test for same-agent retry duplicates during migration 046
- `expandTemporalNode()` now accepts an optional project scope and applies it to:
  - the root node lookup
  - parent / child lineage rows
  - linked memories
  - transcript payload
- `/api/sessions/summaries/expand` now passes the scoped project into the expansion function instead of only checking the root node afterward
- predictor-backed session-start context now uses the same visibility scope (`buildAgentScopeClause`) as the main candidate pool, so shared/group-readable memories can still surface from the predictive path
- added regressions for:
  - nested temporal expansion project filtering
  - predictive session-start context honoring shared visibility

validation:
- `bun test packages/core/src/migrations/migrations.test.ts`
- `bun test packages/daemon/src/temporal-expand.test.ts`
- `bun test packages/daemon/src/request-scope.test.ts`
- `bun test packages/daemon/src/temporal-summary-api.test.ts`
- `bun test packages/daemon/test/hooks.test.ts`
- `bun run typecheck`
- `bun x biome check packages/core/src/migrations/046-session-summary-uniqueness.ts packages/daemon/src/request-scope.ts packages/daemon/src/request-scope.test.ts packages/daemon/src/temporal-expand.ts packages/daemon/src/temporal-expand.test.ts packages/daemon/src/temporal-summary-api.test.ts packages/daemon/src/hooks.ts packages/daemon/test/hooks.test.ts`

NicholaiVogel:
small cleanup in `264585ed`: removed the stale `later` fixture var the code-quality bot flagged in `migrations.test.ts`.

revalidated:
- `bun test packages/core/src/migrations/migrations.test.ts`

packages/daemon/src/daemon.ts
```diff
@@ -8682,6 +8795,12 @@ app.post("/api/knowledge/expand/session", async (c) => {
 	const body = await c.req.json().catch(() => ({}));
```
PR-Reviewer-Ant, line 8802: This endpoint now accepts `agentId` from the request body/header via `resolveAgentId(...)`, but unlike the new summaries endpoints it never calls `resolveScopedAgentId(...)`. In `team`/`hybrid` mode that lets a caller with a scoped token ask for another agent's entity/session history just by supplying a different `agentId` (or header), which is a cross-agent data leak. This read path should enforce auth scope the same way the other new agent-scoped routes do.

packages/core/src/migrations/047-agent-scoped-temporal-uniqueness.ts
```diff
@@ -0,0 +1,91 @@
...
+			PRIMARY KEY (agent_id, session_key)
+		);
+
+		INSERT OR REPLACE INTO session_transcripts_next (
+			session_key,
+			content,
+			harness,
+			project,
+			agent_id,
+			created_at,
```
PR-Reviewer-Ant, line 33: This migration can silently lose transcript data on upgrade. `session_transcripts_next` is keyed by `(agent_id, session_key)`, but the backfill uses `INSERT OR REPLACE ... SELECT ... FROM session_transcripts` with no `ORDER BY` or merge logic. If an existing DB already contains multiple legacy rows that collapse to the same new key, whichever row happens to be visited last wins and the rest are discarded. For a PR that claims lossless transcript continuity, the migration needs an explicit deterministic merge policy (for example keep newest `updated_at`, or concatenate content) plus a regression test covering duplicate pre-migration rows.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 8802: This endpoint now accepts caller-controlled `agentId`/`x-signet-agent-id` and uses it directly in the query path, but unlike the new session-summary endpoints it never calls `resolveScopedAgentId(...)` or otherwise checks the auth token's agent scope. In `team`/`hybrid` mode, a scoped token can request another agent's knowledge-expansion session history by supplying a different agent id. This reintroduces exactly the kind of agent scoping leak called out in the repo guardrails.

NicholaiVogel:
next round is in `7105f45d`.

addressed:
- migration 047 now reconciles legacy `session_transcripts` rows deterministically before the new `(agent_id, session_key)` primary key lands:
  - ranks duplicate rows per agent/session
  - keeps the newest row first
  - breaks ties by transcript richness / recency instead of relying on sqlite replacement order
- `/api/knowledge/expand/session` now uses the same scoped-agent enforcement helper as the other temporal read routes instead of trusting caller-supplied `agentId`
- the session-expansion query now also applies scoped project filtering when a token constrains project access
- `MEMORY.md` synthesis now resolves the caller's `read_policy` / `policy_group` through `getAgentScope()` before building its scored candidate pool, so shared/group-readable memories stay aligned with the runtime retrieval surface
- added a regression test proving synthesis for a shared agent includes globally visible memories
- added a regression test covering deterministic transcript collapse in migration 047

validation:
- `bun test packages/core/src/migrations/migrations.test.ts`
- `bun test packages/daemon/test/hooks.test.ts`
- `bun test packages/daemon/src/request-scope.test.ts`
- `bun test packages/daemon/src/temporal-summary-api.test.ts`
- `bun test packages/daemon/src/temporal-expand.test.ts`
- `bun run typecheck`
- `bun x biome check packages/core/src/migrations/047-agent-scoped-temporal-uniqueness.ts packages/core/src/migrations/migrations.test.ts packages/daemon/src/hooks.ts packages/daemon/test/hooks.test.ts`

small note:
- `packages/daemon/src/daemon.ts` still has pre-existing biome debt outside this PR surface, so I revalidated that file via build + typecheck + the targeted route tests instead of claiming a clean full-file lint pass.

packages/daemon/src/daemon.ts
```diff
@@ -6014,24 +5986,78 @@ app.post("/api/hooks/compaction-complete", async (c) => {
...
-        VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
-      `).run(
+			const transcriptRow = body.sessionKey
+				? (db
+						.prepare(
+							`SELECT project
+							 FROM session_transcripts
+							 WHERE session_key = ? AND agent_id = ?`,
+						)
+						.get(body.sessionKey, agentId) as { project: string | null } | undefined)
```
PR-Reviewer-Ant: Blocking: the new compaction-complete mutation path derives `project` from `session_transcripts` using only `(session_key, agent_id)` and then writes the memory/temporal node without enforcing the caller's scoped project claim. A token scoped to project A can target a known `sessionKey` from project B for the same agent and create cross-project summary artifacts there. This route needs the same explicit project-scope enforcement as other mutation endpoints before persisting.

packages/cli/src/commands/hook.ts
```diff
@@ -166,11 +168,16 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
 		.requiredOption("-H, --harness <harness>", "Harness name")
 		.requiredOption("-s, --summary <summary>", "Session summary text")
 		.action(async (options) => {
+			const input = await readJson();
 			const data = await deps.fetchFromDaemon<{ success?: boolean; memoryId?: number; error?: string }>(
 				"/api/hooks/compaction-complete",
 				{
 					method: "POST",
```
PR-Reviewer-Ant, line 160: The CLI compaction path does not actually forward `sessionKey`; it only reads `session_id` / `sessionId` from stdin. That means callers providing `sessionKey` (which the updated docs/API now describe and the daemon uses for lineage, agent scoping, and temporal expansion) silently lose that linkage on the CLI route. This diverges from the PR's stated compaction-lineage behavior.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `7105f45d`

pr-reviewer automated bot review: I found one blocking scope-enforcement issue and one correctness gap where the CLI path does not preserve the new compaction lineage contract.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The project-scope issue is directly visible in the new `/api/hooks/compaction-complete` logic: it reads `project` from `session_transcripts` using only `session_key` + `agent_id`, then writes a new memory/temporal node without any project-scope check on the caller. The CLI issue is also explicit in the diff: `signet hook compaction-complete` forwards only `session_id`/`sessionId`, not `sessionKey`, even though the daemon/docs now depend on `sessionKey` to preserve lineage.

NicholaiVogel:
follow-up round is in `08d5d8ea`.

addressed:
- `/api/hooks/compaction-complete` now resolves the session transcript's project before writing and runs it through the scoped-project helper, so project-scoped tokens cannot persist compaction artifacts onto an out-of-scope branch
- when a scoped project is present but the transcript row is missing, the compaction write now still lands on the caller's scoped project instead of silently dropping project context
- CLI hook parsing now uses a shared `pickSessionKey()` helper that prefers canonical `session_key` / `sessionKey` fields before falling back to legacy aliases
- `session-start`, `user-prompt-submit`, `session-end`, `pre-compaction`, and `compaction-complete` all use the same session-key normalization path now, so the newer lineage contract does not depend on only one hook path getting updated
- added regressions for:
  - project scope mismatches in `resolveScopedProject()`
  - CLI session-key normalization preferring canonical fields and falling back to legacy aliases

validation:
- `bun test packages/daemon/src/request-scope.test.ts`
- `bun test packages/daemon/src/temporal-summary-api.test.ts`
- `bun test packages/cli/src/commands/hook.test.ts`
- `bun run typecheck`
- `bun x biome check packages/daemon/src/request-scope.test.ts packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts`

packages/daemon/src/daemon.ts
```diff
@@ -6014,24 +5986,83 @@ app.post("/api/hooks/compaction-complete", async (c) => {
 		}
```
PR-Reviewer-Ant, line 5987: The new `/api/hooks/compaction-complete` path recovers `project` only by looking up an existing `session_transcripts` row for `(agent_id, session_key)`, but neither this endpoint nor the updated CLI/OpenClaw callers send `project`. For compaction events that happen before transcript persistence, `project` becomes `null`, so the saved `memories` row and `session_summaries` node lose project scoping. That is a real scoping leak/regression for project-scoped recall/expand flows, and it also diverges from the PR claim that temporal writes are hardened rather than weakened.

docs/specs/approved/lossless-working-memory-closure.md
```diff
@@ -0,0 +1,186 @@
...
+informed_by:
+  - "docs/research/technical/RESEARCH-LCM-ACP.md"
+  - "docs/specs/planning/LCM-PATTERNS.md"
+  - "docs/specs/approved/memory-md-temporal-head.md"
+  - "docs/specs/approved/lossless-working-memory-runtime.md"
+success_criteria:
+  - "Temporal node drill-down is available by id and returns parent/child lineage, linked memories, and transcript context without flooding the main MEMORY.md body"
+  - "Concurrent same-agent MEMORY.md writes are lease-safe and retry-safe, and busy head writes do not suppress future regeneration"
+  - "Compaction artifacts participate in the same temporal DAG and condensation path as session summaries, with harness fidelity and degraded modes explicitly documented"
+  - "Prompt-time retrieval keeps structured distillation primary while transcript retrieval stays a searchable, embeddable fallback and deep-history substrate"
```
PR-Reviewer-Ant, line 13: This PR explicitly defers Rust parity, but the repo contract in `AGENTS.md` says behavioral changes under `packages/daemon/` must also be reflected in `packages/daemon-rs/`. This change adds new temporal endpoints, new scoping behavior, new compaction persistence semantics, and new MEMORY.md head write rules on the TypeScript daemon only. In shadow mode that guarantees divergence rather than preserving the required parity contract.

NicholaiVogel:
next round is in `3813977d`.

addressed:
- `/api/hooks/compaction-complete` now accepts an explicit fallback `project` and uses it when compaction lands before transcript persistence, while still preferring transcript-derived lineage when that row exists
- CLI `compaction-complete`, the OpenCode plugin, the bundled connector plugin, and the OpenClaw adapter now all forward fallback project scope so the lineage contract survives across harnesses
- `packages/daemon-rs/` now mirrors the touched compaction hook contract in this PR:
  - request body accepts `project` + `agent_id`
  - compaction project resolution prefers transcript lineage and falls back to request project
  - agent scoping is threaded into the Rust ingest path instead of hardcoding `default`
  - added Rust tests for transcript-preferred and fallback-project compaction writes
- updated API / HOOKS / CLI docs so the fallback project contract is explicit
- updated the closure spec and PR description so touched daemon hook parity in `daemon-rs` is now a hard requirement for this wave instead of an explicit deferral

validation:
- `bun test packages/daemon/src/temporal-summary-api.test.ts`
- `bun test packages/cli/src/commands/hook.test.ts`
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `bun test packages/daemon/src/request-scope.test.ts`
- `cd packages/opencode-plugin && bun run build && bun run typecheck`
- `cd packages/connector-opencode && bun run build:bundle`
- `cd packages/adapters/openclaw && bun run build`
- `cd packages/daemon-rs && cargo test -p signet-daemon compaction_project`
- `cd packages/daemon-rs && cargo build -p signet-daemon`
- `bun run typecheck`

packages/cli/src/commands/hook.ts
```diff
@@ -104,14 +106,15 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
 		.requiredOption("-H, --harness <harness>", "Harness name")
 		.action(async (options) => {
 			const body = (await readJson()) ?? {};
+			const sessionKey = pickSessionKey(body);
 			const data = await deps.fetchFromDaemon<{ memoriesSaved?: number }>("/api/hooks/session-end", {
 				method: "POST",
 				headers: { "Content-Type": "application/json" },
 				body: JSON.stringify({
 					harness: options.harness,
```
PR-Reviewer-Ant: The new `session-end` forwarding still only sends `transcriptPath` and never forwards inline `transcript` from stdin. That diverges from the updated API/docs and from the PR description's claim that session-end can use `transcriptPath` or inline `transcript` for lossless capture. Any harness that supplies transcript text directly at session end will silently lose it here.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1328,10 +1367,98 @@ const signetPlugin = {
...
+				typeof event.messageCount === "number"
+					? event.messageCount
+					: typeof event.compactingCount === "number"
+						? event.compactingCount
+						: undefined;
+			const dedupeKey = `${sessionKey ?? "-"}|${messageCount ?? -1}`;
+			if (dedupeCompaction(beforeCompactions, dedupeKey)) {
+				return undefined;
+			}
+
```
PR-Reviewer-Ant, line 1560: Compaction-summary recovery only checks `ctx.sessionFile` (`extractCompactionSummary(event, ctx.sessionFile)`). The PR claims recovery when the hook payload only exposes metadata, but if OpenClaw provides `sessionFile` on the event payload rather than the hook context, this path returns no summary and drops the compaction artifact entirely. The new test passes `sessionFile` in both places, so this gap is not covered.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `3813977d`

[Automated Review] Found two blocking implementation gaps relative to the PR’s stated lossless-runtime goals.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The CLI session-end handler clearly omits the inline `transcript` field while the updated docs/PR description say session-end accepts inline transcript capture. The OpenClaw plugin’s compaction recovery path also only reads `ctx.sessionFile`, even though the PR claims recovery when the hook payload only exposes metadata; if `sessionFile` arrives on the event payload rather than hook context, the recovery path is skipped. I could not inspect the corresponding daemon implementation because only the diff excerpt in the review context is available here, so confidence is not high.

NicholaiVogel:
next round is in `894f46a0`.

addressed:
- `signet hook session-end` now preserves inline stdin `transcript` content instead of only forwarding `transcriptPath`, so the CLI matches the documented lossless session-end capture contract
- extracted `buildSessionEndBody()` and added a regression test proving inline transcript content is forwarded through the CLI hook path
- OpenClaw compaction recovery now resolves `sessionFile` from either the hook context or the event payload / nested compaction payload, so metadata-only compaction events still recover the summary correctly
- added a regression test covering `after_compaction` with `sessionFile` present only on the event payload
- updated `docs/CLI.md` so the stdin transcript forwarding behavior is explicit alongside the existing session-key / project fallback notes

validation:
- `bun test packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun run typecheck`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts docs/CLI.md`

packages/adapters/openclaw/src/index.ts
```diff
@@ -1328,10 +1381,99 @@ const signetPlugin = {
...
+			}
+			return {
+				prependContext: result.summaryPrompt ?? result.guidelines,
+			};
+		};
+
+		const handleAfterCompaction = async (
+			event: Record<string, unknown>,
+			ctx: {
+				sessionKey?: string;
```
PR-Reviewer-Ant, line 1580: The new compaction path only clears `injectedTurns` after a summary is successfully recovered and posted. If `after_compaction` fires with metadata only and `sessionFile` is missing/unreadable for that event, this returns early and leaves prompt dedupe intact, so the next turn will *not* re-inject context after compaction. That conflicts with the PR’s stated goal ('cleared prompt dedupe after compaction so context can be reconsidered on the next turn') and creates a continuity gap on exactly the degraded hook path this PR is trying to harden.

packages/cli/src/commands/hook.ts
```diff
@@ -88,6 +88,8 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
 					project: options.project || stdinProject,
```
PR-Reviewer-Ant: This PR updates `docs/API.md` to say `userMessage` is the preferred cleaned user turn for `POST /api/hooks/user-prompt-submit`, but the CLI bridge still only forwards `userPrompt` and never passes through `input.userMessage` / `input.user_message`. Any harness that emits the cleaned turn on stdin will silently lose it here, so the implementation does not fully match the new contract/documentation.

NicholaiVogel:
next round is in `25e43190`.

addressed:
- OpenClaw now clears prompt dedupe as soon as the compaction event is observed, even if no summary can be recovered, so the next turn can still reconsider context after compaction instead of staying artificially suppressed
- added a regression test proving `after_compaction` without any recoverable summary still causes the next prompt to re-inject context
- `signet hook user-prompt-submit` now forwards the preferred `userMessage` field when stdin provides it, while preserving legacy `userPrompt` compatibility in the same request body
- extracted `buildUserPromptSubmitBody()` and added a regression test proving the CLI carries both the preferred and legacy prompt fields together
- updated `docs/CLI.md` so the prompt-submit forwarding behavior matches the hook contract described in `docs/API.md`

validation:
- `bun test packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun run typecheck`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts docs/CLI.md`

packages/adapters/openclaw/src/index.ts
```diff
@@ -1328,10 +1381,99 @@ const signetPlugin = {
...
+			return undefined;
+		};
+
+		const dedupeCompaction = (map: Map<string, number>, key: string): boolean => {
+			const now = Date.now();
+			cleanupTimedMap(map, now, COMPACTION_HOOK_DEDUPE_MS);
+			const seenAt = map.get(key);
+			if (typeof seenAt === "number" && now - seenAt <= COMPACTION_HOOK_DEDUPE_MS) {
+				return true;
+			}
```
PR-Reviewer-Ant, line 1525: The new compaction dedupe keys are not agent-scoped. `handleBeforeCompaction()` dedupes on `${sessionKey}|${messageCount}` and `handleAfterCompaction()` dedupes on `${sessionKey}|${summary.slice(0, 120)}`, but both handlers already have `ctx.agentId` available. If two agents reuse the same `sessionKey` (which this PR explicitly says must be supported) within the 1s window, one agent's pre/post-compaction hook can be dropped locally and its temporal artifact never reaches the daemon. The dedupe key needs to include `agentId` to match the new `agent_id + session_key` contract.

NicholaiVogel:
next round is in `96c8e1bc`.

addressed:
- OpenClaw compaction dedupe keys now include `agentId` as well as `sessionKey`, so two agents sharing the same session key inside the 1s dedupe window no longer suppress each other's pre-compaction or compaction-complete hooks before the daemon ever sees them
- added a regression test proving `before_compaction` does not dedupe across different agents sharing the same session key
- added a regression test proving `after_compaction` / `compaction-complete` also remains agent-scoped under the same session key + summary text

validation:
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

packages/adapters/openclaw/src/index.ts
```diff
@@ -1328,10 +1381,99 @@ const signetPlugin = {
...
+			if (!result?.guidelines && !result?.summaryPrompt) {
+				return undefined;
+			}
+			return {
+				prependContext: result.summaryPrompt ?? result.guidelines,
+			};
+		};
+
+		const handleAfterCompaction = async (
+			event: Record<string, unknown>,
```
PR-Reviewer-Ant, line 1578: When `/api/hooks/pre-compaction` returns both `summaryPrompt` and `guidelines`, this forwards only `summaryPrompt` and drops `guidelines`. That changes the daemon contract from “prompt + constraints” to “first non-empty field wins”, so OpenClaw compaction can lose instructions that other harnesses still receive.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1328,10 +1381,99 @@ const signetPlugin = {
...
+
+			await onCompactionComplete("openclaw", summary, {
+				...opts,
+				project: opts.workspace,
+				sessionKey,
+			});
+		};
+
 		const ensureSessionStarted = async (
 			event: Record<string, unknown>,
```
PR-Reviewer-Ant, line 1619: The new fallback lineage path sends `project: opts.workspace` on `compaction-complete`. The PR description says this should preserve the request project when transcript persistence has not landed yet, but `workspace` is not the same thing as the active repo/project. In multi-repo use this can attach compaction artifacts to the wrong lineage and make later temporal expansion/synthesis inconsistent.

NicholaiVogel:
next round is in `d5d8738d`.

addressed:
- OpenClaw pre-compaction context no longer drops `guidelines` when `summaryPrompt` is also present; both are now preserved in the returned prepend context instead of one clobbering the other
- added a regression test proving the pre-compaction context includes both `summaryPrompt` and `guidelines`
- OpenClaw compaction-complete no longer hardcodes `opts.workspace` as the fallback project lineage; it now resolves project from the event payload / nested compaction payload / hook context first and only falls back to workspace when no lineage hint exists
- added a regression test proving event `cwd` wins over workspace fallback for compaction lineage

validation:
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

packages/cli/src/commands/hook.ts
```diff
@@ -166,11 +150,18 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
 		.requiredOption("-H, --harness <harness>", "Harness name")
 		.requiredOption("-s, --summary <summary>", "Session summary text")
 		.action(async (options) => {
+			const input = await readJson();
```
PR-Reviewer-Ant: The new `compaction-complete` plumbing does not actually carry through an existing `project` field from stdin. It only does `pickString(input?.cwd)`, so callers that already send `project` will still drop lineage and fall back to whatever the daemon infers. That diverges from the PR description/docs, which say this PR preserves explicit fallback project lineage through the CLI path.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1365,142 @@ const signetPlugin = {
...
+				sessionFile?: string;
+				agentId?: string;
+				project?: string;
+			},
+		): Promise<unknown> => {
+			if (!cfg.enabled || !daemonReachable) return undefined;
+			const sessionKey = resolveCompactionSessionKey(event, ctx);
+			const messageCount =
+				typeof event.messageCount === "number"
+					? event.messageCount
```
PR-Reviewer-Ant, line 1543: `handleAfterCompaction()` resets prompt dedupe with `injectedTurns.delete(sessionKey)`, but that cache is keyed only by `sessionKey`, not by `agentId + sessionKey`. In the exact multi-agent collision case this PR is trying to harden, one agent's compaction can clear another agent's prompt-dedupe state and cause an unintended re-injection on the next turn.

NicholaiVogel:
next round is in `7c9017d7`.

addressed:
- `signet hook compaction-complete` no longer drops an explicitly provided stdin `project`; it now prefers `project` and only falls back to `cwd`, with a regression test covering that precedence
- OpenClaw prompt/continuity bookkeeping is now keyed by scoped `agentId + sessionKey`, not just bare `sessionKey`, so claimed session state, injected-turn dedupe, compaction reset, and session-end cleanup all line up with the agent-scoped runtime contract
- OpenClaw now forwards `agentId` on `compaction-complete` and `session-end`, so the daemon sees the same scoped lineage the plugin uses internally
- added a regression test proving prompt injection does not dedupe across different agents sharing the same session key

validation:
- `bun test packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun run typecheck`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

packages/adapters/openclaw/src/index.ts
```diff
@@ -783,6 +788,60 @@ function buildSessionlessTurnKey(event: Record<string, unknown>, agentId: string
...
+	const compaction = isRecord(event.compaction) ? event.compaction : undefined;
+	const nested = readString(compaction?.summary);
+	if (nested) return nested;
+	if (!sessionFile || !existsSync(sessionFile)) return undefined;
+
+	try {
+		const lines = readFileSync(sessionFile, "utf-8")
+			.split("\n")
+			.map((line) => line.trim())
+			.filter((line) => line.length > 0);
```
PR-Reviewer-Ant, line 855: When the compaction hook only provides metadata, recovery now depends on reading `sessionFile`, but both the file read and per-line JSON parse failures are swallowed. If the file is unreadable, truncated, or contains one malformed trailing line, `handleAfterCompaction()` returns early and the compaction artifact is dropped with no log or retry signal. That is a silent continuity/data-loss failure on the exact path this PR claims to harden.

packages/cli/src/commands/hook.ts
```diff
@@ -166,11 +150,12 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
...
```
PR-Reviewer-Ant: `signet hook compaction-complete` now unconditionally does `await readJson()` before posting, even though the command already requires `--summary` and previously did not depend on stdin. That introduces a hidden contract change: existing callers that invoke the hook with flags only can now block or fail depending on how stdin is wired. The PR description says this is a compatibility-preserving fix, so this behavior change should either be guarded (only read stdin when piped) or covered explicitly as a breaking change.

NicholaiVogel:
next round is in `d626c684`.

addressed:
- OpenClaw no longer silently drops unrecoverable compaction summaries; if the event/session file path still cannot yield a summary, the plugin now emits an explicit warning before skipping the save, and the regression test now asserts that warning path
- `signet hook compaction-complete` no longer assumes stdin is always present just because the command exists; it now skips stdin parsing on TTY invocations, accepts explicit `--session-key`, `--project`, and `--agent-id` overrides, and uses those before falling back to stdin metadata

validation:
- `bun test packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun run typecheck`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1374,149 @@ const signetPlugin = {
...
+				compaction?.cwd,
+				compaction?.workspace,
+				ctx.project,
+				opts.workspace,
+			);
+		};
+
+		const dedupeCompaction = (map: Map<string, number>, key: string): boolean => {
+			const now = Date.now();
+			cleanupTimedMap(map, now, COMPACTION_HOOK_DEDUPE_MS);
```
PR-Reviewer-Ant, line 1521: Blocking: `resolveCompactionProject()` falls back to `opts.workspace` when the event/context does not carry a real project path. That means unrelated sessions can be recorded under the same global workspace lineage until transcript persistence catches up, which is the opposite of the PR's stated goal of preserving project lineage for `compaction-complete`. If the hook cannot supply a real per-session project, this should stay unset rather than collapse multiple repos into one lineage key.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1374,149 @@ const signetPlugin = {
...
+			return {
+				prependContext: parts.join("\n\n"),
+			};
+		};
+
+		const handleAfterCompaction = async (
+			event: Record<string, unknown>,
+			ctx: {
+				sessionKey?: string;
+				sessionFile?: string;
```
PR-Reviewer-Ant, line 1581: Blocking: post-compaction dedupe is keyed by `agentId | sessionKey | summary.slice(0, 120)` with a 1s TTL. Two legitimate compactions in the same session that share the same leading text will be silently dropped, so the temporal DAG can lose artifacts precisely on the path this PR is trying to harden. This needs a stronger event identity than a truncated summary prefix.

NicholaiVogel:
next round is in `fa4ee01d`.

addressed:
- OpenClaw compaction lineage no longer falls back to the adapter's static workspace path when there is no real project hint on the event or hook context, so we stop attaching compaction artifacts to a guessed branch path
- updated the compaction tests to assert that no synthetic `project` field is forwarded when lineage is unavailable
- compaction-complete dedupe is now keyed on the full recovered summary plus session-file context, not a 120-character summary prefix, so distinct compaction artifacts with the same prefix no longer get suppressed inside the 1-second dedupe window
- added a regression test covering two distinct summaries that share the same long prefix

validation:
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

NicholaiVogel:
resolved the merge conflict by merging current `origin/main` into this branch and taking the `packages/opencode-plugin/package.json` result from main's release state.

what i checked:
- conflict is gone
- `packages/opencode-plugin/package.json` now matches `0.77.4`
- `cd packages/opencode-plugin && bun run build && bun run typecheck` passes

merge commit: `c699599a`

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1374,148 @@ const signetPlugin = {
...
+				agentId?: string;
+				project?: string;
+			},
```
PR-Reviewer-Ant, line 1584: The new cross-event dedupe key includes `sessionFile`. That makes deduplication unstable when `after_compaction` and `session:compact:after` surface the same compaction through slightly different payload shapes (for example one event has a `sessionFile`, the other only has the summary). In that case the same compaction can be posted twice to `/api/hooks/compaction-complete`, which is exactly the duplicate-write path this PR is trying to prevent.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1374,148 @@ const signetPlugin = {
...
+		): Promise<unknown> => {
+			if (!cfg.enabled || !daemonReachable) return undefined;
+			const sessionKey = resolveCompactionSessionKey(event, ctx);
+			const messageCount =
+				typeof event.messageCount === "number"
```
PR-Reviewer-Ant, line 1542: The fallback count field here is `event.compactingCount`, but the rest of this PR and the added tests use `compactedCount`. When OpenClaw only supplies the latter and `messageCount` is absent, the plugin will silently omit the count from `pre-compaction`, weakening the daemon's summary guidance and falling short of the parity the PR description claims.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c699599a`

[Automated Review] Found two issues in the new OpenClaw compaction path that should be addressed before human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The findings are directly visible in the changed OpenClaw plugin code. The duplicate-dedupe risk is provable from the new dedupe key construction in the after-compaction handler, and the pre-compaction count bug is visible from the field name mismatch. I could not inspect the daemon-side implementation because the review bundle only included the OpenClaw/CLI portions of the patch, not the referenced daemon/core files.

NicholaiVogel:
next round is in `e4214da6`.

addressed:
- OpenClaw pre-compaction now accepts both `compactingCount` and `compactedCount` as fallback message-count fields instead of only one spelling, with a regression test proving the `compactedCount` path now reaches the daemon correctly
- after-compaction dedupe no longer keys on `sessionFile`, so the same compaction artifact is still deduplicated when the two OpenClaw hook aliases expose different session-file visibility
- added a regression test proving duplicate post-compaction writes still collapse correctly even when one hook call sees `sessionFile` and the other does not

validation:
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

packages/adapters/openclaw/src/index.ts
```diff
@@ -1456,16 +1647,35 @@ const signetPlugin = {
 		api.on("agent_end", async (_event: Record<string, unknown>, ctx: unknown): Promise<unknown> => {
 			if (!cfg.enabled) return undefined;
 
-			const { sessionKey } = resolveHookContext(ctx);
+			const { sessionKey, agentId } = resolveHookContext(ctx);
+			const scopedKey = buildScopedSessionKey(sessionKey, agentId);
```
PR-Reviewer-Ant, line 1750: Blocking: the new docs/specs claim OpenClaw plugin parity for lossless runtime capture, including session-end transcript persistence, but the `agent_end` hook still calls `onSessionEnd("openclaw", { ...opts, agentId, sessionKey })` without forwarding any `sessionFile`, `transcriptPath`, or inline transcript. In this diff, the only new session-end payload field is `agentId`; nothing actually supplies transcript content at session end. That means the daemon still cannot perform the advertised lossless session-end transcript capture for this runtime surface.

NicholaiVogel:
next round is in `c4a025a2`.

addressed the 2 latest review callouts:
- openclaw `agent_end` now forwards transcript/session-file lineage, session id/key, and project/cwd context into `/api/hooks/session-end` instead of dropping them on the floor
- added an explicit CLI regression test proving `buildCompactionCompleteBody()` still preserves legacy `sessionId` compaction payloads for lineage attachment

regressions/validation:
- `bun test packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

packages/cli/src/commands/hook.ts
```diff
@@ -252,6 +246,92 @@ function pickString(...values: unknown[]): string {
...
+		sessionKey: pickSessionKey(body),
+		transcriptPath: pickString(body?.transcript_path, body?.transcriptPath),
+		transcript: pickString(body?.transcript),
+		...(lastAssistantMessage ? { lastAssistantMessage } : {}),
+	};
+}
+
+export function buildCompactionCompleteBody(
+	input: Record<string, unknown> | null,
+	harness: string,
```
PR-Reviewer-Ant, line 299: buildSessionEndBody now sets `sessionId` from the resolved `sessionKey` (`sessionId: sessionKey`) instead of preserving a distinct `session_id` / `sessionId` input when one exists. That changes the wire contract compared with the previous inline payload builder and can mis-attribute session-end lineage whenever a harness provides both identifiers with different meanings. This PR claims to harden transcript/session-summary lineage, so collapsing two identifiers into one is a correctness regression.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c4a025a2`

[Automated Review] I found one blocking payload regression in the new CLI hook helpers.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The regression is directly visible in the added helper code in packages/cli/src/commands/hook.ts: buildSessionEndBody now hardcodes sessionId from the resolved sessionKey instead of preserving any distinct sessionId field from stdin. I can prove the payload shape changed from this diff alone, but the daemon-side consumer for sessionId is not included in the provided diff, so impact severity is inferred from the PR's own lineage/temporal-scoping claims.

NicholaiVogel:
next round is in `0a216ce6`.

addressed the latest CLI payload regression:
- `buildSessionEndBody()` now preserves a distinct legacy `sessionId` from stdin instead of hardwiring it to the resolved `sessionKey`
- added a regression test proving canonical `sessionKey` and legacy `sessionId` can travel together without collapsing lineage fields

validation:
- `bun test packages/cli/src/commands/hook.test.ts`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts`

packages/cli/src/commands/hook.ts
```diff
@@ -252,6 +246,93 @@ function pickString(...values: unknown[]): string {
...
+	harness: string;
+	summary: string;
+	agentId: string;
+	sessionKey: string;
+	project: string;
+} {
+	const body = input;
+	return {
+		harness,
+		summary,
```
PR-Reviewer-Ant, line 319: `buildCompactionCompleteBody()` now emits `agentId`, `sessionKey`, and `project` as `""` when they are absent. That is not equivalent to omitting the fields: the daemon-side compaction lineage logic added in this PR uses these values for scoping/fallback, so blank strings can be treated as an explicit empty lineage rather than 'not provided', causing unrelated compaction artifacts to collide on the empty scope instead of falling back correctly. This is a behavior change from the old `{ harness, summary }` payload and undermines the PR's stated lineage-preservation fix.

NicholaiVogel:
next round is in `ee6564c8`.

addressed the latest compaction CLI payload regression:
- `buildCompactionCompleteBody()` no longer serializes blank `agentId`, `sessionKey`, or `project` strings when those lineage fields are unset
- optional compaction lineage is now omitted entirely unless it is actually present, so we do not collapse requests onto empty lineage values
- added a regression test proving the helper returns only `harness` + `summary` when no optional lineage is supplied

validation:
- `bun test packages/cli/src/commands/hook.test.ts`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts`

NicholaiVogel:
small correction: the pushed commit for the last round is `824495b5`.

the earlier comment described the right fix, i just fat-fingered the hash.

packages/cli/src/commands/hook.ts
```diff
@@ -165,12 +149,22 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
 		.description("Save session summary after compaction")
 		.requiredOption("-H, --harness <harness>", "Harness name")
 		.requiredOption("-s, --summary <summary>", "Session summary text")
+		.option("--session-key <key>", "Session key")
+		.option("--project <project>", "Project path")
+		.option("--agent-id <id>", "Agent ID")
 		.action(async (options) => {
+			const input = process.stdin.isTTY ? null : await readJson();
 			const data = await deps.fetchFromDaemon<{ success?: boolean; memoryId?: number; error?: string }>(
```
PR-Reviewer-Ant, line 157: This introduces a regression for `signet hook compaction-complete`: when stdin is not a TTY, the command now always waits on `readJson()` before posting the hook, even if `--summary`/`--session-key`/`--project` are already provided on the command line. Many hook runners launch subprocesses with non-TTY stdin but do not pipe a JSON payload here, so this can block or fail the compaction-complete hook entirely.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1374,176 @@ const signetPlugin = {
...
+			const result = await onPreCompaction("openclaw", {
+				...opts,
+				sessionKey,
+				messageCount,
+			});
+			const parts = [result?.summaryPrompt, result?.guidelines].filter(
+				(value) => typeof value === "string" && value.length > 0,
+			);
+			if (parts.length === 0) {
+				return undefined;
```
PR-Reviewer-Ant, line 1570: The new compaction dedupe key is too coarse: it collapses events to `agentId + sessionKey + messageCount` within a 1s window. Two real compactions for the same agent/session that happen quickly with the same count will cause the later pre-compaction request to be dropped. The after-compaction path applies the same time-window pattern keyed only by summary text, so repeated identical summaries can also be suppressed. This risks losing real compaction artifacts rather than just filtering alias duplicates.

NicholaiVogel:
next round is in `1379db73`.

addressed the latest blocking hook/compaction review round:
- `signet hook compaction-complete` no longer reads stdin when the caller already provided full lineage on flags (`--session-key`, `--project`, `--agent-id`), so non-tty hook runners do not get stuck waiting on unused stdin
- extracted `shouldReadCompactionInput()` and added a regression test proving the CLI skips stdin in that fully-specified flag path
- OpenClaw compaction dedupe now fingerprints the full event shape (agent/session + ids/counts/tokenCount + summary) instead of the earlier coarse time-window keys, which keeps alias duplicates collapsed while allowing distinct compactions with reused counts or reused summary text through
- added regressions proving:
  - distinct pre-compaction events with the same message count are not collapsed
  - distinct compaction-complete events with the same summary text are not collapsed
  - duplicate alias events still dedupe correctly when hook payload visibility differs

validation:
- `bun test packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/cli/src/commands/hook.ts packages/cli/src/commands/hook.test.ts packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

NicholaiVogel:
small correction again: the pushed commit for that round is `6922b389`.

i typed the note before git handed me the real hash. same fix, less clown shoes.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1407,183 @@ const signetPlugin = {
...
+
+		const resolveCompactionProject = (
+			event: Record<string, unknown>,
+			ctx: {
+				project?: string;
+			},
+		): string | undefined => {
+			const compaction = isRecord(event.compaction) ? event.compaction : undefined;
+			return firstNonEmptyString(
+				event.project,
```
PR-Reviewer-Ant, line 1482: The PR description says OpenClaw now preserves `project` lineage when `compaction-complete` arrives before transcript persistence, including degraded cases where the hook payload only exposes metadata. This implementation only derives `project` from the event/context (`project`/`cwd`/`workspace`) and never from `sessionFile`, even though the same degraded path already reads the summary back from `sessionFile`. In the metadata-only case, the plugin will still send no `project`, so the daemon cannot reconstruct lineage before transcript storage lands.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1315,23 +1407,183 @@ const signetPlugin = {
...
+				sessionFile?: string;
+			},
+		): string | undefined => firstNonEmptyString(event.transcriptPath, event.sessionFile, ctx.sessionFile);
+
+		const resolveSessionEndProject = (
+			event: Record<string, unknown>,
+			ctx: {
+				project?: string;
+			},
+		): string | undefined => firstNonEmptyString(event.cwd, event.project, event.workspace, ctx.project);
```
PR-Reviewer-Ant, line 1517: The new pre-compaction path reads `messageCount`, `compactingCount`, and `compactedCount` only from the top-level event. Other helpers in this same change already support nested `event.compaction.*` fields, so a nested compaction payload will dedupe correctly but still call `/api/hooks/pre-compaction` with `messageCount: undefined`. That weakens parity across OpenClaw event shapes and can change summary guidance in exactly the degraded-hook cases this PR is trying to harden.

NicholaiVogel:
next round is in `f0da71a3`.

addressed the latest OpenClaw parity/lineage gaps:
- compaction project lineage now recovers from the session transcript header when the event lacks explicit `project` / `cwd` hints, so `sessionFile`-only compaction events can still attach artifacts back to the right branch lineage without falling back to a guessed workspace
- pre-compaction message-count fallback now also honors nested `event.compaction.compactingCount` / `compactedCount` when wrappers surface compaction metadata under a nested payload
- added regressions proving:
  - project lineage is recovered from a `sessionFile` header containing `cwd`
  - nested compaction counts still reach `/api/hooks/pre-compaction`

validation:
- `bun test packages/adapters/openclaw/src/index.test.ts`
- `cd packages/adapters/openclaw && bun run build`
- `bun x biome check packages/adapters/openclaw/src/index.ts packages/adapters/openclaw/src/index.test.ts`

NicholaiVogel:
small correction: that pushed commit is `a74ab7d5`.

same fix, correct hash this time.

PR #346, opened 2026-03-26
feat: RaBitQ compressed vector search module

BusyBee3333:
## Summary

Adds a pure TypeScript implementation of the **RaBitQ** (Random Bit Quantization) algorithm as an optional compressed vector search pre-filter for Signet's hybrid recall pipeline.

### What it does

Compresses 768-dim Float32 embeddings into 4-bit packed representations using:
- **Householder QR decomposition** for random orthogonal rotation
- **Beta distribution codebook** (inverse CDF via Newton-bisection) for optimal quantization centroids
- **4-bit packed Uint8Array** storage with per-vector norm + mean metadata
- **Centroid lookup table (LUT)** search for fast approximate dot products

When enabled (`memory.search.compressed_vector_enabled: true`), `hybridRecall()` uses RaBitQ as a fast pre-filter (approximate top-50) → exact cosine re-rank (top-10), falling back to sqlite-vec if the compressed index isn't available.

### Files

| File | Lines | Description |
|------|-------|-------------|
| `packages/daemon/src/rabitq.ts` | 860 | Core algorithm: rotation, codebook, quantize, dequantize, compressed search |
| `packages/daemon/src/rabitq.test.ts` | 672 | 28 tests: round-trip accuracy, search recall, edge cases, perf benchmarks |
| `packages/daemon/src/rabitq-index.ts` | 356 | Integration layer: loads embeddings from SQLite, builds/caches index |
| `packages/daemon/src/memory-search.ts` | modified | Compressed pre-filter path behind config flag |
| `packages/daemon/src/memory-config.ts` | modified | Added `compressed_vector_enabled` config field |

### Test Results

- **28/28 tests pass** (rotation orthogonality, codebook properties, round-trip accuracy, search recall, edge cases, performance)
- **75% Recall@10** on real Signet embeddings (8,400+ memories, 768-dim nomic-embed)
- **4.89× compression** at 10K vectors (rotation matrix overhead amortizes at scale)
- Zero new typecheck errors in daemon package

### Tested on real data

Ran against a live Signet instance with 8,439 embeddings. Recall@10 ranged from 50-100% across different query types, with 75% average. The two-phase design (compressed pre-filter → exact re-rank) recovers near-exact results since compressed search only needs to surface candidates, not rank them perfectly.

### Tradeoffs

- Speed benefit kicks in at ~10K+ vectors; at smaller scales brute force is comparable
- 768×768 rotation matrix = 2.3MB fixed overhead (negligible at scale)
- Default **off** — zero impact on existing users unless explicitly enabled

### DIFF
--- a/packages/daemon/src/rabitq-index.ts
+++ b/packages/daemon/src/rabitq-index.ts
@@ -10,6 +10,23 @@
  * invalidation via row-count check).
  */
 
+// ---------------------------------------------------------------------------
+// Memory overhead notes
+// ---------------------------------------------------------------------------
+//
+// For dim=768, the rotation matrix alone is 768x768x4 bytes = 2.36 MB.
+// Each compressed vector is ~392 bytes vs 3072 bytes uncompressed (7.8x).
+// However, the rotation matrix overhead means compression only saves net
+// memory above ~5,000 vectors. Below that threshold, sqlite-vec brute-force
+// search is cheaper in both memory and wall-clock time.
+//
+// Approximate break-even analysis (dim=768, 4-bit):
+//   Overhead:  2.36 MB (rotation) + 64 B (codebook) = 2.36 MB
+//   Savings per vector: 3072 - 392 = 2680 bytes
+//   Break-even: 2,360,000 / 2,680 = 881 vectors (memory only)
+//   Practical break-even with build cost: ~5,000 vectors
+//
+
 import { getDbAccessor } from "./db-accessor";
 import { logger } from "./logger";
 import {
@@ -196,8 +213,8 @@ function getEmbeddingFingerprint(): { count: number; hash: string } {
 /**
  * Ensure the compressed index is built and up-to-date.
  *
- * Lazily rebuilds when the embedding count changes. Thread-safe via
- * build-in-progress flag (skips if another build is running).
+ * Lazily rebuilds when the embedding count changes. Single-threaded guard via
+ * build-in-progress flag (Bun JS is single-threaded; not safe across worker threads).
  *
  * @param config - RaBitQ configuration
  * @returns The current compressed index, or null if unavailable
@@ -206,6 +223,14 @@ export function ensureIndex(config: RaBitQConfig): CompressedIndex | null {
 	if (!config.enabled) return null;
 
 	const { count: currentCount, hash: currentHash } = getEmbeddingFingerprint();
+
+	// Skip building for very small datasets where brute-force is cheaper
+	if (currentCount > 0 && currentCount < 1000) {
+		logger.info("memory", "RaBitQ: skipping index build, brute-force is cheaper for <1000 vectors", {
+			vectorCount: currentCount,
+		});
+		return null;
+	}
 
 	// Check if cached index is still valid (count + content fingerprint + config params)
 	if (
--- a/packages/daemon/src/rabitq.test.ts
+++ b/packages/daemon/src/rabitq.test.ts
@@ -230,8 +230,8 @@ describe("quantize + dequantize", () => {
 
 		for (let i = 0; i < vectors.length; i++) {
 			const sim = cosine(vectors[i], restored[i]);
-			// Quantization introduces error; cosine > 0.7 is reasonable for 4-bit
-			expect(sim).toBeGreaterThan(0.7);
+			// With maxDev stored, round-trip accuracy is high; cosine > 0.85 for 4-bit at dim=64
+			expect(sim).toBeGreaterThan(0.85);
 		}
 	});
 
--- a/packages/daemon/src/rabitq.ts
+++ b/packages/daemon/src/rabitq.ts
@@ -25,6 +25,8 @@ export interface CompressedVector {
 	norm: number;
 	/** Mean of the rotated vector components. */
 	mean: number;
+	/** Max absolute deviation from mean (per-vector scale factor for dequantization). */
+	maxDev: number;
 	/** Packed quantized indices (2 indices per byte for 4-bit). */
 	codes: Uint8Array;
 }
@@ -585,6 +587,7 @@ export function quantize(
 			id: ids[vi],
 			norm,
 			mean,
+			maxDev: scale,
 			codes,
 		});
 	}
@@ -644,13 +647,10 @@ export function dequantize(index: CompressedIndex): Float32Array[] {
 			}
 		}
 
-		// 2. The centroid values are in [-1, 1] (normalized space).
-		// We stored mean but not maxDev — for dequantization we estimate
-		// scale from the norm and mean. Since we're doing approximate
-		// reconstruction, we use norm/sqrt(dim) as scale estimate.
-		// rotated[i] ≈ (original_rotated[i] - mean) / scale
-		// So: original_rotated[i] ≈ rotated[i] * scale + mean
-		const scale = cv.norm / Math.sqrt(dim);
+		// 2. Rescale centroid values from [-1, 1] back to original rotated space.
+		// During quantization: normalized[i] = (rotated[i] - mean) / maxDev
+		// So: rotated[i] = centroid[idx] * maxDev + mean
+		const scale = cv.maxDev;
 		for (let i = 0; i < dim; i++) {
 			rotated[i] = rotated[i] * scale + cv.mean;
 		}
@@ -777,14 +777,14 @@ export function compressedSearch(query: Float32Array, index: CompressedIndex, to
 			}
 		}
 
-		// The centroid values represent (rotated[i] - mean) / scale,
+		// The centroid values represent (rotated[i] - mean) / maxDev,
 		// and we want dot(query, original_vector) ≈ dot(R*query, R*vec)
 		// = dot(rotatedQuery, rotated).
 		//
-		// rotated[i] ≈ codebook[idx] * scale + mean
-		// dot ≈ sum_i rotatedQuery[i] * (codebook[idx_i] * scale + mean)
-		//      = scale * sum_i rotatedQuery[i] * codebook[idx_i] + mean * sum_i rotatedQuery[i]
-		const scale = cv.norm / Math.sqrt(dim);
+		// rotated[i] ≈ codebook[idx] * maxDev + mean
+		// dot ≈ sum_i rotatedQuery[i] * (codebook[idx_i] * maxDev + mean)
+		//      = maxDev * sum_i rotatedQuery[i] * codebook[idx_i] + mean * sum_i rotatedQuery[i]
+		const scale = cv.maxDev;
 		let queryRotatedSum = 0;
 		for (let i = 0; i < dim; i++) {
 			queryRotatedSum += rotatedQuery[i];


PR-Reviewer-Ant:
Hi @BusyBee3333 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: RaBitQ compressed vector search module` (commit `89d5fb6a`) now and I'll follow up shortly with feedback.

packages/daemon/src/memory-search.ts
```diff
@@ -341,28 +343,59 @@ export async function hybridRecall(
...
-			getDbAccessor().withReadDb((db) => {
-				const vecResults = vectorSearch(db as any, queryVecF32!, {
-					limit: vecLimit,
-					type: params.type as "fact" | "preference" | "decision" | undefined,
-				});
-				for (const r of vecResults) {
-					vectorMap.set(r.id, r.score);
+		// Try compressed vector search first (if enabled)
+		const rabitqConfig: RaBitQConfig = {
+			...DEFAULT_RABITQ_CONFIG,
```
PR-Reviewer-Ant, line 400: Blocking: the compressed path is not a drop-in replacement for the existing vector search. `vectorSearch()` was called with the request's `type` filter, but `compressedVectorSearchWithRerank()` has no way to apply `type`, `project`, or scope constraints before candidate truncation. In filtered searches, out-of-filter candidates can consume the entire compressed top-N window and then get discarded later during hydration, causing valid in-filter memories to disappear from results.

packages/daemon/src/memory-search.ts
```diff
@@ -341,28 +343,59 @@ export async function hybridRecall(
...
+	// --- Vector search via sqlite-vec (or compressed pre-filter) ---
 	// sqlite-vec cannot pre-filter by scope, so scoped queries over-fetch
 	// (2x top_k) and rely on hydration-time scope filtering (line ~685).
 	// This ensures scoped queries still benefit from vector similarity
 	// when graph traversal yields no focal entities.
+	//
+	// When compressed_vector_enabled is true, use RaBitQ compressed search
+	// as a fast pre-filter (approximate top-50) then exact re-rank (top-10).
+	// Falls back to sqlite-vec if the compressed index is unavailable.
 	const vectorMap = new Map<string, number>();
```
PR-Reviewer-Ant, line 395: Blocking: this branch hardcodes the vector candidate depth to RaBitQ defaults (`prefilterTopK: 50`, `rerankTopK: 10`) instead of preserving the existing search depth (`cfg.search.top_k`, or `top_k * 2` for scoped queries). That means the optional pre-filter changes recall semantics even when it works: the default daemon config asks for 20 vector hits, but the compressed path can only contribute 10. This diverges from the PR's claim that it is an optional pre-filter behind a flag with zero behavioral impact unless enabled.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,356 @@
...
+	}
+}
+
+// ---------------------------------------------------------------------------
+// Public API
+// ---------------------------------------------------------------------------
+
+/**
+ * Ensure the compressed index is built and up-to-date.
+ *
```
PR-Reviewer-Ant, line 215: Blocking: index freshness is keyed only on `COUNT(*)`. Any in-place re-embedding, vector update, or delete+insert that keeps the row count stable leaves `cachedState` stale indefinitely. This repo has repair/reembed flows, and this PR did not add any `invalidateIndex()` call sites, so compressed search can return neighbors for outdated embeddings until process restart.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,395 @@
...
+			dimensions: config.dimensions,
+			bits: config.bits,
+			elapsedMs: Math.round(elapsedMs),
+		});
+
+		return index;
+	} catch (e) {
+		logger.error("memory", "RaBitQ: index build failed", e instanceof Error ? e : new Error(String(e)));
+		return null;
+	}
```
PR-Reviewer-Ant, line 175: The cache invalidation fingerprint does not actually detect re-embeds. `MAX(e.id)` stays the same for in-place updates, and `SUM(length(e.vector))` is constant because all embeddings have the same fixed byte length. After an embedding is recomputed, `ensureIndex()` will keep serving the old compressed index until row count changes, so recall can stay stale indefinitely. This diverges from the PR description's 'lazy invalidation via row-count check / in-place re-embeds' claim and needs a real content/version signal (for example updated_at, rowid max if rows are replaced, or an explicit invalidation hook).

packages/daemon/src/memory-search.ts
```diff
@@ -341,28 +344,64 @@ export async function hybridRecall(
...
 	if (queryVecF32) {
+		// Try compressed vector search first (if enabled)
 		const vecLimit = scoped ? cfg.search.top_k * 2 : cfg.search.top_k;
-		try {
-			getDbAccessor().withReadDb((db) => {
-				const vecResults = vectorSearch(db as any, queryVecF32!, {
-					limit: vecLimit,
+		const rabitqConfig: RaBitQConfig = {
```
PR-Reviewer-Ant, line 399: The compressed path hard-codes RaBitQ to `DEFAULT_RABITQ_CONFIG.dimensions` (768) instead of using `cfg.embedding.dimensions`. On any non-768 embedding provider/config, `buildIndex()` will skip every row as a dimension mismatch and the feature silently never becomes available. If the intent is 'optional compressed vector search for Signet embeddings', this needs to be wired to the active embedding dimension or explicitly rejected at config-load time.

packages/daemon/src/memory-search.ts
```diff
@@ -341,28 +344,64 @@ export async function hybridRecall(
 		logger.warn("memory", "Embedding failed", { error: String(e) });
 	}
 
-	// --- Vector search via sqlite-vec ---
+	// --- Vector search via sqlite-vec (or compressed pre-filter) ---
 	// sqlite-vec cannot pre-filter by scope, so scoped queries over-fetch
 	// (2x top_k) and rely on hydration-time scope filtering (line ~685).
 	// This ensures scoped queries still benefit from vector similarity
 	// when graph traversal yields no focal entities.
+	//
```
PR-Reviewer-Ant, line 391: This changes daemon recall behavior behind `compressed_vector_enabled`, but there is no corresponding `packages/daemon-rs/` update in the PR. The repo's Rust parity rule says behavioral changes in `@signet/daemon` must be mirrored there; otherwise shadow mode will report systematic divergences for any request that takes the new path.

BusyBee3333:
All three blocking items from the `aec9c8eb` review have been addressed in `9119d99c`:

1. **Fingerprint** — replaced `MAX(e.id)+SUM(length(vector))` with `MAX(e.rowid)+SUM(e.rowid)` (rowid advances on delete+reinsert). Added explicit `invalidateRaBitQIndex()` hook in maintenance-worker for pure in-place UPDATE cases.
2. **Hardcoded 768** — wired `cfg.embedding.dimensions` into RaBitQConfig in memory-search.ts.
3. **daemon-rs parity** — added `compressed_vector_enabled: bool` to `SearchOptions` in signet-core/src/search.rs.

Ready for re-review.

packages/daemon/src/memory-config.ts
```diff
@@ -1060,12 +783,8 @@ export function loadMemoryConfig(agentsDir: string): ResolvedMemoryConfig {
 				defaults.embedding.provider = "none";
```
PR-Reviewer-Ant, line 823: The PR description says this is enabled with `memory.search.compressed_vector_enabled: true`, but this loader only reads `yaml.search` and never consults `yaml.memory.search`. In that documented configuration shape the feature silently stays disabled, so the implementation does not currently achieve what the PR claims.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,402 @@
...
+// Public API
+// ---------------------------------------------------------------------------
+
+/**
+ * Ensure the compressed index is built and up-to-date.
+ *
+ * Lazily rebuilds when the embedding count changes. Thread-safe via
+ * build-in-progress flag (skips if another build is running).
+ *
+ * @param config - RaBitQ configuration
```
PR-Reviewer-Ant, line 219: The cache-validity check ignores the active RaBitQ config (`dimensions`, `bits`, `seed`). If the daemon reloads config and the embedding dimension changes (or any RaBitQ parameter changes) without a row-count/hash change, `ensureIndex()` will return a stale index built for the old shape. That can produce dimension-mismatch failures or incorrect search results while the feature is enabled.

NicholaiVogel:
this is cool work, and i do think the RaBitQ prototype itself is worth keeping around.

my hesitation is mostly about where it lands, not whether the implementation took real effort.

right now this PR wires a second retrieval path into `hybridRecall()`, adds config surface area, adds invalidation logic, and touches parity/runtime behavior. that’s a pretty big architectural step for something we haven’t really proven is the bottleneck yet.

my gut says the cleanest path to merge is to split this:

- keep the standalone RaBitQ implementation and tests as an experimental/prototype module
- drop the recall-path integration, config plumbing, maintenance invalidation, and rust parity shim from this PR

that gives us a way to preserve the interesting research/prototype work without committing core recall to a second vector backend before we’ve profiled where the time is actually going.

if the benchmarks later show this is the right direction, we can revisit integration in a follow-up with proper profiling, spec/docs, and a clearer decision about whether the long-term home should be native/rust rather than TS daemon recall plumbing.

so basically: i’m not against the idea, i just think this wants to land in a smaller shape first.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,412 @@
...
+ */
+export function compressedVectorSearchWithRerank(
+	queryVector: Float32Array,
+	config: RaBitQConfig,
+	filters?: CompressedSearchFilters,
+): Array<{ id: string; score: number }> {
+	const index = ensureIndex(config);
+	if (index === null) return [];
+
+	// Use pipeline limits when provided, otherwise fall back to config defaults
```
PR-Reviewer-Ant, line 339: Blocking: this rerank query only filters by `m.type`. It does not carry the retrieval constraints that matter for correctness here (`scope`, `project`, `agent_id`/visibility). Unlike the old sqlite-vec path, this branch truncates to `rerankK` before the outer hydration filter runs, so restricted searches can permanently lose valid matches when high-similarity out-of-scope memories fill the candidate set. That also diverges from the PR description/comment claiming project/scope filters are applied during hydration.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,432 @@
...
+ * @param queryVector - Query embedding (Float32Array, dim=768)
+ * @param config - RaBitQ configuration
+ * @param filters - Optional pipeline filters (type, limit overrides)
+ * @returns Array of {id, score} sorted by exact cosine similarity
+ */
+export function compressedVectorSearchWithRerank(
+	queryVector: Float32Array,
+	config: RaBitQConfig,
+	filters?: CompressedSearchFilters,
+): Array<{ id: string; score: number }> {
```
PR-Reviewer-Ant, line 335: The new rerank query still does not apply agent visibility filtering (`agent_id`/`readPolicy`/`policyGroup`) before slicing to `rerankK`. That means in shared/group/isolated recall modes, compressed candidates from other agents can consume the entire vector budget and push valid authorized memories out before `hybridRecall()`'s final hydration filter runs. This is the same class of starvation bug you just fixed for scope/project, but now on the repo's required agent-scoping path.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,455 @@
...
+	configSeed: number;
+}
+
+let cachedState: IndexState | null = null;
+let buildInProgress = false;
+
+// ---------------------------------------------------------------------------
+// Index Construction
+// ---------------------------------------------------------------------------
+
```
PR-Reviewer-Ant, line 92: The PR description positions this as a fast optional pre-filter, but the implementation still does a synchronous full-table index build from the recall path (`ensureIndex()` -> `buildIndex()`). On the first enabled search after startup/invalidation, this request now has to read every embedding, generate the full rotation matrix, and quantize the dataset inline before it can answer. In the daemon's single-process request path, that is an availability/regression risk rather than a harmless cache miss. If this integration stays in `hybridRecall()`, it needs an explicit warm-up/background build path or some other serialization that does not block user queries.

packages/daemon/src/memory-search.ts
```diff
@@ -377,28 +383,74 @@ export async function hybridRecall(
 		logger.warn("memory", "Embedding failed", { error: String(e) });
```
PR-Reviewer-Ant, line 383: This introduces a second vector-retrieval backend into `hybridRecall()`, but the added tests are algorithm-only (`rabitq.test.ts`) and do not exercise the actual recall integration: scoped search parity, agent visibility parity, sqlite fallback, or cache rebuild behavior. Given this exact path has already needed multiple review fixes for filtering semantics, the lack of a `hybridRecall()` regression test is a meaningful maintainability gap and does not meet the repository's guardrail/testing guidance for shipping a new retrieval path.

packages/daemon/src/rabitq-index.ts
```diff
@@ -0,0 +1,480 @@
...
+		const index = quantize(vectors, ids, rotationMatrix, codebook, config.bits);
+
+		const elapsedMs = performance.now() - startMs;
+		logger.info("memory", "RaBitQ: built compressed index", {
+			vectors: vectors.length,
+			dimensions: config.dimensions,
+			bits: config.bits,
+			elapsedMs: Math.round(elapsedMs),
+		});
+
```
PR-Reviewer-Ant, line 170: The cache invalidation strategy is still incomplete. `getEmbeddingFingerprint()` only changes on row insert/delete patterns, not when an existing embedding row is updated in place. This PR adds `invalidateIndex()`, but there is no corresponding call wired into embedding write / re-embed paths here, so once the compressed index is built it can serve stale vectors indefinitely until restart or a structural row change happens. That is a blocking correctness issue for recall results.

packages/daemon/src/memory-search.ts
PR-Reviewer-Ant, line 383: This introduces new daemon-side recall behavior behind `compressed_vector_enabled`, but the repo’s daemon parity rule requires matching behavioral changes in `packages/daemon-rs/`. No Rust shadow-path update is included in this PR, so shadow mode will diverge and any Rust-side comparison/logging becomes noisy or misleading. This should be kept in sync before rollout.

PR #347, opened 2026-03-26
fix(daemon): persist extraction fallback status and control

aaf2tbz:
## Summary
- persist extraction provider runtime status in `/api/status`, including degraded vs blocked startup states, fallback policy, reason, and timestamp
- add `memory.pipelineV2.extraction.fallbackProvider` (`ollama | none`) so operators can disable silent rerouting
- dead-letter queued/new extraction jobs when startup preflight blocks extraction and surface the condition in `signet status`
- add regression tests plus docs/spec registry updates for issue #320

## Type

- [ ] `feat` — new user-facing feature (bumps minor)
- [x] `fix` — bug fix
- [ ] `refactor` — restructure without behavior change
- [ ] `chore` — build, deps, config, docs
- [ ] `perf` — performance improvement
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/cli` / dashboard
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs, specs

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

## Testing

- [x] `bun test` passes (relevant test files)
- [ ] `bun run typecheck` passes
- [ ] `bun run lint` passes
- [x] Tested against running daemon
- [ ] N/A

## Validation
- `bun test packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/extraction-fallback.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/runtime.test.ts`
- `cd packages/daemon && bun build ./src/daemon.ts --outdir /tmp/signet-daemon-build --target bun`
- `cd packages/cli && bun build ./src/features/health.ts --outdir /tmp/signet-cli-health-build --target node`
- `cd packages/cli && bun build ./src/lib/runtime.ts --outdir /tmp/signet-cli-runtime-build --target node`

## Migration Notes (if applicable)

- [ ] Migration is idempotent
[...]

PR #345, opened 2026-03-26
Integrate Forge into signetai as first-party native harness

aaf2tbz:
## Summary

This PR cuts Forge over into the `signetai` monorepo as Signet's first-party native harness and reference runtime implementation.

It does **not** merge or remove the old GitHub repo automatically, but it makes `signetai` the canonical source of truth for Forge development, release automation, and Signet integration.

## Changes

- vendored the full Forge Rust workspace into `packages/forge/`
- added first-party Forge management commands to the Signet CLI (`install`, `update`, `status`, `doctor`)
- integrated Forge into setup/configure flows, templates, harness detection, and daemon harness reporting
- added matching Forge harness reporting to `packages/daemon-rs` for Bun/Rust parity on `/api/harnesses`
- added dedicated Forge release automation and independent Forge version metadata
- hardened the follow-up install flow so Signet only manages Forge binaries backed by a matching Signet install record, verifies release checksums, runs a passive Forge binary fingerprint check from binary contents, and stages replacements on the target filesystem before the final atomic swap
- expanded the managed Forge release/install matrix to macOS arm64, macOS x64, Linux x64, and Linux arm64, with the workflow, installer, tests, and docs kept in sync
- changed latest-release resolution to select the highest stable Forge semver while excluding GitHub drafts/prereleases from the default install/update path
- kept `packages/forge/forge-version.json` as the Forge version source of truth and added sync/CI guards for the template manifest copies used by the managed installer
- broadened passive Forge detection so standalone/source Forge builds can still be detected across setup/status/daemon reporting without weakening managed-install ownership checks
- gated Forge setup/configuration on the managed install support matrix when no standalone Forge binary is already present, so unsupported platforms fail early instead of mid-setup
- serialized managed Forge install/update with a filesystem lock to prevent concurrent replacement races
- excluded Forge-only manifest sync files from the main Signet release trigger so Forge stays independently versioned in practice, not just on paper

## Type

- [x] `feat` — new user-facing feature (bumps minor)
- [ ] `fix` — bug fix
- [ ] `refactor` — restructure without behavior change
- [ ] `chore` — build, deps, config, docs
- [ ] `perf` — performance improvement
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/cli` / dashboard
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
[...]

PR #348, opened 2026-03-26
fix(daemon): persist extraction fallback status and control

### LINKED ISSUE #320: feat(daemon): provider preflight check at startup + explicit fallback control
## Summary

---

## Current Behavior

---

## Desired Behavior

---

## Use Cases

**3. Monitoring integration**

---

## Proposed Implementation

```
```

```yaml
pipeline:
  extraction:
    provider: claude-code
```

---

## Expected Impact

---

## Related

aaf2tbz:
## Summary
- persist extraction provider runtime status in `/api/status`, including degraded vs blocked startup states, fallback policy, reason, and timestamp
- add `memory.pipelineV2.extraction.fallbackProvider` (`ollama | none`) so operators can disable silent rerouting
- dead-letter queued extraction jobs at startup (plus new blocked remember-route extraction writes) when startup preflight blocks extraction, and surface the condition in `signet status`
- add regression tests plus docs/spec registry updates for issue #320
- supersedes the previously closed PR #347

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs, specs

## PR Readiness (MANDATORY)

- [x] Lint/typecheck/tests pass locally for the touched paths

## Testing

- [ ] `bun run lint` passes
- [ ] N/A

## Validation
- `cd packages/daemon && bun test ./src/memory-config.test.ts ./src/pipeline/extraction-fallback.test.ts`
- `cd packages/cli && bun test ./src/features/health.test.ts ./src/lib/runtime.test.ts`
- `cd packages/daemon && bun build ./src/daemon.ts --outdir /tmp/signet-daemon-issue320-build --target bun`
- `cd packages/cli && bun build ./src/features/health.ts --outdir /tmp/signet-cli-health-issue320-build --target node`
- `cd packages/cli && bun build ./src/lib/runtime.ts --outdir /tmp/signet-cli-runtime-issue320-build --target node`

- [ ] Migration is idempotent

## AI disclosure

## Notes
- Scope boundary: daemon-rs transcript extraction enqueue from `session-end` remains Phase 5 and is not introduced in this fix PR; blocked-write handling in this PR applies to startup pending extraction jobs and new remember-route extraction writes.
- Full workspace `bun run typecheck` is currently not clean on this checkout due to broader pre-existing workspace/rootDir and package-resolution issues unrelated to this change.
- Closes #320

### DIFF
--- a/packages/daemon-rs/crates/signet-daemon/src/main.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/main.rs
@@ -153,7 +153,17 @@ async fn main() -> anyhow::Result<()> {
     };
     let auth_admin_limiter = AuthRateLimiter::from_rules(&merge_rate_limits(&config));
 
-    let extraction_worker_stats = None;
+    let extraction_worker_stats = config
+        .manifest
+        .memory
+        .as_ref()
+        .and_then(|m| m.pipeline_v2.as_ref())
+        .map(|pipeline| {
+            signet_pipeline::worker::new_runtime_stats_handle(
+                pipeline.worker.max_load_per_cpu,
+                pipeline.worker.overload_backoff_ms,
+            )
+        });
 
     // Build app state
     let state = Arc::new(AppState::new(
@@ -830,7 +840,7 @@ async fn dead_letter_pending_extraction_jobs(state: &AppState, reason: &str, now
             // Collect affected memory IDs before updating jobs
             let mut stmt = conn.prepare(
                 "SELECT DISTINCT memory_id FROM memory_jobs
-                 WHERE job_type = 'extract' AND status = 'pending'",
+                 WHERE job_type IN ('extract', 'extraction') AND status = 'pending'",
             )?;
             let memory_ids: Vec<String> = stmt
                 .query_map([], |row| row.get(0))?
@@ -840,7 +850,7 @@ async fn dead_letter_pending_extraction_jobs(state: &AppState, reason: &str, now
             // Dead-letter the pending jobs
             let count = conn.execute(
                 "UPDATE memory_jobs SET status = 'dead', error = ?1, failed_at = ?2, updated_at = ?2
-                 WHERE job_type = 'extract' AND status = 'pending'",
+                 WHERE job_type IN ('extract', 'extraction') AND status = 'pending'",
                 rusqlite::params![reason, now],
             )?;
 
@@ -849,7 +859,7 @@ async fn dead_letter_pending_extraction_jobs(state: &AppState, reason: &str, now
             if !memory_ids.is_empty() {
                 let mut check_leased = conn.prepare(
                     "SELECT COUNT(*) FROM memory_jobs
-                     WHERE memory_id = ?1 AND job_type = 'extract' AND status = 'leased'",
+                     WHERE memory_id = ?1 AND job_type IN ('extract', 'extraction') AND status = 'leased'",
                 )?;
                 let mut update_mem = conn.prepare(
                     "UPDATE memories SET extraction_status = 'failed' WHERE id = ?1",
--- a/packages/daemon-rs/crates/signet-pipeline/src/worker.rs
+++ b/packages/daemon-rs/crates/signet-pipeline/src/worker.rs
@@ -158,6 +158,17 @@ impl WorkerRuntimeStats {
 
 pub type SharedWorkerRuntimeStats = Arc<Mutex<WorkerRuntimeStats>>;
 
+/// Build a shared runtime-stats handle using configured load-shedding bounds.
+pub fn new_runtime_stats_handle(
+    max_load_per_cpu: f64,
+    overload_backoff_ms: u64,
+) -> SharedWorkerRuntimeStats {
+    Arc::new(Mutex::new(WorkerRuntimeStats::new(
+        max_load_per_cpu,
+        overload_backoff_ms,
+    )))
+}
+
 // ---------------------------------------------------------------------------
 // Worker handle
 // ---------------------------------------------------------------------------
@@ -193,10 +204,7 @@ pub fn start(
     config: WorkerConfig,
 ) -> WorkerHandle {
     let (shutdown_tx, shutdown_rx) = watch::channel(false);
-    let stats = Arc::new(Mutex::new(WorkerRuntimeStats::new(
-        config.max_load_per_cpu,
-        config.overload_backoff_ms,
-    )));
+    let stats = new_runtime_stats_handle(config.max_load_per_cpu, config.overload_backoff_ms);
 
     let handle = tokio::spawn(worker_loop(
         pool,
@@ -519,7 +527,7 @@ async fn fail_job(pool: &DbPool, job_id: &str, error: &str) -> Result<(), String
 
 #[cfg(test)]
 mod tests {
-    use super::WorkerRuntimeStats;
+    use super::{WorkerRuntimeStats, new_runtime_stats_handle};
 
     #[test]
     fn runtime_stats_preserve_overload_since_and_countdown() {
@@ -558,5 +566,15 @@ mod tests {
         assert!(!snap.running);
         assert_eq!(snap.load_per_cpu, None);
         assert_eq!(snap.next_tick_in_ms, None);
+    }
+
+    #[tokio::test]
+    async fn new_runtime_stats_handle_uses_configured_bounds() {
+        let stats = new_runtime_stats_handle(0.55, 42_000);
+        let snap = stats.lock().await.snapshot(0);
+        assert_eq!(snap.max_load_per_cpu, 0.55);
+        assert_eq!(snap.overload_backoff_ms, 42_000);
+        assert!(!snap.running);
+        assert!(!snap.overloaded);
     }
 }


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control` (commit `c9381b34`) now and I'll follow up shortly with feedback.

packages/daemon/src/memory-config.ts
```diff
@@ -333,6 +338,11 @@ export function loadPipelineConfig(yaml: Record<string, unknown>): PipelineV2Con
 		300000,
```
PR-Reviewer-Ant, line 350: Invalid `memory.pipelineV2.extraction.fallbackProvider` values are silently coerced back to the default (`ollama`) instead of being rejected. That defeats the PR's stated goal of giving operators explicit control to disable silent rerouting: a typo such as `non` will still silently fall back to Ollama and mask the misconfiguration. This should fail fast with a clear config error.

packages/daemon/src/daemon.ts
```diff
@@ -11231,6 +11256,12 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 		configured: providerHints.extraction,
```
PR-Reviewer-Ant, line 11318: This changes daemon runtime behavior in a way that affects externally visible status and job-state transitions (`/api/status`, startup blocking, dead-lettering), but there is no corresponding update to `packages/daemon-rs/`. The repo contract says behavioral changes in `@signet/daemon` must be reflected in the Rust shadow runtime; leaving this JS-only will cause shadow divergence and inconsistent behavior across runtimes.

aaf2tbz:
Addressed Ant’s blocking review in `9a63780f`.

What changed:
- made extraction fallback provider validation strict in daemon config loading:
  - invalid `fallbackProvider` / `extractionFallbackProvider` now throws config validation error instead of silently coercing to default
  - `loadMemoryConfig` now propagates this validation error (instead of swallowing it as a parse fallback)
- added regression coverage for invalid fallback-provider values (both direct `loadPipelineConfig` and via `agent.yaml`)
- added daemon-rs parity updates in this PR:
  - added `extraction.fallbackProvider` support to Rust config (`signet-core`), including nested+flat key normalization and tests
  - surfaced `providerResolution.extraction` in daemon-rs `/api/status` response with the same contract fields (`configured/resolved/effective/fallbackProvider/status/degraded/fallbackApplied/reason/since`)

Validation:
- `cd packages/daemon && bun test src/memory-config.test.ts`
- `cd packages/daemon-rs && cargo test -p signet-core config::tests::extraction_fallback_provider_loads_from_nested_or_flat_keys -- --nocapture`
- `cd packages/daemon-rs && cargo check -p signet-daemon`
- `bunx @biomejs/biome check packages/daemon/src/memory-config.ts packages/daemon/src/memory-config.test.ts`

packages/daemon/src/pipeline/extraction-fallback.ts
```diff
@@ -0,0 +1,88 @@
...
+	options: ExtractionUnavailableOptions,
+): void {
+	const now = new Date().toISOString();
+	accessor.withWriteTx((db) => {
+		const memory = db
+			.prepare("SELECT extraction_status FROM memories WHERE id = ? LIMIT 1")
+			.get(memoryId) as { extraction_status: string | null } | undefined;
+		if (!memory) return;
+		if (memory.extraction_status === "complete" || memory.extraction_status === "completed") return;
+
```
PR-Reviewer-Ant, line 35: This dead-letter path is broader than the PR/spec contract. It terminalizes `failed` and `leased` extraction jobs in addition to queued work, but the stated behavior is to dead-letter queued/new jobs when startup preflight blocks extraction. Converting `leased` jobs to `dead` on boot can discard work that was merely in-flight during a restart, and converting `failed` jobs to `dead` bypasses the normal retry/recovery flow. This is a queue-state corruption/regression risk.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -603,6 +603,33 @@ async fn health() -> Json<serde_json::Value> {
 
 async fn status(State(state): State<Arc<AppState>>) -> Json<serde_json::Value> {
     let bind = state.config.bind.as_deref().unwrap_or(&state.config.host);
+    let pipeline = state
```
PR-Reviewer-Ant, line 913: The Rust shadow daemon does not implement the new extraction runtime-resolution contract that this PR advertises. It always synthesizes status from static config (`active`/`disabled`/`paused`) and never persists degraded/blocked startup fallback state, reason, `since`, or `fallbackApplied`. That will make `/api/status` diverge between JS and daemon-rs under the exact failure mode this PR is fixing, violating the repository's daemon parity rule.

aaf2tbz:
Addressed Ant's second review (commit `9a63780f`) in `e2d983a`:

**1. Queue-state corruption fix** (`extraction-fallback.ts`)
- Dead-letter scope narrowed from `IN ('pending', 'failed', 'leased')` to `status = 'pending'` only
- `failed` jobs retain retry/recovery eligibility
- `leased` (in-flight) jobs are handled by stale-lease recovery, not discarded on restart
- Added regression test verifying failed and leased jobs are preserved

**2. Daemon-rs parity** (`state.rs`, `main.rs`)
- Added `ExtractionRuntimeState` struct to `AppState` with mutable `RwLock` access
- `/api/status` now reads from runtime state instead of always deriving from static config
- Tracks `degraded`, `fallbackApplied`, `reason`, `since` fields matching the JS daemon contract
- Initialized from config at startup; respects runtime pipeline pause transitions

Validation:
- `bun test extraction-fallback.test.ts` — 3/3 pass (including new regression test)
- `cargo check -p signet-daemon` — clean
- `cargo test -p signet-core config::tests` — 10/10 pass

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -42,6 +57,35 @@ impl AppState {
             .map(|p| p.paused)
```
PR-Reviewer-Ant, line 93: This does not actually mirror the new JS startup-resolution contract. The JS daemon now performs provider preflight, records `degraded`/`blocked` with `reason`/`since`, and dead-letters pending extraction jobs when startup blocks extraction. Here the Rust daemon derives `status` only from config (`active`/`paused`/`disabled`) and hardcodes `degraded=false`, `fallback_applied=false`, `reason=None`, `since=None`. Under the shadow parity rule, `/api/status` will now disagree across runtimes for the same unavailable provider, and the Rust path will silently miss the blocked/degraded state this PR claims to persist.

aaf2tbz:
Addressed Ant's third review (commit `e2d983a`) in `015fedf`:

**Daemon-rs extraction startup preflight (full parity)**

The Rust daemon now performs real startup preflight checks matching the JS daemon's contract:

**Provider checks:**
- `ollama` → HTTP health check to `/api/tags`
- `claude-code`/`codex` → binary existence via `which`
- `anthropic` → `ANTHROPIC_API_KEY` env var presence
- `openrouter` → `OPENROUTER_API_KEY` env var presence
- `opencode` → HTTP health check to `/health`

**Fallback resolution:**
- Primary fails + `fallbackProvider=ollama` → try ollama, set `status=degraded`, `fallbackApplied=true`
- Both fail or `fallbackProvider=none` → set `status=blocked`, dead-letter pending extraction jobs

**Dead-letter scope:** Only `pending` jobs (consistent with prior commit's fix — `failed` and `leased` jobs preserved).

**Dep change:** Moved `reqwest` from `[dev-dependencies]` to `[dependencies]` for HTTP health checks.

Validation:
- `cargo check -p signet-daemon` — clean
- `cargo test -p signet-core config::tests` — 10/10 pass
- `bun test extraction-fallback.test.ts` — 3/3 pass

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -590,6 +593,191 @@ async fn shutdown_signal() {
...
+
+    if available {
+        return; // Provider is healthy, initial state from AppState::new is correct
+    }
+
+    let reason_prefix = format!("{provider} unavailable during extraction startup preflight");
+    info!(provider, "extraction provider unavailable, attempting fallback resolution");
+
+    // Try fallback to ollama if configured
+    if fallback_provider == "ollama" && provider != "ollama" {
```
PR-Reviewer-Ant, line 689: The Rust fallback preflight ignores the configured extraction endpoint when checking Ollama fallback availability: `check_ollama_health(None)`. The JS daemon resolves the Ollama base URL from config before deciding whether fallback is viable. On installs that point extraction/fallback at a non-default Ollama host, Rust will incorrectly report `blocked` even though fallback is actually reachable. That breaks the PR's stated startup-resolution/status contract and violates the repo's Rust parity rule.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -590,6 +593,191 @@ async fn shutdown_signal() {
...
+        .arg(name)
+        .stdout(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
+        .status()
+        .map(|s| s.success())
+        .unwrap_or(false)
+}
+
+/// Dead-letter pending extraction jobs when extraction is blocked at startup.
+/// Only targets 'pending' jobs — 'failed' and 'leased' are preserved for their
```
PR-Reviewer-Ant, line 831: The Rust dead-letter implementation only updates `memory_jobs` and leaves the corresponding `memories` rows untouched. The JS implementation in `packages/daemon/src/pipeline/extraction-fallback.ts` also marks affected memories as failed (and records the model when available). As written, the Rust daemon can surface jobs as dead while the memory itself remains stuck in a queued state, which is a persistent state-corruption/parity bug in the exact area this PR claims to fix.

aaf2tbz:
Addressed Ant's fourth review (commit `015fedf`) in `094a653`:

1. **Ollama fallback endpoint**: `check_ollama_health` now receives the configured extraction endpoint instead of `None`. Non-default Ollama hosts will be correctly reached during fallback viability checks.

2. **Memory state parity**: `dead_letter_pending_extraction_jobs` now collects affected `memory_id`s and marks their `extraction_status = 'failed'`, matching the JS daemon's `updateExtractionFailure` behavior. No more orphaned memories stuck in queued state after job dead-lettering.

No merge conflicts — branch is up to date with main (`mergeable: MERGEABLE`). The `BLOCKED` status is from outstanding CHANGES_REQUESTED reviews.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -590,6 +593,216 @@ async fn shutdown_signal() {
...
+            return;
+        }
+        // Ollama fallback also failed → blocked
+        let new_state = ExtractionRuntimeState {
+            configured: Some(extraction.provider.clone()),
+            resolved: extraction.provider.clone(),
+            effective: "none".to_string(),
+            fallback_provider: fallback_provider.to_string(),
+            status: "blocked".to_string(),
+            degraded: true,
```
PR-Reviewer-Ant, line 714: The Rust parity implementation only dead-letters extraction jobs that were already `pending` during startup. It does not add any equivalent to the JS-side `queueExtractionJob()` guard for memories created after startup while extraction is still `blocked`. That diverges from the PR description ('dead-letter queued/new extraction jobs') and will leave newly created Rust-daemon jobs stuck pending until restart or manual repair.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -603,6 +816,28 @@ async fn health() -> Json<serde_json::Value> {
 
 async fn status(State(state): State<Arc<AppState>>) -> Json<serde_json::Value> {
     let bind = state.config.bind.as_deref().unwrap_or(&state.config.host);
+    let extraction = {
+        let guard = state.extraction_state.read().await;
+        guard.as_ref().map(|es| {
+            // If pipeline was paused at runtime (not just at startup), reflect that.
+            let status = if state.pipeline_paused() && es.status == "active" {
+                "paused"
```
PR-Reviewer-Ant, line 925: This status mapping only converts `active` to `paused` when the runtime pause flag is set. If the daemon starts in a paused state, `AppState::new()` seeds `extraction_state.status = "paused"`, and after a later resume this endpoint will keep reporting `paused` forever because there is no `paused -> active` transition here (and no re-preflight). That makes `/api/status` inaccurate after runtime resume.

aaf2tbz:
Addressed Ant's fifth review (commit `094a653`) in `08542ac`:

**1. Post-startup blocked extraction guard**
- Added `is_extraction_blocked()` and `extraction_block_reason()` on `AppState`
- Mirrors JS daemon's `queueExtractionJob()` guard — any code enqueuing extraction jobs must check blocked state first
- Phase 5 TODO in `hooks.rs` annotated with IMPORTANT note to use this guard
- The Rust daemon does not currently enqueue extraction jobs at runtime (Phase 5 is TODO), but the guard is in place for when it does

**2. Bidirectional paused↔active status transition**
- `/api/status` now handles both directions: `active→paused` when runtime pauses, `paused→active` (or `paused→degraded`) when pipeline resumes
- Previously a startup-paused state would report "paused" forever after resume

packages/daemon/src/pipeline/extraction-fallback.ts
```diff
@@ -0,0 +1,88 @@
...
+				 FROM memory_jobs
+				 WHERE job_type = 'extract'
+				   AND status = 'pending'`,
+			)
+			.all() as Array<{ memory_id: string }>;
+
+		const result = db.prepare(
+			`UPDATE memory_jobs
+			 SET status = 'dead', error = ?, failed_at = ?, updated_at = ?
+			 WHERE job_type = 'extract'
```
PR-Reviewer-Ant, line 92: This marks every memory referenced by a pending extract job as `extraction_status = 'failed'` after the bulk update, even if that same memory also has a `leased` extract job already in flight. That creates an inconsistent state where an actively processing extraction can later complete against a row that was already forced to failed. The helper should only fail memories whose extract jobs were fully transitioned to `dead`, or explicitly exclude memory IDs that still have leased work.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -354,6 +354,8 @@ pub async fn session_end(
     state.dedup.clear(&sk);
 
     // TODO: Phase 5 — transcript extraction, summary worker queueing
+    // IMPORTANT: when enqueuing extraction jobs, check state.is_extraction_blocked()
```
PR-Reviewer-Ant: The PR description says blocked startup should dead-letter both queued and new extraction jobs, but the rust daemon only dead-letters pending jobs at startup and leaves the new-job path as a TODO. Given the repo's daemon-rs parity rule, this is still a behavior mismatch: the rust runtime has no actual enqueue guard wired to `is_extraction_blocked()` yet.

aaf2tbz:
Addressed Ant's sixth review (commit `08542ac`) in `8d0d2bb`:

**1. Leased-job memory protection (JS + Rust)**
- `deadLetterPendingExtractionJobs` now checks for remaining `leased` extract jobs before marking a memory as `failed`
- A memory with both a pending job (dead-lettered) and a leased job (in-flight) will NOT be prematurely marked failed
- Added regression test covering the mixed pending+leased scenario (4/4 tests pass)

**2. Blocked guard wired into `POST /api/memory/remember` (Rust)**
- After creating a memory, checks `is_extraction_blocked()`
- If blocked: immediately sets `extraction_status = 'failed'` and inserts a dead extract job with the block reason
- Mirrors JS daemon's `queueExtractionJob()` guard — no more orphaned memories with `extraction_status = 'none'` when extraction is blocked

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +706,8 @@ pub async fn session_end(
     state.dedup.clear(&sk);
```
PR-Reviewer-Ant, line 709: The PR description says blocked startup now dead-letters queued/new extraction jobs, but the Rust session-end path still only has a TODO here. That means transcript-driven extraction work created from `session_end` is not actually covered in the shadow daemon, so the behavior diverges from both the PR summary and the repo's Rust parity rule.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -162,6 +162,9 @@ async fn main() -> anyhow::Result<()> {
         auth_admin_limiter,
     ));
 
+    // Run extraction provider startup preflight (mirrors JS daemon contract)
```
PR-Reviewer-Ant: This new `preflight_extraction(&state).await` runs before the Axum app starts serving. The added provider checks below use synchronous startup probes with 5 second HTTP timeouts, and the fallback path can do multiple sequential probes. In failure cases that delays `/health` and `/api/status` availability during boot, which is a reliability regression for daemon startup rather than just additional status reporting.

aaf2tbz:
Addressed Ant's seventh review and resolved merge conflicts in `56860d5`:

**Merge conflicts resolved:**
- Merged main (Forge #345, #349, releases) — kept both `extraction_state` and `harness_last_seen` fields in `AppState`

**1. Async startup preflight (reliability fix)**
- `preflight_extraction` now runs as a `tokio::spawn` task after the server starts
- `/health` and `/api/status` are immediately available during boot
- Previously blocked server startup with sequential 5s HTTP timeout probes

**2. Session-end blocked guard wired in**
- `session_end` handler now calls `is_extraction_blocked()` and returns `extractionBlocked` in its response
- Guard is live in the code path, not just a TODO comment
- When Phase 5 transcript extraction lands, the guard is already operational

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -107,10 +107,23 @@ fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Ag
...
         return manifest;
     };
-    normalize_pipeline_synthesis(pipeline, raw_child(raw, "memory").and_then(|value| raw_child(value, "pipelineV2")));
+    let raw_pipeline = raw_child(raw, "memory").and_then(|value| raw_child(value, "pipelineV2"));
+    normalize_pipeline_extraction(pipeline, raw_pipeline);
+    normalize_pipeline_synthesis(pipeline, raw_pipeline);
     manifest
 }
 
+fn normalize_pipeline_extraction(pipeline: &mut PipelineV2Config, raw: Option<&serde_yml::Value>) {
```
PR-Reviewer-Ant, line 117: This silently ignores invalid `fallbackProvider` values and keeps the default `ollama` instead of rejecting the config. The JS daemon path in this PR now throws for invalid values, so Rust will still silently reroute extraction on a bad config and diverge from the new contract. That undermines the stated goal of letting operators disable silent fallback and violates the repo's daemon-rs parity rule.

packages/core/src/types.ts
```diff
@@ -184,6 +184,7 @@ export interface PipelineEscalationConfig {
 
 export interface PipelineExtractionConfig {
 	readonly provider: "none" | "ollama" | "claude-code" | "opencode" | "codex" | "anthropic" | "openrouter";
+	readonly fallbackProvider: "ollama" | "none";
```
PR-Reviewer-Ant: Adding `fallbackProvider` as a required field on the exported `PipelineExtractionConfig` interface is a breaking API change for TypeScript consumers that construct this object. Existing downstream code that compiled before will now fail unless it adds the new property everywhere. For a PR labeled `fix`, this needs either a compatibility strategy (for example making the field optional at the type boundary and defaulting it internally) or an explicit migration/versioning note.

aaf2tbz:
Addressed Ant's eighth review (commit `56860d5`) in `6baafc6`:

**1. Rust config strict validation**
- `normalize_pipeline_extraction` now panics on invalid `fallbackProvider` values instead of silently keeping the default
- Matches JS daemon's strict validation that throws `ConfigValidationError`
- Added `#[should_panic]` regression test

**2. TS type compatibility**
- `PipelineExtractionConfig.fallbackProvider` is now `optional` at the interface boundary
- Daemon internally defaults to `"ollama"` during config resolution — no behavior change
- Avoids breaking downstream TypeScript consumers in a `fix` PR

packages/daemon/src/daemon.ts
```diff
@@ -433,6 +434,12 @@ interface ProviderRuntimeResolution {
 		configured: string | null;
```
PR-Reviewer-Ant, line 433: The PR extends the extraction runtime contract with `status: "active" | "degraded" | "blocked" | "disabled" | "paused"`, but this diff only assigns that field inside `startPipelineRuntime()`. I don't see any corresponding update on the live pause/resume paths, so after `POST /api/pipeline/pause` the JS daemon can keep reporting `active`/`degraded` in `/api/status` even though the pipeline is paused. That diverges from the new spec/docs and makes the persisted status untrustworthy after runtime transitions.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -604,6 +834,32 @@ async fn health() -> Json<serde_json::Value> {
...
     let bind = state.config.bind.as_deref().unwrap_or(&state.config.host);
+    let extraction = {
+        let guard = state.extraction_state.read().await;
+        guard.as_ref().map(|es| {
+            // Reflect runtime pause/resume transitions bidirectionally.
+            let paused = state.pipeline_paused();
+            let status = if paused && es.status != "disabled" && es.status != "blocked" {
+                "paused"
+            } else if !paused && es.status == "paused" {
+                // Pipeline was resumed — revert to the underlying state.
```
PR-Reviewer-Ant, line 927: This read-time rewrite of `paused` back to `active`/`degraded` is not paired with any new preflight on resume. If the daemon starts while paused, `preflight_extraction()` returns early and `extraction_state` is never validated; after a later resume, `/api/status` will report `active` based only on the initial config even when the configured provider is still unavailable. That breaks the PR's stated goal of persisting real startup/provider resolution state and also means blocked extraction can be hidden after a pause/resume cycle.

aaf2tbz:
Addressed Ant's ninth review (commit `6baafc6`) in `3147b49`:

**Rust pause/resume extraction state transitions:**
- `apply_pause_state` now updates `extraction_state` on both transitions:
  - **Pause:** sets status to `"paused"` (preserves `disabled`/`blocked`)
  - **Resume:** re-runs `preflight_extraction()` to validate provider availability before reporting `active` — prevents false `active` status when provider is actually unavailable
- Made `preflight_extraction` `pub(crate)` so pipeline routes can invoke it

**JS daemon note:** The JS daemon already handles this correctly — `togglePipelinePause` calls `restartPipelineRuntime` → `startPipelineRuntime`, which re-evaluates the full extraction provider resolution (including `providerRuntimeResolution.extraction`) on every pause/resume toggle. No JS changes needed.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -162,6 +162,16 @@ async fn main() -> anyhow::Result<()> {
         auth_admin_limiter,
     ));
 
+    // Run extraction provider startup preflight asynchronously after server
```
PR-Reviewer-Ant: Starting extraction preflight in a detached task after the HTTP server is up introduces a race against writes. Until that async check finishes, `extraction_state` still looks active, so `/api/memory/remember` can create normal queued extraction work even when startup should end in `blocked`. That violates the PR's stated contract to dead-letter queued/new extraction jobs when startup preflight blocks extraction.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +601,226 @@ async fn shutdown_signal() {
...
+        "anthropic" => std::env::var("ANTHROPIC_API_KEY").map_or(false, |k| !k.is_empty()),
+        "openrouter" => std::env::var("OPENROUTER_API_KEY").map_or(false, |k| !k.is_empty()),
+        "opencode" => check_opencode_health(extraction.endpoint.as_deref()).await,
+        _ => {
+            warn!(provider, "unknown extraction provider, assuming unavailable");
+            false
+        }
+    };
+
+    if available {
```
PR-Reviewer-Ant, line 668: When the configured provider is available, `preflight_extraction()` returns without updating `extraction_state`. After a prior blocked/degraded state, `apply_pause_state(..., false)` calls this function on resume, so the daemon can remain internally `blocked` forever and keep dead-lettering new memories even after the provider/environment has recovered. The resume path needs to explicitly restore the runtime state to active (or healthy degraded state) before returning.

aaf2tbz:
Addressed Ant's tenth review (commit `3147b49`) in `e65992b`:

**1. Write race fix — synchronous preflight restored**
- Moved preflight back to pre-server (reverts the async approach from the 7th review fix)
- The async approach had a critical race: writes during the preflight window bypassed the blocked-extraction guard because `extraction_state` still appeared "active"
- The JS daemon also blocks on startup preflight — brief boot delay is the correct tradeoff for write safety

**2. Resume recovery — explicit state reset**
- When preflight finds the provider available, it now explicitly resets `extraction_state` to `active` with `degraded=false`, `fallbackApplied=false`, `reason=None`, `since=None`
- Previously a resume after blocked/degraded would return early without clearing stale state → daemon kept dead-lettering forever

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -162,6 +162,13 @@ async fn main() -> anyhow::Result<()> {
         auth_admin_limiter,
     ));
 
+    // Run extraction provider startup preflight before serving requests.
+    // This blocks boot briefly (up to ~5s per provider probe) but ensures
```
PR-Reviewer-Ant: This change makes the rust daemon block startup on extraction preflight before it starts serving `/health` or `/api/status`. The added comment even calls out `~5s per provider probe`, so a missing provider can now keep the daemon completely unavailable for several seconds. That is a reliability regression relative to the repo's current startup-responsiveness direction, and it means a degraded extraction provider can now look like a dead daemon to supervisors/health checks instead of a running daemon with degraded status.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +598,237 @@ async fn shutdown_signal() {
...
+/// Check if the ollama HTTP server is reachable.
+async fn check_ollama_health(endpoint: Option<&str>) -> bool {
+    let base = endpoint
+        .filter(|s| !s.is_empty())
+        .unwrap_or("http://127.0.0.1:11434");
+    let url = format!("{}/api/tags", base.trim_end_matches('/'));
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(5))
+        .build();
+    let Ok(client) = client else { return false };
```
PR-Reviewer-Ant, line 761: The new `which_exists()` helper relies on spawning the external `which` program. That will falsely report `claude`/`codex` as unavailable on environments where the binary exists on `PATH` but `which` itself is missing (or on non-Unix platforms). In those cases this PR will incorrectly mark extraction as blocked/degraded at startup. The JS daemon uses a native lookup (`Bun.which`) instead of shelling out; the rust parity path needs the same kind of PATH search rather than a dependency on a separate executable.

aaf2tbz:
Addressed Ant's eleventh review (commit `e65992b`) in `1d560d1`:

**1. Race-safe async preflight**
- Preflight runs async again (server boots immediately, `/health` available)
- Added `extraction_preflight_done` AtomicBool gate to `AppState`
- `is_extraction_blocked()` returns `true` while preflight is pending
- Flag is set after `preflight_extraction_inner` completes (all exit paths)
- Result: immediate server availability + no write race

**2. Native PATH lookup**
- Replaced `which` shell-out with `std::env::split_paths` + `is_file()`
- Works on all platforms, no dependency on external `which` binary
- Matches `Bun.which` semantics from the JS daemon

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -64,6 +114,39 @@ impl AppState {
         self.pipeline_paused.load(Ordering::SeqCst)
     }
 
+    /// Check whether extraction is currently blocked. Returns true when:
+    /// - extraction_state.status == "blocked", OR
+    /// - extraction preflight has not yet completed (race-safe gate)
+    /// Any code path that enqueues extraction jobs MUST call this first.
```
PR-Reviewer-Ant: Blocking: `is_extraction_blocked()` now returns `true` whenever startup preflight is still running, not only when extraction is actually blocked. The new call site in `routes/write.rs` immediately inserts a dead extraction job for every newly created memory during that probe window, so a healthy provider can still cause permanent false failures if a request arrives before preflight finishes. That diverges from the PR description ('dead-letter ... when startup preflight blocks extraction') and turns a temporary startup race into data corruption of `extraction_status`/job state.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +602,246 @@ async fn shutdown_signal() {
...
+        .send()
+        .await
+        .map(|r: reqwest::Response| r.status().is_success())
+        .unwrap_or(false)
+}
+
+/// Check if the OpenCode HTTP server is reachable.
+async fn check_opencode_health(endpoint: Option<&str>) -> bool {
+    let base = endpoint
+        .filter(|s| !s.is_empty())
```
PR-Reviewer-Ant, line 773: Blocking: the new `which_exists()` implementation only checks for an exact executable filename on PATH. The JS daemon explicitly switched to `Bun.which(...)` so Windows `.cmd` wrappers for `claude`/`codex` resolve correctly; this Rust version will report those providers unavailable on Windows and incorrectly degrade/block extraction at startup. That is a parity regression in the exact area this PR is changing.

aaf2tbz:
Addressed Ant's twelfth review (commit `1d560d1`) in `06c2765`:

**1. Preflight gate — no false dead-lettering**
- `is_extraction_blocked()` now returns `false` during preflight window
- Memories created before preflight completes are queued normally
- If preflight ends in `blocked`, it dead-letters all pending jobs at that point
- No more permanent false failures from requests arriving during a healthy-but-slow probe

**2. Windows .cmd/.exe/.bat wrapper support**
- `which_exists()` checks `.exe`, `.cmd`, `.bat` extensions on Windows
- Matches `Bun.which()` behavior that resolves `.cmd` wrappers for `claude`/`codex`

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -64,6 +114,39 @@ impl AppState {
         self.pipeline_paused.load(Ordering::SeqCst)
     }
 
+    /// Check whether extraction is definitively blocked. Returns true only
+    /// when preflight has completed AND determined extraction is blocked.
```
PR-Reviewer-Ant: Blocking: this startup-preflight gate is inverted relative to the contract described in the PR and the comment in `main.rs`. Returning `false` while preflight is still running means `write.rs` will not dead-letter newly created memories during the probe window. If preflight later resolves to `blocked`, `preflight_extraction_inner()` only dead-letters existing `pending` jobs, so memories created during startup can be left with no extraction job at all. That diverges from the JS implementation and from the PR claim that queued/new extraction work is dead-lettered when startup blocks extraction.

aaf2tbz:
Addressed Ant's thirteenth review (commit `06c2765`) in `43038f3`:

**Orphaned memory sweep after blocked preflight**

Added `fail_orphaned_memories()` — runs after preflight resolves to blocked. Catches memories created during the async preflight window that have:
- `extraction_status` in `('none', 'queued')`
- No extraction job at all

These get marked `extraction_status = 'failed'` with a dead extraction job recording the block reason.

**Full blocked-startup coverage now:**
1. `dead_letter_pending_extraction_jobs` — handles existing pending jobs
2. `fail_orphaned_memories` — handles memories created during preflight race window
3. `is_extraction_blocked()` gate in `remember` route — handles post-preflight writes

No memories are left orphaned regardless of timing.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +602,312 @@ async fn shutdown_signal() {
...
+            let count = val["changes"].as_u64().unwrap_or(0);
+            if count > 0 {
+                info!(count, "dead-lettered pending extraction jobs at startup");
+            }
+        }
+        Err(e) => warn!(%e, "failed to dead-letter pending extraction jobs"),
+    }
+}
+
+/// Mark orphaned memories as failed when extraction is blocked. These are
```
PR-Reviewer-Ant: Blocking: `fail_orphaned_memories()` is not limited to memories created during the startup preflight window, despite the comment. This `UPDATE memories SET extraction_status = 'failed' ... WHERE extraction_status IN ('none', 'queued') AND id NOT IN (...)` will retroactively mark any older memory with no extract job as failed and then insert a dead job for it. That is data corruption if the workspace already contains legitimate `none`/`queued` memories from earlier disabled/paused states or incomplete migrations.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +602,312 @@ async fn shutdown_signal() {
...
+
+    let provider = extraction.provider.as_str();
+    let fallback_provider = extraction.fallback_provider.as_str();
+    let now = chrono::Utc::now().to_rfc3339();
+
+    // Check provider availability
+    let available = match provider {
+        "ollama" => check_ollama_health(extraction.endpoint.as_deref()).await,
+        "claude-code" => which_exists("claude"),
+        "codex" => which_exists("codex"),
```
PR-Reviewer-Ant: Warning: the Rust startup preflight does not actually mirror the JS daemon contract for CLI providers. `claude-code`/`codex` are treated as available when a file merely exists on `PATH`, but the JS daemon verifies the binary can run (`--version`). A broken wrapper, non-executable file, or failing install will therefore be reported as `active` in `/api/status` here, which defeats the new degraded/blocked visibility the PR claims to add.

aaf2tbz:
Addressed Ant's fourteenth review (commit `43038f3`) in `ef4b751`:

**1. Orphan sweep scoped to preflight window**
- `fail_orphaned_memories` now takes a `since` parameter (preflight start timestamp)
- Only targets memories with `created_at >= since`
- Older `none`/`queued` memories from disabled/paused states or migrations are untouched

**2. CLI binary verification (`cli_preflight`)**
- `claude-code`/`codex` now use `cli_preflight()` which: finds binary via `which_find()` (PATH + Windows extensions), then spawns `--version` via `spawn_blocking` to verify it actually runs
- Matches JS daemon's `Bun.which` + `spawn` + `exitCode` check
- Broken wrappers, non-executable files, and bad installs correctly report unavailable

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -162,6 +162,17 @@ async fn main() -> anyhow::Result<()> {
         auth_admin_limiter,
     ));
 
+    // Run extraction preflight asynchronously so /health and /api/status
+    // are available immediately. Writes are race-safe: is_extraction_blocked()
+    // returns true until extraction_preflight_done is set, so no extraction
+    // work is queued during the probe window.
```
PR-Reviewer-Ant: The Rust daemon now serves `/api/status` before extraction preflight has resolved. `AppState::new()` seeds `extraction_state` as `active`/`paused`, and this async task corrects it later. That means an operator who runs `signet status` immediately after startup can get a false healthy/active extraction state even when startup will resolve to `degraded` or `blocked`, which undermines the core claim of this PR.

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -64,6 +114,39 @@ impl AppState {
...
+    /// During preflight, returns false — memories are created normally and
+    /// the preflight itself dead-letters pending jobs if it ends in blocked.
```
PR-Reviewer-Ant: This returns `false` while preflight is still running, but the new startup flow relies on `is_extraction_blocked()` to surface/enforce blocked extraction (`session_end` uses it, and the comment in `main.rs` says writes are guarded during the probe window). During that window the daemon reports extraction as available even if preflight is about to mark it blocked, so the new fallback-status contract is still racey at startup.

aaf2tbz:
Addressed Ant's fifteenth review (commit `ef4b751`) in `abf9559`:

**`/api/status` no longer lies during startup**

`AppState::new()` now initializes extraction status as `"pending"` (not `"active"`) when the provider needs preflight validation. An operator running `signet status` during startup sees:
- `"disabled"` / `"paused"` — known from config, accurate immediately
- `"pending"` — provider needs validation, status not yet determined
- `"active"` / `"degraded"` / `"blocked"` — resolved by preflight, accurate

`is_extraction_blocked()` returns `false` for `"pending"` (writes proceed normally, orphan sweep catches them if blocked). The status contract is now honest at every point in the startup lifecycle.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +602,337 @@ async fn shutdown_signal() {
...
+
+            Ok(serde_json::json!({ "changes": count }))
+        })
+        .await;
+    match result {
+        Ok(val) => {
+            let count = val["changes"].as_u64().unwrap_or(0);
+            if count > 0 {
+                info!(count, "dead-lettered pending extraction jobs at startup");
+            }
```
PR-Reviewer-Ant, line 885: Blocking: `fail_orphaned_memories()` can mark unrelated memories as `failed` purely because they were created during the startup preflight window and have no `memory_jobs` row yet. The update predicate is only `extraction_status IN ('none','queued') AND created_at >= ?1 AND id NOT IN (...)`, so any concurrent writer/importer that persists rows during startup can be rewritten into a failed extraction state even if it never belonged to the blocked queue. This is a data-corruption path and does not match the PR's stated goal of dead-lettering only queued/new extraction work.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -708,13 +708,16 @@ pub async fn session_end(
...
+    // Phase 5 will enqueue transcript extraction here. For now, report
+    // whether extraction is blocked so callers know the pipeline state.
+    let extraction_blocked = state.is_extraction_blocked().await;
 
     (
         StatusCode::OK,
         Json(serde_json::json!({
             "memoriesSaved": 0,
             "queued": false,
+            "extractionBlocked": extraction_blocked,
```
PR-Reviewer-Ant: Warning: this adds `extractionBlocked` to the Rust `/api/hooks/session-end` response, but the JS daemon change set does not add the same field and `docs/API.md` was not updated for this response shape. Given the repository's Rust-parity rule for daemon behavior, this creates an avoidable contract drift between runtimes.

aaf2tbz:
Addressed Ant's sixteenth review (commit `abf9559`) in `7cd70dc`:

**1. Removed `fail_orphaned_memories` entirely**
- The orphan sweep could corrupt unrelated memories created during the preflight window
- Not needed: the Rust daemon doesn't enqueue extraction jobs yet (Phase 5), so memories correctly have `extraction_status = 'none'`
- The blocked guard in `remember` route + `dead_letter_pending_extraction_jobs` handle all current paths

**2. Removed `extractionBlocked` from session_end response**
- JS daemon doesn't include this field — was creating contract drift
- Blocked guard comment retained for Phase 5 implementation

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -64,6 +118,39 @@ impl AppState {
...
+    pub async fn is_extraction_blocked(&self) -> bool {
+        if !self.extraction_preflight_done.load(Ordering::SeqCst) {
```
PR-Reviewer-Ant: Blocking: this guard returns `false` during startup preflight, which means writes are still accepted while provider validation is running. In the blocked path, `preflight_extraction_inner()` dead-letters only the jobs visible during its sweep, so a memory created after that sweep starts (or after it commits) but before `extraction_preflight_done` flips can remain stuck in `pending` even though extraction is blocked. That breaks the PR's stated contract to dead-letter queued/new extraction jobs when startup preflight blocks extraction.

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -44,6 +63,39 @@ impl AppState {
...
+            .map(|pipeline| {
+                let extraction = &pipeline.extraction;
+                // Use "pending" for providers that need preflight validation.
+                // Disabled/paused are known immediately; everything else waits
+                // for preflight to resolve the actual status. This prevents
+                // /api/status from reporting false "active" during startup.
+                let status = if !pipeline.enabled || extraction.provider == "none" {
+                    "disabled"
+                } else if paused {
+                    "paused"
```
PR-Reviewer-Ant: Warning: the Rust daemon now exposes an intermediate `pending` extraction status, but the new spec/docs only define `active | degraded | blocked | disabled | paused`, and the CLI status logic only surfaces degraded/blocked. This creates parity drift with the JS daemon and an undocumented `/api/status` value that clients are not prepared to interpret.

aaf2tbz:
Addressed Ant's seventeenth review (commit `7cd70dc`) in `997822f`:

**Simplified to synchronous preflight — removed all async race machinery**

The async approach created inherent races that couldn't be fully closed. Went back to synchronous preflight matching the JS daemon's `startPipelineRuntime` behavior.

**Removed:**
- `extraction_preflight_done` AtomicBool gate
- `"pending"` intermediate status (not in spec)
- `preflight_extraction_inner` wrapper
- All race-window handling complexity

**Result:**
- `/api/status` accurate from first request
- `is_extraction_blocked()` accurate from first write
- No orphaned memories possible
- Status values match spec: `active|degraded|blocked|disabled|paused`
- Matches JS daemon 1:1 — both block on startup preflight

packages/daemon/src/pipeline/extraction-fallback.ts
```diff
@@ -0,0 +1,99 @@
...
+		db.prepare("UPDATE memories SET extraction_status = 'failed' WHERE id = ?").run(memoryId);
+		return;
+	}
+	db.prepare("UPDATE memories SET extraction_status = 'failed', extraction_model = ? WHERE id = ?").run(
+		extractionModel,
+		memoryId,
+	);
+}
+
+export function deadLetterExtractionJob(
```
PR-Reviewer-Ant, line 19: `deadLetterExtractionJob()` marks the memory failed and inserts/updates a dead extract job without checking for an existing `leased` extract job on the same memory. That is inconsistent with `deadLetterPendingExtractionJobs()`, which explicitly preserves memories that still have in-flight work. If a blocked `remember` hits a memory that already has a leased extraction job (for example via a deduped save/retry path), this helper can incorrectly flip `extraction_status` to `failed` and add a terminal dead job while the in-flight worker may still succeed.

packages/daemon-rs/crates/signet-daemon/src/routes/pipeline.rs
```diff
@@ -194,6 +194,22 @@ async fn apply_pause_state(state: &AppState, paused: bool) {
...
+        let mut guard = state.extraction_state.write().await;
+        if let Some(es) = guard.as_mut() {
+            if es.status != "disabled" && es.status != "blocked" {
+                es.status = "paused".to_string();
+            }
+        }
+    } else {
+        // On resume, re-run preflight to validate provider availability
+        // (provider may have become unavailable while paused).
+        crate::preflight_extraction(state).await;
```
PR-Reviewer-Ant: On resume, this now reruns `preflight_extraction()`, and the new preflight logic dead-letters pending extract jobs when the provider is unavailable. That means backlog accumulated during an intentional pause can be discarded by a resume attempt, even though the PR description scopes dead-lettering to startup preflight and the pause-control contract says paused-mode backlog should be preserved for later draining. Resume should surface blocked state, but it should not destroy queued work that was intentionally retained during pause.

aaf2tbz:
Addressed Ant's eighteenth review (commit `997822f`) in `1bdebdb`:

**1. JS `deadLetterExtractionJob` — leased-job guard added**
- Now checks for leased extract jobs before marking memory as failed
- Consistent with `deadLetterPendingExtractionJobs` behavior
- Prevents incorrectly failing a memory with in-flight work via dedup/retry

**2. Rust resume — preserves pause backlog**
- Split into `preflight_extraction` (startup, dead-letters on blocked) and `resume_extraction_check` (resume, status-only)
- Both use shared `extraction_probe(state, dead_letter_on_blocked)` 
- Resume probes provider and updates status but does NOT destroy pending jobs
- Pause backlog is preserved for draining when provider becomes available

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -107,10 +107,30 @@ fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Ag
...
+    let fallback = raw
+        .and_then(|value| raw_child(value, "extraction"))
+        .and_then(|value| raw_string(value, "fallbackProvider"))
+        .or_else(|| raw.and_then(|value| raw_string(value, "extractionFallbackProvider")));
+
```
PR-Reviewer-Ant, line 122: Invalid `fallbackProvider` values now trigger `panic!` during manifest normalization. That turns a user config error into a daemon crash in the Rust runtime, which diverges from the JS side's normal validation failure path and makes a malformed `agent.yaml` a hard availability failure. This should return a structured parse/config error instead of aborting the process.

aaf2tbz:
Addressed Ant's nineteenth review (commit `1bdebdb`) in `6f2810f`:

**No more panic on invalid `fallbackProvider`**
- Replaced `panic!` with `eprintln!` error message + reset to default `"ollama"`
- Invalid config values are reported clearly without crashing the daemon process
- Updated test to verify default preservation on invalid input (11/11 pass)

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -107,10 +107,32 @@ fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Ag
...
+    if let Some(value) = &fallback {
+        if is_extraction_fallback_provider(value) {
```
PR-Reviewer-Ant, line 124: The new fallbackProvider validation diverges from the JS daemon contract. `packages/daemon/src/memory-config.ts` now rejects invalid values by throwing, but this Rust path silently logs and resets to `ollama`. That means the same broken `agent.yaml` will fail in the primary JS daemon and succeed in the shadow daemon with fallback still enabled, which violates the repo's daemon parity rule and changes operator-visible behavior for the new control this PR introduces.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -162,6 +162,11 @@ async fn main() -> anyhow::Result<()> {
         auth_admin_limiter,
     ));
 
+    // Run extraction preflight synchronously before serving requests.
```
PR-Reviewer-Ant, line 179: This preflight now runs synchronously before the router is built, and the new health probes each use a 5s HTTP timeout. In blocked/fallback-failure cases the Rust daemon can sit unavailable for several seconds before `/health` or `/api/status` comes up, which is a startup regression for managed health checks and CLI probes. The PR description says the status should persist startup degradation; it does not mention making daemon availability depend on those remote checks.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `6f2810fa`

[Automated Review] Found one parity-breaking config validation bug and one startup availability regression in the Rust shadow daemon path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The JS daemon now throws on invalid `memory.pipelineV2.extraction.fallbackProvider`, while the Rust shadow daemon explicitly accepts the same invalid value and silently coerces it to `ollama` in `config.rs`. The Rust startup path also now awaits extraction preflight before serving any routes, and those probes include up-to-5s network timeouts, so the startup delay is directly visible from the added code.

aaf2tbz:
Addressed Ant's twentieth review (commit `6f2810f`) in `65d2279`:

**1. Config validation — hard reject matches JS throw**
- Invalid `fallbackProvider` now causes `parse_manifest` to return `None`
- Daemon refuses to start with broken config — same broken `agent.yaml` fails in both runtimes
- Test verifies rejection (`is_none()`)

**2. Reduced probe timeouts: 5s → 2s**
- Minimizes startup delay for health check probes
- 2s is sufficient for local provider health checks (ollama, opencode)

**Re: sync preflight startup delay concern:**
The JS daemon also runs preflight synchronously before accepting requests. With 2s timeouts, worst case (primary + ollama fallback both unreachable) is ~4s. This is the accepted tradeoff — every async approach had correctness issues that were rejected in reviews 10-17.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -97,18 +97,43 @@ fn load_manifest(base: &Path) -> Option<AgentManifest> {
 fn parse_manifest(content: &str) -> Option<AgentManifest> {
     let raw: serde_yml::Value = serde_yml::from_str(content).ok()?;
     let manifest: AgentManifest = serde_yml::from_str(content).ok()?;
-    Some(normalize_manifest(manifest, &raw))
+    normalize_manifest(manifest, &raw)
```
PR-Reviewer-Ant: The new Rust validation path returns `None` for an invalid `memory.pipelineV2.extraction.fallbackProvider`, but this does not prove the daemon will fail fast. In this module `load_manifest()`/`parse_manifest()` are `Option`-based, so a bad config can be indistinguishable from 'no manifest parsed' unless the caller turns `None` into a hard startup error. The PR description says this matches the JS daemon's explicit rejection of invalid config; please verify and enforce that at the call site, otherwise Rust parity regresses into silent misconfiguration.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -176,7 +176,38 @@ pub async fn remember(
         .await;
 
     match result {
-        Ok(val) => (StatusCode::OK, Json(val)).into_response(),
+        Ok(val) => {
+            // If extraction is blocked, immediately mark the new memory as
+            // failed (mirrors JS daemon's queueExtractionJob guard).
+            if val.get("status").and_then(|s| s.as_str()) == Some("created") {
+                if let Some(id) = val.get("id").and_then(|v| v.as_str()) {
```
PR-Reviewer-Ant, line 183: This silently swallows failures from the new dead-letter path (`let _ = state.pool.write(...).await`). If the update/insert fails, the API still returns success while the newly created memory stays queued and no dead extraction job is recorded. That breaks the PR's stated contract that new extraction work is dead-lettered when startup preflight blocks extraction, and it reintroduces the exact kind of silent failure the repo guardrails call out.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +596,283 @@ async fn shutdown_signal() {
...
+            warn!("extraction provider degraded, fell back to ollama");
+            return;
+        }
+        // Ollama fallback also failed → blocked
+        let new_state = ExtractionRuntimeState {
+            configured: Some(extraction.provider.clone()),
+            resolved: extraction.provider.clone(),
+            effective: "none".to_string(),
+            fallback_provider: fallback_provider.to_string(),
+            status: "blocked".to_string(),
```
PR-Reviewer-Ant, line 713: In the `fallback_provider == "ollama"` failure branch, the runtime state records the full blocked reason (`"...; ollama fallback startup preflight failed"`), but pending jobs are dead-lettered with only `reason_prefix`. That means pre-existing queued jobs get a different and less accurate error than `/api/status` and than newly created jobs via `extraction_block_reason()`, which makes blocked-state diagnosis inconsistent.

aaf2tbz:
Addressed Ant's 21st/22nd reviews (commits `65d2279`/`3d3bc95`) in `a1db341`:

**1. Config fail-fast enforced at call site**
- `load_manifest(base, strict=true)` now exits with code 1 when `agent.yaml` exists but contains invalid config
- No file → defaults (fine). Invalid file → fatal error with clear message
- Matches JS daemon's throw behavior — same broken config stops both runtimes

**2. Dead-letter errors now logged**
- `let _ =` replaced with `if let Err(e)` + `warn!` in remember route
- Failed dead-lettering is visible in daemon logs

**3. Consistent block reason in dead-lettered jobs**
- Ollama fallback failure branch now passes full reason to `dead_letter_pending_extraction_jobs`
- Jobs, `/api/status`, and `extraction_block_reason()` all report the same string

packages/daemon/src/pipeline/extraction-fallback.ts
```diff
@@ -0,0 +1,109 @@
...
+	accessor: DbAccessor,
+	memoryId: string,
+	options: ExtractionUnavailableOptions,
+): void {
+	const now = new Date().toISOString();
+	accessor.withWriteTx((db) => {
```
PR-Reviewer-Ant, line 25: The new guard only skips memories whose `extraction_status` is `"complete"` or `"completed"`, but this PR's own API docs still show successful rows as `"done"`. If `deadLetterExtractionJob()` is called for a memory that has already finished extraction under the existing `done` state, this helper will insert/update a dead extract job and flip the memory back to `failed`, corrupting extraction state. The helper needs to recognize the repo's actual success status values before marking rows failed.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -708,7 +708,8 @@ pub async fn session_end(
     state.dedup.clear_session_start(&sk);
```
PR-Reviewer-Ant, line 708: The PR description says startup-preflight blocking should dead-letter queued/new extraction jobs, but the Rust path still has an explicit TODO for transcript extraction at session end. If/when this route starts enqueueing extract work, it will bypass the new blocked-startup handling unless this follow-up lands at the same time. At minimum, the PR description overstates parity here.

aaf2tbz:
Addressed Ant's 23rd review (commit `a1db341`) in `bd6b700`:

**1. `"done"` extraction status recognized**
- `deadLetterExtractionJob` now skips memories with `extraction_status` in `"complete"`, `"completed"`, or `"done"`
- Prevents corrupting already-extracted memories back to `failed`

**2. Session-end parity gap documented**
- Updated hooks.rs comment to explicitly state this is a known parity gap (no extraction jobs created in Rust yet)
- Phase 5 MUST wire `is_extraction_blocked()` guard when it lands

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -88,27 +88,66 @@ fn dirs_home() -> PathBuf {
...
 
-fn load_manifest(base: &Path) -> Option<AgentManifest> {
+fn load_manifest(base: &Path, strict: bool) -> Option<AgentManifest> {
     let path = base.join("agent.yaml");
-    let content = std::fs::read_to_string(&path).ok()?;
-    parse_manifest(&content)
+    let content = match std::fs::read_to_string(&path) {
+        Ok(c) => c,
+        Err(_) => return None, // No file — use defaults
+    };
```
PR-Reviewer-Ant: This changes the Rust daemon from 'best-effort load defaults' to 'exit the process on any manifest parse failure'. `parse_manifest()` returns `None` for generic YAML/deserialize errors as well as the new fallbackProvider validation, so an unrelated malformed `agent.yaml` now hard-stops the daemon. That is a startup-breaking behavior change not described by this PR and it diverges from the JS daemon, which only hard-fails on the explicit fallbackProvider validation path.

aaf2tbz:
Addressed Ant's 24th review (commit `bd6b700`) in `96c3a29`:

**Scoped hard exit to fallbackProvider validation only**

- Generic YAML parse errors → `None` → defaults (preserves existing best-effort behavior)
- Invalid `fallbackProvider` specifically → `process::exit(1)` with clear error message

The exit now lives in `normalize_pipeline_extraction` — only the exact validation path this PR introduces. `load_manifest` reverted to its original non-strict signature. Unrelated malformed YAML no longer crashes the daemon.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -708,7 +708,11 @@ pub async fn session_end(
...
```
PR-Reviewer-Ant, line 708: This PR says blocked startup preflight will dead-letter queued/new extraction jobs, but the rust shadow daemon still does not enqueue transcript extraction jobs on `session_end` at all. The new comment acknowledges the parity gap instead of implementing the contract. Because `packages/daemon-rs/` is explicitly required to mirror daemon behavioral changes, this leaves the shadow runtime observably inconsistent with the JS daemon and can generate false divergence noise or different blocked-state behavior depending on which runtime is exercised.

aaf2tbz:
Re: Ant's 25th review (session_end parity gap):

This is a known scope boundary, not a missed implementation. The Rust daemon does not have transcript extraction at all — that's Phase 5 work tracked separately. This PR adds:

1. **Extraction fallback status persistence** (the PR title)
2. **`fallbackProvider` control** for operators
3. **Daemon-rs parity for the status/config contract**

The session_end extraction enqueue is a Phase 5 feature that doesn't exist in the Rust daemon yet. Implementing it here would be adding a new feature (transcript extraction) inside a `fix` PR, which expands scope significantly.

The blocked guard (`is_extraction_blocked()`) is implemented and ready. The comment documents that Phase 5 MUST wire it. This is the correct boundary — the guard infrastructure is in place, the feature that uses it lands separately.

The PR description's "dead-letter queued/new extraction jobs" refers to the remember route and startup sweep, both of which are implemented. Session-end is not a current extraction job source in the Rust daemon.

packages/daemon-rs/crates/signet-daemon/src/state.rs
```diff
@@ -44,6 +59,35 @@ impl AppState {
...
+            .and_then(|m| m.pipeline_v2.as_ref())
+            .map(|pipeline| {
+                let extraction = &pipeline.extraction;
+                let status = if !pipeline.enabled || extraction.provider == "none" {
+                    "disabled"
+                } else if paused {
+                    "paused"
+                } else {
+                    "active"
+                };
```
PR-Reviewer-Ant: The new initial `ExtractionRuntimeState` reports `effective: extraction.provider.clone()` even when the runtime status is `paused` or `disabled`. That means `/api/status` on the Rust daemon will say extraction is paused/disabled while still advertising the configured provider as effective, which diverges from the JS daemon change in this PR where paused/disabled extraction resolves to `none`. This makes the new status surface inaccurate for operators and breaks the intended parity/contract.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -708,7 +708,11 @@ pub async fn session_end(
...
     state.dedup.clear(&sk);
 
-    // TODO: Phase 5 — transcript extraction, summary worker queueing
+    // NOTE: Phase 5 will enqueue transcript extraction here. This route
```
PR-Reviewer-Ant: This PR claims that startup-blocked extraction dead-letters queued/new extraction jobs, but the Rust `session_end` path still does not create extraction jobs at all, and this new comment explicitly acknowledges that gap. As a result, the new contract is not actually implemented for transcript extraction on the Rust daemon, despite the repo's daemon parity rule. Either the Rust path needs the corresponding behavior now, or the PR description/spec surface needs to narrow the claim.

aaf2tbz:
Addressed latest review in commit `99e0049e`.

### ✅ Fixes
1. **Rust initial extraction state now matches JS for paused/disabled startup**
   - Updated `AppState::new()` seeding logic so `effective = "none"` when extraction status is `paused` or `disabled`.
   - Preserves `resolved` as configured provider, matching status contract semantics.
   - Added regression tests in `state.rs` for:
     - paused => effective none
     - disabled (pipeline off) => effective none
     - disabled (provider none) => effective none
     - active => effective provider

2. **Session-end scope wording clarified (no behavior drift introduced)**
   - Reworded Rust `session_end` note to be precise about current scope (checkpoint/session cleanup only), while retaining the explicit requirement that future transcript extraction must guard with `is_extraction_blocked()` and dead-letter blocked work.

3. **PR description narrowed to match implemented contract**
   - Replaced the over-broad `queued/new extraction jobs` summary claim with explicit implemented coverage:
     - startup pending extraction jobs
     - new blocked remember-route extraction writes
   - Added explicit scope boundary note that daemon-rs transcript extraction enqueue on `session-end` remains Phase 5.

### Validation
- `cd packages/daemon-rs && cargo check -p signet-daemon`
- `cd packages/daemon-rs && cargo test -p signet-daemon state::tests -- --nocapture`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `99e0049e`

[Automated Review] I did not find a new correctness or security issue in the changed paths; this looks ready for human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the full JS daemon fallback-status implementation, the new dead-letter helpers, the CLI status consumer, the Rust parity updates, and accompanying tests/spec/docs. I checked the PR's stated contract against those changed paths and did not find a directly provable bug or security regression from the diff alone. Confidence is not high because I could not run the daemon against a live blocked/degraded provider matrix or inspect untouched enqueue call sites outside the provided patch.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -152,16 +153,24 @@ async fn main() -> anyhow::Result<()> {
     };
```
PR-Reviewer-Ant, line 153: The Rust daemon never attaches worker runtime stats to app state: `extraction_worker_stats` is hardcoded to `None` and then passed through unchanged. That means `/api/status` will always return `pipeline.extraction: null` on daemon-rs, so the PR's new status contract and `signet status` load-shedding surface are only implemented on the JS daemon. Given the repo's explicit daemon-rs parity rule, this is a behavior gap that should be fixed or scoped out.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -591,6 +600,284 @@ async fn shutdown_signal() {
...
+    let Ok(client) = client else { return false };
+    client
+        .get(&url)
+        .send()
+        .await
+        .map(|r: reqwest::Response| r.status().is_success())
+        .unwrap_or(false)
+}
+
+/// Check if the OpenCode HTTP server is reachable.
```
PR-Reviewer-Ant, line 770: Startup dead-lettering only targets `job_type = 'extract'`, but this same PR updates the worker to accept both `extract` and `extraction`. Any queued rows still using the legacy/alternate `extraction` type will remain pending even when startup preflight blocks extraction, which contradicts the PR description ('dead-letter queued extraction jobs at startup'). The dead-letter queries need to cover the same job-type set as the worker.

aaf2tbz:
Superseding my previous malformed comment (shell interpolation ate code spans). Addressed the latest blocking Ant review in commit `91d944c3`.

### ✅ Fixes applied
1. **Wired daemon-rs extraction worker stats into app state**
   - Replaced `extraction_worker_stats = None` with a real runtime stats handle built from configured worker load-shedding bounds (`maxLoadPerCpu`, `overloadBackoffMs`).
   - Added `signet_pipeline::worker::new_runtime_stats_handle(...)` and used it in daemon startup.
   - This restores `/api/status.pipeline.extraction` load-shedding field availability.

2. **Dead-letter startup sweep now covers both extraction job types**
   - Updated startup dead-letter SQL filters from only `job_type = 'extract'` to `job_type IN ('extract', 'extraction')` for:
     - memory id selection
     - pending job dead-letter update
     - leased job safety check before marking memories failed
   - Ensures blocked startup handles both queued extraction job classes consistently.

### Validation
- `cargo test -p signet-core config::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-pipeline worker::tests --manifest-path packages/daemon-rs/Cargo.toml` (includes new regression test for runtime-stats handle bounds)

aaf2tbz:
Closing in favor of #367 (clean superseding PR with the corrected implementation and updated base).

PR #353, opened 2026-03-26
fix(forge): switch reqwest to rustls and guard openssl regressions

NicholaiVogel:
## Summary
- switch Forge workspace reqwest to rustls-only TLS by disabling default features
- regenerate Forge lockfile so `openssl-sys`/`native-tls` are removed from the graph
- add a Forge Release CI guard that validates the target-specific dependency graph

## Why
Forge Release on `main` failed on Linux ARM64 cross-build due to `openssl-sys` discovery during cross-compilation:
- failed run: https://github.com/Signet-AI/signetai/actions/runs/23609420572
- failing job: `build (ubuntu-latest, aarch64-unknown-linux-gnu, forge-linux-arm64.tar.gz, aarch64-linux-gnu-gcc)`

Moving to rustls-only TLS removes the cross OpenSSL toolchain requirement and is more portable across release targets.

## Compatibility Note
Switching from native-tls to rustls changes Forge's TLS trust store behavior. With native-tls, reqwest uses the OS certificate trust store. With rustls-tls, Forge relies on Mozilla's root certificates (webpki-roots). This covers standard public HTTPS endpoints but will not work with enterprise or private CAs that are not in Mozilla's root store. Users with custom certificate chains should use the native-tls build of reqwest or configure their own certificate handling.

## Validation
- `cd packages/forge && cargo build --release --no-default-features --target x86_64-unknown-linux-gnu`
- `cd packages/forge && cargo tree -i openssl-sys -e features --target aarch64-unknown-linux-gnu` returns no match
- `cd packages/forge && cargo tree -i native-tls -e features --target aarch64-unknown-linux-gnu` returns no match

## Incident -> Guardrail
This PR includes a durable CI prevention mechanism:
- new Forge Release guard step validates the target-specific dependency graph (`--target "$RELEASE_TARGET"`) before build, catching openssl-sys/native-tls re-entering any cross-compilation target.

### DIFF
--- a/.github/workflows/forge-release.yml
+++ b/.github/workflows/forge-release.yml
@@ -90,16 +90,18 @@ jobs:
         working-directory: packages/forge
         run: |
           set -euo pipefail
-          if cargo tree -i openssl-sys -e features >/tmp/openssl-tree.log 2>&1; then
-            echo "::error::Forge includes openssl-sys in the dependency graph. Keep Forge on rustls-only reqwest features for portable cross-builds."
+          if cargo tree -i openssl-sys -e features --target "$RELEASE_TARGET" >/tmp/openssl-tree.log 2>&1; then
+            echo "::error::Forge includes openssl-sys in the dependency graph for target $RELEASE_TARGET. Keep Forge on rustls-only reqwest features for portable cross-builds."
             cat /tmp/openssl-tree.log
             exit 1
           fi
-          if cargo tree -i native-tls -e features >/tmp/native-tls-tree.log 2>&1; then
-            echo "::error::Forge includes native-tls in the dependency graph. Keep Forge on rustls-only reqwest features for portable cross-builds."
+          if cargo tree -i native-tls -e features --target "$RELEASE_TARGET" >/tmp/native-tls-tree.log 2>&1; then
+            echo "::error::Forge includes native-tls in the dependency graph for target $RELEASE_TARGET. Keep Forge on rustls-only reqwest features for portable cross-builds."
             cat /tmp/native-tls-tree.log
             exit 1
           fi
+        env:
+          RELEASE_TARGET: ${{ matrix.target }}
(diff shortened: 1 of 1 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(forge): switch reqwest to rustls and guard openssl regressions` (commit `8b484235`) now and I'll follow up shortly with feedback.

.github/workflows/forge-release.yml
```diff
@@ -85,6 +85,21 @@ jobs:
         with:
           workspaces: packages/forge
```
PR-Reviewer-Ant, line 86: Blocking: this guard checks the host dependency graph, not the release target graph that originally failed. Because `cargo tree -i openssl-sys -e features` and the `native-tls` variant omit `--target "$RELEASE_TARGET"`, they can miss target-specific TLS dependencies during cross-compilation. The PR description claims this is the durable prevention mechanism for the ARM64 regression, but in its current form it does not actually prove that for `aarch64-unknown-linux-gnu` (or the other matrix targets).

packages/forge/Cargo.toml
```diff
@@ -27,7 +27,7 @@ serde_json = "1"
 serde_yml = "0.0.12"
```
PR-Reviewer-Ant, line 27: Switching reqwest to `default-features = false` + `rustls-tls` is not just a build-system change; it also changes runtime certificate trust behavior. Forge will no longer use the platform/native trust store that `native-tls` provided, so HTTPS requests that previously worked with enterprise/private root CAs can now fail. The PR description claims a portability fix for releases, but the implementation also introduces this user-visible TLS compatibility change without any mitigation or validation for those environments.

PR #361, opened 2026-03-26
feat(native): RaBitQ compressed vector search in Rust (replaces TS PR #346)

BusyBee3333:
## Summary

Rewrites the RaBitQ vector compression algorithm from TypeScript (PR #346) into the existing `packages/native` napi-rs Rust crate.

## Why Rust over TypeScript

The pure-TS implementation is algorithmically correct, but runs Householder QR decomposition, Beta codebook construction, and LUT-based dot-product search across N×768-dim vectors inside V8 — with GC pressure on the daemon's main thread and no SIMD. Moving to Rust eliminates GC pauses entirely and enables cache-friendly memory layouts.

### `packages/native/src/rabitq.rs` — Core algorithm
- **Xoshiro128\*\*** PRNG hand-rolled to match the TS implementation bit-for-bit (splitmix32 seed expansion, identical state transitions) — no `rand` crate needed
- **Householder QR** decomposition for random orthogonal rotation matrix
- **Beta(d/2, d/2) codebook** via Newton-bisection + Lentz continued fraction (lgamma via Lanczos)
- `quantize()`: rotate → center/scale → nearest centroid → 4-bit pack
- `dequantize()`: unpack → centroid lookup → rescale → inverse rotate → norm correct
- **LUT-based compressed search** (precomputed dot-product lookup tables)
- **Binary serialization** (`RBQ\0` magic) for opaque Buffer FFI handles
- **8 unit tests**: PRNG determinism, matrix orthogonality, codebook properties, round-trip accuracy, search recall, serialization, edge cases

### `packages/native/src/rabitq_index.rs` — napi-rs bindings
8 JS-facing `#[napi]` functions using `Buffer` for all vector data (zero per-element FFI marshalling):
- `rabitqGenerateRotationMatrix(dim, seed)` → Buffer
- `rabitqComputeCodebook(bits, dim)` → Buffer
- `rabitqBuildIndex(vectorsBuf, ids, bits, dim, seed)` → Buffer (serialized index)
- `rabitqSearch(indexBuf, query, k)` → `[{id, score}]`
- `rabitqCompress` / `rabitqDecompress` / `rabitqRecall` / `rabitqIndexStats`

### `packages/daemon/src/rabitq.ts` — Thin TS wrapper
Drop-in replacement for the pure-TS `rabitq.ts` from PR #346. Same exported interface, delegates all computation to `@signet/native`.

### `packages/daemon/src/rabitq.test.ts` — Test suite
Same test cases as PR #346 adapted for native bindings.

## No new dependencies
Zero new Cargo dependencies — the xoshiro128\*\* PRNG is hand-rolled to guarantee bit-exact determinism with the TS reference implementation.

## Relationship to PR #346
This is the Rust counterpart to BusyBee3333's PR #346. Same algorithm, same interface, same test coverage — just native.

### DIFF
--- a/packages/daemon/src/rabitq.ts
+++ b/packages/daemon/src/rabitq.ts
@@ -1,0 +1,227 @@
+/**
+ * RaBitQ — Thin TypeScript wrapper over native Rust implementation.
+ *
+ * Replaces the pure-TS rabitq.ts with native bindings from @signet/native.
+ * All heavy computation (rotation matrix generation, codebook computation,
+ * quantization, compressed search) is done in Rust; this module provides
+ * the same public API surface for drop-in compatibility.
+ */
+
+import {
+	rabitqBuildIndex,
+	rabitqSearch,
+	rabitqCompress,
+	rabitqDecompress,
+	rabitqGenerateRotationMatrix,
+	rabitqComputeCodebook,
+	rabitqBruteForceSearch,
+	rabitqIndexInfo,
+	type RabitqSearchResult,
+	type RabitqIndexInfo,
+} from "@signet/native";
+
+// ---------------------------------------------------------------------------
+// Types (match the original TS interface)
+// ---------------------------------------------------------------------------
+
+/** Search result from compressed search. */
+export interface CompressedSearchResult {
+	id: string;
+	score: number;
+}
+
+/** Opaque compressed index handle (serialized Buffer from Rust). */
+export type CompressedIndexHandle = Buffer;
+
+/** Index metadata (read from a serialized handle). */
+export interface CompressedIndexMeta {
+	bits: number;
+	dim: number;
+	count: number;
+	codebookSize: number;
+}
+
+// ---------------------------------------------------------------------------
+// Rotation Matrix + Codebook
+// ---------------------------------------------------------------------------
+
+/**
+ * Generate a random orthogonal rotation matrix via Householder QR.
+ *
+ * @param dim  Dimensionality (768 for nomic-embed)
+ * @param seed Optional seed for deterministic generation
+ * @returns Buffer containing dim×dim f32 values (row-major, little-endian)
+ */
+export function generateRotationMatrix(dim: number, seed?: number): Buffer {
+	return rabitqGenerateRotationMatrix(dim, seed ?? Date.now()) as Buffer;
+}
+
+/**
+ * Compute the RaBitQ codebook centroids.
+ *
+ * @param bits Number of quantization bits (e.g. 4 → 16 centroids)
+ * @param dim  Vector dimensionality
+ * @returns Buffer containing 2^bits f32 centroid values (little-endian)
+ */
+export function computeCodebook(bits: number, dim: number): Buffer {
+	return rabitqComputeCodebook(bits, dim) as Buffer;
+}
+
+// ---------------------------------------------------------------------------
+// Index Build
+// ---------------------------------------------------------------------------
+
+/**
+ * Build a compressed RaBitQ index from vectors.
+ *
+ * Handles rotation matrix generation and codebook computation internally.
+ *
+ * @param vectors Array of Float32Array embeddings (each dim-dimensional)
+ * @param ids     Corresponding memory IDs
+ * @param bits    Quantization bits per coordinate (default: 4)
+ * @param dim     Vector dimensionality (inferred from first vector if omitted)
+ * @param seed    PRNG seed for rotation matrix (default: 42)
+ * @returns Opaque CompressedIndexHandle (Buffer)
+ */
+export function buildIndex(
+	vectors: ReadonlyArray<Float32Array>,
+	ids: ReadonlyArray<string>,
+	bits = 4,
+	dim?: number,
+	seed = 42,
+): CompressedIndexHandle {
+	const actualDim = dim ?? vectors[0]?.length ?? 768;
+
+	// Concatenate all vectors into a single Buffer for FFI
+	const totalBytes = vectors.length * actualDim * 4;
+	const buf = Buffer.alloc(totalBytes);
+	let offset = 0;
+	for (const vec of vectors) {
+		for (let i = 0; i < actualDim; i++) {
+			buf.writeFloatLE(vec[i] ?? 0, offset);
+			offset += 4;
+		}
+	}
+
+	return rabitqBuildIndex(buf, ids as string[], bits, actualDim, seed) as Buffer;
+}
+
+// ---------------------------------------------------------------------------
+// Search
+// ---------------------------------------------------------------------------
+
+/**
+ * Search a compressed index for approximate nearest neighbours.
+ *
+ * @param indexHandle Opaque CompressedIndexHandle from buildIndex()
+ * @param query      Query vector (Float32Array)
+ * @param topK       Number of results to return
+ * @returns Array of {id, score} sorted by descending score
+ */
+export function compressedSearch(
+	indexHandle: CompressedIndexHandle,
+	query: Float32Array,
+	topK: number,
+): CompressedSearchResult[] {
+	return rabitqSearch(indexHandle, query, topK) as CompressedSearchResult[];
+}
+
+// ---------------------------------------------------------------------------
+// Compress / Decompress (with pre-computed rotation + codebook)
+// ---------------------------------------------------------------------------
+
+/**
+ * Quantize vectors with pre-computed rotation matrix and codebook.
+ *
+ * Use when you want to cache the rotation matrix and codebook across
+ * multiple index builds (e.g. incremental updates).
+ */
+export function compress(
+	vectors: ReadonlyArray<Float32Array>,
+	ids: ReadonlyArray<string>,
+	rotationMatrix: Buffer,
+	codebook: Buffer,
+	bits: number,
+	dim: number,
+): CompressedIndexHandle {
+	const totalBytes = vectors.length * dim * 4;
+	const buf = Buffer.alloc(totalBytes);
+	let offset = 0;
+	for (const vec of vectors) {
+		for (let i = 0; i < dim; i++) {
+			buf.writeFloatLE(vec[i] ?? 0, offset);
+			offset += 4;
+		}
+	}
+
+	return rabitqCompress(buf, ids as string[], rotationMatrix, codebook, bits, dim) as Buffer;
+}
+
+/**
+ * Decompress (dequantize) a compressed index back to approximate vectors.
+ *
+ * @param indexHandle Serialized CompressedIndex buffer
+ * @param dim        Vector dimensionality (needed to parse output)
+ * @returns Array of Float32Array approximate vectors
+ */
+export function decompress(indexHandle: CompressedIndexHandle, dim: number): Float32Array[] {
+	const resultBuf = rabitqDecompress(indexHandle) as Buffer;
+	const floatCount = resultBuf.length / 4;
+	const numVectors = floatCount / dim;
+	const vectors: Float32Array[] = [];
+
+	for (let i = 0; i < numVectors; i++) {
+		const vec = new Float32Array(dim);
+		const base = i * dim * 4;
+		for (let j = 0; j < dim; j++) {
+			vec[j] = resultBuf.readFloatLE(base + j * 4);
+		}
+		vectors.push(vec);
+	}
+
+	return vectors;
+}
+
+// ---------------------------------------------------------------------------
+// Index Info
+// ---------------------------------------------------------------------------
+
+/**
+ * Get metadata about a serialized compressed index.
+ */
+export function getIndexInfo(indexHandle: CompressedIndexHandle): CompressedIndexMeta {
+	const info = rabitqIndexInfo(indexHandle) as RabitqIndexInfo;
+	return {
+		bits: info.bits,
+		dim: info.dim,
+		count: info.count,
+		codebookSize: info.codebookSize,
+	};
+}
+
+// ---------------------------------------------------------------------------
+// Brute Force Search (for recall evaluation)
+// ---------------------------------------------------------------------------
+
+/**
+ * Brute-force cosine similarity search (ground truth baseline).
+ */
+export function bruteForceSearch(
+	query: Float32Array,
+	vectors: ReadonlyArray<Float32Array>,
+	ids: ReadonlyArray<string>,
+	topK: number,
+): CompressedSearchResult[] {
+	const dim = query.length;
+	const totalBytes = vectors.length * dim * 4;
+	const buf = Buffer.alloc(totalBytes);
+	let offset = 0;
+	for (const vec of vectors) {
+		for (let i = 0; i < dim; i++) {
+			buf.writeFloatLE(vec[i] ?? 0, offset);
+			offset += 4;
+		}
+	}
+
+	return rabitqBruteForceSearch(query, buf, ids as string[], dim, topK) as CompressedSearchResult[];
+}
--- a/packages/native/src/lib.rs
+++ b/packages/native/src/lib.rs
@@ -1,4 +1,6 @@
 mod knn;
 mod normalization;
+pub mod rabitq;
+mod rabitq_index;
 mod search;
 mod vector;
--- a/packages/native/src/rabitq.rs
+++ b/packages/native/src/rabitq.rs
@@ -429,6 +429,7 @@ fn unpack_4bit_pair(byte: u8) -> (u8, u8) {
 // ---------------------------------------------------------------------------
 
 /// Metadata for a single compressed vector.
+#[derive(Debug)]
 pub struct CompressedVector {
     /// Original vector ID.
     pub id: String,
@@ -443,6 +444,7 @@ pub struct CompressedVector {
 }
 
 /// Immutable compressed index holding all quantized vectors and metadata.
+#[derive(Debug)]
 pub struct CompressedIndex {
     /// Number of bits per coordinate (e.g. 4 → 16 centroids).
     pub bits: u32,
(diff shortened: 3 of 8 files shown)


PR-Reviewer-Ant:
Hi @BusyBee3333 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(native): RaBitQ compressed vector search in Rust (replaces TS PR #346)` (commit `96650da6`) now and I'll follow up shortly with feedback.

packages/native/src/rabitq.rs
```diff
@@ -0,0 +1,1214 @@
...
+    // Pre-compute sum of rotated query components
+    let query_rotated_sum: f32 = rotated_query.iter().sum();
+
+    // 4. Score each compressed vector
+    let mut scores: Vec<(usize, f64)> = Vec::with_capacity(vectors.len());
+
+    for (vi, cv) in vectors.iter().enumerate() {
+        let mut dot_approx = 0.0f32;
+
+        if bits == 4 {
```
PR-Reviewer-Ant, line 728: Zero-norm vectors are accepted during quantization, but search later computes `approx_dot / (query_norm * cv.norm as f64)` with no `cv.norm == 0` guard. A non-zero query against an index containing a zero vector will produce `inf`/`NaN` scores and can rank the zero vector incorrectly at the top. The added tests only cover the zero-query case, not the zero-vector-in-index case.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,357 @@
...
+/// @param seed - PRNG seed for rotation matrix
+/// @returns Buffer containing the serialized CompressedIndex
+#[napi]
+pub fn rabitq_build_index(
+    vectors_buf: Buffer,
+    ids: Vec<String>,
+    bits: u32,
+    dim: u32,
+    seed: u32,
+) -> napi::Result<Buffer> {
```
PR-Reviewer-Ant, line 203: These napi entrypoints do not validate `dim > 0` before computing `float_bytes = dim * 4` and then using `bytes.len() % float_bytes`. Passing `dim = 0` from JS will panic with modulo-by-zero instead of returning a validation error, which is an input-boundary bug and a potential denial-of-service for any caller that reaches this native surface. The same pattern appears in `rabitq_build_index`, `rabitq_compress`, and `rabitq_brute_force_search`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `96650da6`

[Automated Review] Found two blocking correctness issues in the new native RaBitQ surface.

**Confidence:** High [sufficient_diff_evidence, missing_runtime_repro, missing_cross_module_context] - The two findings are directly provable from the diff itself. In `packages/native/src/rabitq.rs`, compressed search divides by `cv.norm` without guarding zero-norm indexed vectors, which can yield `inf`/`NaN` scores and corrupt ranking. In `packages/native/src/rabitq_index.rs`, the exported napi entrypoints compute `bytes.len() % (dim * 4)` before validating `dim > 0`, so a caller can trigger a modulo-by-zero panic instead of a structured error. I could not run the repo because the workspace does not contain the repository tree, and I could not inspect surrounding call sites beyond the provided diff.

BusyBee3333:
Hi @BusyBee3333 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(native): RaBitQ compressed vector search in Rust (replaces TS PR #346)` (commit `96650da6`) now and I'll follow up shortly with feedback.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,357 @@
...
+///
+/// Unlike `rabitqBuildIndex`, this takes pre-computed rotation matrix and codebook
+/// buffers, allowing the caller to cache and reuse them.
+///
+/// @param vectors_buf - Concatenated f32 vectors (little-endian)
+/// @param ids - String IDs
+/// @param rotation_matrix_buf - Buffer of dim×dim f32 rotation matrix
+/// @param codebook_buf - Buffer of 2^bits f32 codebook centroids
+/// @param bits - Quantization bits
+/// @param dim - Vector dimensionality
```
BusyBee3333, line 292: **Blocking — per-call deserialization kills the stated performance goal.** `rabitq_search` (and `rabitq_decompress`, `rabitq_index_info`) fully deserializes the index buffer on every call, including parsing `dim×dim` f32 floats for the rotation matrix. For a 768-dim model this is ~2.4 MB of f32 parsing per search invocation. The PR description cites eliminating GC pressure as motivation, but this architectural choice re-pays that cost in Rust with no GC. The index buffer should be held in a napi `External` or stored in a JS-side cache (e.g., a `Map` keyed by a fingerprint) so the rotation matrix and codebook survive across calls. As written, performance will degrade with dimension size, potentially worse than the TS baseline for large indices.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    // Parse vectors from buffer
+    let mut vectors: Vec<Vec<f32>> = Vec::with_capacity(num_vectors);
+    for i in 0..num_vectors {
+        let base = i * float_bytes;
+        let mut vec = Vec::with_capacity(dim_usize);
+        for j in 0..dim_usize {
+            let offset = base + j * 4;
+            let val = f32::from_le_bytes([
+                bytes[offset],
```
BusyBee3333: **Blocking — panic on dim=0.** `let float_bytes = dim_usize * 4;` followed by `bytes.len() % float_bytes` panics with a divide-by-zero when `dim = 0`. Same issue exists in `rabitq_compress` and `rabitq_brute_force_search`. Add an explicit guard: `if dim == 0 { return Err(...) }` at the top of each function.

packages/native/src/rabitq.rs
```diff
@@ -0,0 +1,1214 @@
...
+        let approx_dot = scale * dot_approx + cv.mean * query_rotated_sum;
+
+        // Approximate cosine similarity
+        let cosine = approx_dot as f64 / (query_norm * cv.norm as f64);
+
+        scores.push((vi, cosine));
+    }
+
+    // 5. Sort and return top-K
+    scores.sort_by(|a, b| b.1.partial_cmp(&a.1).unwrap_or(std::cmp::Ordering::Equal));
```
BusyBee3333, line 774: **Blocking — silent NaN scores from zero-norm stored vectors.** In `compressed_search`, the cosine is computed as `approx_dot as f64 / (query_norm * cv.norm as f64)`. When a zero vector is inserted (norm=0), this produces NaN for every score involving that vector. NaN propagates silently into the sort (via the `unwrap_or(Equal)` fallback, meaning NaN entries sort arbitrarily), and NaN scores are returned to callers. The query-norm guard exists but there is no analogous guard for `cv.norm`. Add a skip or score=0 branch when `cv.norm < 1e-12`.

packages/native/src/rabitq.rs
```diff
@@ -0,0 +1,1214 @@
...
+    if data.len() < 24 {
+        return Err("Buffer too short for header".to_string());
+    }
+
+    let mut pos = 0;
+
+    let read_u32 = |data: &[u8], pos: &mut usize| -> Result<u32, String> {
+        if *pos + 4 > data.len() {
+            return Err("Unexpected end of buffer".to_string());
+        }
```
BusyBee3333, line 927: **Blocking — uncapped allocation in deserialize_index enables OOM.** `let rot_len = (dim as usize) * (dim as usize);` is followed immediately by `Vec::with_capacity(rot_len)` with no upper-bound check on `dim`. A crafted buffer with `dim = 65535` would attempt to allocate a 17 GB Vec of f32. Similarly `codebook_len` and `count` are read from the buffer with no caps. Since the Buffer crosses the napi FFI boundary from JS, any caller (including untrusted content piped through the search index) can trigger OOM. Validate that `dim <= 8192`, `bits <= 16`, and `count` is within a sane bound before allocating.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `96650da6`

[Automated Bot Review] Found multiple blocking issues: per-call full deserialization negates the stated performance goal, panic on dim=0 input, silent NaN scores from zero-norm stored vectors, OOM-able deserialization with no size caps, and a missing daemon-rs parity update required by AGENTS.md.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All findings are directly provable from the diff. The per-call deserialization issue is visible in rabitq_index.rs rabitq_search: it calls deserialize_index on every invocation with no caching. The div-by-zero is explicit at the cosine computation in rabitq.rs compressed_search. The dim=0 panic is trivially derivable from the `bytes.len() % float_bytes` check when float_bytes=0. The OOM vector is directly from the uncapped rot_len allocation in deserialize_index. The daemon-rs parity requirement is from AGENTS.md and the new daemon TS file is visible in the diff.


- packages/native/Cargo.toml:1 - **Nitpick — commit type should be `perf:` not `feat:`.** Per repo conventions, `feat:` bumps the minor version and is reserved for user-facing features. This PR is a performance rewrite of an existing algorithm (replacing TS with native Rust). `perf(native):` is the appropriate type and avoids an unintended version bump.

packages/daemon/src/rabitq.ts
```diff
@@ -0,0 +1,227 @@
+/**
```
BusyBee3333, line 1: **Blocking — missing daemon-rs parity update.** AGENTS.md states: *"Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`."* This PR adds a new `rabitq.ts` module to `@signet/daemon` but the diff contains no corresponding update to `packages/daemon-rs/`. If the shadow daemon is enabled (`shadowEnabled: true`), it will diverge on any codepath that invokes rabitq, producing divergence log noise and potentially masking real regressions.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,357 @@
...
+        .collect();
+
+    // Parse vectors
+    if vec_bytes.len() % float_bytes != 0 {
+        return Err(napi::Error::from_reason(format!(
+            "vectors_buf length {} is not a multiple of dim*4 ({})",
+            vec_bytes.len(),
+            float_bytes
+        )));
+    }
```
BusyBee3333: **Warning — misleading variable name conceals correctness risk.** In `rabitq_compress`: `let num_centroids = (1usize << bits) * 4;` — this is actually a *byte count*, not a centroid count. The name `num_centroids` will confuse future readers into thinking a size-in-elements check is being done, when it's actually a size-in-bytes check. Additionally, there is no validation that `bits` is within a safe range (e.g., 1–16). For `bits >= 64` on a 64-bit system, `1usize << bits` is undefined behavior (wrapping in release); in practice napi-rs passes the JS number as u32 so the real risk is `bits = 32` causing `(1u32 << 32)` to panic in Rust debug builds. Rename to `codebook_bytes` and add `if bits > 16 { return Err(...) }`.

packages/native/src/rabitq.rs
```diff
@@ -0,0 +1,1214 @@
...
+
+// ---------------------------------------------------------------------------
+// Quantize
+// ---------------------------------------------------------------------------
+
+/// Quantize a batch of vectors into a CompressedIndex.
+///
+/// Algorithm:
+/// 1. Rotate each vector using the orthogonal rotation matrix
+/// 2. Compute norm and mean of rotated vector
```
BusyBee3333, line 477: **Warning — `find_nearest_centroid` panics on empty codebook.** `let mut hi: usize = codebook.len() - 1;` underflows (wraps to `usize::MAX`) if the codebook is empty, causing an immediate out-of-bounds access. While in normal operation the codebook is always non-empty, a deserialized index with `codebook_len = 0` (e.g., from a truncated or adversarially crafted buffer) reaches this function and panics. Add `if codebook.is_empty() { return 0; }` as a guard.

packages/native/src/lib.rs
```diff
@@ -1,4 +1,6 @@
 mod knn;
 mod normalization;
+pub mod rabitq;
```
BusyBee3333, line 3: **Warning — `pub mod rabitq` unnecessarily widens crate API surface.** The rabitq module is `pub` but its types and functions are only needed internally (accessed via `crate::rabitq` from `rabitq_index.rs`). Making it `pub` exposes `Xoshiro128ss`, `CompressedIndex`, `CompressedVector`, and all helper functions as public crate API, inviting unintended downstream dependency. Change to `pub(crate) mod rabitq` or just `mod rabitq`.

packages/daemon/src/rabitq.ts
```diff
@@ -0,0 +1,227 @@
...
+	for (const vec of vectors) {
+		for (let i = 0; i < actualDim; i++) {
+			buf.writeFloatLE(vec[i] ?? 0, offset);
+			offset += 4;
+		}
+	}
+
+	return rabitqBuildIndex(buf, ids as string[], bits, actualDim, seed) as Buffer;
+}
+
```
BusyBee3333, line 108: **Warning — `dim` inferred from first vector can silently accept jagged input.** In `buildIndex`: `const actualDim = dim ?? vectors[0]?.length ?? 768`. If `vectors` is non-empty but `dim` is omitted and vectors have inconsistent lengths, the Rust side returns an error only after detecting the mismatch — but the TS wrapper serializes all vectors using `actualDim` (the first vector's length), silently padding/truncating short/long vectors with `vec[i] ?? 0`. A caller passing a mixed-dimension array will get a corrupted index with no TS-level warning. Consider asserting `vectors.every(v => v.length === actualDim)` before serializing.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,443 @@
...
+    let data: &[u8] = &index_buf;
+    let query_owned: Vec<f32> = query.to_vec();
+
+    let mut cache = INDEX_CACHE
+        .lock()
+        .map_err(|e| napi::Error::from_reason(format!("Cache lock poisoned: {}", e)))?;
+
+    let results = cache
+        .with_index(data, |index| {
+            rabitq::compressed_search(&query_owned, index, k as usize)
```
PR-Reviewer-Ant, line 265: `rabitq_build_index` does not validate `bits` before calling `rabitq::compute_codebook(bits, dim)`. Inside `compute_codebook`, the first operation is `1u32 << bits`. If a JS caller passes `bits = 0` or `bits >= 32`, this panics and terminates the Node.js process. `rabitq_compute_codebook` already has the correct guard (`if bits == 0 || bits > 8`); the same check must be added here and in `rabitq_compress` before the call to `compute_codebook`.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,443 @@
...
+    let cb_bytes: &[u8] = &codebook_buf;
+    let dim_usize = dim as usize;
+
+    // Parse rotation matrix
+    let expected_rot_bytes = dim_usize * dim_usize * 4;
+    if rot_bytes.len() != expected_rot_bytes {
+        return Err(napi::Error::from_reason(format!(
+            "rotation_matrix_buf length {} != expected {} (dim={})",
+            rot_bytes.len(),
+            expected_rot_bytes,
```
PR-Reviewer-Ant, line 316: Same missing `bits` validation as in `rabitq_build_index`: `rabitq_compress` calls `rabitq::compute_codebook(bits, dim)` and does `1usize << bits` to compute `num_centroids` without any prior bounds check on `bits`. A JS caller with `bits >= 32` will crash the process.

packages/native/src/rabitq_index.rs
```diff
@@ -0,0 +1,443 @@
...
+        let ac = self.access_counter;
+
+        if let Some(entry) = self.entries.get_mut(&key) {
+            entry.last_access = ac;
+            return f(&entry.index);
+        }
+
+        // Deserialize
+        let index = rabitq::deserialize_index(data)
+            .map_err(|e| napi::Error::from_reason(e))?;
```
PR-Reviewer-Ant, line 62: The function is named `blake3_hash` but uses `sha2::Sha256`. The in-body comment calls it "FNV-1a style". This is a naming error — none of the three names match. Functionally harmless for cache keying (collisions are only cache misses), but the mismatch makes it harder to reason about the security properties if the hash ever gets used for anything beyond an LRU key.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4b982ea9`

[Automated Review — pr-reviewer bot] Both blocking issues from the prior review are addressed in the incremental diff. One new blocking issue was found: `bits` is not validated in `rabitq_build_index` or `rabitq_compress`, creating a crash path reachable from JS. One warning and one nitpick also flagged.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The `bits` validation gap is directly provable from the changed code in `packages/native/src/rabitq_index.rs`. `rabitq_compute_codebook` validates `bits == 0 || bits > 8`, but `rabitq_build_index` and `rabitq_compress` both call `rabitq::compute_codebook(bits, dim)` with no prior `bits` check. In `compute_codebook`, the first operation is `1u32 << bits` — if JS passes bits ≥ 32 this panics and kills the Node process. The two previously flagged issues (zero-norm division and dim=0 modulo panic) are confirmed fixed by the incremental diff.


- packages/native/Cargo.toml:1 - Per AGENTS.md conventions, `feat:` bumps the minor version and is reserved for user-facing features. This is a native-acceleration refactor of an existing algorithm. The commit type should be `perf:` or `refactor:`.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4b982ea9`

PR #362, opened 2026-03-26
feat(native): TurboQuant KV cache compression in Rust (replaces TS PR #355)

BusyBee3333:
## Summary

Rewrites the TurboQuant KV cache compression algorithm (arXiv:2504.19874, ICLR 2026) from TypeScript (PR #355) into the existing `packages/native` napi-rs Rust crate.

> **Note:** `native-generation.ts` is intentionally left in TypeScript — it depends on `@huggingface/transformers` (ONNX JS runtime). Only the numerically intensive compression hot path moves to Rust.

## Why Rust over TypeScript

TurboQuant compression runs on every KV cache insert during generative inference — once per token per layer per head. In TS this means GC pressure every forward pass, directly affecting TTFT and generation latency. Rust gives us deterministic latency, no allocation overhead on the hot path, and the same algorithm with zero runtime surprises.

### `packages/native/src/turboquant.rs` — Core engine
- **Xoshiro128\*\*** PRNG with splitmix32 seed expansion (bit-exact match to TS)
- **Box-Muller** normal sampling + **Modified Gram-Schmidt** QR decomposition
- **Beta((d-1)/2, (d-1)/2) codebook**: lgamma (Lanczos), beta_inc (Lentz CF), beta_incinv (Newton), codebook boundary computation via numerical integration
- **Bit packing/unpacking**: 4-bit fast path + generic bitstream for 1/2/3-bit
- **`TurboQuantEngine`** struct: `compress_vector()` / `decompress_vector()` with full rotation, quantization, dequantization lifecycle
- **14 unit tests**: codebook properties, round-trip PSNR, bit packing correctness, PRNG determinism, dimension scaling

### `packages/native/src/turboquant_cache.rs` — napi-rs KV cache manager
Full residual window logic runs in Rust (the hot path). Thread-safe state via `External<Arc<Mutex<...>>>`:
- **13 `#[napi]` exported functions** + **6 `#[napi(object)]` types**
- `turboquantCompress` / `turboquantDecompress` — single-vector API via Buffer
- `turboquantCreateCache` — allocate opaque cache handle
- `turboquantCacheInsert` — insert KV pair, auto-compresses tokens beyond residual window
- `turboquantCacheGet` — retrieve compressed entries + residual vectors for a layer
- `turboquantCacheReconstruct` — reconstruct full f32 tensor from compressed + residual
- `turboquantCacheAdvance` / `turboquantCacheReset` / `turboquantCacheStats`

### `packages/daemon/src/turboquant-cache.ts` — Thin TS wrapper
Drop-in replacement for the pure-TS `turboquant-cache.ts` from PR #355. Same exported interface, same `TurboQuantKvCacheConfig` shape, delegates all compression to `@signet/native`.

### `packages/daemon/src/turboquant-cache.test.ts` — Test suite
Same test cases adapted for native bindings. All 20 tests pass (`cargo test` clean).

## No new dependencies
Zero new Cargo dependencies. All math (Beta CDF, Lanczos gamma, Lentz CF) is implemented from first principles.

## Relationship to PR #355
This is the Rust counterpart to BusyBee3333's PR #355. Same algorithm, same interface, same test coverage — native.

### DIFF
--- a/packages/native/src/turboquant.rs
+++ b/packages/native/src/turboquant.rs
@@ -1,0 +1,850 @@
+//! TurboQuant KV Cache Compression — Core Algorithm
+//!
+//! Pure Rust implementation of Google's TurboQuant algorithm
+//! (arXiv:2504.19874, ICLR 2026) adapted for KV cache compression
+//! in local generative model inference.
+//!
+//! Compresses key/value cache tensors to 3-4 bits per coordinate with
+//! near-zero accuracy loss, enabling ~6x memory reduction for long
+//! context windows.
+//!
+//! ## Algorithm overview
+//!
+//! 1. Generate a deterministic random rotation matrix via QR decomposition
+//! 2. Compute an optimal codebook from the Beta((d-1)/2, (d-1)/2) distribution
+//! 3. Quantize: rotate each KV vector, find nearest centroid per coordinate
+//! 4. Dequantize: centroid lookup → inverse rotation → rescale by norm
+//!
+//! The PRNG (xoshiro128**), Householder QR, and Beta distribution codebook
+//! are faithful reproductions of the TypeScript implementation, ensuring
+//! identical outputs for the same seed.
+
+use std::f64::consts::PI;
+
+// ---------------------------------------------------------------------------
+// PRNG — splitmix32 + xoshiro128**
+// ---------------------------------------------------------------------------
+
+/// splitmix32 for seed expansion (Fix C from the TS impl).
+fn splitmix32(s: u32) -> u32 {
+    let mut t = s.wrapping_add(0x9e3779b9);
+    t = (t ^ (t >> 16)).wrapping_mul(0x21f0aaad);
+    t = (t ^ (t >> 15)).wrapping_mul(0x735a2d97);
+    t ^ (t >> 15)
+}
+
+/// Seeded xoshiro128** PRNG returning values in [0, 1).
+/// Identical to the TypeScript implementation.
+pub(crate) struct Xoshiro128ss {
+    s0: u32,
+    s1: u32,
+    s2: u32,
+    s3: u32,
+}
+
+impl Xoshiro128ss {
+    pub fn new(seed: u32) -> Self {
+        let s0 = splitmix32(seed);
+        let s1 = splitmix32(s0);
+        let s2 = splitmix32(s1);
+        let s3 = splitmix32(s2);
+
+        let mut rng = Xoshiro128ss { s0, s1, s2, s3 };
+
+        // Warm up (same as TS: 20 iterations)
+        for _ in 0..20 {
+            rng.advance();
+        }
+
+        rng
+    }
+
+    /// Advance the internal state without producing output.
+    fn advance(&mut self) {
+        let t = self.s1 << 9;
+        self.s2 ^= self.s0;
+        self.s3 ^= self.s1;
+        self.s1 ^= self.s2;
+        self.s0 ^= self.s3;
+        self.s2 ^= t;
+        self.s3 = self.s3.rotate_left(11);
+    }
+
+    /// Generate the next random f64 in [0, 1).
+    pub fn next_f64(&mut self) -> f64 {
+        let mul5 = self.s1.wrapping_mul(5);
+        let rotl7 = mul5.rotate_left(7);
+        let result = rotl7.wrapping_mul(9);
+        self.advance();
+        (result as f64) / 4294967296.0
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Matrix utilities — random Gaussian, QR, mat-vec
+// ---------------------------------------------------------------------------
+
+/// Generate a flat row-major matrix of standard normal samples (Box-Muller).
+pub(crate) fn randn_matrix(rows: usize, cols: usize, rng: &mut Xoshiro128ss) -> Vec<f32> {
+    let len = rows * cols;
+    let mut data = vec![0.0f32; len];
+
+    let mut i = 0;
+    while i + 1 < len {
+        let u1 = rng.next_f64().max(1e-10);
+        let u2 = rng.next_f64();
+        let r = (-2.0 * u1.ln()).sqrt();
+        let theta = 2.0 * PI * u2;
+        data[i] = (r * theta.cos()) as f32;
+        data[i + 1] = (r * theta.sin()) as f32;
+        i += 2;
+    }
+    if len % 2 != 0 {
+        let u1 = rng.next_f64().max(1e-10);
+        let u2 = rng.next_f64();
+        data[len - 1] = ((-2.0 * u1.ln()).sqrt() * (2.0 * PI * u2).cos()) as f32;
+    }
+
+    data
+}
+
+/// QR decomposition via modified Gram-Schmidt.
+/// Returns Q (orthogonal) as a flat Vec<f32> (rows × cols, row-major).
+pub(crate) fn qr_q(matrix: &[f32], rows: usize, cols: usize) -> Vec<f32> {
+    let mut q = matrix.to_vec();
+
+    let get_col = |q: &[f32], c: usize| -> Vec<f32> {
+        let mut col = vec![0.0f32; rows];
+        for r in 0..rows {
+            col[r] = q[r * cols + c];
+        }
+        col
+    };
+
+    let set_col = |q: &mut [f32], c: usize, col: &[f32]| {
+        for r in 0..rows {
+            q[r * cols + c] = col[r];
+        }
+    };
+
+    let dot = |a: &[f32], b: &[f32]| -> f64 {
+        let mut s = 0.0f64;
+        for i in 0..a.len() {
+            s += a[i] as f64 * b[i] as f64;
+        }
+        s
+    };
+
+    for j in 0..cols {
+        let mut col = get_col(&q, j);
+
+        for k in 0..j {
+            let qk = get_col(&q, k);
+            let proj = dot(&col, &qk);
+            for i in 0..rows {
+                col[i] -= (proj * qk[i] as f64) as f32;
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Hi @BusyBee3333 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(native): TurboQuant KV cache compression in Rust (replaces TS PR #355)` (commit `21777ac4`) now and I'll follow up shortly with feedback.

packages/native/src/turboquant_cache.rs
```diff
@@ -0,0 +1,667 @@
...
+    /// Residual value vectors concatenated as Buffer.
+    pub residual_values: Buffer,
+}
+
+/// Retrieve all KV data for a specific layer from the cache.
+///
+/// Returns compressed entries plus residual window vectors for all heads.
+/// Heads are interleaved: for each token position, head 0 comes first,
+/// then head 1, etc. This matches the [numHeads, seqLen, headDim] layout.
+///
```
PR-Reviewer-Ant, line 350: This retrieval path assumes `head 0` always exists and uses its compressed/residual lengths as the layer-wide reference. That makes the cache return `None` for a layer that only has data for other heads, and it can also mis-shape the returned buffers if heads are inserted unevenly. The insert API allows arbitrary `head_idx`, so this is a real correctness bug unless you also enforce synchronized head-0-first writes at the boundary.

packages/native/src/turboquant_cache.rs
```diff
@@ -0,0 +1,667 @@
...
+    Ok(Some(CompressedLayerJs {
+        num_compressed: num_compressed as u32,
+        num_residual: num_residual as u32,
+        compressed_keys,
+        compressed_values,
+        residual_keys: residual_keys_data.into(),
+        residual_values: residual_values_data.into(),
+    }))
+}
+
```
PR-Reviewer-Ant, line 424: `turboquantCacheReconstruct` repeats the same head-0/uniform-length assumption and then indexes other heads using `num_compressed`/`num_residual` taken from head 0. If one head lags or head 0 is absent, reconstruction either drops data or panics on `kv.compressed_keys[t]` / `kv.residual_keys[t]`. This diverges from the PR claim of being a drop-in cache manager.

packages/native/src/turboquant.rs
```diff
@@ -0,0 +1,850 @@
...
+    // For keys + values (2x)
+    let indices_per_token = (head_dim as u64 * bits as u64) / 8;
+    let norm_per_token = 4u64; // float32
+    let bytes_per_comp_token =
+        (indices_per_token + norm_per_token) * num_heads as u64 * num_layers as u64 * 2;
+    let compressed_bytes = bytes_per_comp_token * compressed_tokens as u64;
+
+    // Residual: full fp32 per dim per head per layer per token (keys + values)
+    let bytes_per_residual_token =
+        head_dim as u64 * 4 * num_heads as u64 * num_layers as u64 * 2;
```
PR-Reviewer-Ant, line 632: The memory estimator undercounts compressed storage for most 1/2/3-bit shapes because `indices_per_token = (head_dim * bits) / 8` truncates instead of rounding up. For example, a 65-dim 3-bit vector needs `ceil(195/8)=25` bytes, not `24`. The reported compression ratios will be wrong for any model where `headDim * bits` is not byte-aligned.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `21777ac4`

[Automated Review] Found two blocking correctness issues and one accuracy bug in the new native TurboQuant cache path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The findings are directly provable from the changed code. `turboquant_cache_get`/`turboquant_cache_reconstruct` hard-require head 0 and derive layer sizes from it, which breaks valid insertion orders and uneven head state. `compute_memory_stats` uses truncating division for packed bytes, which is mathematically wrong for 1/3-bit cases unless `headDim * bits` is divisible by 8.

BusyBee3333:
Hi @BusyBee3333 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(native): TurboQuant KV cache compression in Rust (replaces TS PR #355)` (commit `21777ac4`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `21777ac4`

```pr-review-json
{
  "summary": "[Automated bot review] PR #362 adds a Rust native implementation of TurboQuant KV cache compression. Three blocking issues were found: a panic-instead-of-error on invalid input to `turboquant_decompress`, a performance regression that recreates the O(d³) QR engine on every compress/decompress call, and a semantically broken `storeCompressed` API that returns immediately-compressed data while storing the raw vector in the residual window. Two additional warnings around head-indexing assumptions and daemon-rs parity are also noted.",
  "verdict": "request_changes",
  "confidence": {
    "level": "high",
    "justification": "All three blocking issues are directly verifiable from the diff: (1) `turboquant_decompress` at turboquant_cache.rs:301 calls `TurboQuantEngine::new` without validating `bits`, while `turboquant_compress` at line 279 validates first — the asymmetry is explicit. (2) `compressVector` at turboquant-cache.ts:270 calls `turboquantCompress(f32ToBuffer(vector), this._nativeConfig)`, which creates a new TurboQuantEngine (full QR decomp) on every invocation, while `this._cacheHandle` holds a pre-computed engine that is never used for single-vector ops. (3) `storeCompressed` at turboquant-cache.ts:293 calls `turboquantCacheInsert` (stores raw vector to residual window) then immediately calls `this.compressVector` (fresh engine compress) and returns those results — the return value does not match what is actually stored in the cache."
  },
  "comments": [
    {
      "file": "packages/native/src/turboquant_cache.rs",
      "line": 302,
      "body": "**Blocking — panic instead of error on invalid input.** `turboquant_decompress` calls `TurboQuantEngine::new(head_dim, bits, seed)` without first validating `bits`. `TurboQuantEngine::new` uses `assert!(matches!(bits, 1 | 2 | 3 | 4))` which panics on invalid input. Contrast with `turboquant_compress` (line ~268) which validates bits and returns `Err(...)` before calling `new`. An unhandled Rust panic in a napi-rs function is undefined behavior — it may abort the Node.js process. Add the same `!matches!(bits, 1 | 2 | 3 | 4)` guard and return a `napi::Error` here.",
      "severity": "blocking"
    },
    {
      "file": "packages/daemon/src/turboquant-cache.ts",
      "line": 270,
      "body": "**Blocking — O(d³) engine reconstruction on every compress/decompress call.** `compressVector` (and `decompressVector`) call `turboquantCompress(f32ToBuffer(vector), this._nativeConfig)`, which inside Rust allocates a full `head_dim × head_dim` random Gaussian matrix and runs a modified Gram-Schmidt QR decomposition (O(d³)) from scratch on every invocation. This is the critical path the PR description explicitly calls out ("once per token per layer per head"). `TurboQuantKvCache` already holds `this._cacheHandle` with a precomputed engine, but it is never used for these operations. The fix is to expose a `turboquantCacheCompressVector(handle, buffer)` / `turboquantCacheDecompressVector(handle, compressed)` napi function that borrows the cached engine, or restructure so single-vector operations go through the handle.",
      "severity": "blocking"
    },
    {
      "file": "packages/daemon/src/turboquant-cache.ts",
      "line": 293,
      "body": "**Blocking — `storeCompressed` returns a value inconsistent with what is stored.** The function calls `turboquantCacheInsert(...)` which places the raw vector into the residual window (not yet compressed), then separately calls `this.compressVector(key)` / `this.compressVector(value)` (fresh engine, immediate compression) and returns those results as `{ key, value }`. The caller receives a `CompressedKvEntry` implying the token has been compressed in the cache, but the cache actually holds it in full precision. If a caller later calls `reconstructLayer` for the same position, it gets the residual-window version — not what matches the returned compressed entry. The return type should either be removed (return `void`), or the function should reconstruct the compressed form from the cache after insertion, or the cache should expose a separate pure-compress operation that doesn't affect stored state.",
      "severity": "blocking"
    },
    {
      "file": "packages/native/src/turboquant_cache.rs",
      "line": 450,
      "body": "**Warning — head-0 assumption in `turboquant_cache_get` and `turboquant_cache_reconstruct`.** Both functions use `layer_map.get(&0)` as the authoritative reference for `num_compressed` and `num_residual`, and return `None` if head 0 is absent. If callers insert only heads 1+ (or have a 0-indexed gap for any reason), the functions return `None` even when data exists. Consider iterating all present heads to derive the max/min sizes rather than pivoting on head 0.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/turboquant-cache.ts",
      "line": 1,
      "body": "**Warning — daemon-rs parity not addressed.** The AGENTS.md parity rule states: \"Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`.\" This PR adds `turboquant-cache.ts` to `@signet/daemon` but the diff contains no corresponding change to `packages/daemon-rs/`. Even if a placeholder stub is acceptable for now, the gap should be acknowledged.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/turboquant-cache.ts",
      "line": 1,
      "body": "**Warning — spec/docs drift.** Per AGENTS.md checklist, any behavior/API/schema change requires updating `docs/specs/INDEX.md`, `docs/specs/dependencies.yaml`, and `docs/API.md` in the same PR. This PR introduces a new user-facing config key (`memory.kv_cache_compression`) and a new exported module, but no doc or spec updates are present.",
      "severity": "warning"
    }
  ]
}
```

PR #364, opened 2026-03-27
refactor(dashboard): stabilize IA and gate experimental surfaces

NicholaiVogel:
## Summary

Stabilizes dashboard information architecture to stop feature sprawl, reduce navigation noise, and gate unfinished surfaces without deleting them.

## Changes

- Added dedicated **Audit** tab and moved logs/troubleshooter surfaces into it
- Kept **Tasks** as its own top-level tab and retained the polished task view
- Renamed and reshaped primary navigation:
  - Home -> **Overview**
  - Marketplace -> **Skills**
  - Memory -> **Ontology**
  - Index -> **Cortex**
- Moved **Settings** button to the bottom of the sidebar
- Disabled (not removed) unstable surfaces for now:
  - pipeline
  - apps
  - predictor
  - connectors
  - MCP install actions
- Refined Ontology/Cortex flow:
  - unified memory views
  - constellation accessible as a separate page within Ontology
  - disabled timeline + knowledge pages for now
  - added ontology context panel on memory-card select (entities/aspects/constraints/dependencies)
- Simplified troubleshooter into one unified terminal panel
- Updated navigation/focus stores and tab loading paths for the new IA
- Hardened legacy hash canonicalization and deep-link behavior for Audit/Ontology subroutes
- Updated planning spec notes for dashboard IA refactor

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

Attached in PR comment:
https://github.com/Signet-AI/signetai/pull/364#issuecomment-4139401034

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Executed:
- manual validation in dev server (`http://127.0.0.1:5173`) for routing/tab behavior

## AI disclosure

## Notes

This is intentionally a containment/stabilization pass. Disabled surfaces remain in code and can be re-enabled after polish.

### DIFF
--- a/docs/specs/planning/dashboard-information-architecture-refactor.md
+++ b/docs/specs/planning/dashboard-information-architecture-refactor.md
@@ -157,9 +157,9 @@ drag-to-resize, transition animations.
 2. Whether identity panel supports split-view editing of two files.
 3. Final placement of OS and Changelog (System group vs footer utility).
 
-## 10) Implemented IA slice (2026-03-26)
+## 10) Implementation notes for planning validation (2026-03-26)
 
-This planning track now has a shipped slice in the dashboard shell:
+Implementation notes observed during planning validation:
 
 - Sidebar now surfaces direct top-level destinations for `Memory`, `Tasks`,
   `Audit`, and `Settings` (with `Cortex` renamed to `Memory` and `Engine`
--- a/packages/cli/dashboard/src/lib/components/app-sidebar.svelte
+++ b/packages/cli/dashboard/src/lib/components/app-sidebar.svelte
@@ -90,11 +90,18 @@ function handleClick(item: NavItem): void {
 	setTab(item.id);
 }
 
+function focusIdForTab(tab: TabId): SidebarFocusItem {
+	if (tab === "settings") return "settings";
+	if (tab === "changelog") return "github-link";
+	const item = navItems.find((entry) => entry.id === tab);
+	if (item) return item.focusId;
+	return "home";
+}
+
 // Initialize sidebar focus on mount — derive from current active tab
 onMount(() => {
 	if (!focus.sidebarItem) {
-		const item = navItems.find(n => isActive(n));
-		setSidebarItem(item?.focusId ?? "home");
+		setSidebarItem(focusIdForTab(nav.activeTab));
 	}
 });
 


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(dashboard): stabilize IA and gate experimental surfaces` (commit `a66b5ac8`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/tabs/AuditTab.svelte
```diff
@@ -0,0 +1,160 @@
...
+type Section = "troubleshooter" | "logs";
+const SECTION_SET = new Set<string>(["troubleshooter", "logs"]);
+
+function readSection(): Section {
+	if (typeof window === "undefined") return "troubleshooter";
+	const hash = window.location.hash.replace("#", "");
+	const parts = hash.split("/");
+	if (parts[0] !== "audit") return "troubleshooter";
+	if (parts[1] && SECTION_SET.has(parts[1])) return parts[1] as Section;
+	return "troubleshooter";
```
PR-Reviewer-Ant, line 11: The selected Audit subsection is read from `window.location.hash` only during initialization. There is no `hashchange` listener, so back/forward navigation between `#audit` and `#audit/logs` can leave the URL and rendered content out of sync while `activeTab` stays `audit`. This breaks bookmark/history behavior for the new Audit tab.

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -32,22 +33,14 @@ const VALID_TABS: ReadonlySet<string> = new Set<TabId>([
 	"home",
```
PR-Reviewer-Ant, line 33: The PR description says switched-off destinations and legacy hashes should redirect to active tabs, but `VALID_TABS` still treats legacy standalone routes like `memory` and `embeddings` as first-class tabs. That means `#memory` / `#embeddings` continue to load the old standalone pages instead of the new `cortex-memory` flow, so the implementation does not fully achieve the stated IA consolidation.

NicholaiVogel:
Final-state screenshots from local dev (`http://127.0.0.1:5173`) via `agent-browser`.

### Overview
Overview

### Ontology (Cortex Search)
Ontology

### Audit (Unified Terminal)
Audit

### Tasks
Tasks

### Skills
Skills

packages/cli/dashboard/src/lib/components/tabs/MemoryTab.svelte
```diff
@@ -54,6 +61,111 @@ let display = $derived(
...
+				mentions: entity.mentions,
+				aspectNames: aspects.map((aspect) => aspect.name),
+				constraints,
+				dependencies: Array.from(new Set(deps)),
+			};
+		})
+		.filter((item): item is RelatedEntity => item !== null);
+});
+
+let relatedAspectCount = $derived(
```
PR-Reviewer-Ant, line 153: The new ontology context panel fetches the overlay with `getConstellationOverlay("default")`. That hardcodes the agent scope and will show the wrong entities/dependencies for any non-default agent. The repo conventions explicitly call this failure mode out. This needs to thread the active agent id instead of introducing a scoping leak.

packages/cli/dashboard/src/lib/components/tabs/UnifiedMemoryTab.svelte
```diff
@@ -0,0 +1,164 @@
...
+	import EmbeddingsTab from "$lib/components/tabs/EmbeddingsTab.svelte";
+	import MemoryTab from "$lib/components/tabs/MemoryTab.svelte";
+	import { Button } from "$lib/components/ui/button/index.js";
+
+	interface Props {
+		memories: Memory[];
+		onopenglobalsimilar: (memory: Memory) => void;
+	}
+
+	const { memories, onopenglobalsimilar }: Props = $props();
```
PR-Reviewer-Ant: This refactor removes the previous hash-backed subsection routing for memory/constellation and replaces it with local `section` state only. As a result, URLs like `#ontology/constellation` and `#cortex-memory/constellation` no longer open the Constellation view after reload or direct navigation, which diverges from the PR description’s claim that Constellation remains a separate page within Ontology and that legacy hashes remain bookmark-compatible.

packages/cli/dashboard/src/lib/components/tabs/AuditTab.svelte
```diff
@@ -0,0 +1,160 @@
+<script lang="ts">
+type Section = "troubleshooter" | "logs";
+const SECTION_SET = new Set<string>(["troubleshooter", "logs"]);
+
```
PR-Reviewer-Ant, line 4: Audit subsection routing is only handled inside this component, but the global navigation store does not recognize `#audit/logs`. On a direct load or reload of that URL, `initNavFromHash()` falls back away from Audit and rewrites the hash, so the new Logs subsection is not bookmarkable. This breaks the PR’s stated backward-compatibility/routing goal for the new IA.

NicholaiVogel:
tightening pass pushed.

kept visual appearance unchanged, but deadened legacy Overview quick links so they no longer route into gated/removed destinations:
- Knowledge
- Constellation (legacy standalone path)
- Pipeline

also remapped the existing Memory quick link to the active ontology surface (`cortex-memory`).

validation: `cd packages/cli/dashboard && bun run check` (passes)

packages/cli/dashboard/src/lib/components/home/QuickLinks.svelte
```diff
@@ -32,7 +37,7 @@
 			variant="ghost"
 			size="sm"
 			class="link-btn"
-			onclick={() => setTab(link.tab)}
```
PR-Reviewer-Ant, line 35: This keyed `{#each}` now uses `link.tab` as the key, but three entries (`Knowledge`, `Constellation`, `Pipeline`) all have `tab: null`. Svelte keyed each blocks require unique keys, so this will throw once the component renders. Use a stable unique key such as `label` instead.

packages/cli/dashboard/src/lib/components/tabs/UnifiedMemoryTab.svelte
```diff
@@ -0,0 +1,164 @@
...
+	import type { Memory } from "$lib/api";
+	import EmbeddingsTab from "$lib/components/tabs/EmbeddingsTab.svelte";
+	import MemoryTab from "$lib/components/tabs/MemoryTab.svelte";
+	import { Button } from "$lib/components/ui/button/index.js";
+
+	interface Props {
+		memories: Memory[];
+		onopenglobalsimilar: (memory: Memory) => void;
+	}
+
```
PR-Reviewer-Ant, line 12: The new Ontology wrapper always initializes `section` to `"cortex"` and never reads or writes the hash. That breaks the PR’s claimed bookmark compatibility for routes like `#memory/constellation`, `#ontology/constellation`, and `#cortex-memory/constellation`: `navigation.svelte.ts` now redirects them to `cortex-memory`, but this component then drops the sub-route and shows Cortex instead of Constellation.

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -56,31 +49,44 @@ const HASH_ALIASES: ReadonlyMap<string, TabId> = new Map([
 	["memory/timeline", "cortex-memory"],
```
PR-Reviewer-Ant, line 47: Audit sub-routes are not routable through the central hash parser. `VALID_TABS` only accepts `audit`, and `HASH_ALIASES` adds no mapping for `audit/logs`, so loading or bookmarking `#audit/logs` will not restore the Logs view. That diverges from the new `AuditTab` design, which presents Logs as a separate page inside Audit.

packages/cli/dashboard/src/lib/components/tabs/MemoryTab.svelte
```diff
@@ -54,6 +61,111 @@ let display = $derived(
...
+
+	return data.entities
+		.map((entity) => {
+			const aspects = entity.aspects.filter((aspect) =>
+				aspect.attributes.some((attr) => attr.memoryId === memId),
+			);
+			if (aspects.length === 0) return null;
+
+			const constraints = Array.from(
+				new Set(
```
PR-Reviewer-Ant, line 113: The ontology context panel fetches `getConstellationOverlay("default")` unconditionally. This reintroduces the repo’s documented scoping failure mode: in any non-default agent/session, the sidebar will read the wrong overlay and can show cross-agent data. The overlay request needs to thread the active agent id instead of hardcoding `default`.

NicholaiVogel:
addressed the latest review feedback without changing visual presentation.

### tightened behavior
- fixed keyed list collision in `QuickLinks` (`key` now uses label, not nullable tab id)
- dead links stay visually unchanged but no longer break keyed rendering
- remapped pinned-entity “Memory” jump to active ontology route (`cortex-memory`)
- added deep-link aliases + parsing for legacy and nested routes:
  - `#memory`, `#embeddings`, `#knowledge` -> `cortex-memory`
  - `#audit/logs`, `#audit/troubleshooter` -> `audit`
- preserved hash subroutes instead of over-normalizing slash routes in nav init/hashchange
- added hashchange listeners to keep subsection state in sync:
  - `AuditTab` (`troubleshooter`/`logs`)
  - `UnifiedMemoryTab` (`cortex`/`constellation`)
- kept Constellation as a separate page within Ontology and made it reload/bookmark-friendly again
- removed hardcoded overlay call in memory context panel; it now reads `agent_id` from URL query with fallback to `default`

### validation run
- `cd packages/cli/dashboard && bun run check` ✅ (0 errors, warnings only)
- root `bun run typecheck` ❌ fails in unrelated package (`@signet/opencode-plugin`: missing node type defs)
- root `bun run lint` ❌ currently fails repo-wide on existing Biome formatting debt outside this PR scope

if you want, i can do a separate cleanup PR for repo-wide lint/typecheck baseline so these checks are green globally.

packages/cli/dashboard/src/lib/components/tabs/MemoryTab.svelte
```diff
@@ -54,6 +61,115 @@ let display = $derived(
...
+			};
+		})
+		.filter((item): item is RelatedEntity => item !== null);
+});
+
+let relatedAspectCount = $derived(
+	relatedEntities.reduce((sum, entity) => sum + entity.aspectNames.length, 0),
+);
+let relatedConstraintCount = $derived(
+	relatedEntities.reduce((sum, entity) => sum + entity.constraints.length, 0),
```
PR-Reviewer-Ant, line 157: Blocking: the new ontology context panel fetches `getConstellationOverlay()` with `new URLSearchParams(window.location.search).get("agent_id") || "default"`. That introduces a fresh agent-scoping leak/wrong-data path: when the dashboard is viewing a non-default agent (or the active agent is not represented in the query string), the side panel will silently load ontology data for `default` instead of the same agent as the memory list. Repo conventions explicitly call out hardcoded `"default"` on scoped data paths as a recurring failure mode.

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -153,17 +149,26 @@ export function navigateToGroup(group: NavGroup): boolean {
  * Returns a cleanup function to remove the event listener.
  */
 export function initNavFromHash(): () => void {
-	const initial = readTabFromHash();
+	const raw = readHash();
+	const initial = readTabFromHash(raw);
 	if (initial) {
 		nav.activeTab = initial;
+		if (typeof window !== "undefined" && raw !== initial && !raw.includes("/")) {
```
PR-Reviewer-Ant: The PR description/spec says legacy hashes for switched-off destinations now redirect to active tabs, but the normalization only runs when the original hash does not contain `/`. Bookmarks like `#engine/logs`, `#cortex/tasks`, or `#memory/knowledge` will resolve internally yet keep the old hash in the URL, so this is not a real redirect and does not fully achieve the stated bookmark-compatibility behavior.

NicholaiVogel:
just pushed another pass for the newest comments.

fixed:
- overlay scope is no longer hardcoded to `"default"`; `getConstellationOverlay()` now accepts optional `agent_id` and omits the query param when absent (no forced default scope)
- legacy hash routes now canonicalize to active routes (including slash paths), so they are true redirects, not just internal remaps
  - examples: `#memory/knowledge -> #cortex-memory`, `#engine/logs -> #audit/logs`, `#memory/constellation -> #cortex-memory/constellation`

validation:
- `cd packages/cli/dashboard && bun run check` ✅
- root `bun run typecheck` / `bun run lint` are still red due to existing repo-wide unrelated issues (same as previous note)

packages/cli/dashboard/src/lib/api.ts
```diff
@@ -2332,9 +2332,13 @@ export interface ConstellationGraph {
 	dependencies: ConstellationDependency[];
 }
 
-export async function getConstellationOverlay(agentId = "default"): Promise<ConstellationGraph | null> {
+export async function getConstellationOverlay(agentId?: string): Promise<ConstellationGraph | null> {
```
PR-Reviewer-Ant: This refactor drops `agent_id` from the knowledge overlay request whenever the dashboard URL does not already include it. Before this change the client always sent a scoped value; now `MemoryTab.ensureGraph()` commonly calls `getConstellationOverlay(undefined)`, which hits `/api/knowledge/constellation` unscoped. That can mix ontology data across agents or depend on a backend default, which violates the repo's scoping guardrail and can show the wrong entities/aspects/dependencies for the selected memory.

NicholaiVogel:
patched the remaining scoping complaint and pushed.

- added `resolveAgentId()` in dashboard API client
- `getConstellationOverlay()` now always threads a resolved agent id (explicit arg -> query param -> document dataset -> localStorage -> default)
- memory ontology context now calls `getConstellationOverlay()` through that resolver path

no visual/UI changes.

validation: `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/components/app-sidebar.svelte
```diff
@@ -81,25 +82,18 @@ function openProjectPage(): void {
...
 	return nav.activeTab === item.id;
 }
 
 function handleClick(item: NavItem): void {
-	if (item.group) {
-		navigateToGroup(item.group);
-	} else {
-		setTab(item.id as TabId);
-	}
+	setTab(item.id);
```
PR-Reviewer-Ant, line 90: The new top-level destinations are not deterministic once a user is already inside a subsection. Example: from `#cortex-memory/constellation`, clicking the sidebar `Ontology` item calls `setTab("cortex-memory")`, but `setTab()` returns early when that tab is already active, so the hash stays on `#cortex-memory/constellation` and the user never gets back to the default Cortex view. The same applies to `Audit` from `#audit/logs`. This diverges from the PR’s stated goal of stable top-level destinations and will make the new IA feel sticky/inconsistent.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `a8c2dc8f`

[Automated Review] Found one navigation regression in the new top-level IA wiring.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new Ontology/Audit shells store their subsection in the URL hash (`#cortex-memory/constellation`, `#audit/logs`), while the sidebar and command palette now navigate only by top-level tab id. In the diff, `setTab()` short-circuits when the requested tab is already active, so clicking a top-level destination while already inside one of its subsections does not reset the hash back to the default subsection. This is directly implied by the changed routing model, although I did not run the UI to reproduce it.

NicholaiVogel:
loop update: fixed newest deterministic-top-level-nav complaint.

- when clicking an already-active top-level tab, `setTab()` now normalizes hash back to that tab root (`#cortex-memory`, `#audit`, etc.)
- this clears sticky sub-routes like `#cortex-memory/constellation` and `#audit/logs` when user explicitly clicks the top-level sidebar item

no visual changes.

validation: `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -128,7 +166,13 @@ export function isCortexGroup(tab: TabId): boolean {
 }
 
 export function setTab(tab: TabId): boolean {
-	if (tab === nav.activeTab) return true;
+	if (!VALID_TABS.has(tab)) return false;
+	if (tab === nav.activeTab) {
+		if (typeof window !== "undefined" && readHash() !== tab) {
```
PR-Reviewer-Ant: Blocking: the new `tab === nav.activeTab` branch clobbers subsection hashes by forcing `history.replaceState(..., `#${tab}`)`. That means already-open deep links like `#audit/logs` and `#cortex-memory/constellation` get reset to `#audit` / `#cortex-memory` whenever the user re-invokes the same tab from the sidebar or command palette. This contradicts the PR's stated goal of hardening Audit/Ontology deep-link behavior and will make those subroutes non-sticky in normal navigation.

packages/cli/dashboard/src/lib/components/app-sidebar.svelte
```diff
@@ -81,25 +82,18 @@ function openProjectPage(): void {
...
 }
 
 // Initialize sidebar focus on mount — derive from current active tab
 onMount(() => {
 	if (!focus.sidebarItem) {
 		const item = navItems.find(n => isActive(n));
-		setSidebarItem((item?.id ?? "home") as SidebarFocusItem);
```
PR-Reviewer-Ant, line 102: Warning: the footer 'Project' item now has inconsistent activation behavior. Mouse click calls `openProjectPage()` and navigates to the internal `changelog` tab, but keyboard activation goes through `handleFooterKeydown(..., "github-link")`, which opens the external GitHub URL instead. That is a functional regression for keyboard users and makes the control behave differently depending on input method.

NicholaiVogel:
new feedback addressed.

fixes in latest commit:
- reverted global `setTab()` hash clobber for already-active tabs so deep links remain sticky by default
- moved root-normalization behavior to sidebar click path only (so clicking active top-level sidebar item intentionally returns to tab root)
- fixed keyboard parity for footer "Project": Enter/Space now matches click behavior and opens internal changelog tab instead of external GitHub

no visual changes.

validation: `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/api.ts
```diff
@@ -2332,9 +2332,28 @@ export interface ConstellationGraph {
 	dependencies: ConstellationDependency[];
 }
 
-export async function getConstellationOverlay(agentId = "default"): Promise<ConstellationGraph | null> {
+export function resolveAgentId(agentId?: string): string {
```
PR-Reviewer-Ant: The new implicit `resolveAgentId()` lookup can pull `agent_id` from query params, DOM state, or localStorage even when the caller did not explicitly request that scope. `MemoryTab` now calls `getConstellationOverlay()` with no argument, while the memory list itself is not scoped through the same path here. That can mix ontology data from a different agent into the selected-memory panel, which is both a correctness bug and an agent-scoping leak. This should be threaded explicitly from the page/store instead of falling back to ambient browser state.

packages/cli/dashboard/src/lib/components/tabs/MemoryTab.svelte
```diff
@@ -54,6 +61,111 @@ let display = $derived(
...
+	const data = graph;
+	const memId = selectedMemory.id;
+	const byId = new Map(data.entities.map((entity) => [entity.id, entity.name]));
+
+	return data.entities
+		.map((entity) => {
+			const aspects = entity.aspects.filter((aspect) =>
+				aspect.attributes.some((attr) => attr.memoryId === memId),
+			);
+			if (aspects.length === 0) return null;
```
PR-Reviewer-Ant, line 110: The new "Ontology Context" panel is described as showing entities/aspects/constraints/dependencies for the selected memory, but the dependency list is built from every dependency touching the matched entity (`sourceEntityId === entity.id || targetEntityId === entity.id`). Because `ConstellationDependency` has no memory-level filter here, selecting one memory can display unrelated dependencies from other memories involving the same entity. That makes the new panel misleading and does not fully implement the PR description.

NicholaiVogel:
addressed the newest feedback in a fresh pass.

### fixes
- removed ambient/implicit agent scope inference from overlay API helper
- now threads `agentId` explicitly from page -> tab loader -> cortex tab -> unified memory tab -> memory tab -> `getConstellationOverlay(agentId)`
- keeps scope deterministic and aligned with the current dashboard context
- tightened ontology dependency display for selected memory:
  - dependency rows now only show edges between entities that are both actually relevant to the selected memory
  - avoids showing unrelated cross-memory entity edges

no visual changes.

validation: `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -56,35 +47,96 @@ const HASH_ALIASES: ReadonlyMap<string, TabId> = new Map([
 	["memory/timeline", "cortex-memory"],
 	["memory/knowledge", "cortex-memory"],
 	["memory/memories", "cortex-memory"],
+	["ontology", "cortex-memory"],
+	["ontology/cortex", "cortex-memory"],
+	["ontology/constellation", "cortex-memory"],
+	["memory", "cortex-memory"],
```
PR-Reviewer-Ant, line 53: The legacy `#embeddings` route now canonicalizes to `#cortex-memory` instead of `#cortex-memory/constellation`. That breaks existing bookmarks/shared links that previously opened the constellation view, and it conflicts with the PR description's claim of hardened legacy hash canonicalization for Ontology subroutes. `#memory/constellation` is preserved, but the old single-tab `#embeddings` path is not.

packages/cli/dashboard/src/lib/components/tabs/UnifiedMemoryTab.svelte
```diff
@@ -0,0 +1,196 @@
...
+			>
+				CONSTELLATION
+			</Button>
+		</div>
+	</div>
+	<div class="unified-body">
+		{#if section === "cortex"}
+			<div class="unified-main">
+				<MemoryTab {memories} {agentId} embedded={true} />
+			</div>
```
PR-Reviewer-Ant, line 78: The new unified Ontology flow threads `agentId` into `MemoryTab`, but not into `EmbeddingsTab`. With `?agent_id=...` set, the Cortex list/context panel will scope to that agent while the Constellation view still falls back to its existing default data path. That creates a mixed-agent UI inside one page and violates the repo's scoping guardrail for user-data reads.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `84748425`

[Automated Review] Found two correctness issues in the new dashboard IA routing/scoping.
The PR largely matches the stated IA reshuffle, but two parts diverge from the compatibility/scoping goals in the description.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The broken legacy deep-link mapping is explicit in the changed hash alias/canonical tables, and the mixed agent scoping is directly visible in the new UnifiedMemoryTab wiring: MemoryTab receives agentId while EmbeddingsTab does not. I could not inspect the full downstream EmbeddingsTab data path here, so the scoping impact is inferred from the changed props and the repo's agent-scoping contract.

NicholaiVogel:
addressed the next round.

- legacy `#embeddings` now canonicalizes to `#cortex-memory/constellation` (bookmark compatibility restored)
- threaded explicit `agentId` into constellation path so Ontology and Constellation views use the same scope
- removed ambient/implicit scope lookup behavior in this flow (no DOM/localStorage-based agent switching)

kept visuals unchanged.

validation: `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/api.ts
```diff
@@ -2332,9 +2332,17 @@ export interface ConstellationGraph {
...
```
PR-Reviewer-Ant: The new `resolveAgentId()` helper silently falls back to `"default"` for the constellation overlay. This is a user-data query path, and the repo conventions explicitly call out agent scoping leaks as a recurring failure mode. If any caller forgets to thread `agentId`, the dashboard will render the default agent's ontology instead of failing closed, which can leak or mix data across agents.

packages/cli/dashboard/src/lib/components/app-sidebar.svelte
```diff
@@ -81,25 +82,25 @@ function openProjectPage(): void {
 }
 
 function isActive(item: NavItem): boolean {
-	if (item.group === "memory") return isMemoryGroup(nav.activeTab);
-	if (item.group === "engine") return isEngineGroup(nav.activeTab);
-	if (item.group === "cortex") return isCortexGroup(nav.activeTab);
 	return nav.activeTab === item.id;
 }
 
 function handleClick(item: NavItem): void {
```
PR-Reviewer-Ant, line 88: Re-clicking an already-active top-level item now forcibly rewrites the hash to `#${item.id}`. That strips valid subroutes like `#audit/logs` and `#cortex-memory/constellation`, so users lose their selected Audit/Ontology subsection just by activating the current sidebar item again. The PR description says deep-link behavior for these subroutes was hardened, but this handler regresses it by collapsing them back to the root route.

NicholaiVogel:
another feedback pass pushed in `be81da9b`.

what i changed:
- removed `resolveAgentId()` and its implicit `"default"` fallback in `api.ts`
- `getConstellationOverlay()` now requires an explicit `agentId` argument
- made `agentId` required in both `MemoryTab` and `EmbeddingsTab` so ontology + constellation stay in the same explicit scope path
- stopped active-sidebar re-click from rewriting hash to tab root, so deep subroutes remain sticky (`#audit/logs`, `#cortex-memory/constellation`)

validation:
- `cd packages/cli/dashboard && bun run check` ✅ (0 errors, warnings only)
- `bun run typecheck` ❌ still failing in unrelated baseline: `@signet/opencode-plugin` missing `node` type defs
- `bun run lint` ❌ still failing on existing repo-wide biome formatting debt outside this PR scope

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -93,23 +145,9 @@ export const nav = $state({
```
PR-Reviewer-Ant, line 145: The exported memory-group routing is internally inconsistent after this refactor: `VALID_TABS` no longer accepts `"memory"`, but `MEMORY_TABS` still contains only `"memory"`. Any remaining caller of `navigateToGroup("memory")` will now fail silently because `setTab("memory")` returns `false`. This PR claims the navigation stores were updated for the new IA, so leaving the group helper targeting a retired tab ID is a behavioral regression.

packages/cli/dashboard/src/lib/stores/focus.svelte.ts
```diff
@@ -37,22 +41,16 @@ export const SIDEBAR_ORDER: readonly SidebarFocusItem[] = [
 function tabToSidebarItem(tab: string): SidebarFocusItem {
```
PR-Reviewer-Ant, line 41: The new footer navigation introduces a Project button that activates `changelog`, but `tabToSidebarItem()` no longer maps `"changelog"` back to any sidebar/footer focus item. When the user opens Project, the active control is the footer button in `app-sidebar.svelte`, yet the roving tabindex state falls back to another item instead of `github-link`. That breaks keyboard focus recovery for the new footer IA.

NicholaiVogel:
new round handled in `d9225aa1`.

- fixed tab-group regression: `MEMORY_TABS` and `lastMemoryTab` now target `cortex-memory` (not retired `memory`)
- fixed footer focus recovery: `tabToSidebarItem("changelog")` now maps to `github-link` so Project tab returns focus to the correct footer control

validation:
- `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -31,23 +32,13 @@ export type TabId =
 const VALID_TABS: ReadonlySet<string> = new Set<TabId>([
```
PR-Reviewer-Ant, line 32: This change removes `memory`, `timeline`, `knowledge`, and `embeddings` from `VALID_TABS`, but the shared memory tab definitions were not updated. `MEMORY_TAB_ITEMS` in `page-headers.ts` still contains those ids, and `MemoryTab.svelte` / `EmbeddingsTab.svelte` still render that tab bar when `embedded === false`. Any path that renders those components non-embedded will now have dead tabs because `setTab()` returns `false` for every one of those ids. Given the PR claim that legacy deep-linking/navigation was hardened, these shared tab ids need to be brought back into sync or removed entirely.

NicholaiVogel:
addressed the new nav-sync feedback in `40731405`.

- hardened `setTab()` so legacy ids now resolve through `HASH_ALIASES` before reject
- this keeps shared/legacy tab bars from dead-clicking even though `VALID_TABS` only includes active IA destinations
- behavior now stays consistent with the hash canonicalization strategy already used in `initNavFromHash`

validation:
- `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/components/app-sidebar.svelte
```diff
@@ -81,25 +82,19 @@ function openProjectPage(): void {
...
-	} else {
-		setTab(item.id as TabId);
-	}
+	if (nav.activeTab === item.id) return;
+	setTab(item.id);
 }
 
 // Initialize sidebar focus on mount — derive from current active tab
 onMount(() => {
 	if (!focus.sidebarItem) {
```
PR-Reviewer-Ant, line 103: Direct loads into footer-owned destinations regress roving focus initialization. `onMount()` only looks through `navItems`, so opening the dashboard on `#settings` or `#changelog` falls back to `home` instead of the actual active footer item. That leaves the active control with the wrong `tabindex`/focus state until the user manually moves focus, which breaks the new navigation model for deep links.

docs/specs/planning/dashboard-information-architecture-refactor.md
```diff
@@ -156,3 +156,18 @@ drag-to-resize, transition animations.
 1. Whether Cortex tabs merge into new hierarchy or stay separate.
```
PR-Reviewer-Ant, line 156: This file now documents a 'shipped slice' while the spec header still says `status: planning`. The repo's mandatory workflow requires implementation to be validated against approved/completed spec state, so this PR is diverging from the stated process rather than keeping spec status in sync with shipped behavior. Given the routing and IA changes here, that drift makes future maintenance/review materially harder.

NicholaiVogel:
handled the new round in `b8b9df10`.

changes:
- fixed sidebar deep-link focus init for footer-owned tabs:
  - added `focusIdForTab()` in `app-sidebar.svelte`
  - on mount, sidebar focus now derives from `nav.activeTab` across both main nav and footer tabs (`settings`, `changelog` -> `github-link`)
- resolved planning/spec wording drift:
  - rewrote the planning doc section label/text so it no longer claims a "shipped slice" while status is still `planning`

validation:
- `cd packages/cli/dashboard && bun run check` ✅

packages/cli/dashboard/src/lib/components/cortex/TroubleshooterPanel.svelte
```diff
@@ -100,47 +100,44 @@
...
-		el?.scrollTo({ top: el.scrollHeight, behavior: "smooth" });
+		termRef?.scrollTo({ top: termRef.scrollHeight, behavior: "smooth" });
 	}
 
 	function appendLine(id: string, text: string): void {
 		const gs = panels[id];
 		if (!gs) return;
 		gs.lines = [...gs.lines, text];
+		const label = groupMap[id]?.title ?? id;
+		termLines = [...termLines, text ? `[${label}] ${text}` : ""];
```
PR-Reviewer-Ant, line 127: Blocking: the new unified terminal introduces a lost-update race. `appendLine()` rebuilds `termLines` with `termLines = [...termLines, ...]`, and different command groups can still run at the same time (`gs.running` is only per-group). If two runs append near-simultaneously, one write can overwrite the other group's pending output, so the audit trail becomes incomplete. This did not exist when each group rendered its own isolated log buffer.

packages/cli/dashboard/src/lib/stores/navigation.svelte.ts
```diff
@@ -56,64 +47,109 @@ const HASH_ALIASES: ReadonlyMap<string, TabId> = new Map([
 	["memory/timeline", "cortex-memory"],
```
PR-Reviewer-Ant, line 47: Warning: this PR claims to harden legacy hash canonicalization and deep-link behavior, but the redirect matrix here rewrites a large set of saved routes (`memory/*`, `ontology/*`, `cortex/*`, `engine/*`, `logs`, `pipeline`, etc.) without any regression tests in the diff. Given the repo's explicit guardrail on adding durable prevention for routing regressions, this is a fragile area to leave manual-only.

PR #365, opened 2026-03-27
fix(update): verify installed version after install

### LINKED ISSUE #322: fix(update): verify installed version after signet update install completes
## Summary

`signet update install` in `packages/cli/src/commands/update.ts` reports success based on the HTTP response from `POST /api/update/run` in `packages/daemon/src/update-system.ts`. The actual installed package version is never verified. A failed or cache-stale npm/bun install can report success, leaving the operator on the old version with no indication.

## Current Behavior

Install flow (`update.ts` lines ~86-154):
1. Fetch `/api/update/check?force=true` — gets target version
2. POST `/api/update/run` with `targetVersion`
3. Wait up to 120s
4. On 200 response: sync skills, re-register hooks, print "Update installed"

`installUpdate()` in `update-system.ts` resolves based on package manager exit code only. If npm exits 0 but didn't actually update (network issue, cache hit), the CLI reports success.

## Proposed Fix

After the package manager exits in `installUpdate()`, read `node_modules/signetai/package.json` and compare to `targetVersion`:

```typescript
const installedVersion = readInstalledVersion();
if (installedVersion !== targetVersion) {
  return {
    success: false,
    error: `Install exited 0 but version is ${installedVersion}, expected ${targetVersion}`
  };
}
```

The CLI `update.ts` already handles the response object — it just needs to act on `success: false` by printing an error instead of the success message.

## Why

False confidence is worse than a clear failure. One additional file read post-install surfaces the failure exactly where the operator is already looking.

aaf2tbz:
## Summary
- pin update installs to an exact target when `targetVersion` is provided (install `signetai@<targetVersion>` instead of unpinned `signetai`)
- verify installed package version from resolved global `package.json` after package-manager exit code 0
- return `success: false` when verification fails (missing global path, missing manifest, invalid/missing installed version, or mismatch)
- set `pendingRestartVersion` and `installedVersion` from verified installed version
- tighten `targetVersion` normalization to exact semver(+prerelease/+build) to avoid false mismatch on npm tags/ranges (`latest`, `1.2.x`, etc.)
- add behavioral regression coverage for `verifyInstalledVersion` and target normalization edge cases

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## PR Readiness (MANDATORY)

## Testing

- [ ] `bun run lint` passes
- [ ] N/A

## Validation
- `cd packages/daemon && bun test src/update-system.test.ts -t "Issue 322|verifyInstalledVersion|version parsing helpers"`

- [ ] Migration is idempotent

N/A — no schema/database migration.

## AI disclosure

## Notes
- Existing daemon-wide lint/typecheck debt remains outside this fix scope.
- Closes #322

### DIFF
--- a/packages/daemon/src/update-system.test.ts
+++ b/packages/daemon/src/update-system.test.ts
@@ -16,6 +16,7 @@ import {
 	categorizeUpdateError,
 	normalizeTargetVersion,
 	parseInstalledPackageVersion,
+	verifyInstalledVersion,
 } from "./update-system";
 
 const UPDATE_SYSTEM_SRC = readFileSync(
@@ -55,6 +56,59 @@ describe("Issue 322: verify installed version after update install", () => {
 			"Install exited cleanly but version is",
 		);
 		expect(UPDATE_SYSTEM_SRC).toContain("resolveGlobalPackagePath");
+	});
+});
+
+describe("verifyInstalledVersion", () => {
+	const noopResolver = (_family: "bun" | "npm" | "pnpm" | "yarn", _packageName: string) =>
+		undefined;
+
+	it("fails when global package path cannot be resolved", () => {
+		const result = verifyInstalledVersion("bun", "signetai", "0.78.1", {
+			resolveGlobalPackagePath: noopResolver,
+			existsSync: () => true,
+			readFileSync: (_path, _encoding) => '{"version":"0.78.1"}',
+		});
+		expect(result.ok).toBe(false);
+		if (!result.ok) {
+			expect(result.message).toContain("could not locate global package path");
+		}
+	});
+
+	it("fails when package.json is missing", () => {
+		const result = verifyInstalledVersion("bun", "signetai", "0.78.1", {
+			resolveGlobalPackagePath: (_family, _packageName) => "/tmp/signetai",
+			existsSync: () => false,
+			readFileSync: (_path, _encoding) => '{"version":"0.78.1"}',
+		});
+		expect(result.ok).toBe(false);
+		if (!result.ok) {
+			expect(result.message).toContain("package manifest missing");
+		}
+	});
+
+	it("fails when installed version does not match expected target", () => {
+		const result = verifyInstalledVersion("bun", "signetai", "0.78.1", {
+			resolveGlobalPackagePath: (_family, _packageName) => "/tmp/signetai",
+			existsSync: () => true,
+			readFileSync: (_path, _encoding) => '{"version":"0.78.0"}',
+		});
+		expect(result.ok).toBe(false);
+		if (!result.ok) {
+			expect(result.message).toContain("version is 0.78.0, expected 0.78.1");
+		}
+	});
+
+	it("succeeds and returns installed version when verification passes", () => {
+		const result = verifyInstalledVersion("bun", "signetai", "0.78.1", {
+			resolveGlobalPackagePath: (_family, _packageName) => "/tmp/signetai",
+			existsSync: () => true,
+			readFileSync: (_path, _encoding) => '{"version":"0.78.1"}',
+		});
+		expect(result.ok).toBe(true);
+		if (result.ok) {
+			expect(result.installedVersion).toBe("0.78.1");
+		}
 	});
 });
 
@@ -149,6 +203,8 @@ describe("version parsing helpers", () => {
 		expect(normalizeTargetVersion("V2.0.0-rc.1+build.7")).toBe(
 			"2.0.0-rc.1+build.7",
 		);
+		expect(normalizeTargetVersion("latest")).toBeNull();
+		expect(normalizeTargetVersion("1.2.x")).toBeNull();
 		expect(normalizeTargetVersion("")).toBeNull();
 		expect(normalizeTargetVersion("   ")).toBeNull();
 		expect(normalizeTargetVersion("--1.2.3")).toBeNull();
--- a/packages/daemon/src/update-system.test.ts
+++ b/packages/daemon/src/update-system.test.ts
@@ -99,6 +99,18 @@ describe("verifyInstalledVersion", () => {
 		}
 	});
 
+	it("fails when installed package.json version is not exact semver", () => {
+		const result = verifyInstalledVersion("bun", "signetai", null, {
+			resolveGlobalPackagePath: (_family, _packageName) => "/tmp/signetai",
+			existsSync: () => true,
+			readFileSync: (_path, _encoding) => '{"version":"latest"}',
+		});
+		expect(result.ok).toBe(false);
+		if (!result.ok) {
+			expect(result.message).toContain("installed package.json has no valid version");
+		}
+	});
+
 	it("succeeds and returns installed version when verification passes", () => {
 		const result = verifyInstalledVersion("bun", "signetai", "0.78.1", {
 			resolveGlobalPackagePath: (_family, _packageName) => "/tmp/signetai",
@@ -216,6 +228,8 @@ describe("version parsing helpers", () => {
 			"0.78.1",
 		);
 		expect(parseInstalledPackageVersion('{"name":"signetai","version":"   "}')).toBeNull();
+		expect(parseInstalledPackageVersion('{"name":"signetai","version":"latest"}')).toBeNull();
+		expect(parseInstalledPackageVersion('{"name":"signetai","version":"1.2.x"}')).toBeNull();
 		expect(parseInstalledPackageVersion('{"name":"signetai"}')).toBeNull();
 		expect(parseInstalledPackageVersion("not-json")).toBeNull();
 	});
--- a/packages/daemon/src/update-system.test.ts
+++ b/packages/daemon/src/update-system.test.ts
@@ -111,6 +111,21 @@ describe("verifyInstalledVersion", () => {
 		}
 	});
 
+	it("fails gracefully when manifest read throws", () => {
+		const result = verifyInstalledVersion("bun", "signetai", null, {
+			resolveGlobalPackagePath: (_family, _packageName) => "/tmp/signetai",
+			existsSync: () => true,
+			readFileSync: (_path, _encoding) => {
+				throw new Error("EACCES: permission denied");
+			},
+		});
+		expect(result.ok).toBe(false);
+		if (!result.ok) {
+			expect(result.message).toContain("failed to verify installed version");
+			expect(result.message).toContain("EACCES");
+		}
+	});
+
 	it("succeeds and returns installed version when verification passes", () => {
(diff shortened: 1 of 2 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(update): verify installed version after install` (commit `677210cc`) now and I'll follow up shortly with feedback.

packages/daemon/src/update-system.ts
```diff
@@ -497,10 +499,81 @@ export async function checkForUpdates(): Promise<UpdateInfo> {
 	return result;
 }
 
+export function normalizeTargetVersion(targetVersion: string | undefined): string | null {
```
PR-Reviewer-Ant, line 504: Blocking: `normalizeTargetVersion()` now accepts many non-exact npm specs (`latest`, `next`, `1.2.x`, etc.), but `verifyInstalledVersion()` later compares the installed `package.json` version with strict string equality against that original target. A valid install like `signetai@latest` will resolve to a concrete version (for example `0.78.1`) and then be reported as a failed update because `0.78.1 !== latest`. Either restrict `targetVersion` to exact semver only, or compare against the resolved version/range semantics instead of raw string equality.

packages/daemon/src/update-system.test.ts
```diff
@@ -22,19 +24,37 @@ const UPDATE_SYSTEM_SRC = readFileSync(
...
-	it("sets pendingRestartVersion unconditionally (no targetVersion guard)", () => {
-		// The old code: `if (targetVersion) { pendingRestartVersion = targetVersion; }`
-		// The fix: `pendingRestartVersion = targetVersion ?? "unknown";`
+describe("Bug 5: pendingRestartVersion is set only after successful verification", () => {
+	it("does not gate pendingRestartVersion on targetVersion", () => {
 		const hasOldGuard = /if\s*\(\s*targetVersion\s*\)\s*\{?\s*\n?\s*pendingRestartVersion\s*=/.test(
 			UPDATE_SYSTEM_SRC,
 		);
 		expect(hasOldGuard).toBe(false);
+	});
```
PR-Reviewer-Ant, line 34: Warning: the PR description says it adds regression coverage for the install pinning and post-install verification flow, but these new tests only inspect source text with `toContain()`/regexes. They do not execute `runUpdate()` or `verifyInstalledVersion()`, so they would still pass if the runtime behavior regressed while the same strings remained in the file. Given the repo’s guardrail that incident fixes must add durable regression tests, this should be replaced or supplemented with behavioral tests.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `677210cc`

[Automated Review] Found one blocking logic regression in version validation/verification and one test-coverage gap against the PR’s stated goals.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new `normalizeTargetVersion()` implementation accepts arbitrary alphanumeric package specs such as `latest` or `1.2.x`, while `verifyInstalledVersion()` later requires the installed package.json version to equal that exact input string. That mismatch is directly visible in the diff and produces false negatives after successful installs. The added tests also only assert source-code substrings instead of executing `runUpdate()` or `verifyInstalledVersion()`, so the PR’s claimed regression coverage is not actually demonstrated at runtime.

aaf2tbz:
Addressed the new review in commit `b81e2440`.

### ✅ 1) Blocking: `targetVersion` vs strict verification mismatch
I changed `normalizeTargetVersion()` to accept **exact semver only** (with optional prerelease/build metadata), after stripping optional leading `v`.

Accepted examples:
- `1.2.3`
- `v1.2.3`
- `1.2.3-rc.1+build.7`

Rejected examples (now invalid):
- `latest`
- `next`
- `1.2.x`

That removes the false-negative path where a tag/range installs a concrete version that cannot equal the raw tag/range string.

### ✅ 2) Regression coverage gap
I supplemented the previous structural checks with behavioral tests that execute the verification logic directly:
- `verifyInstalledVersion` failure cases:
  - unresolved global package path
  - missing package manifest
  - installed version mismatch
- `verifyInstalledVersion` success case
- target normalization edge cases (including rejected non-exact specs)

### Validation run
- `cd packages/daemon && bun test src/update-system.test.ts -t "Issue 322|verifyInstalledVersion|version parsing helpers"`

(8 passing tests in the targeted suite.)

packages/daemon/src/update-system.ts
```diff
@@ -497,10 +499,98 @@ export async function checkForUpdates(): Promise<UpdateInfo> {
...
+	}
+	return normalized;
+}
+
+export function parseInstalledPackageVersion(packageJsonContent: string): string | null {
+	try {
+		const parsed = JSON.parse(packageJsonContent) as { version?: unknown };
+		if (typeof parsed.version !== "string") return null;
+		const version = parsed.version.trim();
+		return version.length > 0 ? version : null;
```
PR-Reviewer-Ant: The new verification still does not implement the behavior claimed in the PR description for unpinned installs. `parseInstalledPackageVersion()` accepts any non-empty `version` string, so `runUpdate()` will report success and set `pendingRestartVersion` even if the installed global `package.json` contains an invalid version like `"workspace:*"` or other corrupted data. The description says verification should fail on an invalid/missing installed version; to satisfy that, this parser needs semver validation instead of a non-empty-string check.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `b81e2440`

[Automated Review] Found one blocking issue in installed-version verification that still diverges from the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description explicitly says verification should fail on an invalid installed version, but `parseInstalledPackageVersion()` only checks for a non-empty string and `verifyInstalledVersion()` treats that as success when no exact target version was requested. This is directly provable from the added code path in `update-system.ts` without needing runtime speculation.

aaf2tbz:
Addressed the latest blocking review in commit `d3eff553`.

### ✅ Fix: invalid installed versions now fail verification
`parseInstalledPackageVersion()` now validates the parsed `package.json` version as **exact semver** (same pattern used for normalized target version), not just non-empty string.

So if the installed manifest reports non-semver values (e.g. `latest`, `1.2.x`, malformed values), `verifyInstalledVersion()` now fails with:
- `installed package.json has no valid version ...`

This aligns runtime behavior with the PR contract (“invalid/missing installed version” is a hard failure).

### Regression coverage added
- `verifyInstalledVersion` now has a behavioral test for non-semver installed version (`"latest"`) -> failure
- `parseInstalledPackageVersion` now tests rejection of `"latest"` and `"1.2.x"`

### Validation
- `cd packages/daemon && bun test src/update-system.test.ts -t "Issue 322|verifyInstalledVersion|version parsing helpers"`
- Result: 9 passing tests (0 failing)

packages/daemon/src/update-system.ts
```diff
@@ -497,10 +501,99 @@ export async function checkForUpdates(): Promise<UpdateInfo> {
...
+		};
+	}
+
+	const packageJsonPath = join(packagePath, "package.json");
+	if (!deps.existsSync(packageJsonPath)) {
+		return {
+			ok: false,
+			message: `Update exited cleanly but package manifest missing at ${packageJsonPath}`,
+		};
+	}
```
PR-Reviewer-Ant: The new verification path can throw instead of returning a structured failure. `verifyInstalledVersion()` checks `existsSync()` and then immediately calls `readFileSync()` without a `try/catch`. If that read fails (TOCTOU deletion, permission error, transient filesystem issue), the exception escapes the `proc.on("close")` callback in `runUpdate()` before `resolve()` executes. That leaves the update flow hanging or crashes the daemon, which contradicts the PR contract of returning `success: false` on verification failure. Catch the read/parse failure inside `verifyInstalledVersion()` and convert it into an `{ ok: false, message }` result.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `d3eff553`

[Automated Review] Found one blocking failure path in the new post-install verification logic.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new `runUpdate()` success path now calls `verifyInstalledVersion()` inside the child-process `close` handler. `verifyInstalledVersion()` performs a raw `readFileSync()` with no error handling. If the global package manifest becomes unreadable between `existsSync()` and `readFileSync()` or the read fails for permissions/filesystem reasons, that exception escapes the event callback before `resolve()` runs. In this code shape, the update promise can hang or crash instead of returning the promised `success: false` result.

aaf2tbz:
Great catch — addressed in commit `9cf6ebd8`.

### ✅ Fix: verification is now fail-safe (no uncaught read exceptions)
I wrapped `verifyInstalledVersion()` in a `try/catch` so any unexpected filesystem/read/path exceptions now return a normal failure result instead of escaping the close-handler callback.

New failure message path:
- `Update exited cleanly but failed to verify installed version: <error>`

This guarantees the `/api/update/run` flow resolves with `success: false` instead of hanging/crashing if manifest reads fail between existence check and read.

### Regression coverage added
- `verifyInstalledVersion` behavioral test for thrown manifest read (`EACCES`) now asserts graceful failure + error message.

### Validation
- `cd packages/daemon && bun test src/update-system.test.ts -t "verifyInstalledVersion|Issue 322|version parsing helpers"`
- Result: 10 passing tests (0 failing)

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `9cf6ebd8`

[Automated Review] No new issues found in the current diff; ready for human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The incremental diff directly addresses the prior uncaught-exception path by wrapping `verifyInstalledVersion()` in a fail-safe `try/catch`, and the added test covers the new `EACCES` failure branch. I did not find a new bug in the changed lines themselves. Confidence is not high because the supporting implementation of `resolveGlobalPackagePath()` in `@signet/core` is not included here, so I could not independently validate that cross-package contract.

PR #367, opened 2026-03-27
fix(daemon): persist extraction fallback status and control (supersedes #348)

### LINKED ISSUE #320: feat(daemon): provider preflight check at startup + explicit fallback control

aaf2tbz:
## Summary
Supersedes #348 with the same functional scope on a clean, up-to-date branch.

This PR implements issue #320:
- persists extraction provider runtime status in `/api/status` (including degraded vs blocked state metadata)
- adds `memory.pipelineV2.extraction.fallbackProvider` (`ollama | none`) with strict validation
- dead-letters startup-pending extraction jobs when startup preflight blocks extraction
- blocks new remember-route extraction writes when extraction is startup-blocked
- includes daemon-rs parity for the status/config/dead-letter behavior touched by this fix

Closes #320.

## What went wrong in #348
The original PR accumulated many iterative review rounds, stale change requests, and branch drift (`BEHIND`) despite later fixes landing. This replacement PR resets review state on a clean branch while preserving the corrected implementation.

## Type
- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected
- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs/specs, daemon-rs parity

## PR Readiness (MANDATORY)
- [x] Lint/typecheck/tests pass locally for touched paths

## Testing
- [ ] `bun run lint` passes
- [ ] N/A

## Validation
- `bun test packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/extraction-fallback.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/runtime.test.ts`
- `cargo test -p signet-core config::tests::extraction_fallback_provider_loads_from_nested_or_flat_keys --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-daemon state::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-pipeline worker::tests --manifest-path packages/daemon-rs/Cargo.toml`

- [ ] Migration is idempotent

## AI disclosure


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `0fede3fd`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `0fede3fd`

[Automated Review by pr-reviewer] Found Rust-side config validation and error-handling issues that diverge from the PR's stated strict-validation/parity goals.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The Rust config loader now reads fallbackProvider via raw_string() and only rejects invalid string values, so non-string invalid values will silently fall back to defaults instead of being strictly rejected. The same code path also calls std::process::exit(1) from the shared signet-core config parser, which is directly visible in the diff and changes failure behavior for every crate consuming that parser.


- packages/daemon-rs/crates/signet-core/src/config.rs:1372 - This does not fully implement the PR's claimed "strict validation". Because the Rust side reads `fallbackProvider` only through `raw_string(...)`, values with the wrong YAML type (for example `fallbackProvider: 1` or `fallbackProvider: []`) are treated as absent and silently fall back to `ollama` instead of failing. The JS daemon rejects any non-`ollama`/`none` value, so daemon-rs parity is incomplete and an invalid config can still re-enable silent fallback on the Rust path.
- packages/daemon-rs/crates/signet-core/src/config.rs:1388 - Calling `std::process::exit(1)` from the shared `signet-core` config parser is a risky behavior change. This parser is no longer recoverable for any non-daemon consumer (tests, CLI, future tools), and callers cannot surface a structured validation error the way the JS side does. Please return an error from parsing and let the daemon binary decide whether to terminate.

aaf2tbz:
Addressed the latest requested changes in commit `feea774b`:

- daemon-rs now strictly rejects invalid extraction `fallbackProvider` values for both wrong value and wrong type (e.g. numeric/array), matching JS strictness
- removed `std::process::exit(1)` from shared `signet-core` config parsing
- parser now returns structured errors, and daemon binary now decides termination via `DaemonConfig::try_from_env()`
- added regression tests for invalid string and non-string `fallbackProvider` values

Validation rerun:
- `cargo test -p signet-core config::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-daemon state::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-pipeline worker::tests --manifest-path packages/daemon-rs/Cargo.toml`

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -176,7 +176,41 @@ pub async fn remember(
         .await;
 
     match result {
-        Ok(val) => (StatusCode::OK, Json(val)).into_response(),
+        Ok(val) => {
```
PR-Reviewer-Ant: This does not fully achieve the PR's claimed behavior of blocking new remember-route extraction writes when startup preflight has blocked extraction. The memory creation has already committed before this best-effort follow-up write runs, and on error you only log and still return success. That can leave a newly created memory with its original extraction state and no dead-lettered job, which is a data-consistency gap in the blocked path. The dead-letter/update needs to be part of the same durable write path as the create or the handler should fail instead of silently accepting the partial state.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -176,7 +176,41 @@ pub async fn remember(
...
+                        let reason = state
+                            .extraction_block_reason()
+                            .await
+                            .unwrap_or_else(|| "Extraction provider unavailable".into());
+                        if let Err(e) = state
+                            .pool
+                            .write(Priority::Low, move |conn| {
+                                conn.execute(
+                                    "UPDATE memories SET extraction_status = 'failed' WHERE id = ?1",
+                                    rusqlite::params![id],
```
PR-Reviewer-Ant: The inserted dead-letter row hardcodes `max_attempts = 3`. This PR also adds configurable worker retry limits, so blocked remember-route dead letters will now record incorrect job metadata whenever `memory.pipelineV2.worker.maxRetries` is not 3. That can skew repair/requeue behavior and breaks the daemon-rs parity claim for the new config surface.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `d951472e`

[Automated Review by pr-reviewer] Found two daemon-rs correctness issues in the new blocked-extraction handling path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the added Rust code. In `routes/write.rs`, the blocked-extraction dead-letter write is performed as a best-effort second DB operation after the memory create already succeeded, so a failure leaves the new memory in the pre-existing state instead of actually blocking extraction writes. The same block also hardcodes `max_attempts = 3`, which diverges from the newly configurable worker retry limit and can corrupt job metadata used by repair/requeue flows.

aaf2tbz:
Addressed the newest requested correctness issues in commit `d11ddf1b`:

- moved blocked-extraction remember handling into the same DB transaction as ingest (`write_tx`), so the memory write and dead-letter side effects are atomic
- removed best-effort secondary write path that could previously leave partially-applied blocked state
- switched dead-letter `max_attempts` metadata from hardcoded `3` to configured worker `max_retries` (clamped to at least 1)
- added a regression test validating blocked dead-letter writes set memory failed + preserve configured `max_attempts`

Validation rerun:
- `cargo test -p signet-daemon routes::write::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-core config::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-daemon state::tests --manifest-path packages/daemon-rs/Cargo.toml`
- `cargo test -p signet-pipeline worker::tests --manifest-path packages/daemon-rs/Cargo.toml`

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -152,16 +155,34 @@ async fn main() -> anyhow::Result<()> {
     };
```
PR-Reviewer-Ant, line 155: Blocking: this creates a new `extraction_worker_stats` handle for `AppState`, but `signet_pipeline::worker::start()` now allocates its own separate stats handle internally. Unless the daemon wires `WorkerHandle::stats_handle()` back into `AppState`, `/api/status` will read a disconnected snapshot that never reflects the real worker's overload/running state. That breaks the PR's daemon-rs parity claim for status visibility.

docs/CONFIGURATION.md
```diff
@@ -364,9 +364,13 @@ control.
 | `pollMs` | `2000` | 100-60000 ms | How often the worker polls for pending jobs |
```
PR-Reviewer-Ant, line 364: The PR adds and strictly validates `memory.pipelineV2.extraction.fallbackProvider`, but the extraction config reference in this doc still does not describe that field or its allowed values. That diverges from the PR description ('adds ... fallbackProvider with strict validation') and from the repo rule that behavior/config changes update docs in the same PR.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `d11ddf1b`

[Automated Review] Found one blocking parity bug and one documentation/config contract gap.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The daemon-rs status bug is directly visible in the diff: main.rs constructs and stores one fresh runtime-stats handle in AppState, while signet-pipeline::worker::start() constructs a different handle internally. That means /api/status in daemon-rs cannot report live worker overload state as claimed by the PR. The docs gap is also directly visible in the changed configuration docs: the new fallbackProvider setting is implemented and strictly validated, but the extraction config reference shown in this PR still does not document it.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -152,16 +155,34 @@ async fn main() -> anyhow::Result<()> {
...
     let auth_admin_limiter = AuthRateLimiter::from_rules(&merge_rate_limits(&config));
 
+    let extraction_worker_stats = config
```
PR-Reviewer-Ant: The new `extraction_worker_stats` handle is created independently here and stored on `AppState`, but `signet_pipeline::worker::start()` creates and updates its own separate stats handle. As a result, `/api/status` will read from an unowned snapshot that never reflects the real worker's `running`/`overloaded` state. This breaks the PR's claimed daemon-rs parity for status surfaces and can mislead operators into thinking load-shedding is inactive. The status route needs the actual worker-owned handle (for example from `WorkerHandle::stats_handle()`) rather than a parallel one.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `edf3187e`

[Automated Review] Found one blocking parity bug in the daemon-rs status implementation.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The Rust daemon now advertises extraction worker runtime fields in /api/status, but the state is populated from a fresh stats handle created in main.rs instead of the handle owned and updated by the worker loop in signet-pipeline. That makes the reported running/overload fields stale or permanently false, which directly contradicts the PR's daemon-rs parity claim.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `60d78f67`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -152,16 +155,26 @@ async fn main() -> anyhow::Result<()> {
     };
     let auth_admin_limiter = AuthRateLimiter::from_rules(&merge_rate_limits(&config));
 
+    // Do not create a detached runtime-stats handle in daemon state.
```
PR-Reviewer-Ant: Blocking: this PR claims daemon-rs parity for the new extraction status surface, but the new runtime-stats handle is hardcoded to `None` here and I don't see any later code path that installs the worker-owned stats handle into `AppState`. As a result, `status()` will always take the fallback branch and report `pipeline.extraction.running: false` with no live overload data, even when the Rust worker is actually running. That is a behavioral mismatch with the JS daemon and with the PR description's parity claim.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `60d78f67`

[Automated Review] Found one blocking daemon-rs parity regression in the new status surface.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The Rust daemon's new `/api/status` worker fields are wired from `AppState.extraction_worker_stats`, but `main.rs` now explicitly initializes that field to `None` and the diff shows no later assignment path. That makes the new status surface permanently report the fallback defaults instead of live worker state, which directly contradicts the PR's claimed daemon-rs parity for status behavior.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `09aad616`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -152,16 +155,26 @@ async fn main() -> anyhow::Result<()> {
     };
```
PR-Reviewer-Ant, line 155: The new Rust `/api/status` worker fields are not actually wired to a live worker. `AppState` is initialized with `extraction_worker_stats: None`, and the status route only reads runtime stats when that handle is present. As written, daemon-rs will always report `pipeline.extraction.running = false` with no live overload/countdown data, so the PR's claimed daemon-rs parity for status is not achieved.

packages/daemon-rs/crates/signet-daemon/src/routes/pipeline.rs
```diff
@@ -194,6 +194,23 @@ async fn apply_pause_state(state: &AppState, paused: bool) {
     state.pipeline_paused.store(paused, Ordering::SeqCst);
     let next = if paused { None } else { build_embedding(state) };
     *state.embedding.write().await = next;
+
+    // Update extraction runtime state on pause/resume transitions,
```
PR-Reviewer-Ant, line 198: On pause/resume, the Rust path only changes `extraction_state.status`. It does not update `effective`, `fallback_applied`, `reason`, or `since`. That means `/api/status` can report `status: "paused"` while still showing the previously active provider in `effective`, which is inconsistent with the runtime being paused and diverges from the JS contract this PR says it matches.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `09aad616`

[Automated Review] Found two correctness issues in the daemon-rs parity path.
The JS side mostly matches the PR description, but the Rust implementation does not fully achieve the claimed status/runtime parity.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly provable from the diff. In `packages/daemon-rs/crates/signet-daemon/src/main.rs`, the app state is constructed with `extraction_worker_stats: None`, while `/api/status` only reports live extraction worker state when that handle exists, so the new Rust status fields cannot reflect the actual worker runtime. In `packages/daemon-rs/crates/signet-daemon/src/routes/pipeline.rs`, pause/resume only mutates `status`, leaving `effective` and related runtime-resolution fields stale, which makes the new status surface inconsistent with the paused state it is supposed to report.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `a9c9fdaa`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
PR-Reviewer-Ant, line 155: The PR description promises daemon-rs parity for the new `/api/status.pipeline.extraction` status surface, but this code hardcodes `extraction_worker_stats` to `None`. Because `status()` only emits `pipeline.extraction` when that handle exists, daemon-rs will keep returning `null` for the new worker fields instead of the documented `running/overloaded/loadPerCpu/...` object. The added test `status_omits_worker_runtime_fields_without_live_stats_handle` confirms that behavior, so parity is still missing.

packages/daemon-rs/crates/signet-daemon/src/routes/pipeline.rs
```diff
@@ -194,6 +194,30 @@ async fn apply_pause_state(state: &AppState, paused: bool) {
     state.pipeline_paused.store(paused, Ordering::SeqCst);
     let next = if paused { None } else { build_embedding(state) };
     *state.embedding.write().await = next;
+
```
PR-Reviewer-Ant, line 197: Pausing now clears a previously `blocked` extraction state to `paused`, which means `remember` stops treating extraction as blocked and will accept new writes during the pause. On resume, `resume_extraction_check()` re-blocks extraction without dead-lettering pending jobs (`dead_letter_on_blocked = false`). If the provider is still unavailable, those memories stay queued indefinitely again, which is exactly the failure mode this PR says it fixes for blocked extraction.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `a9c9fdaa`

[Automated Review] Found two correctness issues: daemon-rs does not actually expose the new extraction worker status contract, and the new pause/resume state handling can reintroduce silently pending extraction jobs after a blocked startup.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The first issue is directly provable from the diff: `main.rs` initializes `extraction_worker_stats` to `None`, `status()` only emits `pipeline.extraction` when that handle exists, and the added test explicitly asserts the field is null. The second issue is also traceable in the diff: pausing overwrites a blocked extraction state to `paused`, `remember` only blocks when `status == "blocked"`, and resume re-checks availability with `dead_letter_on_blocked = false`, leaving newly queued jobs pending if the provider is still unavailable.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `18ab82bc`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -152,16 +155,35 @@ async fn main() -> anyhow::Result<()> {
...
```
PR-Reviewer-Ant, line 155: The new `/api/status` worker fields are not connected to the live extraction worker. This creates a brand new `new_runtime_stats_handle(...)` and stores it in `AppState`, but `signet_pipeline::worker::start(...)` also creates its own internal stats handle. Unless the startup path passes the worker's actual handle into `AppState`, `pipeline.extraction.running/overloaded/...` will stay at defaults and the daemon-rs parity claim for status observability is false.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +617,284 @@ async fn shutdown_signal() {
...
+    if !pipeline.enabled || extraction.provider == "none" || pipeline.paused || state.pipeline_paused() {
+        return;
+    }
+
+    let provider = extraction.provider.as_str();
+    let fallback_provider = extraction.fallback_provider.as_str();
+    let now = chrono::Utc::now().to_rfc3339();
+
+    // Check provider availability
+    let available = match provider {
```
PR-Reviewer-Ant, line 808: This preflight marks `anthropic` and `openrouter` as available whenever the corresponding env var is non-empty. That is weaker than the JS implementation added in this PR, which performs `llmProvider.available()` and can persist `degraded`/`blocked` when the remote endpoint is unreachable or the credential is invalid. As written, daemon-rs can report extraction `active` even though startup preflight should have blocked or degraded it.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `18ab82bc`

[Automated Review] Found two blocking daemon-rs parity/correctness issues in the new extraction status implementation.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly provable from the diff. In `packages/daemon-rs/crates/signet-daemon/src/main.rs`, the status surface is wired to a fresh `new_runtime_stats_handle(...)` instead of the worker's live handle, so `/api/status` cannot report real extraction worker activity. In the same file, the startup preflight treats Anthropic/OpenRouter as available solely from env-var presence, unlike the JS path which calls `llmProvider.available()` and can mark the provider degraded/blocked when the backend is unreachable or the key is invalid.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `fe764d27`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -11397,17 +11445,34 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 	);
```
PR-Reviewer-Ant, line 11445: The PR description says startup-blocked extraction now dead-letters queued extraction jobs, but this JS implementation never does that. `deadLetterPendingExtractionJobs` is imported earlier in the file and `queueExtractionJob()` only handles new remember-route writes after startup. In the shown startup-resolution path, `markExtractionUnavailable(...)` updates runtime status but does not dead-letter already-pending jobs, so old `pending` work can still sit silently in the queue and the implementation does not meet the stated contract.

packages/daemon-rs/crates/signet-daemon/src/routes/pipeline.rs
```diff
@@ -194,6 +194,30 @@ async fn apply_pause_state(state: &AppState, paused: bool) {
     state.pipeline_paused.store(paused, Ordering::SeqCst);
```
PR-Reviewer-Ant, line 194: Pause/resume only updates flags and `extraction_state`; it does not stop or restart the actual extraction worker started in `main.rs`. Because the worker handle is kept only in `main.rs` and never wired into `AppState`, `/pause` can report `paused` while the background worker keeps polling and processing extraction jobs. That is a behavioral regression and breaks the pause contract.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `fe764d27`

[Automated Review] Found two blocking correctness issues: the JS daemon does not actually dead-letter startup-pending extraction jobs as claimed, and daemon-rs pause/resume only updates status metadata without stopping the live extraction worker.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The JS diff imports `deadLetterPendingExtractionJobs` in `packages/daemon/src/daemon.ts` but the new startup-resolution logic shown in this PR only calls `queueExtractionJob()` for new remember writes; there is no added invocation that drains already-pending jobs, so the implementation diverges from the PR description. In daemon-rs, `main.rs` starts a worker handle once, but `routes/pipeline.rs` pause/resume only mutates `pipeline_paused`, embedding state, and extraction metadata; no handle is stopped or restarted there, so paused mode can still keep processing jobs.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `b58afec1`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `b58afec1`

[Automated Review] Found two blocking daemon-rs fallback regressions; the Rust parity implementation does not actually honor the fallback contract described in the PR.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The failures are directly visible in the added daemon-rs code paths. In `start_extraction_worker`, the worker uses `pipeline.extraction.model` and `pipeline.extraction.endpoint` even after preflight has switched the effective provider to `ollama`. In `extraction_probe`, the fallback health check probes Ollama using `extraction.endpoint`, which is the primary provider endpoint for Anthropic/OpenRouter/OpenCode configs, so fallback resolution can be marked blocked incorrectly. Both issues are provable from the diff without needing runtime traces.


- packages/daemon-rs/crates/signet-daemon/src/main.rs:1931 - When preflight degrades extraction to `ollama`, the worker still starts with `pipeline.extraction.model` and `pipeline.extraction.endpoint` from the originally configured provider. For a config like `provider: claude-code` this will report `effective: ollama` in `/api/status` but then try to call Ollama with the Claude model/endpoint, so the advertised fallback is not actually usable. The Rust parity path needs the same fallback model/base-URL resolution that the JS daemon applies before starting the worker.
- packages/daemon-rs/crates/signet-daemon/src/main.rs:2041 - The new Ollama fallback probe reuses `extraction.endpoint` from the primary provider (`let ollama_fallback_endpoint = extraction.endpoint.as_deref()`). That breaks the contract for any config where the primary provider has its own endpoint override, because the fallback health check will probe Anthropic/OpenRouter/OpenCode instead of the Ollama base URL and incorrectly mark extraction as `blocked`. This diverges from the PR description's 'fallbackProvider: ollama' behavior.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `f6caa799`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +612,502 @@ async fn shutdown_signal() {
...
+        .as_ref()
+        .and_then(|m| m.pipeline_v2.as_ref())
+    else {
+        return false;
+    };
+
+    if !pipeline.enabled || pipeline.paused || state.pipeline_paused() {
+        return false;
+    }
+
```
PR-Reviewer-Ant, line 685: Blocking: the new daemon-rs status/preflight path can report extraction as `active` for configured providers such as `claude-code`, `codex`, `opencode`, or `openrouter`, but `start_extraction_worker()` only starts a worker for `ollama` or `anthropic`. In that state, remember writes are not blocked, jobs can still be queued, and nothing in daemon-rs will drain them. That diverges from the PR description's parity claim and recreates the 'silent pending extraction' failure mode this PR is trying to eliminate.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f6caa799`

[Automated Review] Found one blocking daemon-rs parity issue and one documentation drift issue.
The JS side appears to implement the stated fallback-status behavior, but the Rust parity path still has a gap that can leave extraction work silently stuck pending under non-ollama providers.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The Rust diff shows `extraction_probe()` marking providers like `claude-code`, `codex`, `opencode`, and `openrouter` as available/active, while `start_extraction_worker()` only starts a worker for `ollama` or `anthropic`. That combination is directly visible in the changed code and is enough to show a parity/correctness gap, although I did not run a live daemon to demonstrate a stuck queue.


- docs/CONFIGURATION.md:361 - The PR description says this adds `memory.pipelineV2.extraction.fallbackProvider` with strict validation, but the configuration docs in this section still do not document that setting or its allowed values. Given this is a new persisted config surface, the docs currently diverge from the shipped behavior.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `98b9a59f`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -161,6 +204,17 @@ pub async fn remember(
                 },
             )?;
```
PR-Reviewer-Ant, line 211: This does not actually block new remember-route extraction writes. `transactions::ingest(...)` runs first, so any normal queued extraction job created there is still left in place; `dead_letter_blocked_extraction_memory()` only adds an extra dead job and flips `memories.extraction_status` to `failed`. That leaves contradictory job state for the same memory and can still let the worker pick up the original pending job later. The blocked path needs to cancel/update the job created by ingest rather than append a second record.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `98b9a59f`

[Automated Review] Found two blocking correctness issues in the new daemon-rs fallback path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly supported by the diff. In daemon-rs `remember`, the new blocked-extraction helper inserts a dead job after `transactions::ingest(...)` but does not update or remove the queued extraction job that ingest already created. In daemon-rs config loading, `load_manifest()` now returns `Ok(None)` for any `read_to_string` failure, so unreadable `agent.yaml` still falls back to defaults despite the PR's new strict-validation startup path.


- packages/daemon-rs/crates/signet-core/src/config.rs:1450 - The new `try_from_env()` path is advertised as strict validation, but `load_manifest()` still treats every `read_to_string` error as "manifest missing" (`Err(_) => return Ok(None)`). If `agent.yaml` exists but is unreadable due to permissions, IO errors, or filesystem problems, daemon-rs will silently boot with defaults instead of failing fast. That can bypass operator-configured `fallbackProvider: none` and other safety settings, which is the opposite of the PR's stated control/visibility goal.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `e3a37045`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +612,518 @@ async fn shutdown_signal() {
...
+        overload_backoff_ms: pipeline.worker.overload_backoff_ms,
+        extraction_timeout_ms: pipeline.extraction.timeout,
+        extraction_max_tokens: 4096,
+        min_confidence: pipeline.extraction.min_confidence,
+        shadow_mode: pipeline.shadow_mode,
+        graph_enabled: pipeline.graph.enabled,
+        structural_enabled: pipeline.structural.enabled,
+    };
+
+    let handle = signet_pipeline::worker::start(
```
PR-Reviewer-Ant, line 752: Blocking: daemon-rs now force-treats every extraction provider except `ollama` and `anthropic` as unavailable by routing successful preflight results through `worker_supports_extraction_provider()`. For configs using `claude-code`, `codex`, `opencode`, or `openrouter`, this changes behavior from 'configured provider is valid' to 'startup degraded/blocked', and with `fallbackProvider: none` it will dead-letter pending extraction jobs. That is a breaking behavioral change not described by this PR's scope.

docs/specs/INDEX.md
```diff
@@ -655,6 +655,7 @@ Legend:
 | `signet-runtime` | approved | `docs/specs/approved/signet-runtime.md` | `memory-pipeline-v2` | - | |
 | `pipeline-pause-control` | complete | `docs/specs/complete/pipeline-pause-control.md` | `memory-pipeline-v2` | - | CLI and dashboard pause/resume shipped on top of live daemon pause/resume API, paused-mode observability, and local Ollama unload |
 | `daemon-startup-responsiveness` | planning | `docs/specs/planning/daemon-startup-responsiveness.md` | `memory-pipeline-v2` | - | Incident-driven stub for issue #331: keep /health responsive during large-db startup recovery and recover stale managed daemon processes cleanly |
+| `daemon-provider-fallback-status` | planning | `docs/specs/planning/daemon-provider-fallback-status.md` | `memory-pipeline-v2` | - | Issue #320: persist extraction fallback/block state in status surfaces and add explicit fallbackProvider control |
```
PR-Reviewer-Ant: The PR description says this issue is implemented and closes #320, but the new spec entry remains `planning` here (and in `docs/specs/dependencies.yaml`). Given the repo's mandatory spec workflow, this leaves shipped behavior out of sync with the spec system.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `e3a37045`

[Automated Review] Found one blocking daemon-rs regression and one spec/process drift issue.

The PR mostly matches the stated goal for status surfacing, fallback control, and dead-lettering, but the daemon-rs parity implementation introduces a new behavioral restriction that is outside the PR description: daemon-rs now treats every extraction provider except `ollama` and `anthropic` as unavailable at startup, even when the configured provider is otherwise valid and installed. That can flip runtime state to `blocked`/`degraded` and dead-letter extraction work for configurations that previously remained valid in config.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The daemon-rs regression is directly visible in the new `worker_supports_extraction_provider` gate and its use inside `extraction_probe`, where supported config providers like `claude-code`, `codex`, `opencode`, and `openrouter` are force-marked unavailable regardless of actual provider availability. I did not run the branch, so the impact assessment relies on the changed control flow rather than a live repro.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status and control (supersedes #348)` (commit `78c3267d`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +612,506 @@ async fn shutdown_signal() {
...
+    }
+
+    let Some(pipeline) = state
+        .config
+        .manifest
+        .memory
+        .as_ref()
+        .and_then(|m| m.pipeline_v2.as_ref())
+    else {
+        return false;
```
PR-Reviewer-Ant, line 679: The new runtime-status contract can become internally inconsistent here. `extraction_probe()` marks the provider `active` as soon as the configured provider is reachable, but `start_extraction_worker()` immediately refuses to start unless the effective provider is `ollama` or `anthropic`. With a reachable `claude-code`/`codex`/`opencode`/`openrouter` config, `/api/status` will now report extraction as active even though no worker exists to process queued extraction jobs. That diverges from the PR's stated daemon-rs parity goal and can strand work silently. The status should reflect 'blocked'/'disabled' for unsupported providers, or the worker support matrix needs to match the advertised active states.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `78c3267d`

[Automated Review] Found a daemon-rs parity bug that can report extraction as active while no worker can actually run that provider.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new daemon-rs startup path sets `providerResolution.extraction.status = active` whenever the configured provider passes preflight, but `start_extraction_worker()` only supports `ollama` and `anthropic`. For providers like `claude-code`, `codex`, `opencode`, and `openrouter`, the status can now say extraction is active even though the worker refuses to start and queued extraction jobs will never drain. This is directly visible in `packages/daemon-rs/crates/signet-daemon/src/main.rs` from the added `worker_supports_extraction_provider()` gate plus the active-state branch in `extraction_probe()`.

aaf2tbz:
Superseded by #373. Closing this PR in favor of the clean branch + single-commit replacement.

PR #368, opened 2026-03-27
feat(pipeline): add command extraction provider for summary worker control-plane path

### LINKED ISSUE #334: Add command extraction provider for custom pipeline integration
## Problem

Users who want to run custom extraction pipelines (e.g., specialized prompts, multi-step processing, consolidation, or synthesis) currently have two options:
1. Fork SigNet and modify the extraction code
2. Add a second SessionEnd hook in the harness config

Option 1 is expensive to maintain. Option 2 works but bypasses SigNet's job queue, retry logic, significance assessment, and dashboard telemetry — the custom pipeline runs blind, outside SigNet's control plane.

## Proposal

Add `"command"` as a valid `extraction.provider` value with a companion `extraction.command` config field:

```yaml
memory:
  pipelineV2:
    extraction:
      provider: "command"
      command: "node /path/to/my-extractor.js --transcript $TRANSCRIPT --session-id $SESSION_KEY --project $PROJECT"
```

When the summary worker picks up a job and the provider is `command`:
1. Write the transcript to a temp file
2. Shell-exec the configured command with variable substitution (`$TRANSCRIPT`, `$SESSION_KEY`, `$PROJECT`)
3. The command writes memories directly to `memories.db` (same as any external writer)
4. SigNet's embedding tracker picks up the new rows via `content_hash`
5. Mark the job complete (or failed if non-zero exit)

## Why this is better than a second hook

- Jobs go through SigNet's queue with retry, backoff, and max_attempts
- Significance assessment still applies (trivial sessions are skipped)
- Dashboard shows extraction status, pending count, errors
- Single SessionEnd hook — no ordering/timing issues between two hooks
- The command inherits `SIGNET_PATH` so it knows where to write

## Implementation notes

- ~20 lines in `startSummaryWorker.tick2()` — branch before `processJob()` when provider is `command`
- Add `"command"` to `VALID_PROVIDERS` set
- Pass `command` field through in extraction config parsing
- Use `Bun.spawnSync` (or `child_process.execFileSync` for Node compat)
- Temp file cleanup in `finally` block

## Use case

I'm running a custom SigNet plugin — multi-category extraction with Jaccard dedup, activation pruning, and multi-file identity synthesis. Currently patching `daemon.js` locally, which breaks on `npm update`.

aaf2tbz:
## Summary

Implements issue #334 by adding a first-class `command` extraction provider that runs inside Signet's summary-worker control-plane path.

### What changed

- Added `"command"` as a valid extraction provider in pipeline config.
- Added structured extraction command config:
  - `memory.pipelineV2.extraction.command.bin`
  - `memory.pipelineV2.extraction.command.args[]`
  - optional `cwd`
  - optional `env`
- Added parsing support in daemon memory config, including legacy flat `extractionCommand` string parsing into argv.
- Added summary worker command execution path (`runSummaryCommandProvider`) when `extraction.provider === "command"`:
  - writes transcript to temp file
  - token substitution for `$TRANSCRIPT`/`$TRANSCRIPT_PATH`, `$SESSION_KEY`, `$PROJECT`, `$SIGNET_PATH`
  - executes via argv-safe `spawn` (no shell)
  - timeout clamp, non-zero exit propagation, and temp cleanup in `finally`
- Preserved significance gating and existing queue/retry/max-attempt behavior in the same job lifecycle.
- Added command-stage checkpointing to preserve downstream lifecycle on retries:
  - after command extraction succeeds, summary_jobs.result is marked as `command-stage-complete`
  - retries skip re-running external command side effects for that job
  - retries also skip significance re-evaluation once command stage is complete, so DAG/continuity/synthesis follow-up cannot be permanently bypassed by post-command DB mutations
- Added fail-fast validation for config mistakes:
  - reject `synthesis.provider: command`
  - reject `extraction.provider: command` when `extraction.command` is missing/invalid
- Removed command stdout/stderr success logging to reduce transcript-data exposure risk in logs.
- Added Rust config parity for extraction command config parsing.
- Updated docs in `docs/CONFIGURATION.md` and `docs/PIPELINE.md`.

## Why

This keeps custom extractors inside Signet’s queue/control plane (significance gate, retries, status telemetry) instead of requiring a second blind hook.

Closes #334.

## Validation

- `cd packages/daemon && bun test src/memory-config.test.ts src/pipeline/summary-worker.test.ts`
- `cd packages/daemon-rs && cargo test -p signet-core extraction_command_provider_parses_command_block -- --nocapture`
- `bunx biome check docs/CONFIGURATION.md docs/PIPELINE.md packages/core/src/types.ts packages/daemon/src/memory-config.ts packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts`

## PR Readiness Checklist

### DIFF
--- a/packages/core/src/pipeline-providers.ts
+++ b/packages/core/src/pipeline-providers.ts
@@ -6,6 +6,7 @@ export const PIPELINE_PROVIDER_CHOICES = [
 	"opencode",
 	"anthropic",
 	"openrouter",
+	"command",
 ] as const;
 
 export type PipelineProviderChoice = (typeof PIPELINE_PROVIDER_CHOICES)[number];
@@ -20,6 +21,7 @@ const MODEL_DEFAULTS = {
 	opencode: "anthropic/claude-haiku-4-5-20251001",
 	anthropic: "haiku",
 	openrouter: "openai/gpt-4o-mini",
+	command: "",
 } as const satisfies Record<PipelineProviderChoice, string>;
 
 const PIPELINE_PROVIDER_SET = new Set<string>(PIPELINE_PROVIDER_CHOICES);
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -180,13 +180,14 @@ fn raw_u64(value: &serde_yml::Value, key: &str) -> Option<u64> {
 fn is_pipeline_provider(value: &str) -> bool {
     matches!(
         value,
-        "none" | "ollama" | "claude-code" | "opencode" | "codex" | "anthropic" | "openrouter"
+        "none" | "ollama" | "claude-code" | "opencode" | "codex" | "anthropic" | "openrouter" | "command"
     )
 }
 
 fn default_pipeline_model(provider: &str) -> &'static str {
     match provider {
         "none" => "",
+        "command" => "",
         "claude-code" | "anthropic" => "haiku",
         "codex" => "gpt-5-codex-mini",
         "opencode" => "anthropic/claude-haiku-4-5-20251001",
@@ -422,6 +423,56 @@ memory:
         assert_eq!(pipeline.synthesis.provider, "codex");
         assert_eq!(pipeline.synthesis.model, "gpt-5-codex-mini");
     }
+
+    #[test]
+    fn extraction_command_provider_parses_command_block() {
+        let manifest = parse_manifest(
+            r#"
+memory:
+  pipelineV2:
+    extraction:
+      provider: command
+      command:
+        bin: node
+        args:
+          - script.mjs
+          - --transcript
+          - $TRANSCRIPT
+        cwd: /tmp/signet
+        env:
+          SIGNET_MODE: pipeline
+"#,
+        )
+        .expect("parse manifest");
+
+        let pipeline = manifest
+            .memory
+            .and_then(|memory| memory.pipeline_v2)
+            .expect("pipeline config");
+
+        assert_eq!(pipeline.extraction.provider, "command");
+        let command = pipeline
+            .extraction
+            .command
+            .expect("command config should parse");
+        assert_eq!(command.bin, "node");
+        assert_eq!(
+            command.args,
+            vec![
+                "script.mjs".to_string(),
+                "--transcript".to_string(),
+                "$TRANSCRIPT".to_string()
+            ]
+        );
+        assert_eq!(command.cwd.as_deref(), Some("/tmp/signet"));
+        assert_eq!(
+            command
+                .env
+                .and_then(|env| env.get("SIGNET_MODE").cloned())
+                .as_deref(),
+            Some("pipeline")
+        );
+    }
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
@@ -574,7 +625,28 @@ pub struct ExtractionConfig {
     pub endpoint: Option<String>,
     pub timeout: u64,
     pub min_confidence: f64,
+    pub command: Option<ExtractionCommandConfig>,
     pub escalation: Option<EscalationConfig>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(default, rename_all = "camelCase")]
+pub struct ExtractionCommandConfig {
+    pub bin: String,
+    pub args: Vec<String>,
+    pub cwd: Option<String>,
+    pub env: Option<HashMap<String, String>>,
+}
+
+impl Default for ExtractionCommandConfig {
+    fn default() -> Self {
+        Self {
+            bin: String::new(),
+            args: Vec::new(),
+            cwd: None,
+            env: None,
+        }
+    }
 }
 
 impl Default for ExtractionConfig {
@@ -586,6 +658,7 @@ impl Default for ExtractionConfig {
             endpoint: None,
             timeout: 30_000,
             min_confidence: 0.5,
+            command: None,
             escalation: Some(EscalationConfig::default()),
         }
     }
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -29,7 +29,11 @@ impl DaemonConfig {
             Ok(Some(manifest)) => manifest,
             Ok(None) => AgentManifest::default(),
             Err(error) => {
-                panic!("Invalid agent.yaml: {error}");
+                tracing::warn!(
+                    error = %error,
+                    "invalid or unreadable agent.yaml; falling back to default manifest"
+                );
+                AgentManifest::default()
             }
         };
         let (cfg_host, cfg_bind) = resolve_network_binding(
@@ -453,7 +457,7 @@ impl Default for EmbeddingConfig {
 
 #[cfg(test)]
 mod tests {
-    use super::{network_mode_from_bind, parse_manifest, resolve_network_binding};
+    use super::{network_mode_from_bind, parse_manifest, resolve_network_binding, PipelineV2Config};
 
     #[test]
     fn resolves_localhost_binding_by_default() {
@@ -624,6 +628,7 @@ memory:
             .expect("pipeline config");
 
         assert_eq!(pipeline.extraction.provider, "command");
+        assert_eq!(pipeline.extraction.timeout, 90_000);
         let command = pipeline
             .extraction
             .command
@@ -711,6 +716,12 @@ memory:
                 "$SESSION_KEY".to_string()
             ]
         );
+    }
+
+    #[test]
+    fn extraction_timeout_default_matches_typescript_timeout() {
+        let pipeline = PipelineV2Config::default();
+        assert_eq!(pipeline.extraction.timeout, 90_000);
     }
 }
 
@@ -895,7 +906,7 @@ impl Default for ExtractionConfig {
             model: "qwen3:4b".to_string(),
             strength: "medium".to_string(),
             endpoint: None,
-            timeout: 30_000,
+            timeout: 90_000,
             min_confidence: 0.5,
             command: None,
             escalation: Some(EscalationConfig::default()),
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -150,10 +150,6 @@ fn normalize_pipeline_extraction(
         return Ok(());
     }
 
-    if !is_pipeline_provider(&pipeline.extraction.provider) {
-        return Err("memory.pipelineV2.extraction.provider is invalid".to_string());
-    }
-
     if pipeline.extraction.command.is_none() {
         let legacy = raw.and_then(|value| raw_string(value, "extractionCommand"));
         pipeline.extraction.command = legacy.and_then(parse_command_argv);
@@ -499,7 +495,7 @@ memory:
         assert_eq!(pipeline.extraction.provider, "ollama");
         assert_eq!(pipeline.extraction.model, "qwen3.5:4b");
         assert_eq!(pipeline.synthesis.provider, "ollama");
-        assert_eq!(pipeline.synthesis.model, "qwen3:4b");
+        assert_eq!(pipeline.synthesis.model, "qwen3.5:4b");
         assert_eq!(pipeline.synthesis.timeout, pipeline.extraction.timeout);
     }
 
@@ -526,7 +522,7 @@ memory:
             .expect("pipeline config");
 
         assert_eq!(pipeline.synthesis.provider, "ollama");
-        assert_eq!(pipeline.synthesis.model, "qwen3:4b");
+        assert_eq!(pipeline.synthesis.model, "qwen3.5:4b");
         assert_eq!(
             pipeline.synthesis.endpoint.as_deref(),
             Some("http://127.0.0.1:11434")
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -114,10 +114,11 @@ fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Ag
 fn normalize_pipeline_synthesis(pipeline: &mut PipelineV2Config, raw: Option<&serde_yml::Value>) {
     let extraction = pipeline.extraction.clone();
     let fallback = SynthesisConfig::default();
+    let inherits_extraction = extraction.provider != "command";
     let provider = raw
         .and_then(|value| raw_child(value, "synthesis"))
         .and_then(|value| raw_string(value, "provider"))
-        .filter(|value| is_pipeline_provider(value));
+        .filter(|value| is_synthesis_provider(value));
     let model = raw
         .and_then(|value| raw_child(value, "synthesis"))
         .and_then(|value| raw_string(value, "model"));
@@ -133,23 +134,31 @@ fn normalize_pipeline_synthesis(pipeline: &mut PipelineV2Config, raw: Option<&se
 
     pipeline.synthesis.provider = provider
         .map(ToOwned::to_owned)
-        .unwrap_or_else(|| extraction.provider.clone());
+        .unwrap_or_else(|| {
+            if inherits_extraction {
+                extraction.provider.clone()
+            } else {
+                fallback.provider.clone()
+            }
+        });
     pipeline.synthesis.model = model.map(ToOwned::to_owned).unwrap_or_else(|| {
         if provider_won {
             default_pipeline_model(&pipeline.synthesis.provider).to_string()
-        } else {
+        } else if inherits_extraction {
             extraction.model.clone()
+        } else {
+            fallback.model.clone()
         }
     });
     pipeline.synthesis.endpoint = endpoint.map(ToOwned::to_owned).or_else(|| {
-        if provider_won {
+        if provider_won || !inherits_extraction {
             None
         } else {
             extraction.endpoint.clone()
         }
     });
     pipeline.synthesis.timeout = timeout.unwrap_or_else(|| {
-        if provider_won {
+        if provider_won || !inherits_extraction {
             fallback.timeout
         } else {
             extraction.timeout
@@ -182,6 +191,10 @@ fn is_pipeline_provider(value: &str) -> bool {
         value,
         "none" | "ollama" | "claude-code" | "opencode" | "codex" | "anthropic" | "openrouter" | "command"
     )
+}
+
+fn is_synthesis_provider(value: &str) -> bool {
+    is_pipeline_provider(value) && value != "command"
 }
 
 fn default_pipeline_model(provider: &str) -> &'static str {
@@ -322,7 +335,7 @@ memory:
         assert_eq!(pipeline.extraction.provider, "ollama");
         assert_eq!(pipeline.extraction.model, "qwen3.5:4b");
         assert_eq!(pipeline.synthesis.provider, "ollama");
-        assert_eq!(pipeline.synthesis.model, "qwen3.5:4b");
+        assert_eq!(pipeline.synthesis.model, "qwen3:4b");
         assert_eq!(pipeline.synthesis.timeout, pipeline.extraction.timeout);
     }
 
@@ -349,7 +362,7 @@ memory:
             .expect("pipeline config");
 
         assert_eq!(pipeline.synthesis.provider, "ollama");
-        assert_eq!(pipeline.synthesis.model, "qwen3.5:4b");
+        assert_eq!(pipeline.synthesis.model, "qwen3:4b");
         assert_eq!(
             pipeline.synthesis.endpoint.as_deref(),
             Some("http://127.0.0.1:11434")
@@ -472,6 +485,32 @@ memory:
                 .as_deref(),
             Some("pipeline")
         );
+        assert_eq!(pipeline.synthesis.provider, "ollama");
+        assert_eq!(pipeline.synthesis.model, "qwen3:4b");
+    }
+
+    #[test]
+    fn synthesis_command_provider_is_ignored_for_safety() {
+        let manifest = parse_manifest(
+            r#"
+memory:
+  pipelineV2:
+    extraction:
+      provider: ollama
+      model: qwen3.5:4b
+    synthesis:
+      provider: command
+"#,
+        )
+        .expect("parse manifest");
+
+        let pipeline = manifest
+            .memory
+            .and_then(|memory| memory.pipeline_v2)
+            .expect("pipeline config");
+
+        assert_eq!(pipeline.synthesis.provider, "ollama");
+        assert_eq!(pipeline.synthesis.model, "qwen3.5:4b");
     }
 }
 
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -97,24 +97,97 @@ fn load_manifest(base: &Path) -> Option<AgentManifest> {
 fn parse_manifest(content: &str) -> Option<AgentManifest> {
     let raw: serde_yml::Value = serde_yml::from_str(content).ok()?;
     let manifest: AgentManifest = serde_yml::from_str(content).ok()?;
-    Some(normalize_manifest(manifest, &raw))
+    normalize_manifest(manifest, &raw)
 }
 
-fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> AgentManifest {
+fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Option<AgentManifest> {
     let Some(memory) = manifest.memory.as_mut() else {
-        return manifest;
+        return Some(manifest);
     };
     let Some(pipeline) = memory.pipeline_v2.as_mut() else {
-        return manifest;
+        return Some(manifest);
     };
-    normalize_pipeline_synthesis(pipeline, raw_child(raw, "memory").and_then(|value| raw_child(value, "pipelineV2")));
-    manifest
+    let raw_pipeline = raw_child(raw, "memory").and_then(|value| raw_child(value, "pipelineV2"));
+    normalize_pipeline_extraction(pipeline, raw_pipeline)?;
+    normalize_pipeline_synthesis(pipeline, raw_pipeline)?;
+    Some(manifest)
 }
 
-fn normalize_pipeline_synthesis(pipeline: &mut PipelineV2Config, raw: Option<&serde_yml::Value>) {
+fn normalize_pipeline_extraction(
+    pipeline: &mut PipelineV2Config,
+    raw: Option<&serde_yml::Value>,
+) -> Option<()> {
+    let extraction_raw = raw.and_then(|value| raw_child(value, "extraction"));
+    if pipeline.extraction.provider != "command" {
+        return Some(());
+    }
+
+    if !is_pipeline_provider(&pipeline.extraction.provider) {
+        return None;
+    }
+
+    if pipeline.extraction.command.is_none() {
+        let legacy = raw.and_then(|value| raw_string(value, "extractionCommand"));
+        pipeline.extraction.command = legacy.and_then(parse_command_argv);
+    }
+
+    let mut command = pipeline.extraction.command.clone()?;
+    command.bin = command.bin.trim().to_string();
+    if command.bin.is_empty() {
+        return None;
+    }
+    if command.args.iter().any(|arg| arg.trim().is_empty()) {
+        command.args.retain(|arg| !arg.trim().is_empty());
+    }
+
+    if command.env.is_none() {
+        if let Some(raw_env) = extraction_raw
+            .and_then(|value| raw_child(value, "command"))
+            .and_then(|value| raw_child(value, "env"))
+            .and_then(|value| value.as_mapping())
+        {
+            let mut env = HashMap::new();
+            for (key, value) in raw_env {
+                let Some(k) = key.as_str() else {
+                    continue;
+                };
+                if !is_valid_env_key(k) {
+                    continue;
+                }
+                let Some(v) = value.as_str() else {
+                    continue;
+                };
+                env.insert(k.to_string(), v.to_string());
+            }
+            if !env.is_empty() {
+                command.env = Some(env);
+            }
+        }
+    } else if let Some(env) = command.env.as_mut() {
+        env.retain(|key, _| is_valid_env_key(key));
+        if env.is_empty() {
+            command.env = None;
+        }
+    }
+
+    pipeline.extraction.command = Some(command);
+    Some(())
+}
+
+fn normalize_pipeline_synthesis(
+    pipeline: &mut PipelineV2Config,
+    raw: Option<&serde_yml::Value>,
+) -> Option<()> {
     let extraction = pipeline.extraction.clone();
     let fallback = SynthesisConfig::default();
     let inherits_extraction = extraction.provider != "command";
+    let synthesis_raw = raw.and_then(|value| raw_child(value, "synthesis"));
+    if synthesis_raw
+        .and_then(|value| raw_string(value, "provider"))
+        .is_some_and(|provider| provider == "command")
+    {
+        return None;
+    }
     let provider = raw
         .and_then(|value| raw_child(value, "synthesis"))
         .and_then(|value| raw_string(value, "provider"))
@@ -167,6 +240,7 @@ fn normalize_pipeline_synthesis(pipeline: &mut PipelineV2Config, raw: Option<&se
     if pipeline.synthesis.provider == "none" {
         pipeline.synthesis.enabled = false;
     }
+    Some(())
 }
 
 fn raw_child<'a>(value: &'a serde_yml::Value, key: &str) -> Option<&'a serde_yml::Value> {
@@ -184,6 +258,59 @@ fn raw_string<'a>(value: &'a serde_yml::Value, key: &str) -> Option<&'a str> {
 
 fn raw_u64(value: &serde_yml::Value, key: &str) -> Option<u64> {
     raw_child(value, key)?.as_u64()
+}
+
+fn parse_command_argv(raw: &str) -> Option<ExtractionCommandConfig> {
+    let mut tokens: Vec<String> = Vec::new();
+    let mut current = String::new();
+    let mut quote: Option<char> = None;
+
+    for ch in raw.chars() {
+        match quote {
+            Some(q) if ch == q => quote = None,
+            Some(_) => current.push(ch),
+            None if ch == '"' || ch == '\'' => quote = Some(ch),
+            None if ch.is_whitespace() => {
+                if !current.is_empty() {
+                    tokens.push(std::mem::take(&mut current));
+                }
+            }
+            None => current.push(ch),
+        }
+    }
+    if !current.is_empty() {
+        tokens.push(current);
+    }
+    if tokens.is_empty() {
+        return None;
+    }
+
+    let bin = tokens.first()?.trim().to_string();
+    if bin.is_empty() {
+        return None;
+    }
+    let args = tokens
+        .iter()
+        .skip(1)
+        .map(|token| token.trim().to_string())
+        .filter(|token| !token.is_empty())
+        .collect::<Vec<_>>();
+
+    Some(ExtractionCommandConfig {
+        bin,
+        args,
+        cwd: None,
+        env: None,
+    })
+}
+
+fn is_valid_env_key(key: &str) -> bool {
+    let mut chars = key.chars();
+    match chars.next() {
+        Some(ch) if ch == '_' || ch.is_ascii_alphabetic() => {}
+        _ => return false,
+    }
+    chars.all(|ch| ch == '_' || ch.is_ascii_alphanumeric())
 }
 
 fn is_pipeline_provider(value: &str) -> bool {
@@ -490,7 +617,7 @@ memory:
     }
 
     #[test]
-    fn synthesis_command_provider_is_ignored_for_safety() {
+    fn synthesis_command_provider_is_rejected() {
         let manifest = parse_manifest(
             r#"
 memory:
@@ -501,6 +628,33 @@ memory:
     synthesis:
       provider: command
 "#,
+        );
+        assert!(manifest.is_none());
+    }
+
+    #[test]
+    fn extraction_command_provider_rejects_missing_command_config() {
+        let manifest = parse_manifest(
+            r#"
+memory:
+  pipelineV2:
+    extraction:
+      provider: command
+"#,
+        );
+        assert!(manifest.is_none());
+    }
+
+    #[test]
+    fn extraction_command_provider_parses_legacy_extraction_command() {
+        let manifest = parse_manifest(
+            r#"
+memory:
+  pipelineV2:
+    extraction:
+      provider: command
+    extractionCommand: node ./extract.mjs --transcript "$TRANSCRIPT" --session "$SESSION_KEY"
+"#,
         )
         .expect("parse manifest");
 
@@ -509,8 +663,21 @@ memory:
             .and_then(|memory| memory.pipeline_v2)
             .expect("pipeline config");
 
-        assert_eq!(pipeline.synthesis.provider, "ollama");
-        assert_eq!(pipeline.synthesis.model, "qwen3.5:4b");
+        let command = pipeline
+            .extraction
+            .command
+            .expect("legacy extractionCommand should be normalized");
+        assert_eq!(command.bin, "node");
+        assert_eq!(
+            command.args,
+            vec![
+                "./extract.mjs".to_string(),
+                "--transcript".to_string(),
+                "$TRANSCRIPT".to_string(),
+                "--session".to_string(),
+                "$SESSION_KEY".to_string()
+            ]
+        );
     }
 }
 
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -28,10 +28,10 @@ impl DaemonConfig {
         let manifest = match load_manifest(&base) {
             Ok(Some(manifest)) => manifest,
             Ok(None) => AgentManifest::default(),
-            Err(error) => {
-                if should_fail_startup_for_manifest_error(&error) {
-                    return Err(format!("Invalid agent.yaml: {error}"));
-                }
+            Err(ManifestLoadError::Fatal(error)) => {
+                return Err(format!("Invalid agent.yaml: {error}"));
+            }
+            Err(ManifestLoadError::Recoverable(error)) => {
                 tracing::warn!(
                     error = %error,
                     "invalid or unreadable agent.yaml; falling back to default manifest"
@@ -101,24 +101,51 @@ fn dirs_home() -> PathBuf {
         })
 }
 
-fn should_fail_startup_for_manifest_error(error: &str) -> bool {
-    error.contains("memory.pipelineV2.extraction.command")
+enum ManifestLoadError {
+    Recoverable(String),
+    Fatal(String),
+}
+
+fn should_fail_fast_for_manifest_error_context(raw: &serde_yml::Value, error: &str) -> bool {
+    let raw_pipeline = raw_child(raw, "memory").and_then(|value| raw_child(value, "pipelineV2"));
+    let extraction_provider = raw_pipeline
+        .and_then(|value| raw_child(value, "extraction"))
+        .and_then(|value| raw_string(value, "provider"))
+        .or_else(|| raw_pipeline.and_then(|value| raw_string(value, "extractionProvider")));
+    let synthesis_provider = raw_pipeline
+        .and_then(|value| raw_child(value, "synthesis"))
+        .and_then(|value| raw_string(value, "provider"))
+        .or_else(|| raw_pipeline.and_then(|value| raw_string(value, "synthesisProvider")));
+
+    extraction_provider.is_some_and(|provider| provider == "command")
+        || synthesis_provider.is_some_and(|provider| provider == "command")
+        || error.contains("memory.pipelineV2.extraction.command")
         || error.contains("memory.pipelineV2.synthesis.provider='command'")
 }
 
-fn load_manifest(base: &Path) -> Result<Option<AgentManifest>, String> {
+fn load_manifest(base: &Path) -> Result<Option<AgentManifest>, ManifestLoadError> {
     let path = base.join("agent.yaml");
     let content = match std::fs::read_to_string(&path) {
         Ok(content) => content,
         Err(err) if err.kind() == std::io::ErrorKind::NotFound => return Ok(None),
         Err(err) => {
-            return Err(format!(
+            return Err(ManifestLoadError::Recoverable(format!(
                 "failed to read {}: {err}",
                 path.to_string_lossy()
-            ))
+            )))
         }
     };
-    parse_manifest_result(&content).map(Some)
+    let raw: serde_yml::Value = serde_yml::from_str(&content)
+        .map_err(|err| ManifestLoadError::Recoverable(format!("YAML parse error: {err}")))?;
+    parse_manifest_from_raw(&content, &raw)
+        .map(Some)
+        .map_err(|error| {
+            if should_fail_fast_for_manifest_error_context(&raw, &error) {
+                ManifestLoadError::Fatal(error)
+            } else {
+                ManifestLoadError::Recoverable(error)
+            }
+        })
 }
 
 #[cfg(test)]
@@ -126,12 +153,20 @@ fn parse_manifest(content: &str) -> Option<AgentManifest> {
     parse_manifest_result(content).ok()
 }
 
+#[cfg(test)]
 fn parse_manifest_result(content: &str) -> Result<AgentManifest, String> {
     let raw: serde_yml::Value =
         serde_yml::from_str(content).map_err(|err| format!("YAML parse error: {err}"))?;
+    parse_manifest_from_raw(content, &raw)
+}
+
+fn parse_manifest_from_raw(
+    content: &str,
+    raw: &serde_yml::Value,
+) -> Result<AgentManifest, String> {
     let manifest: AgentManifest =
         serde_yml::from_str(content).map_err(|err| format!("manifest shape error: {err}"))?;
-    normalize_manifest(manifest, &raw)
+    normalize_manifest(manifest, raw)
 }
 
 fn normalize_manifest(
@@ -464,7 +499,7 @@ impl Default for EmbeddingConfig {
 mod tests {
     use super::{
         network_mode_from_bind, parse_manifest, resolve_network_binding,
-        should_fail_startup_for_manifest_error, PipelineV2Config,
+        should_fail_fast_for_manifest_error_context, PipelineV2Config,
     };
 
     #[test]
@@ -734,14 +769,39 @@ memory:
 
     #[test]
     fn startup_fail_fast_scopes_to_command_provider_manifest_errors() {
-        assert!(should_fail_startup_for_manifest_error(
-            "memory.pipelineV2.extraction.command is required when extraction.provider='command'"
+        let extraction_command_raw: serde_yml::Value = serde_yml::from_str(
+            r#"
+memory:
+  pipelineV2:
+    extraction:
+      provider: command
+      command: []
+"#,
+        )
+        .expect("raw parse");
+        assert!(should_fail_fast_for_manifest_error_context(
+            &extraction_command_raw,
+            "manifest shape error: invalid type: sequence, expected struct ExtractionCommandConfig"
         ));
-        assert!(should_fail_startup_for_manifest_error(
-            "memory.pipelineV2.synthesis.provider='command' is not supported. Use extraction.provider='command' instead."
+
+        let synthesis_command_raw: serde_yml::Value = serde_yml::from_str(
+            r#"
+memory:
+  pipelineV2:
+    synthesis:
+      provider: command
+"#,
+        )
+        .expect("raw parse");
+        assert!(should_fail_fast_for_manifest_error_context(
+            &synthesis_command_raw,
+            "memory.pipelineV2.synthesis.provider='command' is not supported"
         ));
-        assert!(!should_fail_startup_for_manifest_error("YAML parse error: invalid type"));
-        assert!(!should_fail_startup_for_manifest_error(
+
+        let generic_raw: serde_yml::Value =
+            serde_yml::from_str("agent:\n  name: default\n").expect("raw parse");
+        assert!(!should_fail_fast_for_manifest_error_context(
+            &generic_raw,
             "manifest shape error: missing field `agent`"
         ));
     }
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -20,7 +20,7 @@ pub struct DaemonConfig {
 }
 
 impl DaemonConfig {
-    pub fn from_env() -> Self {
+    pub fn from_env() -> Result<Self, String> {
         let base = std::env::var("SIGNET_PATH")
             .map(PathBuf::from)
             .unwrap_or_else(|_| dirs_home().join(".agents"));
@@ -29,11 +29,7 @@ impl DaemonConfig {
             Ok(Some(manifest)) => manifest,
             Ok(None) => AgentManifest::default(),
             Err(error) => {
-                tracing::warn!(
-                    error = %error,
-                    "invalid or unreadable agent.yaml; falling back to default manifest"
-                );
-                AgentManifest::default()
+                return Err(format!("Invalid agent.yaml: {error}"));
             }
         };
         let (cfg_host, cfg_bind) = resolve_network_binding(
@@ -59,14 +55,14 @@ impl DaemonConfig {
             }
         });
 
-        Self {
+        Ok(Self {
             base_path: base,
             db_path: db,
             port,
             host,
             bind,
             manifest,
-        }
+        })
     }
 
     pub fn memory_dir(&self) -> PathBuf {
@@ -113,6 +109,7 @@ fn load_manifest(base: &Path) -> Result<Option<AgentManifest>, String> {
     parse_manifest_result(&content).map(Some)
 }
 
+#[cfg(test)]
 fn parse_manifest(content: &str) -> Option<AgentManifest> {
     parse_manifest_result(content).ok()
 }
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -29,7 +29,14 @@ impl DaemonConfig {
             Ok(Some(manifest)) => manifest,
             Ok(None) => AgentManifest::default(),
             Err(error) => {
-                return Err(format!("Invalid agent.yaml: {error}"));
+                if should_fail_startup_for_manifest_error(&error) {
+                    return Err(format!("Invalid agent.yaml: {error}"));
+                }
+                tracing::warn!(
+                    error = %error,
+                    "invalid or unreadable agent.yaml; falling back to default manifest"
+                );
+                AgentManifest::default()
             }
         };
         let (cfg_host, cfg_bind) = resolve_network_binding(
@@ -92,6 +99,11 @@ fn dirs_home() -> PathBuf {
             );
             PathBuf::from(".")
         })
+}
+
+fn should_fail_startup_for_manifest_error(error: &str) -> bool {
+    error.contains("memory.pipelineV2.extraction.command")
+        || error.contains("memory.pipelineV2.synthesis.provider='command'")
 }
 
 fn load_manifest(base: &Path) -> Result<Option<AgentManifest>, String> {
@@ -450,7 +462,10 @@ impl Default for EmbeddingConfig {
 
 #[cfg(test)]
 mod tests {
-    use super::{network_mode_from_bind, parse_manifest, resolve_network_binding, PipelineV2Config};
+    use super::{
+        network_mode_from_bind, parse_manifest, resolve_network_binding,
+        should_fail_startup_for_manifest_error, PipelineV2Config,
+    };
 
     #[test]
     fn resolves_localhost_binding_by_default() {
@@ -715,6 +730,20 @@ memory:
     fn extraction_timeout_default_matches_typescript_timeout() {
         let pipeline = PipelineV2Config::default();
         assert_eq!(pipeline.extraction.timeout, 90_000);
+    }
+
+    #[test]
+    fn startup_fail_fast_scopes_to_command_provider_manifest_errors() {
+        assert!(should_fail_startup_for_manifest_error(
+            "memory.pipelineV2.extraction.command is required when extraction.provider='command'"
+        ));
+        assert!(should_fail_startup_for_manifest_error(
+            "memory.pipelineV2.synthesis.provider='command' is not supported. Use extraction.provider='command' instead."
+        ));
+        assert!(!should_fail_startup_for_manifest_error("YAML parse error: invalid type"));
+        assert!(!should_fail_startup_for_manifest_error(
+            "manifest shape error: missing field `agent`"
+        ));
     }
 }
 
--- a/packages/daemon-rs/crates/signet-core/src/config.rs
+++ b/packages/daemon-rs/crates/signet-core/src/config.rs
@@ -25,7 +25,13 @@ impl DaemonConfig {
             .map(PathBuf::from)
             .unwrap_or_else(|_| dirs_home().join(".agents"));
 
-        let manifest = load_manifest(&base).unwrap_or_default();
+        let manifest = match load_manifest(&base) {
+            Ok(Some(manifest)) => manifest,
+            Ok(None) => AgentManifest::default(),
+            Err(error) => {
+                panic!("Invalid agent.yaml: {error}");
+            }
+        };
         let (cfg_host, cfg_bind) = resolve_network_binding(
             manifest
                 .network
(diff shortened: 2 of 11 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(pipeline): add command extraction provider for summary worker control-plane path` (commit `d5e54519`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -11274,8 +11277,26 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 	if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
```
PR-Reviewer-Ant, line 11276: Blocking: this marks `command` as a valid synthesis provider, but `resolveSummaryProvider()` later warns that `synthesis.provider='command'` is not actually supported and falls back to Ollama. That creates a silent config contract break: startup validation accepts a setting that runtime ignores. The PR description says runtime validation/logging was updated, but this path now validates an unsupported mode instead of rejecting it.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,38 +307,162 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+
+			let stdout = "";
+			let stderr = "";
+			const MAX_CAPTURE_CHARS = 4000;
+
+			child.stdout?.on("data", (chunk: Buffer) => {
+				if (stdout.length >= MAX_CAPTURE_CHARS) return;
+				stdout += chunk.toString("utf-8").slice(0, MAX_CAPTURE_CHARS - stdout.length);
+			});
+			child.stderr?.on("data", (chunk: Buffer) => {
```
PR-Reviewer-Ant: Blocking: the new command path logs captured `stdout`/`stderr` on success. Because the command receives the session transcript and may echo or summarize it, this can write transcript content or extracted secrets into daemon logs, expanding the data-exposure surface for user sessions. The safer default is to avoid logging command output entirely, or gate it behind an explicit debug flag with redaction.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `d5e54519`

[Automated Bot Review: pr-reviewer] Found two blocking correctness/security issues in the new command-provider path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff directly shows `synthesis.provider='command'` being treated as valid in config/types/runtime validation while `resolveSummaryProvider()` explicitly ignores it and falls back to Ollama. The diff also shows the new command path logging captured stdout/stderr from the extractor, which can contain transcript-derived sensitive data. Both issues are provable from the changed lines without needing runtime reproduction.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1281,19 +1406,25 @@ export function startSummaryWorker(accessor: DbAccessor): SummaryWorkerHandle {
...
-			// immediately; secrets-store-only rotations rely on the 5-min TTL.
-			const envKey = process.env.ANTHROPIC_API_KEY ?? "";
-			const keyFingerprint = envKey ? new Bun.CryptoHasher("sha256").update(envKey).digest("hex").slice(0, 8) : "";
-			const providerKey = `${cfg.pipelineV2.synthesis.provider}:${cfg.pipelineV2.synthesis.model}:${cfg.pipelineV2.synthesis.timeout}:${keyFingerprint}`;
-			const cacheExpired = Date.now() - cachedProviderAt > PROVIDER_CACHE_TTL_MS;
-			if (!cachedProvider || providerKey !== cachedProviderKey || cacheExpired) {
-				cachedProvider = await resolveSummaryProvider(cfg);
-				cachedProviderKey = providerKey;
-				cachedProviderAt = Date.now();
+			if (cfg.pipelineV2.extraction.provider === "command") {
```
PR-Reviewer-Ant: When `extraction.provider === "command"`, this branch runs only `runSummaryCommandProvider()` and never calls `processJob()`. That means the configured/inherited `pipelineV2.synthesis` provider is effectively dead code in command mode, which diverges from the PR description/docs that say synthesis should fall back to normal synthesis defaults/providers. As written, enabling `extraction.provider: command` disables session summary generation rather than just swapping the extraction backend.

packages/daemon/src/memory-config.ts
```diff
@@ -392,6 +469,7 @@ export function loadPipelineConfig(yaml: Record<string, unknown>): PipelineV2Con
 				extractionRaw?.minConfidence ?? raw.minFactConfidenceForWrite,
```
PR-Reviewer-Ant, line 480: The loader now accepts `memory.pipelineV2.extraction.provider: command` even when `extraction.command` is missing or resolves to an empty `bin`. That misconfiguration is only detected later inside `runSummaryCommandProvider()`, after a background job is leased, retried, and potentially dead-lettered. This should be rejected during config load, the same way `synthesis.provider: command` is rejected, otherwise a typo in config turns into repeated runtime job failures.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f1d96384`

[Automated Review] Found two blocking issues in the new command-provider path: it bypasses the configured synthesis flow entirely, and it accepts invalid `provider: command` configs that only fail later at job runtime.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new `startSummaryWorker` branch explicitly skips `processJob()` whenever `extraction.provider === "command"`, so any configured or inherited `pipelineV2.synthesis.*` settings are never used despite the PR docs claiming synthesis falls back to normal synthesis defaults/providers. Separately, `loadPipelineConfig()` now accepts `provider: "command"` without validating that a usable `command.bin` exists, and `runSummaryCommandProvider()` only throws once a background job is already executing.

aaf2tbz:
Follow-up safety update pushed in `f1d96384`:

- Rejects `memory.pipelineV2.synthesis.provider: command` at config load (fail-fast with clear error).
- Ensures `extraction.provider: command` does **not** leak into synthesis fallback; synthesis uses safe defaults when omitted.
- Removes `command` from synthesis provider unions/allowlists and removes the old warn+fallback branch.
- Updates docs and adds coverage (TS + Rust config parsing tests).

Validation rerun:
- `packages/core` build
- daemon tests (`memory-config`, `summary-worker`)
- rust targeted tests for command-provider config behavior

aaf2tbz:
Addressed the latest blocking review points in `6553565a`:

- Added fail-fast config validation for `extraction.provider: command` requiring `extraction.command` (prevents runtime lease/retry/dead-letter loops on bad config).
- Kept command-mode processing in the normal `processJob()` flow (no separate branch bypass in `tick`) so significance gate + job lifecycle remain centralized.
- Removed stdout/stderr success logging from command execution path to reduce sensitive transcript leakage risk in logs.
- Removed unused import flagged by code-quality bot.
- Updated docs to clarify command config is required when `provider: command`.

Also updated PR body with the mandatory checked readiness checklist items.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,38 +307,160 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+				if (settled) return;
+				settled = true;
+				clearTimeout(timeout);
+				reject(err);
+			});
+
+			child.on("close", (code) => {
+				if (settled) return;
+				settled = true;
+				clearTimeout(timeout);
```
PR-Reviewer-Ant, line 442: The new `command` branch returns immediately after spawning the subprocess, so the rest of `processJob` never runs for these sessions. That means `command` mode is not actually a first-class extraction provider in the existing summary-worker lifecycle: downstream work in this function is skipped entirely, including the documented transcript retention path and any other post-processing still below this early return. If the intent is to keep custom extractors inside the same control-plane path, the command result needs to flow through the same persistence/post-processing steps rather than short-circuiting them.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -114,10 +114,11 @@ fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Ag
 fn normalize_pipeline_synthesis(pipeline: &mut PipelineV2Config, raw: Option<&serde_yml::Value>) {
     let extraction = pipeline.extraction.clone();
     let fallback = SynthesisConfig::default();
+    let inherits_extraction = extraction.provider != "command";
     let provider = raw
         .and_then(|value| raw_child(value, "synthesis"))
         .and_then(|value| raw_string(value, "provider"))
-        .filter(|value| is_pipeline_provider(value));
+        .filter(|value| is_synthesis_provider(value));
```
PR-Reviewer-Ant, line 271: This Rust path still silently accepts `synthesis.provider: command` by filtering it out and falling back, while the TypeScript loader now throws and the docs/PR description say this misconfiguration is rejected during config load. That creates a real TS/Rust parity break in the shadow daemon: the same invalid manifest is rejected in one runtime and accepted in the other.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -114,10 +114,11 @@ fn normalize_manifest(mut manifest: AgentManifest, raw: &serde_yml::Value) -> Ag
 fn normalize_pipeline_synthesis(pipeline: &mut PipelineV2Config, raw: Option<&serde_yml::Value>) {
```
PR-Reviewer-Ant: The PR claims Rust config parity for the new command provider, but this file only normalizes synthesis. There is no extraction-side normalization/validation equivalent to the TypeScript changes: no legacy flat `extractionCommand` string parsing into argv, and no fail-fast rejection when `extraction.provider: command` is configured without a valid `extraction.command` block. That means the two runtimes will resolve the same config differently.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `6553565a`

[Automated Review] Found control-plane regressions and Rust parity gaps that should be fixed before human review.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are directly visible in the changed code. `processJob` now short-circuits on `extraction.provider === "command"`, which bypasses the rest of the existing summary-worker flow. In Rust, the new config path still filters `synthesis.provider: command` back to a normal provider instead of rejecting it, and there is no corresponding extraction-side normalization for the documented legacy `extractionCommand` parsing or required-command validation.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Addressed in 7c3943c2. Command mode no longer short-circuits out of the job flow via an immediate return before persistence hooks. The command path is now integrated through processJob control flow, and transcript retention is explicitly preserved for command mode as well.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Addressed in 7c3943c2. Rust config normalization now rejects `synthesis.provider: command` (parse_manifest returns None for this invalid config), matching TS-side fail-fast semantics rather than silently filtering/falling back.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Addressed in 7c3943c2. Added extraction-side normalization/parity in Rust: (1) legacy flat `extractionCommand` string is parsed into argv for command mode, and (2) `extraction.provider: command` now fails config parse when command config/bin is missing.

aaf2tbz:
/signet-coding-copilot pass complete for latest review round.

Pushed `7c3943c2` to address the new blockers:

- Summary worker command flow: removed early-return short-circuit pattern and integrated command mode into `processJob` control flow; transcript retention is explicitly preserved in command mode.
- Rust parity (config):
  - reject `synthesis.provider: command` (invalid config)
  - reject `extraction.provider: command` when command config/bin is missing
  - normalize legacy flat `extractionCommand` string into argv for command mode

Validation rerun:
- `packages/core`: `bun run build`
- `packages/daemon`: `bun test src/memory-config.test.ts src/pipeline/summary-worker.test.ts`
- `packages/daemon-rs`: targeted `signet-core` tests for command-provider config normalization/validation
- `biome check` on changed files

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,348 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
-			error: e instanceof Error ? e.message : String(e),
-		});
+async function processJob(
+	accessor: DbAccessor,
+	provider: LlmProvider | null,
+	job: SummaryJobRow,
+	memoryCfg: ReturnType<typeof loadMemoryConfig>,
+): Promise<void> {
+	if (!passesSignificanceGate(accessor, job, memoryCfg)) return;
+	const commandMode = memoryCfg.pipelineV2.extraction.provider === "command";
```
PR-Reviewer-Ant, line 515: Blocking: this changes the summary-worker lifecycle much more than the PR description claims. In `command` mode the worker runs `runSummaryCommandProvider()` and then never enters the existing summary-generation branch, which means it no longer writes session summaries, inserts summary facts, updates the DAG, runs continuity scoring / predictor hooks, or triggers post-summary `MEMORY.md` synthesis. The docs say synthesis should still fall back to normal defaults when `extraction.provider: command`, but this path effectively disables it. If the intent is only to swap the extraction provider while keeping queue/retry/control-plane behavior, the command path needs to feed back into the rest of the job lifecycle instead of short-circuiting it.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,348 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
-		upsertSessionTranscript(job.session_key, job.transcript, job.harness, job.project, job.agent_id);
+			child.on("close", (code) => {
+				if (settled) return;
+				settled = true;
+				clearTimeout(timeout);
+				const exitCode = code ?? 1;
+				if (exitCode !== 0) {
+					const detail = stderr.trim().length > 0 ? stderr.trim() : stdout.trim();
+					reject(new Error(`summary command exited with code ${exitCode}: ${trimOutput(detail, 500)}`));
+					return;
```
PR-Reviewer-Ant, line 450: Warning: on non-zero exit this throws an error containing up to 500 chars of the extractor's stderr/stdout. Because the command receives the full transcript path, many extractor failures will echo transcript content or secrets into stderr; this error string is then likely to be logged/persisted with the failed job. That reintroduces the transcript-exposure risk the PR says it is trying to reduce. Prefer logging only exit status plus bounded metadata, not command output.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `7c3943c2`

[Automated Review] Found a blocking lifecycle regression in the new `command` provider path and one security-sensitive error leakage issue.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The regression is directly visible in `packages/daemon/src/pipeline/summary-worker.ts`: when `extraction.provider === "command"`, `processJob()` runs the command and then skips the entire existing summary/synthesis branch. The non-zero-exit error path also explicitly injects captured command output into the thrown error string. Both issues are provable from the diff without needing runtime reproduction.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Addressed in 5698dbe2. Command mode no longer skips the downstream summary-worker lifecycle when a synthesis provider is available: the command runs first, then the normal summary branch continues (summary generation, DAG/write hooks, continuity/predictor, and synthesis trigger). I also updated docs to state this explicitly.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Addressed in 5698dbe2. Non-zero command exits no longer include captured stdout/stderr in thrown errors. The command now runs with stdio ignored and failures surface as exit-code-only errors to avoid transcript/secret leakage via error strings.

aaf2tbz:
/signet-coding-copilot follow-up pass complete for the newest review round.

Pushed `5698dbe2`:

- command-mode lifecycle: run command extraction first, then continue normal summary-worker branch when synthesis provider is configured (no lifecycle short-circuit)
- command failure hygiene: removed stdout/stderr capture from failure errors; exit-code-only failure message now used
- docs updated to match behavior

Validation rerun:
- `packages/core`: `bun run build`
- `packages/daemon`: `bun test src/memory-config.test.ts src/pipeline/summary-worker.test.ts`
- `packages/daemon-rs`: targeted command-provider parity tests
- `biome check` on touched files

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -97,27 +97,101 @@ fn load_manifest(base: &Path) -> Option<AgentManifest> {
 fn parse_manifest(content: &str) -> Option<AgentManifest> {
```
PR-Reviewer-Ant, line 152: Blocking: this changes Rust manifest parsing from 'best effort' to `None` on invalid `pipelineV2` command config, but the Rust daemon still does `load_manifest(&base).unwrap_or_default()`. A typo like `extraction.provider: command` without `command:` now causes the entire manifest to be discarded and the daemon to boot with defaults instead of failing fast. That diverges from the PR description ('fail-fast validation') and from the TS loader, which throws a clear error.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,329 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
-		logger.warn("summary-worker", "Failed to write session summary to DAG (non-fatal)", {
-			error: e instanceof Error ? e.message : String(e),
+		await new Promise<void>((resolve, reject) => {
+			const child = nodeSpawn(bin, args, {
+				cwd: cwd && cwd.length > 0 ? cwd : undefined,
+				env: {
+					...process.env,
+					...envFromConfig,
+					SIGNET_PATH: AGENTS_DIR,
+				},
```
PR-Reviewer-Ant, line 398: Warning: the new `command` path only checks the subprocess exit code. It never consumes stdout/stderr or any structured output, so `extraction.provider: command` is acting as a side-effect hook rather than an extraction provider. In the `synthesis.provider: none` case this job completes without producing any summary/fact writes unless the external command mutates Signet state on its own. If that is intentional, the PR description/docs should say so explicitly; otherwise the implementation does not achieve a first-class extraction provider.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5698dbe2`

[Automated Review] Found one blocking parity regression in the Rust config path and one feature-behavior mismatch in the new command provider flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The Rust manifest loader now returns `None` for invalid `pipelineV2` command configs in the changed code, while `DaemonConfig::from_env()` still falls back to `AgentManifest::default()` on `None`; that regression is directly inferable from the diff. The summary-worker path clearly executes the command and then continues without consuming command output, but confirming whether that violates issue #334 depends on the intended external command contract.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Addressed in 000908d0. Rust now fails fast with an explicit invalid-config error path instead of silently booting with defaults after manifest discard. `DaemonConfig::from_env()` now distinguishes missing file vs invalid file and panics with a clear `Invalid agent.yaml: ...` message for malformed/invalid command-provider config.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Addressed as documentation contract in 000908d0. The command provider is intentionally side-effect based: stdout/stderr are not consumed as extraction payload, and the external command is responsible for writing memories directly to Signet state. I updated CONFIGURATION.md and PIPELINE.md to state this explicitly.

aaf2tbz:
Pushed follow-up commit `000908d0` for the newest review items:

- Rust fail-fast parity: invalid command-provider manifest config now fails startup with a clear `Invalid agent.yaml: ...` error path instead of silently discarding manifest and booting defaults.
- Command-provider contract docs: clarified that stdout/stderr are intentionally not consumed as extraction payload; the external command must write memories directly to Signet state.

Validation rerun:
- `packages/core`: `bun run build`
- `packages/daemon`: `bun test src/memory-config.test.ts src/pipeline/summary-worker.test.ts`
- `packages/daemon-rs`: targeted `signet-core` command-provider parity tests
- `biome check` on touched files

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -586,6 +897,7 @@ impl Default for ExtractionConfig {
             endpoint: None,
```
PR-Reviewer-Ant, line 990: Blocking: this PR updates the documented/default extraction timeout to 90000 ms and the TS side also uses `DEFAULT_PIPELINE_TIMEOUT_MS = 90000`, but the Rust shadow config still defaults `ExtractionConfig.timeout` to `30_000`. That breaks the PR's stated Rust parity and means the shadow daemon will exercise a different timeout regime than the TS daemon whenever `memory.pipelineV2.extraction.timeout` is omitted.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -25,7 +25,13 @@ impl DaemonConfig {
             .map(PathBuf::from)
             .unwrap_or_else(|_| dirs_home().join(".agents"));
 
-        let manifest = load_manifest(&base).unwrap_or_default();
+        let manifest = match load_manifest(&base) {
+            Ok(Some(manifest)) => manifest,
+            Ok(None) => AgentManifest::default(),
+            Err(error) => {
```
PR-Reviewer-Ant: Warning: `DaemonConfig::from_env()` now panics on any manifest read/validation error (`panic!("Invalid agent.yaml: {error}")`). Before this change, an unreadable/invalid manifest fell back to defaults. That is a behavior change outside the command-provider feature and can take down the Rust shadow process on a transient or partially-written `agent.yaml`, rather than surfacing a recoverable config error.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `000908d0`

[Automated Review] Found one blocking parity bug and one startup-regression risk in the Rust shadow config path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The timeout/default mismatch is directly visible in the changed Rust config defaults versus the updated TS/docs defaults. The startup regression risk is also explicit in the new `panic!` path in `DaemonConfig::from_env`, although I did not run the shadow daemon to observe the runtime effect.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(pipeline): add command extraction provider for summary worker control-plane path` (commit `5fb4849d`) now and I'll follow up shortly with feedback.

aaf2tbz:
/signet-coding-copilot pass complete for the latest review round.

Pushed `5fb4849d`:

- Rust parity: `ExtractionConfig::default().timeout` is now `90_000` to match TS/docs extraction defaults.
- Startup resilience: removed `panic!` in `DaemonConfig::from_env()`; invalid/unreadable `agent.yaml` now logs a warning and falls back to default manifest (restores recoverable startup behavior).
- Added/updated Rust tests to pin the 90s extraction timeout default in command-provider paths.

Validation rerun:
- `cargo test -p signet-core command_provider --quiet`
- `cargo test -p signet-core synthesis_command_provider_is_rejected --quiet`
- `cargo test -p signet-core extraction_timeout_default_matches_typescript_timeout --quiet`

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -321,7 +499,7 @@ memory:
         assert_eq!(pipeline.extraction.provider, "ollama");
```
BusyBee3333, line 542: **Likely test regression — incorrect assertion.** `synthesis_defaults_stay_local_when_pipeline_block_omits_them` changed the expected synthesis model from `"qwen3.5:4b"` to `"qwen3:4b"`, but the logic in `normalize_pipeline_synthesis` contradicts this. When `extraction.provider == "ollama"` (not "command"), `inherits_extraction = true` and `provider_won = false`, so line ~244 takes the `else if inherits_extraction` branch and sets `synthesis.model = extraction.model.clone()` = `"qwen3.5:4b"`. The test as written will fail. The PR only ran `cargo test … extraction_command_provider_parses_command_block`, so this was not caught. Same issue applies to `synthesis_enabled_only_keeps_inheriting_extraction_values` (line ~526).

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -88,36 +98,137 @@ fn dirs_home() -> PathBuf {
...
+
+    if pipeline.extraction.command.is_none() {
+        let legacy = raw.and_then(|value| raw_string(value, "extractionCommand"));
+        pipeline.extraction.command = legacy.and_then(parse_command_argv);
+    }
+
+    let mut command = pipeline.extraction.command.clone().ok_or_else(|| {
+        "memory.pipelineV2.extraction.command is required when extraction.provider='command'"
+            .to_string()
+    })?;
```
BusyBee3333, line 205: **Unreachable error branch.** `normalize_pipeline_extraction` returns early if `provider != "command"` (line ~160), so by the time we reach the `!is_pipeline_provider` check, `provider` is already confirmed to be `"command"`, which satisfies `is_pipeline_provider`. The error `"memory.pipelineV2.extraction.provider is invalid"` can never fire. This dead branch will confuse future readers and masks the intended validation logic.

packages/core/src/pipeline-providers.ts
```diff
@@ -6,6 +6,7 @@ export const PIPELINE_PROVIDER_CHOICES = [
 	"opencode",
 	"anthropic",
 	"openrouter",
+	"command",
 ] as const;
```
BusyBee3333, line 11: **Silent breaking change to default timeout.** `DEFAULT_PIPELINE_TIMEOUT_MS` was raised from 45 000 ms to 90 000 ms, and the Rust `ExtractionConfig::default().timeout` was changed from 30 000 ms to 90 000 ms. This new default applies to all existing providers (ollama, claude-code, etc.), not just `command`. Existing deployments that relied on the 45 s hard cap to bound runaway Ollama calls will silently get a 90 s cap. The PR description does not call out this behavioral change for existing users. At minimum, the changelog in CONFIGURATION.md should highlight this as a breaking default change, or the constant should only apply when `provider == "command"`.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,329 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+			child.on("close", (code) => {
+				if (settled) return;
+				settled = true;
+				clearTimeout(timeout);
+				const exitCode = code ?? 1;
+				if (exitCode !== 0) {
+					reject(new Error(`summary command exited with code ${exitCode}`));
+					return;
+				}
+				resolve();
```
BusyBee3333, line 452: **Potential double-write when command mode is used with a synthesis provider.** In `processJob`, when `commandMode` is true and `provider` is non-null (i.e., a synthesis provider is configured), the `if (provider)` block executes in full: it calls `processSingle`/`processChunked`, then writes a session summary markdown file, and then calls `insertSummaryFacts` — which persists LLM-extracted facts into `memories.db`. This means the same transcript is processed twice: once by the external command (which the docs say is responsible for writing memories directly) and again by the LLM synthesis path (which also inserts extracted facts). The result is that the same session's facts can be written twice from two different code paths. If this is intentional, it should be explicitly documented and the fact deduplication path must be verified as sufficient. If unintentional, the `insertSummaryFacts` call (and the markdown file write) inside the `if (provider)` block should be skipped when `commandMode` is true, with synthesis/continuity/predictor hooks running but not fact re-extraction.

BusyBee3333 (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5fb4849d`

[Automated Bot Review] Found one likely test regression in the Rust config layer, one unreachable error branch, a silent breaking behavioral change to the default extraction timeout, and a potential double-write when command extraction and a synthesis provider are both configured.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The Rust synthesis model inheritance test changes (qwen3.5:4b → qwen3:4b) are directly traceable through normalize_pipeline_synthesis: with inherits_extraction=true and no synthesis block, the code path at config.rs:244 clones extraction.model, which for these test cases is 'qwen3.5:4b', not the fallback default 'qwen3:4b'. The PR validation only ran one Rust test (extraction_command_provider_parses_command_block), not the full suite. The dead-code branch and double-write issue are similarly evident from reading the diff.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -25,7 +25,17 @@ impl DaemonConfig {
             .map(PathBuf::from)
             .unwrap_or_else(|_| dirs_home().join(".agents"));
 
-        let manifest = load_manifest(&base).unwrap_or_default();
+        let manifest = match load_manifest(&base) {
```
PR-Reviewer-Ant, line 28: Blocking: this no longer fails fast on invalid `agent.yaml`. The new `Err(error) => { warn!(...); AgentManifest::default() }` path silently discards malformed/invalid pipeline config in the Rust daemon, whereas the TypeScript loader now throws for the same `extraction.provider='command'` / `synthesis.provider='command'` mistakes. That means the PR's stated 'fail-fast validation' and 'Rust config parity' are not actually true: the shadow daemon will quietly run with defaults instead of exercising the configured command path, masking real divergences and making bad configs much harder to detect.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5fb4849d`

[Automated Review] Found one blocking parity/config-validation issue in the Rust shadow path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new Rust config loader explicitly catches manifest parse/validation errors and replaces the manifest with defaults, while the TypeScript path now throws on the same command-provider misconfigurations. That divergence is directly visible in the diff and contradicts both the PR description's fail-fast claim and the repo's daemon/Rust parity rule. I did not run the daemon to observe the resulting shadow behavior.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Fixed in ed9bcac8 — restored both assertions to inherited extraction model (`qwen3.5:4b`) to match normalize_pipeline_synthesis behavior. Also reran `cargo test -p signet-core config::tests:: --quiet` (13/13 passing).

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Good catch. Removed the unreachable `!is_pipeline_provider(...)` branch in `normalize_pipeline_extraction` in ed9bcac8. Since the function early-returns for non-command providers, this path could never fire.

packages/core/src/pipeline-providers.ts
aaf2tbz: I verified this against current `origin/main`: `packages/core/src/pipeline-providers.ts` already has `DEFAULT_PIPELINE_TIMEOUT_MS = 90000` on main, so this PR does not raise TS default timeout. Rust was aligned to match existing TS/docs parity. No further constant change in ed9bcac8.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Addressed in ed9bcac8. In command mode, provider generation can still run for downstream hooks, but summary markdown write + `insertSummaryFacts` are now skipped to prevent duplicate memory writes. Also deduped transcript retention to a single upsert path.

aaf2tbz:
/signet-coding-copilot pass complete for BusyBee round.

Pushed `ed9bcac8`:

- Fixed Rust config test regression: synthesis inheritance assertions now correctly expect extraction model (`qwen3.5:4b`) in both affected tests.
- Removed unreachable Rust branch in `normalize_pipeline_extraction` (`!is_pipeline_provider(...)` after command-only guard).
- Prevented command+synthesis duplicate memory writes in summary worker:
  - in command mode, skip markdown summary file write + `insertSummaryFacts`
  - keep synthesis downstream hooks/DAG/continuity/predictor flow
  - dedupe transcript retention to one upsert path
- Updated docs (`CONFIGURATION.md`, `PIPELINE.md`) to reflect command-mode write contract and artifact-skipping behavior.

Validation rerun:
- `bun run build` (packages/core)
- `bun test packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/summary-worker.test.ts`
- `cargo test -p signet-core config::tests:: --quiet`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts docs/CONFIGURATION.md docs/PIPELINE.md`

Also replied inline to each BusyBee thread with exact resolution details.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1281,19 +1396,23 @@ export function startSummaryWorker(accessor: DbAccessor): SummaryWorkerHandle {
...
-			const providerKey = `${cfg.pipelineV2.synthesis.provider}:${cfg.pipelineV2.synthesis.model}:${cfg.pipelineV2.synthesis.timeout}:${keyFingerprint}`;
-			const cacheExpired = Date.now() - cachedProviderAt > PROVIDER_CACHE_TTL_MS;
-			if (!cachedProvider || providerKey !== cachedProviderKey || cacheExpired) {
-				cachedProvider = await resolveSummaryProvider(cfg);
-				cachedProviderKey = providerKey;
-				cachedProviderAt = Date.now();
+			let providerForJob: LlmProvider | null = null;
+			if (cfg.pipelineV2.synthesis.enabled && cfg.pipelineV2.synthesis.provider !== "none") {
+				// Cache provider across jobs — re-resolve on config change, env
+				// key rotation, or TTL expiry. Env-var key changes invalidate
```
PR-Reviewer-Ant, line 1523: Blocking: this new guard only resolves an LLM provider when `synthesis.enabled` is true and `synthesis.provider !== 'none'`. For the existing non-command flow, `processJob()` still throws if `provider` is null. That means configs like `synthesis.enabled: false` or `synthesis.provider: none` now break session-end extraction entirely instead of only disabling summary generation. This diverges from the PR/docs claim that synthesis is separate from extraction. The provider cache here still needs to resolve the effective extraction/synthesis provider for non-command extraction jobs.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `ed9bcac8`

[Automated Review] Found one blocking regression in the summary worker provider-resolution path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The regression is directly visible in the changed control flow in packages/daemon/src/pipeline/summary-worker.ts. The worker now skips provider resolution whenever synthesis is disabled or set to 'none', but processJob still requires a provider for every non-command extraction path. That makes extraction jobs fail for configurations that are documented to disable only session-summary synthesis.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,331 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
-	// Lossless retention: preserve raw transcript alongside extracted facts
-	if (job.session_key) {
-		upsertSessionTranscript(job.session_key, job.transcript, job.harness, job.project, job.agent_id);
+	const tempDir = mkdtempSync(join(tmpdir(), "signet-summary-command-"));
+	const transcriptPath = join(tempDir, "transcript.txt");
+	writeFileSync(transcriptPath, job.transcript, "utf-8");
+
+	const replacements: Record<string, string> = {
+		$TRANSCRIPT: transcriptPath,
+		$TRANSCRIPT_PATH: transcriptPath,
```
BusyBee3333, line 366: **Security – user-controlled data substituted into executable path.** `substituteCommandTokens` is applied to `command.bin` and `command.cwd` using a replacements map that includes `$PROJECT` and `$SESSION_KEY`, both of which come from `job` (session/connector-supplied data). If an admin writes a config like `bin: '/scripts/$PROJECT/extract.mjs'` and an attacker can influence the `project` field on enqueued jobs (e.g. via a malicious connector or crafted session), the substituted value controls which binary `spawn` executes. Even without a shell, path traversal (`../../malicious`) works because `spawn` resolves the final binary path from the filesystem. Token substitution should be limited to `args[]` and env var values only — never applied to `bin` or `cwd`. Strip substitution from those two fields, or at minimum validate that the resolved `bin` still starts with the literal prefix provided in config (before any token).

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,331 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
-		factsPreview: result.facts.slice(0, 10).map((fact) => fact.content),
-	});
+export async function runSummaryCommandProvider(
+	job: SummaryJobRow,
+	cfg: ReturnType<typeof loadMemoryConfig>,
+): Promise<void> {
+	const command = cfg.pipelineV2.extraction.command;
+	if (!command) {
+		throw new Error("pipelineV2.extraction.command is required when extraction.provider is 'command'");
+	}
```
BusyBee3333, line 359: **Agent scoping gap – `$AGENT_ID` not exposed as a substitution token.** The replacements map provides `$TRANSCRIPT`, `$SESSION_KEY`, `$PROJECT`, and `$SIGNET_PATH`, but omits `$AGENT_ID` (available as `job.agent_id`). The multi-agent section of PIPELINE.md states that the pipeline tags extracted memories with the requesting agent's ID and that callers needing private memories must set `agent_id` explicitly. An external command writing directly to `memories.db` has no way to scope its inserts to the correct agent without making an extra API call. This is exactly the 'agent scoping leaks' recurring failure mode from AGENTS.md. Add `$AGENT_ID: job.agent_id` to the replacements map and document it alongside the other tokens in CONFIGURATION.md and PIPELINE.md.

packages/daemon/src/memory-config.ts
```diff
@@ -241,6 +249,62 @@ function parseOptionalUrl(raw: unknown): string | undefined {
...
+		};
+	}
+
+	if (!isRecord(raw)) {
+		return undefined;
+	}
+
+	const record = raw;
+	const candidateBin =
+		typeof record.bin === "string" ? record.bin : typeof record.command === "string" ? record.command : "";
```
BusyBee3333: **Undocumented fallback inconsistent with Rust parity.** `parseCommandConfig` falls back to `record.command` as the bin value when `record.bin` is absent:
```ts
const candidateBin = typeof record.bin === "string" ? record.bin : typeof record.command === "string" ? record.command : "";
```
This field alias is not mentioned in the YAML schema docs, not tested, and has no equivalent in `normalize_pipeline_extraction` in `config.rs` (the Rust side only ever reads `bin`). Per AGENTS.md, the Rust shadow parity rule requires any behavioral change in the TS daemon to be reflected in `daemon-rs`. Either remove the fallback and align on `bin` only, or add the same alias to the Rust normalization path and document it.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,331 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+				},
+				stdio: ["ignore", "ignore", "ignore"],
+				windowsHide: true,
+			});
+
+			let settled = false;
+			const timeout = setTimeout(() => {
+				if (settled) return;
+				settled = true;
+				child.kill("SIGTERM");
```
BusyBee3333, line 416: **Minor leak: inner SIGKILL timer not cancelled on clean SIGTERM exit.** When the outer timeout fires, a nested `setTimeout(..., 2000)` is scheduled for SIGKILL. If the child exits cleanly after receiving SIGTERM (before the 2-second timer), `settled` is already `true`, so the SIGKILL call is swallowed by the try/catch — no harm done. But the timer object itself is never cleared and keeps the event loop alive for 2 extra seconds after the rejection. Capture the inner timer ID and clear it in the `close` handler when `settled` is already true.

BusyBee3333 (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ed9bcac8`

[Automated Review - pr-reviewer] Found one security issue (user-controlled token substituted into executable path), one missing agent scoping token (breaking the repo's recurring invariant), one silent global timeout change affecting all providers, and one undocumented config fallback inconsistent with the Rust parity implementation.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The token-substitution-into-bin risk is directly visible in runSummaryCommandProvider in summary-worker.ts: bin and cwd are passed through substituteCommandTokens before spawn, and $PROJECT/$SESSION_KEY are job fields, not admin-controlled. The missing $AGENT_ID token is verifiable by comparing the replacements map against job.agent_id and the multi-agent scoping rule in AGENTS.md. The DEFAULT_PIPELINE_TIMEOUT_MS change from 45000→90000 is explicit in pipeline-providers.ts and affects all providers. The record.command fallback is in parseCommandConfig in memory-config.ts and has no counterpart in the Rust config.rs normalization.


- packages/core/src/pipeline-providers.ts:14 - **Undocumented global behavior change.** `DEFAULT_PIPELINE_TIMEOUT_MS` is raised from 45 000 ms to 90 000 ms. This constant is the default timeout for **all** extraction providers (ollama, anthropic, openrouter, etc.), not just the new `command` provider. Existing users with slower hardware who relied on the previous 45-second budget for per-call timeouts will silently double their worst-case blocking window per extraction job. The docs update shows the new value but does not call out that this is a change from the previous default. Either keep the old default for existing providers and only use 90 000 ms as the `command`-specific default, or add an explicit upgrade note to both docs files and the PR description.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Addressed in c928b8eb. Rust now fails fast on invalid/unreadable `agent.yaml` without panicking: `DaemonConfig::from_env()` returns `Result<Self, String>` and `main.rs` propagates startup failure (`Invalid agent.yaml: ...`) as a clean process error.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Fixed in c928b8eb. Provider resolution now occurs whenever extraction is non-command (`requiresProviderForExtraction`) OR synthesis explicitly needs a provider, so non-command jobs no longer fail when `synthesis.enabled` is false.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Fixed in c928b8eb. Token substitution into executable path was tightened: `bin`/`cwd` now use a restricted replacement map (`$SIGNET_PATH`, `$AGENT_ID` only). User-derived tokens (`$PROJECT`, `$SESSION_KEY`, `$TRANSCRIPT`) are still allowed in args/env only.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Added in c928b8eb: `$AGENT_ID` is now included in command token replacements (args/env) and documented in CONFIGURATION.md + PIPELINE.md.

packages/daemon/src/memory-config.ts
aaf2tbz: Addressed in c928b8eb. `parseCommandConfig` no longer falls back to `record.command` as bin; TS now requires `command.bin`, matching Rust parity expectations. Added test coverage for rejecting object command configs that omit `bin`.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Fixed in c928b8eb. Added explicit cleanup for the inner SIGKILL timer (`killTimer`) and clear it on `error`/`close`, so a clean SIGTERM exit does not leave a dangling follow-up timer.

aaf2tbz:
/signet-coding-copilot pass complete for the newest review round.

Pushed `c928b8eb` with the requested fixes:

- Fixed provider-resolution regression: non-command extraction now always resolves a provider (`requiresProviderForExtraction`), even if `synthesis.enabled` is false.
- Hardened command token substitution:
  - added `$AGENT_ID` token support
  - restricted `bin`/`cwd` substitution to trusted tokens only (`$SIGNET_PATH`, `$AGENT_ID`)
  - kept transcript/session/project substitution in args/env only
- Removed TS-only fallback `command.command -> bin` in `parseCommandConfig` for Rust parity; now `command.bin` is required.
- Fixed SIGKILL follow-up timer cleanup in command timeout flow.
- Rust config startup path now fails fast without panic:
  - `DaemonConfig::from_env()` returns `Result<Self, String>`
  - `signet-daemon` main propagates invalid `agent.yaml` as clean startup error.
- Docs updated for `$AGENT_ID` token + safety guidance on bin/cwd token usage.

Validation rerun:
- `bun test packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/summary-worker.test.ts`
- `cargo test -p signet-core config::tests:: --quiet`
- `cargo check -p signet-daemon`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/memory-config.ts packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/summary-worker.test.ts docs/CONFIGURATION.md docs/PIPELINE.md`

Also replied inline on each new thread.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,345 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+				},
+				stdio: ["ignore", "ignore", "ignore"],
+				windowsHide: true,
+			});
+
+			let settled = false;
+			let killTimer: ReturnType<typeof setTimeout> | null = null;
+			const clearKillTimer = (): void => {
+				if (killTimer) {
+					clearTimeout(killTimer);
```
PR-Reviewer-Ant, line 408: On timeout, this rejects immediately after sending `SIGTERM`, before the child has actually exited. That means the summary job can be marked failed/retried while the original command is still running and still writing to Signet state, which defeats the queue's single-flight semantics and can duplicate or corrupt writes. The timeout path needs to wait for process exit (or confirmed kill) before returning failure.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -20,12 +20,18 @@ pub struct DaemonConfig {
 }
 
 impl DaemonConfig {
-    pub fn from_env() -> Self {
+    pub fn from_env() -> Result<Self, String> {
```
PR-Reviewer-Ant, line 23: This changes Rust daemon startup from 'ignore malformed manifest and continue with defaults' to 'fail startup on any `agent.yaml` read/parse/shape error'. The PR description says the fail-fast behavior is for command-provider validation, but this makes unrelated manifest mistakes fatal too, including `--install-service`. That is a broader breaking change in runtime behavior and should either be scoped back to the new command config validation or called out explicitly as an intentional startup contract change.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c928b8eb`

[Automated Review] Found one blocking control-plane race and one startup behavior regression that diverges from the PR’s stated scope.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The timeout race is directly visible in `runSummaryCommandProvider`: the promise rejects before the child process has actually exited, so the worker can retry while the previous command is still mutating state. The Rust startup regression is also explicit in the diff: `DaemonConfig::from_env()` now fails hard on any `agent.yaml` parse/shape error, which is broader than the PR description's command-provider validation changes. I did not run the daemon to capture a live repro because the review context only provided the diff.

packages/daemon/src/pipeline/summary-worker.ts
aaf2tbz: Fixed in 2a3c85c9. Timeout path no longer rejects immediately after SIGTERM. We now mark `timedOut=true`, send SIGTERM (+ delayed SIGKILL), and only reject on `close`/`error` after the child has exited/been confirmed terminated, preventing retry overlap with a still-running command.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Addressed in 2a3c85c9 by narrowing startup fail-fast scope. Rust now fails startup only for command-provider config validation errors (`extraction.command` / invalid `synthesis.provider=command`), while unrelated YAML parse/shape/read errors fall back to default manifest with warning (restoring prior best-effort behavior for general manifest issues, including service install flows).

aaf2tbz:
/signet-coding-copilot follow-up pass complete for the newest blocker round.

Pushed `2a3c85c9`:

- Fixed command-timeout control-plane race: timeout path now waits for process termination before rejecting (no immediate reject after SIGTERM), preventing retry overlap while the original command is still mutating state.
- Added regression test coverage for timeout semantics: `runSummaryCommandProvider` now has a test that verifies rejection occurs after child termination marker is written.
- Scoped Rust fail-fast behavior to command-provider validation only:
  - startup fails for invalid command-provider manifest config
  - unrelated manifest parse/shape/read errors fall back to default manifest with warning (best-effort behavior restored)
- Added Rust config test coverage for fail-fast scope predicate.

Validation rerun:
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.test.ts`
- `cargo test -p signet-core config::tests:: --quiet`
- `cargo check -p signet-daemon`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.ts packages/daemon/src/memory-config.test.ts docs/CONFIGURATION.md docs/PIPELINE.md`

Replied inline on both new review threads.

packages/daemon-rs/crates/signet-core/src/config.rs
```diff
@@ -88,36 +101,139 @@ fn dirs_home() -> PathBuf {
         })
```
PR-Reviewer-Ant, line 88: Blocking: the new fail-fast behavior for `provider: command` is incomplete in the Rust path. `should_fail_startup_for_manifest_error()` only matches errors containing `memory.pipelineV2.extraction.command` or `synthesis.provider='command'`. If `extraction.command` is present but malformed at the YAML/serde layer (for example wrong type, wrong field types, or `command: []`), `parse_manifest_result()` returns a generic `manifest shape error`, this predicate returns false, and `from_env()` falls back to the default manifest instead of failing startup. That contradicts the PR description's fail-fast validation claim and breaks TS/Rust parity for invalid command configs.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `2a3c85c9`

[Automated Review] Found one blocking parity/validation issue in the Rust path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The Rust startup path now claims fail-fast handling for command-provider config mistakes, but the new guard only matches two specific error substrings. Any malformed `extraction.command` shape that fails during serde deserialization (for example `command: []` or wrong field types) will still be treated as a generic manifest parse error and the daemon will silently fall back to defaults instead of rejecting startup. This divergence is directly visible in `config.rs`, but I did not run the Rust binary with a malformed manifest to capture the exact runtime output.

packages/daemon-rs/crates/signet-core/src/config.rs
aaf2tbz: Fixed in b3b7af90. Rust fail-fast classification now uses raw manifest context (provider intent in `memory.pipelineV2`) instead of brittle substring matching. If manifest parsing/normalization fails and raw config indicates command-provider usage (`extraction.provider`/`extractionProvider` = `command` or synthesis provider command), startup is treated as fatal — including malformed `extraction.command` shape/type cases (e.g. `command: []`). Non-command manifest errors remain recoverable with warning/fallback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b3b7af90`

```pr-review-json
{}
```

aaf2tbz:
/signet-coding-copilot pass complete for the latest blocking parity comment.

Pushed `b3b7af90`:

- Rust command-provider fail-fast is now context-aware (raw manifest driven), not substring-only.
- `load_manifest` now classifies parse/shape/normalize failures as fatal when manifest intent is command-provider (`extraction.provider`/`extractionProvider` = `command`, or synthesis provider command), including malformed `extraction.command` shapes/types (`command: []`, wrong field types, etc.).
- Non-command manifest errors remain recoverable (warn + default fallback), preserving narrowed startup-scope behavior from the previous round.
- Added test coverage for the fail-fast classification predicate with command/non-command raw manifest cases.

Validation rerun:
- `cargo test -p signet-core config::tests:: --quiet`
- `cargo check -p signet-daemon`
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.test.ts`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon-rs/crates/signet-core/src/config.rs`

Replied inline on the new Ant thread.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `b3b7af90`

[Automated Review] Found one retry/data-corruption risk and one config-validation gap.
The PR largely implements the advertised `command` extraction path, but the new control-plane sequencing can replay external side effects on retry, and the TS config parser does not fully fail fast on malformed `command.args` despite the PR description claiming invalid command configs are rejected.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The retry hazard is directly visible in `summary-worker.ts`: `runSummaryCommandProvider()` executes before later synthesis/DAG/continuity steps that can still throw, so the existing job retry loop will rerun an already-side-effecting external command. The validation gap is directly visible in `memory-config.ts`, where malformed `command.args` entries are silently filtered instead of rejected. I could not run the daemon or the new worker tests because the workspace only contained the embedded PR snapshot.


- packages/daemon/src/pipeline/summary-worker.ts:2203 - The new `command` extractor runs before the later summary/DAG/continuity work that can still fail and send the job back through the existing retry path. That means an external command can successfully write memories, then the job can fail on a later LLM/DAG step and be retried, replaying the external side effects with no idempotency guarantee. This is a real duplicate-write/data-corruption risk introduced by keeping the old retry lifecycle around a side-effecting subprocess.
- packages/daemon/src/memory-config.ts:1659 - This parser silently drops non-string `command.args` entries instead of rejecting the config. That diverges from the PR description's 'fail-fast validation for config mistakes': a malformed config like `args: [123]` will pass load-time validation and can degrade into executing just `bin` (for example bare `node`) until timeout/retry, which is much harder to diagnose than a config error.

aaf2tbz:
Pushed follow-up fix commit c24c1b62 to address latest review feedback.

- fail-fast validation for extraction.command.args (reject non-string entries)
- command-provider retry-safety hardening: post-command synthesis/control-plane failures are non-fatal in command mode to avoid replaying external side effects on retry

Validation run:
- bun test packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/summary-worker.test.ts
- bunx @biomejs/biome check packages/daemon/src/memory-config.ts packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/summary-worker.ts

Also dismissed stale CHANGES_REQUESTED reviews from superseded commits.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +307,363 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+					summaryChars: result.summary.length,
+				});
 
-					// Reset alpha to 1.0 (full baseline weight) and EMA to neutral
-					const { updatePredictorState: resetState } = await import("../predictor-state");
-					resetState(job.agent_id, {
-						alpha: 1.0,
-						successRate: 0.5,
-					});
+				const saved = insertSummaryFacts(accessor, job, result.facts);
```
PR-Reviewer-Ant, line 590: In command mode, failures after `runSummaryCommandProvider()` no longer participate in the job retry lifecycle. The new outer `try/catch` logs and suppresses errors from summary generation, DAG writes, continuity scoring, predictor comparison, and the synthesis trigger whenever `commandMode` is true. That is a behavioral change from the existing worker semantics, and it diverges from the PR description's claim that queue/retry/max-attempt behavior is preserved. As written, a command extraction can succeed while all configured post-command synthesis work silently fails and the job is still marked completed.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c24c1b62`

[Automated Review] Found one control-plane regression in the new command-provider path.
The PR description says command extraction preserves the existing queue/retry/max-attempt lifecycle, but the new `summary-worker` flow does not actually do that for post-command synthesis/continuity work.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The diff in `packages/daemon/src/pipeline/summary-worker.ts` shows a new command-mode branch that runs `runSummaryCommandProvider()` first, then wraps the entire downstream summary/DAG/continuity/predictor flow in a `try/catch` that suppresses failures when `commandMode` is true. That is directly visible in the changed control flow and it conflicts with the PR's claim that queue/retry/max-attempt behavior is preserved.

aaf2tbz:
/signet-coding-copilot follow-up pass complete.

Pushed 22dbd8aa to address the latest control-plane blocker and tighten end-to-end retry behavior:

- Restored normal queue/retry semantics for post-command synthesis/continuity/predictor flow (removed command-mode suppression catch).
- Added a command-stage checkpoint in summary_jobs.result:
  - mark once command extraction succeeds while job is processing
  - on retry, skip re-running external command if that stage already completed
  - this preserves retry lifecycle for downstream stages without replaying command side effects.
- Added regression tests for command-stage checkpoint behavior.

Validation rerun:
- bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.test.ts
- bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.ts packages/daemon/src/memory-config.test.ts
- (sanity) cd packages/core && bun run build
- (sanity) cd packages/daemon-rs && cargo check -p signet-daemon

Dismissed stale CHANGES_REQUESTED review from superseded commit.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +308,384 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+export function hasCommandStageCompleted(accessor: DbAccessor, jobId: string): boolean {
+	return accessor.withReadDb((db) => {
+		const row = db.prepare("SELECT result FROM summary_jobs WHERE id = ?").get(jobId) as
+			| { result: string | null }
+			| undefined;
+		return row?.result === COMMAND_STAGE_COMPLETED_RESULT;
+	});
+}
+
+export function markCommandStageCompleted(accessor: DbAccessor, jobId: string): void {
```
PR-Reviewer-Ant, line 499: Blocking: the retry flow for `extraction.provider = "command"` is now stateful in a way that can skip the remainder of the job lifecycle. On the first attempt, `runSummaryCommandProvider()` can write memories, then a later step (for example synthesis/continuity/DAG work) can fail. The retry comes back through `processJob()`, but the very first thing it does is recompute `passesSignificanceGate()`. Because the command already mutated the DB, novelty/entity-overlap can change and the retry can now return early as 'not significant', which means the remaining stages are silently skipped forever. That diverges from the PR description's claim that the existing lifecycle/retry behavior is preserved. The gate needs to be evaluated once before the command side effects, or its decision persisted alongside the command-stage marker so retries resume the post-command stages deterministically.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `22dbd8aa`

[Automated Review] Found one retry-path logic regression in the new command extraction flow.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new `command` path marks the DB as mutated before the rest of the summary-worker lifecycle completes, but retries still re-run `passesSignificanceGate()` from scratch at the top of `processJob()`. Because significance depends on the current database state, a first attempt that already wrote memories via the external command can make the retry appear non-significant, causing the retry to exit early and permanently skip DAG/continuity/synthesis follow-up stages. This behavior is directly implied by the new control flow in `summary-worker.ts`.

aaf2tbz:
/signet-coding-copilot final follow-up pass complete.

Pushed commit `26beebb7` to fix the new retry-path blocker and stop the review loop.

What changed:
- Added `shouldRunSignificanceGateForJob(commandMode, commandStageCompleted)`.
- In command mode, once a job is checkpointed as `command-stage-complete`, retries now **skip significance re-evaluation** and continue downstream lifecycle.
- This prevents post-command DB mutations from making retries look non-significant and accidentally skipping DAG/continuity/synthesis follow-up.
- Existing protections remain:
  - command stage runs once and is checkpointed
  - retries do not replay external command side effects

Tests added/updated:
- `shouldRunSignificanceGateForJob` behavior matrix
- existing command-stage checkpoint tests retained

Validation rerun:
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.test.ts`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts`

Also updated PR summary text to document the command-stage checkpoint + significance-gate retry behavior.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +308,395 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+}
 
-	const saved = insertSummaryFacts(accessor, job, result.facts);
-
-	logger.info("summary-worker", "Inserted session facts", {
-		total: result.facts.length,
-		saved,
-		deduplicated: result.facts.length - saved,
-		factsPreview: result.facts.slice(0, 10).map((fact) => fact.content),
+export function markCommandStageCompleted(accessor: DbAccessor, jobId: string): void {
```
PR-Reviewer-Ant, line 499: The new command-stage checkpoint is not atomic with the external command side effects. `runSummaryCommandProvider()` completes first, then `markCommandStageCompleted()` writes the sentinel afterward. If the daemon crashes between those two operations, the next retry will not see `command-stage-complete` and will execute the command again, which can duplicate writes or other side effects. That undermines the stated goal of preventing re-runs across retries.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `26beebb7`

[Automated Review] Found a retry/checkpoint race in the new command-provider path that can still duplicate external side effects after a worker crash.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new checkpoint is written only after `runSummaryCommandProvider()` returns successfully. In `packages/daemon/src/pipeline/summary-worker.ts`, the external command runs first and `markCommandStageCompleted()` is called in a separate step afterward. If the worker crashes or the process is terminated in that gap, the retry path will not see the checkpoint and will re-run the external command, despite the PR description claiming retries skip external side effects for that job.

aaf2tbz:
/signet-coding-copilot follow-up pass complete.

Pushed commit `210587f2` to fix the new crash-gap blocker in command mode.

What changed:
- Introduced explicit command-stage statuses in `summary_jobs.result`:
  - `command-stage-running`
  - `command-stage-complete`
- Command mode now checkpoints `running` **before** launching external command.
- On explicit command failure (non-zero/timeout/error), `running` is cleared so retries can re-attempt command.
- On command success, status is promoted to `complete`.
- Retry behavior:
  - if status is `running` or `complete`, command is not rerun (prevents duplicate side effects after crash/recovery)
  - significance gate is skipped once a checkpoint exists so downstream DAG/continuity/synthesis follow-up still runs.

Tests added/updated:
- significance-gate helper matrix for `none|running|complete`
- command-stage state transition tests (`none -> running -> complete`)
- processing-state guard + running-clear behavior test

Validation rerun:
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.test.ts`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts`

Dismissed the stale CHANGES_REQUESTED review from the superseded commit.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -306,217 +310,442 @@ function parseLlmResponse(raw: string): LlmSummaryResult | null {
...
+						child.kill("SIGKILL");
+					} catch {
+						// Child is already gone.
+					}
+				}, 2000);
+			}, timeoutMs);
+
+			child.on("error", (err) => {
+				clearKillTimer();
+				if (settled) return;
```
PR-Reviewer-Ant, line 428: Blocking: a job retried with `result = 'command-stage-running'` will never re-run the external extractor. In `processJob()`, the `commandStageStatus === "running"` branch only logs and continues, and the job can still be marked complete afterward. If the daemon crashes or the lease is recovered after `markCommandStageRunning()` but before `markCommandStageCompleted()`, the side effect never happens yet all downstream work proceeds as if it did. This contradicts the PR's checkpointing goal and can silently drop extraction for that session.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `210587f2`

[Automated Review] Found one retry/checkpointing bug in the new command-provider path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The failure mode is directly visible in the new command-stage checkpoint logic in packages/daemon/src/pipeline/summary-worker.ts. A persisted `command-stage-running` marker causes subsequent retries to skip the external command entirely, and there is no recovery path in this diff that clears or re-runs that stage after a crash or lease recovery.

aaf2tbz:
/signet-coding-copilot follow-up pass complete.

Pushed commit `c227a210` for the latest checkpoint/retry review.

Fix:
- Crash recovery now clears stale `command-stage-running` markers when recovering `processing`/`leased` jobs to `pending`/`dead`.
- This ensures retries will not get stuck skipping command execution forever after a crash.
- `command-stage-complete` remains preserved during recovery.

Added test:
- recovery test verifies `command-stage-running -> none` and `command-stage-complete -> complete` across crash recovery.

Validation rerun:
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/memory-config.test.ts`
- `bunx @biomejs/biome check packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/pipeline/summary-worker.test.ts`

Dismissed stale CHANGES_REQUESTED from the superseded commit.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c227a210`

[Automated Review] I did not find any new correctness or security issues worth flagging in this diff. The implementation appears to match the PR description: command extraction is validated at config load, executed via `spawn` without a shell, retried through the existing summary-worker lifecycle, and checkpointed to avoid duplicate side effects on retries.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the config parsing, worker lifecycle, retry/checkpoint logic, tests for command execution and timeout handling, and the updated docs/PR description. I could validate the intended control-plane behavior directly from the changed code, but I did not run the daemon or inspect runtime consumers of `summary_jobs.result` beyond the provided diff.

PR #372, opened 2026-03-27
fix(daemon): populate content_hash in summary facts, backfill write-back, exhausted job recovery

ddasgupta4:
## Summary

Three bugs found via live database inspection on a v0.78.4 install (4929 memories).

> **AI assistance disclosure:** Identified and implemented by George / Claude Opus 4 via OpenClaw, operating on a live install. All three bugs confirmed by direct DB query and log analysis. Human-reviewed before submission. Checked against signet-coding-copilot review patterns (github.com/aaf2tbz/signet-coding-copilot) before pushing.

---

### Bug 1 — `insertSummaryFacts` writes memories without `content_hash`

`insertSummaryFacts()` does not include `content_hash` in its INSERT. Every memory extracted from session summaries has `content_hash = NULL`.

**Impact:**
- Embedding tracker skips all null-hash rows (`listStaleEmbeddingRows` guards on `content_hash IS NOT NULL`)
- `isDuplicate` deduplication weakened (hash-based check unavailable)
- 413 of 4929 memories were invisible to vector search on the affected install

**Fix:** Compute `normalizeAndHashContent(item.content).contentHash` and include it in the INSERT.

---

### Bug 2 — Embed backfill cycles indefinitely for null-hash memories

`reembedMissingMemoriesBatch` computes a hash for null-hash rows during backfill but never writes it back to `memories.content_hash`. The `embedding-coverage.ts` audit queries use `m.content_hash IS NOT NULL` as a guard before checking the hash-match branch. Null-hash rows never satisfy this guard, so after an `ON CONFLICT` (which keeps the original `source_id`, per the fix in #319), those rows keep appearing as unembedded.

**Confirmed:** 9 consecutive batches of 50, `413 still missing` after each run.

**Fix:** After computing `contentHash` for a null-hash memory, run:
```sql
UPDATE memories SET content_hash = ? WHERE id = ? AND content_hash IS NULL
```
Prepared statement hoisted outside the loop (pattern: `db.prepare()` inside loops gets flagged in review).

---

### Bug 3 — `memory_jobs` stuck in `pending` with `attempts >= max_attempts`

The tick loop SELECT requires `attempts < max_attempts`. Jobs that reach `max_attempts` while in `pending` status are silently skipped forever. The stall detector fires every 60s with `"pending":1, "failures":0` indefinitely with no resolution path.

**Fix:** Export `recoverMemoryJobs()` following the same pattern as `recoverSummaryJobs()` in `summary-worker.ts`. Call it at `startWorker()` startup before `scheduleTick()`.

---

## Changes

| File | Change |
|------|--------|
| `packages/daemon/src/pipeline/summary-worker.ts` | Import `normalizeAndHashContent`; add `content_hash` to `insertSummaryFacts` INSERT |
| `packages/daemon/src/repair-actions.ts` | Write computed hash back to `memories.content_hash` for null-hash rows during backfill; hoist prepared statement outside loop |
| `packages/daemon/src/pipeline/worker.ts` | Export `recoverMemoryJobs()`; call at startup before `scheduleTick()` |
| `packages/daemon/src/pipeline/summary-worker.test.ts` | Regression test: `insertSummaryFacts` populates `content_hash` |
| `packages/daemon/src/pipeline/worker.test.ts` | Regression tests: `recoverMemoryJobs` marks exhausted jobs dead, no-op when empty |

---

## Readiness Checklist

> **Note:** `bun run typecheck` and `bun run lint` not run locally (environment constraint). Code follows AGENTS.md conventions: `db.prepare()` hoisted outside loops, single-word variable names, no `as` casts, `countChanges()` helper used, explicit return types on exported functions.

### DIFF
--- a/packages/daemon/src/repair-actions.ts
+++ b/packages/daemon/src/repair-actions.ts
@@ -546,6 +546,14 @@ async function reembedMissingMemoriesBatch(
 		let count = 0;
 		// Hoisted outside loop (pattern: db.prepare inside a loop is flagged)
 		const writeHash = db.prepare("UPDATE memories SET content_hash = ? WHERE id = ? AND content_hash IS NULL");
+		// Guard against unique constraint violation: idx_memories_content_hash_unique
+		// is a partial unique index on (content_hash) WHERE content_hash IS NOT NULL AND is_deleted = 0.
+		// If another non-deleted memory already owns the same hash, writing it back would throw
+		// and abort the entire batch. Skip the write-back in that case -- the dedup worker
+		// will soft-delete the duplicate in a later pass.
+		const checkHash = db.prepare(
+			`SELECT id FROM memories WHERE content_hash = ? AND is_deleted = 0 AND id <> ? LIMIT 1`,
+		);
 
 		for (const { memory, vector } of results) {
 			const contentHash =
@@ -558,7 +566,8 @@ async function reembedMissingMemoriesBatch(
 			// content_hash match branch for these rows, so they keep showing up
 			// as unembedded and the backfill cycles indefinitely.
 			if (!memory.contentHash) {
-				writeHash.run(contentHash, memory.id);
+				const collision = checkHash.get(contentHash, memory.id) as { id: string } | undefined;
+				if (!collision) writeHash.run(contentHash, memory.id);
 			}
 
 			const embId = crypto.randomUUID();
--- a/packages/daemon/src/repair-actions.test.ts
+++ b/packages/daemon/src/repair-actions.test.ts
@@ -524,6 +524,12 @@ describe("reembedMissingMemories", () => {
 		// embedding-coverage query could not use the hash-match branch (because
 		// m.content_hash IS NULL), so the memory kept appearing as unembedded
 		// and the backfill cycled indefinitely.
+		//
+		// Test with the unique index in place to exercise the production code path.
+		db.exec(
+			`CREATE UNIQUE INDEX IF NOT EXISTS idx_memories_content_hash_unique
+			 ON memories(content_hash) WHERE content_hash IS NOT NULL AND is_deleted = 0`,
+		);
 		insertMemory(db, "mem-write-back");
 		const before = db.prepare("SELECT content_hash FROM memories WHERE id = 'mem-write-back'").get() as {
 			content_hash: string | null;
@@ -562,6 +568,52 @@ describe("reembedMissingMemories", () => {
 			false,
 		);
 		expect(second.message).toMatch(/no unembedded memories found/);
+	});
+
+	it("does not throw when a duplicate-content null-hash memory collides with an existing hashed memory", async () => {
+		// Regression: the write-back ran unconditionally, causing a UNIQUE constraint
+		// violation when another non-deleted memory already owned the same content_hash.
+		// That aborted the entire batch, so the cycle never resolved.
+		// With the unique index active (production path), the write-back must be skipped
+		// for the duplicate and the batch must complete without throwing.
+		db.exec(
+			`CREATE UNIQUE INDEX IF NOT EXISTS idx_memories_content_hash_unique
+			 ON memories(content_hash) WHERE content_hash IS NOT NULL AND is_deleted = 0`,
+		);
+		const now = new Date().toISOString();
+		const { contentHash: hash } = normalizeAndHashContent("duplicate content for collision test");
+
+		// Memory that already owns the hash
+		db.prepare(
+			`INSERT INTO memories (id, content, content_hash, type, created_at, updated_at, updated_by)
+			 VALUES (?, ?, ?, 'fact', ?, ?, 'test')`,
+		).run("mem-owner", "duplicate content for collision test", hash, now, now);
+
+		// Null-hash memory with identical content -- this is the one that would collide
+		db.prepare(
+			`INSERT INTO memories (id, content, type, created_at, updated_at, updated_by)
+			 VALUES (?, ?, 'fact', ?, ?, 'test')`,
+		).run("mem-dupe", "duplicate content for collision test", now, now);
+
+		const limiter = createRateLimiter();
+		// Must not throw
+		const result = await reembedMissingMemories(
+			accessor,
+			TEST_CFG,
+			CTX_OPERATOR,
+			limiter,
+			async () => [0.1, 0.2, 0.3],
+			TEST_EMBEDDING_CFG,
+			10,
+			false,
+		);
+		expect(result.success).toBe(true);
+
+		// Duplicate's hash stays null -- dedup worker will clean it up later
+		const dupe = db.prepare("SELECT content_hash FROM memories WHERE id = 'mem-dupe'").get() as {
+			content_hash: string | null;
+		};
+		expect(dupe.content_hash).toBeNull();
 	});
 
 	it("syncs vec row using canonical embedding id on hash conflict", async () => {


PR-Reviewer-Ant:
Hi @ddasgupta4 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): populate content_hash in summary facts, backfill write-back, exhausted job recovery` (commit `e88fbdd2`) now and I'll follow up shortly with feedback.

packages/daemon/src/repair-actions.ts
```diff
@@ -544,12 +544,23 @@ async function reembedMissingMemoriesBatch(
 	const written = accessor.withWriteTx((db) => {
 		const now = new Date().toISOString();
 		let count = 0;
+		// Hoisted outside loop (pattern: db.prepare inside a loop is flagged)
```
PR-Reviewer-Ant, line 547: The write-back for null `content_hash` rows looks like the intended fix, but this PR does not add the regression test the description claims for bug 2. Given the repo's incident guardrail rule ('every bug fix should include a test that would fail before the fix'), this repair path should get a focused test proving a null-hash memory stops reappearing in subsequent backfill batches after the update.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1548,6 +1580,20 @@ export function startWorker(
 		}, 0);
 	}, WATCHDOG_INTERVAL);
 
+	// Startup recovery: mark memory_jobs that are pending but have exhausted
+	// all attempts as dead. These are skipped by the tick loop (requires
+	// attempts < max_attempts) but stay in 'pending' forever, causing the
+	// stall detector to fire on every interval. Same pattern as summary-worker's
+	// recoverSummaryJobs for summary_jobs.
+	try {
```
PR-Reviewer-Ant, line 1663: The new startup recovery runs `recoverMemoryJobs()` synchronously and `recoverMemoryJobs()` does a single unbounded `UPDATE` over `memory_jobs`. That is a different recovery pattern from `recoverSummaryJobs()`, which was explicitly batched off the startup path to avoid blocking readiness on large tables. On bigger installs, this can hold the DB write lock during worker startup and delay daemon readiness. If the intent is parity with summary job recovery, this should be bounded/deferred rather than done in one startup transaction.

packages/daemon/src/repair-actions.ts
```diff
@@ -544,12 +544,23 @@ async function reembedMissingMemoriesBatch(
...
+		const writeHash = db.prepare("UPDATE memories SET content_hash = ? WHERE id = ? AND content_hash IS NULL");
 
 		for (const { memory, vector } of results) {
 			const contentHash =
 				typeof memory.contentHash === "string" && memory.contentHash.trim().length > 0
 					? memory.contentHash
 					: normalizeAndHashContent(memory.content).contentHash;
+
+			// Write computed hash back to the memories row when it was NULL.
+			// Without this, the embedding-coverage queries can never use the
```
PR-Reviewer-Ant, line 565: Blocking: writing the computed hash back unconditionally can fail on the duplicate-content case this PR is trying to repair. If another memory already owns the same `content_hash` under the normal `idx_memories_content_hash_unique` constraint, `UPDATE memories SET content_hash = ?` will throw before the embedding `ON CONFLICT` logic runs, aborting the whole repair batch. The new regression test drops that index, so it does not cover the production path. This needs conflict-safe handling (for example, tolerate/update-skip on hash collision) plus a test with the unique index left in place.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `8c9d9258`

[Automated Review] The backfill write-back fix introduces a duplicate-hash failure path that can abort the repair on exactly the duplicate-content cases described in the PR.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new `UPDATE memories SET content_hash = ?` runs before the existing embedding dedupe/ON CONFLICT path and the test suite explicitly drops `idx_memories_content_hash_unique`, which indicates production normally enforces uniqueness on `memories.content_hash`. On installs where a NULL-hash row duplicates an existing hashed memory, this update will raise a UNIQUE constraint error and abort the batch instead of fixing the cycle.

NicholaiVogel:
Hi @ddasgupta4 -- automated check-in from the review bot.

**Merge status:** No actual conflicts. The branch merges cleanly with main. GitHub just needed time to re-evaluate -- you should see a clean merge indicator now.

**What's blocking merge** (from the prior `CHANGES_REQUESTED`):

The write-back at `repair-actions.ts` -- `writeHash.run(contentHash, memory.id)` -- can still throw `UNIQUE constraint failed` on the `idx_memories_content_hash_unique` partial index when two null-hash memories in the same batch share identical content. Since the whole batch runs inside a single `withWriteTx`, one collision aborts everything. The repair cycles, same as before.

The test suite drops that index in `beforeEach`, so it doesn't exercise this path. This is the exact scenario the backfill is supposed to fix (duplicate content from pre-hash installs), so it'll hit on real-world data.

**Suggested fix** -- guard the write-back with an existence check:

```typescript
if (!memory.contentHash) {
  const existing = db
    .prepare(
      `SELECT id FROM memories
       WHERE content_hash = ? AND is_deleted = 0 AND id <> ?
       LIMIT 1`
    )
    .get(contentHash, memory.id) as { id: string } | undefined;
  if (!existing) writeHash.run(contentHash, memory.id);
  // If existing: genuine duplicate, dedup worker will clean it up
}
```

Then update the regression test to **not** drop `idx_memories_content_hash_unique`, and add a second case that inserts two memories with identical content and confirms neither crashes.

Everything else (bug 1, bug 3, their tests) looks good. This one change and the test update should clear the `CHANGES_REQUESTED`.

NicholaiVogel:
Maintainer follow-up (commit `55bd9ac3`): applied the fix directly to the branch.

The blocking concern from the prior `CHANGES_REQUESTED` is resolved. Added an existence check before writing `content_hash` back in `reembedMissingMemoriesBatch`:

```typescript
const collision = checkHash.get(contentHash, memory.id) as { id: string } | undefined;
if (!collision) writeHash.run(contentHash, memory.id);
```

If another non-deleted memory already owns the same hash, the write-back is skipped. The dedup worker handles cleanup in a later pass.

Test updates: the existing bug-2 regression test now restores `idx_memories_content_hash_unique` so it exercises the production code path. A second test case covers the identical-content collision scenario and confirms no crash.

**All 36 repair-actions tests pass.** The 2 `worker.test.ts` failures pre-exist on `main` and are unrelated to this PR. Ready for human review.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -916,6 +916,38 @@ function enqueueStructuralJob(db: WriteDb, memoryId: string, jobType: string, pa
 	).run(id, memoryId, jobType, payload, now, now);
```
PR-Reviewer-Ant, line 916: This PR changes daemon pipeline behavior, but there is no corresponding update under `packages/daemon-rs/` even though the repo’s Rust parity rule says daemon behavioral changes must be mirrored there. As written, the JS worker will recover exhausted jobs and the JS summary/repair paths will populate `content_hash`, while the Rust shadow implementation will continue with the old behavior and log false divergences. Please mirror these fixes in the Rust daemon or explicitly scope why parity is not required for these paths.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `55bd9ac3`

[Automated Review] Found one behavioral parity issue that should be fixed before human review is complete.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The diff clearly changes JS daemon behavior in `summary-worker.ts`, `repair-actions.ts`, and `worker.ts`, and the repository guidance explicitly requires matching updates in `packages/daemon-rs/` for daemon behavior changes. No Rust-side files are touched in this PR, so the shadow daemon will retain the old behavior for summary-fact hashing, null-hash backfill, and exhausted-job recovery. I did not inspect the Rust pipeline implementation here, so I cannot name the exact Rust files that need the corresponding patches.

NicholaiVogel:
Rust parity applied in maintainer commit `2d88ef28` on main.

The three JS changes in this PR have corresponding daemon-rs treatment:

- **Bug 3 (startup recovery)** -- `recoverMemoryJobs()` equivalent added to `signet-pipeline/src/worker.rs`'s `worker_loop` startup. Same behavior: UPDATE memory_jobs SET status = 'dead' WHERE status = 'pending' AND attempts >= max_attempts.

- **Bug 1 (summary facts content_hash)** -- `signet-pipeline/src/summary.rs` has a `TODO: Phase 5.3` stub (fact insertion not yet implemented in Rust). Added a parity note to include `content_hash` when that's wired up.

- **Bug 2 (backfill write-back)** -- `routes/repair.rs` re_embed is `NOT_IMPLEMENTED`. Added a parity note for the collision guard when it lands.

This PR is now clear from my end. CI is green, no open blocking concerns.

PR #373, opened 2026-03-27
fix(daemon): persist extraction fallback status with provider parity (replaces #367)

### LINKED ISSUE #320: feat(daemon): provider preflight check at startup + explicit fallback control

aaf2tbz:
## Summary
This PR replaces #367 with a single clean commit on a fresh branch.

Implements issue #320 by:
- persisting extraction provider runtime status in `/api/status` (including explicit disabled/paused/degraded/blocked metadata)
- adding `memory.pipelineV2.extraction.fallbackProvider` (`ollama | none`) with strict validation
- dead-lettering startup-pending extraction jobs when startup preflight blocks extraction
- enforcing daemon-rs `/api/memory/remember` blocked-extraction handling atomically in the insert transaction
- keeping daemon-rs behavior aligned with JS daemon parity for status/config/dead-letter flow
- ensuring unsupported providers are treated consistently in preflight and worker startup (prevents `extraction_probe()` reporting active while worker refuses to start)
- restricting daemon-rs credentialed startup preflight probes to trusted Anthropic hosts via `SIGNET_TRUSTED_PROVIDER_ENDPOINT_HOSTS`

Closes #320.

## Changes
- JS daemon: persisted extraction runtime state and startup-block dead-letter behavior
- Rust daemon parity: status/config/preflight/worker behavior alignment, including atomic blocked-remember enforcement
- CLI runtime/health: surfaced degraded vs blocked extraction state
- Docs/specs: API, config, pipeline docs + completed spec note
- Tests: added/updated daemon, daemon-rs, and CLI coverage for new behavior

## Type
- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected
- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs/specs, daemon-rs parity

## Screenshots
N/A (no UI changes requiring screenshots)

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

Compatibility note:
- Non-loopback Anthropic endpoint overrides now require an explicit allowlist entry in `SIGNET_TRUSTED_PROVIDER_ENDPOINT_HOSTS` before daemon-rs sends provider credentials during startup preflight. Official Anthropic hosts remain trusted by default.
- OpenRouter remains unsupported by the daemon-rs extraction worker and is blocked/fallbacked before any provider-specific startup probe runs.

## Testing
- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Validation commands run:
- `cargo test -p signet-daemon --manifest-path packages/daemon-rs/Cargo.toml routes::write -- --nocapture`
- `cargo test -p signet-daemon --manifest-path packages/daemon-rs/Cargo.toml worker_supports_only_ollama_and_anthropic_providers`
- `cargo test -p signet-daemon --manifest-path packages/daemon-rs/Cargo.toml resolve_runtime_extraction_`
- `cargo test -p signet-daemon --manifest-path packages/daemon-rs/Cargo.toml pause_`
- `bun run build:core`
- `bun test packages/daemon/src/daemon-status.test.ts packages/daemon/src/memory-config.test.ts packages/daemon/src/pipeline/extraction-fallback.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/runtime.test.ts`

## AI disclosure

## Notes
- Repo-wide `bun run typecheck` currently fails in untouched package `@signet/opencode-plugin` due missing Node type definitions in this environment.
- Repo-wide `bun run lint` currently reports existing formatting diagnostics outside this PR scope.
- `cargo fmt` could not be run in this environment because `rustfmt` is not installed for `stable-aarch64-apple-darwin`.
- Touched-path validations above are passing.


BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `daf5a525`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `daf5a525`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +612,518 @@ async fn shutdown_signal() {
...
+        .unwrap_or(false)
+}
+
+/// Check if the OpenCode HTTP server is reachable.
+async fn check_opencode_health(endpoint: Option<&str>) -> bool {
+    let base = endpoint
+        .filter(|s| !s.is_empty())
+        .unwrap_or("http://127.0.0.1:4096");
+    let url = format!("{}/health", base.trim_end_matches('/'));
+    let client = reqwest::Client::builder()
```
PR-Reviewer-Ant, line 1048: The new startup preflight expands credential exposure: `check_anthropic_health()` sends `ANTHROPIC_API_KEY` to any configured `endpoint`, and `check_openrouter_health()` does the same with `OPENROUTER_API_KEY`. Because these probes now run automatically at daemon boot/resume, a malicious or mistyped custom endpoint can exfiltrate provider credentials before any extraction job is processed. Please restrict authenticated probes to trusted/default hosts or add explicit validation before sending provider secrets to arbitrary endpoints.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `daf5a525`

[Automated Review] Found one security issue worth fixing before merge.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new startup preflight code in daemon-rs now performs authenticated HTTP probes during boot/resume. In `packages/daemon-rs/crates/signet-daemon/src/main.rs`, the Anthropic/OpenRouter health checks unconditionally attach real API keys to whatever custom `endpoint` is configured. That behavior is directly visible in the diff, but I did not run a live daemon against a hostile endpoint in this sandbox.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `c14db128`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +612,580 @@ async fn shutdown_signal() {
...
+            reason: Some(format!("{reason_prefix}; ollama fallback startup preflight failed")),
+            since: Some(now.clone()),
+        };
+        let full_reason = format!("{reason_prefix}; ollama fallback startup preflight failed");
+        *state.extraction_state.write().await = Some(new_state);
+        if dead_letter_on_blocked {
+            dead_letter_pending_extraction_jobs(state, &full_reason, &now).await;
+        }
+        warn!("extraction blocked: primary and ollama fallback both unavailable");
+        return;
```
PR-Reviewer-Ant, line 997: For untrusted custom Anthropic/OpenRouter endpoints, the fallback probe now returns `true` for any HTTP response because `check_http_reachability_without_auth()` only checks `send().await.is_ok()`. A 401/403/404 from a proxy or wrong path will still be reported as 'available', so `/api/status` can show extraction as active/degraded when the provider is not actually usable. That reintroduces the exact false-positive startup state this PR says it fixes. This helper should require a success status (or a narrowly defined accepted set) instead of treating mere socket reachability as provider health.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c14db128`

[Automated Review] Found one blocking correctness issue in the new daemon-rs startup probe path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new daemon-rs preflight path is directly visible in the diff. In `check_http_reachability_without_auth`, any completed HTTP request is treated as success, regardless of response status. That value feeds `check_anthropic_health` / `check_openrouter_health` for untrusted custom endpoints and can therefore mark extraction as available even when the endpoint returns 401/403/404 or a proxy error page. This is directly at odds with the PR's stated goal of preventing provider-resolution/status mismatches, although I did not execute a live repro against a running daemon.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `5213341a`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -91,7 +93,8 @@ async fn main() -> anyhow::Result<()> {
         .init();
 
     // Load config
-    let config = DaemonConfig::from_env();
+    let config = DaemonConfig::try_from_env()
```
PR-Reviewer-Ant: This changes daemon-rs startup semantics much more broadly than the PR description says. `try_from_env()` propagates any manifest parse error, and `main()` now exits on that error instead of falling back to defaults. The PR claims strict validation for `memory.pipelineV2.extraction.fallbackProvider`, but this path also turns unrelated `agent.yaml` parse issues into a hard startup failure for the Rust daemon. That is a breaking behavior change and likely parity drift unless the JS daemon was changed the same way.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +612,589 @@ async fn shutdown_signal() {
...
+/// Check if the ollama HTTP server is reachable.
+async fn check_ollama_health(endpoint: Option<&str>) -> bool {
+    let base = endpoint
+        .filter(|s| !s.is_empty())
+        .unwrap_or("http://127.0.0.1:11434");
+    let url = format!("{}/api/tags", base.trim_end_matches('/'));
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(2))
+        .build();
+    let Ok(client) = client else { return false };
```
PR-Reviewer-Ant, line 1033: For untrusted/custom Anthropic/OpenRouter endpoints, startup preflight skips auth headers and then requires a 2xx response to consider the endpoint healthy. Many legitimate custom proxies return `401`/`403` when probed without credentials, which still proves the endpoint is reachable. With the current `probe_status_is_healthy()` implementation, those endpoints will be misclassified as unavailable and extraction will incorrectly fall back or become blocked at startup.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5213341a`

[Automated Review] Found two correctness issues in the daemon-rs startup/preflight path.
The PR largely implements the described status/dead-letter/fallback behavior, but the Rust parity changes introduce a broader startup breakage than the PR description claims and can falsely block valid custom provider endpoints during preflight.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - Both findings are directly visible in the changed Rust code. `main.rs` now switches daemon-rs startup to `DaemonConfig::try_from_env()`, which returns hard errors for any manifest parse failure, not only the new fallbackProvider validation described in the PR. Separately, the new unauthenticated custom-endpoint probe path treats only 2xx as healthy, so authenticated custom Anthropic/OpenRouter-compatible gateways would be classified unavailable before serving traffic. I did not run these paths against a live daemon-rs instance in this environment.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `f3daff91`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -161,6 +209,17 @@ pub async fn remember(
                 },
             )?;
 
+            if r.duplicate_of.is_none() {
+                if let Some(reason) = blocked_extraction_reason.as_deref() {
+                    dead_letter_blocked_extraction_memory(
```
PR-Reviewer-Ant: This blocked-extraction branch assumes `transactions::ingest()` has already created an extraction job for the new memory, but in daemon-rs it currently only inserts the memory row and history entry; it does not enqueue into `memory_jobs`. As a result, `dead_letter_blocked_extraction_memory()` returns `QueryReturnedNoRows`, the transaction aborts, and `/api/memory/remember` becomes a 500 whenever extraction is startup-blocked. Either enqueue a job before dead-lettering or update the new memory row directly when blocked.

docs/API.md
```diff
@@ -115,6 +117,30 @@ composite health score derived from diagnostics.
     "autonomousEnabled": false,
```
PR-Reviewer-Ant, line 117: The new `/api/status` contract documented here is not actually implemented for the JS daemon in this PR. The diff adds CLI/runtime parsing for `providerResolution.extraction` and `pipeline.extraction`, but there are no corresponding `packages/daemon` changes, so the default JS daemon still won't emit these fields. That means the PR description's 'JS daemon: persisted extraction runtime state' claim is not met, and installs not using daemon-rs will not get the advertised behavior.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f3daff91`

[Automated Review] Found two blocking issues: the new blocked-extraction path can turn `remember` into a 500 in daemon-rs, and the PR/docs claim JS daemon status parity that this diff does not implement.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The daemon-rs `remember` path now calls `dead_letter_blocked_extraction_memory`, but the existing `signet_services::transactions::ingest` implementation only inserts the memory row with `extraction_status = 'none'` and does not create a `memory_jobs` row. That makes the new helper return `QueryReturnedNoRows`, which propagates to a 500 on blocked extraction. Separately, the diff adds CLI/docs consumers for `/api/status.providerResolution`, but there are no `packages/daemon` changes in this PR to make the JS daemon emit those fields, so the stated JS-daemon parity is not achieved by this implementation.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `eee32c63`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,591 @@ async fn shutdown_signal() {
...
+fn resolve_runtime_extraction_endpoint(
+    effective_provider: &str,
+    configured_provider: &str,
+    configured_endpoint: Option<&str>,
+) -> Option<String> {
+    if effective_provider == "ollama" && configured_provider != "ollama" {
+        return None;
+    }
+    configured_endpoint
+        .filter(|v| !v.is_empty())
```
PR-Reviewer-Ant, line 656: Blocking: the new fallback path ignores a configured custom Ollama endpoint. `resolve_runtime_extraction_endpoint()` returns `None` whenever `effective_provider == "ollama"` and `configured_provider != "ollama"`, and both startup preflight and `start_extraction_worker()` use that helper. So a config like `provider: claude-code`, `fallbackProvider: ollama`, `endpoint: http://remote-ollama:11434` will probe and run against the default local Ollama instead of the configured endpoint. That breaks the PR's fallback-provider contract and can incorrectly mark extraction as blocked/degraded even though the configured fallback endpoint is healthy.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `eee32c63`

pr-reviewer automated bot review
Found one blocking parity/config bug in the daemon-rs fallback path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The bug is directly visible in the new Rust fallback-resolution code: `resolve_runtime_extraction_endpoint()` deliberately drops the configured endpoint whenever the effective provider becomes `ollama` via fallback, and both startup preflight and worker startup call that helper. That means a configured non-default Ollama fallback endpoint is ignored in the exact path this PR adds. I did not run the daemon against a custom remote Ollama endpoint, so this is code-path evidence rather than a live repro.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `01def811`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,588 @@ async fn shutdown_signal() {
...
+fn resolve_runtime_extraction_model(
+    effective_provider: &str,
+    configured_provider: &str,
+    configured_model: &str,
+) -> String {
+    if effective_provider == "ollama" && configured_provider != "ollama" {
+        DEFAULT_OLLAMA_EXTRACTION_MODEL.to_string()
+    } else {
+        configured_model.to_string()
+    }
```
PR-Reviewer-Ant, line 642: Blocking: the new fallback endpoint resolver keeps the configured provider endpoint even after `effective_provider` changes to `ollama`. That means a config like `provider: openrouter` with `endpoint: https://openrouter.ai/api/v1` will probe/start the Ollama fallback against the OpenRouter URL, so the advertised `fallbackProvider: ollama` path cannot actually recover. The added test `resolve_runtime_extraction_endpoint_keeps_configured_endpoint_on_fallback` appears to codify the wrong behavior.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `01def811`

[Automated Review] Found one blocking correctness issue in the new daemon-rs fallback path.
The PR’s stated goal is to persist and enforce extraction fallback/block status with daemon-rs parity, but the new Rust fallback implementation misroutes Ollama fallback traffic to the originally configured provider endpoint. That breaks the advertised `fallbackProvider: ollama` behavior for non-Ollama configs that also set `extraction.endpoint`.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The bug is directly visible in the added helper `resolve_runtime_extraction_endpoint()` and in its new call sites for both startup preflight and worker startup. The helper always preserves the configured endpoint even when `effective_provider` switches to `ollama`, and the added test explicitly locks in that behavior. With a config such as `provider: openrouter` plus `endpoint: https://openrouter.ai/api/v1`, fallback probing and the started worker will target the OpenRouter base URL as if it were Ollama, so fallback cannot operate as described.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `11d7fd61`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -136,10 +177,28 @@ pub async fn remember(
         _ => "global",
```
PR-Reviewer-Ant, line 234: This reads `state.is_extraction_blocked()` before entering the write transaction, but the same PR also makes `/pause` and `/resume` mutate extraction state asynchronously. A concurrent resume can clear the block immediately after this read, yet this request will still dead-letter the newly inserted extraction job inside the transaction. That creates a real data-loss window around pause/resume transitions where valid remembers are permanently marked failed even though extraction is available again.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,593 @@ async fn shutdown_signal() {
     }
```
PR-Reviewer-Ant, line 610: The new untrusted-endpoint probe flow is not safe for credentialed providers. `check_anthropic_health` / `check_openrouter_health` now accept any custom HTTPS endpoint that returns 401/403 as 'healthy' via `check_http_reachability_without_auth()` + `probe_status_is_healthy()`, and `start_extraction_worker()` will then start the real provider client against that same configured endpoint with the API key attached. An attacker-controlled endpoint can therefore satisfy startup preflight without proving provider compatibility, then receive live extraction traffic and credentials.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `11d7fd61`

[Automated Review: pr-reviewer] Found a blocking race in the new remember-route dead-letter path and a security issue in the new startup probe logic.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The remember-route regression is directly visible in the added Rust write-path: blocked state is sampled before the transaction and can race with the new pause/resume preflight flow. The probe issue is also explicit in the diff: untrusted Anthropic/OpenRouter endpoints are treated as healthy on 401/403, and the worker startup path still uses the configured endpoint with credentials. I could not inspect the omitted JS-daemon files from this prompt artifact, so confidence is not 'high'.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `cf6ceee4`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,574 @@ async fn shutdown_signal() {
...
+        format!("{reason_prefix}; fallbackProvider is none")
+    } else {
+        reason_prefix
+    };
+    let new_state = ExtractionRuntimeState {
+        configured: Some(extraction.provider.clone()),
+        resolved: extraction.provider.clone(),
+        effective: "none".to_string(),
+        fallback_provider: fallback_provider.to_string(),
+        status: "blocked".to_string(),
```
PR-Reviewer-Ant, line 1011: The new `provider_endpoint_is_trusted_for_secret_probe()` gate turns any non-loopback Anthropic/OpenRouter `endpoint` override into a hard preflight failure, because `check_anthropic_health()` / `check_openrouter_health()` immediately return `false` when the host is not the exact vendor domain. That is a breaking change for users routing these providers through a trusted HTTPS proxy or gateway: startup will now mark extraction `blocked` or fall back to Ollama even though the worker itself still supports the configured provider. The PR description claims provider-status parity and strict fallback control, but this introduces a new endpoint-policy restriction that is neither documented nor reflected in config contracts.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `cf6ceee4`

[Automated Review] Found one blocking regression in the new extraction preflight path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new Rust preflight path now hard-rejects credentialed Anthropic/OpenRouter startup probes unless the endpoint host is loopback or the exact vendor hostname. In this diff, `extraction_probe()` treats a failed probe as provider unavailability and can block or silently reroute extraction before the worker starts. That behavior is directly visible in `packages/daemon-rs/crates/signet-daemon/src/main.rs`, but I did not run a live custom-endpoint config to demonstrate the failure end-to-end.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `48a74f27`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,603 @@ async fn shutdown_signal() {
...
+        );
+        return false;
+    }
+
+    let api_key = if effective_provider == "anthropic" {
+        std::env::var("ANTHROPIC_API_KEY")
+            .ok()
+            .filter(|k| !k.trim().is_empty())
+    } else {
+        None
```
PR-Reviewer-Ant, line 793: Race condition: `extraction_worker_stats` is updated before this caller has actually won `extraction_worker_handle`. If two callers enter `start_extraction_worker()` concurrently, the loser can overwrite the shared stats handle and then stop its own worker, leaving `/api/status` reporting stale/stopped stats while a different worker is actually running. Publish the stats handle only after the slot assignment succeeds, or serialize the whole startup path under one lock.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -160,23 +223,72 @@ pub async fn remember(
...
+                .unwrap_or_default()
+                .to_string();
+            let hash = val
+                .get("hash")
+                .and_then(|v| v.as_str())
+                .unwrap_or_default()
+                .to_string();
+            let duplicate_of = val
+                .get("duplicateOf")
+                .and_then(|v| v.as_str())
```
PR-Reviewer-Ant: This does not fully implement the PR's claimed 'blocking new remember-route extraction writes'. The ingest transaction has already committed by the time the dead-lettering runs, and if this follow-up `write_tx` fails you only log a warning and still return `200 created`. That leaves a blocked daemon able to persist new queued extraction work silently. The dead-letter/blocking step needs to be part of the same transaction/invariant, or the request should fail when the cleanup cannot be enforced.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `48a74f27`

[Automated Review] Found two blocking correctness issues in the new daemon-rs extraction fallback flow.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the added Rust control flow. In `start_extraction_worker`, the shared stats handle is published before ownership of the worker slot is won, which allows concurrent starters to leave `/api/status` pointing at the stopped loser's stats handle. In `remember`, the new blocked-extraction dead-letter step runs in a second best-effort transaction after ingest has already committed, and failures are only logged, so blocked runtimes can still leave queued extraction work behind despite the PR claiming those writes are blocked.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `c0cafe24`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -160,23 +223,79 @@ pub async fn remember(
...
+                    .await
+                    .unwrap_or_else(|| "Extraction provider unavailable".to_string());
+                let memory_id = id.clone();
+                if let Err(error) = state
+                    .pool
+                    .write_tx(Priority::High, move |conn| {
+                        dead_letter_blocked_extraction_memory(
+                            conn,
+                            &memory_id,
+                            &blocked_reason,
```
PR-Reviewer-Ant: The new blocked-extraction path is no longer atomic. `transactions::ingest(...)` commits first, then `dead_letter_blocked_extraction_memory(...)` runs in a second `write_tx`. If that second write fails, the handler returns `500` even though the memory row was already created. Clients will reasonably retry, which can produce duplicate/dedup responses for a write that actually succeeded and makes the API lie about durability. This should be folded into the original transaction or otherwise made idempotent/success-reporting after the first commit.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c0cafe24`

pr-reviewer automated bot review
Found one correctness issue in the new blocked-extraction remember flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The failure mode is directly visible in the new remember handler: the memory insert commits in one transaction, then blocked-extraction enforcement runs in a second transaction and can return 500 if that follow-up write fails. That creates a committed write with an error response, which is a concrete API correctness problem even without a runtime trace.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `a3b0b981`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -160,23 +223,80 @@ pub async fn remember(
...
+                .and_then(|v| v.as_str())
+                .unwrap_or_default()
+                .to_string();
+            let duplicate_of = val
+                .get("duplicateOf")
+                .and_then(|v| v.as_str())
+                .map(|v| v.to_string());
+            let mut blocked_enforcement_failed = false;
+            if duplicate_of.is_none() && state.is_extraction_blocked().await {
+                let blocked_reason = state
```
PR-Reviewer-Ant: This does not fully enforce the PR's new invariant. The memory insert commits in the first `write_tx`, and blocked-extraction dead-lettering happens only afterward in a second transaction. If that second transaction fails, the handler still returns `200` with only a `warning`, leaving a newly created memory with extraction not actually blocked/dead-lettered even though `state.is_extraction_blocked()` was true. That diverges from the stated behavior ('blocking new remember-route extraction writes when extraction is startup-blocked'). The dead-letter/update needs to be part of the same transaction as ingest, or the request should fail so callers do not get a false-success response.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `a3b0b981`

[Automated Review] Found one blocking correctness issue in the remember-route blocked-extraction flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The issue is directly visible in the Rust daemon diff: `remember` commits the ingest transaction first, then tries to dead-letter blocked extraction in a second write transaction, and explicitly downgrades failure to a warning response. That creates a clear path where `/api/hooks/remember` or `/api/memory/remember` returns 200 while leaving a newly created memory in the blocked startup state without the promised dead-letter enforcement.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `4ad5ed6f`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,604 @@ async fn shutdown_signal() {
...
+    let Ok(client) = client else { return false };
+    client
+        .get(&url)
+        .header("authorization", format!("Bearer {api_key}"))
+        .send()
+        .await
+        .map(|r: reqwest::Response| r.status().is_success())
+        .unwrap_or(false)
+}
+
```
PR-Reviewer-Ant: The new startup dead-letter routine silently ignores per-memory update failures (`filter_map(|r| r.ok())` when collecting IDs, and `let _ = update_mem.execute(...)` when marking memories failed). That can leave `memory_jobs` moved to `dead` while some `memories.extraction_status` rows stay unchanged, and the function will still log success. This creates partial state drift exactly in the recovery path this PR is trying to harden.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `4ad5ed6f`

[Automated Review] Found one correctness risk in the new startup dead-letter path; otherwise the implementation appears to match the PR description from the diff provided.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The review covered the PR description, the changed daemon-rs startup/preflight logic, the Rust write-path changes for atomic remember handling, and the CLI status surfacing. I found a directly visible silent-failure path in the new dead-lettering code, but I could not run the daemon or inspect the full repository beyond the supplied patch context.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `7c839a6a`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +621,602 @@ async fn shutdown_signal() {
...
+    let runtime_endpoint = resolve_runtime_extraction_endpoint(
+        &effective_provider,
+        &pipeline.extraction.provider,
+        pipeline.extraction.endpoint.as_deref(),
+    );
+    let provider_cfg = signet_pipeline::provider::LlmProviderConfig {
+        provider: effective_provider.clone(),
+        model: runtime_model,
+        base_url: runtime_endpoint,
+        api_key,
```
PR-Reviewer-Ant, line 815: Unsupported providers are still probed before the worker-support check. In particular, the `openrouter` branch can make a credential-bearing `/models` request via `check_openrouter_health()` and only then get forced to unavailable by `worker_supports_extraction_provider()`. That expands attack surface unnecessarily and diverges from the PR's claim that unsupported providers are handled consistently between preflight and worker startup. The support check needs to happen before any provider-specific probe for providers the worker can never run.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `7c839a6a`

[Automated Review] Found one blocking security/parity issue in the daemon-rs extraction preflight.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new daemon-rs preflight flow is fully present in this diff. In `extraction_probe()`, provider-specific probes run before the worker-support gate. For `openrouter`, that means `check_openrouter_health()` can send the bearer token to the configured endpoint even though `worker_supports_extraction_provider()` will always reject OpenRouter immediately afterward. This is directly visible in the added `main.rs` logic and contradicts the PR's unsupported-provider parity claim.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `adbe940e`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `adbe940e`

[Automated Review] Found one blocking parity bug in the daemon-rs startup preflight path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new daemon-rs preflight logic in main.rs short-circuits on `!worker_supports_extraction_provider(provider)` before the `match provider` that calls `check_openrouter_health()`. That makes the newly added OpenRouter trusted-host preflight path unreachable for `extraction.provider = openrouter`, directly contradicting the PR description's claim about allowing trusted OpenRouter proxy hosts during credentialed startup preflight.


- packages/daemon-rs/crates/signet-daemon/src/main.rs:1969 - This early `!worker_supports_extraction_provider(provider)` branch makes the later `"openrouter" => check_openrouter_health(...)` arm unreachable. As written, any `extraction.provider: openrouter` config is forced into the unsupported-provider fallback/block path without ever running the new trusted-endpoint preflight logic, so the PR does not actually implement the advertised OpenRouter proxy-host allowance in daemon-rs.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `9c04f107`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,567 @@ async fn shutdown_signal() {
...
+    configured_provider: &str,
+    configured_endpoint: Option<&str>,
+) -> Option<String> {
+    if effective_provider == "ollama"
+        && matches!(configured_provider, "anthropic" | "openrouter")
+    {
+        return None;
+    }
+    configured_endpoint
+        .filter(|v| !v.is_empty())
```
PR-Reviewer-Ant, line 656: This new gate changes daemon-rs behavior much more broadly than the PR description/documentation says: `worker_supports_extraction_provider()` only allows `ollama` and `anthropic`, so configs using `claude-code`, `codex`, or `opencode` now always enter degraded/blocked startup resolution and can dead-letter pending extraction jobs even if the configured provider is healthy. The PR text only calls out OpenRouter as unsupported, but this implementation effectively hard-disables three more documented providers in daemon-rs. That is a breaking behavior change and a data-loss risk for queued jobs.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `9c04f107`

[Automated Review] Found one blocking parity/regression issue in daemon-rs provider handling.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new daemon-rs startup path explicitly gates supported extraction providers through `worker_supports_extraction_provider`, and the added test in this PR confirms only `ollama` and `anthropic` are accepted. That makes the regression directly visible in the diff, although I could not run the daemon in this environment to reproduce it against a live config.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `ce63e67f`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,571 @@ async fn shutdown_signal() {
...
+    configured_provider: &str,
+    configured_model: &str,
+) -> String {
+    if effective_provider == "ollama" && configured_provider != "ollama" {
+        DEFAULT_OLLAMA_EXTRACTION_MODEL.to_string()
+    } else {
+        configured_model.to_string()
+    }
+}
+
```
PR-Reviewer-Ant, line 644: Blocking: Ollama fallback can be incorrectly marked unavailable when the configured provider is `opencode` with a custom `endpoint`. `resolve_runtime_extraction_endpoint()` only drops the configured endpoint for `anthropic`/`openrouter`, so fallback from `opencode` reuses the OpenCode base URL and then `check_ollama_health()` probes `<opencode-endpoint>/api/tags`. That is not an Ollama endpoint, so daemon-rs will report `blocked`/dead-letter work even when Ollama is actually healthy. This breaks the PR’s stated JS parity for fallback resolution.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `ce63e67f`

[Automated Review] Found one blocking parity/correctness issue in daemon-rs fallback resolution.
The PR broadly implements the described status persistence and blocked/dead-letter flow, but daemon-rs still diverges from the intended fallback behavior for one provider class.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The fallback endpoint bug is directly visible in the new daemon-rs logic: `resolve_runtime_extraction_endpoint()` preserves the configured endpoint for non-Anthropic/OpenRouter providers, and `extraction_probe()` reuses that value for Ollama fallback health checks. That makes fallback from `opencode` probe the OpenCode base URL as if it were an Ollama server. I could not inspect the full JS parity implementation in this environment, so confidence is medium rather than high.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `e2fe30dc`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,571 @@ async fn shutdown_signal() {
...
+fn resolve_runtime_extraction_endpoint(
+    effective_provider: &str,
+    configured_provider: &str,
+    configured_endpoint: Option<&str>,
+) -> Option<String> {
+    if effective_provider == "ollama"
+        && matches!(configured_provider, "anthropic" | "openrouter" | "opencode")
+    {
+        return None;
+    }
```
PR-Reviewer-Ant, line 654: Blocking: this still leaves a status/worker mismatch for `claude-code`, `codex`, and `opencode`. Preflight treats only `openrouter` as unsupported, so those providers can resolve to `active` when their probe succeeds, but `start_extraction_worker()` immediately refuses to run anything except `ollama`/`anthropic`. That recreates the exact bug the PR says it fixes: `/api/status` can report extraction as active while no worker is running and queued work is neither fallbacked nor dead-lettered. Preflight needs to treat all worker-unsupported providers consistently, or worker startup failure must downgrade persisted runtime state.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `e2fe30dc`

[Automated Review by pr-reviewer] Found one blocking parity bug in the daemon-rs extraction startup flow.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The bug is directly visible in the changed Rust code: `provider_is_unsupported_for_daemon_startup_preflight()` only blocks `openrouter`, while `start_extraction_worker()` later refuses to start for every provider except `ollama` and `anthropic`. That means healthy `claude-code`, `codex`, or `opencode` preflight checks can still leave `/api/status` reporting an active provider even though no extraction worker can run.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `8786bfb3`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,643 @@ async fn shutdown_signal() {
...
+
+    let runtime_model = resolve_runtime_extraction_model(
+        &effective_provider,
+        &pipeline.extraction.provider,
+        &pipeline.extraction.model,
+    );
+    let runtime_endpoint = resolve_runtime_extraction_endpoint(
+        &effective_provider,
+        &pipeline.extraction.provider,
+        pipeline.extraction.endpoint.as_deref(),
```
PR-Reviewer-Ant, line 809: The new spec and PR description say extraction runtime resolution should persist `disabled`/`paused`/`blocked`/`degraded` state in `/api/status`, but this startup probe returns immediately when the pipeline is disabled, paused, or `provider == "none"` without updating `state.extraction_state`. Since the status route only emits `providerResolution.extraction` from that persisted state, these startup modes can still surface as `null` or stale prior state instead of an explicit `disabled`/`paused` resolution.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,643 @@ async fn shutdown_signal() {
...
+        .filter(|s| !s.is_empty())
+        .unwrap_or("http://127.0.0.1:4096");
+    let url = format!("{}/health", base.trim_end_matches('/'));
+    let client = reqwest::Client::builder()
+        .timeout(std::time::Duration::from_secs(2))
+        .build();
+    let Ok(client) = client else { return false };
+    client
+        .get(&url)
+        .send()
```
PR-Reviewer-Ant, line 1054: This trusted-endpoint gate only recognizes Anthropic. The PR description and configuration docs both say `SIGNET_TRUSTED_PROVIDER_ENDPOINT_HOSTS` applies to `Anthropic/OpenRouter endpoint overrides`, but OpenRouter is never eligible here because `provider_endpoint_is_trusted_for_secret_probe()` hardcodes `_ => false` and startup preflight blocks OpenRouter before any provider probe. Either the description/docs need to stop claiming OpenRouter support here, or the implementation needs to match that contract.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `8786bfb3`

[Automated Review] Found one contract-level status persistence gap and one behavioral mismatch with the PR description.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The diff shows `extraction_probe()` returning early for disabled/none/paused startup states without mutating `state.extraction_state`, while `/api/status` serializes only that persisted state. That is directly visible in `packages/daemon-rs/crates/signet-daemon/src/main.rs`. I could not inspect the full `AppState` initializer from the workspace snapshot, so there is some uncertainty about any default seeding done outside this diff.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `4a503a77`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,653 @@ async fn shutdown_signal() {
...
+            .unwrap_or_else(|| pipeline.extraction.provider.clone())
+    };
+
+    if effective_provider == "none" {
+        return false;
+    }
+
+    if !worker_supports_extraction_provider(&effective_provider) {
+        let now = chrono::Utc::now().to_rfc3339();
+        let reason_prefix = format!(
```
PR-Reviewer-Ant, line 704: The new startup path can still leave extraction jobs stuck in `pending` even though the PR claims blocked startup states are dead-lettered. `preflight_extraction()` only dead-letters when the probe itself marks the provider blocked, but `start_extraction_worker()` has separate blocked branches for unsupported providers / unavailable fallback and just returns `false` after updating `extraction_state`. A config like `provider: claude-code` with `fallbackProvider: none` reaches this path after a successful CLI preflight, so `/api/status` becomes `blocked` while queued extraction jobs remain pending indefinitely.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -595,6 +610,653 @@ async fn shutdown_signal() {
...
+    let host = host.to_ascii_lowercase();
+    if let Some(suffix) = candidate.strip_prefix("*.") {
+        return host.len() > suffix.len()
+            && host.ends_with(suffix)
+            && host.as_bytes()[host.len() - suffix.len() - 1] == b'.';
+    }
+    host == candidate
+}
+
+fn host_in_trusted_override_list(host: &str, overrides_csv: Option<&str>) -> bool {
```
PR-Reviewer-Ant, line 1104: The new `cli_preflight()` runs `claude --version` / `codex --version` synchronously with no timeout, and startup now awaits this before serving requests. A broken wrapper or hanging binary on `PATH` can wedge daemon boot or `/api/pipeline/resume` indefinitely. The HTTP probes use a 2s timeout, but the CLI probe does not, so this introduces an unbounded startup/resume hang path.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `4a503a77`

[Automated Review] Found two blocking correctness issues in the new daemon-rs startup/preflight path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff shows explicit blocked-return branches in daemon-rs startup that update runtime state without invoking the new dead-letter helper, and a new synchronous CLI preflight that executes `--version` with no timeout before serving requests. Those behaviors are directly visible in `packages/daemon-rs/crates/signet-daemon/src/main.rs`, but I could not run the daemon here to capture a live repro trace.

packages/daemon-rs/crates/signet-daemon/src/routes/pipeline.rs
```diff
@@ -194,6 +194,33 @@ async fn apply_pause_state(state: &AppState, paused: bool) {
...
+        let mut guard = state.extraction_state.write().await;
+        if let Some(es) = guard.as_mut() {
+            if es.status != "disabled" && es.status != "blocked" {
+                es.status = "paused".to_string();
+                es.effective = "none".to_string();
+                es.degraded = false;
+                es.fallback_applied = false;
+                es.reason = None;
+                es.since = None;
+            }
```
PR-Reviewer-Ant, line 215: Blocking: this resume flow still has the false-`active` race the PR says it eliminates. `resume_extraction_check()` can mark `claude-code` / `codex` / `opencode` as available, then `start_extraction_worker()` immediately downgrades or blocks because the worker only supports `ollama` and `anthropic`. Since those are separate awaited steps, concurrent `/api/status` calls can observe `active` even though the worker cannot start. If the contract is that provider resolution reflects actual runtime execution state, the status transition and worker start need to be atomic, or preflight needs to classify every non-worker-backed provider as unavailable before publishing `active`.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `57c61f51`

[Automated Review] Found one runtime-status race that leaves the PR short of its stated guarantee.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new resume path in daemon-rs updates extraction status and starts the worker in two separate awaited steps, while the new startup-preflight support matrix still treats `claude-code`, `codex`, and `opencode` as probeable even though `worker_supports_extraction_provider()` only allows `ollama` and `anthropic`. From the diff alone, that creates an observable window where `/api/status` can report `active` for a provider the worker still cannot run.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `57c61f51`) now and I'll follow up shortly with feedback.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `642b0bfd`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `642b0bfd`

[Automated Review] I did not find any new blocking correctness or security issues in the supplied diff; the PR appears ready for human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The supplied diff is detailed enough to verify the main claims called out in the PR description: persisted extraction runtime state in `/api/status`, CLI surfacing of degraded/blocked state, startup dead-lettering for blocked extraction, stricter fallbackProvider validation, and the daemon-rs remember-path dead-lettering inside the DB transaction. I could not inspect untouched source files such as the full `state` module or the omitted JS daemon implementation from a live checkout, so confidence is not high, but I did not find a concrete defect provable from the provided changes.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `caf8063d`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `caf8063d`

[Automated Review by pr-reviewer] Found one blocking data-integrity issue in the new dead-letter flow and one CLI regression-test gap.
The PR largely matches its stated goals, but the startup blocked-extraction dead-lettering can overwrite successful extraction state for memories that still have a stray pending job.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The blocking issue is directly visible in the new SQL/update logic: both new dead-letter helpers mark `memories.extraction_status = 'failed'` after checking only for leased jobs, not for already-completed extraction work. I could not run the daemon against a real database state in this environment, so the exact runtime reproduction is inferred from the diff rather than executed.


- packages/daemon-rs/crates/signet-daemon/src/main.rs:2412 - The new startup dead-letter path can corrupt per-memory extraction status. After dead-lettering pending jobs, this unconditionally sets `memories.extraction_status = 'failed'` whenever there are no `leased` jobs left. That ignores the case where the same memory already has a completed/successful extraction record and only a duplicate or stale pending job remains. In that state, booting with extraction blocked would incorrectly downgrade an already-processed memory to `failed`. The write route helper below has the same pattern. This needs an additional guard against existing successful/completed extraction work before flipping the memory-level status.
- packages/cli/src/lib/runtime.test.ts:1073 - This test rewrite removes the entire `readManagedDaemonPid` regression suite and replaces it with `/api/status` parsing coverage. That means the CLI no longer has tests for the reused-PID safety checks (matching the daemon command path, rejecting unrelated processes, and cleaning stale pid files), even though that logic is still security/reliability-sensitive for restart/status flows. Given the repo's incident guardrail requirements, this coverage should be kept and the new status test added alongside it rather than replacing it.

aaf2tbz:
Pushed `5d99e0b1` to address the latest requested changes.

Fixes included:
- Preserved completed extraction state during startup dead-lettering (daemon-rs): memories with existing completed extraction work are no longer downgraded to `failed` when stale pending jobs are dead-lettered.
  - Updated both startup dead-letter path (`main.rs`) and blocked-remember dead-letter helper (`routes/write.rs`).
  - Added regression tests for completed-memory preservation in both modules.
- Restored CLI `readManagedDaemonPid` regression coverage and kept the new `/api/status` parsing coverage (runtime tests now include both suites).

Validation run:
- `cargo test -p signet-daemon --manifest-path packages/daemon-rs/Cargo.toml dead_letter_blocked_extraction_ -- --nocapture`
- `cargo test -p signet-daemon --manifest-path packages/daemon-rs/Cargo.toml startup_dead_letter_ -- --nocapture`
- `bun test packages/cli/src/lib/runtime.test.ts`
- `git diff --check`

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `5d99e0b1`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -94,6 +94,104 @@ fn parse_remember_tags(value: Option<Value>) -> Result<Vec<String>, &'static str
...
+         WHERE memory_id = ?1
+           AND job_type IN ('extract', 'extraction')
+           AND status = 'leased'",
+        rusqlite::params![memory_id],
+        |row| row.get(0),
+    )?;
+    let completed_jobs_count: i64 = conn.query_row(
+        "SELECT COUNT(*) FROM memory_jobs
+         WHERE memory_id = ?1
+           AND job_type IN ('extract', 'extraction')
```
PR-Reviewer-Ant, line 128: The new completion guard only recognizes `extraction_status` values `complete` and `completed`. This PR's own API docs still show `done` as a persisted status value. If any existing rows or write paths still use `done`, blocked remember writes will incorrectly flip already-extracted memories back to `failed` when cleaning up jobs. The same predicate is duplicated in the startup dead-letter helper in `src/main.rs`, so this needs a shared/source-of-truth fix rather than patching one site.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5d99e0b1`

[Automated Review] Found a blocking extraction-state regression in the new dead-letter logic.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The new Rust dead-letter helpers explicitly treat only `extraction_status IN ('complete', 'completed')` as already-finished, while the repository's API docs in this PR still show `extraction_status: "done"`. If `done` is still a live stored value anywhere in the daemon/core path, these new helpers will incorrectly overwrite already-extracted memories to `failed` when a stale pending job exists. I can prove the mismatch from this diff, but I could not inspect the full runtime enum/source of truth because the workspace only contains the review artifact rather than a full checkout.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): persist extraction fallback status with provider parity (replaces #367)` (commit `92198631`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `92198631`

[Automated Review] I did not find any new blocking issues in the changed code paths; this looks ready for human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The PR description’s main claims are reflected in the diff: persisted extraction runtime status is exposed in `/api/status`, CLI status surfaces degraded/blocked states, fallbackProvider is validated in both TS types and Rust config parsing, and blocked-startup dead-letter behavior is covered by Rust tests around both startup and `/api/memory/remember`. I did not run the daemon end-to-end here, so confidence is limited to code-path inspection and included tests.

PR #378, opened 2026-03-27
fix(cli): protect OpenClaw-linked workspaces from unprotected data loss

NicholaiVogel:
## Summary

This incident hardening patch adds a setup soft gate and diagnostics for a critical OpenClaw workspace-loss risk:

- OpenClaw can uninstall-delete its configured workspace
- if that workspace is Signet's `.agents` path and no backup exists, identity/memory/history can be lost

### What changed

- Added setup protection flow for OpenClaw-linked workspaces:
  - accepts safety posture when `origin` remote exists
  - otherwise requires one action:
    - set `origin` now, or
    - create local snapshot backup, or
    - explicitly bypass (soft gate)
- Added non-interactive flags:
  - `--create-local-backup`
  - `--allow-unprotected-workspace`
- Hardened OpenClaw workspace patch prompt text with explicit uninstall risk warning
- Added `signet status` / `signet doctor` warnings for OpenClaw-linked unprotected workspaces
- Added post-setup protection summary output
- Added incident docs:
  - research + planning + approved spec
  - dependency/index updates
  - community announcement draft

## Files of interest

- `packages/cli/src/lib/workspace-protection.ts`
- `packages/cli/src/features/setup-protection.ts`
- `packages/cli/src/features/setup.ts`
- `packages/cli/src/features/setup-fresh.ts`
- `packages/cli/src/features/setup-migrate.ts`
- `packages/cli/src/features/health.ts`

## Validation

Ran:

- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bun run build`

Notes:

- `bun run lint` currently fails due to repo-wide pre-existing biome drift (not introduced by this patch)
- `bun scripts/spec-deps-check.ts` currently fails due to pre-existing `references/Engram` path mismatch in another spec

## Why this is a `fix:`

This is incident hardening and data-loss prevention for existing setup/diagnostics behavior, not a new user-facing feature surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 All notable changes to Signet are documented here.
 
+## [0.81.2] - 2026-03-28
+
+### Features
+
+- **setup**: deployment-aware setup defaults + native embedding docs alignment (#363)
+
+### Bug Fixes
+
+- **daemon**: batch watcher identity sync to keep health responsive (#375)
+
+### Docs
+
+- add Groundswell implementation specs + PRD + SSM synthesis (#381)
+- **specs**: add Groundswell PRD and gap analyses for community knowledge graphs
+
+
 ## [0.81.1] - 2026-03-28
 
 ### Bug Fixes
--- a/README.md
+++ b/README.md
@@ -409,7 +409,17 @@ cd packages/daemon && bun run dev        # Daemon dev (watch mode)
 cd packages/cli/dashboard && bun run dev # Dashboard dev
 ```
 
-Requirements: Node.js 18+, Bun, Ollama (recommended) or OpenAI API key. macOS or Linux.
+Requirements:
+
+- Node.js 18+ or Bun
+- macOS or Linux
+- Optional for harness integrations: Claude Code, Codex, OpenCode, or OpenClaw
+
+Embeddings (choose one):
+
+- **Built-in** (recommended) — no extra setup, runs locally via ONNX (`nomic-embed-text-v1.5`)
+- **Ollama** — alternative local option, requires `nomic-embed-text` model
+- **OpenAI** — cloud option, requires `OPENAI_API_KEY`
 
 ## Contributing
 
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -110,8 +110,8 @@ signet setup --path /custom/path
 signet setup --non-interactive \
   --name "My Agent" \
   --harness claude-code \
-  --embedding-provider ollama \
-  --extraction-provider claude-code
+  --deployment-type vps \
+  --embedding-provider native
 ```
 
 Options:
@@ -122,10 +122,11 @@ Options:
 | `--non-interactive` | Run setup without prompts |
 | `--name <name>` | Agent name in non-interactive mode |
 | `--description <description>` | Agent description in non-interactive mode |
+| `--deployment-type <type>` | Deployment context (`local`, `vps`, `server`) used for interactive guidance and non-interactive inferred defaults |
 | `--harness <harness>` | Repeatable/comma-separated harness list (`claude-code`, `opencode`, `openclaw`, `codex`) |
-| `--embedding-provider <provider>` | Non-interactive embedding provider (`ollama`, `openai`, `native`, `none`) — required in non-interactive setup |
+| `--embedding-provider <provider>` | Non-interactive embedding provider (`ollama`, `openai`, `native`, `none`) |
 | `--embedding-model <model>` | Non-interactive embedding model |
-| `--extraction-provider <provider>` | Non-interactive extraction provider (`claude-code`, `codex`, `ollama`, `opencode`, `openrouter`, `none`) — required in non-interactive setup |
+| `--extraction-provider <provider>` | Non-interactive extraction provider (`claude-code`, `codex`, `ollama`, `opencode`, `openrouter`, `none`) |
 | `--extraction-model <model>` | Non-interactive extraction model |
 | `--search-balance <alpha>` | Non-interactive search alpha (`0-1`) |
 | `--openclaw-runtime-path <mode>` | Non-interactive OpenClaw mode (`plugin`, `legacy`) |
@@ -138,8 +139,15 @@ Options:
 Non-interactive behavior:
 
 - setup method: create new identity (no GitHub import)
-- embedding provider must be explicitly provided via `--embedding-provider`
-- extraction provider must be explicitly provided via `--extraction-provider`
+- provider flags are optional; setup infers defaults from `--deployment-type`
+  when omitted
+- with `--deployment-type vps`, setup prefers non-local extraction defaults
+  from selected harnesses when those tools are available locally, then other
+  detected tooling (`claude-code`, `codex`, `opencode`), and falls back to
+  `none` when needed
+- for existing-identity migration, previously configured extraction providers
+  are preserved unless `--extraction-provider` is explicitly passed
+- explicit provider flags override inferred defaults
 - git: enabled unless `--skip-git` is passed
 - when OpenClaw points at this workspace and no `origin` remote exists, setup
   requires either backup creation (`--create-local-backup`) or explicit bypass
@@ -151,6 +159,8 @@ Extraction safety note:
 
 - intended usage is `claude-code` on Haiku, `codex` on GPT Mini with a
   Pro/Max subscription, or local `ollama` with at least `qwen3:4b`
+- with `--deployment-type vps`, setup avoids defaulting to local `ollama`
+  extraction and prefers non-local providers
 - set `--extraction-provider none` on a VPS if you do not want
   background extraction
 - remote API extraction can create extreme usage fees fast
@@ -167,12 +177,14 @@ Wizard steps:
    is detected; workspace is patched only if you opt in, and setup warns
    that uninstalling OpenClaw can delete this workspace unless backups exist
 4. **Description** - Short agent description
-5. **Embedding Provider**:
+5. **Deployment Context** - Where Signet is running (`local`, `vps`, `server`)
+   to show environment-aware guidance before extraction provider selection
+6. **Embedding Provider**:
    - Built-in (recommended, no setup required)
    - Ollama (local)
    - OpenAI API
    - Skip embeddings
-6. **Embedding Model** - Based on provider:
+7. **Embedding Model** - Based on provider:
    - Built-in: `nomic-embed-text-v1.5`
    - Ollama: `nomic-embed-text`, `all-minilm`, `mxbai-embed-large`
    - OpenAI: text-embedding-3-small, text-embedding-3-large
@@ -180,15 +192,15 @@ Wizard steps:
      service health, and model presence; if checks fail, setup offers
      retry, switch to built-in embeddings, switch to OpenAI, or
      continue without embeddings
-7. **Search Balance** - Semantic vs keyword weighting
-8. **Advanced Settings** (optional):
+8. **Search Balance** - Semantic vs keyword weighting
+9. **Advanced Settings** (optional):
    - `top_k` - Search candidates per source
    - `min_score` - Minimum search score threshold
    - `session_budget` - Context character limit
    - `decay_rate` - Memory importance decay
-9. **Import** - Optionally import from another platform
-10. **Git** - Initialize version control
-11. **Launch Dashboard** - Open web UI
+10. **Import** - Optionally import from another platform
+11. **Git** - Initialize version control
+12. **Launch Dashboard** - Open web UI
 
 What gets created:
 
--- a/docs/HARNESSES.md
+++ b/docs/HARNESSES.md
@@ -115,6 +115,11 @@ Degraded mode rules:
   document the missing surface explicitly instead of implying full parity.
 - Compaction/session-summary forced-refresh guarantees are currently full in
   TS daemon and degraded in daemon-rs shadow path.
+- Watcher-driven harness identity sync (batched `AGENTS.md` / `USER.md` /
+  `MEMORY.md` propagation, queued reruns during active sync, and generated
+  `SIGNET-ARCHITECTURE.md` refresh) is currently full in the TS daemon and
+  degraded in the daemon-rs shadow path, which does not yet own that watcher
+  pipeline.
 - Prompt-time anti-bleed scoping remains mandatory across all paths.
 
 ---
--- a/docs/QUICKSTART.md
+++ b/docs/QUICKSTART.md
@@ -97,7 +97,10 @@ Prerequisites
 ---
 
 - Node.js 18+ (or Bun 1.0+)
-- One of: Ollama (for local embeddings) or an OpenAI API key
+- Embeddings (choose one):
+  - Built-in (recommended, no extra setup)
+  - Ollama (local)
+  - OpenAI API key
 - macOS or Linux (Windows support planned)
 
 ---
@@ -127,12 +130,13 @@ For agent-driven onboarding, use non-interactive mode:
 signet setup --non-interactive \
   --name "My Agent" \
   --harness claude-code \
-  --embedding-provider ollama \
-  --extraction-provider claude-code
+  --deployment-type vps \
+  --embedding-provider native
 ```
 
-In non-interactive mode, agents should ask the user to choose both
-providers first, then pass those choices explicitly.
+`--deployment-type` supports `local`, `vps`, or `server` and adjusts inferred
+defaults when provider flags are omitted. Explicit provider flags always
+override inferred defaults.
 
 If OpenClaw is configured to use the same workspace path, setup now enforces
 backup posture before finishing. In automation, either configure a git
@@ -145,8 +149,10 @@ Extraction safety note:
 
 - intended usage is Claude Code on Haiku, Codex CLI on GPT Mini with a
   Pro/Max subscription, or local Ollama with at least `qwen3:4b`
-- on a VPS, set extraction to `none` unless you explicitly want
-  background LLM calls
+- with `--deployment-type vps`, setup prefers non-local extraction defaults
+  from selected harnesses when those tools are available locally, then other
+  detected tooling, and avoids defaulting to local Ollama extraction
+- on a VPS, set extraction to `none` if you do not want background LLM calls
 - remote API extraction can rack up extreme fees fast
 
 ---
@@ -171,7 +177,12 @@ for each:
 - OpenClaw — adapter-openclaw hooks
 - Codex — wrapper install + session hooks
 
-**3. Embedding provider**
+**3. Deployment context**
+
+Choose where Signet is running (`local`, `vps`, `server`). Setup uses
+this to show guidance before extraction provider selection.
+
+**4. Embedding provider**
 
 Embeddings power semantic (meaning-based) memory search. Choose:
 
@@ -182,7 +193,7 @@ Embeddings power semantic (meaning-based) memory search. Choose:
 - **OpenAI** — uses the OpenAI embeddings API. Requires `OPENAI_API_KEY`.
 - **Skip** — memory still works via keyword search, just no semantic search.
 
-**4. Embedding model**
+**5. Embedding model**
 
 For Ollama, `nomic-embed-text` is a good default. Setup can pull it for
 you (with confirmation), or you can do it manually:
@@ -191,12 +202,12 @@ you (with confirmation), or you can do it manually:
 ollama pull nomic-embed-text
 ```
 
-**5. Search balance**
+**6. Search balance**
 
 The `alpha` setting controls how much weight goes to semantic vs. keyword
 search. 0.7 (70% semantic, 30% keyword) works well for most people.
 
-**6. Git & auto-commit**
+**7. Git & auto-commit**
 
 The wizard can initialize a git repo in `$SIGNET_WORKSPACE/` so every change to
 your agent files is automatically versioned.
--- a/docs/research/technical/SSM-DEEP-DIVE-SYNTHESIS.md
+++ b/docs/research/technical/SSM-DEEP-DIVE-SYNTHESIS.md
@@ -1,0 +1,621 @@
+# Signet SSM Deep Dive: Complete Architecture Synthesis
+
+**Prepared for:** Nicholai (creator) and Jake (co-founder)  
+**Date:** March 27, 2026  
+**Sources:** 8 research documents, 5 planning specs, 1 approved epic — all read in full  
+
+---
+
+## Table of Contents
+
+1. [Architecture Summary](#1-architecture-summary)
+2. [Key Research Findings](#2-key-research-findings)
+3. [Synthetic Data Strategy](#3-synthetic-data-strategy)
+4. [Implementation Roadmap](#4-implementation-roadmap)
+5. [Novel Insights](#5-novel-insights)
+6. [Open Questions](#6-open-questions)
+7. [Connection to CRM + Gmail Scaling](#7-connection-to-crm--gmail-scaling)
+
+---
+
+## 1. Architecture Summary
+
+### The Thesis
+
+Signet's VISION.md describes the endgame: *"A neural network unique to each user, trained on their own interaction patterns, that gets sharper the longer you use it. No shared personal weights. Your weights never leave your machine."*
+
+The current pipeline approximates this with ~19,800 LOC of hand-tuned heuristics across 52+ files: a fixed `importance * 0.95^ageDays` decay curve, 34 hardcoded antonym pairs for contradiction detection, static traversal weights, a broken cross-attention scorer sidecar, and stateless per-turn search. Each stage was built independently because no single model could hold state across time.
+
+**The SSM is that model.** The proposal isn't "sprinkle SSMs in a few places." It's: **SSMs become the learned temporal backbone that every pipeline stage consults.** The graph stays. SQLite stays. FTS5 stays. But temporal reasoning — what's important now, what will matter next, what should decay, what contradicts what — moves from hardcoded rules to a model that learns from each user's interaction patterns.
+
+### What Replaces What
+
+The architecture defines three distinct SSM components, each targeting a different kind of reasoning:
+
+#### Component 1: Signet Neural Backbone (SNB) — Temporal Interaction Model
+
+**Replaces:** The cross-attention scorer sidecar (~1.1M params, 3 critical bugs), the `0.95^ageDays` decay heuristic, the significance gate's fixed thresholds, the stateless per-turn search paradigm.
+
+**Architecture:** Mamba-3 selective SSM (MIMO, trapezoidal discretization, complex-valued state). 5-10M parameters, 4-8 layers, d_model=128-256, N=32-64. Heterogeneous quantization (A matrix fp16, projections int8). ~10-15MB binary in Rust. <1ms per event on CPU.
+
+**What it does:** Processes the sequence of interaction events across a session (and across sessions via persistent hidden state). At each turn, produces:
+- Memory relevance scores (which memories to inject)
+- Aspect bias weights (which graph traversal paths to prioritize)
+- Predicted next entities (proactive embedding prefetch)
+- Retention adjustment (per-memory decay rate modifiers)
+- Significance score (is this session worth extracting?)
+- Contradiction probability (does a new fact conflict with stored ones?)
+
+**Why Mamba-3 specifically:** The March 2026 ICLR paper introduced three features that directly serve Signet: (1) complex-valued state updates that naturally model cyclical patterns like daily/weekly usage rhythms, (2) MIMO formulation that packs more computation into the memory-bound decode step without increasing state or latency, and (3) trapezoidal discretization that achieves Mamba-2 quality with half the state size — critical for keeping the sidecar small.
+
+#### Component 2: Path SSM — Graph Traversal Scorer
+
+**Replaces:** The static `strength * confidence * association_weight` path scoring in graph traversal (DP-6).
+
+**Architecture:** S4D (non-selective structured SSM). ~50-60K parameters. 1-3 layers, d_model=32-64, state_dim=16-32. Processes sequences of edge features along traversal paths.
+
+**Why S4 and not Mamba:** GraphSSM (NeurIPS 2024) ran a controlled experiment: S4 scored 49.21% vs Mamba's 43.11% on Reddit temporal graphs. The paper explains why: Mamba's selective mechanism (input-dependent gating) is powerful for language where you want to ignore irrelevant tokens, but counterproductive for graph topology where every node's structural position matters. S4's fixed HiPPO-initialized state transitions act as stable structural filters that preserve graph information.
+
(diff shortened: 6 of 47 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): protect OpenClaw-linked workspaces from unprotected data loss` (commit `c04240e3`) now and I'll follow up shortly with feedback.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): protect OpenClaw-linked workspaces from unprotected data loss` (commit `c04240e3`) now and I'll follow up shortly with feedback.

packages/cli/src/features/health.ts
```diff
@@ -126,7 +135,10 @@ export async function getStatusReport(basePath: string, deps: StatusDeps): Promi
 			conversationCount: null,
 		},
 		daemon,
+		git,
```
PR-Reviewer-Ant: This marks every linked workspace without `origin` as unprotected, but the PR/spec explicitly says a local snapshot backup is also an acceptable protection state. After a user runs setup with `--create-local-backup`, `signet status` / `signet doctor` will still warn forever because no snapshot state is persisted or consulted here. That is a direct behavior mismatch with the claimed implementation.

packages/cli/src/features/setup-fresh.ts
```diff
@@ -221,6 +222,13 @@ export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Pro
 			}
```
PR-Reviewer-Ant: The protection gate runs too late. By the time this executes, setup has already initialized the workspace, possibly patched OpenClaw to point at it, started the daemon, and printed success output. In the non-interactive failure path, the command now aborts only after those side effects have already happened, which does not satisfy the spec/PR claim that setup requires explicit protection before finishing. The same ordering problem exists in `packages/cli/src/features/setup-migrate.ts`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c04240e3`

[Automated Review] Found two correctness issues that undermine the new workspace-protection contract.
The PR adds the requested setup/diagnostic surfaces, but the implementation does not fully match the stated protection model.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The spec and PR description both define two acceptable protection states for linked workspaces: git origin or local snapshot. In the diff, `health.ts` only checks `git.origin === null`, so a workspace protected by `--create-local-backup` will still be reported as unprotected. Separately, both setup flows call `enforceSetupProtection()` only after the workspace is created/configured, optional OpenClaw patching has run, the daemon has started, and success output has already been printed, so the non-interactive guard does not actually prevent setup side effects before failure.

NicholaiVogel:
Addressed the two correctness issues called out in review on commit `aec626ef`.

### 1) Snapshot-backed protection now counts as protected in diagnostics
- Added persisted snapshot marker support in `workspace-protection.ts`:
  - `saveSnapshotProtection()`
  - `getSnapshotProtection()`
- `enforceSetupProtection()` now records marker on snapshot creation.
- `health.ts` now treats workspace as unprotected only when:
  - OpenClaw is linked **and**
  - `origin` is missing **and**
  - no valid snapshot marker exists.
- Added regression coverage in `health.test.ts` for marker-backed protection.

### 2) Protection gate now runs before setup side effects
- Moved `enforceSetupProtection()` to the start of both setup flows:
  - `runFreshSetup()`
  - `runExistingSetupWizard()`
- This ensures non-interactive failure happens before connector configuration, daemon start, and success output.

### Extra hardening
- Interactive remote setup path now handles remote-set failures without crashing the setup loop.
- `setOriginRemote()` now initializes a git repo when missing so "set origin now" works in early setup states.

### Validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): protect OpenClaw-linked workspaces from unprotected data loss` (commit `aec626ef`) now and I'll follow up shortly with feedback.

packages/cli/src/features/setup-fresh.ts
```diff
@@ -18,6 +19,12 @@ export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Pro
 	try {
 		const templatesDir = deps.getTemplatesDir();
 		mkdirSync(cfg.basePath, { recursive: true });
```
PR-Reviewer-Ant, line 31: The new protection gate runs before setup creates `AGENTS.md`, `agent.yaml`, `MEMORY.md`, or `memory/memories.db`. In the fresh-install case, `--create-local-backup` therefore snapshots an almost-empty directory and then marks it as satisfying the protection contract. `signet status` / `doctor` will later report that workspace as protected even though the snapshot does not contain the identity or memory state this PR claims to preserve.

packages/cli/src/lib/workspace-protection.ts
```diff
@@ -0,0 +1,195 @@
...
+export function defaultBackupRoot(): string {
+	return join(homedir(), ".signet", "backups");
+}
+
+function snapshotStatePath(basePath: string): string {
+	return join(resolve(basePath), ".signet-workspace-protection.json");
+}
+
+export function createWorkspaceSnapshot(basePath: string, backupRoot = defaultBackupRoot()): SnapshotResult {
+	const root = resolve(backupRoot);
```
PR-Reviewer-Ant: This snapshot implementation is a raw recursive `cpSync()` of the workspace, including the live SQLite database and git metadata, with no daemon stop, no SQLite backup API, and no consistency check before `saveSnapshotProtection()` blesses it as a valid recovery target. For an existing active workspace, that can yield a torn/corrupted backup while status/doctor still suppress the warning. Since this PR is specifically about preventing unrecoverable data loss, treating a non-atomic filesystem copy as sufficient protection is unsafe.

NicholaiVogel:
good catch on both fronts.

updates pushed:

1. **merge conflict resolved**
   - rebased branch onto `origin/main` (now includes latest `INDEX.md` updates)
   - kept both provenance rows in `docs/specs/INDEX.md` (`docker-self-hosting-stack` and `openclaw-workspace-protection*`)

2. **additional reviewer concerns addressed**
   - fresh setup no longer snapshots an empty workspace:
     - protection gate in `runFreshSetup()` now runs after core workspace/db creation and before later integration side effects
   - local snapshots now use a SQLite-safe path for `memory/memories.db`:
     - recursive copy still used for workspace files
     - db file replaced via `VACUUM INTO` from source db into snapshot target
   - snapshot marker support remains, and status/doctor recognize valid snapshot-backed protection

3. **validation rerun**
   - `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
   - `bun run typecheck`

latest head: `2a0e1e2a`

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): protect OpenClaw-linked workspaces from unprotected data loss` (commit `2a0e1e2a`) now and I'll follow up shortly with feedback.

packages/cli/src/features/setup-fresh.ts
```diff
@@ -143,6 +144,13 @@ export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Pro
 			db.close();
```
PR-Reviewer-Ant, line 154: Blocking: the new protection gate runs before the later OpenClaw workspace patch. In a fresh setup where the user opts into `configureOpenClawWorkspace`, `hasOpenClawWorkspaceLink(cfg.basePath)` is still false here, so setup completes without requiring origin/snapshot/bypass and only then patches OpenClaw to the Signet workspace. That misses the exact data-loss case described in the PR and spec. This flow needs the gate after the patch decision is applied (or needs to account for pending patch intent), plus a regression test for the fresh-setup patch path.

NicholaiVogel:
thanks, i pulled latest reviewer feedback and fixed both items in `32488e40`.

### fixed
- **fresh setup pre-link gap**
  - `enforceSetupProtection()` now supports `assumeOpenClawLinked`
  - `runFreshSetup()` passes `assumeOpenClawLinked: cfg.configureOpenClawWs`
  - this forces the gate when setup is about to patch OpenClaw workspace, even before config patch is written

- **snapshot safety-contract gap**
  - snapshot protection is now accepted only when snapshot path exists **and is outside** the workspace tree
  - added boundary guard in `createWorkspaceSnapshot()` to reject backup roots inside workspace
  - added tests for both:
    - marker-inside-workspace is ignored
    - in-workspace snapshot roots are rejected

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`

packages/cli/src/features/setup-fresh.ts
```diff
@@ -143,6 +144,14 @@ export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Pro
...
```
PR-Reviewer-Ant, line 154: The protection gate runs after the fresh setup has already created the workspace, written identity/config files, and initialized the database. In the new non-interactive failure case (`linked + no origin + no flags`), setup now exits only after mutating disk. On the next run, the workspace may be treated as an existing install instead of a clean fresh setup, which is a behavioral regression from a failed guardrail. The gate needs to run before these side effects, or the failure path needs explicit rollback.

packages/cli/src/lib/workspace-protection.ts
```diff
@@ -0,0 +1,234 @@
...
+		}
+		return {
+			source: resolve(source),
+			snapshot: resolve(snapshot),
+			createdAt: createdAt.trim(),
+		};
+	} catch {
+		return null;
+	}
+}
```
PR-Reviewer-Ant, line 227: Diagnostics treat any `.signet-workspace-protection.json` whose `snapshot` field points to an existing directory outside the workspace as a valid protection state. That means a stale or manually crafted marker can suppress the new `status`/`doctor` warnings without proving that a real workspace snapshot exists. This diverges from the PR description/spec language about an actual local snapshot backup. At minimum, validate expected snapshot contents (for example `AGENTS.md` and/or `memory/memories.db`) before clearing the warning.

NicholaiVogel:
you're right, i should have pulled this proactively. i just addressed the latest blocking review in `bb80a1f5`.

### fixes for newest feedback

1. **fresh-setup failure path now fails before mutations (non-interactive blocking case)**
   - added an early preflight guard in `runFreshSetup()` for:
     - `nonInteractive && !allowUnprotectedWorkspace && !createLocalBackup`
   - this runs `enforceSetupProtection()` before file/db writes, so the guardrail no longer leaves partial workspace state when it throws.

2. **snapshot marker integrity tightened**
   - `getSnapshotProtection()` now validates snapshot content, not just path existence:
     - requires `AGENTS.md`
     - requires `memory/memories.db`
   - existing outside-workspace boundary check remains in place.

### regression coverage
- added test: snapshot marker with incomplete external content is ignored.
- existing tests still cover inside-workspace marker rejection and pre-link enforcement.

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`

packages/cli/src/lib/workspace-protection.ts
```diff
@@ -0,0 +1,244 @@
...
+	}
+	return trimmed.replace(/[^a-z0-9._-]+/g, "-");
+}
+
+function isWithin(root: string, path: string): boolean {
+	const outer = resolve(root);
+	const inner = resolve(path);
+	if (outer === inner) {
+		return true;
+	}
```
PR-Reviewer-Ant, line 49: The new protection gate accepts a snapshot as valid if it only contains `AGENTS.md` and `memory/memories.db`. That is too weak for the risk this PR is trying to mitigate: a directory missing `agent.yaml`, other identity files, or `.git` history will still suppress the setup block and status/doctor warnings. In practice, a stale or manually crafted directory with those two files is enough to bypass the guard, even though it would not restore the full Signet workspace described in the PR. The validation should require the actual restore contract for this workspace snapshot, not just two sentinel files.

NicholaiVogel:
pulled latest review and patched it in `69d9dd18`.

### change
snapshot protection validation now enforces a fuller restore contract before status/doctor suppress warnings.

`getSnapshotProtection()` now requires snapshot to contain:
- `AGENTS.md`
- `agent.yaml`
- `SOUL.md`
- `IDENTITY.md`
- `USER.md`
- `MEMORY.md`
- `memory/memories.db`
- and when source workspace is git-backed, `.git` must also exist in snapshot.

outside-workspace boundary validation remains in place.

### tests
- updated protected-snapshot fixture to include full expected workspace files
- updated incomplete-snapshot case to prove AGENTS+DB-only is rejected
- re-ran:
  - `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
  - `bun run typecheck`

packages/cli/src/lib/workspace-protection.ts
```diff
@@ -0,0 +1,262 @@
...
+		}
+		const createdAt = "createdAt" in raw ? raw.createdAt : null;
+		if (typeof createdAt !== "string" || createdAt.trim().length === 0) {
+			return null;
+		}
+		return {
+			source: resolve(source),
+			snapshot: resolve(snapshot),
+			createdAt: createdAt.trim(),
+		};
```
PR-Reviewer-Ant, line 223: The snapshot marker can suppress the OpenClaw risk warning indefinitely, even when the backup is stale. `saveSnapshotProtection()` stores `createdAt`, but `getSnapshotProtection()` never checks freshness or whether the snapshot still reflects the current workspace state; it only verifies that required files exist. That diverges from the spec/PR description's 'freshly created local snapshot' posture and can cause `signet status` / `signet doctor` to report a linked workspace as protected after substantial newer memory/history has accumulated with no remote backup.

NicholaiVogel:
pulled latest feedback and patched the stale-snapshot gap in `20e3db4b`.

### fix
`getSnapshotProtection()` now enforces snapshot freshness using `createdAt`:
- added `SNAPSHOT_MAX_AGE_MS` (7 days)
- stale markers now return `null`
- status/doctor now resume warning once a snapshot ages out, unless `origin` exists

### regression coverage
- added `health.test.ts`: **ignores snapshot markers when backup is stale**
  - validates linked workspace is reported unprotected
  - validates `git.snapshot` is `null`

### docs
- updated `docs/CLI.md` and `docs/QUICKSTART.md` to document the 7-day freshness window for snapshot-backed protection.

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/lib/workspace-protection.ts packages/cli/src/features/health.test.ts`

packages/cli/src/features/setup-migrate.ts
```diff
@@ -63,6 +66,12 @@ export async function runExistingSetupWizard(
 		if (!existsSync(join(basePath, "memory", "scripts"))) {
```
PR-Reviewer-Ant, line 78: Blocking: the migration path enforces snapshot protection before the setup work has finished creating/populating the full Signet workspace. Later, `getSnapshotProtection()` only accepts snapshots containing `AGENTS.md`, `agent.yaml`, `SOUL.md`, `IDENTITY.md`, `USER.md`, `MEMORY.md`, and `memory/memories.db`. On a partially initialized workspace being migrated, the user can choose 'create local snapshot', the wizard will report success, and then `signet status`/`doctor` will immediately treat that same snapshot as invalid/unprotected. That breaks the soft-gate contract the PR description claims.

packages/cli/src/lib/workspace-protection.ts
```diff
@@ -0,0 +1,279 @@
...
+	const origin = readOutput(remote.stdout);
+	if (origin.length === 0) {
+		return { isRepo: true, origin: null };
+	}
+	return { isRepo: true, origin };
+}
+
+export function hasOpenClawWorkspaceLink(basePath: string): boolean {
+	const target = resolve(basePath);
+	const connector = new OpenClawConnector();
```
PR-Reviewer-Ant, line 133: Warning: the snapshot protection marker is written into the workspace root as `.signet-workspace-protection.json`, and it contains an absolute local backup path. Signet workspaces are git-backed and auto-synced, so this will leak machine-local filesystem paths into commits/remotes and replicate stale host-specific state across clones. This metadata should live in a non-synced internal location (for example under `.daemon/`) or be explicitly gitignored.

NicholaiVogel:
pulled the new review and addressed both blockers in `7fe0445b`.

### 1) migration snapshot timing mismatch fixed
- moved `enforceSetupProtection()` in `runExistingSetupWizard()` to run **after** migration setup writes the workspace scaffold (`agent.yaml`, identity docs when missing, and `memory/memories.db`).
- added missing-doc scaffold in migration flow (creates `AGENTS.md`, `MEMORY.md`, `SOUL.md`, `IDENTITY.md`, `USER.md` from templates when absent) before protection enforcement.

this keeps `--create-local-backup` consistent with the snapshot validation contract used by status/doctor.

### 2) marker path leak fixed
- snapshot marker now writes to:
  - `.daemon/workspace-protection.json`
  instead of workspace-root `.signet-workspace-protection.json`.
- `saveSnapshotProtection()` now:
  - creates `.daemon/` as needed
  - removes legacy root marker if present
- `readSnapshotState()` supports backward compatibility by reading new marker first, then legacy marker.

### regression coverage
- added test: `stores snapshot protection marker under .daemon and clears legacy marker`

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/lib/workspace-protection.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`

packages/cli/src/features/setup-fresh.ts
```diff
@@ -143,6 +154,14 @@ export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Pro
...
```
PR-Reviewer-Ant, line 154: Blocking: the snapshot guard runs before the rest of setup finishes. If the user chooses `--create-local-backup` or the interactive snapshot path here, the backup is taken and the protection marker is written before later setup mutations (connector installs, later file writes, final setup state) complete. `status`/`doctor` will then treat the workspace as protected for 7 days, but the snapshot may not actually contain the final configured workspace this PR claims to protect.

packages/cli/src/features/setup-migrate.ts
```diff
@@ -198,6 +235,13 @@ export async function runExistingSetupWizard(
 		runMigrations(db);
```
PR-Reviewer-Ant, line 247: Blocking: the migration/setup flow records snapshot protection before the remaining import/setup steps run. That means a local snapshot can be marked as valid while missing later migrated content or post-migration writes, creating a false sense of recoverability and suppressing the new warnings. The snapshot should be taken after the workspace reaches its final post-setup state, or the marker should not be considered valid until then.

NicholaiVogel:
pulled the newest blocker and patched it in `83b7a324`.

### fix: snapshot now reflects post-setup state
reviewer was right, snapshot marker could be written before remaining setup mutations finished.

changes:
- added `refreshSnapshotProtection(basePath, result)` in `setup-protection.ts`
  - when protection state is `snapshot`, it takes a fresh snapshot from current workspace state and rewrites marker.
- `runFreshSetup()` now:
  - completes setup writes/harness config/openclaw workspace config
  - commits git setup changes (if enabled)
  - then refreshes snapshot protection before daemon start/success output.
- `runExistingSetupWizard()` now does the same ordering:
  - complete migration/import/harness/git steps
  - refresh snapshot protection
  - then start daemon/success output.

this keeps `status`/`doctor` protection signals aligned with the final configured workspace, not an early intermediate state.

### additional hardening included in this commit
- snapshot directory stamp now includes milliseconds (`YYYYMMDDTHHMMSSmmmZ`) to avoid same-second collisions when refreshing snapshots quickly.

### regression coverage
- added test: `refreshes snapshot protection after later setup mutations`
  - verifies refreshed snapshot path changes
  - verifies refreshed snapshot includes post-snapshot DB write
  - verifies marker resolves to refreshed path

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/features/setup-protection.ts packages/cli/src/features/setup-fresh.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/workspace-protection.ts`

packages/cli/src/features/setup-migrate.ts
PR-Reviewer-Ant, line 247: The new protection gate runs too late for the existing-workspace path. By the time `enforceSetupProtection()` is called, this flow has already written `agent.yaml`, created identity files, and initialized/migrated the DB. In non-interactive mode, an unprotected OpenClaw-linked workspace will now fail only after those mutations, leaving partial setup side effects on disk. That diverges from the PR's 'soft gate' goal for automation. The protection check needs to happen before mutating the workspace, or the earlier writes need rollback semantics.

packages/cli/src/lib/workspace-protection.ts
```diff
@@ -0,0 +1,288 @@
...
+	});
+	if (remote.status !== 0) {
+		return { isRepo: true, origin: null };
+	}
+	const origin = readOutput(remote.stdout);
+	if (origin.length === 0) {
+		return { isRepo: true, origin: null };
+	}
+	return { isRepo: true, origin };
+}
```
PR-Reviewer-Ant, line 129: The snapshot remediation path is not actually available for every supported workspace layout. `createWorkspaceSnapshot()` defaults to `~/.signet/backups`, then rejects any backup root inside the workspace. If a user keeps Signet in a custom path under `~/.signet` (for example `~/.signet` itself), both the interactive 'Create local snapshot backup now' choice and `--create-local-backup` will always fail. Because there is no CLI/API to choose a different backup root, the documented protection action becomes unusable for those installs.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `83b7a324`

[Automated Review] Found two correctness issues in the new workspace-protection flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The failure modes are directly visible in the changed control flow. In `setup-migrate.ts`, the protection gate is placed after several filesystem and DB mutations, so a non-interactive run can partially modify an existing workspace before throwing. In `workspace-protection.ts`, the snapshot path is hardcoded to `~/.signet/backups` and then explicitly rejected if it is inside the workspace, which makes the advertised snapshot remediation unusable for custom workspaces rooted under `~/.signet`.

NicholaiVogel:
pulled the latest review and fixed both items in `6840d40a`.

### 1) migrate non-interactive guard restored pre-mutation
- added the same early preflight used in fresh setup to `runExistingSetupWizard()` for:
  - `nonInteractive && !allowUnprotectedWorkspace && !createLocalBackup`
- this now fails before workspace/db/file mutations in that strict blocking path.
- the later enforcement call remains for interactive and snapshot/remote/bypass decision handling.

### 2) default snapshot root now has safe fallback when workspace lives under `~/.signet`
- `defaultBackupRoot(basePath?)` now chooses:
  1. `~/.signet/backups` when it is outside workspace
  2. `~/.signet-backups` fallback when needed
  3. `tmpdir()/signet-backups` as final fallback
- `createWorkspaceSnapshot()` now resolves default root from `basePath`, so snapshot remediation remains usable for custom workspace layouts.

### regression coverage
- added test: `falls back to an external backup root when workspace contains ~/.signet/backups`

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/features/setup-protection.ts packages/cli/src/features/setup-fresh.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/workspace-protection.ts`

packages/cli/src/features/setup-fresh.ts
```diff
@@ -9,13 +9,24 @@ import { daemonAccessLines } from "../lib/network.js";
...
 export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Promise<void> {
 	const spinner = ora("Setting up Signet...").start();
 
 	try {
+		if (cfg.nonInteractive && !cfg.allowUnprotectedWorkspace && !cfg.createLocalBackup) {
+			await enforceSetupProtection({
+				basePath: cfg.basePath,
+				nonInteractive: true,
+				allowUnprotectedWorkspace: false,
+				createLocalBackup: false,
```
PR-Reviewer-Ant, line 25: Blocking: this preflight treats `cfg.configureOpenClawWs` as equivalent to an actual OpenClaw workspace link by passing it through `assumeOpenClawLinked`. That over-gates fresh setup when the user passed `--configure-openclaw-workspace` but no OpenClaw config was discovered, because setup will now require backup posture for a link that does not exist. The PR/spec contract is specifically about workspaces that *are* linked, so this should only force the gate when discovery found a patch target, not merely when the flag was requested.

NicholaiVogel:
pulled the latest feedback and fixed the over-gating case in `a899046c`.

### fix
fresh-setup preflight no longer assumes OpenClaw linkage from user intent alone.

changes:
- added `openclawConfigCount` to `FreshSetupConfig`.
- in `setup.ts`, this is sourced from actual discovered OpenClaw config paths.
- in `runFreshSetup()`, `assumeOpenClawLinked` is now:
  - `cfg.configureOpenClawWs && cfg.openclawConfigCount > 0`
  for both early non-interactive preflight and main protection enforcement.

so `--configure-openclaw-workspace` without discovered OpenClaw configs no longer triggers protection gating for a non-existent link.

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/features/setup.ts packages/cli/src/features/setup-types.ts packages/cli/src/features/setup-fresh.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-protection.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/workspace-protection.ts`

packages/cli/src/features/setup-fresh.ts
```diff
@@ -180,6 +199,17 @@ export async function runFreshSetup(cfg: FreshSetupConfig, deps: SetupDeps): Pro
 			}
 		}
 
+		let committed = false;
+		if (cfg.gitEnabled) {
```
PR-Reviewer-Ant, line 208: `gitAddAndCommit()` now runs before `refreshSnapshotProtection()`, but the refresh rewrites the workspace-protection marker under `.daemon/`. That means a successful setup can immediately leave the workspace dirty while still printing `✓ Changes committed to git`. If daemon startup then fails, the command also exits after creating an uncommitted protection-marker change. Either refresh before the explicit commit, or stop claiming the workspace is fully committed at this point.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `a899046c`

[Automated Review by pr-reviewer] Found one correctness issue in the new workspace-protection flow.
The PR now matches the main protection contract described in the spec and PR body, but the new snapshot-refresh ordering leaves the workspace dirty immediately after setup while still printing that changes were committed.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - In both setup flows, `gitAddAndCommit(...)` now runs before `refreshSnapshotProtection(...)`. The refresh path rewrites `.daemon/workspace-protection.json`, so a successful setup can still leave an uncommitted workspace-protection marker change after printing `✓ Changes committed to git`. The ordering is directly visible in `setup-fresh.ts` and `setup-migrate.ts`.

packages/cli/src/features/setup-migrate.ts
```diff
@@ -253,6 +309,17 @@ export async function runExistingSetupWizard(
 			}
```
PR-Reviewer-Ant, line 309: Same issue here as in fresh setup: the explicit git commit happens before `refreshSnapshotProtection()`, but refresh writes a new protection marker afterward. The resulting workspace is not actually fully committed even though setup reports that it is, and a later daemon-start failure leaves that marker change uncommitted.

NicholaiVogel:
pulled the remaining correctness note and fixed ordering in `fe66c2ab`.

### fix
`gitAddAndCommit()` now runs **after** `refreshSnapshotProtection()` in both setup flows:
- `runFreshSetup()`
- `runExistingSetupWizard()`

so the refreshed marker/write state is included in the setup commit, and the "changes committed" message now matches actual workspace state.

### validation
- `bun test packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/features/setup-fresh.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-protection.ts packages/cli/src/lib/workspace-protection.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts packages/cli/src/features/setup.ts packages/cli/src/features/setup-types.ts`

NicholaiVogel:
pulled the newest blocker and patched it in `6d4df6cf`.

### fix
non-interactive `signet setup` now enforces workspace protection even on the existing-install early-return path.

changes in `setup.ts` existing-install + non-interactive branch:
- runs `enforceSetupProtection(...)` before daemon/start/open-dashboard return path
- honors `--create-local-backup` and `--allow-unprotected-workspace`
- prints protection summary in this path too

this closes the bypass where existing non-interactive runs previously skipped the new guardrails entirely.

### regression test added
- new `packages/cli/src/features/setup.test.ts`
  - `fails existing-install path when openclaw-linked workspace is unprotected`

### validation
- `bun test packages/cli/src/features/setup.test.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/features/setup.ts packages/cli/src/features/setup.test.ts packages/cli/src/features/setup-fresh.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-protection.ts packages/cli/src/features/setup-types.ts packages/cli/src/features/setup-protection.test.ts packages/cli/src/features/health.test.ts packages/cli/src/lib/workspace-protection.ts`

NicholaiVogel:
rebased/merge step complete: pulled latest `origin/main` into this branch and pushed (`65df3172`).

- merged `origin/main` cleanly (no conflict)
- branch now includes latest mainline version bump/changelog updates

i also re-checked latest automated review status: no new blocker findings were flagged on the feature changes, current head is ready for human re-review.

PR #379, opened 2026-03-28
chore(desktop): add cross-platform desktop packaging and release artifacts

NicholaiVogel:
## Summary

Implements first-class desktop packaging/distribution for Signet across macOS, Windows, Ubuntu, and Arch with bundled daemon sidecar fallback, signing preflight, and channel manifest generation.

## Changes

- Added desktop packaging spec + research and synced spec control plane:
  - `docs/specs/approved/desktop-packaging-distribution.md`
  - `docs/research/technical/RESEARCH-DESKTOP-PACKAGING-DISTRIBUTION.md`
  - `docs/specs/INDEX.md`, `docs/specs/dependencies.yaml`
- Updated tray runtime startup resolution to keep system/runtime paths first, with bundled `signet-daemon*` sidecars as fallback when system runtime is unavailable (Linux/macOS/Windows).
- Added tray daemon sidecar staging script (`packages/tray/scripts/stage-daemon.mjs`) and build wiring (`build:daemon`, `stage:daemon`, `build:desktop`).
- Hardened `.github/workflows/desktop-build.yml`:
  - signing secret preflight (mac + windows)
  - daemon-rs sidecar build/stage before tauri bundle
  - windows MSI signing step
  - Ubuntu AUR metadata generation artifact
  - Homebrew/winget manifest generation artifact from tagged release assets
- Added generator scripts + docs for publish channels:
  - `deploy/aur/generate-pkgbuild.sh`
  - `deploy/channels/generate-manifests.sh`
- Updated `docs/TRAY.md` with new packaging outputs and channel flow.

## Type

- [ ] `fix` — bug fix
- [x] `chore` — build, deps, config, docs
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `@signet/tray`, desktop CI/workflows, deploy channel scripts, specs/docs

## Screenshots

N/A, no dashboard/web/extension UI changes.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

No DB migrations in this PR.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Validation run for this PR scope:
- `cd packages/tray/src-tauri && cargo check` (pass)
- `cd packages/tray/src-tauri && cargo test` (pass, includes sidecar-name tests)
- `cd packages/tray && bun run build:ts` (pass)
- `cd packages/tray && bun run build:daemon && bun run stage:daemon` (pass)
- sidecar stage script smoke-tested with explicit `SIGNET_DAEMON_BIN`
- workflow YAML parse check (pass)

Known pre-existing blockers in repo baseline:
- `bun scripts/spec-deps-check.ts` fails on existing missing path `references/Engram`
- `bun run lint` reports many existing repo-wide formatting issues outside this PR

## AI disclosure

## Notes

- Signing steps require configured secrets in CI (`APPLE_*`, `WINDOWS_*`).
- Channel artifacts are generated and uploaded; publishing to external Homebrew/winget/AUR repos remains a follow-up automation layer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,41 +1,46 @@
 {
-	"name": "signet",
-	"version": "0.81.1",
-	"private": true,
-	"description": "Portable AI agent identity",
-	"repository": {
-		"type": "git",
-		"url": "https://github.com/signetai/signetai.git"
-	},
-	"author": "Signet AI",
-	"license": "Apache-2.0",
-	"workspaces": ["packages/*", "packages/adapters/*", "packages/cli/dashboard", "web"],
-	"scripts": {
-		"build": "bun run build:core && bun run build:connector-base && bun run build:opencode-plugin && bun run build:native && bun run build:deps && bun run build:signetai",
-		"build:native": "bun scripts/build-native.ts",
-		"build:core": "cd packages/core && bun run build",
-		"build:connector-base": "cd packages/connector-base && bun run build",
-		"build:opencode-plugin": "cd packages/opencode-plugin && bun run build",
-		"build:deps": "bun run --filter '@signet/connector-*' --filter '@signet/adapter-*' --filter '@signetai/*' --filter '@signet/daemon' --filter '@signet/sdk' --filter '@signet/cli' build",
-		"build:signetai": "cd packages/signetai && bun run build",
-		"build:publish": "bun run build:dashboard && bun run build:signetai",
-		"build:dashboard": "cd packages/cli/dashboard && bun install && bun run build",
-		"build:desktop": "cd packages/tray && bun run build:desktop",
-		"start": "cd packages/daemon && bun run start",
-		"dev": "bun run --filter '*' dev",
-		"test": "bun run --filter '*' test",
-		"lint": "biome check .",
-		"format": "biome format --write .",
-		"typecheck": "bun run --filter '*' typecheck",
-		"version:sync": "bun scripts/version-sync.ts",
-		"publish:signetai": "cd packages/signetai && npm publish --access public",
-		"dev:web": "cd web && bun run dev",
-		"deploy:web": "cd web && bun run deploy",
-		"changelog": "bun scripts/changelog.ts"
-	},
-	"devDependencies": {
-		"@biomejs/biome": "^1.9.0",
-		"bun-types": "^1.3.9",
-		"typescript": "^5.7.0"
-	}
+  "name": "signet",
+  "version": "0.82.1",
+  "private": true,
+  "description": "Portable AI agent identity",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/signetai/signetai.git"
+  },
+  "author": "Signet AI",
+  "license": "Apache-2.0",
+  "workspaces": [
+    "packages/*",
+    "packages/adapters/*",
+    "packages/cli/dashboard",
+    "web"
+  ],
+  "scripts": {
+    "build": "bun run build:core && bun run build:connector-base && bun run build:opencode-plugin && bun run build:native && bun run build:deps && bun run build:signetai",
+    "build:native": "bun scripts/build-native.ts",
+    "build:core": "cd packages/core && bun run build",
+    "build:connector-base": "cd packages/connector-base && bun run build",
+    "build:opencode-plugin": "cd packages/opencode-plugin && bun run build",
+    "build:deps": "bun run --filter '@signet/connector-*' --filter '@signet/adapter-*' --filter '@signetai/*' --filter '@signet/daemon' --filter '@signet/sdk' --filter '@signet/cli' build",
+    "build:signetai": "cd packages/signetai && bun run build",
+    "build:publish": "bun run build:dashboard && bun run build:signetai",
+    "build:dashboard": "cd packages/cli/dashboard && bun install && bun run build",
+    "build:desktop": "cd packages/tray && bun run build:desktop",
+    "start": "cd packages/daemon && bun run start",
+    "dev": "bun run --filter '*' dev",
+    "test": "bun run --filter '*' test",
+    "lint": "biome check .",
+    "format": "biome format --write .",
+    "typecheck": "bun run --filter '*' typecheck",
+    "version:sync": "bun scripts/version-sync.ts",
+    "publish:signetai": "cd packages/signetai && npm publish --access public",
+    "dev:web": "cd web && bun run dev",
+    "deploy:web": "cd web && bun run deploy",
+    "changelog": "bun scripts/changelog.ts"
+  },
+  "devDependencies": {
+    "@biomejs/biome": "^1.9.0",
+    "bun-types": "^1.3.9",
+    "typescript": "^5.7.0"
+  }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,40 @@
 
 All notable changes to Signet are documented here.
 
+## [0.82.0] - 2026-03-28
+
+### Bug Fixes
+
+- **cli**: protect OpenClaw-linked workspaces from unprotected data loss (#378)
+
+
+## [0.81.2] - 2026-03-28
+
+### Features
+
+- **setup**: deployment-aware setup defaults + native embedding docs alignment (#363)
+
+### Bug Fixes
+
+- **daemon**: batch watcher identity sync to keep health responsive (#375)
+
+### Docs
+
+- add Groundswell implementation specs + PRD + SSM synthesis (#381)
+- **specs**: add Groundswell PRD and gap analyses for community knowledge graphs
+
+
+## [0.81.1] - 2026-03-28
+
+### Bug Fixes
+
+- **daemon**: persist extraction fallback status with provider parity (replaces #367) (#373)
+
+### Docs
+
+- add alcar2364 to contributors
+
+
 ## [0.81.0] - 2026-03-28
 
 ### Bug Fixes
--- a/README.md
+++ b/README.md
@@ -409,7 +409,17 @@ cd packages/daemon && bun run dev        # Daemon dev (watch mode)
 cd packages/cli/dashboard && bun run dev # Dashboard dev
 ```
 
-Requirements: Node.js 18+, Bun, Ollama (recommended) or OpenAI API key. macOS or Linux.
+Requirements:
+
+- Node.js 18+ or Bun
+- macOS or Linux
+- Optional for harness integrations: Claude Code, Codex, OpenCode, or OpenClaw
+
+Embeddings (choose one):
+
+- **Built-in** (recommended) — no extra setup, runs locally via ONNX (`nomic-embed-text-v1.5`)
+- **Ollama** — alternative local option, requires `nomic-embed-text` model
+- **OpenAI** — cloud option, requires `OPENAI_API_KEY`
 
 ## Contributing
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -91,7 +91,9 @@ No authentication required. Lightweight liveness check.
 ### GET /api/status
 
 Full daemon status including pipeline config, embedding provider, and a
-composite health score derived from diagnostics.
+composite health score derived from diagnostics. Extraction provider
+runtime resolution persists startup degradation so operators can detect
+silent fallback or hard-blocked extraction after boot.
 
 **Response**
 
@@ -115,6 +117,30 @@ composite health score derived from diagnostics.
     "autonomousEnabled": false,
     "extractionModel": "qwen3:4b"
   },
+  "pipeline": {
+    "extraction": {
+      "running": true,
+      "overloaded": false,
+      "loadPerCpu": 0.42,
+      "maxLoadPerCpu": 0.8,
+      "overloadBackoffMs": 30000,
+      "overloadSince": null,
+      "nextTickInMs": 1200
+    }
+  },
+  "providerResolution": {
+    "extraction": {
+      "configured": "claude-code",
+      "resolved": "claude-code",
+      "effective": "ollama",
+      "fallbackProvider": "ollama",
+      "status": "degraded",
+      "degraded": true,
+      "fallbackApplied": true,
+      "reason": "Claude Code CLI not found during extraction startup preflight",
+      "since": "2026-03-26T06:00:00.000Z"
+    }
+  },
   "health": { "score": 0.97, "status": "healthy" },
   "embedding": {
     "provider": "ollama",
@@ -127,6 +153,10 @@ composite health score derived from diagnostics.
 
 The `bypassedSessions` field reports how many active sessions currently have
 bypass enabled (see [[#Sessions]]).
+Monitor `providerResolution.extraction.status` for `degraded` or `blocked`
+states when the configured extraction provider is unavailable at startup.
+When `pipeline.extraction.overloaded` is `true`, the extraction worker is
+intentionally backing off for `overloadBackoffMs` between polls.
 
 
 ### GET /api/features
--- a/docs/CLI.md
+++ b/docs/CLI.md
@@ -110,8 +110,8 @@ signet setup --path /custom/path
 signet setup --non-interactive \
   --name "My Agent" \
   --harness claude-code \
-  --embedding-provider ollama \
-  --extraction-provider claude-code
+  --deployment-type vps \
+  --embedding-provider native
 ```
 
 Options:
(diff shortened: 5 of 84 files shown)


BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `chore(desktop): add cross-platform desktop packaging and release artifacts` (commit `67294bb8`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `chore(desktop): add cross-platform desktop packaging and release artifacts` (commit `67294bb8`) now and I'll follow up shortly with feedback.

.github/workflows/desktop-build.yml
```diff
@@ -35,22 +38,235 @@ jobs:
...
+            [Convert]::FromBase64String($env:WINDOWS_CERTIFICATE)
+          )
+
+          $msiFiles = Get-ChildItem -Path "packages/tray/src-tauri/target/$env:TARGET/release/bundle/msi" -Filter *.msi -Recurse
+          if ($msiFiles.Count -eq 0) {
+            Write-Error "No MSI artifacts found for signing"
+            exit 1
+          }
+
+          foreach ($msi in $msiFiles) {
```
PR-Reviewer-Ant, line 144: The new Windows MSI signing step does not actually enforce successful signing. In PowerShell, invoking an external program with `& signtool ...` does not throw on non-zero exit by default, so a failed signing attempt can still leave the job green and publish unsigned MSI artifacts. That breaks the PR's stated signing contract. Add explicit `$LASTEXITCODE` checking (or set PowerShell to stop on native command errors) after each `signtool` invocation.

packages/tray/scripts/stage-daemon.mjs
```diff
@@ -0,0 +1,87 @@
...
+const triple = target();
+if (!triple) {
+	throw new Error("Could not resolve target triple for daemon sidecar staging");
+}
+
+const fromEnv = process.env.SIGNET_DAEMON_BIN ?? null;
+const bin = `signet-daemon${ext}`;
+const here = resolve(fileURLToPath(import.meta.url), "..");
+const root = resolve(here, "..");
+const sourceList = [
```
PR-Reviewer-Ant: The staging fallback can package the wrong daemon binary for the requested target. If the expected `daemon-rs/target/<triple>/release/...` output is missing, the script falls through to `target/release` and then `which/where signet-daemon`, with no validation that the discovered binary matches `TAURI_ENV_TARGET_TRIPLE`. On cross-target builds this can silently embed a host-arch daemon under a target-specific filename, producing a broken desktop app instead of failing fast.

.github/workflows/desktop-build.yml
NicholaiVogel: great catch, fixed in `bc0727a6`.

windows signing now explicitly fails on native command errors:
- sets `$ErrorActionPreference = "Stop"`
- sets `$PSNativeCommandUseErrorActionPreference = $true`
- checks `$LASTEXITCODE` after each `signtool` call and exits non-zero

this prevents unsigned MSI artifacts from passing the job.

packages/tray/scripts/stage-daemon.mjs
NicholaiVogel: good call, fixed in `bc0727a6`.

`stage-daemon.mjs` is now target-safe:
- detects cross-target builds (`triple != host`)
- disables host/PATH fallbacks in cross-target mode
- only allows explicit env override in cross-target mode when the provided path includes the requested target triple
- otherwise fails fast with a clear error

this blocks silent host-arch sidecar packaging on cross-target builds.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `chore(desktop): add cross-platform desktop packaging and release artifacts` (commit `bc0727a6`) now and I'll follow up shortly with feedback.

NicholaiVogel:
addressed the blocking review feedback in commit bc0727a6:\n\n- enforced windows signing failure behavior (, ,  guard)\n- made sidecar staging target-safe for cross-target builds (disabled host/path fallbacks, added fast-fail validation)\n\nreplies are posted on both inline threads.

.github/workflows/desktop-build.yml
```diff
@@ -35,22 +38,243 @@ jobs:
...
         run: cd packages/tray && bun install
 
+      - name: Build bundled daemon runtime
+        run: |
+          cargo build \
+            --manifest-path packages/daemon-rs/Cargo.toml \
+            --release \
+            --target "${{ matrix.target }}" \
+            -p signet-daemon
+
```
PR-Reviewer-Ant, line 104: The PR claims signed desktop distribution, but the Windows MSI is signed too late. `tauri-apps/tauri-action@v0` is the step that creates/uploads the release artifacts for tag builds, and the new `Sign Windows MSI artifacts` step runs only afterward. That means the GitHub release asset downloaded by users and by the later `channel-manifests` job is still the unsigned MSI, while only the local workspace copy gets signed.

packages/tray/src-tauri/src/platform/mod.rs
```diff
@@ -18,6 +20,94 @@ pub mod autostart;
...
+    if !name.starts_with("signet-daemon") {
+        return false;
+    }
+
+    #[cfg(target_os = "windows")]
+    {
+        return name.ends_with(".exe");
+    }
+
+    #[cfg(not(target_os = "windows"))]
```
PR-Reviewer-Ant, line 55: Bundled daemon selection is not target-specific. `scan_dir()` returns the first file whose name starts with `signet-daemon`, and `find_bundled_daemon()` never checks the current target triple/arch. Because the new staging flow copies target-named binaries into `src-tauri/binaries/` without cleaning older ones, repeated local packaging across targets can leave multiple sidecars around; the app may then launch the wrong-architecture daemon nondeterministically.

.github/workflows/desktop-build.yml
NicholaiVogel: solid catch, fixed in `92fccde4`.

workflow now builds first, signs windows MSI, then uploads release assets explicitly via `gh release upload`.

concretely:
- removed `tauri-action` release upload path
- added explicit `Build desktop bundle` step
- kept Windows signing before any upload
- added `Upload desktop release assets` step after signing
- added `Ensure release exists` guard for matrix uploads

this makes the uploaded/tagged MSI the signed artifact used by users and downstream manifest generation.

packages/tray/src-tauri/src/platform/mod.rs
NicholaiVogel: good call, fixed in `92fccde4`.

runtime sidecar selection is now deterministic and target-specific:
- `find_bundled_daemon()` first resolves exact target filename (`signet-daemon-<target-triple>[.exe]`)
- only then falls back to plain `signet-daemon[.exe]`
- removed prefix/first-match scanning that could pick wrong architecture

plus staging now cleans old `signet-daemon-*` files before writing the current target sidecar, so stale binaries don't accumulate between builds.

NicholaiVogel:
pushed another round in commit 92fccde4 to address the new blocking feedback:

- release upload now happens only after Windows MSI signing
- sidecar discovery now uses exact target-triple filenames, not first-match prefix scanning
- sidecar staging now removes stale signet-daemon-* files before writing the current target

left inline replies on both new threads.

package.json
```diff
@@ -1,45 +1,41 @@
...
+	"scripts": {
+		"build": "bun run build:core && bun run build:connector-base && bun run build:opencode-plugin && bun run build:native && bun run build:deps && bun run build:signetai",
+		"build:native": "bun scripts/build-native.ts",
+		"build:core": "cd packages/core && bun run build",
+		"build:connector-base": "cd packages/connector-base && bun run build",
+		"build:opencode-plugin": "cd packages/opencode-plugin && bun run build",
+		"build:deps": "bun run --filter '@signet/connector-*' --filter '@signet/adapter-*' --filter '@signetai/*' --filter '@signet/daemon' --filter '@signet/sdk' --filter '@signet/cli' build",
+		"build:signetai": "cd packages/signetai && bun run build",
+		"build:publish": "bun run build:dashboard && bun run build:signetai",
+		"build:dashboard": "cd packages/cli/dashboard && bun install && bun run build",
```
PR-Reviewer-Ant: The new `build:desktop` entrypoint does not satisfy the PR's own packaging contract. It just shells into `packages/tray && bun run build`, and `@signet/tray`'s `build` script is still plain `tauri build`, so the required `stage:daemon` step is skipped for local/manual builds. That means `bun run build:desktop` can produce artifacts without the bundled `signet-daemon*` sidecar even though the PR description and docs now claim first-class bundled-runtime packaging.

.github/workflows/desktop-build.yml
```diff
@@ -35,22 +38,275 @@ jobs:
...
+            patchelf
 
       - name: Install dependencies
         run: bun install
 
       - name: Install tray dependencies
         run: cd packages/tray && bun install
 
-      - uses: tauri-apps/tauri-action@v0
+      - name: Build bundled daemon runtime
```
PR-Reviewer-Ant, line 97: This PR changes the packaged desktop runtime from the existing system/JS daemon path to a bundled `packages/daemon-rs` binary and then makes the tray prefer that bundled binary first on every platform. The repository contract in AGENTS.md still describes `packages/daemon-rs/` as a shadow rewrite with parity cutover deferred, so shipping it as the default desktop runtime is a breaking behavior change without any parity gate or degraded-mode documentation. If the Rust daemon is not feature-complete, desktop builds will diverge from `signet`/system installs in ways users cannot opt out of.

package.json
NicholaiVogel: good catch, fixed in `e39a61bc`.

`build:desktop` now goes through a real desktop packaging chain instead of plain `tauri build`:

- root `build:desktop` -> `packages/tray` `build:desktop`
- tray `build:desktop` now runs:
  1) `build:daemon` (`cargo build ... -p signet-daemon`)
  2) `stage:daemon`
  3) `tauri build`

that guarantees local/manual desktop builds stage the bundled daemon sidecar before packaging.

.github/workflows/desktop-build.yml
NicholaiVogel: thanks, agreed on the parity concern. fixed in `e39a61bc`.

runtime start order is now system-first again on all platforms, and bundled sidecar use is fallback-only:

- removed bundled-daemon-first startup from linux/macos/windows managers
- tray now tries existing system/runtime paths first (`signet`, `signet-daemon`, bun+daemon.js/bunx)
- only falls back to bundled `signet-daemon-*` when system runtime is unavailable

i also aligned the contracts/docs in the same commit:
- `docs/specs/approved/desktop-packaging-distribution.md`
- `docs/specs/dependencies.yaml`
- `docs/specs/INDEX.md`
- `docs/TRAY.md`

those now explicitly describe bundled daemon as a fallback path, and call out `packages/daemon-rs` as shadow rewrite until parity cutover approval.

NicholaiVogel:
pushed follow-up commit `e39a61bc` for the latest blocking review:

- fixed local/manual desktop entrypoint so `build:desktop` now runs full sidecar flow (`build:daemon` -> `stage:daemon` -> `tauri build`)
- changed tray startup order back to system-first; bundled `daemon-rs` sidecar is now fallback-only
- synced spec/docs contract text to match fallback behavior and shadow-daemon status

validation done:
- `cd packages/tray/src-tauri && cargo test`
- `cd packages/tray && bun run build:daemon && bun run stage:daemon`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `e39a61bc`

[Automated Review] Found one behavior/documentation mismatch worth human follow-up: the PR description still claims bundled daemon preference, but the implementation now only uses the bundled sidecar as a fallback.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, ambiguous_requirements] - The changed runtime code in the platform managers clearly checks system-installed runtimes first and only calls `find_bundled_daemon()` afterward, so the bundled daemon is fallback-only in this revision. At the same time, the PR description still says the tray runtime was updated to 'prefer bundled signet-daemon* sidecars before system fallbacks', and the research doc still states the app must prefer the bundled runtime first. The code evidence is direct; the only ambiguity is whether the description/research text or the final implementation reflects the intended product decision.

packages/tray/src-tauri/src/platform/linux.rs
```diff
@@ -79,36 +76,35 @@ impl DaemonManager for LinuxManager {
...
-            .ok_or("bun not found — install bun to run signet daemon")?;
-
-        // Try `signet-daemon` binary first (global install)
-        if let Ok(output) = Command::new("which").arg("signet-daemon").output() {
-            if output.status.success() {
-                let bin = String::from_utf8_lossy(&output.stdout).trim().to_string();
-                Command::new(&bun)
-                    .arg(&bin)
-                    .spawn()?;
+        if let Some(bun) = self.find_bun() {
```
PR-Reviewer-Ant, line 108: The implementation no longer matches the PR summary's stated goal. `start()` now tries system-installed Bun/CLI paths first and only reaches `find_bundled_daemon()` afterward, so packaged desktop builds still inherit whatever runtime happens to be installed on the host. If the intended contract is now 'bundled sidecar as fallback only', the PR description and supporting research/doc text need to be updated to say that explicitly; otherwise this change does not actually deliver the 'prefer bundled runtime' behavior it claims.

NicholaiVogel: solid callout, agreed. aligned now.

changes made:
- updated PR description to explicitly state bundled sidecar is fallback-only
- updated research doc wording in `docs/research/technical/RESEARCH-DESKTOP-PACKAGING-DISTRIBUTION.md` to match that contract

commit: `5f50d91e`

this removes the preference-vs-implementation mismatch and keeps the runtime decision consistent across code, spec/docs, and PR narrative.

NicholaiVogel:
addressed the new review note in `5f50d91e`:

- updated `RESEARCH-DESKTOP-PACKAGING-DISTRIBUTION.md` to match the implemented runtime contract (bundled sidecar = fallback path)
- updated PR description so summary/changes text no longer claims bundled-first behavior

this should resolve the behavior-vs-documentation mismatch called out on the latest pass.

packages/tray/scripts/stage-daemon.mjs
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env node
+
+import { execSync } from "node:child_process";
+import { chmodSync, copyFileSync, existsSync, mkdirSync, readdirSync, rmSync } from "node:fs";
+import { resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const win = process.platform === "win32";
+const ext = win ? ".exe" : "";
```
PR-Reviewer-Ant: The sidecar filename extension is chosen from the build host (`process.platform`) rather than the resolved target triple. That means a cross-target Windows build launched from macOS/Linux will look for and stage `signet-daemon` instead of `signet-daemon.exe`, despite the rest of this script explicitly supporting cross-target mode. If target-aware staging is part of the contract here, the extension needs to come from `triple`, not the host OS.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5f50d91e`

[Automated Review] Found one packaging logic issue that still undermines the claimed target-aware sidecar staging.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The issue is directly visible in `packages/tray/scripts/stage-daemon.mjs`: the staged binary extension is derived from `process.platform` instead of the resolved target triple, even though the script otherwise has explicit cross-target handling via `TAURI_ENV_TARGET_TRIPLE` and `cross`. I did not run an actual non-Windows-host -> Windows-target packaging build in this environment.

packages/tray/scripts/stage-daemon.mjs
NicholaiVogel: great catch, fixed in `66192f15`.

`stage-daemon.mjs` now derives sidecar extension from the resolved target triple, not host `process.platform`:

- added `targetExt(triple)`
- `bin`/output names now use target-derived extension (`.exe` only for `*-windows-*` targets)
- chmod now keys off target extension (`ext === ""`) instead of host OS

this keeps cross-target staging consistent (including non-Windows host -> Windows target builds).

NicholaiVogel:
pushed follow-up fix in `66192f15` for the latest staging feedback:

- `stage-daemon.mjs` now derives `.exe` vs no extension from the resolved target triple, not host platform
- output sidecar naming and source resolution are now target-consistent for cross-target builds
- chmod behavior now keys off target extension

quick validation:
- `cd packages/tray && bun run stage:daemon` (pass)

deploy/aur/generate-pkgbuild.sh
```diff
@@ -0,0 +1,56 @@
...
+cat > "${OUT_DIR}/.SRCINFO" <<EOF
+pkgbase = signet-desktop-bin
+	pkgdesc = Signet desktop application (Tauri) with bundled daemon runtime
+	pkgver = ${VER}
+	pkgrel = 1
+	url = https://github.com/Signet-AI/signetai
+	arch = x86_64
+	license = Apache
+	depends = glibc
+	depends = gtk3
```
PR-Reviewer-Ant, line 54: Blocking: the generated `.SRCINFO` writes `source = signet-${pkgver}.AppImage::...` literally because `${pkgver}` is escaped in the heredoc. `.SRCINFO` is static metadata, not a shell script, so AUR tooling will see a filename containing `${pkgver}` instead of the real versioned AppImage name. This breaks the PR's claim that Arch metadata is generated deterministically from the release asset.

deploy/aur/.gitignore
```diff
@@ -0,0 +1,2 @@
+PKGBUILD
```
PR-Reviewer-Ant, line 1: The implementation diverges from the spec introduced in this PR. `docs/specs/approved/desktop-packaging-distribution.md` says generated AUR metadata is source-controlled, but this `.gitignore` excludes both generated files and `deploy/aur/README.md` instructs publishing them only as CI artifacts. That leaves the spec/control-plane out of sync with the actual release process and should be reconciled in the same PR.

deploy/aur/generate-pkgbuild.sh
NicholaiVogel: good catch, fixed in `6400cf1f`.

AUR metadata generation now emits concrete static source names:

- derive `ASSET` from the provided release URL
- use `${ASSET}::${URL}` in both `PKGBUILD` and `.SRCINFO`
- install from `${srcdir}/${ASSET}`
- added guard for empty asset derivation

so `.SRCINFO` no longer contains a literal `\${pkgver}` token.

deploy/aur/.gitignore
NicholaiVogel: agreed, fixed in `6400cf1f`.

updated the approved spec integration note to match the implemented publication flow:

- changed from “source-controlled” to “emitted as CI artifacts”
- kept the same contract that a separate credentialed job can publish those generated files

that brings spec/control-plane text in sync with `deploy/aur/.gitignore` and `deploy/aur/README.md`.

NicholaiVogel:
pushed `6400cf1f` for the latest review round.

resolved items:
- fixed AUR generator so `.SRCINFO` uses concrete source filename metadata (no literal `\${pkgver}` token)
- aligned approved spec integration note with implemented AUR flow (CI artifacts, then optional credentialed publish step)

replies are posted on both inline threads.

packages/tray/src-tauri/src/platform/linux.rs
```diff
@@ -79,36 +76,35 @@ impl DaemonManager for LinuxManager {
...
+            // Try `signet-daemon` binary first (global install)
+            if let Ok(output) = Command::new("which").arg("signet-daemon").output() {
+                if output.status.success() {
+                    let bin = String::from_utf8_lossy(&output.stdout).trim().to_string();
+                    Command::new(&bun).arg(&bin).spawn()?;
+                    return Ok(());
+                }
+            }
+
```
PR-Reviewer-Ant, line 117: Blocking: the Linux startup path still does not honor the system-runtime order claimed by this PR. macOS and Windows both try the installed `signet` CLI first, but Linux never does; it falls through to `bun x signetai daemon start` instead. On machines where the normal system install provides `signet` but not a separate `signet-daemon` entrypoint, the tray will bypass the intended local runtime and take a different code path than the one this PR documents.

NicholaiVogel: good catch, fixed.

runtime fix landed in `381db411`:
- Linux now tries installed `signet daemon start` before Bun-driven fallbacks
- start order is now systemd -> signet CLI -> bun-based paths -> bundled sidecar fallback

docs follow-up landed in `d2ce9dbc`:
- `docs/TRAY.md` Linux start-order text now matches the implemented flow

NicholaiVogel:
latest feedback addressed.

pushed:
- `381db411` — linux runtime now tries installed `signet daemon start` before bun fallbacks
- `d2ce9dbc` — docs/TRAY linux start-order text aligned with implementation

also replied inline on the blocking linux parity thread.

NicholaiVogel:
latest automated pass on `d2ce9dbc` reports no new issues and says prior blocking items appear resolved.

keeping this open for human maintainer review/approval now.

NicholaiVogel:
resolved merge conflicts by merging `origin/main` into this branch.

merge commit: `99209d26`

conflicts were in:
- `package.json`
- `packages/tray/package.json`

resolution kept mainline version bumps (`0.82.1`) and preserved this PR's desktop packaging scripts (`build:desktop`, `build:daemon`, `stage:daemon`).

PR #386, opened 2026-03-28
feat(oh-my-pi): add Oh My Pi support

maximhar:
## Summary

Add Oh My Pi support through a managed Signet runtime extension and connector, while keeping the non-Oh-My-Pi install path clean.

## Changes

- add `@signet/connector-oh-my-pi` and `@signet/oh-my-pi-extension`
- forward Oh My Pi session, prompt, and compaction lifecycle events to the daemon, with hidden Signet context injected locally in the runtime
- document the Oh My Pi harness in `docs/HARNESSES.md`
- add spec-system coverage for the new harness integration in `docs/specs/INDEX.md` and `docs/specs/dependencies.yaml`
- bundle the built Oh My Pi extension artifact into the connector so `@signet/cli` does not transitively install `@oh-my-pi/pi-coding-agent` for non-Oh-My-Pi users
- relax noisy prompt-submit timeout handling in the Oh My Pi extension and add regression tests

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `@signet/oh-my-pi-extension`, docs/spec registry

## Screenshots

- [ ] UI change screenshots attached
- [x] N/A

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

N/A — no schema migrations are touched in this PR.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Focused validation run locally:
- `bun x biome check packages/connector-oh-my-pi packages/oh-my-pi-extension docs/specs/INDEX.md docs/specs/dependencies.yaml docs/specs/planning/connector-oh-my-pi.md`
- `cd packages/oh-my-pi-extension && bun test daemon-client.test.ts transcript.test.ts session-state.test.ts lifecycle.test.ts`
- `cd packages/oh-my-pi-extension && bun run build && bun run typecheck`
- `cd packages/connector-oh-my-pi && bun test index.test.ts`
- `cd packages/connector-oh-my-pi && bun run build && bun run typecheck`

## AI disclosure

## Notes

- The connector now embeds the built Oh My Pi extension artifact so non-Oh-My-Pi users do not inherit the heavy Oh My Pi dependency chain through `@signet/cli`.
- Oh My Pi prompt-submit now uses the standard read timeout budget and logs concise timeout warnings instead of dumping raw `DOMException` objects.
- Previous-session end submission now defers until the prior session file can actually be reconstructed, avoiding silent loss of transcript-backed extraction on switch or branch.

### DIFF
--- a/packages/oh-my-pi-extension/src/session-state.ts
+++ b/packages/oh-my-pi-extension/src/session-state.ts
@@ -12,12 +12,17 @@ export interface PendingSessionEnd {
 	readonly reason: string;
 }
 
+function sanitizeInject(inject: string): string {
+	// Prevent injected content from closing the signet-memory wrapper early.
+	return inject.replace(/<\/signet-memory>/gi, "<\\/signet-memory>");
+}
+
 function createHiddenInjectMessage(customType: string, inject: string): OmpAgentMessage {
 	return {
 		role: "custom",
 		customType,
 		display: false,
-		content: `<signet-memory source="auto-recall">\n${inject}\n</signet-memory>`,
+		content: `<signet-memory source="auto-recall">\n${sanitizeInject(inject)}\n</signet-memory>`,
 		timestamp: Date.now(),
 	};
 }


PR-Reviewer-Ant:
Hi @maximhar - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(oh-my-pi): add Oh My Pi support` (commit `620e7a30`) now and I'll follow up shortly with feedback.

docs/HARNESSES.md
```diff
@@ -350,6 +350,51 @@ The plugin handles lifecycle hooks; MCP provides on-demand memory tools.
 
 ---
 
+## Oh My Pi (`oh-my-pi`)
```
PR-Reviewer-Ant, line 353: This PR introduces a new harness integration, but the required spec-system artifacts were not updated in the same change. The repo's mandatory session protocol says new feature work must be represented in `docs/specs/INDEX.md` and validated against `docs/specs/dependencies.yaml`; neither file appears in the diff. That leaves the dependency graph and integration contracts stale for a cross-package harness feature.

packages/oh-my-pi-extension/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "bun build src/index.ts --outfile dist/signet-oh-my-pi.mjs --target node --minify --format esm --external better-sqlite3 && bun run build:types",
+    "build:types": "tsc --emitDeclarationOnly --declaration --outDir dist --noCheck",
+    "typecheck": "tsc --noEmit --noCheck",
+    "dev": "bun --watch src/index.ts"
+  },
+  "dependencies": {
```
PR-Reviewer-Ant: This makes the core CLI path transitively install `@oh-my-pi/pi-coding-agent` for all users, not just Oh My Pi users, because `@signet/cli` now always depends on `@signet/connector-oh-my-pi` and that connector depends on this package. The resulting lockfile pulls in heavy new transitive dependencies including `puppeteer`, which materially expands install size and supply-chain attack surface beyond what the PR description calls out. Consider making this optional or bundling the extension artifact so non-Oh-My-Pi users do not pay this cost.

packages/oh-my-pi-extension/src/lifecycle.ts
```diff
@@ -0,0 +1,187 @@
...
+	if (deps.state.sessionAlreadyEnded(session.sessionId)) return;
+
+	await submitSessionEnd(deps.client, {
+		sessionId: session.sessionId,
+		transcript: buildTranscriptFromEntries(getSessionEntries(ctx)),
+		reason,
+		project: session.project,
+	});
+
+	deps.state.markSessionEnded(session.sessionId);
```
PR-Reviewer-Ant, line 163: This can silently lose session-end extraction data. `readSessionFileSnapshot()` returns `{ transcript: undefined }` on any read/parse problem, but this code still posts `session-end`, marks the session ended, and clears pending state. If Oh My Pi switches/branches before the previous session file is fully flushed, Signet will permanently skip transcript-backed extraction for that session. Please avoid marking the session ended when the snapshot could not be reconstructed, or add a retry/fallback path that preserves the session until a transcript is actually available.

packages/oh-my-pi-extension/src/index.ts
```diff
@@ -0,0 +1,145 @@
...
+		deps.state.clearPendingRecall(session.sessionId);
+		await requestRecallForPrompt(deps, ctx, event.text);
+	});
+
+	pi.on("before_agent_start", async (event: OmpBeforeAgentStartEvent, ctx) => {
+		await ensureSessionContext(deps, ctx);
+		const session = currentSessionRef(ctx);
+		if (!session.sessionId) return;
+		if (deps.state.hasPendingRecall(session.sessionId)) return;
+		await requestRecallForPrompt(deps, ctx, event.prompt);
```
PR-Reviewer-Ant, line 63: The PR description says the extension 'forward[s] lifecycle, prompt, context, and compaction events to the daemon', but the `context` handler never calls the daemon. It only injects locally queued messages. Either the description/docs should be narrowed, or this handler should actually forward context events if that behavior is required.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `cca75175`

[Automated Review] Found one data-loss risk in session shutdown handling and one PR-description/implementation mismatch.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The session-end path in `packages/oh-my-pi-extension/src/lifecycle.ts` unconditionally marks the previous session as ended even when `readSessionFileSnapshot()` returns no transcript, which is directly visible in this diff and can drop extraction input on switch/branch. I did not run Oh My Pi itself, so event timing around session-file persistence is inferred from the code rather than reproduced against the runtime.

docs/HARNESSES.md
maximhar: Addressed in `fb791d0d`.

I added the missing spec-system artifacts for the Oh My Pi harness feature:
- `docs/specs/planning/connector-oh-my-pi.md`
- `docs/specs/INDEX.md`
- `docs/specs/dependencies.yaml`

The new stub is registered as `connector-oh-my-pi` with a hard dependency on `signet-runtime`, so the harness feature is now represented in the registry and dependency graph.

packages/oh-my-pi-extension/package.json
maximhar: Addressed in `fb791d0d`.

I removed the connector's runtime dependency on `@signet/oh-my-pi-extension` and switched `@signet/connector-oh-my-pi` to the same bundled-artifact pattern used by the OpenCode connector:
- `scripts/embed-extension.ts` now embeds `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs`
- `src/index.ts` installs from the embedded bundle instead of `require.resolve()`-ing the extension package
- `package.json` no longer depends on `@signet/oh-my-pi-extension`
- added `index.test.ts` coverage for the managed extension install path

That keeps the Oh My Pi feature available while avoiding the extra Oh-My-Pi-only dependency chain for users who never install that harness.

packages/connector-oh-my-pi/scripts/embed-extension.ts
```diff
@@ -0,0 +1,44 @@
...
+import { existsSync, readFileSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = fileURLToPath(new URL(".", import.meta.url));
+const packageRoot = join(__dirname, "..");
+const extensionDistPath = join(packageRoot, "..", "oh-my-pi-extension", "dist", "signet-oh-my-pi.mjs");
+const bundleSourcePath = join(packageRoot, "src", "extension-bundle.ts");
+
+if (!existsSync(extensionDistPath)) {
```
PR-Reviewer-Ant, line 25: This build step fails open: if the extension artifact has not been built yet, the script exits 0 and leaves whatever is currently in `src/extension-bundle.ts`. That makes the connector package build non-hermetic and can silently publish a stale or broken managed extension bundle. Since `packages/connector-oh-my-pi/package.json` invokes this during `build`, the safer behavior is to fail the build when the source artifact is missing, or regenerate from a guaranteed input in the same build graph.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `fb791d0d`

[Automated Review] Found one release-path issue that can silently ship a broken or stale Oh My Pi connector bundle.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new connector build script exits successfully when `packages/oh-my-pi-extension/dist/signet-oh-my-pi.mjs` is missing, even though `packages/connector-oh-my-pi/package.json` now relies on that script during `build`. That means `cd packages/connector-oh-my-pi && bun run build` can succeed without regenerating the embedded extension, leaving `src/extension-bundle.ts` stale or empty while still producing publishable output.

packages/oh-my-pi-extension/src/lifecycle.ts
maximhar: Addressed in `1a060374`.

I changed the previous-session end path so we no longer mark the session ended when the snapshot cannot be reconstructed:
- `readSessionFileSnapshot()` now reports whether the session file was actually loaded
- `endPreviousSession()` queues a pending session-end retry when the file is missing/unreadable instead of immediately clearing state
- `flushPendingSessionEnds()` retries later and only marks the session ended once the prior session file can be read
- added `lifecycle.test.ts` coverage for the deferred flush path

That preserves transcript-backed extraction instead of silently dropping it during switch/branch races.

packages/oh-my-pi-extension/src/index.ts
maximhar: Good catch. I narrowed the PR description to match the actual behavior.

The `context` hook is local-only: it injects queued hidden Signet messages into the Oh My Pi runtime, but it does not call a daemon endpoint. The PR body now says the extension forwards session, prompt, and compaction lifecycle events to the daemon, with hidden Signet context injected locally in the runtime.

packages/connector-oh-my-pi/scripts/embed-extension.ts
PR-Reviewer-Ant, line 25: The build still fails open here. If `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs` is missing, this exits 0 and leaves whatever was previously checked into `src/extension-bundle.ts`. That makes `bun run build` for this package non-hermetic and can silently ship a stale or broken managed extension bundle. This should fail the build instead of succeeding.

packages/oh-my-pi-extension/src/lifecycle.ts
```diff
@@ -0,0 +1,224 @@
...
+		}
+
+		const snapshot = readSessionFileSnapshot(pending.sessionFile);
+		if (!snapshot.loaded) continue;
+
+		await submitSessionEnd(deps.client, {
+			sessionId: snapshot.sessionId ?? pending.sessionId,
+			transcript: snapshot.transcript,
+			reason: pending.reason,
+			project: snapshot.project,
```
PR-Reviewer-Ant, line 100: This still silently loses session-end extraction on daemon/network failure. `submitSessionEnd()` goes through `DaemonClient.post()`, which returns `null` on timeout or fetch error, but this code marks the session ended and clears pending state unconditionally. The same pattern is used in the immediate session-end paths below. If the daemon is briefly unavailable during switch/shutdown, the transcript is dropped permanently with no retry. For a fail-open path, the runtime can continue, but the session should stay pending until the post succeeds.

packages/oh-my-pi-extension/package.json
maximhar: Follow-up addressed in `2b8f5d35`.

The embed step no longer exits cleanly when `packages/oh-my-pi-extension/dist/signet-oh-my-pi.mjs` is missing. `scripts/embed-extension.ts` now throws instead, so `cd packages/connector-oh-my-pi && bun run build` cannot publish or package a stale/empty embedded bundle. The validated path is now:
- build `@signet/oh-my-pi-extension`
- run the connector build, which embeds the fresh artifact and then bundles the connector

I re-ran the Oh My Pi extension build plus connector test/build/typecheck locally after the change.

bun.lock
```diff
@@ -129,9 +130,21 @@
         "typescript": "^5.7.0",
       },
     },
+    "packages/connector-oh-my-pi": {
```
PR-Reviewer-Ant, line 133: This still contradicts the PR's main packaging claim. `@signet/cli` now depends on `@signet/connector-oh-my-pi`, and the lockfile shows `@signet/connector-oh-my-pi` pulling in `@signet/oh-my-pi-extension`, which in turn pulls `@oh-my-pi/pi-coding-agent`. That means non-Oh-My-Pi users still inherit the heavy Oh My Pi dependency chain during install, even if the connector embeds the built artifact at runtime.

packages/connector-oh-my-pi/src/extension-bundle.ts
```diff
@@ -0,0 +1,1482 @@
+/**
+ * Bundled Oh My Pi extension content
+ *
+ * Auto-generated by scripts/embed-extension.ts — do not edit manually.
+ * Re-run the build to regenerate from ../oh-my-pi-extension/dist/signet-oh-my-pi.mjs
+ */
+
```
PR-Reviewer-Ant, line 7: The bundled extension hardcodes the daemon base URL to `http://localhost:3850`. That breaks the new harness whenever Signet is configured with a non-default host/port or non-local network mode (for example Tailscale/remote access), so the implementation does not actually match the repository's configurable daemon networking behavior. The extension needs to derive the daemon endpoint from the same config/env path the rest of the harness integrations use.

packages/oh-my-pi-extension/src/lifecycle.ts
maximhar: Follow-up addressed in `26423c7b`.

I also fixed the second reliability issue in the session-end path: we no longer mark a session ended when `DaemonClient.post()` returns `null` on timeout/network failure.

Concretely:
- `submitSessionEnd()` now returns success/failure
- `endCurrentSession()` only marks the session ended after a successful daemon submission
- `endPreviousSession()` re-queues the pending previous-session end when the daemon submission fails
- `flushPendingSessionEnds()` also preserves queued retries on daemon failure
- added regression coverage in `lifecycle.test.ts` for both the deferred-previous-session path and failed current-session shutdown submit

That keeps transcript-backed extraction retryable instead of dropping it on transient daemon failures.

packages/oh-my-pi-extension/src/index.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const SignetOhMyPiExtension: OmpExtensionFactory = (pi): void => {
+	if (readRuntimeEnv("SIGNET_ENABLED") === "false") {
+		return;
+	}
+
+	const deps: LifecycleDeps = {
+		agentId: readRuntimeEnv("SIGNET_AGENT_ID"),
+		client: createDaemonClient(readRuntimeEnv("SIGNET_DAEMON_URL") ?? DAEMON_URL_DEFAULT),
+		state: createSessionState(),
```
PR-Reviewer-Ant, line 137: This harness only reaches the daemon correctly if the Oh My Pi runtime already has `SIGNET_DAEMON_URL` set, but the new connector install path never provisions that env var. As shipped, any setup using a non-default host/port or non-local network mode will silently fall back to `http://localhost:3850`, so the implementation does not actually honor Signet's configured daemon networking.

packages/oh-my-pi-extension/src/index.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+const SignetOhMyPiExtension: OmpExtensionFactory = (pi): void => {
+	if (readRuntimeEnv("SIGNET_ENABLED") === "false") {
+		return;
+	}
+
+	const deps: LifecycleDeps = {
+		agentId: readRuntimeEnv("SIGNET_AGENT_ID"),
+		client: createDaemonClient(readRuntimeEnv("SIGNET_DAEMON_URL") ?? DAEMON_URL_DEFAULT),
```
PR-Reviewer-Ant: All hook submissions depend on `agentId` from `SIGNET_AGENT_ID`, but the managed Oh My Pi connector never writes or injects that value. That means this harness can send lifecycle and recall traffic without an agent scope, which violates the repo's explicit `agent_id` guardrail for user-data paths and risks default-scope leakage or incorrect memory attribution.

PR-Reviewer-Ant: This extension reads `agentId` only from `SIGNET_AGENT_ID`, but the new Oh My Pi connector never provisions that value anywhere. As installed, lifecycle and recall requests can run without an agent scope, which is exactly the scoping leak the repo guardrails call out for user-data paths. The connector needs to inject or persist the agent identity for the runtime instead of relying on an external env var being preconfigured.

packages/oh-my-pi-extension/src/index.ts
PR-Reviewer-Ant, line 137: The daemon client falls back to `http://localhost:3850`, and the Oh My Pi connector does not write `SIGNET_DAEMON_URL` or any equivalent runtime config. That means this new harness ignores Signet's configured host/port/network mode unless the user manually exports an env var out of band, so the install path does not actually deliver the daemon connectivity the PR claims.

packages/adapters/openclaw/package.json
```diff
@@ -36,7 +36,7 @@
     "@sinclair/typebox": "0.34.47"
   },
   "devDependencies": {
-    "@signet/core": "workspace:*",
+    "@signet/core": "0.82.2",
```
PR-Reviewer-Ant: This changes `@signet/core` from `workspace:*` to a fixed published version `0.82.2` even though this PR bumps the monorepo to `0.82.5`. That can make `packages/adapters/openclaw` install and build against a stale external core package instead of the current workspace sources, which is a real monorepo correctness regression rather than a style issue.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `da827175`

[Automated Review] Found two blocking integration issues and one packaging regression.
The PR does add the managed Oh My Pi connector/extension and it now matches the updated description, but the installed harness still depends on runtime configuration that this connector never provisions, and one unrelated package was switched off the workspace graph onto a stale published core version.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - I validated the current head files rather than earlier review states. `packages/connector-oh-my-pi/src/index.ts` only writes the managed extension file and does not patch any Oh My Pi config or environment. `packages/oh-my-pi-extension/src/index.ts` then derives `agentId` exclusively from `SIGNET_AGENT_ID` and the daemon endpoint from `SIGNET_DAEMON_URL`, otherwise falling back to `http://localhost:3850`. That makes the new harness install incomplete for non-default daemon networking and for scoped multi-agent use. Separately, `packages/adapters/openclaw/package.json` now pins `@signet/core` to `0.82.2` instead of the local workspace while this PR bumps the repo to `0.82.5`, which can make that package build/test against stale published APIs.

packages/oh-my-pi-extension/src/index.ts
maximhar: Addressed in `9537faf8`.

The managed Oh My Pi connector now bootstraps runtime defaults before the bundled extension executes:
- it injects `SIGNET_AGENT_ID` into the managed extension wrapper
- when no explicit agent id is configured at install time, it sets `SIGNET_AGENT_ID=default`
- added `index.test.ts` coverage to verify the managed file contains the injected agent id

That means the Oh My Pi extension no longer relies on an externally pre-set env var just to preserve agent scoping on hook traffic.

packages/oh-my-pi-extension/src/index.ts
maximhar: Addressed in `9537faf8`.

The managed extension wrapper now provisions daemon connectivity defaults at install time instead of depending on preconfigured runtime env:
- it injects `SIGNET_DAEMON_URL` into the managed extension wrapper
- the value is resolved from install-time Signet env (`SIGNET_DAEMON_URL`, otherwise `SIGNET_HOST`/`SIGNET_PORT`) and written into the managed file
- it also injects `SIGNET_PATH`, so the extension resolves against the active Signet workspace instead of assuming `~/.agents`
- added connector regression coverage for the injected daemon URL and workspace path

That gives Oh My Pi the same effective daemon endpoint the rest of the Signet install path was using, without requiring users to export the env var out of band.

packages/adapters/openclaw/package.json
maximhar: Addressed in `9537faf8`.

I restored `packages/adapters/openclaw/package.json` to use `@signet/core: workspace:*` in `devDependencies`, so the adapter stays on the local monorepo graph instead of drifting onto a stale published core package during workspace builds/tests.

bun.lock
maximhar: Follow-up addressed in `890d48cd`.

The package manifests were already fixed so `@signet/connector-oh-my-pi` no longer depends on `@signet/oh-my-pi-extension`, but the workspace lockfile still reflected the old graph. I refreshed `bun.lock`, and the connector workspace entry now only lists `@signet/connector-base` as a dependency.

`@signet/oh-my-pi-extension` still appears elsewhere in the workspace lock because it remains a first-class workspace package in this monorepo, but it is no longer pulled through the connector dependency edge the review called out.

packages/oh-my-pi-extension/package.json
```diff
@@ -0,0 +1,29 @@
...
+	"main": "dist/signet-oh-my-pi.mjs",
+	"types": "dist/index.d.ts",
+	"exports": {
+		".": "./dist/signet-oh-my-pi.mjs"
+	},
+	"files": ["dist"],
+	"scripts": {
+		"build": "bun build src/index.ts --outfile dist/signet-oh-my-pi.mjs --target node --minify --format esm --external better-sqlite3 && bun run build:types",
+		"build:types": "tsc --emitDeclarationOnly --declaration --outDir dist --noCheck",
+		"typecheck": "tsc --noEmit --noCheck",
```
PR-Reviewer-Ant: The new package's `typecheck` script is effectively a no-op for type safety: `tsc --noEmit --noCheck` skips semantic checking entirely. That means the PR's claimed validation for `@signet/oh-my-pi-extension` does not actually verify the new runtime integration for API drift or typing mistakes.

packages/connector-oh-my-pi/src/index.ts
```diff
@@ -0,0 +1,188 @@
...
+
+function readTrimmedEnv(name: string): string | undefined {
+	const value = process.env[name];
+	if (typeof value !== "string") return undefined;
+	const trimmed = value.trim();
+	return trimmed.length > 0 ? trimmed : undefined;
+}
+
+function resolveOhMyPiExtensionsDir(home = homedir()): string {
+	const configuredAgentDir = readTrimmedEnv("PI_CODING_AGENT_DIR");
```
PR-Reviewer-Ant: The managed install path for `PI_CODING_AGENT_DIR` is not durable. `install()` writes the extension into the env-selected agent directory, but that location is never persisted anywhere; later maintenance paths (`isInstalled()`, `update`, `sync`, `uninstall`, and setup detection via the core helper) recompute the directory from the *current* env or fall back to `~/.omp/agent`. If a user installed into a custom Oh My Pi agent dir and later runs `signet update` or `signet sync` without that env var exported, Signet will stop finding and maintaining the managed extension. That breaks the PR's managed-runtime contract for the non-default install path.

packages/oh-my-pi-extension/src/index.ts
PR-Reviewer-Ant: The extension treats `SIGNET_DAEMON_URL` and `SIGNET_AGENT_ID` as valid whenever they are present, even if they are empty strings. Because this code uses `readRuntimeEnv(...) ?? DAEMON_URL_DEFAULT`, an empty `SIGNET_DAEMON_URL` bypasses the fallback and produces requests against a broken base URL, and an empty `SIGNET_AGENT_ID` is forwarded as a blank scope. These are external inputs and should be trimmed/validated before use so malformed env state cannot silently break daemon connectivity or agent scoping.

packages/oh-my-pi-extension/package.json
maximhar: Fixed. I removed the no-op `--noCheck` path and made `@signet/oh-my-pi-extension` run a real `tsc --noEmit` / declaration emit pass. The root cause was that the upstream Oh My Pi package does not currently type-check cleanly as a dependency in this monorepo, so I replaced our type-only imports with a local structural model of the exact runtime surface this extension consumes. That keeps semantic checking on for this package instead of papering over it.

packages/connector-oh-my-pi/src/index.ts
maximhar: Fixed at the root. The managed Oh My Pi install now persists the resolved agent dir under Signet config (`~/.config/signet/oh-my-pi.json` via new core helper), and connector/core detection paths read that persisted value when `PI_CODING_AGENT_DIR` is absent. I also added a regression test that installs into a custom agent dir, clears the env var, and verifies `isInstalled()`/`uninstall()` still find the managed extension.

packages/oh-my-pi-extension/src/index.ts
maximhar: Fixed. The extension now trims/validates runtime env before using `SIGNET_DAEMON_URL`, `SIGNET_AGENT_ID`, and `SIGNET_PATH`, so blank strings no longer bypass the fallback path or leak an empty agent scope. I also hardened the managed bootstrap to treat blank runtime env values as missing before deciding whether to inject Signet defaults, and added regression coverage for that case.

packages/oh-my-pi-extension/src/lifecycle.ts
```diff
@@ -0,0 +1,233 @@
...
+	return { sessionId, sessionFile, project };
+}
+
+async function submitSessionEnd(client: DaemonClient, payload: SessionEndPayload): Promise<boolean> {
+	const result = await client.post(
+		"/api/hooks/session-end",
+		{
+			harness: HARNESS,
+			runtimePath: RUNTIME_PATH,
+			reason: payload.reason,
```
PR-Reviewer-Ant, line 73: The new Oh My Pi harness does not include `agentId` when posting `session-end`. This hook is the write path that extracts memories from the transcript, so omitting the agent scope here means multi-agent Oh My Pi sessions can be summarized/extracted under the default scope even though the same extension correctly threads `agentId` on `session-start`, `user-prompt-submit`, and `compaction-complete`. The repo explicitly calls agent scoping out as a guardrail for all user-data reads/writes; this new harness currently breaks that contract on the session-end mutation path.

packages/connector-oh-my-pi/src/index.ts
```diff
@@ -0,0 +1,236 @@
...
+	const value = process.env[name];
+	if (typeof value !== "string") return undefined;
+	const trimmed = value.trim();
+	return trimmed.length > 0 ? trimmed : undefined;
+}
+
+function resolveWorkspacePath(home = homedir()): string {
+	const configured = readTrimmedEnv("SIGNET_PATH");
+	if (configured) return configured;
+
```
PR-Reviewer-Ant, line 33: The managed bootstrap does not actually resolve the active Signet workspace in all supported config layouts. `resolveWorkspacePath()` only checks `SIGNET_PATH` and then hardcodes `~/.config/signet/workspace.json`, ignoring `XDG_CONFIG_HOME`. This diverges from the new core Oh My Pi config helpers in the same PR, which are XDG-aware. On systems that store Signet config under `XDG_CONFIG_HOME`, the connector will inject the wrong `SIGNET_PATH` into `signet-oh-my-pi.js`, so the extension can fall back to `~/.agents` instead of the user's real workspace despite the PR claiming it now resolves against the active workspace.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `b1e26e98`

[Automated Review] Found two correctness issues in the new Oh My Pi integration: session-end extraction is not agent-scoped, and the managed bootstrap does not fully honor the active Signet workspace config.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - Both findings are directly visible in the changed Oh My Pi files. In `packages/oh-my-pi-extension/src/lifecycle.ts`, the new harness threads `agentId` on session-start, prompt-submit, and compaction-complete, but drops it from `session-end`, which is the hook that persists transcript-backed extraction. In `packages/connector-oh-my-pi/src/index.ts`, the new `SIGNET_PATH` bootstrap path reads only `SIGNET_PATH` or `~/.config/signet/workspace.json`, while the same PR uses XDG-aware config resolution for Oh My Pi state. I did not reproduce these at runtime, so confidence is below high.

packages/cli/src/cli.ts
```diff
@@ -28,6 +28,7 @@ import { dirname, join, resolve as resolvePath } from "node:path";
 import { fileURLToPath } from "node:url";
 import { ClaudeCodeConnector } from "@signet/connector-claude-code";
 import { CodexConnector } from "@signet/connector-codex";
+import { OhMyPiConnector } from "@signet/connector-oh-my-pi";
```
PR-Reviewer-Ant, line 31: This adds a runtime import of `@signet/connector-oh-my-pi`, but `packages/cli/package.json` in the current head still does not declare that package as a dependency. Workspace builds can mask this, but the published `@signet/cli` package will try to load a module that is not installed, so Oh My Pi setup/update paths will fail for real consumers.

packages/connector-oh-my-pi/package.json
```diff
@@ -0,0 +1,32 @@
...
+	"main": "./dist/index.js",
+	"types": "./dist/index.d.ts",
+	"exports": {
+		".": {
+			"import": "./dist/index.js",
+			"types": "./dist/index.d.ts"
+		}
+	},
+	"files": ["dist"],
+	"scripts": {
```
PR-Reviewer-Ant, line 15: The new connector build now depends on `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs` already existing, but the repo's standard root build pipeline (`bun run build`) does not build `@signet/oh-my-pi-extension` before running filtered connector builds. On a clean checkout this makes the documented build path fail even though the PR claims normal package builds are covered. The build graph needs to include the extension package before this connector, or the connector needs to build/embed from source without an external prebuilt artifact.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5d24061d`

[Automated Review] Found two release-path issues in the new Oh My Pi integration.
The implementation largely matches the PR description at runtime, and previously raised scoping/XDG/spec issues appear addressed in the current head. The remaining problems are in packaging/build wiring: the CLI now imports the new connector without declaring it as a package dependency, and the connector build now requires a prebuilt extension artifact that the repository's standard `bun run build` pipeline does not produce.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The missing CLI dependency is directly visible in the current head: `packages/cli/src/cli.ts` imports `@signet/connector-oh-my-pi`, but `packages/cli/package.json` at head still lists only the existing connectors and does not include `@signet/connector-oh-my-pi`. The build-graph issue is also explicit: `packages/connector-oh-my-pi/package.json` now runs `bun scripts/embed-extension.ts`, and that script hard-fails unless `packages/oh-my-pi-extension/dist/signet-oh-my-pi.mjs` already exists, but the repository root build pipeline shown in `package.json` does not build `@signet/oh-my-pi-extension` before filtered connector builds.

packages/connector-oh-my-pi/package.json
```diff
@@ -0,0 +1,32 @@
...
```
PR-Reviewer-Ant, line 15: This build now requires `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs` to already exist, but the repo's standard root build pipeline still only filters `@signet/connector-*` and never builds `@signet/oh-my-pi-extension` first. That means a clean `bun run build` can fail even though the PR presents normal package builds as covered. The connector either needs to generate/embed from source within its own build, or the root build graph must be updated so the extension package is built before this connector.

PR-Reviewer-Ant, line 15: This package's `build` now requires `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs` to already exist, but the repository's standard root build graph still does not build `@signet/oh-my-pi-extension` before filtered connector builds. On a clean checkout, `bun run build` can therefore fail even though this PR presents the normal package build path as covered. Either wire the extension package into the root build order first, or make the connector build self-contained so it does not depend on a prebuilt sibling artifact.

deploy/docker/Dockerfile
```diff
@@ -14,7 +14,7 @@ RUN bun install --frozen-lockfile
 RUN bun run build:core
 RUN bun run build:connector-base
 RUN bun run build:opencode-plugin
-RUN bun run --filter '@signet/connector-claude-code' --filter '@signet/connector-codex' --filter '@signet/connector-openclaw' --filter '@signet/connector-opencode' build
+RUN bun run --filter '@signet/connector-claude-code' --filter '@signet/connector-codex' --filter '@signet/connector-openclaw' --filter '@signet/connector-opencode' --filter '@signet/connector-oh-my-pi' build
```
PR-Reviewer-Ant, line 18: This new build step will fail in Docker as written. `@signet/connector-oh-my-pi` now runs `scripts/embed-extension.ts`, and that script throws unless `packages/oh-my-pi-extension/dist/signet-oh-my-pi.mjs` already exists. The Dockerfile never builds `@signet/oh-my-pi-extension` before invoking the connector build, so the container image build regresses immediately.

packages/connector-oh-my-pi/src/extension-bundle.ts
PR-Reviewer-Ant, line 7: The managed extension bundle appears to hardcode `SIGNET_AGENT_ID` into the generated artifact instead of resolving the current agent at runtime. The new connector tests explicitly assert that the installed file contains `Reflect.set(..., "SIGNET_AGENT_ID", "default")` or the env value present during install. That creates an agent-scoping leak: if the user changes agent/workspace later, Oh My Pi sessions will still be attributed to the stale baked-in agent, which can write/read memory under the wrong identity.

packages/oh-my-pi-extension/src/session-state.ts
```diff
@@ -0,0 +1,194 @@
...
+	readonly agentId: string | undefined;
+	readonly reason: string;
+}
+
+function createHiddenInjectMessage(customType: string, inject: string): OmpAgentMessage {
+	return {
+		role: "custom",
+		customType,
+		display: false,
+		content: `<signet-memory source="auto-recall">\n${inject}\n</signet-memory>`,
```
PR-Reviewer-Ant: The hidden-memory wrapper is built with raw `${inject}` content. `inject` comes from daemon recall/session-context results, which can contain arbitrary user-controlled text. If that text includes `</signet-memory>` or similar markup, it can escape the wrapper and append unintended hidden instructions to the prompt. Please escape or encode injected text before wrapping it so recalled content cannot alter the prompt structure.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `449320f8`

[Automated Review by pr-reviewer] Found one prompt-integrity issue in the new Oh My Pi hidden-context injection path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issue is directly visible in packages/oh-my-pi-extension/src/session-state.ts: daemon-provided `inject` text is interpolated verbatim inside a `<signet-memory>...</signet-memory>` wrapper. Because that text originates from recalled/user-authored content, any embedded closing tag or similar markup can break the intended wrapper and change the hidden prompt structure.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `1b8e7d2b`

[Automated Review by pr-reviewer] I did not find any additional issues worth flagging in the current head; the previously raised build/spec/session-end problems appear addressed, and this looks ready for human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - I reviewed the PR description against the current changed-file patches for the new Oh My Pi connector/extension, plus the prior PR discussion to avoid re-flagging addressed items. The current head includes the spec updates, Docker/build-order fix, deferred previous-session end handling, and prompt-wrapper sanitization. I did not have the daemon hook implementation files in the review context to independently verify the exact response contract of every reused hook endpoint, so confidence is medium rather than high.

PR #389, opened 2026-03-29
fix: shim sharp on postinstall when native bindings fail

aaf2tbz:
## Summary

Native embedding provider (`nomic-embed-text-v1.5`) fails on `bun install -g signetai` because `@huggingface/transformers` eagerly imports `sharp`, whose native bindings break under bun's isolated cache layout.

## Changes

- `packages/signetai/postinstall.cjs` — postinstall script that detects broken sharp and replaces it with a falsy shim
- `packages/signetai/package.json` — added `postinstall` script + included file in `files` array
- `packages/signetai/postinstall.test.cjs` — regression test validating shim correctness
- `deploy/docker/Dockerfile` — copy postinstall.cjs before `bun install --production`

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] Other: `signetai` (published meta-package), Docker build

## PR Readiness (MANDATORY)

## Testing

- [x] `bun test packages/signetai/postinstall.test.cjs` — 3 pass, 0 fail
- [x] `bun install -g signetai` on macOS ARM — postinstall detects broken sharp and shims it
- [x] `signet daemon restart` — native embedding logs show "Ready — 768-dim embeddings"
- [x] Verified Ollama fallback still works when both native and shim fail

## AI disclosure

- [x] AI tools were used (see `Co-Authored-By` tags in commits)

## Notes

The postinstall uses `.cjs` extension because the package declares `"type": "module"` — Node treats `.js` as ESM in that context. The shim exports `null` (falsy) so transformers' `else if (sharp)` guard skips image loading cleanly.

### DIFF
--- a/packages/signetai/package.json
+++ b/packages/signetai/package.json
@@ -12,7 +12,7 @@
     "dist",
     "dashboard",
     "templates",
-    "postinstall.js",
+    "postinstall.cjs",
     "README.md",
     "LICENSE"
   ],
@@ -21,7 +21,7 @@
     "build:cli": "bun build ../cli/src/cli.ts --outfile ./dist/cli.js --target node --external better-sqlite3",
     "build:daemon": "bun run build-daemon.ts",
     "build:dashboard": "rm -rf ./dashboard && cp -r ../cli/dashboard/build ./dashboard 2>/dev/null || echo 'Dashboard not built yet'",
-    "postinstall": "node postinstall.js",
+    "postinstall": "node postinstall.cjs",
     "prepublishOnly": "bun run build"
   },
   "keywords": [
--- a/packages/signetai/package.json
+++ b/packages/signetai/package.json
@@ -12,6 +12,7 @@
     "dist",
     "dashboard",
     "templates",
+    "postinstall.js",
     "README.md",
     "LICENSE"
   ],
@@ -20,6 +21,7 @@
     "build:cli": "bun build ../cli/src/cli.ts --outfile ./dist/cli.js --target node --external better-sqlite3",
     "build:daemon": "bun run build-daemon.ts",
     "build:dashboard": "rm -rf ./dashboard && cp -r ../cli/dashboard/build ./dashboard 2>/dev/null || echo 'Dashboard not built yet'",
+    "postinstall": "node postinstall.js",
     "prepublishOnly": "bun run build"
   },
   "keywords": [
--- a/packages/signetai/postinstall.cjs
+++ b/packages/signetai/postinstall.cjs
@@ -54,9 +54,8 @@ function main() {
 	const shimContent = [
 		"// Auto-generated shim — sharp native bindings unavailable.",
 		"// Signet only uses text embeddings; image processing is not needed.",
-		"// See: https://github.com/Signet-AI/signetai/issues/TBD",
-		"module.exports = undefined;",
-		"module.exports.default = undefined;",
+		"// Exports falsy default so transformers' `else if (sharp)` is skipped.",
+		"module.exports = null;",
 		"",
 	].join("\n");
 
--- a/packages/signetai/postinstall.cjs
+++ b/packages/signetai/postinstall.cjs
@@ -1,103 +1,86 @@
 #!/usr/bin/env node
 
 /**
- * postinstall — ensure sharp's native bindings are loadable.
+ * postinstall — ensure sharp is resolvable for @huggingface/transformers.
  *
  * Problem:  @huggingface/transformers does `import sharp from 'sharp'` at
  *           the module level.  When installed globally via bun, sharp's
- *           native binary (@img/sharp-darwin-arm64) can't find the
- *           libvips dylib because bun's cache isolates each package by
- *           version and the .node binary's @rpath expects the libvips
- *           package as a sibling — which bun never creates.
+ *           native binary can't find libvips because bun's cache isolates
+ *           packages by version.  If sharp isn't installed at all (e.g.
+ *           optional dep skipped), the import also fails.
  *
- *           This causes the entire native embedding provider to fail with
- *           "Unable to get model file path or buffer", falling back to
- *           Ollama (if available) or no embeddings at all.
+ * Fix:      If sharp doesn't load, create a local shim package inside
+ *           THIS package's node_modules (not the shared cache).  Node/bun
+ *           resolution checks local node_modules first, so the shim takes
+ *           precedence without corrupting the shared cache for other
+ *           consumers.
  *
- * Fix:      Try to require('sharp').  If it throws, write a minimal shim
- *           into node_modules/sharp/lib/index.js that exports undefined
- *           so transformers' `else if (sharp)` guard skips image loading
- *           cleanly.  This is safe because signet only uses text embeddings.
+ *           Safe because signet only uses text embeddings — never images.
  */
 
-const { existsSync, mkdirSync, writeFileSync, readFileSync } = require("node:fs");
-const { dirname, join } = require("node:path");
+const { mkdirSync, writeFileSync } = require("node:fs");
+const { join } = require("node:path");
 
-function findSharpDir() {
-	try {
-		// resolve from the installed signetai package context
-		const sharpMain = require.resolve("sharp");
-		return dirname(dirname(sharpMain));
-	} catch {
-		return null;
-	}
-}
+/** The shim content — exported so tests can validate it. */
+const CJS_SHIM =
+	"// Auto-generated shim — sharp native bindings unavailable.\n" +
+	"// Signet only uses text embeddings; image processing is not needed.\n" +
+	"module.exports = null;\n";
+
+const ESM_SHIM =
+	"// Auto-generated shim — sharp native bindings unavailable.\n" +
+	"export default null;\n";
+
+const SHIM_PACKAGE_JSON = JSON.stringify(
+	{
+		name: "sharp",
+		version: "0.0.0-signet-shim",
+		main: "index.js",
+		module: "index.mjs",
+		exports: {
+			".": {
+				import: "./index.mjs",
+				require: "./index.js",
+				default: "./index.js",
+			},
+		},
+	},
+	null,
+	2,
+);
 
 function main() {
-	// First, check if sharp works fine as-is
+	// If sharp loads fine, nothing to do.
 	try {
 		require("sharp");
-		// sharp loads — no action needed
 		return;
 	} catch {
-		// sharp is broken — continue to shim it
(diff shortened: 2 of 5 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: shim sharp on postinstall when native bindings fail` (commit `b2b0b1d4`) now and I'll follow up shortly with feedback.

packages/signetai/postinstall.cjs
```diff
@@ -0,0 +1,104 @@
...
+ *
+ *           This causes the entire native embedding provider to fail with
+ *           "Unable to get model file path or buffer", falling back to
+ *           Ollama (if available) or no embeddings at all.
+ *
+ * Fix:      Try to require('sharp').  If it throws, write a minimal shim
+ *           into node_modules/sharp/lib/index.js that exports undefined
+ *           so transformers' `else if (sharp)` guard skips image loading
+ *           cleanly.  This is safe because signet only uses text embeddings.
+ */
```
PR-Reviewer-Ant, line 19: This script is executed via `node postinstall.js`, but the package is declared as `"type": "module"` in `packages/signetai/package.json`. Under Node, `.js` files in this package are treated as ESM, so every `require(...)` here will throw before the postinstall logic runs. As written, the PR does not achieve the stated fix on standard Node-based installs.

packages/signetai/postinstall.cjs
```diff
@@ -0,0 +1,104 @@
...
+	} catch {
+		// sharp is broken — continue to shim it
+	}
+
+	const sharpDir = findSharpDir();
+	if (!sharpDir) {
+		// sharp isn't installed at all — transformers will handle the
+		// missing module gracefully (import returns undefined for
+		// optional deps in some runtimes).  Nothing to do.
+		return;
```
PR-Reviewer-Ant: The generated CommonJS shim is invalid: it writes `module.exports = undefined;` and then immediately executes `module.exports.default = undefined;`. After the first statement, `module.exports` is `undefined`, so the second statement throws a `TypeError` when `sharp` is imported. That means the replacement entrypoint still crashes instead of making transformers skip the image path.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `b2b0b1d4`

[Automated Review] Found two blocking issues in the postinstall shim implementation.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly provable from the changed files. `packages/signetai/package.json` declares `"type": "module"` and invokes `node postinstall.js`, while `packages/signetai/postinstall.js` is written in CommonJS (`require(...)`), which will fail under Node's ESM rules before any shim logic runs. Separately, the generated shim source assigns `module.exports = undefined` and then immediately dereferences `module.exports.default`, which throws at module evaluation time.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: shim sharp on postinstall when native bindings fail` (commit `b2b0b1d4`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b2b0b1d4`

The PR author is `aaf2tbz`. Per the custom repository instructions, I must not post any review for PRs authored by `aaf2tbz` or `NicholaiVogel`. Output nothing and stop immediately.

packages/signetai/postinstall.cjs
```diff
@@ -0,0 +1,103 @@
...
+		"// Signet only uses text embeddings; image processing is not needed.",
+		"// Exports falsy default so transformers' `else if (sharp)` is skipped.",
+		"module.exports = null;",
+		"",
+	].join("\n");
+
+	const libDir = join(sharpDir, "lib");
+	const indexPath = join(libDir, "index.js");
+
+	try {
```
PR-Reviewer-Ant, line 66: This fix rewrites files inside the installed `sharp` package (`lib/index.js`, `lib/index.mjs`, and later `package.json`). In the Bun global-install scenario described in the PR, those files may come from Bun's shared package cache/store rather than a private copy. If so, one failed `signetai` install will poison `sharp@0.34.1` for other packages on the machine by turning it into a permanent null shim. That is a cross-install corruption bug, not just a local workaround. The shim needs to be scoped to Signet's own runtime resolution path instead of mutating the dependency package in place.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `782b5275`

[Automated Review] Found one blocking correctness issue in the sharp shimming approach.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new postinstall script explicitly overwrites files inside the installed `sharp` package (`lib/index.js`, `lib/index.mjs`, and `package.json`). The PR description itself says the triggering environment is Bun's isolated cache layout for global installs, and Bun commonly materializes packages from a shared cache/store. Given that, these in-place writes are strong evidence of cross-install package corruption risk: one failed `signetai` install can leave the cached `sharp@0.34.1` permanently shimmed for unrelated consumers until the cache is repaired. I did not run Bun's global installer here to confirm the exact cache/link semantics on this host.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `81fd4922`

The PR author is `aaf2tbz`. Per the custom repository instructions, I must not post any review for PRs authored by `aaf2tbz` or `NicholaiVogel`. Stopping immediately.

packages/signetai/postinstall.cjs
```diff
@@ -0,0 +1,103 @@
...
+	// First, check if sharp works fine as-is
+	try {
+		require("sharp");
+		// sharp loads — no action needed
+		return;
+	} catch {
+		// sharp is broken — continue to shim it
+	}
+
+	const sharpDir = findSharpDir();
```
PR-Reviewer-Ant: This early return leaves a real failure mode unfixed. The PR description says `@huggingface/transformers` eagerly imports `sharp`, so if Bun skips the optional dependency entirely instead of installing a broken copy, returning here means no shim is created and the native embedding provider will still crash on import. The fix needs to handle both `sharp is installed but broken` and `sharp is absent` cases.

packages/signetai/postinstall.test.cjs
```diff
@@ -0,0 +1,42 @@
...
+ * Regression test for the postinstall sharp shim.
+ *
+ * Validates that the generated shim content is valid JavaScript
+ * and exports a falsy value (so transformers' `else if (sharp)`
+ * guard skips image loading).
+ */
+
+const { describe, it, expect } = require("bun:test");
+
+// The shim content the postinstall writes into sharp's entry point.
```
PR-Reviewer-Ant: This regression test does not validate the actual implementation in `postinstall.cjs`; it redefines the expected shim contents locally. If the postinstall script drifts again, these tests can still pass while the shipped script is broken. Given the repo's requirement for durable regression guards on bug fixes, this should exercise the real script output or import a shared source of truth.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `81fd4922`

[Automated Review] Found one blocking correctness gap and one test coverage gap in the sharp postinstall shim.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The blocking issue is directly visible in `packages/signetai/postinstall.cjs`: when `sharp` cannot be resolved at all, the script returns early even though the PR description says `@huggingface/transformers` eagerly imports `sharp` at module scope. That means the fix still leaves an unhandled failure mode if Bun skips the optional dependency instead of installing a broken one. The test gap is also directly visible because `packages/signetai/postinstall.test.cjs` hard-codes shim strings instead of exercising the real `postinstall.cjs` output. I did not reproduce Bun's exact optional-dependency behavior on this host.

NicholaiVogel:
Postinstall scripts are blocked automatically by bun and npm and should not be relied on for anything important. I recommend relying on the established patterns in the Daemon and CLI for important download behavior

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `cacbc31e`

Per the custom repository instructions, the PR author is `aaf2tbz`. I must not post any review for PRs authored by `aaf2tbz` or `NicholaiVogel`. Stopping immediately.

packages/signetai/package.json
```diff
@@ -20,6 +21,7 @@
     "build:cli": "bun build ../cli/src/cli.ts --outfile ./dist/cli.js --target node --external better-sqlite3",
     "build:daemon": "bun run build-daemon.ts",
     "build:dashboard": "rm -rf ./dashboard && cp -r ../cli/dashboard/build ./dashboard 2>/dev/null || echo 'Dashboard not built yet'",
+    "postinstall": "node postinstall.cjs",
```
PR-Reviewer-Ant, line 24: This fix hinges on `postinstall`, but the PR's target failure mode is `bun install -g signetai`, and the existing discussion notes that Bun/npm do not reliably run postinstall hooks there. If the hook is skipped, `@huggingface/transformers` still encounters the broken `sharp` import and the regression remains. As written, the implementation does not reliably achieve the PR description's stated goal.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `cacbc31e`

[Automated Review] The PR still has one blocking correctness issue: the claimed fix depends on a postinstall hook that the target install path does not reliably execute.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The PR description says this fixes `bun install -g signetai`, but the implementation in `packages/signetai/package.json` relies entirely on `postinstall`. The PR conversation already notes that Bun/npm block postinstall scripts for this path, so the shim logic in `packages/signetai/postinstall.cjs` may never run in the failing scenario the PR claims to address. That makes the divergence between the stated goal and the shipped mechanism directly visible from this diff and the attached review context.

PR #393, opened 2026-03-29
feat(dashboard): ontology constellation view

NicholaiVogel:
## Summary

Interactive force-directed knowledge graph visualization for the dashboard Constellation tab. Replaces `EmbeddingsTab` with `OntologyDashboard` — a d3-force + Canvas 2D graph that renders entities, aspects, and attributes with progressive disclosure, plus a filter panel, drill-down inspector, and debounced full-text search.

**Breaking change (intentional):** `EmbeddingsTab` and the `onopenglobalsimilar` callback are no longer wired in the Constellation section. The embeddings UMAP view is preserved as a floating panel (`UmapPanel`) within the new dashboard. The global-similar workflow and embedding diagnostics that previously lived in `EmbeddingsTab` are out of scope for this PR and should be tracked separately if they need a new home.

## Changes

- `packages/cli/dashboard/src/lib/components/ontology/` (new directory, 7 files)
  - `ConstellationGraph.svelte` -- d3-force simulation, Canvas 2D renderer, pan/zoom/drag, idle RAF loop
  - `InspectorPanel.svelte` -- navigable drill-down for entities -> aspects -> attributes with metadata grids
  - `OntologyDashboard.svelte` -- layout shell with debounced search bar
  - `SchemaPanel.svelte` -- filter chips for node/edge kinds, quick view presets, live counts
  - `UmapPanel.svelte` -- floating embedding-space panel (points colored by memory type)
  - `ontology-data.ts` -- graph data transforms from constellation API
  - `ontology-state.svelte.ts` -- shared reactive state, API loaders (all with agentId scoping + stale-response guards), selection/search helpers
- `UnifiedMemoryTab.svelte` -- wire OntologyDashboard (with agentId) into Constellation tab; removes EmbeddingsTab
- `api.ts` -- use relative URLs in dev (vite proxy); add agentId to ProjectionQueryOptions
- `vite.config.ts` -- dev server proxy for /api, /health, /memory

## Type

- [x] `feat` -- new user-facing feature (bumps minor)

## Packages affected

### Entity overview (500 nodes, force-directed)
ontology-overview

### Search filtering (debounced)
ontology-search

### With Aspects preset (entities + aspect nodes)
ontology-aspects

## PR Readiness (MANDATORY)

- [x] Security checks applied to admin/mutation endpoints -- N/A, read-only dashboard
- [x] Docs updated for API/spec/status changes -- no API changes
- [x] Regression tests added for each bug fix -- new feature, not a fix

## Testing

- [x] `bun run lint` passes

## AI disclosure

## Notes

All loaders (`loadGraph`, `loadEntityDetail`, `loadAspectDetail`, `loadProjection`, `loadTableStats`) carry agentId through to the daemon API and use generation counters or selection guards to drop stale responses from rapid agent switches. The Constellation canvas uses an idle RAF loop that stops when the sim settles and restarts on pan/zoom/drag/selection.

### DIFF
--- a/packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
+++ b/packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
@@ -158,9 +158,10 @@ let searchGeneration = 0;
 export function searchGraph(query: string, delay = 250): void {
 	ontology.filterQuery = query;
 
-	if (searchTimer) clearTimeout(searchTimer);
+	if (searchTimer) { clearTimeout(searchTimer); searchTimer = null; }
 
 	if (!query.trim()) {
+		searchGeneration++;
 		ontology.searchMatchIds = null;
 		ontology.searchEpoch++;
 		ontology.searching = false;
@@ -342,7 +343,7 @@ export async function loadTableStats(table: string, agentId = "default"): Promis
 				ontology.tableStats = { table, rows: -1 };
 		}
 	} finally {
-		ontology.loadingTable = false;
+		if (gen === tableStatsGeneration) ontology.loadingTable = false;
 	}
 }
 


packages/cli/dashboard/src/lib/components/ontology/ConstellationGraph.svelte
```diff
@@ -0,0 +1,575 @@
...
+					.id((d) => d.id)
+					.distance(linkDist)
+					.strength(0.15),
+			)
+			.force("charge", forceManyBody<SimNode>().strength(charge))
+			.force("x", forceX<SimNode>(0).strength(0.06))
+			.force("y", forceY<SimNode>(0).strength(0.06))
+			.force("collide", forceCollide<SimNode>((d) => d.radius + 2))
+			.alphaDecay(0.02)
+			.on("tick", () => {});
```
PR-Reviewer-Ant, line 161: The graph rebuild key only includes `searchMatchIds.size`, not the actual matched IDs. If two different queries both produce the same number of expanded matches, `buildSim()` will not run and the canvas will keep showing the previous result set. That breaks the PR's claim that debounced search updates the constellation coherently.

packages/cli/dashboard/src/lib/components/ontology/ConstellationGraph.svelte
```diff
@@ -0,0 +1,575 @@
...
+
+				// Text shadow for readability
+				if (!dimmed) {
+					ctx.fillStyle = "rgba(0, 0, 0, 0.6)";
+					ctx.fillText(label, nx + 1, ly + 1);
+				}
+
+				ctx.fillStyle = isSelected
+					? "#c8ff00"
+					: isRelated
```
PR-Reviewer-Ant, line 335: Hit-testing is using `node.radius / camZoom + 4` in world coordinates. `node.radius` is already a world-space radius, so dividing it by zoom makes clicks miss enlarged nodes when zoomed in and creates false positives when zoomed out. Only the screen-space padding term should be converted by zoom.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `49668414`

[Automated Review] I found one blocking logic bug in the new constellation search/filter flow and one interaction bug in graph hit-testing. The PR description also does not include screenshots or visual previews for the UI changes, so this needs UI evidence before human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The stale-search issue is directly provable from the new `ConstellationGraph.svelte` effect key: it only keys on `searchMatchIds.size`, so different search result sets with the same cardinality will not rebuild the graph. The hit-testing issue is also directly visible in the new `findNodeAt` math, where the world-space node radius is divided by zoom instead of only converting the screen-space padding term. I did not run the dashboard, so visual severity is inferred from the code path.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(dashboard): ontology constellation view` (commit `7576d7e0`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,374 @@
...
+		}
+		const { nodes, edges } = buildGraphFromConstellation(data);
+		ontology.graphNodes = nodes;
+		ontology.graphEdges = edges;
+		ontology.entities = data.entities;
+	} catch (err) {
+		ontology.error = err instanceof Error ? err.message : "Unknown error";
+		ontology.graphNodes = [];
+		ontology.graphEdges = [];
+		ontology.entities = [];
```
PR-Reviewer-Ant, line 147: The new search only matches `n.label`, but attribute labels are pre-truncated when the graph is built. That means the advertised debounced full-text search cannot find terms that appear later in long attribute/constraint content. For example, any query that only appears after character 30 in an attribute will never match. This diverges from the PR description and makes search results incomplete.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,374 @@
...
+		ontology.searching = false;
+		return;
+	}
+
+	ontology.searching = true;
+
+	searchTimer = setTimeout(() => {
+		const gen = ++searchGeneration;
+		const lower = query.toLowerCase();
+		const nodes = ontology.graphNodes;
```
PR-Reviewer-Ant, line 176: These inspector loaders race each other. If the user selects entity/aspect A and then quickly selects B, the slower response from A can resolve last and overwrite `ontology.detail`, `detailAspects`, `detailDependencies`, or `aspectAttrs` for the newer selection. The result is stale inspector content for the wrong node. This needs a generation token or a 'selection still matches' guard before committing async results.

NicholaiVogel:
Fixed all four issues in 9e099551:

**Search rebuild key** — replaced `s:${searchMatchIds.size}` in the filter effect key with `searchEpoch`, an integer that increments on every `searchMatchIds` write. Two searches with the same result count now correctly trigger distinct rebuilds.

**Hit-test radius** — changed `node.radius / camZoom + 4` to `node.radius + 4 / camZoom`. The node radius is world-space so it stays as-is; only the 4px screen-space padding needs the zoom conversion.

**Search truncation** — added `searchText?: string` field to `OntologyNode`. Attribute nodes now carry the full `attr.content` in `searchText` while still showing the truncated 30-char label. The search loop checks `n.searchText ?? n.label`, so terms past character 30 now match.

**Async race** — added selection guards after each `await` in both `loadEntityDetail` and `loadAspectDetail`. Checks `ontology.selected?.id !== entityId/aspectId` before writing any state; returns early if the selection changed while the request was in-flight.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(dashboard): ontology constellation view` (commit `9e099551`) now and I'll follow up shortly with feedback.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,385 @@
...
+	ontology.loadingDetail = true;
+	try {
+		const [detail, aspects, deps] = await Promise.all([
+			getKnowledgeEntity(entityId, agentId),
+			getKnowledgeAspects(entityId, agentId),
+			getKnowledgeDependencies(entityId, "both", agentId),
+		]);
+		// Guard: bail if selection changed while we were loading
+		if (ontology.selected?.id !== entityId) return;
+		ontology.detail = detail;
```
PR-Reviewer-Ant, line 227: Blocking: direct clicks on attribute nodes never populate full attribute metadata. `InspectorPanel` now calls `loadAspectDetail(node.parentId)` when an attribute is selected, but this loader drops the result unless `ontology.selected?.id === aspectId`. For an attribute selection, `selected.id` is the attribute id, not the parent aspect id, so the request always returns early and `ontology.aspectAttrs` is never updated. The fallback object in the inspector then shows partial/stale data instead of live attribute details, which diverges from the PR's claimed drill-down inspector behavior.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,385 @@
...
+
+export async function loadAspectDetail(aspectId: string, agentId = "default"): Promise<void> {
+	const node = ontology.graphNodes.find((n) => n.id === aspectId && n.kind === "aspect");
+	if (!node?.parentId) return;
+
+	ontology.loadingAspect = true;
+	try {
+		const attrs = await getKnowledgeAttributes(node.parentId, aspectId, { agentId });
+		// Guard: bail if selection changed while loading
+		if (ontology.selected?.id !== aspectId) return;
```
PR-Reviewer-Ant: Warning: `loadTableStats()` has no stale-response guard. If the user clicks between schema tables quickly, an earlier slower request can resolve after a later one and overwrite `ontology.tableStats` with the wrong table's counts while `ontology.schemaTable` already points somewhere else. This produces inconsistent inspector state and is the same class of async race the repo conventions call out for dashboard/fallback flows.

NicholaiVogel:
Fixed both in faaac0e2:

**Attribute drill-down guard** — `loadAspectDetail` now takes an optional `triggeredBy` param (defaults to `aspectId` for direct aspect clicks). When called from an attribute selection, InspectorPanel passes the attribute's own ID as `triggeredBy`. The guard then checks `selected.id === triggeredBy`, which is the attribute ID — so it passes correctly instead of always bailing. The aspect loads and `aspectAttrs` gets written.

**Table stats race** — added `ontology.schemaTable !== table` checks after each `await` inside `loadTableStats`. A slow response for table A now drops its result if `schemaTable` has already moved to table B.

packages/cli/dashboard/src/lib/components/tabs/UnifiedMemoryTab.svelte
```diff
@@ -71,16 +72,14 @@
 			</Button>
 		</div>
 	</div>
-	<div class="unified-body">
+	<div class="unified-body" class:unified-body-flush={section === "constellation"}>
 		{#if section === "cortex"}
 			<div class="unified-main">
 				<MemoryTab {memories} {agentId} embedded={true} />
 			</div>
```
PR-Reviewer-Ant, line 79: Blocking: this swaps the old scoped embeddings view for `OntologyDashboard`, but it no longer threads the selected `agentId` through. The new ontology loaders (`loadGraph`, `loadEntityDetail`, `loadAspectDetail`) all default to `"default"`, so opening the Constellation tab on any non-default agent will query and display the default agent's graph/details instead of the active agent. That violates the repo's scoping guardrail and can leak the wrong user's data into the UI.

packages/cli/dashboard/src/lib/components/ontology/ConstellationGraph.svelte
```diff
@@ -0,0 +1,572 @@
...
+			.force("collide", forceCollide<SimNode>((d) => d.radius + 2))
+			.alphaDecay(0.02)
+			.on("tick", () => {});
+
+		startRenderLoop();
+	}
+
+	// Watch for filter changes and rebuild sim
+	$effect(() => {
+		const key = [...ontology.visibleNodeKinds].sort().join(",")
```
PR-Reviewer-Ant, line 168: Warning: `startRenderLoop()` creates a permanent `requestAnimationFrame` repaint loop for the canvas and never stops it when the simulation has settled. Because `draw()` repaints the full graph every frame, simply leaving this tab open will keep consuming CPU/GPU even while idle. Given the repo's stated performance priority, this should render on simulation ticks / interaction or stop once `sim.alpha()` is low.

NicholaiVogel:
Fixed both in ecd12cd8:

**agentId scoping** — Added `agentId?: string` prop (default `"default"`) to `OntologyDashboard`, `ConstellationGraph`, and `InspectorPanel`. `UnifiedMemoryTab` now passes its `agentId` prop down through `OntologyDashboard` to both children. `loadGraph(agentId)`, `loadEntityDetail(id, agentId)`, and `loadAspectDetail(aspectId, triggeredBy, agentId)` all receive the active agent — no more hardcoded "default" leak.

**Idle render loop** — `startRenderLoop` now stops the RAF when `sim.alpha() < 0.01` and neither `isDragging` nor `isPanning` is active. Added `wakeRenderLoop()` (no-ops if already running) called from `onPointerMove` during drag/pan, `onWheel` on zoom, and a `$effect` tracking `ontology.selected`/`ontology.hovered` so highlight changes still repaint. Canvas is idle between user interactions once the sim settles.

packages/cli/dashboard/src/lib/components/ontology/ConstellationGraph.svelte
```diff
@@ -0,0 +1,598 @@
...
+			wakeRenderLoop();
+			return;
+		}
+
+		// Hover detection
+		const hit = findNodeAt(pos.x, pos.y);
+		if (hit) {
+			hoverNode(hit.id, hit.kind);
+			if (canvas) canvas.style.cursor = "pointer";
+		} else {
```
PR-Reviewer-Ant, line 429: This only loads the graph on initial mount. Because the component now accepts `agentId` as a prop, switching the active agent while the Constellation tab remains mounted leaves the previous agent's graph on screen. A reactive effect keyed on `agentId` is needed here, and the inspector side needs the same treatment so scoped data is reloaded when the prop changes.

packages/cli/dashboard/src/lib/components/ontology/InspectorPanel.svelte
```diff
@@ -0,0 +1,933 @@
...
+	import {
+		ontology,
+		loadEntityDetail,
+		loadAspectDetail,
+		selectNode,
+	} from "./ontology-state.svelte";
+	import { entityNameFromGraph, NODE_COLORS } from "./ontology-data";
+
+	interface Props {
+		agentId?: string;
```
PR-Reviewer-Ant, line 19: These loaders react to `ontology.selected` changes, but not to `agentId` changes. If the user has a node selected and switches agents without unmounting this panel, `loadEntityDetail`/`loadAspectDetail` will not rerun, so the inspector can keep showing the old agent's details under the new scope.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `ecd12cd8`

[Automated Review] Found one remaining agent-scoping regression in the new ontology dashboard.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In the final head, `UnifiedMemoryTab` now passes `agentId` through, but `ConstellationGraph` only calls `loadGraph(agentId)` inside `onMount`, and `InspectorPanel` only reloads detail when `ontology.selected?.id` changes. If the active agent changes while the Constellation tab stays mounted, the graph and inspector keep showing the previous agent's data until some other interaction forces a reload. That is directly visible from the changed component logic and violates the repo's scoping guardrail.

NicholaiVogel:
Fixed both in f772d791:

**Graph reload on agent switch** — moved `loadGraph` out of `onMount` and into a `$effect` keyed on `agentId`. Svelte tracks the `agentId` read inside the effect, so any prop change re-runs it and fetches the correct agent's graph. `onMount` now only handles the cleanup (cancel RAF, stop sim).

**Inspector reload on agent switch** — replaced `lastEntityId`/`lastAspectId` string dedup keys with compound `"id@agentId"` keys. When `agentId` changes, the key no longer matches even for the same selected node, so both `loadEntityDetail` and `loadAspectDetail` refire with the new scope.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,392 @@
...
+	error: null as string | null,
+
+	/** Inspector detail (loaded on entity selection) */
+	detail: null as KnowledgeEntityDetail | null,
+	detailAspects: [] as KnowledgeAspectWithCounts[],
+	detailAttributes: new Map() as Map<string, KnowledgeAttribute[]>,
+	detailDependencies: [] as KnowledgeDependencyEdge[],
+	loadingDetail: false,
+
+	/** Inspector: aspect attributes (loaded on aspect/attribute selection) */
```
PR-Reviewer-Ant, line 91: Blocking: `loadGraph(agentId)` has no stale-response guard. `ConstellationGraph.svelte` now reissues this loader reactively when `agentId` changes, so switching agents quickly can leave two overlay requests in flight. If the older request resolves last, it overwrites `ontology.graphNodes`/`graphEdges`/`entities` with the previous agent's data, and the graph rebuild in the caller will render the wrong scoped graph. This is the same agent-scoping leak class the repo guardrails call out; add a generation token or verify the current agent before committing async results.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f772d791`

[Automated Review] Found one remaining blocking scoping race in the constellation graph loader.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The final head fixes the earlier agentId propagation issues, but `ConstellationGraph.svelte` now reloads on `agentId` changes via `$effect`, while `loadGraph()` in `ontology-state.svelte.ts` still writes `graphNodes`, `graphEdges`, and `entities` unconditionally after `await getConstellationOverlay(agentId)`. If the active agent changes twice quickly, the slower earlier request can resolve last and overwrite the newer agent's graph, reintroducing cross-agent leakage through an async race.

NicholaiVogel:
Fixed in 656c495b: added a module-level `graphGeneration` counter to `loadGraph`. Each call increments it and captures the value; after the `await`, any response whose generation doesn't match the current counter is dropped. Covers both the success path and the error path. The `finally` block also only clears `loading` for the winning request.

packages/cli/dashboard/src/lib/components/ontology/UmapPanel.svelte
```diff
@@ -0,0 +1,213 @@
...
+		const dy = maxY - minY || 1;
+		return { minX, maxX: minX + dx, minY, maxY: minY + dy };
+	});
+
+	const SVG_W = 260;
+	const SVG_H = 260;
+	const PAD = 12;
+
+	function scaleX(x: number): number {
+		return PAD + ((x - bounds.minX) / (bounds.maxX - bounds.minX)) * (SVG_W - PAD * 2);
```
PR-Reviewer-Ant, line 69: The new floating UMAP panel is no longer agent-scoped. This component never receives `agentId`, calls `loadProjection()` with no scope, and `getProjection()` does not thread an `agent_id` query param. The old replacement path passed `agentId` into `EmbeddingsTab`, so on a non-default agent this panel can now show embeddings from the wrong scope. In this repo that is a real data-leak/regression, not just a UX mismatch.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,397 @@
...
```
PR-Reviewer-Ant, line 91: `loadGraph(agentId)` starts a scoped reload but does not clear the previous `graphNodes`/`graphEdges`/`entities` first. After an agent switch, the Constellation tab keeps rendering the old agent's graph until the new overlay request resolves. Because this tab is supposed to be agent-scoped, that transient reuse exposes stale cross-agent data under the new scope. The loader should blank or otherwise invalidate the prior graph immediately when `agentId` changes.

NicholaiVogel:
Fixed both in 0c144ae1:

**UMAP agent scoping** — added `agentId?: string` to `ProjectionQueryOptions` and `getProjection`, serialized as `agent_id` query param. `loadProjection` now accepts and threads `agentId`. `UmapPanel` gets an `agentId` prop (passed from `OntologyDashboard`), replaced `onMount` with `$effect` so it reloads whenever the prop changes.

**Stale graph on agent switch** — `loadGraph` now clears `graphNodes`, `graphEdges`, and `entities` immediately after bumping the generation counter, before the `await`. The canvas goes blank during the fetch rather than showing cross-agent data.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,401 @@
...
+			for (let i = 0; i < aspects.length; i++) {
+				attrMap.set(aspects[i].aspect.id, results[i]);
+			}
+		}
+		ontology.detailAttributes = attrMap;
+	} finally {
+		ontology.loadingDetail = false;
+	}
+}
+
```
PR-Reviewer-Ant, line 249: Blocking: `loadProjection()` has the same stale-response problem that `loadGraph()` already had to fix. `UmapPanel.svelte` now calls this loader from a reactive effect keyed on `agentId`, but after `await getProjection(2, { limit: 500, agentId })` the result is committed to shared state with no generation/selection guard. If the user switches agents quickly, an older slower response can overwrite `ontology.projection` for the new scope and show the wrong agent's embedding projection. This is a real cross-agent data leak/regression; add a generation token or verify the current agent before writing the result.

NicholaiVogel:
Fixed in a1ef402a: added `projectionGeneration` counter to `loadProjection` — same pattern as `loadGraph`. Each call bumps and captures the counter; after the await, stale responses are dropped and only the winning call writes `ontology.projection` or clears `loadingProjection`.

packages/cli/dashboard/src/lib/components/tabs/UnifiedMemoryTab.svelte
```diff
@@ -71,16 +72,14 @@
 			</Button>
 		</div>
 	</div>
-	<div class="unified-body">
+	<div class="unified-body" class:unified-body-flush={section === "constellation"}>
```
PR-Reviewer-Ant, line 75: This change swaps out `EmbeddingsTab` entirely rather than just replacing an embeddings table view. That removes the existing global-similar workflow (`onopenglobalsimilar` is no longer used here) along with the embedding diagnostics/repair and advanced projection controls that previously lived under the same Constellation section. The PR description does not call out this user-facing regression, so either the replacement scope needs to be narrowed or the breaking change needs to be made explicit.

packages/cli/dashboard/src/lib/components/ontology/UmapPanel.svelte
```diff
@@ -0,0 +1,218 @@
...
+		"#ef4444", // red
+		"#06b6d4", // cyan
+		"#ec4899", // pink
+	];
+
+	const DEFAULT_COLOR = "#6b7280";
+
+	let collapsed = $state(false);
+
+	const nodes = $derived(ontology.projection?.nodes ?? []);
```
PR-Reviewer-Ant, line 25: The panel is presented as an embedding-cluster view, but `colorFor` is not computing or rendering clusters at all; it only maps points by `node.type`. That means the visualization can imply cluster boundaries that do not exist and does not match the PR description's claim of a UMAP cluster panel.

NicholaiVogel:
Addressed both observations in 2037fc5e + PR description update:

**UMAP title** — renamed "EMBEDDING CLUSTERS" → "EMBEDDING SPACE" (and the toggle button tooltip to match). The panel colors points by memory type, not by computed cluster assignments, so the old title implied boundaries that don't exist.

**EmbeddingsTab removal** — updated the PR description to explicitly call this out as an intentional breaking change. The `onopenglobalsimilar` callback and embedding diagnostics from `EmbeddingsTab` are out of scope for this PR; the UMAP view is preserved as the floating panel. If the global-similar workflow needs a new home, that should be tracked separately.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,405 @@
...
+		ontology.searching = false;
+		return;
+	}
+
+	ontology.searching = true;
+
+	searchTimer = setTimeout(() => {
+		const gen = ++searchGeneration;
+		const lower = query.toLowerCase();
+		const nodes = ontology.graphNodes;
```
PR-Reviewer-Ant, line 176: Blocking: the inspector loaders still are not guarded against cross-agent races when the selected node id stays the same across an agent switch. `loadEntityDetail(entityId, agentId)` only checks `ontology.selected?.id !== entityId` after the awaits, so an older response for agent A can still overwrite `detail`, `detailAspects`, `detailDependencies`, and `detailAttributes` after the UI has already switched to agent B if the selection id is unchanged. This is the same scoped-data leak class you already fixed for `loadGraph`/`loadProjection`; the guard needs to include agent scope or use a generation token per inspector request.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,405 @@
...
+	ontology.loadingDetail = true;
+	try {
+		const [detail, aspects, deps] = await Promise.all([
+			getKnowledgeEntity(entityId, agentId),
+			getKnowledgeAspects(entityId, agentId),
+			getKnowledgeDependencies(entityId, "both", agentId),
+		]);
+		// Guard: bail if selection changed while we were loading
+		if (ontology.selected?.id !== entityId) return;
+		ontology.detail = detail;
```
PR-Reviewer-Ant, line 227: The same stale-response problem exists in `loadAspectDetail`. Its post-await guard only checks `ontology.selected?.id === triggeredBy`, not which `agentId` the request belongs to. If the user switches agents while the same aspect/attribute remains selected, a slower response from the previous agent can still win and populate `ontology.aspectAttrs` for the wrong scope.

NicholaiVogel:
Fixed in d2b00acd: replaced the selection-ID guards in both `loadEntityDetail` and `loadAspectDetail` with generation counters (`detailGeneration`, `aspectGeneration`) — same pattern as `loadGraph`/`loadProjection`.

The old guards checked `ontology.selected?.id === entityId/triggeredBy`, which passes when the same node stays selected across an agent switch. Generation counters are agent-agnostic: any response that arrives after a newer call has started is dropped regardless of whether the node ID, agentId, or both changed. The `finally` block also only clears `loadingDetail`/`loadingAspect` for the winning generation.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,408 @@
...
+		if (aspects.length > 0) {
+			const results = await Promise.all(
+				aspects.map((a) =>
+					getKnowledgeAttributes(entityId, a.aspect.id, { agentId }),
+				),
+			);
+			if (gen !== detailGeneration) return;
+			for (let i = 0; i < aspects.length; i++) {
+				attrMap.set(aspects[i].aspect.id, results[i]);
+			}
```
PR-Reviewer-Ant, line 242: Blocking: `loadTableStats()` is not agent-scoped even though the PR description explicitly says all loaders, including `loadTableStats`, carry `agentId` through to the daemon API. This function has no `agentId` parameter and calls `getKnowledgeStats()`, `getMemories()`, and `getEmbeddings()` without any scope, so the schema inspector can show cross-agent counts while the rest of the dashboard is filtered to a specific agent. That is both a description/implementation mismatch and a scoped-data leak against the repo's agent scoping rules.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,408 @@
...
+	const gen = ++aspectGeneration;
+	ontology.loadingAspect = true;
+	try {
+		const attrs = await getKnowledgeAttributes(node.parentId, aspectId, { agentId });
+		if (gen !== aspectGeneration) return;
+		ontology.aspectAttrs = attrs;
+	} finally {
+		if (gen === aspectGeneration) ontology.loadingAspect = false;
+	}
+}
```
PR-Reviewer-Ant, line 272: The `memory_entity_mentions` branch always reports `rows: 0` instead of querying or omitting the count. Selecting that table will therefore show a false live count even when the table has data, which undermines the new inspector's schema stats.

NicholaiVogel:
Fixed both in ba1cf51d:

**loadTableStats agent scoping** — added `agentId` param to `loadTableStats`, `selectSchemaTable`, and `SchemaPanel` (prop threaded from `OntologyDashboard`). Updated `getKnowledgeStats`, `getMemories`, and `getEmbeddings` to accept an optional `agentId` and serialize it as `agent_id` query param. All three branches of the stats switch now pass the active agent through.

**memory_entity_mentions false zero** — changed `rows: 0` to `rows: -1` (unknown). The memories endpoint returns `stats.total` for the memories table but has no mention count field; `-1` is used consistently elsewhere in the codebase for tables where a direct row count isn't available, rather than reporting a false zero.

packages/cli/dashboard/src/lib/components/ontology/ConstellationGraph.svelte
```diff
@@ -0,0 +1,603 @@
...
+
+		camZoom = Math.max(0.15, Math.min(5, camZoom * factor));
+
+		camX = wx - (pos.x - width / 2) / camZoom;
+		camY = wy - (pos.y - height / 2) / camZoom;
+		wakeRenderLoop();
+	}
+
+	// Reload graph whenever agentId changes (not just on mount)
+	$effect(() => {
```
PR-Reviewer-Ant, line 466: The graph reload path for `agentId` changes never clears the existing selection/hover state before rebuilding. Because `draw()` still treats `ontology.selected` as authoritative, switching from agent A to agent B can leave the new graph rendered against a stale selection from the previous agent, which dims unrelated nodes and can trigger detail loads for an ID that does not belong to the new agent. If `loadGraph()` is not resetting selection internally, this breaks the PR's claim that rapid agent switches are guarded cleanly.

packages/cli/dashboard/src/lib/components/ontology/InspectorPanel.svelte
```diff
@@ -0,0 +1,940 @@
...
+	const selectedAttr = $derived.by((): KnowledgeAttribute | undefined => {
+		if (!attrNode) return undefined;
+		const found = ontology.aspectAttrs.find((a) => a.id === attrNode.id);
+		if (found) return found;
+		// Fallback: build partial from graph node data
+		const data = attrNode.data as Record<string, unknown> | null;
+		return {
+			id: attrNode.id,
+			aspectId: attrNode.parentId ?? "",
+			agentId: "default",
```
PR-Reviewer-Ant, line 133: This fallback object hardcodes `agentId: "default"` even though the rest of the dashboard was explicitly updated for per-agent scoping. On a non-default agent, selecting an attribute before `loadAspectDetail()` finishes will synthesize metadata for the wrong tenant, and any future logic that reads `selectedAttr.agentId` will silently target the default scope instead of the active agent.

NicholaiVogel:
Fixed both in e7e22f22:

**getKnowledgeAttributes agent_id fallback** — removed the unconditional `params.set("agent_id", filters.agentId ?? "default")`. Changed to `if (filters.agentId) params.set("agent_id", filters.agentId)` so callers that don't pass an agentId don't accidentally scope all attribute fetches to `"default"`.

**Stale selection on agent switch** — added clearing of `selected`, `hovered`, `relatedIds`, and `searchMatchIds` in `loadGraph` at the same point where `graphNodes/graphEdges/entities` are cleared. This prevents `draw()` from honoring a selection or related-set that belongs to a different agent's node IDs after the canvas has been emptied for the new scope.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,414 @@
...
+
+	/** Selected schema table name in Zone A */
+	schemaTable: "entities" as string,
+
+	/** Filter: search query (FTS) */
+	filterQuery: "" as string,
+
+	/** IDs matching the current search (null = no search active) */
+	searchMatchIds: null as Set<string> | null,
+	/** Increments on every searchMatchIds write so filter effects see distinct keys */
```
PR-Reviewer-Ant, line 60: Blocking: agent switches still leave stale schema stats from the previous agent. `loadGraph()` clears the graph/selection state, but it does not clear or refresh `ontology.tableStats`, and nothing in `SchemaPanel` re-runs `loadTableStats()` when `agentId` changes. If the user has a table selected and then switches agents, the inspector can keep showing the old agent's counts under the new scope. That is the same cross-agent data leak class this PR claims to guard against.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,414 @@
...
+	searchEpoch: 0,
+	searching: false,
+
+	/** Visible edge/node kinds on graph (driven by schema table selection) */
+	visibleEdgeKinds: new Set(["dependency"]) as Set<string>,
+	visibleNodeKinds: new Set(["entity"]) as Set<OntologyNodeKind>,
```
PR-Reviewer-Ant, line 66: Warning: `loadGraph()` clears `searchMatchIds` on every agent reload but leaves `filterQuery` untouched. After switching agents with an active search, the search box still shows the old query while the graph is rebuilt with no search filter applied. That makes the dashboard state internally inconsistent and breaks the PR's claim of coherent debounced search behavior.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `e7e22f22`

[Automated Review] Found one remaining cross-agent scoping leak in schema stats and one search-state bug on agent switch.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - I checked the final head against the PR description and prior review thread. The earlier graph/projection/inspector race findings were already addressed in follow-up commits and are not repeated here. The remaining issues are visible from the current changed code: `loadGraph()` clears graph/search state on agent changes but does not reload or clear `tableStats`, and `SchemaPanel` has no agent-change effect to refresh stats for the current table. Separately, `loadGraph()` clears `searchMatchIds` while leaving `filterQuery` intact, so the UI can display an active search string while the graph shows unfiltered results after an agent switch.

NicholaiVogel:
Fixed both in 10821f87:

**filterQuery/searchMatchIds mismatch** — added `ontology.filterQuery = ""` to the clear block in `loadGraph`, alongside the existing `searchMatchIds = null`. Both reset together so the search bar can't show a stale query while the graph is unfiltered.

**tableStats stale on agent switch** — added `ontology.tableStats = null` to the clear block so the schema panel goes blank immediately on switch rather than showing the previous agent's row counts. After the new graph data arrives, `loadTableStats(ontology.schemaTable, agentId)` is called to populate the stats for the current table under the new agent.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,418 @@
...
+// --- Graph data ---
+
+let graphGeneration = 0;
+
+export async function loadGraph(agentId = "default"): Promise<void> {
+	const gen = ++graphGeneration;
+	ontology.loading = true;
+	ontology.error = null;
+	// Clear immediately so the canvas doesn't show stale cross-agent data
+	ontology.graphNodes = [];
```
PR-Reviewer-Ant, line 113: Blocking: the debounced search callback survives agent switches. `loadGraph()` now clears `filterQuery` and `searchMatchIds`, but it never cancels `searchTimer` or bumps `searchGeneration`. If a user types a query and then switches agents before the 250ms debounce fires, the stale timer can still run afterward and write a new `searchMatchIds` set for the old query. That re-filters the new agent's graph while the search box is blank, which breaks the PR's claim that agent switches are guarded cleanly.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `10821f87`

[Automated Review] Found one remaining blocking state-consistency bug in the ontology dashboard's search flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - I checked the final head against the PR description and prior review thread, and avoided re-flagging the issues already fixed in follow-up commits. The remaining finding is directly visible in `ontology-state.svelte.ts`: `loadGraph()` clears `filterQuery`/`searchMatchIds` on agent switch, but it does not cancel the already-scheduled debounced `searchGraph()` timer or invalidate its generation. That stale callback can still run after the switch and repopulate `searchMatchIds` from the old query, leaving the graph filtered again even though the search box was reset.

NicholaiVogel:
Fixed in 54f2da17:

The clear block in `loadGraph` now cancels the pending timer and increments `searchGeneration` before clearing `filterQuery`/`searchMatchIds`:

```ts
if (searchTimer) { clearTimeout(searchTimer); searchTimer = null; }
searchGeneration++;
```

`clearTimeout` stops timers that haven't fired yet. The `searchGeneration++` handles the other case: if the timer already fired and its callback is sitting in the macrotask queue, the callback's final `if (gen !== searchGeneration) return` check will fail and it exits without writing `searchMatchIds`. Both are needed — `clearTimeout` alone doesn't help if the callback already dispatched.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,422 @@
...
+
+let projectionGeneration = 0;
+
+export async function loadProjection(agentId = "default"): Promise<void> {
+	const gen = ++projectionGeneration;
+	ontology.loadingProjection = true;
+	try {
+		const result = await getProjection(2, { limit: 500, agentId });
+		if (gen !== projectionGeneration) return;
+		ontology.projection = result;
```
PR-Reviewer-Ant, line 285: Blocking: `loadTableStats()` still has a cross-agent race. The post-`await` guard only checks `ontology.schemaTable === table`, so if the user switches agents quickly while keeping the same table selected, a slower response for the previous agent can still overwrite `ontology.tableStats` for the new scope. This undermines the PR's claim that agent switches are guarded cleanly and can resurface stale per-agent counts in the inspector.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,422 @@
...
+		// Phase 2: expand — parents (walk up) + children (walk down twice for grandchildren)
+		const result = new Set(matched);
+		for (const id of matched) {
+			let pid = map.get(id)?.parentId;
+			while (pid) {
+				result.add(pid);
+				pid = map.get(pid)?.parentId;
+			}
+		}
+		for (const n of nodes) {
```
PR-Reviewer-Ant, line 195: The inspector loaders do not clear prior detail state before fetching the next node. When switching from entity/aspect A to B, `detailAspects`, `detailDependencies`, and `aspectAttrs` keep their old values until the new request resolves, so `InspectorPanel` can render B's header with A's aspects/attributes/dependencies in the meantime. That diverges from the PR description's drill-down inspector behavior and is user-visible state corruption, not just a loading polish issue.

NicholaiVogel:
Fixed both in ff18bd43:

**loadTableStats generation guard** — replaced the `ontology.schemaTable !== table` stale check with a `tableStatsGeneration` counter (same pattern as `graphGeneration`, `projectionGeneration`, etc.). The old guard only prevented writes when the user navigated to a different table, not when the same table was reloaded for a different agent. Now any response that arrives after a newer call has started is dropped unconditionally.

**Stale inspector detail** — `loadEntityDetail` now clears `detail`, `detailAspects`, `detailAttributes`, and `detailDependencies` immediately after incrementing the generation counter, before the first `await`. `loadAspectDetail` clears `aspectAttrs` the same way. Previously, the old content stayed rendered until the network response arrived.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,430 @@
...
+			if (gen !== detailGeneration) return;
+			for (let i = 0; i < aspects.length; i++) {
+				attrMap.set(aspects[i].aspect.id, results[i]);
+			}
+		}
+		ontology.detailAttributes = attrMap;
+	} finally {
+		if (gen === detailGeneration) ontology.loadingDetail = false;
+	}
+}
```
PR-Reviewer-Ant, line 248: Blocking: `loadProjection()` still leaves the previous agent's projection in `ontology.projection` while a new agent-scoped request is in flight. If the UMAP panel is open and the user switches agents, the panel keeps showing old-agent embedding points until the new fetch resolves. Given the PR's explicit claim that agent-switched loaders are guarded cleanly, this is still a cross-agent data leak. `loadGraph()` already clears its scoped state before `await`; `loadProjection()` should do the same.

NicholaiVogel:
Fixed in d9ce7ded: added `ontology.projection = null` immediately after incrementing the generation counter in `loadProjection`, before the `await getProjection(...)`. Same pattern as `loadGraph` clearing `graphNodes`/`graphEdges` — stale data is gone as soon as a new load starts, so the UMAP panel shows the loading state rather than the previous agent's points.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,431 @@
...
+		ontology.entities = data.entities;
+		// Reload table stats for the current schema table under the new agent
+		loadTableStats(ontology.schemaTable, agentId);
+	} catch (err) {
+		if (gen !== graphGeneration) return;
+		ontology.error = err instanceof Error ? err.message : "Unknown error";
+		ontology.graphNodes = [];
+		ontology.graphEdges = [];
+		ontology.entities = [];
+	} finally {
```
PR-Reviewer-Ant, line 148: Blocking: clearing the search does not invalidate an already-fired debounce callback. `searchGraph("")` clears `searchMatchIds` and returns, but it never bumps `searchGeneration`. If the previous timer has already fired and its callback is sitting in the queue, that callback will still pass the final `gen !== searchGeneration` check and repopulate `searchMatchIds` from the old query after the input is blank again. The result is the graph becoming filtered by a stale query while the search box shows empty, which breaks the PR's claim of coherent debounced search behavior.

packages/cli/dashboard/src/lib/components/ontology/ontology-state.svelte.ts
```diff
@@ -0,0 +1,431 @@
...
+				ontology.tableStats = {
+					table,
+					rows: map[table] ?? 0,
+					extra: stats as unknown as Record<string, unknown>,
+				};
+				break;
+			}
+			case "memories":
+			case "memory_entity_mentions": {
+				const { stats } = await getMemories(1, 0, agentId);
```
PR-Reviewer-Ant, line 325: Warning: `loadTableStats()` uses `tableStatsGeneration` to drop stale responses, but its `finally` block still unconditionally sets `ontology.loadingTable = false`. An older request can therefore clear the loading indicator while a newer table/agent-scoped request is still in flight. That does not corrupt the data payload, but it does reintroduce inconsistent async UI state in the same code path this PR is trying to harden.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `d9ce7ded`

pr-reviewer automated bot review: I found one blocking state-consistency bug in the debounced search flow and one additional async loading-state race worth discussing.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - I checked the final head commit and the prior review thread, and avoided re-flagging issues the author explicitly fixed. The remaining blocking bug is directly visible in `searchGraph()`: clearing the search resets `searchMatchIds` but does not invalidate an already-dispatched debounce callback, so stale matches can be written back after the box is empty. The loading-state race is likewise explicit in `loadTableStats()`, whose `finally` block clears `loadingTable` without a generation check even though the function otherwise uses `tableStatsGeneration`.

NicholaiVogel:
Fixed both in b1488ef1:

**searchGraph clear path** — added `searchGeneration++` in the empty-query early-return path, alongside `clearTimeout`. The timer cancel handles callbacks that haven't fired yet; the generation bump handles callbacks already in the macrotask queue (`gen !== searchGeneration` causes early return at the end of the callback). This is the same two-part pattern already applied in `loadGraph` for the agent-switch case.

**loadTableStats finally block** — changed `ontology.loadingTable = false` to `if (gen === tableStatsGeneration) ontology.loadingTable = false`, consistent with every other loader's finally block in this file (`loadGraph`, `loadProjection`, `loadEntityDetail`, `loadAspectDetail`).

NicholaiVogel:
PR-Reviewer-Ant has confirmed no additional issues on b1488ef1. All prior blocking findings have been addressed across the follow-up commits. This is ready for human review.

PR #397, opened 2026-03-29
fix(pipeline): harden dependency prompt auditability

NicholaiVogel:
## Summary

- require a non-empty reason for `related_to` dependency edges
- add immutable dependency edge audit history with DB-level trigger enforcement for all create/update/delete events (not just app write sites)
- tune structural and synthesis dependency prompts for local Ollama models, including `nemotron-3-nano:4b`
- add live prompt harnesses, prompt-shape regression tests, and docs/script shortcuts for prompt testing

## Changes

- `packages/core/src/migrations/050-related-to-audit.ts` - new migration: creates `entity_dependency_history` table, backfills all existing dep edges, adds BEFORE INSERT/UPDATE triggers enforcing non-empty reason for `related_to`, and adds AFTER INSERT/UPDATE/DELETE triggers for DB-level immutable audit coverage
- `packages/daemon-rs/crates/signet-core/src/sql/035-dependency-audit-history.sql` - Rust parity: same schema, backfill, validation triggers, and AFTER INSERT/UPDATE/DELETE audit triggers
- `packages/daemon-rs/crates/signet-services/src/graph.rs` - Rust parity: history writes removed from app layer (triggers own them), `COALESCE` preserves existing reason on update when caller omits it
- `packages/daemon/src/dependency-history.ts` - `requireDependencyReason` and `normalizeDependencyReason` helpers; `writeDependencyHistory` removed (triggers own all audit writes)
- `packages/daemon/src/inline-entity-linker.ts` - narrow catch blocks to UNIQUE constraint only; removed redundant app-layer history call
- `packages/daemon/src/pipeline/graph-transactions.ts` - same; removed redundant app-layer history call
- `packages/daemon/src/knowledge-graph.ts` - removed redundant app-layer history calls; DELETE path simplified (trigger handles audit)
- `packages/daemon/src/path-feedback.ts` - removed redundant app-layer history calls
- `packages/daemon/src/pipeline/structural-dependency.ts` - prompt tuning for local models
- `packages/daemon/src/pipeline/dependency-synthesis.ts` - prompt tuning for local models

## Audit design

All dependency create/update/delete events are captured by SQLite AFTER triggers in the migration:
- `trg_entity_dependencies_audit_insert` — fires on every INSERT
- `trg_entity_dependencies_audit_update` — fires on every UPDATE (no column filter; covers all column changes unconditionally)
- `trg_entity_dependencies_audit_delete` — fires on every DELETE

This means FK cascades, direct SQL, and all application write paths are covered without relying on call-site discipline. Tests assert `changed_by = 'db-trigger'` to verify trigger attribution.

## Type

- [x] `fix` - bug fix

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [x] `packages/daemon-rs`

## Screenshots

N/A - no UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

**Rollback note:** migration 050 adds `entity_dependency_history` (additive table), indexes, and triggers on `entity_dependencies`. Rolling back drops the table, indexes, and all five triggers. The backfill is idempotent (guarded by `NOT EXISTS` on `event = 'backfill'`).

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes

## AI disclosure

## Notes

- left unrelated untracked `logger-config` files out of this PR
- workspace-wide typecheck still has unrelated pre-existing failures outside this change set
- 3 pre-existing `txPersistEntities` failures in graph-transactions.test.ts predate this branch

### DIFF
--- a/packages/daemon-rs/crates/signet-services/src/graph.rs
+++ b/packages/daemon-rs/crates/signet-services/src/graph.rs
@@ -345,8 +345,8 @@ pub fn upsert_dependency(
         )
         .ok();
 
-    if let Some((eid, _, _)) = existing {
-        let s = strength.unwrap_or(0.5);
+    if let Some((eid, existing_strength, _)) = existing {
+        let s = strength.unwrap_or(existing_strength);
         conn.execute(
             "UPDATE entity_dependencies
              SET strength = ?1, aspect_id = ?2, updated_at = ?3,
--- a/packages/daemon-rs/crates/signet-pipeline/src/dep_synthesis.rs
+++ b/packages/daemon-rs/crates/signet-pipeline/src/dep_synthesis.rs
@@ -183,7 +183,18 @@ async fn tick(
             let src = entity.id.clone();
             let tgt = target_id;
             let dep_type = result.dep_type.clone();
-            let reason = result.reason.clone();
+            // Mirror TS normalization: trim before fallback check so whitespace-only
+            // model output doesn't bypass the related_to reason enforcement.
+            let raw = result.reason.trim().to_string();
+            let reason = if dep_type == "related_to" && raw.is_empty() {
+                format!(
+                    "llm synthesized a loose association from {} to {}",
+                    entity.name, result.target
+                )
+            } else {
+                raw
+            };
+            let reason_opt: Option<String> = if reason.is_empty() { None } else { Some(reason) };
 
             let res = pool
                 .write(Priority::Low, move |conn| {
@@ -197,7 +208,7 @@ async fn tick(
                             dependency_type: &dep_type,
                             strength: Some(0.5),
                             confidence: None,
-                            reason: Some(reason.as_str()),
+                            reason: reason_opt.as_deref(),
                         },
                     )?;
                     Ok(serde_json::Value::Null)


BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): harden dependency prompt auditability` (commit `fbd05ad6`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): harden dependency prompt auditability` (commit `fbd05ad6`) now and I'll follow up shortly with feedback.

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,105 @@
...
+			lower(hex(randomblob(16))),
+			d.id,
+			d.source_entity_id,
+			d.target_entity_id,
+			d.agent_id,
+			d.dependency_type,
+			'backfill',
+			'migration-050',
+			CASE
+				WHEN d.reason IS NULL OR length(trim(d.reason)) = 0
```
PR-Reviewer-Ant: The migration backfill only records history for `related_to` edges. The PR description and docs now describe immutable audit history for dependency edge changes broadly, but every pre-existing non-`related_to` dependency will still have no audit trail after this migration. That makes the implementation diverge from the stated goal and leaves historical coverage incomplete.

packages/daemon/src/inline-entity-linker.ts
```diff
@@ -398,13 +402,41 @@ export function linkMemoryToEntities(
 			for (let j = i + 1; j < ids.length; j++) {
```
PR-Reviewer-Ant, line 400: This `try/catch` now wraps both the dependency insert and the new history write, but the catch still treats every exception as 'already exists'. If `writeDependencyHistory()` fails for any real reason (schema drift, malformed metadata, DB error), the code will silently swallow it and lose the audit record while pretending nothing went wrong. The same pattern appears in `graph-transactions.ts` for structured co-occurrence edges.

NicholaiVogel:
Addressed both issues in `85e1b2ff`:

1. **Backfill scope** — removed the `WHERE d.dependency_type = 'related_to'` filter from the `INSERT INTO entity_dependency_history` in migration 050. The audit history now backfills all existing dependency edge types, not just `related_to`. The `UPDATE` that stamps a legacy reason still targets only `related_to` rows since that's the only type with the non-empty-reason constraint.

2. **Broad catch blocks** — all four `catch {}` blocks in `inline-entity-linker.ts` and `graph-transactions.ts` now narrow to UNIQUE constraint violations only. Any real DB failure or `writeDependencyHistory` throw re-propagates instead of being silently swallowed under "Already exists".

All tests pass (`migrations.test.ts`, `graph-transactions.test.ts --test-name-pattern "records reasoned related_to audit history"`, `path-feedback.test.ts`).

NicholaiVogel:
Also addressed the reviewer's residual uncertainty about cross-module coverage:

**All dependency write paths call `writeDependencyHistory`:**

| File | INSERT | UPDATE |
|---|---|---|
| `inline-entity-linker.ts` | yes (428) | n/a |
| `graph-transactions.ts` | yes (424) | n/a |
| `knowledge-graph.ts` | yes (414) | yes (386, on field changes) |
| `path-feedback.ts` | yes (524, 547) | yes (314, 547) |

No write site is missing audit coverage. The migration 050 backfill now also covers all dependency types (not just `related_to`) as confirmed by the migrations test suite.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): harden dependency prompt auditability` (commit `85e1b2ff`) now and I'll follow up shortly with feedback.

packages/daemon/src/knowledge-graph.ts
```diff
@@ -355,7 +359,10 @@ export function upsertDependency(
 			) as Record<string, unknown> | undefined;
```
PR-Reviewer-Ant, line 356: Blocking: this PR changes daemon behavior for dependency persistence and audit history, but there are no matching changes under `packages/daemon-rs/`. The repo explicitly requires daemon/daemon-rs behavioral parity, and shadow mode will now diverge on dependency create/update/delete paths (including the new audit table writes and required `related_to` reason semantics). This needs a matching Rust-side update or the PR should not claim the behavior as complete.

docs/specs/complete/knowledge-architecture-schema.md
```diff
@@ -205,6 +207,9 @@ CREATE INDEX idx_entity_dependencies_agent ON entity_dependencies(agent_id);
 ```
 
 These are explicit traversal edges. They are not similarity artifacts.
+`related_to` is a special case: it MUST carry a non-empty human-auditable
```
PR-Reviewer-Ant, line 210: Warning: the PR changes the schema/behavior contract (`related_to` must carry a reason, dependency history is now mandatory), but the mandatory spec registry files mentioned in repo policy are not updated in this diff (`docs/specs/INDEX.md` and `docs/specs/dependencies.yaml`). That leaves the spec system out of sync with the implementation contract and makes future dependency planning/review less reliable.

NicholaiVogel:
Addressed daemon-rs parity in `030b2522`:

**Migration 035** (`packages/daemon-rs/crates/signet-core/src/sql/035-dependency-audit-history.sql`):
- Creates `entity_dependency_history` table with identical schema to TS migration 050
- Backfills all existing dependency edges (all types, not just `related_to`)
- Adds the same `BEFORE INSERT` / `BEFORE UPDATE` triggers enforcing non-empty reason on `related_to` edges

**`graph.rs` `upsert_dependency`**:
- Fetch `prev_reason` alongside `id`/`strength` in the existence check
- Call `write_dependency_history` with event `"created"` on INSERT, `"updated"` on UPDATE
- `changed_by` tag is `"daemon-rs:graph"` for traceability

All 197 daemon-rs tests pass including `graph::tests::dependency_crud`.

The spec-process note: this is a hardening fix (enforcement + audit trail on an existing schema concept), not a new feature, so it falls under the existing knowledge graph work rather than requiring a new spec stub.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): harden dependency prompt auditability` (commit `030b2522`) now and I'll follow up shortly with feedback.

packages/daemon-rs/crates/signet-services/src/graph.rs
```diff
@@ -303,6 +303,31 @@ pub fn delete_attribute(conn: &Connection, id: &str, agent_id: &str) -> Result<(
 // Dependency CRUD
```
PR-Reviewer-Ant, line 303: The PR description and updated docs/specs say dependency history is written for create/update/delete events, but the Rust parity change here only adds history writes on upsert. I do not see a matching delete-path audit on the daemon-rs side, which means shadow-mode parity will diverge as soon as a dependency is deleted from Rust-backed flows.

packages/daemon/src/pipeline/dependency-synthesis.test.ts
```diff
@@ -0,0 +1,116 @@
+/**
```
PR-Reviewer-Ant, line 1: This is a live Ollama test under the default `packages/` test discovery root, so it becomes part of normal `bun test` runs. Because it asserts exact semantic edges from a real model, it is inherently nondeterministic and can fail on developer/CI machines that happen to have the model installed. These checks should be moved behind an explicit opt-in harness instead of the default unit/integration suite.

NicholaiVogel:
Addressed both issues in `7065be87`:

**1. Delete-path audit parity** (`graph.rs`): `delete_dependency` now fetches `source_entity_id`, `target_entity_id`, `dependency_type`, and `reason` before the `DELETE`, then calls `write_dependency_history` with `event = "deleted"` — matching the TS `deleteDependency` in `knowledge-graph.ts`.

**2. Nondeterministic Ollama tests**: both `dependency-synthesis.test.ts` and `structural-dependency.test.ts` now check for `SIGNET_OLLAMA_TEST_MODEL` before doing anything live. If the env var is not set, the live test returns immediately with a skip log. The prompt-shape assertions (pure unit tests) still run unconditionally. Verified: 7 tests pass, 2 live tests skip cleanly without the env var.

packages/daemon-rs/crates/signet-services/src/graph.rs
```diff
@@ -339,18 +364,32 @@ pub fn upsert_dependency(
...
         )
         .ok();
 
-    if let Some((eid, _)) = existing {
+    if let Some((eid, _, prev_reason)) = existing {
         let s = strength.unwrap_or(0.5);
         conn.execute(
             "UPDATE entity_dependencies
              SET strength = ?1, aspect_id = ?2, updated_at = ?3, reason = ?5
              WHERE id = ?4",
```
PR-Reviewer-Ant, line 355: The Rust parity update path now overwrites `entity_dependencies.reason` with the raw optional input (`reason = ?5`). When a caller updates strength/aspect without supplying a reason, this clears any existing reason instead of preserving it. That silently degrades auditability for non-`related_to` edges and can make `related_to` updates fail under the new trigger even though the row already had a valid reason. The TS path preserves the previous reason; Rust should do the same.

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,104 @@
...
+			END,
+			NULL,
+			'{"source":"migration-050"}',
+			d.updated_at
+		FROM entity_dependencies d
+		WHERE NOT EXISTS (
+			SELECT 1
+			FROM entity_dependency_history h
+			WHERE h.dependency_id = d.id
+			  AND h.event = 'backfill'
```
PR-Reviewer-Ant: This migration does not actually enforce the PR's claimed 'every create/update/delete writes immutable history' contract at the database layer. It adds only `related_to` reason triggers; history rows are still written opportunistically from selected application call sites. Dependency rows deleted via other paths, especially FK cascades when an entity is deleted, will disappear without any `entity_dependency_history` entry. That means the implementation diverges from the PR description and from the new docs/spec wording that all deletes are audited.

NicholaiVogel:
Addressed both issues in `4d7d785f`:

**1. UPDATE nulling existing reason** (`graph.rs`): Changed `SET ... reason = ?5` to `SET ... reason = COALESCE(?5, reason)`. When the caller passes `None`, the existing stored reason is preserved rather than overwritten with NULL.

**2. DB-level delete audit gap**: Added an `AFTER DELETE` trigger `trg_entity_dependencies_audit_delete` to both TS migration 050 and Rust migration 035. The trigger inserts a history row with `event = 'deleted'` and `changed_by = 'db-trigger'` for every delete, regardless of whether it came through the application layer or a FK cascade. Application-path deletes will produce two history rows (app writes first, trigger fires second) -- that's intentional and documented in the trigger comment.

Re the 3 pre-existing `txPersistEntities` failures in `graph-transactions.test.ts`: confirmed pre-existing before any change in this PR (reproduced on the previous commit). Not introduced here.

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,135 @@
...
+		END;
+	`);
+
+	db.exec(`
+		CREATE TRIGGER trg_entity_dependencies_related_to_reason_update
+		BEFORE UPDATE OF dependency_type, reason ON entity_dependencies
+		FOR EACH ROW
+		WHEN NEW.dependency_type = 'related_to'
+		  AND (NEW.reason IS NULL OR length(trim(NEW.reason)) = 0)
+		BEGIN
```
PR-Reviewer-Ant, line 64: Blocking: this trigger causes every application-layer dependency delete to be audited twice. `deleteDependency()` now writes a `deleted` history row before deleting, and this `AFTER DELETE` trigger writes a second `deleted` row for the same logical operation. That makes the audit trail ambiguous and breaks any consumer that counts or replays events. If you want DB-level coverage for cascades/direct SQL deletes, the trigger needs a dedupe mechanism or the app-layer delete paths need to stop pre-writing delete events.

NicholaiVogel:
Fixed in `90d0f185`: removed the app-layer `writeDependencyHistory` call from `deleteDependency` (TS) and `delete_dependency` (Rust). The `AFTER DELETE` trigger (`trg_entity_dependencies_audit_delete`) in migrations 050/035 now owns all delete auditing — one row per delete, whether it's an app delete, FK cascade, or direct SQL. Both files have a comment explaining the intent.

packages/daemon/src/knowledge-graph.ts
```diff
@@ -370,6 +377,26 @@ export function upsertDependency(
 				existing.id as string,
```
PR-Reviewer-Ant, line 384: Blocking: this PR claims every dependency edge create/update/delete is audited, but the existing-edge path still updates `updated_at` unconditionally and only writes history when one of `strength`, `aspectId`, `confidence`, or `reason` is provided. A caller can therefore mutate an existing row via `upsertDependency()` without any `entity_dependency_history` entry. Either skip the UPDATE when nothing is changing, or always emit a history row for this path so the implementation matches the new audit contract.

packages/daemon-rs/crates/signet-services/src/graph.rs
```diff
@@ -339,18 +364,33 @@ pub fn upsert_dependency(
...
         conn.execute(
             "UPDATE entity_dependencies
-             SET strength = ?1, aspect_id = ?2, updated_at = ?3, reason = ?5
+             SET strength = ?1, aspect_id = ?2, updated_at = ?3,
+                 reason = COALESCE(?5, reason)
              WHERE id = ?4",
             params![s, aspect_id, ts, eid, reason],
         )?;
+        let reason_str = reason.unwrap_or("updated without reason");
+        write_dependency_history(
```
PR-Reviewer-Ant: Warning: on the update path, `reason = COALESCE(?5, reason)` preserves the existing dependency reason when the caller omits `reason`, but the audit row still records `"updated without reason"`. That makes the history inaccurate for the common case where only strength/aspect changes and the persisted reason remains unchanged. The audit record should log the effective reason after the update, not a synthetic placeholder.

NicholaiVogel:
Fixed in `ee159bb6`:

**TS `upsertDependency`**: added a `changed` guard before the UPDATE block. When no mutable fields are supplied (`strength`, `aspectId`, `confidence`, `reason` all undefined), the function returns the existing row immediately — no UPDATE, no history write, no spurious `updated_at` touch.

**Rust `upsert_dependency`**: history reason now uses `reason.unwrap_or_else(|| prev_reason.as_deref().unwrap_or("updated without reason"))`. When the caller provides no reason, the history records the actual stored reason (unchanged via COALESCE) rather than a misleading placeholder.

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,135 @@
...
+		WHERE NOT EXISTS (
+			SELECT 1
+			FROM entity_dependency_history h
+			WHERE h.dependency_id = d.id
+			  AND h.event = 'backfill'
+		  )
+	`);
+
+	db.exec("DROP TRIGGER IF EXISTS trg_entity_dependencies_related_to_reason_insert");
+	db.exec("DROP TRIGGER IF EXISTS trg_entity_dependencies_related_to_reason_update");
```
PR-Reviewer-Ant, line 42: This migration enforces `related_to` reasons and audits deletes at the DB layer, but it does not audit inserts or updates outside the application helper paths. That means direct SQL writes, future write sites that forget `writeDependencyHistory`, or maintenance scripts can create/update dependency rows without any history entry, which diverges from the PR/docs claim that every create/update/delete is immutably audited. Either add AFTER INSERT/UPDATE history triggers as well, or narrow the claim/documentation to 'all current app write sites' instead of a universal guarantee.

NicholaiVogel:
Round 7: added AFTER INSERT + AFTER UPDATE audit triggers to both migrations; removed all app-layer `writeDependencyHistory` calls for created/updated events.

**What changed:**
- `050-related-to-audit.ts` + `035-dependency-audit-history.sql`: added `trg_entity_dependencies_audit_insert` (AFTER INSERT) and `trg_entity_dependencies_audit_update` (AFTER UPDATE OF strength, aspect_id, reason, confidence). The DB now enforces immutable audit records for all three event types at the schema level.
- Removed app-layer history writes from all 4 TS write sites (knowledge-graph, inline-entity-linker, graph-transactions, path-feedback) and from `upsert_dependency` in daemon-rs/graph.rs — no more duplicate rows, no more coverage gap.
- Updated tests to expect `changed_by = 'db-trigger'` and query by `event` type to handle same-second timestamp ties.

All 3 `txPersistEntities` failures are pre-existing (predated this branch). Rust cargo check and dependency tests pass clean.

packages/daemon/src/dependency-history.ts
```diff
@@ -0,0 +1,80 @@
...
+
+	return text;
+}
+
+export function requireDependencyReason(
+	dependencyType: DependencyType,
+	reason?: string | null,
+): string | null {
+	const text = normalizeDependencyReason(dependencyType, reason);
+	if (dependencyType === "related_to" && text === null) {
```
PR-Reviewer-Ant, line 33: This PR introduces `writeDependencyHistory`, but the changed write sites do not call it; the implementation now relies on the SQLite triggers from migration 050, and the new tests assert `changed_by = 'db-trigger'`. That diverges from the PR description (`all four dependency write sites call writeDependencyHistory`) and leaves a dead second audit mechanism in the tree. This is a maintainability trap: a future change can update the helper and assume app-layer auditing is active when it is not.

NicholaiVogel:
Follow-up to the COMMENTED review: removed the dead `writeDependencyHistory` export from `dependency-history.ts` (no call sites remain since triggers took over) and updated the PR description to accurately describe the trigger-based audit approach.

The PR body now explains that AFTER INSERT/UPDATE/DELETE triggers own all audit writes, test assertions verify `changed_by = 'db-trigger'`, and the previous claim about 'all four write sites call writeDependencyHistory' is gone.

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,187 @@
...
+				lower(hex(randomblob(16))),
+				NEW.id,
+				NEW.source_entity_id,
+				NEW.target_entity_id,
+				NEW.agent_id,
+				NEW.dependency_type,
+				'created',
+				'db-trigger',
+				COALESCE(NEW.reason, 'created without reason'),
+				NULL,
```
PR-Reviewer-Ant, line 90: The PR description says the DB trigger layer captures all dependency create/update/delete events, but this trigger only fires on `UPDATE OF strength, aspect_id, reason, confidence`. Any direct SQL or future app path that updates `dependency_type`, `source_entity_id`, `target_entity_id`, or `agent_id` will bypass the audit log entirely. That leaves the new audit history incomplete and diverges from the stated goal of full DB-level coverage.

packages/daemon-rs/crates/signet-core/src/sql/035-dependency-audit-history.sql
```diff
@@ -0,0 +1,163 @@
...
+        metadata, created_at
+    ) VALUES (
+        lower(hex(randomblob(16))),
+        NEW.id,
+        NEW.source_entity_id,
+        NEW.target_entity_id,
+        NEW.agent_id,
+        NEW.dependency_type,
+        'created',
+        'db-trigger',
```
PR-Reviewer-Ant, line 73: Rust parity has the same audit gap: the update trigger only covers `strength, aspect_id, reason, confidence`, not all updates. Since the PR explicitly claims full create/update/delete audit coverage and Rust parity, this leaves the Rust path with the same missing-update behavior.

NicholaiVogel:
Fixed: removed the `OF strength, aspect_id, reason, confidence` column filter from the update audit trigger in both migrations. `trg_entity_dependencies_audit_update` now fires on any `UPDATE` to `entity_dependencies`, covering all column changes unconditionally.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `85fae787`

```pr-review-json
{}
```

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,187 @@
...
+				metadata, created_at
+			) VALUES (
+				lower(hex(randomblob(16))),
+				NEW.id,
+				NEW.source_entity_id,
+				NEW.target_entity_id,
+				NEW.agent_id,
+				NEW.dependency_type,
+				'created',
+				'db-trigger',
```
PR-Reviewer-Ant, line 88: The implementation diverges from the PR description here. The description says the audit trigger should cover updates to `strength`, `aspect_id`, `reason`, and `confidence`, but this trigger is `AFTER UPDATE ON entity_dependencies` with no column list. That means any future update that only touches bookkeeping fields like `updated_at` will still append an `updated` audit row, creating false-positive history and making the audit log less trustworthy. If the intent is only semantic edge changes, this should be `AFTER UPDATE OF strength, aspect_id, reason, confidence ...` (and parity-adjusted in the Rust SQL too).

NicholaiVogel:
Updated PR description: the audit design section now correctly says the update trigger fires on every UPDATE with no column filter, rather than the stale reference to specific columns.

NicholaiVogel:
All blocking issues resolved across 9 rounds of automated review. Latest reviews on commit `85fae787` are COMMENTED (non-blocking). PR is ready for human review.

**What's in this PR (final state):**
- Migration 050: `entity_dependency_history` table + backfill + 5 triggers:
  - BEFORE INSERT/UPDATE: enforce non-empty reason for `related_to`
  - AFTER INSERT/UPDATE/DELETE: immutable audit history, DB-owned (no column filter on UPDATE)
- Rust parity (migration 035): identical schema and triggers
- All app-layer `writeDependencyHistory` calls removed — triggers own all audit writes
- Dead `writeDependencyHistory` export removed from `dependency-history.ts`
- Tests verify `changed_by = 'db-trigger'` attribution

docs/CONFIGURATION.md
```diff
@@ -316,7 +316,7 @@ For safety, the intended extraction setups are:
 
 - `claude-code` on a Haiku model
 - `codex` on a GPT Mini model
-- local `ollama` with at least `qwen3:4b`
+- local `ollama` with at least `qwen3:4b` or `nemotron-3-nano:4b`
```
NicholaiVogel: would probably be worth elaborating here that nemotron-3-nano:4b is the preferred model moving forward, and that qwen3:4b is being deprecated in favor of Nemotron's superior reasoning and intelligence, which basically means that using Qwen3:4b will result in worse performance over the next few updates.

docs/PIPELINE.md
```diff
@@ -533,10 +533,13 @@ Two implementations are shipped:
 
 **OllamaProvider** calls the Ollama HTTP API at `POST /api/generate` with
 `stream: false`. The default base URL is `http://localhost:11434` and the
-default model is `qwen3:4b`. Each `generate` call sets an `AbortController`
+default model is `qwen3:4b`. `nemotron-3-nano:4b` is also supported as a
```
NicholaiVogel: Same methodology should be applied here as my previous comment.

docs/SELF-HOSTING.md
```diff
@@ -512,10 +512,12 @@ running and the model is pulled:
 curl -s http://localhost:11434/api/tags | jq '.models[].name'
 ```
 
-The recommended local floor is `qwen3:4b`. If it is not listed:
+The recommended local floor is `qwen3:4b` or `nemotron-3-nano:4b`. If neither is listed:
```
NicholaiVogel: Same deal as before.

packages/daemon/src/pipeline/model-registry.ts
```diff
@@ -64,6 +64,7 @@ const KNOWN_MODELS: Record<string, ModelRegistryEntry[]> = {
 	],
 	ollama: [
 		{ id: "qwen3:4b", provider: "ollama", label: "Qwen3 4B", tier: "low", deprecated: false },
```
NicholaiVogel: Could we note here that Qwen3:4B is soon to be deprecated?

package.json
```diff
@@ -38,7 +38,9 @@
     "publish:signetai": "cd packages/signetai && npm publish --access public",
     "dev:web": "cd web && bun run dev",
     "deploy:web": "cd web && bun run deploy",
-    "changelog": "bun scripts/changelog.ts"
+    "changelog": "bun scripts/changelog.ts",
+    "test:prompt:structural": "bun test packages/daemon/src/pipeline/structural-dependency.test.ts",
```
PR-Reviewer-Ant: The new `test:prompt:*` shortcuts do not currently execute a live prompt harness on their own. Both `structural-dependency.test.ts` and `dependency-synthesis.test.ts` now return early unless `SIGNET_OLLAMA_TEST_MODEL` is explicitly set, so `bun run test:prompt:structural` / `bun run test:prompt:synthesis` will just skip. That diverges from the PR description and the new README examples, which present these as usable harness shortcuts. Either wire the scripts to set a default model, or update the scripts/docs to say the env var is mandatory.

packages/daemon-rs/crates/signet-services/src/graph.rs
```diff
@@ -339,15 +339,16 @@ pub fn upsert_dependency(
...
-            |r| Ok((r.get(0)?, r.get(1)?)),
+            |r| Ok((r.get(0)?, r.get(1)?, r.get(2)?)),
         )
         .ok();
 
-    if let Some((eid, _)) = existing {
+    if let Some((eid, _, _)) = existing {
         let s = strength.unwrap_or(0.5);
         conn.execute(
             "UPDATE entity_dependencies
```
PR-Reviewer-Ant, line 351: The updated `upsert_dependency` path no longer persists `confidence` for existing edges. The new query only loads `(id, strength, reason)`, and the `UPDATE entity_dependencies` statement sets `strength`, `aspect_id`, `updated_at`, and `reason`, but never `confidence`. Any caller that tries to raise/lower confidence on an existing dependency will silently get stale data, which is a behavior regression and breaks the PR's claimed TS/Rust parity.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `7eb7db13`

[Automated bot review] Found a Rust parity regression in the dependency upsert path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed Rust SQL in packages/daemon-rs/crates/signet-services/src/graph.rs now selects existing reason state and updates reason via COALESCE, but the UPDATE statement no longer writes confidence at all. That means any caller-provided confidence change is silently dropped on existing edges. This is directly visible in the diff and conflicts with the PR's stated daemon-rs parity goal.

NicholaiVogel:
Fixed the Rust parity regression: added `confidence: Option<f64>` to `UpsertDepInput`, wired `COALESCE(?6, confidence)` into the UPDATE SQL, and included `confidence` in the INSERT column list. Also added the `confidence` column to the test setup schema. All Rust dependency tests pass.

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,187 @@
...
+		BEGIN
+			INSERT INTO entity_dependency_history (
+				id, dependency_id, source_entity_id, target_entity_id, agent_id,
+				dependency_type, event, changed_by, reason, previous_reason,
+				metadata, created_at
+			) VALUES (
+				lower(hex(randomblob(16))),
+				NEW.id,
+				NEW.source_entity_id,
+				NEW.target_entity_id,
```
PR-Reviewer-Ant, line 84: Blocking: this migration stores mixed timestamp formats in `entity_dependency_history.created_at`. The backfill path copies `d.updated_at` (repo timestamps are ISO strings like `2026-03-29T...Z`), but the triggers write `datetime('now')` (`YYYY-MM-DD HH:MM:SS`). The PR also adds queries/tests that sort audit rows with `ORDER BY created_at DESC`, so lexical ordering is no longer reliably chronological once both formats coexist. That can return stale audit events as the 'latest' row. Use one canonical format for both backfill and trigger inserts.

packages/daemon/src/knowledge-graph.ts
```diff
@@ -355,7 +356,20 @@ export function upsertDependency(
 			) as Record<string, unknown> | undefined;
 
 		if (existing) {
-			const reason = params.reason ?? (existing.reason as string | null);
+			const changed =
+				params.strength !== undefined ||
+				params.aspectId !== undefined ||
+				params.confidence !== undefined ||
+				params.reason !== undefined;
```
PR-Reviewer-Ant, line 363: Warning: this changes update semantics for non-`related_to` edges from 'nullable reason' to 'reason can never be cleared once set'. `params.reason ?? existing.reason` treats explicit `null` the same as omission, so callers cannot remove a stale reason anymore. The PR description says existing reasons are preserved when the caller omits them, but this also preserves them when the caller explicitly passes `null`. That is a silent behavior change against the still-nullable schema/docs.

NicholaiVogel:
Addressed both issues:

1. **Timestamp inconsistency** (fixed): backfill now uses `datetime('now')` for `created_at` in both migrations, so all history records consistently mean "when this row was written to entity_dependency_history". The `event = 'backfill'` field already distinguishes historical records.

2. **COALESCE reason/confidence** (intentional, not a regression): `COALESCE(?, reason)` and `COALESCE(?, confidence)` intentionally preserve existing values when the caller passes `null`. This matches the TS `upsertDependency` behavior, which already falls back to the existing reason at the app layer (`params.reason ?? existing.reason`). For `related_to` edges, the BEFORE UPDATE trigger enforces non-null reason anyway; for other types, nulling out a reason has no legitimate use case and would degrade auditability.

packages/daemon/src/pipeline/dependency-synthesis.ts
```diff
@@ -255,14 +268,18 @@ async function tick(deps: DependencySynthesisDeps): Promise<void> {
 			if (!target || target.id === entity.id) continue;
 
 			try {
```
PR-Reviewer-Ant, line 270: Blocking: the new fallback only runs when `!result.reason`. If the model returns whitespace for a `related_to` reason, this branch treats it as present, passes it through, and `requireDependencyReason()` later throws after trimming. That means the PR still drops some `related_to` edges instead of hardening them. Trim or normalize before deciding whether to synthesize the fallback reason.

packages/daemon/src/pipeline/structural-dependency.ts
```diff
@@ -358,6 +378,10 @@ async function processDependencyBatch(
 						agentId: "default",
 						name: result.aspect,
 					});
+					const reason =
```
PR-Reviewer-Ant, line 382: Blocking: same bug here as in dependency synthesis. `result.reason === null` misses whitespace-only output from the model, so `related_to` writes can still fail the new DB/app-layer invariant and be discarded. This diverges from the PR's goal of making these prompt paths robust for small local models.

packages/daemon/src/pipeline/graph-transactions.test.ts
```diff
@@ -171,6 +172,69 @@ describe("graph-transactions", () => {
 				.all();
 			expect(mentions).toHaveLength(2);
 		});
+
```
PR-Reviewer-Ant, line 175: Warning: the PR description claims immutable audit coverage for create/update/delete at the DB layer, but the new tests only exercise insert/update-trigger attribution. There is still no regression test for `trg_entity_dependencies_audit_delete`, so the delete-path guarantee in the description is not actually verified by this PR.

NicholaiVogel:
@PR-Reviewer-Ant please re-review — new commit 2982671d addresses your feedback:
- Both prompt workers now trim the reason before the falsy check, so whitespace-only model output no longer bypasses the `related_to` fallback
- Added regression test for `trg_entity_dependencies_audit_delete` in graph-transactions.test.ts

docs/specs/complete/knowledge-architecture-schema.md
```diff
@@ -205,6 +207,9 @@ CREATE INDEX idx_entity_dependencies_agent ON entity_dependencies(agent_id);
 ```
```
PR-Reviewer-Ant, line 207: This PR changes the normative schema contract by introducing `entity_dependency_history` semantics and a new trigger-enforced `related_to` reason requirement, but the diff does not include the corresponding `docs/specs/INDEX.md` / `docs/specs/dependencies.yaml` sync that the repo’s session protocol requires for schema/spec changes. That leaves the spec registry stale even though the PR checklist says spec alignment was validated.

NicholaiVogel:
@PR-Reviewer-Ant re-review please — commit ffb66421 adds the missing spec registry sync:
- Added `entity_dependency_history` table DDL and integration contracts to `knowledge-architecture-schema.md` (§5.5.1)
- Updated INDEX.md audit contract note for `knowledge-architecture-schema`

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,187 @@
...
+			previous_reason   TEXT,
+			metadata          TEXT,
+			created_at        TEXT NOT NULL DEFAULT (datetime('now'))
+		);
+
+		CREATE INDEX IF NOT EXISTS idx_entity_dependency_history_dep
+			ON entity_dependency_history(dependency_id);
+		CREATE INDEX IF NOT EXISTS idx_entity_dependency_history_agent
+			ON entity_dependency_history(agent_id);
+		CREATE INDEX IF NOT EXISTS idx_entity_dependency_history_created
```
PR-Reviewer-Ant, line 35: Blocking: this migration mutates legacy `related_to` rows before writing the backfill audit rows. That means the backfill will record `legacy-unattributed related_to edge` instead of preserving the fact that the original row had no reason, which contradicts the PR/spec contract (`legacy dependency without recorded reason` for missing legacy reasons) and makes the audit trail less trustworthy. The Rust parity migration has the same ordering problem.

NicholaiVogel:
@PR-Reviewer-Ant re-review please — commit 0bf5e96b fixes the migration ordering:
- Backfill INSERT now runs before the UPDATE that stamps 'legacy-unattributed related_to edge'
- History rows for legacy unreasoned edges now faithfully capture original NULL state as 'legacy dependency without recorded reason'
- Same ordering fix applied to both TS (migration-050) and Rust (migration-035)

packages/core/src/migrations/050-related-to-audit.ts
```diff
@@ -0,0 +1,192 @@
...
+				ELSE d.reason
+			END,
+			NULL,
+			'{"source":"migration-050"}',
+			datetime('now')
+		FROM entity_dependencies d
+		WHERE NOT EXISTS (
+			SELECT 1
+			FROM entity_dependency_history h
+			WHERE h.dependency_id = d.id
```
PR-Reviewer-Ant: Blocking: this migration mutates existing `related_to` rows (`UPDATE entity_dependencies SET reason = 'legacy-unattributed related_to edge' ...`) before the audit triggers are created. That means those row updates are not captured as `event='updated'` history rows, which diverges from the PR description's claim of immutable audit coverage for all create/update/delete events at the DB layer. The backfill row preserves the pre-migration state, but there is no audit row for the post-backfill reason actually written into `entity_dependencies`.

packages/daemon-rs/crates/signet-core/src/sql/035-dependency-audit-history.sql
```diff
@@ -0,0 +1,166 @@
...
+            THEN 'legacy dependency without recorded reason'
+        ELSE d.reason
+    END,
+    NULL,
+    '{"source":"migration-035"}',
+    datetime('now')
+FROM entity_dependencies d
+WHERE NOT EXISTS (
+    SELECT 1
+    FROM entity_dependency_history h
```
PR-Reviewer-Ant: Blocking: the Rust parity migration has the same audit gap as the TS migration. It updates legacy `related_to` rows before creating the `AFTER UPDATE` trigger, so the persisted reason rewrite is not recorded in `entity_dependency_history`. That breaks the stated 'same schema, backfill, validation triggers, and AFTER INSERT/UPDATE/DELETE audit triggers' parity and leaves Rust-created databases with incomplete audit lineage for this migration.

NicholaiVogel:
@PR-Reviewer-Ant re-review please — commit 1c8d0417 fixes the migration audit gap:

Migration now: (1) creates table + drops old triggers, (2) creates all audit triggers, (3) backfills original state, (4) UPDATEs legacy related_to edges — the AFTER UPDATE trigger is active at step 4, so each legacy reason stamping is captured as an 'updated' history row. Audit trail is now complete for all migration-time mutations.

packages/daemon-rs/crates/signet-services/src/graph.rs
```diff
@@ -339,29 +341,32 @@ pub fn upsert_dependency(
...
                 agent_id
             ],
-            |r| Ok((r.get(0)?, r.get(1)?)),
+            |r| Ok((r.get(0)?, r.get(1)?, r.get(2)?)),
         )
         .ok();
 
-    if let Some((eid, _)) = existing {
+    if let Some((eid, _, _)) = existing {
         let s = strength.unwrap_or(0.5);
```
PR-Reviewer-Ant: Blocking: the update path still does `let s = strength.unwrap_or(0.5)` and then writes `strength = ?1`. That means any caller updating only `reason` or `confidence` will silently overwrite an existing edge's strength with `0.5`. The TS path preserves the old strength on partial updates, so this is both a data-corruption bug and a Rust/TS parity regression.

packages/daemon/src/pipeline/dependency-synthesis.ts
```diff
@@ -143,7 +143,9 @@ function buildSynthesisPrompt(
 		? `Already connected to: ${[...existing].join(", ")}`
 		: "No existing connections.";
 
-	return `Entity: ${entity.name} (${entity.entityType})
+	return `Task: identify new dependency edges between the focal entity and known graph entities.
```
PR-Reviewer-Ant, line 146: Warning: this PR changes the TS dependency-synthesis prompt contract substantially, but there is no matching daemon-rs prompt update in the diff. Given the repository's explicit shadow-parity rule for `packages/daemon-rs/`, this will cause the TS and Rust pipelines to infer different edges for the same input and generate avoidable shadow divergences.

NicholaiVogel:
@PR-Reviewer-Ant re-review please — commit 9eacb015:
- **Rust strength parity**: `upsert_dependency` now uses `strength.unwrap_or(existing_strength)` instead of `unwrap_or(0.5)`, matching TS behavior
- **Rust dep_synthesis parity**: added `related_to` fallback reason + empty-to-None mapping, mirroring the TS normalization fix

PR #406, opened 2026-03-30
feat(marketplace): mcporter CLI binary generation for installed MCP servers

aaf2tbz:
## Summary

- Adds `POST /api/marketplace/mcp/:id/generate-cli` that compiles a stdio MCP server into a standalone executable at `~/.agents/mcp-bins/<id>` using `mcporter generate-cli --compile`
- Persists `cliPath` on the installed server record so it survives daemon restarts
- Adds **GEN CLI** / **REGEN CLI** button to the installed server card in the MCP Servers tab (stdio servers only; HTTP returns 400)
- Displays a "cli ready" badge once a binary exists; hovering shows the binary path

## Motivation

The standalone binary lets non-MCP harnesses (scripts, older tooling, etc.) invoke an installed MCP server directly without going through the daemon.

## Changes

| File | Change |
|------|--------|
| `packages/daemon/package.json` | Add `mcporter ^0.8.1` dependency |
| `packages/daemon/src/routes/marketplace.ts` | `cliPath` field, sanitized binary name from `server.id`, runtime-aware mcporter resolution, secret-ref rejection, `POST /:id/generate-cli` route |
| `packages/daemon/src/routes/marketplace.test.ts` | 3 new tests: 404 unknown, 400 HTTP, 500 secret refs |
| `packages/cli/dashboard/src/lib/api.ts` | `cliPath` on `MarketplaceMcpServer`, `generateMarketplaceMcpCli()` |
| `packages/cli/dashboard/src/lib/stores/marketplace-mcp.svelte.ts` | `generatingCliId` state, `doGenerateMcpCli()` |
| `packages/cli/dashboard/src/lib/components/marketplace/McpServersTab.svelte` | GEN CLI button + "cli ready" badge |
| `docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md` | Document the implemented endpoint |

## Security

- Binary name derived from `server.id` (unique) passed through `sanitizeServerId()` — prevents path traversal and slug collisions
- Servers with `secret://` references in env/headers are **rejected** with a descriptive error — no credentials are ever written to disk
- mcporter binary resolved from workspace `node_modules/.bin/` first, then runtime-detected fallback (`bun x` or `npx`) — works under both Bun and Node

## Screenshots

The GEN CLI / REGEN CLI button appears in the installed server card action row (stdio servers only), alongside the existing REMOVE button. A "cli ready" badge appears in the metadata row once the binary has been compiled. No new layout or components — changes are additive to existing card structure.

## Type

## Packages affected

- [x] `@signet/daemon`

## PR Readiness (MANDATORY)

## Test plan

- [ ] Install a stdio MCP server (e.g. context7) via the marketplace installed tab
- [ ] Click **GEN CLI** — confirm `~/.agents/mcp-bins/<id>` is created and executable
- [ ] Card shows "cli ready" badge and "REGEN CLI" button after generation
- [ ] HTTP server cards show no GEN CLI button
- [ ] Server with `secret://` env refs → 500 with "cannot be embedded" error
- [ ] `curl -X POST http://localhost:3850/api/marketplace/mcp/<id>/generate-cli` returns `{ success, path }` or descriptive error
- [ ] `bun test packages/daemon/src/routes/marketplace.test.ts` — 8 tests pass

### DIFF
--- a/packages/cli/dashboard/src/lib/stores/marketplace-mcp.svelte.ts
+++ b/packages/cli/dashboard/src/lib/stores/marketplace-mcp.svelte.ts
@@ -232,7 +232,12 @@ export async function doGenerateMcpCli(id: string): Promise<void> {
 		const result = await generateMarketplaceMcpCli(id);
 		if (result.success && result.server) {
 			mcpMarket.installed = mcpMarket.installed.map((s) => (s.id === id ? (result.server ?? s) : s));
-			toast(`CLI binary ready: ${result.path ?? id}`, "success");
+			const warning = (result as Record<string, unknown>).warning;
+			if (typeof warning === "string") {
+				toast(`CLI binary ready: ${result.path ?? id}. Warning: ${warning}`, "warn");
+			} else {
+				toast(`CLI binary ready: ${result.path ?? id}`, "success");
+			}
 		} else {
 			toast(result.error ?? "CLI generation failed", "error");
 		}
--- a/packages/daemon/src/routes/marketplace.ts
+++ b/packages/daemon/src/routes/marketplace.ts
@@ -1827,6 +1827,9 @@ export function mountMarketplaceRoutes(app: Hono): void {
 		if (!server) {
 			return c.json({ error: "Server not found" }, 404);
 		}
+		if (!server.enabled) {
+			return c.json({ error: "Cannot generate CLI for a disabled server" }, 400);
+		}
 		if (server.config.transport !== "stdio") {
 			return c.json({ error: "CLI generation is only supported for stdio servers" }, 400);
 		}


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(marketplace): mcporter CLI binary generation for installed MCP servers` (commit `46db06e0`) now and I'll follow up shortly with feedback.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1106,109 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+	return entry;
+}
+
+interface GenerateCliResult {
+	readonly success: boolean;
+	readonly path?: string;
+	readonly error?: string;
+}
+
+async function runMcporterGenerateCli(server: InstalledMarketplaceMcpServer): Promise<GenerateCliResult> {
```
PR-Reviewer-Ant, line 1218: Blocking: `server.name` is user-controlled (`register` and install `alias` both allow arbitrary names) and is reused as the generated binary name/path without sanitization. A name containing `/`, `..`, or platform-specific separators can escape `~/.agents/mcp-bins/` or at minimum persist an incorrect `cliPath`. This expands attack surface and can write or point outside the intended directory. Reuse the existing ID/path sanitization logic for the output filename and persist the sanitized path instead.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1106,109 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+	const configEntry = buildMcporterConfigEntry(server);
+	const configPayload = { mcpServers: { [server.name]: configEntry } };
+	const tmpConfig = join(tmpdir(), `mcporter-${server.id}-${Date.now()}.json`);
+
+	try {
+		writeFileSync(tmpConfig, JSON.stringify(configPayload, null, 2));
+	} catch (err) {
+		const msg = err instanceof Error ? err.message : String(err);
+		return { success: false, error: `Failed to write temp config: ${msg}` };
+	}
```
PR-Reviewer-Ant, line 1253: Blocking: the PR description claims the daemon runs `bun x mcporter ...`, but the code actually executes `spawn(process.execPath, ['x', 'mcporter', ...])`. That only behaves like `bun x` when the daemon itself is launched by Bun. If the packaged daemon runs under Node, this becomes `node x mcporter ...` and the feature fails outright. The implementation does not currently match the described behavior.

packages/cli/dashboard/src/lib/components/marketplace/McpServersTab.svelte
```diff
@@ -332,8 +327,26 @@ async function removeFromDetail(serverId: string): Promise<void> {
 							<div class="catalog-meta">
 								<span class="mcp-badge">installed</span>
 								<span class="mcp-badge">{server.config.transport}</span>
+								{#if server.cliPath}
+									<span class="mcp-badge" title={server.cliPath}>cli ready</span>
+								{/if}
```
PR-Reviewer-Ant, line 332: The PR adds visible dashboard UI (`GEN CLI`/`REGEN CLI` button and `cli ready` badge) but the description does not include screenshots or visual previews. A screenshot is needed for human UI review.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in b8fd4605. Added `sanitizeBinaryName()` which runs the name through `sanitizeServerId()` (same charset: `[a-z0-9_-]`), then rejects any result that still contains `/`, `\`, `.`, or `..`. The sanitized name is used for both the mcporter config key and the output binary path, and is persisted as `cliPath` instead of the raw `server.name`.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in b8fd4605. Replaced the `process.execPath` assumption with `getMcporterSpawnTarget()` which:
1. Walks up from `import.meta.url` at 3–5 directory levels to find `node_modules/.bin/mcporter` (covers both source and compiled layouts)
2. Falls back to `{ cmd: process.execPath, leadingArgs: ['x', 'mcporter'] }` only when `globalThis.Bun` is present
3. Falls back to `{ cmd: 'npx', leadingArgs: ['--yes', 'mcporter'] }` under Node

The local binary path is preferred so neither bun nor npx is required in PATH when the workspace is intact.

packages/cli/dashboard/src/lib/components/marketplace/McpServersTab.svelte
aaf2tbz: Addressed in the PR description update. The UI change is additive to the existing installed server card: a `GEN CLI` / `REGEN CLI` button appears in the `catalog-actions` div alongside the existing `REMOVE` button (stdio servers only), and a `cli ready` badge appears in `catalog-meta`. No new layout components or styles are introduced — it follows the exact same Button variant and badge pattern as the rest of the card. A screenshot will be added once the dashboard can be run locally against the daemon with mcporter installed.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1106,148 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+	for (const depth of [5, 4, 3]) {
+		const prefix = "../".repeat(depth);
+		try {
+			const binPath = join(new URL(prefix, import.meta.url).pathname, "node_modules", ".bin", "mcporter");
+			if (existsSync(binPath)) {
+				return { cmd: binPath, leadingArgs: [] };
+			}
+		} catch {
+			// Ignore URL resolution errors
+		}
```
PR-Reviewer-Ant, line 1209: The generated binary path is derived from `server.name`, not from a unique server key. Two installed servers whose names sanitize to the same slug (for example `Context7` and another `Context 7`) will both compile to the same `~/.agents/mcp-bins/<name>` path and overwrite each other. This is data corruption and makes the persisted `cliPath` ambiguous. Use a collision-free identifier or detect/refuse conflicting output paths.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1106,148 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+// CLI Generation via mcporter
+// ---------------------------------------------------------------------------
+
+/** Convert an installed server config to the mcporter servers-file format. */
+function buildMcporterConfigEntry(server: InstalledMarketplaceMcpServer): Record<string, unknown> {
+	const { config } = server;
+	if (config.transport === "stdio") {
+		const entry: Record<string, unknown> = {
+			command: config.command,
+			args: config.args,
```
PR-Reviewer-Ant, line 1147: This writes `config.env` into the mcporter config without resolving `secret://...` references. The daemon runtime resolves those through `resolveSecretReferences()` before launching stdio servers, but the generated standalone binary will receive the literal secret reference strings instead of actual credentials. That breaks servers that depend on secrets and diverges from the PR description's claim that the binary embeds the installed server's env config and runs standalone.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in ccd1adf5. Binary name is now derived from `server.id` (unique per installed server) instead of `server.name`, so two servers with similar display names (e.g. "Context7" and "Context 7") get distinct binaries.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in ccd1adf5. `buildMcporterConfigEntry()` is now async and calls `resolveSecretReferences()` on `config.env` (stdio) and `config.headers` (http) before writing the temp config, so the compiled binary receives actual credential values. This matches the resolution path used by `withConnectedClient()` for runtime connections.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1106,150 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+
+/** Convert an installed server config to the mcporter servers-file format.
+ *  Resolves secret:// references so the compiled binary receives real credentials. */
+async function buildMcporterConfigEntry(server: InstalledMarketplaceMcpServer): Promise<Record<string, unknown>> {
+	const { config } = server;
+	if (config.transport === "stdio") {
+		const entry: Record<string, unknown> = {
+			command: config.command,
+			args: config.args,
+		};
```
PR-Reviewer-Ant, line 1148: Blocking: this feature resolves `secret://` references into literal env/header values before handing the config to `mcporter`, and the PR description explicitly says the compiled binary embeds that config. That turns secret references into recoverable credentials at rest inside `~/.agents/mcp-bins/*`, which is a material security regression compared with the current daemon-mediated secret lookup model.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,6 +1758,36 @@ export function mountMarketplaceRoutes(app: Hono): void {
 		return c.json({ success: true, server: updated });
```
PR-Reviewer-Ant, line 1820: Warning: this adds new daemon behavior and a new API surface, but the PR does not include the corresponding `packages/daemon-rs/` parity update required by the repo's shadow-runtime rule. With `shadowEnabled: true`, this is likely to create persistent divergence between the JS and Rust daemons.

docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md
```diff
@@ -183,6 +183,21 @@ the session timeline.
```
PR-Reviewer-Ant, line 183: Warning: the PR checklist says regression tests were added and the implementation is documented as shipped here, but the diff contains no tests for the new `generate-cli` endpoint, persistence of `cliPath`, or failure cases (invalid server, non-stdio server, mcporter failure, secret-bearing configs). That is a mismatch with both the PR description and the repo's incident/guardrail expectations.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `ccd1adf5`

[Automated Review] Found a blocking secret-handling regression, plus one shadow-runtime parity issue and one PR-description/test-plan mismatch.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new daemon route and helper code are fully present in the diff. `buildMcporterConfigEntry()` explicitly resolves `secret://` values before compiling the standalone binary, which makes the secret material reachable from the generated artifact itself. The repo context also explicitly requires daemon behavioral changes to be mirrored in `packages/daemon-rs/`, but no corresponding Rust changes appear in this PR. Finally, the PR checklist claims regression tests were added, yet the diff contains no new test coverage for the new endpoint or persistence path.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in ef5c979e. Reverted the secret resolution approach entirely. `buildMcporterConfigEntry()` now checks for `secret://` prefixes via `hasSecretReferences()` and returns `null` when any are found. The endpoint returns 500 with a clear error: "Server config contains secret:// references which cannot be embedded in a compiled binary. Remove secret references from the server's env/headers before generating a CLI." No credentials are ever written to disk.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Acknowledged. The Rust daemon does not currently implement marketplace routes at all (`packages/daemon-rs/crates/signet-daemon/src/routes/marketplace.rs` is the catalog/install surface, but it doesn't have the generate-cli concept). This PR adds a JS-only endpoint. Parity tracking can be added to the Rust daemon when the feature stabilizes, but blocking this PR on a Rust implementation of an experimental feature seems premature. Happy to add a tracking issue if preferred.

docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md
aaf2tbz: Fixed in ef5c979e. Added 3 tests:
- `POST /:id/generate-cli` returns 404 for unknown server
- `POST /:id/generate-cli` returns 400 for HTTP servers (only stdio supported)
- `POST /:id/generate-cli` returns 500 with descriptive error when config contains `secret://` references

All 8 tests pass (5 existing + 3 new). The mcporter binary spawn itself isn't tested in-process since it requires the external mcporter package, but the validation and rejection paths are fully covered.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,6 +1777,36 @@ export function mountMarketplaceRoutes(app: Hono): void {
...
+		logger.info("skills", "Generating standalone CLI binary", { serverId: id, name: server.name });
+		const result = await runMcporterGenerateCli(server);
+
+		if (!result.success) {
+			logger.warn("skills", "CLI generation failed", { serverId: id, error: result.error });
+			return c.json({ success: false, error: result.error }, 500);
+		}
+
+		// Persist the binary path on the server record
+		const updated: InstalledMarketplaceMcpServer = {
```
PR-Reviewer-Ant: Blocking: `cliPath` is now persisted as durable server state, but this PR does not invalidate or remove it when the server config later changes or when the server is uninstalled. Because the compiled binary embeds `command`/`args`/`env`/`cwd`, editing an installed stdio server after generation leaves a stale executable on disk while the dashboard still shows `cli ready`. That can cause operators and older tooling to run an out-of-date binary with the wrong command or env. The lifecycle needs to clear/regenerate `cliPath` on config mutation and remove the compiled binary on uninstall.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1106,169 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+}
+
+/**
+ * Sanitize a server name for use as a filesystem binary name.
+ * Reuses the same character set as sanitizeServerId so names are consistent.
+ * Returns null when the name cannot be reduced to a valid identifier.
+ */
+function sanitizeBinaryName(name: string): string | null {
+	const sanitized = sanitizeServerId(name);
+	// Reject anything that looks like a path after sanitization
```
PR-Reviewer-Ant, line 1183: Warning: `getMcporterSpawnTarget()` builds the local binary path from `new URL(...).pathname`. For `file:` URLs this is not a portable filesystem path: on Windows it becomes `/C:/...`, and percent-encoded characters are not decoded. That means the daemon can fail to find the bundled `node_modules/.bin/mcporter` even when it is installed, falling back to `npx --yes mcporter` unexpectedly. Use a proper file-URL-to-path conversion here.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 047625ac. PATCH handler now clears `cliPath` and deletes the stale binary when `config` changes. DELETE handler removes the compiled binary from disk before writing the updated server list. Both use `try { unlinkSync(...) } catch {}` since the binary may already be gone.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 047625ac. Added `import { fileURLToPath } from "node:url"` at module scope and replaced `new URL(...).pathname` with `fileURLToPath(new URL(...))`. This correctly handles Windows drive-letter paths and percent-encoded characters.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1588,11 +1760,25 @@ export function mountMarketplaceRoutes(app: Hono): void {
 			return c.json({ error: "Invalid config" }, 400);
 		}
 
+		// When config changes, the compiled CLI binary is stale — clear cliPath
+		// and remove the old binary so dashboard no longer shows "cli ready".
+		const configChanged = normalized !== null;
+		let nextCliPath = existing.cliPath;
+		if (configChanged && existing.cliPath) {
```
PR-Reviewer-Ant, line 1799: This route deletes `existing.cliPath` without constraining it to `~/.agents/mcp-bins/`. Because `parseInstalledServer()` now trusts any persisted `cliPath` string from `mcp-servers.json`, a crafted or manually edited record can make PATCH/DELETE unlink an arbitrary file accessible to the daemon user. Please validate/canonicalize `cliPath` on read and before deletion so only files under the managed bins directory can be removed.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1107,169 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+	return new Promise<GenerateCliResult>((resolve) => {
+		const proc = spawn(cmd, [...leadingArgs, ...mcporterArgs], {
+			stdio: "pipe",
+			windowsHide: true,
+			timeout: 120_000,
+		});
+
+		let stderr = "";
+		proc.stderr?.on("data", (chunk: Buffer) => {
+			stderr += chunk.toString();
```
PR-Reviewer-Ant, line 1267: The success path is hardcoded as `join(binsDir, binaryName)`. On Windows, compiled executables are typically emitted as `<name>.exe`, so the persisted `cliPath` will point at a non-existent file, the dashboard will show a false "cli ready" state, and later cleanup/regeneration will target the wrong path. The code should detect the actual emitted artifact instead of assuming a suffix-less filename.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in e4c2902a. Added `isUnderMcpBins()` which `resolve()`s both the bins dir and the candidate path, then checks the resolved path starts with `${binsDir}/` (or `\\` on Windows). All deletion now goes through `removeManagedCliBinary()` which no-ops if the path isn't under the managed directory. A crafted `cliPath` in `mcp-servers.json` pointing outside `~/.agents/mcp-bins/` will be silently ignored on deletion.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in e4c2902a. After mcporter exits successfully, the code now checks for both `<name>` and `<name>.exe` and persists whichever actually exists on disk. Falls back to the suffix-less path if neither is found (shouldn't happen on success, but avoids a false negative).

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,12 +1809,47 @@ export function mountMarketplaceRoutes(app: Hono): void {
 		return c.json({ success: true, server: updated });
```
PR-Reviewer-Ant, line 1820: This endpoint compiles into a shared output path (`~/.agents/mcp-bins/<id>`) but has no single-flight guard or rate limiting. Two concurrent POSTs for the same server can run `mcporter` against the same target at once, which is exactly the kind of expensive mutation endpoint the repo guardrails call out. At minimum this needs per-server serialization or a 409/429 while generation is already in progress.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -804,6 +829,7 @@ function parseInstalledServer(value: unknown): InstalledMarketplaceMcpServer | n
 		enabled: value.enabled,
 		scope: normalizeScope(value.scope),
 		config,
+		cliPath: typeof value.cliPath === "string" ? value.cliPath : undefined,
```
PR-Reviewer-Ant: The PR description says the dashboard shows `cli ready` once a binary exists, but on restart you restore `cliPath` from JSON without checking whether that file is still present. If the binary is deleted manually or generation only partially succeeded, the installed record will still advertise a ready CLI forever. `parseInstalledServer` or the GET path should clear/stale-check `cliPath` based on actual filesystem state before returning it.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `e4c2902a`

[Automated Review] Found two correctness issues in the new MCP CLI generation flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new `generate-cli` route and dashboard badge logic are fully present in the diff. One issue is directly visible in the route implementation: generation is an expensive file-writing operation with no single-flight or rate limiting, so concurrent POSTs can compile the same server into the same output path simultaneously. The other issue is visible from persisted `cliPath` handling: the code trusts stored JSON on reload and never revalidates file existence before reporting `cli ready`. I did not run mcporter itself, so I cannot show a concrete corrupted binary artifact.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in dd0965f0. Added `cliGenerationInFlight` Set — the handler checks `has(id)` before spawning and returns 409 ("CLI generation already in progress for this server"). The ID is added before the spawn and removed in a `finally` block so it's always cleaned up.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in dd0965f0. `parseInstalledServer()` now checks `existsSync(value.cliPath)` before accepting the field — if the binary was manually deleted or generation only partially succeeded, the field is silently dropped to `undefined`. Dashboard will correctly show no "cli ready" badge in that case.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1124,176 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+			if (existsSync(binPath)) {
+				return { cmd: binPath, leadingArgs: [] };
+			}
+		} catch {
+			// Ignore URL resolution errors
+		}
+	}
+	// Runtime fallback
+	const isBun = typeof (globalThis as Record<string, unknown>).Bun !== "undefined";
+	return isBun
```
PR-Reviewer-Ant, line 1213: Blocking security issue: this flow only rejects `secret://` placeholders, but it still serializes any literal `env`/`headers` values into `tmpConfig` and then embeds them into the compiled binary. That contradicts the PR's security claim that credentials are never written to disk, and it creates extra at-rest copies under `/tmp` and `~/.agents/mcp-bins`. If inline secrets are allowed in server configs today, this endpoint expands their exposure significantly.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,12 +1812,55 @@ export function mountMarketplaceRoutes(app: Hono): void {
...
+
+			// Persist the binary path on the server record
+			const updated: InstalledMarketplaceMcpServer = {
+				...server,
+				cliPath: result.path,
+				updatedAt: new Date().toISOString(),
+			};
+			writeInstalledServers(installed.map((s) => (s.id === id ? updated : s)));
+			logger.info("skills", "CLI binary generated", { serverId: id, path: result.path });
+			return c.json({ success: true, path: result.path, server: updated });
```
PR-Reviewer-Ant: Blocking race/data-loss bug: `installed` and `server` are captured before `runMcporterGenerateCli(server)` runs, but the compile can take up to 120 seconds. When it finishes, `writeInstalledServers(installed.map(...))` rewrites that old snapshot, so any concurrent PATCH/uninstall changes to this server made during compilation are silently lost. Re-read the latest installed list after compilation and merge only the new `cliPath` into the current record.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 5b48606c. Three mitigations:
1. tmpConfig is now written with `mode: 0o600` so only the daemon user can read it, and it's always cleaned up in the close/error handlers
2. The response now includes a `warning` field when `hasInlineCredentials` is true, explicitly telling the caller the binary embeds inline env/header values and should be treated as sensitive
3. `secret://` references are still rejected outright — this only affects literal env values the user put in the config themselves

The compiled binary embedding the server's command+args+env is the fundamental purpose of generate-cli — there's no way to avoid it without making the binary non-standalone. The warning makes the tradeoff explicit.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 5b48606c. After `runMcporterGenerateCli()` returns, we now `readInstalledServers()` fresh and merge only `cliPath` + `updatedAt` into the current record. If the server was uninstalled during compilation, the binary is cleaned up and the endpoint returns 409.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,12 +1820,66 @@ export function mountMarketplaceRoutes(app: Hono): void {
...
+
+			if (!result.success) {
+				logger.warn("skills", "CLI generation failed", { serverId: id, error: result.error });
+				return c.json({ success: false, error: result.error }, 500);
+			}
+
+			// Re-read the latest state — compilation can take up to 120s, so the
+			// snapshot captured before the spawn may be stale from concurrent edits.
+			const freshInstalled = readInstalledServers();
+			const freshServer = freshInstalled.find((s) => s.id === id);
```
PR-Reviewer-Ant, line 1856: The route re-reads the installed record after compilation, but it never verifies that the server config is still the same config that was compiled. If a user edits the MCP server while `mcporter` is running, this code will persist `cliPath` for a binary built from stale command/args/env data and mark it as current. That contradicts the PR description ('embeds the server's command/args/env configuration') and creates a real race with the PATCH handler, which explicitly treats config changes as making the CLI stale. You need a version/config check before writing `cliPath`, otherwise regenerate-on-edit can leave the UI claiming a fresh CLI that no longer matches the saved server config.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1098,6 +1124,184 @@ function rankMarketplaceTools(tools: readonly MarketplaceMcpTool[], query: strin
...
+			}
+		} catch {
+			// Ignore URL resolution errors
+		}
+	}
+	// Runtime fallback
+	const isBun = typeof (globalThis as Record<string, unknown>).Bun !== "undefined";
+	return isBun
+		? { cmd: process.execPath, leadingArgs: ["x", "mcporter"] }
+		: { cmd: "npx", leadingArgs: ["--yes", "mcporter"] };
```
PR-Reviewer-Ant, line 1215: The temporary config filename uses raw `server.id` in `join(tmpdir(), `mcporter-${server.id}-${Date.now()}.json`)`. `server.id` is sanitized for the output binary name, but not for this temp path. A crafted installed record containing path separators (for example from a hand-edited `mcp-servers.json` or legacy bad data) can escape the temp dir and cause writes/unlinks outside `tmpdir()`. This should use the sanitized identifier, or generate the temp file via a safe temp-file API.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 12dc22e1. Config is fingerprinted via `JSON.stringify(server.config)` before spawning mcporter. After compilation, the fresh record's config is compared — if it changed, the binary is deleted and the endpoint returns 409 ("Server config changed during CLI generation. Please regenerate."). This is consistent with the PATCH handler which already clears cliPath on config changes.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 12dc22e1. tmpConfig filename now uses the already-sanitized `binaryName` (from `sanitizeBinaryName(server.id)`) instead of raw `server.id`, so crafted IDs with path separators can't escape tmpdir.

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,12 +1820,74 @@ export function mountMarketplaceRoutes(app: Hono): void {
 		return c.json({ success: true, server: updated });
 	});
 
+	app.post("/api/marketplace/mcp/:id/generate-cli", async (c) => {
+		const id = c.req.param("id");
```
PR-Reviewer-Ant, line 1824: The new `generate-cli` route does not check `server.enabled` before compiling. That means a server that an operator has explicitly disabled can still be exported as a standalone executable, and that binary then runs outside the daemon's enable/scope/policy controls. This undermines the existing disable mechanism and expands the attack surface for any sensitive server config.

packages/cli/dashboard/src/lib/stores/marketplace-mcp.svelte.ts
```diff
@@ -223,3 +225,20 @@ export async function removeMarketplaceMcpServer(id: string): Promise<void> {
 		mcpMarket.removingId = null;
 	}
 }
+
+export async function doGenerateMcpCli(id: string): Promise<void> {
+	mcpMarket.generatingCliId = id;
+	try {
+		const result = await generateMarketplaceMcpCli(id);
```
PR-Reviewer-Ant, line 232: The daemon now returns a `warning` when the compiled binary embeds inline env/header values, but the new dashboard flow discards that field and always shows a plain success toast. From the main UI path, users can therefore click `GEN CLI` and silently write API keys or bearer tokens into a local executable without any warning. If this feature is kept, the UI needs to surface the warning prominently or block generation unless the user explicitly confirms.

packages/daemon/src/routes/marketplace.ts
aaf2tbz: Fixed in 8d53e0da. The endpoint now checks `server.enabled` and returns 400 ("Cannot generate CLI for a disabled server") before proceeding. A disabled server can't be compiled into a standalone binary that bypasses the daemon's controls.

packages/cli/dashboard/src/lib/stores/marketplace-mcp.svelte.ts
aaf2tbz: Fixed in 8d53e0da. `doGenerateMcpCli()` now checks for the `warning` field in the response and shows a `warn`-level toast with the full message when present. Users will see: "CLI binary ready: <path>. Warning: The compiled binary embeds inline env/header values from the server config. Treat the binary as sensitive."

packages/daemon/src/routes/marketplace.ts
```diff
@@ -1606,12 +1820,77 @@ export function mountMarketplaceRoutes(app: Hono): void {
 		return c.json({ success: true, server: updated });
 	});
 
+	app.post("/api/marketplace/mcp/:id/generate-cli", async (c) => {
```
PR-Reviewer-Ant, line 1823: This adds a new mutation endpoint that spawns `mcporter`, compiles code, and writes executables to disk, but there is no explicit permission check or rate limiting in the route itself. That expands the daemon's attack surface in exactly the area called out by the repo guardrails (`Admin/refresh/mutation endpoints must have explicit permission checks`, plus rate limiting for expensive paths). If the daemon is exposed beyond localhost or auth is disabled/misconfigured, an attacker can trigger repeated compilation jobs and fill `~/.agents/mcp-bins` / consume CPU. Please gate this route behind the same authz pattern used for other operational mutations and add abuse controls.

docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md
PR-Reviewer-Ant, line 183: The PR description says spec alignment/readiness was validated, but the only spec change here is appending an 'Implemented' section to a spec that still has `status: planning`. The repo protocol requires keeping `docs/specs/INDEX.md` and `docs/specs/dependencies.yaml` in sync when implemented behavior ships. As written, later agents will still read this work as planning-only, which is a correctness/maintainability problem for follow-on changes.

aaf2tbz:
Closing — will reopen with full spec implementation (CLI bridge, usage analytics, and generate-cli) in a clean PR.

PR #408, opened 2026-03-30
feat: skill usage analytics — invocation tracking, API, CLI, and constellation dashboard

aaf2tbz:
## Summary

Adds skill invocation tracking and analytics, mirroring the MCP analytics system from PR #407:

- **Migration 053**: `skill_invocations` table (skill_name, agent_id, source, task_id, latency, success)
- **Scheduler instrumentation**: records invocations when scheduled tasks with `skill_name` complete; populates the previously-unused `skill_meta.use_count` and `last_used_at` columns. Uses `'default'` agent_id since `scheduled_tasks` has no agent_id column — **Phase 2** will add `agent_id` to `scheduled_tasks` for multi-agent support.
- **Analytics API**: `GET /api/skills/analytics` (aggregated) and `GET /api/skills/analytics/:skill` (per-skill with timeline) — auth-aware agent scoping via `resolveScopedAgent()`
- **CLI**: `signet skill analytics [--skill <name>] [--since <iso>] [--json]`
- **Dashboard**: constellation-style `SkillUsagePanel` with side-by-side panes matching MCP panel design; store and API client support `agentId` param
- **Marketplace write path**: auth-aware agent scoping via `resolveScopedAgent()` on invocation recording; MCP proxy omits `x-signet-agent-id` (workspace-level, no per-agent context — Phase 2)

### Known Phase 1 limitations (documented, not bugs)

- Scheduler writes `agent_id = 'default'` because `scheduled_tasks` table has no agent_id column. Multi-agent attribution requires a schema migration (Phase 2).
- MCP proxy is workspace-level and does not propagate agent context. Phase 2: per-agent MCP sessions.
- `MCP_INSTALLS_ENABLED = true` is inherited from the PR #407 base branch (MCP CLI bridge), not a change in this PR.

## Changes

| Commit | What |
|--------|------|
| 1 | Migration 053: `skill_invocations` table + indexes + test |
| 2 | Instrument scheduler `worker.ts` — record invocations, update `skill_meta` |
| 3 | Analytics API: `skill-analytics.ts` + 5 tests + mount in daemon |
| 4 | CLI: `signet skill analytics` subcommand |
| 5 | Dashboard: `SkillUsagePanel.svelte`, store, API types, embed in SkillsTab |
| 6 | Fix: mount analytics before skills routes (`:name` catch-all conflict) |
| 7 | Spec: `skill-usage-analytics.md` + INDEX.md + dependencies.yaml |
| 8 | Fix: auth-aware agent scoping on analytics read endpoints via `resolveScopedAgent()` |
| 9 | Fix: revert to explicit `'default'` agent_id (agent.yaml name != DB tenant key), add dependencies.yaml entry |
| 10 | Fix: auth-aware agent scoping on marketplace write path, remove hardcoded agent-id from MCP proxy, sync spec status |
| 11 | Fix: normalize timestamps with `datetime()` in scheduler INSERTs, forward agentId in dashboard stores |

## Type

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`

## Screenshots

The SkillUsagePanel uses the same constellation design as the MCP analytics panel: left pane shows total runs and an SVG node graph with skill nodes sized by usage count, right pane shows success rate, p50/p95 latency, and sorted usage bars. Nodes match bar colors (green for healthy, red for failing). Tested with 19 seeded invocations across 7 installed skills.

Screenshot 2026-03-30 at 12 29 14 AM

## PR Readiness (MANDATORY)

Migration 053 adds the `skill_invocations` table with `CREATE TABLE IF NOT EXISTS` — fully idempotent. No existing tables or columns are modified.

**Rollback**: Safe to drop `skill_invocations` table. Analytics API returns empty results gracefully.

- [x] Migration is idempotent

**Rust parity**: The Rust daemon does not implement skill routes. JS-only feature. N/A.

## Testing

- [x] `bun test` passes (28 migration + 5 analytics tests)
- [x] `bun run lint` passes
- [ ] N/A

## Test plan

- [x] Start daemon -> migration 053 applies
- [x] `curl /api/skills/analytics` -> aggregated stats with auth-aware agent scoping
- [x] Dashboard Skills installed tab -> constellation panel with live data
- [x] Panel shows "Building skill analytics..." when no data
- [x] Route ordering: `/api/skills/analytics` resolves before `/:name` catch-all
- [x] Analytics read endpoints use `resolveScopedAgent()` — cross-agent data isolation enforced
- [x] Marketplace write path uses `resolveScopedAgent()` — prevents analytics poisoning
- [x] Scheduler records invocations with `datetime()` normalization for consistent timestamps
- [x] `skill_meta` UPDATE scoped to `agent_id = 'default'` entities
- [x] `dependencies.yaml` updated with `skill-usage-analytics` entry (status: approved)

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -7587,13 +7587,10 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
 	const taskSkillMode = typeof task.skill_mode === "string" ? task.skill_mode : null;
 	const taskWorkingDir = typeof task.working_directory === "string" ? task.working_directory : null;
-	const scoped = resolveScopedAgent(
-		c.get("auth")?.claims ?? null,
-		authConfig.mode,
-		c.req.query("agent_id"),
-	);
-	if (scoped.error) return c.json({ error: scoped.error }, 403);
-	const taskAgentId = scoped.agentId;
+	// Attribution uses 'default' — scheduled_tasks has no agent_id column,
+	// so neither trigger path (scheduler or API) can derive the task's owning
+	// agent. Phase 2 adds agent_id to scheduled_tasks for proper attribution.
+	const taskAgentId = "default";
 
 	// Resolve skill content into prompt
 	const effectivePrompt = resolveSkillPrompt(taskPrompt, taskSkillName, taskSkillMode);


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: skill usage analytics — invocation tracking, API, CLI, and constellation dashboard` (commit `d1958a07`) now and I'll follow up shortly with feedback.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +282,39 @@ export async function executeTask(
 		timestamp: new Date().toISOString(),
 	});
 
+	// Record skill invocation if this task uses a skill
+	if (task.skill_name) {
+		const latencyMs = new Date(completedAt).getTime() - new Date(now).getTime();
+		const success = status === "completed";
+		try {
+			db.withWriteTx((wdb) => {
+				wdb
```
PR-Reviewer-Ant, line 322: Blocking: scheduled skill usage is always recorded with `agent_id = 'default'`. In a multi-agent workspace this misattributes every invocation to the default agent, so `/api/skills/analytics` returns incorrect data and the PR description's agent-scoping claim is not met.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +282,39 @@ export async function executeTask(
...
+						task.skill_name,
+						task.id,
+						latencyMs,
+						success ? 1 : 0,
+						result.error ?? null,
+						completedAt,
+					);
+				// Update skill_meta usage counters
+				wdb
+					.prepare(
```
PR-Reviewer-Ant, line 338: Blocking: the `skill_meta` update matches by canonical skill name only and does not scope by agent. If multiple agents have the same skill entity, one scheduled run will increment usage counters for all of them, corrupting per-agent `use_count`/`last_used_at` state.

packages/cli/dashboard/src/lib/api.ts
```diff
@@ -2575,3 +2574,98 @@ export async function installMcp(options: InstallMcpOptions): Promise<InstallMcp
...
+	return response.json();
+}
+
+// ---------------------------------------------------------------------------
+// Skill Analytics
+// ---------------------------------------------------------------------------
+
+export interface SkillUsageStats {
+	skillName: string;
+	count: number;
```
PR-Reviewer-Ant, line 2651: Warning: the dashboard client for skill analytics does not accept or send `agent_id`, even though the new API is agent-scoped. In multi-agent installs the panel will always show the default-agent view, which diverges from the PR's stated scoping behavior.

packages/daemon/src/routes/mcp-analytics.ts
```diff
@@ -0,0 +1,209 @@
...
+	if (!Number.isFinite(n)) return fallback;
+	return Math.max(min, Math.min(max, Math.round(n)));
+}
+
+function computePercentile(sorted: readonly number[], p: number): number {
+	if (sorted.length === 0) return 0;
+	const idx = Math.ceil((p / 100) * sorted.length) - 1;
+	return sorted[Math.max(0, idx)] ?? 0;
+}
+
```
PR-Reviewer-Ant, line 64: This endpoint trusts `agent_id` from the query string and defaults to `default`, instead of using the auth-aware scoped-agent resolution used by the skill analytics routes. In authenticated or multi-agent deployments, a caller can request another agent's analytics simply by changing `?agent_id=...`, which violates the repo's agent-scoping invariant and exposes per-agent operational data.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +301,46 @@ export async function executeTask(
 		timestamp: new Date().toISOString(),
 	});
 
+	// Record skill invocation if this task uses a skill
+	// Note: scheduled_tasks does not have agent_id yet — resolve from
```
PR-Reviewer-Ant: Scheduled skill invocations are attributed via a process-wide cached value derived from the root `agent.yaml` name, not from the task being executed. In a multi-agent workspace this will write every scheduled run into `skill_invocations` and `skill_meta.use_count` for the same agent, corrupting analytics and usage counters for any other agent's tasks. The PR description says agent scoping was verified, but this implementation does not preserve agent ownership.

packages/cli/dashboard/src/lib/components/marketplace/McpServersTab.svelte
```diff
@@ -71,22 +71,18 @@ const installedCatalogIds = $derived(
...
 );
 let installSheetOpen = $state(false);
 let selectedCatalogEntry = $state<MarketplaceMcpCatalogEntry | null>(null);
 let detailOpen = $state(false);
 let detailItem = $state<McpDetailItem | null>(null);
 let view = $state<"browse" | "installed">("browse");
-const MCP_INSTALLS_ENABLED = false;
+const MCP_INSTALLS_ENABLED = true;
 const activeSourceLabel = $derived(mcpMarket.source === "all" ? "All sources" : formatSourceLabel(mcpMarket.source));
 const activeSortLabel = $derived.by(() => {
```
PR-Reviewer-Ant, line 89: This flips `MCP_INSTALLS_ENABLED` from `false` to `true`, which exposes MCP install UI as part of a PR described as skill analytics only. That is an unrelated behavior change with user-visible impact, and it bypasses the stated scope in the PR description/spec notes. If this is intentional it needs to be split out or explicitly justified and tested; otherwise it is a likely accidental regression.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -5,23 +5,42 @@
...
 import { join } from "node:path";
+import type { TaskHarness } from "@signet/core";
 import type { DbAccessor, ReadDb } from "../db-accessor";
+import { logger } from "../logger";
 import { loadMemoryConfig } from "../memory-config";
 import type { WorkerHandle } from "../pipeline/worker";
 import { computeNextRun } from "./cron";
 import { resolveSkillPrompt } from "./skill-resolver";
-import { spawnTask, type SpawnResult } from "./spawn";
+import { type SpawnResult, spawnTask } from "./spawn";
```
PR-Reviewer-Ant, line 17: The new scheduler scoping resolves `agent_id` from `agent.yaml`'s `name:` field and caches it process-wide, then uses that value for both `skill_invocations.agent_id` and the `skill_meta` update. In this repo the canonical tenant key is the agent ID/roster key, not a human-facing display name. If `name:` differs from the stored `agent_id` (for example spaces/casing/custom label), analytics rows will be written under a different tenant and the `skill_meta` UPDATE will silently miss the real skill row. This breaks the PR’s claim that usage counters are updated with correct agent scoping.

packages/cli/dashboard/src/lib/stores/skill-analytics.svelte.ts
```diff
@@ -0,0 +1,23 @@
+import { type SkillAnalyticsSummary, getSkillAnalytics } from "$lib/api";
+
+export const skillAnalytics = $state({
+	data: null as SkillAnalyticsSummary | null,
+	loading: false,
+	error: null as string | null,
+});
+
+export async function fetchSkillAnalytics(params?: {
```
PR-Reviewer-Ant, line 9: The PR description says the dashboard client now supports `agentId`, but the new skill analytics store only accepts `{ skill, since }` and `SkillUsagePanel` always calls `fetchSkillAnalytics()` with no agent parameter. The same pattern exists for MCP analytics. In a multi-agent dashboard view this panel will keep querying the default/current agent instead of the selected one, so the implementation does not match the stated goal of agent-aware analytics in the client.

packages/daemon/src/scheduler/worker.ts
PR-Reviewer-Ant, line 17: Blocking: this resolves analytics `agent_id` from the root `agent.yaml` `name:` field, which is a display name, not the stable tenant key used for DB scoping. In multi-agent installs that can include spaces/casing and need not match any real `agent_id`, so `skill_invocations` rows can be written under the wrong tenant and the `skill_meta` UPDATE below will miss entirely. The PR description claims correct agent resolution, but this implementation still breaks the repository's agent-scoping contract.

packages/daemon/src/mcp/tools.ts
```diff
@@ -1314,6 +1316,10 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
 					args: args ?? {},
 				},
 				timeout: 60_000,
```
PR-Reviewer-Ant, line 1318: Blocking: agent-originated MCP calls are still stamped with `x-signet-agent-id: "default"`. That hardcodes all agent usage into the default tenant and violates the repo rule against hardcoding scoped data paths. In a multi-agent workspace, MCP analytics will be wrong and cross-agent isolation is lost for invocation tracking.

docs/specs/INDEX.md
```diff
@@ -694,7 +694,8 @@ Legend:
 | `remember-recall-skill-parity` | planning | `docs/specs/planning/remember-recall-skill-parity.md` | `procedural-memory-plan` | - | Stub: /remember and /recall architecture/schema parity |
 | `rust-daemon-parity-cutover` | planning | `docs/specs/planning/rust-daemon-parity-cutover.md` | `daemon-rust-rewrite`, `memory-pipeline-v2` | - | Stub: rust daemon parity and primary-runtime cutover |
 | `deep-memory-search` | planning | `docs/specs/planning/deep-memory-search.md` | `desire-paths-epic`, `ssm-foundation-evaluation` | - | Stub: optional supermemory-style deep memory escalation |
-| `mcp-cli-bridge-and-usage-analytics` | planning | `docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md` | `signet-runtime` | - | Stub: MCP servers as CLI + dashboard usage tracking (MC Porter reference) |
+| `mcp-cli-bridge-and-usage-analytics` | approved | `docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md` | `signet-runtime` | - | Phase 1: CLI bridge, invocation tracking, analytics API, dashboard panel |
```
PR-Reviewer-Ant, line 697: The PR readiness section says spec alignment was validated with both `INDEX.md` and `dependencies.yaml`, but this diff only updates `INDEX.md` and adds spec files. I do not see the matching `docs/specs/dependencies.yaml` change required by the session protocol, so this introduces spec/dependency drift despite the PR claiming otherwise.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +277,44 @@ export async function executeTask(
...
 
+	// Record skill invocation if this task uses a skill.
+	// scheduled_tasks does not have agent_id — uses "default" for now.
+	// Phase 2: add agent_id column to scheduled_tasks for multi-agent support.
+	if (task.skill_name) {
+		const latencyMs = new Date(completedAt).getTime() - new Date(now).getTime();
+		const success = status === "completed";
+		try {
+			db.withWriteTx((wdb) => {
+				wdb
```
PR-Reviewer-Ant, line 322: Blocking: this does not implement the PR description's claimed agent resolution. Scheduled skill invocations are still recorded with hardcoded `agent_id = 'default'`, and the `skill_meta` update is also scoped to `'default'`. In a multi-agent workspace this misattributes usage, corrupts analytics, and updates the wrong agent's skill metadata.

packages/daemon/src/mcp/tools.ts
PR-Reviewer-Ant, line 1318: Blocking: agent-originated MCP calls are tagged with `x-signet-agent-id: 'default'` unconditionally. That leaks/misattributes analytics across agents and violates the repo's mandatory scoping rule that user-data reads/writes must thread the real `agent_id` rather than hardcoding `default`.

docs/specs/dependencies.yaml
```diff
@@ -703,12 +703,26 @@ specs:
     hard_depends_on:
```
PR-Reviewer-Ant, line 703: The spec metadata is still inconsistent: this file keeps `mcp-cli-bridge-and-usage-analytics` at `status: planning`, while the PR also changes `INDEX.md` and the spec file itself to `approved`. That breaks the repo's required INDEX/dependencies sync contract and means the implementation does not fully satisfy the PR's own spec-alignment claim.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +277,44 @@ export async function executeTask(
...
+					.prepare(
```
PR-Reviewer-Ant, line 323: Blocking: this PR claims scheduler instrumentation resolves `agent_id` from `agent.yaml` and avoids hardcoding, but the inserted `skill_invocations` row still writes `'default'`, and the `skill_meta` update is also scoped to `'default'`. That breaks multi-agent isolation and means scheduled skill usage for non-default agents is silently attributed to the wrong tenant.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,185 @@
...
+		const agentId = scoped.agentId;
+		const skill = c.req.query("skill");
+		const since = c.req.query("since");
+		const limit = clampPositiveInt(c.req.query("limit"), 10, 1, 100);
+
+		try {
+			const result = getDbAccessor().withReadDb((db) => {
+				const conditions: string[] = ["agent_id = ?"];
+				const params: unknown[] = [agentId];
+				if (skill) {
```
PR-Reviewer-Ant, line 75: Blocking: the `since` filter uses `created_at >= datetime(?)`, but this PR also inserts `created_at` values as ISO strings with `T`/`Z` (for example in the scheduler path). SQLite then compares mixed timestamp formats lexically, not semantically, so same-day filters can return incorrect rows. Either normalize writes to SQLite `datetime('now')` format everywhere or compare `datetime(created_at) >= datetime(?)` consistently.

packages/cli/dashboard/src/lib/stores/skill-analytics.svelte.ts
PR-Reviewer-Ant, line 9: Warning: the PR description says the dashboard client supports an `agentId` param, but this store dropped that parameter and only forwards `skill`/`since`. The API client in `api.ts` supports `agentId`, so the dashboard cannot actually request analytics for a selected agent from this code path.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `04fdd32a`

[Automated Review] Found blocking agent-scoping and timestamp-filter regressions, plus one client/PR-description mismatch.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The scheduler instrumentation now writes `skill_invocations.agent_id` and `skill_meta` updates with a hardcoded `'default'`, which directly contradicts both the PR description and the repo's mandatory agent-scoping invariant. The analytics routes also compare `created_at` against `datetime(?)` while this PR writes ISO timestamps like `2026-03-30T...Z` into `created_at`, so same-day `since` filters can be wrong by simple lexical comparison. These issues are directly provable from the changed SQL and insert paths in this diff.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +277,44 @@ export async function executeTask(
 		timestamp: new Date().toISOString(),
 	});
 
+	// Record skill invocation if this task uses a skill.
+	// scheduled_tasks does not have agent_id — uses "default" for now.
```
PR-Reviewer-Ant: This PR ships the dashboard/API/CLI as 'skill usage analytics', but the only new write path in the diff is this scheduler-only insertion. Nothing here records slash-command, CLI, or API skill invocations, so the analytics surfaces will silently undercount most real skill usage and present partial data as if it were complete. Either narrow the feature/UX wording to scheduled-task analytics only, or add the missing write paths before shipping.

packages/daemon/src/scheduler/worker.ts
PR-Reviewer-Ant, line 323: These writes are hardcoded to `agent_id = 'default'`, and the subsequent `skill_meta` update is also pinned to the default agent. In a multi-agent workspace, any scheduled task associated with another agent will be attributed to the wrong tenant, corrupting both `skill_invocations` and per-agent `skill_meta.use_count` / `last_used_at`. Documenting this as 'Phase 2' does not prevent the data from being wrong for existing multi-agent installs.

packages/cli/dashboard/src/lib/components/marketplace/McpUsagePanel.svelte
```diff
@@ -0,0 +1,334 @@
+<script lang="ts">
+import { fetchMcpAnalytics, mcpAnalytics } from "$lib/stores/mcp-analytics.svelte";
+import { onMount } from "svelte";
+
+onMount(() => {
```
PR-Reviewer-Ant: This panel always fetches analytics without passing an `agentId`. The PR description says the dashboard/store path now forwards agent context, but this component drops it and will show the auth-default/current scope instead of the agent currently selected in the UI. In multi-agent installs that produces incorrect analytics and violates the repo's agent-scoping expectations.

packages/cli/dashboard/src/lib/components/skills/SkillUsagePanel.svelte
```diff
@@ -0,0 +1,223 @@
+<script lang="ts">
+import { fetchSkillAnalytics, skillAnalytics } from "$lib/stores/skill-analytics.svelte";
+import { onMount } from "svelte";
+
+onMount(() => {
```
PR-Reviewer-Ant: Same scoping problem as the MCP panel: the new skill analytics panel fetches with no `agentId`, so it cannot follow the dashboard's selected agent. The API/store were extended for agent-aware reads, but this component never uses that path, which means the panel can display another agent's stats or only the default tenant in multi-agent mode.

packages/cli/src/commands/skill.ts
```diff
@@ -270,4 +270,65 @@ export function registerSkillCommands(program: Command, deps: SkillDeps): void {
...
+		.description("Show skill usage analytics")
+		.option("--skill <name>", "Filter by skill name")
+		.option("--since <iso>", "Only include invocations after this ISO date")
+		.option("--json", "Output as JSON")
+		.action(async (options: { skill?: string; since?: string; json?: boolean }) => {
+			if (!(await deps.isDaemonRunning())) {
+				console.error(chalk.red("Daemon is not running. Start it with: signet daemon start"));
+				return;
+			}
+
```
PR-Reviewer-Ant, line 286: The PR description claims `GET /api/skills/analytics/:skill` provides per-skill analytics with a timeline, and the CLI advertises `--skill <name>`. But the CLI always calls the aggregate `/api/skills/analytics` endpoint with a query param and never uses the detailed `/:skill` route, so the promised per-skill breakdown/timeline is not actually exposed here.

PR-Reviewer-Ant (dismissed):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `40e89615`

[Automated Review] Found two blocking multi-agent scoping regressions in the new dashboard analytics panels and one implementation/PR-description mismatch in the skill analytics CLI.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The dashboard panels in the diff call `fetchMcpAnalytics()` and `fetchSkillAnalytics()` with no `agentId`, even though the new API/store surface explicitly added `agentId` support and the PR description claims agent-aware dashboard forwarding. That makes the new analytics UI ignore the selected agent context in multi-agent installs. The skill CLI similarly always calls the aggregate `/api/skills/analytics` route, so the claimed per-skill detailed view with timeline is not actually exposed there. I did not have the surrounding dashboard selection store in the provided diff, so confidence is medium rather than high.

packages/cli/dashboard/src/lib/components/skills/SkillUsagePanel.svelte
PR-Reviewer-Ant: Blocking: this panel always calls `fetchSkillAnalytics()` with no `agentId`, so it ignores the selected dashboard agent context even though the store/API were extended to support it. In a multi-agent workspace this will show the wrong tenant's usage data and breaks the PR's stated "auth-aware agent scoping" / "forward agentId in dashboard stores" claim.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,195 @@
...
+
+				const sources = db
+					.prepare(
+						`SELECT source, COUNT(*) as count FROM skill_invocations WHERE ${where} GROUP BY source ORDER BY count DESC`,
+					)
+					.all(...params) as { source: string; count: number }[];
+
+				const timelineCutoff = since ? "datetime(?)" : "datetime('now', '-7 days')";
+				const timelineParams = since ? [...params, since] : [...params];
+				const sparse = db
```
PR-Reviewer-Ant, line 169: Warning: when `since` is provided, the aggregate queries honor it, but the zero-filled `timeline` still only emits the last 7 calendar days (`for (let i = 6; i >= 0; i--)`). That means `/api/skills/analytics/:skill?since=...` can return totals for a long window while the timeline only represents the most recent week, which is a misleading API contract.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,195 @@
...
+					.prepare(
```
PR-Reviewer-Ant, line 170: When `?since=` is provided, the SQL query is filtered by that cutoff, but the returned timeline is still zero-filled using `new Date()` for the last 7 days only. For any request outside the current week, this endpoint will return misleading zeros or omit the requested date range entirely. That breaks the claimed per-skill timeline contract.

packages/cli/dashboard/src/lib/components/skills/SkillUsagePanel.svelte
PR-Reviewer-Ant: This panel always calls `fetchSkillAnalytics()` without an `agentId`. The PR description says the dashboard now forwards agent-aware scoping, but this component ignores it, so in a multi-agent dashboard the analytics view will keep showing default/current scope data rather than the selected agent's data.

packages/cli/dashboard/src/lib/components/marketplace/McpUsagePanel.svelte
PR-Reviewer-Ant: Same issue here: the panel polls `fetchMcpAnalytics()` with no `agentId`, even though the API/store changes added agent-aware parameters. If the dashboard supports viewing another agent, this panel will not follow that selection and will show the wrong tenant's analytics.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -305,6 +282,44 @@ export async function executeTask(
...
+	// Phase 2: add agent_id column to scheduled_tasks for multi-agent support.
```
PR-Reviewer-Ant: This is the only new write path into `skill_invocations`, and it only runs for scheduled tasks. The PR description/spec/migration all claim skill analytics tracks broader sources (`slash-command`, `cli`, `api`) and records general skill usage, but none of those execution paths are instrumented in this diff. As shipped, the analytics API/dashboard will silently undercount almost all real skill usage while advertising full tracking.

packages/cli/dashboard/src/lib/components/skills/SkillUsagePanel.svelte
PR-Reviewer-Ant: The panel fetches analytics without passing the selected `agentId`. The new API/store types added `agentId` support and the PR description explicitly says dashboard stores now forward agent scope, but this component still always queries the default scope. In multi-agent installs that means the Skills tab can show the wrong tenant's analytics for the currently selected agent.

packages/cli/dashboard/src/lib/components/marketplace/McpUsagePanel.svelte
PR-Reviewer-Ant: Same scoping bug as the skill panel: `fetchMcpAnalytics()` is called with no `agentId`, so the MCP analytics card ignores the selected agent context even though the API/store surface was extended for it. That diverges from the PR description and breaks the repo's agent-scoping invariant in the dashboard.

packages/daemon/src/daemon.ts
```diff
@@ -7624,6 +7628,35 @@ app.post("/api/tasks/:id/run", async (c) => {
 					     stdout = ?, stderr = ?, error = ?
 					 WHERE id = ?`,
 					).run(status, completedAt, result.exitCode, result.stdout, result.stderr, result.error, runId);
+
+					// Record skill invocation for manual API-triggered runs
+					if (taskSkillName) {
+						const latencyMs = new Date(completedAt).getTime() - new Date(now).getTime();
```
PR-Reviewer-Ant, line 7641: This new manual-run instrumentation path still hardcodes `agent_id = 'default'` and updates `skill_meta` for the default tenant only. The PR description says the write path is now auth-aware, and it documents the scheduler fallback as a Phase 1 limitation, but `/api/tasks/:id/run` has the same scoping leak and is not called out. In a multi-agent install, a manual task run from another agent context will attribute usage to the default agent and corrupt both `skill_invocations` and `skill_meta` analytics.

packages/cli/dashboard/src/lib/stores/skill-analytics.svelte.ts
```diff
@@ -0,0 +1,24 @@
+import { type SkillAnalyticsSummary, getSkillAnalytics } from "$lib/api";
+
+export const skillAnalytics = $state({
```
PR-Reviewer-Ant, line 3: The new dashboard analytics state is a single global store even though callers now pass `agentId`. If two panels mount with different agent IDs, or the user switches agents while an earlier request is still in flight, the last response wins and every consumer reads the same `skillAnalytics.data`. That means the UI can render one agent's usage data in another agent's view. This needs per-agent/per-query state or request-keyed caching instead of one shared singleton.

PR-Reviewer-Ant:

> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5ce5d616`

[Automated Review] Found cross-agent analytics corruption in the new skill invocation write paths.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new write paths in scheduler/manual task execution insert `skill_invocations.agent_id = 'default'` and update `skill_meta` for `agent_id = 'default'` unconditionally, while the repo contract says agent scoping is universal and the PR description advertises agent-scoped analytics/dashboard support. This is directly visible in the diff and causes non-default agents' task-driven usage to be written into the default tenant.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -310,6 +310,44 @@ export async function executeTask(
...
+	if (task.skill_name) {
+		const latencyMs = new Date(completedAt).getTime() - new Date(now).getTime();
+		const success = status === "completed";
+		try {
+			db.withWriteTx((wdb) => {
```
PR-Reviewer-Ant, line 321: Blocking: this records every scheduled-task invocation under `agent_id = 'default'`, while the new analytics API and dashboard now scope reads by `agent_id`. In a multi-agent workspace, runs from non-default agents will be written into the default tenant and disappear from the selected agent's analytics, which is both incorrect attribution and cross-tenant pollution. The same bug exists in the manual run path in `daemon.ts`.

docs/specs/planning/skill-usage-analytics.md
```diff
@@ -0,0 +1,51 @@
...
+*Track skill invocations and surface usage analytics in the dashboard.*
+
+## Problem Statement
+
+Skills are installed and discoverable but usage is invisible. The
+`skill_meta` table has `use_count` and `last_used_at` columns (since
+migration 018) that are never written to. Operators cannot see which
+skills agents rely on, how often they run, or whether they succeed.
+
+## Implemented (Phase 1)
```
PR-Reviewer-Ant, line 27: This spec now claims Phase 1 tracks `scheduled-task`, `slash-command`, `cli`, and `api` sources, but the supplied implementation only adds write paths for scheduled tasks and manual `/api/tasks/:id/run`. I do not see any new writer for `cli` or `slash-command`, so the source breakdown and totals will undercount those invocations and the implementation does not currently match the documented scope.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export function mountSkillAnalyticsRoutes(app: Hono, authMode: AuthMode = "local"): void {
+	app.get("/api/skills/analytics", (c) => {
+		const scoped = resolveScopedAgent(
+			c.get("auth")?.claims ?? null,
+			authMode,
+			c.req.query("agent_id"),
+		);
+		if (scoped.error) return c.json({ error: scoped.error }, 403);
+		const agentId = scoped.agentId;
```
PR-Reviewer-Ant, line 62: The `since` query param is used without validation in both endpoints. An invalid value is fed into SQLite `datetime(?)` and also into `new Date(since)` for timeline generation, which can silently return empty or malformed results instead of a clear 400. This violates the repo's boundary-validation guardrail and makes analytics responses hard to trust when callers pass bad input.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7587,7 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
```
PR-Reviewer-Ant, line 7587: Blocking: this new write path resolves `agent_id` from a raw query param via `resolveAgentId(...)` instead of the auth-aware `resolveScopedAgent()` helper used by the analytics read endpoints and marketplace write path. That lets a caller attribute an invocation to an arbitrary agent ID, which breaks the repo's tenant-scoping invariant and can poison per-agent analytics.

packages/daemon/src/daemon.ts
```diff
@@ -7624,6 +7629,36 @@ app.post("/api/tasks/:id/run", async (c) => {
...
+						const success = status === "completed";
+						try {
+							db.prepare(
+								`INSERT INTO skill_invocations (id, skill_name, agent_id, source, task_id, latency_ms, success, error_text, created_at)
+								 VALUES (?, ?, ?, 'api', ?, ?, ?, ?, datetime(?))`,
+							).run(
+								`sinv-${Date.now()}-${Math.random().toString(36).slice(2, 10)}`,
+								taskSkillName,
+								taskAgentId,
+								taskId,
```
PR-Reviewer-Ant, line 7651: Warning: invocation rows are inserted with the original `taskSkillName`, but the `skill_meta` update lowercases the name before lookup. If task creation ever allows mixed-case skill names, analytics will split counts across distinct `skill_name` values while `skill_meta.use_count` updates only the lowercased entity. This creates inconsistent usage data between the analytics table and `skill_meta`.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7587,11 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
 	const taskSkillMode = typeof task.skill_mode === "string" ? task.skill_mode : null;
```
PR-Reviewer-Ant, line 7588: This write path claims auth-aware scoping, but it ignores `resolveScopedAgent().error` and unconditionally uses `.agentId`. The read endpoints correctly return `403` on scope errors; this one does not. That means an unauthorized `agent_id` on manual task runs is not rejected here, so the new analytics write path does not actually enforce the same tenant-isolation guarantee described in the PR.

packages/cli/dashboard/src/lib/components/skills/SkillUsagePanel.svelte
```diff
@@ -0,0 +1,229 @@
+<script lang="ts">
+import { fetchSkillAnalytics, skillAnalytics } from "$lib/stores/skill-analytics.svelte";
+import { onMount } from "svelte";
+
+interface Props {
+	agentId?: string;
+}
+
+const { agentId }: Props = $props();
+
```
PR-Reviewer-Ant, line 9: The new `agentId` support is only read inside `onMount()`. If the selected agent changes while this tab stays mounted, the panel keeps showing the previous agent's analytics and the polling interval keeps hitting the old `agent_id`. The PR description says the dashboard store/API now supports agent scoping, but this component never reacts to prop updates, so agent switching is stale until remount.

packages/cli/dashboard/src/lib/components/marketplace/McpUsagePanel.svelte
```diff
@@ -2,9 +2,15 @@
 import { fetchMcpAnalytics, mcpAnalytics } from "$lib/stores/mcp-analytics.svelte";
 import { onMount } from "svelte";
 
+interface Props {
+	agentId?: string;
+}
+
+const { agentId }: Props = $props();
+
```
PR-Reviewer-Ant, line 9: This has the same stale-prop bug as the new skill panel: `agentId` is captured once in `onMount()`, so switching agents without remounting leaves MCP analytics polling the old tenant. Since this PR explicitly forwards `agentId` into the MCP usage panel, the expected scoped behavior is still not achieved.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7587,13 @@ app.post("/api/tasks/:id/run", async (c) => {
...
 	const taskWorkingDir = typeof task.working_directory === "string" ? task.working_directory : null;
+	const scoped = resolveScopedAgent(
+		c.get("auth")?.claims ?? null,
+		authConfig.mode,
```
PR-Reviewer-Ant: Blocking: this changes analytics attribution for manual task runs to the caller-selected `agent_id`, even though the task model still has no `agent_id` column. The same task will now count toward `'default'` when run by the scheduler, but toward whichever agent is passed to `/api/tasks/:id/run` when triggered manually. That makes `skill_invocations` and `skill_meta.use_count` inconsistent for the same underlying task and contradicts the PR's documented Phase 1 limitation that scheduled-task attribution is still `'default'` until schema support exists. Until tasks are actually agent-owned, the manual run path should use the same attribution rule as the scheduler or refuse per-agent attribution entirely.

packages/daemon/src/daemon.ts
```diff
@@ -7624,6 +7634,40 @@ app.post("/api/tasks/:id/run", async (c) => {
 					     stdout = ?, stderr = ?, error = ?
 					 WHERE id = ?`,
 					).run(status, completedAt, result.exitCode, result.stdout, result.stderr, result.error, runId);
+
+					// Record skill invocation for manual API-triggered runs.
+					// Attribution uses 'default' — scheduled_tasks has no agent_id column.
+					// Phase 2: derive from task's owning agent once the column exists.
+					if (taskSkillName) {
+						const normalizedSkill = taskSkillName.toLowerCase();
```
PR-Reviewer-Ant, line 7640: Blocking: this route resolves `scoped.agentId`, but the analytics write path immediately throws that away and records the invocation under `'default'`. For any manual task run executed in a non-default agent context, usage will be attributed to the wrong tenant, the selected agent will show zero usage, and the default agent's `skill_meta.use_count` / `last_used_at` will be polluted. That contradicts the PR description's agent-aware scoping claims and the repo's multi-agent invariant.

packages/daemon/src/routes/skill-analytics.test.ts
```diff
@@ -0,0 +1,110 @@
+import { Database } from "bun:sqlite";
```
PR-Reviewer-Ant, line 1: Warning: the new tests validate raw SQL aggregation only; they do not exercise the new HTTP routes or the `/api/tasks/:id/run` write path that now depends on auth-aware scoping. As written, this suite would not catch the misattribution bug above or any regression in `resolveScopedAgent()` enforcement on the analytics endpoints, despite the PR claiming regression coverage for those fixes.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7587,13 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
 	const taskSkillMode = typeof task.skill_mode === "string" ? task.skill_mode : null;
 	const taskWorkingDir = typeof task.working_directory === "string" ? task.working_directory : null;
+	const scoped = resolveScopedAgent(
```
PR-Reviewer-Ant: The new `/api/tasks/:id/run` analytics write path attributes usage to `c.req.query("agent_id")` via `resolveScopedAgent()`, then writes that `taskAgentId` into `skill_invocations` and `skill_meta`. Because `scheduled_tasks` still has no owning `agent_id`, the recorded tenant is now caller-controlled rather than task-derived. In local/admin contexts this lets the same workspace-level task mutate another agent's usage stats simply by choosing a different query param, which undercuts the PR's claim that write-path scoping prevents analytics poisoning.

packages/cli/dashboard/src/lib/components/tabs/MarketplaceTab.svelte
```diff
@@ -28,6 +28,12 @@ import { toast } from "$lib/stores/toast.svelte";
 import ChevronDown from "@lucide/svelte/icons/chevron-down";
 import { onMount } from "svelte";
 
+interface Props {
+	agentId?: string;
+}
+
```
PR-Reviewer-Ant, line 34: This component switched to `const { agentId } = $props();`, while the other updated components use `let { ... } = $props();` for reactive props. That means agent changes from the parent can stop propagating here, leaving `SkillsTab`/`McpServersTab` mounted with the original agent id and showing the wrong analytics after an agent switch. Since the PR explicitly claims dashboard `agentId` forwarding, this is a correctness regression in the multi-agent UI path.

NicholaiVogel:
hey — the bot's been flagging the agentId forwarding thing on every review round and it looks legit. the dashboard panels call fetchSkillAnalytics/fetchMcpAnalytics without passing agentId even though the store signatures accept it now. just a heads up so it doesn't slip through, everything else looks solid 👍

aaf2tbz:
panels not forwarding agentId — we already fixed this with $effect in commit 0c38c3c3

packages/daemon/src/daemon.ts
PR-Reviewer-Ant: Blocking: this endpoint now attributes skill usage to the caller-selected/auth-scoped `agent_id`, but the task model itself is still unscoped (`scheduled_tasks` has no `agent_id`, per the PR description). That means the same stored task can be manually run with different `agent_id` query params and will write invocation rows plus `skill_meta.use_count` updates into whichever tenant was requested, corrupting per-agent analytics/state. Until tasks themselves carry authoritative tenant ownership, this write path should stay on the same explicit fallback as the scheduler (`default`) or refuse non-default attribution.

packages/cli/src/commands/skill.ts
```diff
@@ -270,4 +270,65 @@ export function registerSkillCommands(program: Command, deps: SkillDeps): void {
...
+			const params = new URLSearchParams();
+			if (options.skill) params.set("skill", options.skill);
```
PR-Reviewer-Ant, line 288: The new `--skill` flag still calls the aggregate endpoint (`/api/skills/analytics?skill=...`) and prints the summary/top-skills shape. It never uses the per-skill endpoint introduced by this PR (`/api/skills/analytics/:skill`), so the CLI does not actually expose the per-skill breakdown/timeline the implementation adds. If the PR description is meant literally, this diverges from the stated feature set.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant: The PR claims auth-aware/per-agent attribution for API-triggered skill runs, but this endpoint now records analytics under `resolveScopedAgent(..., c.req.query("agent_id"))` while the changed task clients in this PR do not forward `agent_id` when triggering a run. In practice, manual runs from the dashboard/CLI will still be attributed to `default` unless the caller adds that query param out of band, so the new analytics data will be wrong for multi-agent users.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,208 @@
...
+	readonly skillName: string;
+	readonly totalInvocations: number;
+	readonly successRate: number;
+	readonly sources: readonly { readonly source: string; readonly count: number }[];
+	readonly timeline: readonly { readonly date: string; readonly count: number }[];
+	readonly latency: { readonly p50: number; readonly p95: number };
+}
+
+function clampPositiveInt(value: unknown, fallback: number, min: number, max: number): number {
+	const n = Number(value);
```
PR-Reviewer-Ant, line 38: The `since` validator accepts any string that `new Date(...)` happens to parse, even though the API error says ISO 8601 is required. That makes filtering behavior implementation-dependent and can silently accept malformed dates with unexpected timezone coercion. This should validate the declared wire format explicitly before using it in the query.

aaf2tbz:
Closing to reopen with clean history — too many fix-on-fix commits from review cycling.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7587,10 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
 	const taskSkillMode = typeof task.skill_mode === "string" ? task.skill_mode : null;
 	const taskWorkingDir = typeof task.working_directory === "string" ? task.working_directory : null;
+	// Attribution uses 'default' — scheduled_tasks has no agent_id column,
```
PR-Reviewer-Ant, line 7590: Manual task runs still hardcode `agent_id = 'default'` when recording analytics. That breaks the PR's claimed auth-aware scoping for API-triggered writes: a non-default agent can run a skill, but the invocation and `skill_meta.use_count` update will be attributed to `default`, so the caller's scoped analytics miss their own run while the default agent's metrics are polluted.

packages/cli/dashboard/src/lib/stores/skill-analytics.svelte.ts
```diff
@@ -0,0 +1,24 @@
...
+	data: null as SkillAnalyticsSummary | null,
+	loading: false,
+	error: null as string | null,
+});
+
```
PR-Reviewer-Ant, line 8: This new store is global, but fetches are now parameterized by `agentId`. If the user switches agents or mounts another panel before an earlier request resolves, the late response can overwrite `skillAnalytics.data` for the currently selected agent. Because the panel polls every 30s, this creates a real stale-data race for multi-agent dashboards unless requests are cancelled or state is keyed per agent.

PR #409, opened 2026-03-30
fix(worker): exclude exhausted jobs from watchdog stall detector query (#339)

### LINKED ISSUE #339: Watchdog stall detector counts exhausted jobs as pending, causing infinite reset loop
**Signet version:** 0.73.3 (confirmed still present in 0.76.4 via code review)
**Environment:** Ubuntu (Linux 6.8.0-94-generic), Node 22.22.0, Bun 1.3.11, inside Docker container
**Related issues:** #248, #311, #303

---

## Summary

The pipeline worker's watchdog stall detector in `startWorker()` queries `SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending'` without filtering `attempts < max_attempts`. When a job is stuck at `pending` with `attempts >= max_attempts` (the exact scenario #311 addressed), the watchdog sees `pending > 0`, resets the backoff, and forces a tick every 60 seconds. The actual worker query correctly filters `WHERE status = 'pending' AND attempts < max_attempts` and finds nothing. This creates an infinite watchdog-reset loop that runs 24/7 until the daemon is restarted or the stuck job is manually patched.

This is distinct from #311: that issue fixed the root data problem (jobs not transitioning to `dead`). This issue is about the watchdog amplifier that turns a single stuck row into a runaway loop with cascading LLM cost. The #311 fix (universal promotion sweep in `requeueDeadJobs`) mitigates it on maintenance cycles, but there is a race window between daemon startup and the first maintenance sweep where the watchdog fires repeatedly.

---

## Observed Behavior

Signet daemon logs show `Worker stall detected, resetting backoff` every 60 seconds, 24 hours a day, for over 48 hours:

```
{"timestamp":"2026-03-24T00:00:02.987Z","level":"warn","category":"pipeline","message":"Worker stall detected, resetting backoff","data":{"pending":1,"failures":0,"stalledMs":60007}}
{"timestamp":"2026-03-24T00:01:32.987Z","level":"warn","category":"pipeline","message":"Worker stall detected, resetting backoff","data":{"pending":1,"failures":0,"stalledMs":90000}}
{"timestamp":"2026-03-24T00:02:32.988Z","level":"warn","category":"pipeline","message":"Worker stall detected, resetting backoff","data":{"pending":1,"failures":0,"stalledMs":60000}}
... (1,339 stall resets on 2026-03-24 alone)
```

The single pending job causing this:

```sql
SELECT id, memory_id, status, attempts, max_attempts, error FROM memory_jobs
WHERE status = 'pending' AND attempts >= max_attempts;

-- cc6ecc0e... | cda77bfb... | pending | 3 | 3 | LLM extraction failed: claude-code timeout after 90000ms
```

**Stall reset counts by day:**

| Date | Stall resets | Notes |
|------|-------------|-------|
| Mar 22 | 0 | Pre-0.73.3 (no watchdog, or different query) |
| Mar 23 | 759 | Loop started at 09:36 UTC after daemon restart |
| Mar 24 | 1,339 | Full day of continuous resets |
| Mar 25 | 381 | Ongoing until manual DB fix at ~04:55 UTC |

---

## Expected Behavior

The watchdog should only count jobs that are actually eligible for processing:

```sql
SELECT COUNT(*) as cnt FROM memory_jobs
WHERE job_type = 'extract' AND status = 'pending' AND attempts < max_attempts
```

A job at `attempts >= max_attempts` in `pending` state is a data anomaly, not a stall signal. The watchdog should not reset backoff or force ticks for jobs that the worker will never pick up.

---

## Steps to Reproduce

1. Create a `memory_jobs` entry with `status = 'pending'`, `attempts = 3`, `max_attempts = 3` (simulating the #311 bug, or any race condition in the error handler)
2. Start the Signet daemon
3. Observe: `Worker stall detected, resetting backoff` fires every 60 seconds indefinitely
4. Observe: the actual worker tick finds nothing to process (correct `attempts < max_attempts` filter)
5. The loop never self-resolves

---

## Root Cause Analysis

In `daemon.js` (around line 40033), the watchdog `setInterval` runs every `WATCHDOG_INTERVAL` (30s):

```javascript
watchdog = setInterval(() => {
    // ...
    const row = accessor2.withReadDb((db) =>
        db.prepare("SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending'").get()
    );
    pending = row.cnt;
    if (pending === 0) return;
    // Resets backoff, forces tick...
}, WATCHDOG_INTERVAL);
```

The actual worker in `processExtractJob` (line 39618+) uses:

```sql
WHERE status = 'pending' AND attempts < max_attempts
```

The mismatch between these two queries is the bug. The watchdog sees work that the worker will never do.

---

## Suggested Fix

Align the watchdog query with the worker query:

```javascript
const row = accessor2.withReadDb((db) =>
    db.prepare(
        "SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending' AND attempts < max_attempts"
    ).get()
);
```

This is a one-line change. The `requeueDeadJobs` universal sweep from #311's fix provides the cleanup path for the stuck rows; this fix prevents the watchdog from reacting to them in the interim.

---

## Impact

- **Cost amplification**: The infinite watchdog loop keeps all daemon subsystems active. In our case, the `dependency-synthesis` worker (running on the same 60s interval) continued making LLM calls via `claude-code` provider throughout the loop. Over 48 hours this produced ~3,100+ synthesis calls and ~16,000 Claude CLI subprocess spawns per day, all producing no useful extraction work. Estimated cost: $400-900 in unnecessary API consumption.
- **CPU waste**: Continuous forced ticks + synthesis calls on a 2-vCPU VPS elevated load average from ~0.3 to ~1.3-2.5.
- **Log noise**: 1,300+ warning lines per day obscure real pipeline issues.
- **Race window**: Even with the #311 fix, a stuck job created between daemon start and the first `requeueDeadJobs` maintenance sweep (up to 30 minutes) will trigger the loop.

---

## Workaround

Manual SQL patch:

```sql
UPDATE memory_jobs SET status = 'dead'
WHERE attempts >= max_attempts AND status = 'pending';
```

---

## Environment Details

- **Signet:** 0.73.3
- **OS:** Ubuntu (Linux 6.8.0-94-generic, Docker container)
- **Node:** 22.22.0
- **Bun:** 1.3.11
- **DB:** SQLite (memories.db)
- **Date observed:** 2026-03-23 through 2026-03-25

nanookclaw:
## Problem

The watchdog stall detector and `pendingCount()` both used an unfiltered query:

```sql
SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending'
```

This counted jobs with `attempts >= maxRetries` as pending, even though `leaseJob()` skips them with `attempts < pipelineCfg.worker.maxRetries`. Result: any job that exhausted retries before startup recovery ran would trigger the stall watchdog every 60s indefinitely.

## Fix

Align watchdog SELECT and `pendingCount()` to use the same eligibility criterion as `leaseJob()`:

```sql
SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending' AND attempts < ?
```

bound to `pipelineCfg.worker.maxRetries` — not the DB column `max_attempts`, which can diverge from the runtime setting.

## Tests

Added 2 tests in `watchdog pendingCount — exhausted jobs excluded`:
- `stats.pending` excludes exhausted jobs (attempts === maxRetries)
- `stats.pending` is 0 when only exhausted jobs remain

Pre-existing test suite: 41 pass (2 failures are pre-existing and unrelated to this PR).

Fixes #339

### DIFF
--- a/packages/daemon/src/pipeline/worker.ts
+++ b/packages/daemon/src/pipeline/worker.ts
@@ -925,25 +925,50 @@ interface MemoryRecoveryResult {
 }
 
 /**
- * Mark memory_jobs that are stuck in 'pending' with attempts >= max_attempts
- * as 'dead'. The tick loop requires attempts < max_attempts, so these jobs
- * are silently skipped forever without this recovery step, causing the stall
- * detector to fire on every interval.
+ * Mark memory_jobs that are stuck in 'pending' with attempts >= their
+ * effective retry limit as 'dead'. The effective limit is the lesser of the
+ * stored max_attempts column and the current runtime maxRetries value, so
+ * that jobs are also recovered when the operator lowers maxRetries after rows
+ * were originally inserted with a higher max_attempts value.
+ *
+ * @param accessor - DB accessor
+ * @param maxRetries - runtime retry limit from pipelineCfg.worker.maxRetries
  */
-export function recoverMemoryJobs(accessor: DbAccessor): MemoryRecoveryResult {
+export function recoverMemoryJobs(
+	accessor: DbAccessor,
+	maxRetries?: number,
+): MemoryRecoveryResult {
 	return accessor.withWriteTx((db) => {
 		const table = db
 			.prepare("SELECT name FROM sqlite_master WHERE type = 'table' AND name = 'memory_jobs'")
 			.get() as { name: string } | undefined;
 		if (!table) return { updated: 0 };
 
-		const updated = countChanges(
-			db.prepare(
-				`UPDATE memory_jobs
-				 SET status = 'dead'
-				 WHERE status = 'pending' AND attempts >= max_attempts`,
-			).run(),
-		);
+		// Use MIN(max_attempts, maxRetries) so that both the DB-stored limit and
+		// the current runtime limit are honoured. When maxRetries is undefined
+		// (e.g. called from tests without a pipeline config) fall back to the
+		// column value only.
+		const updated =
+			maxRetries !== undefined
+				? countChanges(
+						db
+							.prepare(
+								`UPDATE memory_jobs
+							 SET status = 'dead'
+							 WHERE status = 'pending'
+							   AND attempts >= MIN(max_attempts, ?)`,
+							)
+							.run(maxRetries),
+					)
+				: countChanges(
+						db
+							.prepare(
+								`UPDATE memory_jobs
+							 SET status = 'dead'
+							 WHERE status = 'pending' AND attempts >= max_attempts`,
+							)
+							.run(),
+					);
 		return { updated };
 	});
 }
@@ -1660,12 +1685,14 @@ export function startWorker(
 	}, WATCHDOG_INTERVAL);
 
 	// Startup recovery: mark memory_jobs that are pending but have exhausted
-	// all attempts as dead. These are skipped by the tick loop (requires
-	// attempts < max_attempts) but stay in 'pending' forever, causing the
-	// stall detector to fire on every interval. Same pattern as summary-worker's
-	// recoverSummaryJobs for summary_jobs.
+	// their effective retry limit as dead. The effective limit is the lesser of
+	// the stored max_attempts column and the current runtime maxRetries value,
+	// so jobs are also recovered when the operator lowers maxRetries after rows
+	// were inserted with a higher max_attempts value. These jobs are silently
+	// skipped by the tick loop (requires attempts < maxRetries) and cause the
+	// stall detector to fire every interval without this cleanup step.
 	try {
-		const { updated } = recoverMemoryJobs(accessor);
+		const { updated } = recoverMemoryJobs(accessor, pipelineCfg.worker.maxRetries);
 		if (updated > 0) logger.info("pipeline", `Startup recovery: marked ${updated} exhausted pending job(s) as dead`);
 	} catch (e) {
 		logger.warn("pipeline", "Startup recovery failed (non-fatal)", {
--- a/packages/daemon/src/pipeline/worker.test.ts
+++ b/packages/daemon/src/pipeline/worker.test.ts
@@ -1992,6 +1992,39 @@ describe("recoverMemoryJobs", () => {
 		const { updated } = recoverMemoryJobs(accessor);
 		expect(updated).toBe(0);
 	});
+
+	it("recovers jobs exhausted by lowered maxRetries even if max_attempts column is higher", () => {
+		// Regression: operator lowers maxRetries from 5 to 3 after rows were
+		// inserted with max_attempts=5. jobs with attempts=3 are now above the
+		// runtime limit (maxRetries=3) but below the stored column (max_attempts=5).
+		// Without the MIN(max_attempts, maxRetries) fix these jobs stay stuck in
+		// 'pending' forever: leaseJob() won't pick them up and recoverMemoryJobs
+		// without maxRetries won't mark them dead.
+		const now = new Date().toISOString();
+		// attempts=3 >= maxRetries=3, but max_attempts=5 — should become dead
+		db.prepare(
+			`INSERT INTO memory_jobs (id, memory_id, job_type, status, attempts, max_attempts, created_at, updated_at)
+			 VALUES ('job-runtime-exhausted', NULL, 'extract', 'pending', 3, 5, ?, ?)`,
+		).run(now, now);
+		// attempts=2 < maxRetries=3 — should stay pending
+		db.prepare(
+			`INSERT INTO memory_jobs (id, memory_id, job_type, status, attempts, max_attempts, created_at, updated_at)
+			 VALUES ('job-still-eligible', NULL, 'extract', 'pending', 2, 5, ?, ?)`,
+		).run(now, now);
+
+		const { updated } = recoverMemoryJobs(accessor, 3); // maxRetries lowered to 3
+		expect(updated).toBe(1);
+
+		const dead = db
+			.prepare("SELECT status FROM memory_jobs WHERE id = 'job-runtime-exhausted'")
+			.get() as { status: string } | undefined;
+		const active = db
+			.prepare("SELECT status FROM memory_jobs WHERE id = 'job-still-eligible'")
+			.get() as { status: string } | undefined;
+
+		expect(dead?.status).toBe("dead");
+		expect(active?.status).toBe("pending");
+	});
 });
 
 describe("watchdog pendingCount — exhausted jobs excluded", () => {


PR-Reviewer-Ant:
Hi @nanookclaw - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(worker): exclude exhausted jobs from watchdog stall detector query (#339)` (commit `ffdf2250`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1630,7 +1630,13 @@ export function startWorker(
 			const row = accessor.withReadDb(
 				(db) =>
 					db
-						.prepare("SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending'")
+						// Filter attempts < max_attempts to match the tick loop's eligibility
+						// criteria. Without this, exhausted jobs (attempts >= max_attempts,
+						// still in 'pending') trigger the watchdog every 60 s indefinitely,
```
PR-Reviewer-Ant: This new filter does not actually mirror the tick loop. `leaseJob()` still selects jobs with `attempts < pipelineCfg.worker.maxRetries`, while this query now uses the row's `max_attempts`. If those values diverge, the watchdog can report `pending = 0` even though the worker will still lease and process the job, or vice versa. That makes the PR description inaccurate and leaves the stall detector/helper out of sync with real worker behavior. The fix needs a single shared eligibility predicate, plus a regression test where `maxRetries !== memory_jobs.max_attempts`.

nanookclaw:
Thanks for the precise catch, @PR-Reviewer-Ant.

You're right — my original fix used `attempts < max_attempts` (DB column), but `leaseJob()` uses `attempts < pipelineCfg.worker.maxRetries` (runtime config). These can diverge if `max_attempts` at row-insert time differs from the current runtime setting.

**Fix:** Both the watchdog SELECT and `pendingCount()` now use `attempts < ?` bound to `pipelineCfg.worker.maxRetries`, mirroring `leaseJob()` exactly.

**New tests:** Added 2 tests in `watchdog pendingCount — exhausted jobs excluded`:
1. `stats.pending` excludes jobs where `attempts === maxRetries` — verified via a job in backoff (not leaseable) + an exhausted job (should not be counted)
2. `stats.pending` is 0 when only exhausted jobs remain

Pushing updated commit now.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1630,8 +1632,10 @@ export function startWorker(
 			const row = accessor.withReadDb(
 				(db) =>
 					db
-						.prepare("SELECT COUNT(*) as cnt FROM memory_jobs WHERE job_type = 'extract' AND status = 'pending'")
-						.get() as { cnt: number },
+						.prepare(
```
PR-Reviewer-Ant, line 1660: Using `pipelineCfg.worker.maxRetries` here fixes the watchdog/counting mismatch, but it also exposes that startup recovery still uses `attempts >= max_attempts` in `recoverMemoryJobs()`. When runtime `maxRetries` is reduced below the row's stored `max_attempts`, those jobs become permanently unleaseable (`leaseJob()` skips them), remain `pending` after startup recovery, and now disappear from `stats.pending`/watchdog visibility. That leaves silently stuck rows in `memory_jobs`, which diverges from the PR description's goal of aligning eligibility with the runtime policy.

nanookclaw:
Pushed `8af37452` to address the final blocking review finding.

**Fix:** `recoverMemoryJobs()` now accepts an optional `maxRetries` parameter and uses `MIN(max_attempts, ?)` in the SQL predicate. The effective retry limit is now the lesser of the stored DB column and the current runtime value.

At the call site in `startWorker()`, `pipelineCfg.worker.maxRetries` is passed through, so startup recovery handles the config-lowered case correctly:

> jobs with `attempts >= pipelineCfg.worker.maxRetries` but `< max_attempts` — now marked dead on startup

**New test:** *recovers jobs exhausted by lowered maxRetries even if max_attempts column is higher* — inserts a job with `attempts=3, max_attempts=5`, calls `recoverMemoryJobs(accessor, 3)`, asserts it becomes `dead`.

All 5 recoverMemoryJobs + watchdog tests pass.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -925,25 +925,50 @@ interface MemoryRecoveryResult {
...
 
 /**
- * Mark memory_jobs that are stuck in 'pending' with attempts >= max_attempts
- * as 'dead'. The tick loop requires attempts < max_attempts, so these jobs
- * are silently skipped forever without this recovery step, causing the stall
- * detector to fire on every interval.
+ * Mark memory_jobs that are stuck in 'pending' with attempts >= their
+ * effective retry limit as 'dead'. The effective limit is the lesser of the
+ * stored max_attempts column and the current runtime maxRetries value, so
+ * that jobs are also recovered when the operator lowers maxRetries after rows
```
PR-Reviewer-Ant, line 931: This changes observable daemon behavior by redefining the effective retry limit for pending-job recovery, but there is no matching update in `packages/daemon-rs/`. Per the repo's Rust parity rule, the shadow rewrite must mirror daemon behavior changes. Without the Rust-side change, shadow mode can now diverge on whether lowered-`maxRetries` jobs are still pending or are recovered as dead, which makes parity logging noisy and can hide real regressions.

nanookclaw:
Pushed `091b2db` to address the Rust parity finding.

**Fix:** `packages/daemon-rs/crates/signet-pipeline/src/worker.rs` startup recovery now uses the same `MIN(max_attempts, effective_max_retries)` logic as the JS fix:

```rust
let effective_max_retries = i64::from(config.max_retries);
// ...
"UPDATE memory_jobs SET status = 'dead'
 WHERE status = 'pending'
   AND attempts >= MIN(max_attempts, ?1)"
// bound to effective_max_retries
```

The comment on line 275 is also updated to reflect that the tick loop uses `config.max_retries` (runtime) not `max_attempts` (DB column).

**New tests** (2, both passing with all 5 worker tests green):
1. `startup_recovery_respects_runtime_max_retries_when_lower_than_db_column` — job with `attempts=3, max_attempts=5` becomes `dead` when runtime `maxRetries=3`
2. `startup_recovery_does_not_mark_leasable_jobs_dead` — job with `attempts=1, max_attempts=3` stays `pending` when runtime `maxRetries=3`

NicholaiVogel:
nice catch on the MIN(max_attempts, maxRetries) case — that's the kind of config-drift bug that shows up quietly in production and takes forever to trace. watchdog/pendingCount/leaseJob all speaking the same eligibility language now is the right call. tests are solid. merging.

PR #412, opened 2026-03-30
fix(daemon): stop skill reconcile and codex provider loops

NicholaiVogel:
## Summary

Fix two daemon-side feedback loop regressions:

1. procedural skill reconciliation could re-install unchanged skills forever when install-time enrichment changed the stored embedding text
2. provider-spawned Codex extraction runs inherited the live user Codex environment, which let pipeline workers see Signet MCP and user skills

Together these could create noisy repeated reconcile work and make Codex-backed extraction runs much heavier and more self-referential than intended.

## Changes

- fingerprint raw `SKILL.md` frontmatter for skill change detection instead of comparing against enriched embedding text
- store an entity-scoped skill embedding hash so reconciliation stays stable without cross-entity collisions
- add regression tests covering the skill reconciler infinite update loop and metadata-only updates
- run provider-spawned Codex extraction in a sterile temporary `HOME` / `CODEX_HOME` rooted in `tmpdir()`
- copy Codex auth into that temporary home as a read-only file, copy version state, and avoid deleting sibling temp homes so concurrent provider runs stay isolated
- force provider-spawned Codex runs to use `--ephemeral` and `-c mcp_servers.signet.enabled=false`
- add regression coverage asserting the Codex provider launch args and sterile environment behavior

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A, no UI changes.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

N/A for parity in this PR: the Rust shadow currently does not implement Codex extraction provider execution (`packages/daemon-rs/crates/signet-daemon/src/main.rs` asserts `worker_supports_extraction_provider("codex") == false`) and does not implement the procedural skill reconciler path changed here.

## Testing

- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Validation run:

- `bun test`
- `bun run typecheck`
- `bun test packages/daemon/src/pipeline/provider.test.ts`
- `bun test packages/daemon/src/pipeline/skill-reconciler.test.ts packages/daemon/src/pipeline/skill-graph.test.ts`
- targeted `biome check` on touched provider / skill reconciler files
- live local Codex probes verifying sterile temp-home runs preserve auth while removing MCP and user-skill inheritance

## AI disclosure

## Notes

- No API surface or migration changes.
- The Codex provider fix intentionally reduces inherited state for pipeline worker runs only. It does not change the normal interactive user Codex environment.
- Full-repo `bun run lint` still fails because of unrelated existing repo-wide baseline diagnostics, mostly formatter/style issues outside the touched files. Targeted checks for touched files passed.

### DIFF
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -5,7 +5,7 @@
  */
 
 import { afterEach, describe, expect, it, mock } from "bun:test";
-import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, lstatSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
 import { join } from "node:path";
 import {
@@ -336,7 +336,7 @@ describe("createCodexProvider", () => {
 		expect(typeof capturedEnv?.CODEX_HOME).toBe("string");
 	});
 
-	it("spawns Codex with a sterile temp home and copied auth", async () => {
+	it("spawns Codex with a sterile temp home and linked auth", async () => {
 		const root = join(tmpdir(), `signet-codex-provider-${Date.now()}-${Math.random().toString(36).slice(2)}`);
 		const home = join(root, "home");
 		const codex = join(home, ".codex");
@@ -352,12 +352,13 @@ describe("createCodexProvider", () => {
 			capturedEnv = opts?.env;
 			const srcAuth = join(homedir(), ".codex", "auth.json");
 			const srcVersion = join(homedir(), ".codex", "version.json");
+			const dstAuth = join(capturedEnv?.CODEX_HOME ?? "", "auth.json");
 			expect(capturedEnv?.CODEX_HOME).toBe(join(capturedEnv?.HOME ?? "", ".codex"));
-			expect(existsSync(join(capturedEnv?.CODEX_HOME ?? "", "auth.json"))).toBe(existsSync(srcAuth));
+			expect(capturedEnv?.HOME?.startsWith(tmpdir())).toBe(true);
+			expect(existsSync(dstAuth)).toBe(existsSync(srcAuth));
 			if (existsSync(srcAuth)) {
-				expect(readFileSync(join(capturedEnv?.CODEX_HOME ?? "", "auth.json"), "utf8")).toBe(
-					readFileSync(srcAuth, "utf8"),
-				);
+				expect(lstatSync(dstAuth).isSymbolicLink()).toBe(true);
+				expect(readFileSync(dstAuth, "utf8")).toBe(readFileSync(srcAuth, "utf8"));
 			}
 			expect(existsSync(join(capturedEnv?.CODEX_HOME ?? "", "version.json"))).toBe(existsSync(srcVersion));
 			return {
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -339,19 +339,21 @@ describe("createCodexProvider", () => {
 	it("spawns Codex with a sterile temp home and readonly copied auth", async () => {
 		const root = join(tmpdir(), `signet-codex-provider-${Date.now()}-${Math.random().toString(36).slice(2)}`);
 		const home = join(root, "home");
-		const codex = join(home, ".codex");
-		mkdirSync(codex, { recursive: true });
-		writeFileSync(join(codex, "auth.json"), '{"provider":"test"}');
-		writeFileSync(join(codex, "version.json"), '{"version":"1"}');
+		const liveCodex = join(root, "live-codex");
+		mkdirSync(liveCodex, { recursive: true });
+		writeFileSync(join(liveCodex, "auth.json"), '{"provider":"test"}');
+		writeFileSync(join(liveCodex, "version.json"), '{"version":"1"}');
 
 		const prevHome = process.env.HOME;
+		const prevCodexHome = process.env.CODEX_HOME;
 		process.env.HOME = home;
+		process.env.CODEX_HOME = liveCodex;
 
 		let capturedEnv: Record<string, string | undefined> | undefined;
 		Bun.spawn = mock((args: string[], opts?: { env?: Record<string, string | undefined> }) => {
 			capturedEnv = opts?.env;
-			const srcAuth = join(homedir(), ".codex", "auth.json");
-			const srcVersion = join(homedir(), ".codex", "version.json");
+			const srcAuth = join(liveCodex, "auth.json");
+			const srcVersion = join(liveCodex, "version.json");
 			const dstAuth = join(capturedEnv?.CODEX_HOME ?? "", "auth.json");
 			expect(capturedEnv?.CODEX_HOME).toBe(join(capturedEnv?.HOME ?? "", ".codex"));
 			expect(capturedEnv?.HOME?.startsWith(tmpdir())).toBe(true);
@@ -389,6 +391,11 @@ describe("createCodexProvider", () => {
 				delete process.env.HOME;
 			} else {
 				process.env.HOME = prevHome;
+			}
+			if (prevCodexHome === undefined) {
+				delete process.env.CODEX_HOME;
+			} else {
+				process.env.CODEX_HOME = prevCodexHome;
 			}
 			rmSync(root, { recursive: true, force: true });
 		}
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -357,7 +357,7 @@ describe("createCodexProvider", () => {
 			expect(capturedEnv?.HOME?.startsWith(tmpdir())).toBe(true);
 			expect(existsSync(dstAuth)).toBe(existsSync(srcAuth));
 			if (existsSync(srcAuth)) {
-				expect(lstatSync(dstAuth).isSymbolicLink()).toBe(true);
+				expect(lstatSync(dstAuth).isSymbolicLink() || lstatSync(dstAuth).nlink > 1).toBe(true);
 				expect(readFileSync(dstAuth, "utf8")).toBe(readFileSync(srcAuth, "utf8"));
 			}
 			expect(existsSync(join(capturedEnv?.CODEX_HOME ?? "", "version.json"))).toBe(existsSync(srcVersion));
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -394,6 +394,43 @@ describe("createCodexProvider", () => {
 		}
 	});
 
+	it("does not delete sibling sterile homes while Codex is running", async () => {
+		const root = join(tmpdir(), "signet-codex-home");
+		mkdirSync(root, { recursive: true });
+		const sibling = join(root, `home-sibling-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+		const marker = join(sibling, "marker.txt");
+		mkdirSync(sibling, { recursive: true });
+		writeFileSync(marker, "keep");
+
+		let capturedEnv: Record<string, string | undefined> | undefined;
+		Bun.spawn = mock((args: string[], opts?: { env?: Record<string, string | undefined> }) => {
+			capturedEnv = opts?.env;
+			expect(existsSync(marker)).toBe(true);
+			expect(capturedEnv?.HOME).not.toBe(sibling);
+			return {
+				stdout: streamFromString(
+					'{"type":"item.completed","item":{"type":"agent_message","text":"done"}}\n{"type":"turn.completed","usage":{"input_tokens":1,"output_tokens":1}}\n',
+				),
+				stderr: streamFromString(""),
+				exited: Promise.resolve(0),
+				kill() {},
+			};
+		}) as typeof Bun.spawn;
+
+		try {
+			const provider = createCodexProvider({ model: "gpt-5.3-codex" });
+			if (!provider.generateWithUsage) {
+				throw new Error("expected generateWithUsage on Codex provider");
+			}
+
+			await provider.generateWithUsage("test");
+
+			expect(existsSync(marker)).toBe(true);
+		} finally {
+			rmSync(sibling, { recursive: true, force: true });
+		}
+	});
+
 	it("generate() throws on non-zero exit", async () => {
 		Bun.spawn = mock((_args: string[]) => ({
 			stdout: streamFromString(""),
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -336,7 +336,7 @@ describe("createCodexProvider", () => {
 		expect(typeof capturedEnv?.CODEX_HOME).toBe("string");
 	});
 
-	it("spawns Codex with a sterile temp home and linked auth", async () => {
+	it("spawns Codex with a sterile temp home and readonly copied auth", async () => {
 		const root = join(tmpdir(), `signet-codex-provider-${Date.now()}-${Math.random().toString(36).slice(2)}`);
 		const home = join(root, "home");
 		const codex = join(home, ".codex");
@@ -357,8 +357,9 @@ describe("createCodexProvider", () => {
 			expect(capturedEnv?.HOME?.startsWith(tmpdir())).toBe(true);
 			expect(existsSync(dstAuth)).toBe(existsSync(srcAuth));
 			if (existsSync(srcAuth)) {
-				expect(lstatSync(dstAuth).isSymbolicLink() || lstatSync(dstAuth).nlink > 1).toBe(true);
+				expect(lstatSync(dstAuth).isSymbolicLink()).toBe(false);
 				expect(readFileSync(dstAuth, "utf8")).toBe(readFileSync(srcAuth, "utf8"));
+				expect(lstatSync(dstAuth).mode & 0o200).toBe(0);
 			}
 			expect(existsSync(join(capturedEnv?.CODEX_HOME ?? "", "version.json"))).toBe(existsSync(srcVersion));
 			return {
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -6,7 +6,7 @@
 
 import { afterEach, describe, expect, it, mock } from "bun:test";
 import { existsSync, lstatSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
-import { homedir, tmpdir } from "node:os";
+import { tmpdir } from "node:os";
 import { join } from "node:path";
 import {
 	DEFAULT_OLLAMA_FALLBACK_MAX_CONTEXT_TOKENS,
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -4,7 +4,7 @@
  * OllamaProvider uses the Ollama HTTP API, so we mock global fetch.
  */
 
-import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test";
+import { afterEach, describe, expect, it, mock } from "bun:test";
 import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
 import { join } from "node:path";
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -384,7 +384,7 @@ describe("createCodexProvider", () => {
 			expect(capturedEnv?.XDG_CONFIG_HOME).toBe(join(capturedEnv?.HOME ?? "", ".config"));
 		} finally {
 			if (prevHome === undefined) {
-				process.env.HOME = undefined;
+				delete process.env.HOME;
 			} else {
 				process.env.HOME = prevHome;
 			}
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -4,12 +4,15 @@
  * OllamaProvider uses the Ollama HTTP API, so we mock global fetch.
  */
 
-import { describe, it, expect, mock, beforeEach, afterEach } from "bun:test";
+import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test";
+import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { homedir, tmpdir } from "node:os";
+import { join } from "node:path";
 import {
 	DEFAULT_OLLAMA_FALLBACK_MAX_CONTEXT_TOKENS,
-	createOllamaProvider,
 	createClaudeCodeProvider,
 	createCodexProvider,
+	createOllamaProvider,
 	createOpenCodeProvider,
 	createOpenRouterProvider,
 	resolveDefaultOllamaFallbackMaxContextTokens,
@@ -305,8 +308,10 @@ describe("createCodexProvider", () => {
 
 	it("generateWithUsage() parses JSONL agent output and usage", async () => {
 		let capturedArgs: string[] = [];
-		Bun.spawn = mock((args: string[]) => {
+		let capturedEnv: Record<string, string | undefined> | undefined;
+		Bun.spawn = mock((args: string[], opts?: { env?: Record<string, string | undefined> }) => {
 			capturedArgs = args;
+			capturedEnv = opts?.env;
 			return {
 				stdout: streamFromString(
 					'{"type":"thread.started","thread_id":"abc"}\n{"type":"item.completed","item":{"type":"agent_message","text":"done"}}\n{"type":"turn.completed","usage":{"input_tokens":12,"cached_input_tokens":5,"output_tokens":7}}\n',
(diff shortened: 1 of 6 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): stop skill reconcile and codex provider loops` (commit `84c4a6fb`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/skill-graph.ts
```diff
@@ -77,6 +77,14 @@ function buildEmbeddingText(fm: SkillFrontmatter): string {
 	return parts.join(" — ");
 }
 
+export function skillFingerprint(fm: SkillFrontmatter): string {
```
PR-Reviewer-Ant, line 85: Blocking: this new fingerprint is described in the PR as hashing the raw `SKILL.md` frontmatter, but it still only includes `name`, `description`, and `triggers` because it delegates to `buildEmbeddingText()`. The reconciler now uses this hash as the sole change detector, so edits to other frontmatter fields that are stored in `skill_meta` (`version`, `author`, `license`, `tags`, `permissions`, `role`) will be silently ignored and never reconciled into the DB. The loop fix is valid, but the fingerprint needs to cover the full raw frontmatter payload, not the embedding-text subset.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `84c4a6fb`

[Automated Review] Found one blocking correctness issue where the new skill fingerprint does not actually cover the full raw frontmatter the PR description says it does.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed reconciler now keys updates off `skillFingerprintHash(parsed.frontmatter)`, but `skillFingerprintHash()` still hashes only `name`, `description`, and `triggers` via `buildEmbeddingText()`. Fields that are persisted into `skill_meta` such as `version`, `author`, `license`, `tags`, `permissions`, and `role` no longer trigger reinstall/update when edited on disk. This divergence is directly visible in the diff across `skill-graph.ts` and `skill-reconciler.ts`.

packages/daemon/src/pipeline/skill-graph.ts
NicholaiVogel: good catch. i changed `skillFingerprint()` to hash the full persisted frontmatter shape instead of the embedding-text subset, so edits to fields like `version`, `author`, `license`, `tags`, `permissions`, and `role` now trigger reconciliation correctly.

i also added a regression test that changes only `version` on disk and verifies `reconcileOnce()` performs an update and writes the new hash. targeted daemon tests and typecheck are passing on `4acf2bf8`.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -4,12 +4,15 @@
  * OllamaProvider uses the Ollama HTTP API, so we mock global fetch.
  */
 
-import { describe, it, expect, mock, beforeEach, afterEach } from "bun:test";
+import { afterEach, beforeEach, describe, expect, it, mock } from "bun:test";
```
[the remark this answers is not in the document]
NicholaiVogel: fixed in `4acf2bf8`. removed the unused `beforeEach` import while addressing the blocking review comment.

packages/daemon/src/pipeline/skill-graph.ts
```diff
@@ -280,11 +303,11 @@ export async function installSkillNode(
 				   dimensions = excluded.dimensions,
 				   source_id = excluded.source_id,
 				   chunk_text = excluded.chunk_text`,
-			).run(embId, hash, blob, embVec.length, entityId, embeddingText, now);
+			).run(embId, rawHash, blob, embVec.length, entityId, embeddingText, now);
```
PR-Reviewer-Ant: Blocking: switching `content_hash` to the raw frontmatter hash makes skill embeddings collide across distinct skill entities that share the same frontmatter. On `ON CONFLICT(content_hash)` this code overwrites `source_id`, `vector`, and `chunk_text`, so a later install can steal the embedding row from another skill/agent and replace its enriched text/vector. This reintroduces bad reconcile behavior for multi-agent or duplicate-skill cases and creates cross-entity data corruption. The hash needs to remain unique per installed skill entity (for example include `agentId`/entity identity), or the schema/update path needs to stop using a global content-hash uniqueness boundary here.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `4acf2bf8`

[Automated Review] Found one blocking correctness issue in the new skill embedding fingerprint/storage flow.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new code now keys `embeddings.content_hash` off raw frontmatter in `installSkillNode()`, while the row remains globally unique and still updates `source_id`, `vector`, and `chunk_text` on conflict. That means two skill entities with identical frontmatter but different owners/enrichment results can overwrite each other's embedding row. This is directly visible in the new `rawHash` write path in `packages/daemon/src/pipeline/skill-graph.ts` and is reachable because the install API already accepts `agentId`.

packages/daemon/src/pipeline/skill-graph.ts
NicholaiVogel: good catch. i changed skill embedding storage so `embeddings.content_hash` is now scoped by entity identity plus the raw frontmatter fingerprint, instead of using the raw frontmatter hash alone. that keeps reconcile stable while avoiding cross-entity collisions on the global `content_hash` uniqueness boundary.

i updated the reconciler to compare against the same entity-scoped hash, added regression coverage around the scoped hash behavior, and reran the targeted daemon tests plus typecheck. fixed in `85b00fbc`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `85b00fbc`

[Automated Bot Review] Found one test-isolation issue in the new Codex provider coverage.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new test directly mutates process.env.HOME and restores it with `process.env.HOME = undefined` when the variable was originally unset. In Node/Bun env handling, that assignment does not reliably remove the variable and can leave a literal `"undefined"` value behind for later tests, which would affect `homedir()`-based behavior. I did not run the suite here, so this is based on the changed test code rather than a reproduced failure.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -324,6 +329,67 @@ describe("createCodexProvider", () => {
...
+				stdout: streamFromString(
+					'{"type":"item.completed","item":{"type":"agent_message","text":"done"}}\n{"type":"turn.completed","usage":{"input_tokens":1,"output_tokens":1}}\n',
+				),
+				stderr: streamFromString(""),
+				exited: Promise.resolve(0),
+				kill() {},
+			};
+		}) as typeof Bun.spawn;
+
+		try {
```
PR-Reviewer-Ant, line 377: If `HOME` was originally unset, restoring it with `process.env.HOME = undefined` can leak `HOME="undefined"` into the rest of the test process instead of removing the variable. That would skew later `homedir()`/path behavior and make unrelated tests flaky. Use `delete process.env.HOME` in this branch.

NicholaiVogel: fixed in `7f31537c`. when the test overrides `HOME`, it now restores the unset case with `delete process.env.HOME` instead of assigning `undefined`, so we do not leak a bogus `HOME` value into later tests. reran the targeted daemon tests and typecheck after the change.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,43 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+} {
+	const root = join(homedir(), ".cache", "signet");
+	mkdirSync(root, { recursive: true });
+	const home = mkdtempSync(join(root, "codex-home-"));
+	const codexHome = join(home, ".codex");
+	mkdirSync(codexHome, { recursive: true });
+
+	const auth = join(homedir(), ".codex", "auth.json");
+	if (existsSync(auth)) {
+		cpSync(auth, join(codexHome, "auth.json"));
```
PR-Reviewer-Ant: This change duplicates the user's Codex auth material onto disk in a second location (`~/.cache/signet/codex-home-*`). Cleanup only happens through the subprocess-exit path, so a daemon crash, kill -9, or host restart can leave stale auth copies behind indefinitely. That expands secret persistence compared with the pre-PR behavior and is worth tightening before relying on this as a 'sterile' execution path.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `7f31537c`

[Automated Review] Found one security issue in the new Codex temp-home isolation flow.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new `createSterileCodexEnv()` path in `packages/daemon/src/pipeline/provider.ts` explicitly copies `~/.codex/auth.json` into a fresh temp directory under `~/.cache/signet/` and only removes it via best-effort process-exit cleanup. That behavior is directly visible in the diff. I did not reproduce a crash/abrupt-exit case here, so the residual-secret risk assessment is based on the code path rather than a runtime trace.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,43 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 	};
 }
 
+function createSterileCodexEnv(baseEnv: Record<string, string | undefined>): {
+	readonly env: Record<string, string | undefined>;
```
PR-Reviewer-Ant, line 202: This changes Codex extraction from a pure subprocess spawn into a path that must be able to create `~/.cache/signet/codex-home-*` on the daemon host. In service/container setups with a read-only or non-writable home directory, `mkdirSync`/`mkdtempSync` will now fail before Codex even starts, which is a functional regression from the previous implementation. Using `tmpdir()` or a fallback path is safer here.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1206,8 +1244,11 @@ export function createCodexProvider(config?: Partial<CodexProviderConfig>): LlmP
 				"exec",
```
PR-Reviewer-Ant, line 1248: This is a daemon behavior change covered by the repository's explicit Rust parity rule, but the PR marks Rust parity as N/A and no matching `packages/daemon-rs/` change is shown. The shadow daemon will now diverge on Codex provider launch semantics (`--ephemeral`, Signet MCP disabled, sterile home) unless the Rust implementation is updated in lockstep.

packages/daemon/src/pipeline/skill-reconciler.ts
```diff
@@ -123,21 +123,15 @@ export async function reconcileOnce(deps: ReconcilerDeps): Promise<{
 				const storedEmb = deps.accessor.withReadDb(
```
PR-Reviewer-Ant, line 123: The raw-frontmatter hash switch also changes daemon reconciliation behavior, but no corresponding `packages/daemon-rs/` update is shown despite the repo's required daemon parity contract. If the Rust shadow still keys skill updates off enriched embedding text, this PR will introduce persistent shadow divergences exactly in the loop this fix is trying to stabilize.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `bc2f730a`

[Automated Review] Found two blocking regressions and one parity gap that should be addressed before human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The writable-home regression is directly visible in `createSterileCodexEnv()` now calling `mkdirSync`/`mkdtempSync` under `homedir()/.cache/signet` before every Codex spawn. The Rust parity gap is also explicit from the daemon-only behavior changes in this diff plus the repository's stated shadow-parity rule, but I do not have the corresponding `packages/daemon-rs/` files in the provided context to verify whether a matching change exists elsewhere.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `f9531ff5`. the sterile Codex home now lives under `tmpdir()` instead of `~/.cache`, and `auth.json` is linked into that temp home instead of copied. that removes the duplicate on-disk auth blob while still letting Codex see the login material it expects. targeted daemon tests and typecheck passed after the change.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `f9531ff5`. the sterile Codex home now uses `tmpdir()` rather than a path under the daemon user's home, so provider-spawned Codex runs no longer depend on a writable home directory just to boot the isolation path. targeted daemon tests and typecheck passed after the update.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: reviewed and clarified in the PR notes. this PR's Codex-provider change is N/A for Rust shadow parity because the Rust daemon does not currently execute Codex extraction providers at all. `packages/daemon-rs/crates/signet-daemon/src/main.rs` still asserts `worker_supports_extraction_provider("codex") == false`, so there is no corresponding Rust launch path to update here.

packages/daemon/src/pipeline/skill-reconciler.ts
NicholaiVogel: reviewed and clarified in the PR notes. this reconciliation change is also N/A for Rust parity in the current shadow setup because the Rust daemon does not implement the procedural skill reconciler / `skill_meta` sync path changed in this PR. there is no corresponding Rust reconciliation loop to update yet, so this should not create a shadow divergence on an existing Rust path.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,43 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+} {
+	const root = join(tmpdir(), "signet-codex-home-");
+	mkdirSync(root, { recursive: true });
+	const home = mkdtempSync(root);
+	const codexHome = join(home, ".codex");
+	mkdirSync(codexHome, { recursive: true });
+
+	const auth = join(homedir(), ".codex", "auth.json");
+	if (existsSync(auth)) {
+		symlinkSync(auth, join(codexHome, "auth.json"));
```
PR-Reviewer-Ant: Blocking: the new auth-preservation path unconditionally creates a filesystem symlink for `auth.json`. That is not a safe cross-platform substitute for the previous copy behavior: on Windows, file symlink creation often requires Developer Mode or elevated privileges, and `symlinkSync()` will throw before the Codex subprocess starts. Since this provider already contains Windows-specific spawn handling, this change can regress Codex-backed extraction on supported Windows daemon installs. Use a Windows-safe fallback (for example copy on win32, or catch `EPERM`/`EINVAL` and fall back) instead of requiring symlink capability.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f9531ff5`

pr-reviewer automated bot review
Found one blocking platform regression in the new Codex sterile-home path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new behavior is explicit in the diff: `createSterileCodexEnv()` now unconditionally uses `symlinkSync()` for `~/.codex/auth.json` before every Codex provider spawn. This is directly reachable from `createCodexProvider()` and affects the daemon's cross-platform subprocess path. On Windows, creating file symlinks commonly fails without Developer Mode or elevated privileges, so this can make Codex extraction fail before the subprocess is even launched.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `27221013`. auth preservation now uses a Windows-safe fallback path: hard link on win32, symlink elsewhere, and copy only if linking fails. that keeps Codex provider startup from depending on symlink privileges while still avoiding duplicate auth blobs where the platform allows it. reran the targeted daemon tests and typecheck after the change.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,52 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+function createSterileCodexEnv(baseEnv: Record<string, string | undefined>): {
+	readonly env: Record<string, string | undefined>;
+	cleanup(): void;
+} {
+	const root = join(tmpdir(), "signet-codex-home-");
+	mkdirSync(root, { recursive: true });
+	const home = mkdtempSync(root);
+	const codexHome = join(home, ".codex");
+	mkdirSync(codexHome, { recursive: true });
+
```
PR-Reviewer-Ant, line 212: Blocking: the new sterile-home flow links `~/.codex/auth.json` into the temporary `CODEX_HOME` instead of giving the subprocess an isolated read-only credential view. On Unix the symlink points writes at the real auth file, and on Windows the hard link is the same underlying file. If the spawned Codex process refreshes or rewrites auth state, it can mutate or corrupt the user's live login material from what is supposed to be an isolated extraction run. This diverges from the PR's isolation goal and introduces a data-corruption/security risk.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `27221013`

pr-reviewer automated bot review
Found one blocking issue in the Codex sterile-home auth handling.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new `createSterileCodexEnv()` path in `packages/daemon/src/pipeline/provider.ts` explicitly exposes the real `~/.codex/auth.json` to provider-spawned Codex runs via `symlinkSync()` on Unix and `linkSync()` on Windows. Any write through that temp-home path will hit the original auth file, so the supposedly isolated extraction run can mutate or corrupt the user's live Codex login state. This behavior is directly reachable from `createCodexProvider()` in the current diff.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `04896bde`. the sterile Codex home now gets its own copied `auth.json` again, but that copy is made read-only inside the temp `CODEX_HOME`, and stale temp homes are swept on the next run. that keeps the extraction run from mutating the user's live Codex auth state while still preserving the isolation path. reran the targeted daemon tests and typecheck after the change.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,49 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
 	};
 }
 
+function createSterileCodexEnv(baseEnv: Record<string, string | undefined>): {
+	readonly env: Record<string, string | undefined>;
+	cleanup(): void;
+} {
+	const root = join(tmpdir(), "signet-codex-home");
```
PR-Reviewer-Ant, line 205: Blocking: `createSterileCodexEnv()` now deletes every existing `tmpdir()/signet-codex-home/home-*` before creating the next temp home. `createCodexProvider()` is still allowed to run multiple subprocesses concurrently (`SIGNET_MAX_LLM_CONCURRENCY`, default 4), so one Codex extraction can remove another in-flight extraction's `HOME`/`CODEX_HOME` mid-run. That turns the new isolation fix into a cross-request race and can cause intermittent auth/config read failures or corrupted execution state under load.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `04896bde`

pr-reviewer automated bot review
Found one blocking race in the new Codex sterile-home setup. The PR also fixes the previously reported hash-collision and linked-auth issues, but this new cleanup path can now break concurrent Codex provider runs.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The regression is explicit in the current diff for packages/daemon/src/pipeline/provider.ts: createSterileCodexEnv() now scans tmpdir()/signet-codex-home and rmSync()s every prior home-* directory before creating a new one. createCodexProvider() still runs under a global semaphore that allows up to 4 concurrent subprocesses, so a second spawn can delete the first spawn's HOME/CODEX_HOME while that Codex process is still running.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `c88cf702`.

i removed the eager sweep of sibling `tmpdir()/signet-codex-home/home-*` directories from `createSterileCodexEnv()`. each Codex provider run now creates its own unique temp home and only cleans up its own directory on exit, so concurrent extraction runs can no longer delete each other's `HOME` / `CODEX_HOME` mid-flight.

i also added a regression test that pre-creates a sibling sterile home and verifies a Codex run leaves it intact while spawning. targeted daemon tests and `bun run typecheck` are passing after the change.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,45 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
+	mkdirSync(root, { recursive: true });
+	const home = mkdtempSync(join(root, "home-"));
+	const codexHome = join(home, ".codex");
+	mkdirSync(codexHome, { recursive: true });
```
PR-Reviewer-Ant, line 209: Blocking: the sterile-home copy path ignores an existing `CODEX_HOME` and always reads from `~/.codex`. That means daemon environments using a custom Codex home will spawn provider runs without the user's real auth/version state, which contradicts the PR's stated goal of preserving auth while isolating the run. Please source auth/version from the effective live Codex home (`process.env.CODEX_HOME` if set, otherwise `join(homedir(), '.codex')`).

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `c88cf702`

[Automated Review] Found one blocking environment-handling regression in the new Codex sterile-home path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The regression is directly visible in `packages/daemon/src/pipeline/provider.ts`: `createSterileCodexEnv()` now hardcodes its auth/version copy source to `join(homedir(), '.codex', ...)`, while `createCodexProvider()` builds the sterile environment from the current process environment. If the daemon is running with a non-default `CODEX_HOME`, the live Codex CLI state is no longer copied into the sterile temp home, so provider-spawned Codex runs can lose auth/version state even though the PR claims those runs preserve Codex auth in isolation.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `a1b19b9f`.

i changed `createSterileCodexEnv()` to source auth/version from the effective live Codex home, `baseEnv.CODEX_HOME` when present, otherwise `join(homedir(), ".codex")`. that keeps provider-spawned Codex runs aligned with daemon environments that use a custom `CODEX_HOME`.

i also tightened the regression test to set a custom live `CODEX_HOME` and verify the sterile temp home copies auth/version from there, not from `~/.codex`. targeted daemon tests and `bun run typecheck` are passing after the change.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -197,6 +198,46 @@ function spawnHidden(cmd: string[], options?: { env?: Record<string, string | un
...
```
PR-Reviewer-Ant, line 209: The sterile-home copy source still does not match the process's effective Codex home in all cases. When `CODEX_HOME` is unset but the daemon is running with a custom `HOME`, this falls back to `homedir()` instead of `${baseEnv.HOME}/.codex`, so provider-spawned Codex runs can lose auth/version state despite the PR claiming those runs preserve auth in isolation. The fallback should derive from the effective environment (`baseEnv.HOME`) before consulting the OS home directory.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `5cc6a676`

pr-reviewer automated bot review
Found one environment-handling regression that still breaks the PR's claimed Codex auth preservation in some daemon setups.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The regression is directly visible in packages/daemon/src/pipeline/provider.ts. createSterileCodexEnv() now honors CODEX_HOME, but when CODEX_HOME is unset it still falls back to node:os homedir() instead of the daemon's effective HOME from baseEnv. In service/container environments that launch the daemon with a custom HOME but no CODEX_HOME, the live Codex state lives under $HOME/.codex, so this code copies from the wrong directory and the spawned Codex run loses the auth/version state that the PR description says should be preserved.

packages/daemon/src/pipeline/provider.ts
NicholaiVogel: fixed in `262d3e65`.

i changed the fallback chain in `createSterileCodexEnv()` to derive the live Codex home from the effective environment first: `baseEnv.CODEX_HOME`, then `join(baseEnv.HOME ?? homedir(), ".codex")`. that keeps sterile provider runs aligned when the daemon is launched with a custom `HOME` but no explicit `CODEX_HOME`.

targeted daemon tests and `bun run typecheck` are passing after the change.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -4,12 +4,15 @@
  * OllamaProvider uses the Ollama HTTP API, so we mock global fetch.
  */
 
-import { describe, it, expect, mock, beforeEach, afterEach } from "bun:test";
+import { afterEach, describe, expect, it, mock } from "bun:test";
+import { existsSync, lstatSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { homedir, tmpdir } from "node:os";
```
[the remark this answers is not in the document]
NicholaiVogel: fixed in `262d3e65`. removed the now-unused `homedir` import from `packages/daemon/src/pipeline/provider.test.ts` while addressing the custom-home fallback review comment.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `94f0c486`

[Automated Review by pr-reviewer] No additional issues found in the latest diff; the current implementation matches the PR description and is ready for human review.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The reviewed diff directly covers the two claimed fixes: skill reconciliation now compares an entity-scoped raw-frontmatter hash instead of enriched embedding text, and the Codex provider now spawns with a temp HOME/CODEX_HOME, copied auth/version state, --ephemeral, and Signet MCP disabled. The previously reported concurrent-home deletion and CODEX_HOME/HOME source regressions are resolved in this revision. Confidence is not high because the mounted review context does not include adjacent runtime modules beyond the provided PR materials, so I could not independently inspect every downstream consumer of the new hashes or the Codex CLI behavior outside these touched paths.

PR #413, opened 2026-03-30
feat: skill usage analytics — invocation tracking, API, CLI, and dashboard

aaf2tbz:
## Summary

Adds skill usage analytics: invocation tracking, aggregated API, CLI command, and constellation-style dashboard panel. Per-agent task ownership and agent-scoped task endpoints.

## Changes

- `packages/core/src/migrations/053-skill-invocations.ts` — skill_invocations table
- `packages/core/src/migrations/054-task-agent-id.ts` — adds agent_id to scheduled_tasks (BREAKING for multi-agent, see Migration Notes)
- `packages/daemon/src/routes/skill-analytics.ts` — analytics API endpoints (agent-scoped reads, strict ISO validation, case-normalized skill lookups)
- `packages/daemon/src/routes/skill-analytics.test.ts` — SQL-level tests (10 tests)
- `packages/daemon/src/routes/skill-analytics-routes.test.ts` — HTTP route-level tests (7 tests)
- `packages/daemon/src/routes/task-scoping.test.ts` — task agent-scoping SQL tests (10 tests)
- `packages/daemon/src/scheduler/worker.ts` — scheduler invocation recording (reads task.agent_id)
- `packages/daemon/src/daemon.ts` — all task CRUD endpoints scoped by agent_id in SQL, task creation via resolveScopedAgent(), PATCH supports agent reassignment (validated through resolveScopedAgent), API task-run invocation recording, stream + runs endpoints agent-scoped
- `packages/cli/src/commands/skill.ts` — `signet skill analytics [--skill] [--since] [--agent] [--json]`
- `packages/cli/dashboard/src/lib/components/skills/SkillUsagePanel.svelte` — dashboard panel ($effect + agentId)
- `packages/cli/dashboard/src/lib/stores/skill-analytics.svelte.ts` — store with request sequencing
- `packages/cli/dashboard/src/lib/stores/mcp-analytics.svelte.ts` — added request sequencing
- `docs/specs/planning/skill-usage-analytics.md` — spec

### Task ownership model
- **Tasks are agent-owned.** All task CRUD/run/stream/runs endpoints scope by `agent_id` in SQL — no existence oracle (non-owned tasks return 404 identical to missing).
- **Task runs are task-owned, not agent-owned.** Runs belong to their parent task. On agent reassignment via PATCH, the full run history transfers with the task. This is intentional — runs are operational artifacts of the task itself.
- **Callers that omit `agent_id` default to `'default'`** via `resolveScopedAgent()`. No existing client breaks because all pre-migration callers were implicitly the default agent.

### Migration 054 (BREAKING for multi-agent)
- Adds `agent_id TEXT NOT NULL DEFAULT 'default'` to `scheduled_tasks`.
- **All existing rows are assigned to `'default'`** because the pre-migration schema had no ownership column — there is no prior data to infer the correct agent.
- **Single-agent (local) deployments are unaffected.**
- **Multi-agent deployments:** after upgrade, tasks are only visible to the default agent. Operators must reassign via PATCH: `PATCH /api/tasks/:id?agent_id=default` with `body: { agentId: "<target>" }`. This is an admin operation done in local mode or with default-agent credentials.

## Type

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`

## PR Readiness (MANDATORY)

## Migration Notes

- [x] Migration is idempotent

**Migration 053** creates `skill_invocations` table. Idempotent (`CREATE TABLE/INDEX IF NOT EXISTS`).

**Migration 054** adds `agent_id TEXT NOT NULL DEFAULT 'default'` to `scheduled_tasks`. Idempotent (checks PRAGMA table_info before ALTER). **BREAKING for multi-agent** — see Design Decisions above.

Rust daemon parity: N/A (Rust rewrite has not reached migration layer).

Rollback 053: `DROP TABLE skill_invocations; DELETE FROM schema_migrations WHERE version = 53;`
Rollback 054: SQLite does not support DROP COLUMN — recreate table without agent_id or accept the column. `DELETE FROM schema_migrations WHERE version = 54;`

## Testing

- [x] `bun test` passes (27 SQL + route + scoping tests across 3 test files)

## AI disclosure


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat: skill usage analytics — invocation tracking, API, CLI, and dashboard` (commit `c85fce0a`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7587,10 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
```
PR-Reviewer-Ant, line 7628: Blocking: this path records every API-triggered invocation against `taskAgentId = "default"` and updates `skill_meta` for the default agent only. That contradicts the PR’s claimed auth-aware agent scoping and makes the new dashboard `agentId` switching incorrect for any non-default agent: analytics queries for another agent will legitimately return zero even when that agent ran the task.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -310,6 +310,45 @@ export async function executeTask(
 		timestamp: new Date().toISOString(),
```
PR-Reviewer-Ant, line 311: Blocking: the scheduler write path also hardcodes `agent_id = 'default'` and increments usage only on the default agent’s skill entity. Given the repo’s universal agent-scoping invariant, this bakes cross-agent attribution drift into persisted analytics data instead of merely deferring UI support.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,208 @@
...
+	return sorted[Math.max(0, idx)] ?? 0;
+}
+
+export function mountSkillAnalyticsRoutes(app: Hono, authMode: AuthMode = "local"): void {
+	app.get("/api/skills/analytics", (c) => {
+		const scoped = resolveScopedAgent(
+			c.get("auth")?.claims ?? null,
+			authMode,
+			c.req.query("agent_id"),
+		);
```
PR-Reviewer-Ant, line 63: Blocking: writes normalize `skill_name` to lowercase, but both analytics endpoints compare the raw `skill` path/query value without normalizing it first. Any mixed-case lookup from the CLI or future UI detail view will miss existing rows and report false zero-usage results.

packages/daemon/src/routes/skill-analytics.test.ts
```diff
@@ -0,0 +1,110 @@
+import { Database } from "bun:sqlite";
```
PR-Reviewer-Ant, line 1: Warning: this file is named as a route test, but it only runs direct SQL against `skill_invocations`. It does not mount the Hono routes, does not assert the claimed 400 on invalid `since`, and does not cover auth-scoped agent resolution. That diverges from the PR description’s test plan and leaves the new API contract unverified.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 7628: Blocking: this hardcodes `taskAgentId = "default"` for manual task runs, but the new analytics endpoints read back with `resolveScopedAgent(..., agent_id)`. For any non-default agent, invocations triggered from that agent's UI/API will be written under `default` and then disappear from that agent's analytics. The PR description and readiness checklist both claim agent scoping was validated, but this write path breaks that contract.

packages/daemon/src/scheduler/worker.ts
PR-Reviewer-Ant, line 311: Blocking: scheduled-task analytics are also written with `agent_id = 'default'` and update `skill_meta` only for the default agent. That means the new feature cannot produce correct per-agent usage data even though the API is scoped by agent. At minimum this diverges from the repo invariant that user-data reads/writes must thread `agent_id`, and it will silently skew counts toward the default tenant.

packages/cli/dashboard/src/lib/stores/skill-analytics.svelte.ts
```diff
@@ -0,0 +1,24 @@
+import { type SkillAnalyticsSummary, getSkillAnalytics } from "$lib/api";
+
+export const skillAnalytics = $state({
+	data: null as SkillAnalyticsSummary | null,
+	loading: false,
+	error: null as string | null,
+});
+
```
PR-Reviewer-Ant, line 8: Warning: this is a singleton store with no request sequencing or cancellation. The new `SkillUsagePanel` refetches on `agentId` changes and on a polling interval, so a slower response for agent A can arrive after switching to agent B and overwrite the shared store with stale data. The same pattern is now used for agent-aware MCP analytics too, so the dashboard can show the wrong tenant's stats after quick agent switches.

packages/daemon/src/daemon.ts
```diff
@@ -7424,6 +7428,7 @@ app.post("/api/tasks", async (c) => {
 			nextRunAt,
```
PR-Reviewer-Ant, line 7443: Blocking: this write path trusts `body.agentId` and persists it directly into `scheduled_tasks`. The new analytics read routes correctly use `resolveScopedAgent(...)`, but task creation does not. In an authenticated multi-agent deployment, a caller can create work attributed to another agent and pollute that agent's analytics/use counts. This needs the same auth-aware agent resolution on the mutation path instead of accepting an arbitrary body field.

packages/daemon/src/routes/skill-analytics.test.ts
```diff
@@ -0,0 +1,167 @@
...
+		const all = db
+			.prepare("SELECT COUNT(*) as c FROM skill_invocations WHERE agent_id = 'default'")
+			.get() as { c: number };
+		expect(all.c).toBe(2);
+
+		const filtered = db
+			.prepare(
+				"SELECT COUNT(*) as c FROM skill_invocations WHERE agent_id = 'default' AND created_at >= datetime(?)",
+			)
+			.get("2025-01-01T00:00:00.000Z") as { c: number };
```
PR-Reviewer-Ant, line 151: These tests do not exercise the HTTP contract they are claiming to cover. They reimplement the SQL and date-validation logic locally instead of hitting `mountSkillAnalyticsRoutes()`, so regressions in `resolveScopedAgent`, 400 handling for bad `since`, or timeline response shaping would still pass. Given the PR description explicitly claims agent scoping and input validation, this should include endpoint-level regression coverage.

packages/daemon/src/daemon.ts
```diff
@@ -7583,6 +7597,7 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillName = typeof task.skill_name === "string" ? task.skill_name : null;
```
PR-Reviewer-Ant, line 7583: Blocking: this manual task-run endpoint now participates in per-agent analytics, but it still does not enforce agent scope before running the task. A caller can hit `/api/tasks/:id/run` with another agent's task id, execute that agent's scheduled skill, and write invocation data under that agent. Since this PR introduces agent-aware task creation and analytics, the run path needs the same `resolveScopedAgent()` / ownership check rather than trusting task ids globally.

PR-Reviewer-Ant, line 7628: Warning: `task.agent_id` falls back to `'default'` here. That silently collapses attribution if the selected row is missing `agent_id` for any reason, and it directly contradicts the PR note that there is 'No hardcoded `default` on any write path'. Because migration 054 makes this column non-null with a default, this path should reject/diagnose a missing `agent_id` instead of recording analytics under the wrong agent.

packages/daemon/src/daemon.ts
```diff
@@ -7374,6 +7378,15 @@ app.post("/api/tasks", async (c) => {
 	const body = await c.req.json();
```
PR-Reviewer-Ant, line 7384: Blocking: this PR starts assigning `agent_id` ownership to scheduled tasks and enforces it for `POST /api/tasks/:id/run`, but there are no matching scoping changes for the rest of the task endpoints in this file. Unless the unchanged `/api/tasks` read/update/delete handlers already thread `agent_id`, tasks remain cross-agent visible/mutable even though analytics now treats them as owned per agent. That violates the repo's agent-scoping invariant and creates a real data-isolation leak.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,208 @@
...
+				}
+				const where = conditions.join(" AND ");
+
+				const totals = db
+					.prepare(
+						`SELECT COUNT(*) as total, COALESCE(SUM(CASE WHEN success = 1 THEN 1 ELSE 0 END), 0) as successes
+					 FROM skill_invocations WHERE ${where}`,
+					)
+					.get(...params) as { total: number; successes: number } | undefined;
+
```
PR-Reviewer-Ant, line 94: The summary query aggregates purely from `skill_invocations` and never joins installed skills / `skill_meta`, so the new dashboard panel can surface historic or now-uninstalled skills. That diverges from the spec and PR description, which both describe the panel as showing usage for installed skills.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,208 @@
...
+	readonly sources: readonly { readonly source: string; readonly count: number }[];
+	readonly timeline: readonly { readonly date: string; readonly count: number }[];
+	readonly latency: { readonly p50: number; readonly p95: number };
+}
+
+function clampPositiveInt(value: unknown, fallback: number, min: number, max: number): number {
+	const n = Number(value);
+	if (!Number.isFinite(n)) return fallback;
+	return Math.max(min, Math.min(max, Math.round(n)));
+}
```
PR-Reviewer-Ant, line 41: The PR text says `since` is validated as ISO 8601, but `isValidIsoDate()` only checks whether `new Date(value)` parses. That accepts many non-ISO date strings depending on runtime parsing rules, so the endpoint does not actually enforce the contract it claims.

packages/daemon/src/daemon.ts
```diff
@@ -7437,12 +7455,16 @@ app.post("/api/tasks", async (c) => {
 app.get("/api/tasks/:id", (c) => {
```
PR-Reviewer-Ant, line 7457: These task-by-id handlers fetch `scheduled_tasks` by bare `id` before enforcing agent scope. In multi-agent/authenticated setups this leaks cross-agent task existence via `404` vs `403` differences and diverges from the repo's scoping invariant. The read/update/delete paths should scope in SQL (or otherwise avoid distinguishable existence checks before authorization).

packages/cli/src/commands/skill.ts
```diff
@@ -270,4 +270,65 @@ export function registerSkillCommands(program: Command, deps: SkillDeps): void {
 			}
 			console.log(readFileSync(path, "utf-8"));
 		});
+
+	// signet skill analytics
+	skillCmd
+		.command("analytics")
+		.description("Show skill usage analytics")
+		.option("--skill <name>", "Filter by skill name")
```
PR-Reviewer-Ant, line 278: The new `signet skill analytics` command cannot target a non-default agent. It only forwards `skill` and `since`, but the daemon route scopes by `agent_id`. In local/CLI usage that means this command will always read the default agent's analytics, which diverges from the PR's 'full per-agent attribution' claim and from the repo's agent-scoping invariant.

packages/daemon/src/routes/skill-analytics.test.ts
```diff
@@ -0,0 +1,171 @@
...
+	let db: Database;
+
+	beforeEach(() => {
+		db = new Database(":memory:");
+		runMigrations(db);
+	});
+
+	afterEach(() => {
+		if (db) db.close();
+	});
```
PR-Reviewer-Ant, line 15: These tests validate read-side aggregation by inserting directly into `skill_invocations`, but they never cover the new write paths in `scheduler/worker.ts` and `/api/tasks/:id/run` that are responsible for recording invocations and bumping `skill_meta.use_count`. As written, the PR can pass with broken instrumentation, so the implementation is not actually validated against the behavior claimed in the PR description.

packages/daemon/src/daemon.ts
```diff
@@ -7347,8 +7351,11 @@ app.get("/api/tasks/:id/stream", (c) => {
 	});
```
PR-Reviewer-Ant, line 7353: Blocking: this PR claims all task endpoints are now scoped by `agent_id`, but `/api/tasks/:id/stream` immediately above this hunk is still unscoped. That leaves a cross-agent existence/data leak if another agent's task ID is known, and it contradicts the stated 'no existence oracle' goal for task APIs.

packages/cli/dashboard/src/lib/components/tabs/SkillsTab.svelte
```diff
@@ -317,6 +319,9 @@ onMount(() => {
 			{sk.searching ? "Searching..." : "Loading..."}
```
PR-Reviewer-Ant, line 319: The panel is rendered only when `sk.installed.length > 0`. That diverges from the PR description's claim that analytics are historical and include invocations from skills that have since been uninstalled: an agent with historical usage but zero currently installed skills will never see the dashboard analytics panel.

docs/specs/dependencies.yaml
```diff
@@ -698,17 +698,31 @@ specs:
 
 
   - id: mcp-cli-bridge-and-usage-analytics
-    status: planning
+    status: approved
     path: docs/specs/planning/mcp-cli-bridge-and-usage-analytics.md
     hard_depends_on:
       - signet-runtime
     soft_depends_on: []
```
PR-Reviewer-Ant, line 705: This change adds a `blocks: [skill-usage-analytics]` edge for `mcp-cli-bridge-and-usage-analytics`, but `docs/specs/INDEX.md` was not updated to reflect that dependency edge. The repo contract requires `INDEX.md` and `dependencies.yaml` to stay in sync for spec changes, so this introduces spec drift in the same PR.

packages/daemon/src/daemon.ts
```diff
@@ -7436,8 +7454,12 @@ app.post("/api/tasks", async (c) => {
 // Get a single task + recent runs
 app.get("/api/tasks/:id", (c) => {
 	const taskId = c.req.param("id");
+	const scoped = resolveScopedAgent(c.get("auth")?.claims ?? null, authConfig.mode, c.req.query("agent_id"));
+	if (scoped.error) return c.json({ error: scoped.error }, 403);
 
-	const task = getDbAccessor().withReadDb((db) => db.prepare("SELECT * FROM scheduled_tasks WHERE id = ?").get(taskId));
+	const task = getDbAccessor().withReadDb((db) =>
+		db.prepare("SELECT * FROM scheduled_tasks WHERE id = ? AND agent_id = ?").get(taskId, scoped.agentId),
+	);
```
PR-Reviewer-Ant, line 7465: Blocking: this changes existing task endpoints to require `agent_id` scoping from the request (`query` for GET/DELETE/RUN, `body.agentId` for PATCH). In local multi-agent use, callers that previously addressed a task by ID alone will now get 404s for any non-`default` task unless they were also updated to send the agent id. This PR updates analytics plumbing, but not the task clients or tests, so it introduces a cross-agent task management regression.

packages/cli/src/commands/skill.ts
```diff
@@ -270,4 +270,67 @@ export function registerSkillCommands(program: Command, deps: SkillDeps): void {
...
+		.option("--since <iso>", "Only include invocations after this ISO date")
+		.option("--agent <id>", "Agent ID to scope analytics (default: 'default')")
+		.option("--json", "Output as JSON")
```
PR-Reviewer-Ant, line 281: The new `--skill` flag still calls `/api/skills/analytics?...` and parses the aggregate summary shape. It never hits the dedicated `/api/skills/analytics/:skill` route added in this PR, so the implementation falls short of the claimed per-skill analytics surface and leaves that endpoint effectively unused outside tests.

aaf2tbz:
Commit `b2232cc5` addresses the remaining review comments:

**Fixed in this commit:**
- `/api/tasks/:id/stream` now scoped by `agent_id` via `resolveScopedAgent` (was unscoped — comment 3012324800)
- `/api/tasks/:id/runs` now scoped by `agent_id` with ownership check before returning runs (was unscoped)
- CLI `--skill` flag now hits the dedicated `/api/skills/analytics/:skill` detail route with sources+timeline output (comments 3012326837, 3012268501)
- Dashboard `SkillUsagePanel` no longer gated on `sk.installed.length > 0` — shows historical analytics even with zero installed skills (comment 3012324805)

**Already fixed in prior commits (ant reviewed stale commits):**
- Write paths use `task.agent_id` from DB row, not hardcoded `default` (fixed in dcfdaf0f+)
- Read paths normalize skill name via `.toLowerCase()` (fixed in b373baea+)
- Store has request sequence counters to prevent stale overwrites (fixed in cebcfbed+)
- CLI has `--agent` flag forwarding `agent_id` to API (fixed in b83fbf4c)
- Task creation uses `resolveScopedAgent()` (fixed in 4f56e7bf)
- Task-by-id endpoints scope in SQL with `AND agent_id = ?` (fixed in 574aabf1)
- `task.agent_id` asserted, not silently defaulted (fixed in 60adabc0)
- ISO validation uses regex `ISO_DATE_RE` + Date parse (fixed in 574aabf1)

**Not applicable / by design:**
- `dependencies.yaml` / `INDEX.md` are in sync: `skill-usage-analytics` has `soft_depends_on: [mcp-cli-bridge-and-usage-analytics]` matching the `blocks` edge
- Analytics aggregates from `skill_invocations` intentionally — historical data for uninstalled skills is a stated Phase 1 feature, not a bug
- Task endpoint scoping is the correct behavior for multi-agent deployments; existing local callers use `default` agent via `resolveScopedAgent` fallback

PR description updated to reflect all of the above.

packages/core/src/migrations/054-task-agent-id.ts
```diff
@@ -0,0 +1,17 @@
...
+ * to the correct agent instead of hardcoding 'default'.
+ */
+
+import type { MigrationDb } from "./index";
+
+export function up(db: MigrationDb): void {
+	const cols = db.prepare("PRAGMA table_info(scheduled_tasks)").all() as ReadonlyArray<Record<string, unknown>>;
+	const colNames = new Set(cols.flatMap((c) => (typeof c.name === "string" ? [c.name] : [])));
+
+	if (!colNames.has("agent_id")) {
```
PR-Reviewer-Ant, line 29: This migration forces every existing `scheduled_tasks` row to `agent_id = 'default'`. In a workspace that already uses multiple agents, that reassigns ownership of all legacy tasks to the default agent, so other agents lose visibility/control over their tasks after upgrade and future skill analytics will be misattributed. That is a breaking data migration, not just a default for new rows.

packages/daemon/src/daemon.ts
```diff
@@ -7251,9 +7255,11 @@ app.post("/api/update/run", async (c) => {
 
 app.get("/api/tasks/:id/stream", (c) => {
 	const taskId = c.req.param("id");
+	const scoped = resolveScopedAgent(c.get("auth")?.claims ?? null, authConfig.mode, c.req.query("agent_id"));
```
PR-Reviewer-Ant, line 7258: This PR changes the scoping contract for all task endpoints (`list`, `detail`, `patch`, `delete`, `run`, `stream`, `runs`) but the added tests only cover the new analytics routes. Because the PR description explicitly claims full agent-scoped task isolation and no existence oracle, these route changes need regression coverage; otherwise a missed caller/path here can silently reintroduce cross-agent leakage or false 404s.

packages/daemon/src/daemon.ts
```diff
@@ -7461,9 +7485,11 @@ app.get("/api/tasks/:id", (c) => {
 app.patch("/api/tasks/:id", async (c) => {
 	const taskId = c.req.param("id");
 	const body = await c.req.json();
```
PR-Reviewer-Ant, line 7487: The new PATCH scoping makes agent reassignment fail. `resolveScopedAgent(..., body.agentId)` is used before loading the existing row, and the lookup is `SELECT * FROM scheduled_tasks WHERE id = ? AND agent_id = ?`. If an existing task belongs to `default` and the client tries to reassign it to `agent-a` by sending `agentId: 'agent-a'`, this query returns 404 before any update can happen. That contradicts the migration note in this PR (`operators should reassign tasks via the PATCH endpoint after upgrade`) and leaves pre-migration multi-agent tasks stuck on `default`.

packages/core/src/migrations/054-task-agent-id.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * Migration 054: Add agent_id to scheduled_tasks
+ *
+ * Allows skill invocation analytics to attribute task runs
+ * to the correct agent instead of hardcoding 'default'.
+ *
+ * Existing rows receive agent_id = 'default' because pre-migration
+ * tasks had no agent ownership column — there is no prior data to
+ * infer the correct agent. In single-agent (local) deployments this
+ * is accurate. In multi-agent deployments that predate this migration,
```
PR-Reviewer-Ant: This migration note promises that existing multi-agent tasks can be repaired via PATCH after upgrade, but the current PATCH implementation cannot actually move a task between agents because it scopes the read by the target `agentId`. Either the route needs a dedicated reassignment path, or this compatibility note is inaccurate. As written, the rollout story in the PR description is not achieved.

packages/core/src/migrations/054-task-agent-id.ts
```diff
@@ -0,0 +1,26 @@
...
+ * the lookup by the caller's agent_id.
+ */
+
+import type { MigrationDb } from "./index";
+
+export function up(db: MigrationDb): void {
+	const cols = db.prepare("PRAGMA table_info(scheduled_tasks)").all() as ReadonlyArray<Record<string, unknown>>;
+	const colNames = new Set(cols.flatMap((c) => (typeof c.name === "string" ? [c.name] : [])));
+
+	if (!colNames.has("agent_id")) {
```
PR-Reviewer-Ant, line 29: Blocking: this migration assigns every legacy `scheduled_tasks` row to `agent_id = 'default'`, but the same PR changes all task endpoints to require `agent_id` matches in SQL. On an existing multi-agent install, any task that originally belonged to a non-default agent becomes invisible/uneditable/unrunnable to its owner immediately after upgrade, and later analytics are attributed to the wrong agent unless the operator performs a manual DB rewrite. That is a breaking behavior change, not just a migration note.

packages/core/src/migrations/054-task-agent-id.ts
```diff
@@ -0,0 +1,29 @@
...
```
PR-Reviewer-Ant, line 29: This is a breaking data migration for existing multi-agent installs. Every pre-existing task is forced to `agent_id = 'default'`, and this PR immediately scopes all task CRUD/run endpoints by `agent_id`, so non-default agents lose visibility/execution of their existing tasks right after upgrade. The PR notes manual SQL repair, but there is no automatic/backfill path and the new scoped PATCH route cannot reassign ownership afterward. This needs a compatibility/backfill strategy before merge.

aaf2tbz:
## Re: Migration 054 breaking change for multi-agent

Ant has flagged this across multiple reviews. Addressing directly:

**Why automatic backfill is impossible:** Pre-migration `scheduled_tasks` has no `agent_id` column — there is zero data to infer which agent owned which task. `DEFAULT 'default'` is the only safe assignment.

**The backfill strategy IS the documented SQL update.** This is standard for schema migrations that add ownership columns to previously-unscoped tables. The migration JSDoc now explicitly calls this out as BREAKING for multi-agent and provides the exact SQL.

**Single-agent (local) deployments** — which are all current production deployments — are completely unaffected. Multi-agent is experimental and operators are expected to handle data migrations.

**Scope:** A CLI `signet task reassign` command is a reasonable future addition but is out of scope for the analytics feature PR.

packages/daemon/src/daemon.ts
```diff
@@ -7495,12 +7522,15 @@ app.patch("/api/tasks/:id", async (c) => {
 		return c.json({ error: "skillMode must be 'inject' or 'slash' when skillName is set" }, 400);
```
PR-Reviewer-Ant, line 7523: The PATCH path lets a caller set `body.agentId` directly without running it through `resolveScopedAgent()`. In authenticated multi-agent deployments, an agent that owns a task can reassign it into another agent's namespace by sending any `agentId`, which breaks the PR's claimed task isolation and creates a cross-tenant write path.

packages/daemon/src/daemon.ts
```diff
@@ -7512,6 +7542,7 @@ app.patch("/api/tasks/:id", async (c) => {
 			nextRunAt,
 			skillName,
```
PR-Reviewer-Ant, line 7549: This `UPDATE scheduled_tasks ... WHERE id = ?` is no longer scoped by `agent_id`, even though the PR description says all task endpoints are scoped in SQL. That makes the PATCH flow race-prone: after the scoped read succeeds, a concurrent reassignment can move the task to another agent and this update will still apply. The write itself needs the same `AND agent_id = ?` ownership guard.

packages/daemon/src/daemon.ts
```diff
@@ -7461,9 +7485,12 @@ app.get("/api/tasks/:id", (c) => {
 app.patch("/api/tasks/:id", async (c) => {
```
PR-Reviewer-Ant, line 7485: The documented migration recovery path for existing multi-agent tasks appears to break under auth. This PATCH route first resolves the current owner from `c.req.query("agent_id")` and only then allows reassignment via `body.agentId`. After migration 054, legacy rows are forced to `agent_id='default'`, so a non-default agent would need permission to act as `default` just to reclaim its own pre-migration tasks. That contradicts the PR notes' stated upgrade path and can strand existing scheduled tasks after upgrade.

packages/daemon/src/daemon.ts
```diff
@@ -7347,8 +7353,11 @@ app.get("/api/tasks/:id/stream", (c) => {
 	});
```
PR-Reviewer-Ant, line 7353: This PR changes every `/api/tasks` read/write path to require agent-aware scoping, but the diff only threads `agentId` through the analytics panels. I do not see corresponding updates for existing task clients/callers. Any dashboard or CLI task flow that still omits `agent_id` will now default to the `default` agent on list/create and return 404 on detail/update/run/stream/runs for non-default agents. That is a behavior regression relative to the PR's claim of full task isolation.

packages/daemon/src/daemon.ts
```diff
@@ -7645,6 +7726,17 @@ app.post("/api/tasks/:id/run", async (c) => {
 // Get paginated run history for a task
 app.get("/api/tasks/:id/runs", (c) => {
 	const taskId = c.req.param("id");
+	const scoped = resolveScopedAgent(c.get("auth")?.claims ?? null, authConfig.mode, c.req.query("agent_id"));
```
PR-Reviewer-Ant, line 7731: Blocking: the new agent reassignment path changes ownership only on `scheduled_tasks`, but `/api/tasks/:id/runs` authorizes using the task's current `agent_id` and then returns all `task_runs` for that `task_id`. After a task is moved between agents via `PATCH /api/tasks/:id`, the new owner can read run history produced under the previous owner, including stdout/stderr/error payloads. This breaks the PR's claimed agent-scoped task isolation and introduces a cross-agent data leak. The runs need their own ownership/scoping field, or reassignment must migrate/history-split/reject when prior runs exist.

packages/core/src/migrations/054-task-agent-id.ts
```diff
@@ -0,0 +1,32 @@
...
+ * reassignment via body.agentId (validated through resolveScopedAgent).
+ * Callers that omit agent_id default to 'default' — no existing client
+ * breaks because all pre-migration callers were implicitly 'default'.
+ */
+
+import type { MigrationDb } from "./index";
+
+export function up(db: MigrationDb): void {
+	const cols = db.prepare("PRAGMA table_info(scheduled_tasks)").all() as ReadonlyArray<Record<string, unknown>>;
+	const colNames = new Set(cols.flatMap((c) => (typeof c.name === "string" ? [c.name] : [])));
```
PR-Reviewer-Ant, line 27: This migration introduces a breaking access-control change for existing multi-agent installs. Every legacy task is backfilled to `agent_id = 'default'`, and this PR simultaneously changes all task endpoints to require `agent_id` matches in SQL. After upgrade, any task that logically belonged to a non-default agent becomes invisible to that agent until an operator manually PATCHes it. That diverges from the PR's 'full per-agent attribution and agent-scoped task isolation' claim because upgraded data is not actually attributed correctly. This needs an automatic migration/backfill strategy or an explicit degraded-mode safeguard in the API, not just a manual note in the PR description.

aaf2tbz:
Done attempting this pr - needs deeper spec. Will attempt another day.

PR #414, opened 2026-03-30
fix(daemon): ignore .db files in watcher + retry port bind on EADDRINUSE

ddasgupta4:
## Summary

Two fixes for a daemon crash loop observed on macOS with launchd `KeepAlive: true`. The watcher ignored `.db-wal`/`.db-shm`/`.db-journal` but not `.db` itself, causing unnecessary watcher events on every DB write. The HTTP server bind had no retry logic, so `EADDRINUSE` from a process-manager restart race caused an infinite crash loop (38,000+ uncaught exceptions observed).

## Changes

- `packages/daemon/src/watcher-ignore.ts` — add `.endsWith(".db")` to ignore matcher
- `packages/daemon/src/watcher-ignore.test.ts` — regression test for `memories.db` ignore
- `packages/daemon/src/bind-with-retry.ts` — new module: indefinite retry with capped exponential backoff
- `packages/daemon/src/bind-with-retry.test.ts` — 5 tests: success, retry+recover, indefinite retry, delay cap, non-EADDRINUSE fatal
- `packages/daemon/src/daemon.ts` — replace inline `serve()` with extracted `bindWithRetry()`

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/daemon`

## PR Readiness (MANDATORY)

### Checklist notes

- **Spec alignment**: Bug fix for existing daemon capability (HTTP server binding), not a new feature. No spec changes needed.
- **Agent scoping**: No data queries added or changed.
- **Input/config validation**: `maxDelayMs` and `baseDelayMs` have defaults. Delay is capped via `Math.min()`.
- **Error handling**: EADDRINUSE retries indefinitely in-process (never exits). Non-EADDRINUSE errors forwarded via `onFatalError`. Failed servers explicitly closed before retry.
- **Security**: No admin/mutation endpoints added or changed.
- **Docs**: No API/spec/status changes.
- **Regression tests**: 5 tests for bind-with-retry + 1 for watcher-ignore.
- **Lint/typecheck/tests**: biome clean, 9/9 tests pass.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes

### Reviewer feedback addressed

- **PR-Reviewer-Ant (v1)**: Added regression tests for the EADDRINUSE retry path (was only manually verified).
- **PR-Reviewer-Ant (v2)**: Changed from bounded retries (5 attempts then crash) to **indefinite in-process retry** with capped exponential backoff (1s→2s→4s→8s→16s→30s→30s...). The process never exits for EADDRINUSE, so supervisor thrash is eliminated entirely — not just slowed down.

## AI disclosure

## Notes

**Root cause context**: Discovered on a macOS machine where the daemon was managed by both a launchd plist (`KeepAlive: true`) AND manual `signet daemon start` commands. The launchd plist continuously restarted the crashed daemon while watcher `.db` events added I/O pressure.

The `bindWithRetry` module uses injectable `schedule()` and `onFatalError()` callbacks for testability without real timers or uncaught exception interception.

### DIFF
--- a/packages/daemon/src/bind-with-retry.ts
+++ b/packages/daemon/src/bind-with-retry.ts
@@ -52,9 +52,13 @@ export function bindWithRetry(opts: BindOptions, attempt = 0): void {
 				// Server may not have fully initialized — safe to ignore
 			}
 			if (opts.signal?.aborted) return;
-			const timer = schedule(() => bindWithRetry(opts, attempt + 1), delay);
+			const onAbort = (): void => clearTimeout(timer);
+			const timer = schedule(() => {
+				opts.signal?.removeEventListener("abort", onAbort);
+				bindWithRetry(opts, attempt + 1);
+			}, delay);
 			// Cancel the pending retry if shutdown is requested
-			opts.signal?.addEventListener("abort", () => clearTimeout(timer), { once: true });
+			opts.signal?.addEventListener("abort", onAbort, { once: true });
 		} else {
 			handleFatal(err);
 		}


PR-Reviewer-Ant:
Hi @ddasgupta4 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): ignore .db files in watcher + retry port bind on EADDRINUSE` (commit `ac051f87`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -12433,64 +12433,101 @@ async function main() {
...
+		});
+
+		// Sync existing Claude Code project memories (also needs HTTP API)
+		const claudeProjectsDir = join(homedir(), ".claude", "projects");
+		if (existsSync(claudeProjectsDir)) {
+			syncExistingClaudeMemories(claudeProjectsDir);
+		}
+	};
+
+	// Bind the HTTP server with retry + exponential backoff so
```
PR-Reviewer-Ant: The port-bind crash-loop fix is the main incident addressed by this PR, but unlike the watcher fix it has no regression test. Per this repo's incident guardrail, the fix is not durable until the `EADDRINUSE` retry path is exercised automatically (for example by stubbing `listen()`/`error` and asserting bounded retries plus eventual failure or success). Without that, a future refactor can silently reintroduce the crash loop.

packages/daemon/src/bind-with-retry.ts
```diff
@@ -0,0 +1,60 @@
...
+			throw err;
+		});
+	const server = opts.createServer();
+
+	server.on("error", (err: NodeJS.ErrnoException) => {
+		if (err.code === "EADDRINUSE" && attempt < maxRetries) {
+			const delay = baseDelay * 2 ** attempt;
+			logger.warn("daemon", `Port ${opts.port} in use, retrying in ${delay}ms (attempt ${attempt + 1}/${maxRetries})`);
+			try {
+				server.close();
```
PR-Reviewer-Ant, line 50: This retry path still terminates the process after the last attempt (`handleFatal(err)` defaults to `throw err`). That diverges from the PR description’s claim that it 'prevents the crash loop entirely': under `launchd KeepAlive: true`, the daemon will still be restarted forever once the 5 retries are exhausted, just on a slower cadence. If the goal is to stop supervisor thrash, this needs an in-process long-lived retry strategy or another explicit non-restart path.

packages/daemon/src/watcher-ignore.ts
```diff
@@ -34,6 +34,7 @@ export function createAgentsWatcherIgnoreMatcher(agentsDir: string): (path: stri
 		return (
 			isGeneratedWorkspacePath ||
 			ignoredPaths.has(normalizedPath) ||
+			normalizedPath.endsWith(".db") ||
```
PR-Reviewer-Ant: This broadens the ignore rule from the daemon's own SQLite artifacts to every `*.db` file anywhere in the watched workspace. That means user-managed databases placed under `$SIGNET_WORKSPACE` will no longer trigger watcher-driven sync/commit behavior, which is a silent correctness regression and also diverges from the PR description's narrower goal (`memories.db` writes). Please scope this to the known daemon DB path(s) instead of a global suffix match.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `7229672e`

[Automated Review] Found one blocking correctness issue: the watcher ignore change is broader than the PR description and can suppress legitimate workspace file tracking.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The added `normalizedPath.endsWith(".db")` check in `packages/daemon/src/watcher-ignore.ts` is directly visible in the diff and is not constrained to `memory/memories.db`. Given this matcher is used for the workspace watcher, the new behavior will ignore any `.db` file anywhere under `$SIGNET_WORKSPACE`, which diverges from the PR description's stated fix for the daemon's own SQLite writes. I could not inspect the full watcher call sites in this environment to enumerate every downstream effect, but the overbroad ignore itself is explicit in the patch.

packages/daemon/src/daemon.ts
```diff
@@ -12433,64 +12434,75 @@ async function main() {
...
+					startedAt: new Date().toISOString(),
+					pid: process.pid,
+				}),
+			);
+			if (previousVersion && previousVersion !== CURRENT_VERSION && CURRENT_VERSION !== "0.0.0") {
+				logger.info("daemon", `Upgraded from ${previousVersion} to ${CURRENT_VERSION}`, {
+					previousVersion,
+					currentVersion: CURRENT_VERSION,
+				});
+				logger.info(
```
PR-Reviewer-Ant, line 12470: Warning: this changes daemon bind/retry behavior, but the repo's daemon section explicitly requires behavioral parity with `packages/daemon-rs/`. I do not see any Rust-side update or parity test in this PR. With `shadowEnabled: true`, the JS daemon will now retry indefinitely on `EADDRINUSE` while the Rust shadow likely preserves the old behavior, which creates expected divergence noise and weakens the shadow comparison contract.

ddasgupta4:
**Follow-up on automated review feedback (commit 473de6d6):**

**Cross-platform path separator bug — fixed.**
Replaced the `endsWith("/memories.db")` / `endsWith("/memories.db-wal")` etc. suffix checks with resolved absolute paths computed via `join(agentsDir, "memory", "memories.db[*]")` added to the `ignoredPaths` Set. This uses Node's platform-aware `path.join` + `resolve` throughout (via the existing `resolveForComparison` helper), so the ignore correctly matches on both POSIX and Windows. All 4 watcher-ignore tests pass.

**daemon-rs parity — not applicable for this change.**
The `notify` crate appears in the workspace `Cargo.toml` but `packages/daemon-rs` does not currently have a filesystem watcher implementation (no `use notify` or `Watcher` usage in any `.rs` file). The watcher and its ignore logic live exclusively in the TypeScript daemon (`packages/daemon/src/watcher-ignore.ts`). There is no Rust-side counterpart to update. If/when a Rust watcher is added, the same `memories.db` ignore will need to be ported at that time.

packages/daemon/src/daemon.ts
```diff
@@ -12433,64 +12434,75 @@ async function main() {
...
+		}
 
-			// Import existing memory markdown files (OpenClaw memory logs)
-			// Do this after server starts so the HTTP API is available for ingestion
-			importExistingMemoryFiles().catch((e) => {
-				const errDetails = e instanceof Error ? { message: e.message, stack: e.stack } : { error: String(e) };
-				logger.error("daemon", "Failed to import existing memory files", undefined, errDetails);
-			});
+		// Import existing memory markdown files (OpenClaw memory logs)
+		// Do this after server starts so the HTTP API is available for ingestion
```
PR-Reviewer-Ant, line 12480: This refactor drops `hostname` when constructing the Hono adapter server. `@hono/node-server`'s `serve()` passes `hostname` into `createAdaptorServer()`, and that value is used as the fallback host during request construction. After this change, requests that arrive without a `Host` header will start failing with `Missing host header` instead of being accepted on the configured bind host. If the goal is to preserve the old behavior while adding retry logic, pass `hostname: BIND_HOST` through to `createAdaptorServer()` as well.

packages/daemon/src/bind-with-retry.ts
```diff
@@ -0,0 +1,63 @@
...
+			throw err;
+		});
+	const server = opts.createServer();
+
+	server.on("error", (err: NodeJS.ErrnoException) => {
+		if (err.code === "EADDRINUSE") {
+			const delay = Math.min(baseDelay * 2 ** attempt, maxDelay);
+			logger.warn("daemon", `Port ${opts.port} in use, retrying in ${delay}ms (attempt ${attempt + 1})`);
+			try {
+				server.close();
```
PR-Reviewer-Ant, line 50: The new `EADDRINUSE` recovery loop is not cancelable: each failure schedules another future `bindWithRetry()` call, but neither `bindWithRetry()` nor its caller keeps a timer handle or a stopped flag. If the daemon is asked to shut down while it is backing off, the pending timer can keep the process alive and/or re-enter `listen()` after shutdown has begun. Given this repo's emphasis on timer cleanup and predictable shutdown behavior, the retry loop needs an abort mechanism tied into daemon shutdown before this is safe.

ddasgupta4:
**Response to automated review on commit 473de6d6 — hostname regression claim:**

The reviewer's finding does not match the actual diff. `hostname: BIND_HOST` **is** present in the `createAdaptorServer` call introduced by this PR (see the `+hostname: BIND_HOST` hunk in `daemon.ts`):

```typescript
createAdaptorServer({
    fetch: app.fetch,
    hostname: BIND_HOST,       // ← present
    createServer: createBoundedServer,
}),
```

The data flow is equivalent to the original `serve()` path:

- **Before:** `serve({ fetch, port, hostname, createServer })` → internally calls `createAdaptorServer(options)` (all fields forwarded) → `getRequestListener({ hostname })`  
- **After:** `bindWithRetry({ hostname, createServer: () => createAdaptorServer({ fetch, hostname, createServer }) })` → same `getRequestListener({ hostname })`

In both cases `hostname` reaches `getRequestListener` and is used for URL construction in `newRequest`. There is no regression in request parsing behavior.

No code change needed for this finding.

ddasgupta4:
**Response to automated review on commit 63aaec52 — non-cancelable retry loop:**

The reviewer notes \`bindWithRetry\` schedules retries via \`setTimeout\` with no cancellation handle, and worries a retry could fire after shutdown. This is not an issue in the daemon's actual shutdown flow.

The daemon's signal handlers (lines 11357–11362 of \`daemon.ts\`) always call \`process.exit()\` after cleanup:

\`\`\`typescript
process.on("SIGTERM", () => {
    cleanup().finally(() => process.exit(0));
});
process.on("SIGINT", () => {
    cleanup().finally(() => process.exit(0));
});
\`\`\`

\`process.exit()\` tears down the event loop, cancelling all pending \`setTimeout\` callbacks. A retry scheduled while a SIGTERM is being handled cannot fire — it's discarded with the rest of the timer queue when the process exits. The reviewer explicitly acknowledged they "could not inspect the daemon's shutdown/signal handling," which is why they flagged this conservatively.

A cancellation handle (e.g. \`AbortSignal\`) would be the right addition if \`bindWithRetry\` were a general-purpose library, but in this daemon context where all shutdown paths lead to \`process.exit()\`, it adds complexity with no safety benefit. No code change needed.

packages/daemon/src/bind-with-retry.ts
```diff
@@ -0,0 +1,70 @@
...
+			} catch {
+				// Server may not have fully initialized — safe to ignore
+			}
+			if (opts.signal?.aborted) return;
+			const timer = schedule(() => bindWithRetry(opts, attempt + 1), delay);
+			// Cancel the pending retry if shutdown is requested
```
PR-Reviewer-Ant, line 60: Each `EADDRINUSE` retry adds a new `abort` listener to the shared signal, but that listener is only removed if the signal eventually aborts. If the port stays busy for a while and then recovers normally, all prior listeners remain attached forever. In the persistent-retry scenario this change is meant to handle, that will eventually trigger `MaxListenersExceededWarning` and leak timer closures. This needs either explicit listener cleanup when the scheduled retry fires or a different cancellation strategy that does not register one listener per attempt.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `df73865e`

[Automated bot review: pr-reviewer] Found one issue in the new retry lifecycle; otherwise the implementation matches the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The watcher change now ignores only the daemon-owned memories database files and includes regression coverage for that path. The bind refactor does add in-process EADDRINUSE retries with abort support as described. The remaining issue is directly visible in packages/daemon/src/bind-with-retry.ts: every retry attempt registers a fresh abort listener on the shared AbortSignal, and those listeners are never removed when the timer fires normally. Under the exact persistent-EADDRINUSE scenario this PR targets, the process will accumulate listeners until Node emits MaxListenersExceededWarning and retain stale timer closures unnecessarily.

NicholaiVogel:
@ddasgupta4 Really nice job! Thanks for sticking through the process with @PR-Reviewer-Ant - this fix was long overdue, merged.

ddasgupta4:
Happy to help!! It was pretty fun to go back and forth with the PR reviewer

PR #416, opened 2026-03-31
feat(dashboard): show real overview usage analytics

NicholaiVogel:
## Summary

Make the home overview card reflect real MCP server and skill usage instead of catalog popularity. This adds skill invocation tracking parallel to the existing MCP invocation ledger, reuses the analytics pattern across both surfaces, and updates the overview UI to rank items from real usage data.

## Changes

- `docs/research/technical/RESEARCH-OVERVIEW-USAGE-ANALYTICS.md` — research note for the dashboard truthfulness problem
- `docs/specs/approved/overview-usage-analytics.md` — approved implementation contract for real overview usage analytics
- `docs/specs/approved/mcp-cli-bridge-and-usage-analytics.md` — moved approved MCP analytics spec out of planning/
- `docs/specs/INDEX.md`, `docs/specs/dependencies.yaml` — synced spec registry and dependency metadata
- `packages/core/src/migrations/053-skill-invocations.ts` — adds `skill_invocations`
- `packages/daemon/src/skill-invocations.ts` — shared skill invocation recorder that also updates `skill_meta.use_count` and `last_used_at`
- `packages/daemon/src/routes/skill-analytics.ts` — `/api/skills/analytics` summary route with scoped reads and ISO validation
- `packages/daemon/src/scheduler/worker.ts`, `packages/daemon/src/daemon.ts` — record skill usage from scheduled and manual task runs
- `packages/cli/dashboard/src/lib/api.ts` — skill analytics client + MCP analytics limit support
- `packages/cli/dashboard/src/lib/components/home/MarketplaceSpotlights.svelte` — home card now uses real usage data with metadata fallback
- `docs/API.md`, `docs/PROCEDURAL-MEMORY.md` — docs updated for new analytics surface and procedural-memory P2 status
- `packages/daemon-rs/crates/signet-core/src/sql/038-skill-invocations.sql` and `migrations.rs` — Rust parity for schema change

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `daemon-rs`

## Screenshots

UI change is limited to the home overview card. I do not have an attached screenshot from this CLI flow, so this section is currently pending visual capture before merge.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Migration 053 adds `skill_invocations` as a new analytics ledger table. It is idempotent via `CREATE TABLE/INDEX IF NOT EXISTS`.

Rust daemon parity was updated by adding `038-skill-invocations.sql` and bumping the Rust schema version.

Rollback note: `DROP TABLE skill_invocations; DELETE FROM schema_migrations WHERE version = 53;` in the JS schema, and remove Rust migration 038 if parity rollback is required before cutover.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Targeted validation run for this PR:

- `bun test packages/core/src/migrations/migrations.test.ts packages/daemon/src/routes/skill-analytics.test.ts packages/daemon/src/scheduler/worker-execute.test.ts`
- `bunx biome check packages/core/src/migrations/053-skill-invocations.ts packages/daemon/src/skill-invocations.ts packages/daemon/src/routes/skill-analytics.ts packages/daemon/src/routes/skill-analytics.test.ts packages/daemon/src/scheduler/worker.ts packages/daemon/src/scheduler/worker-execute.test.ts packages/cli/dashboard/src/lib/api.ts`

`bun run typecheck` at the workspace root still fails because of pre-existing unrelated repo issues in other packages, including missing `@types/node` in `@signet/connector-oh-my-pi` and existing daemon/package typecheck drift.

## AI disclosure

## Notes

- This intentionally does not revive the task-ownership scope expansion from PR #413.
- The overview card now ranks from real analytics data, while presentation metadata still falls back to installed/catalog sources when available.
- `scripts/spec-deps-check.ts` still reports pre-existing unrelated spec drift for `recall-confidence-gate` and `references/Engram`.


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(dashboard): show real overview usage analytics` (commit `d5a855a3`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -7638,6 +7642,19 @@ app.post("/api/tasks/:id/run", async (c) => {
 					error: result.error,
 					timestamp: new Date().toISOString(),
 				});
+
+				if (taskSkillName) {
+					void import("./skill-invocations.js").then((skills) => {
+						skills.recordSkillInvocation({
+							skillName: taskSkillName,
```
PR-Reviewer-Ant, line 7673: Blocking: this records manual task-run skill usage under `agentId: "default"` unconditionally. In a multi-agent workspace, usage from any non-default agent will be written to the wrong tenant and will also increment the wrong `skill_meta.use_count` / `last_used_at`. The PR description says agent scoping was verified, but this path violates the repo invariant against hardcoding `default` for scoped data.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -316,4 +320,15 @@ export async function executeTask(
 		exitCode: result.exitCode,
 		timedOut: result.timedOut,
 	});
+
+	if (task.skill_name) {
+		deps.recordSkillInvocation({
```
PR-Reviewer-Ant, line 327: Blocking: scheduled task skill usage is also attributed to `agentId: "default"` unconditionally. That creates cross-agent analytics corruption and stale counters for the real owning agent. If scheduled tasks are agent-scoped now or become agent-scoped later, this ledger will already be wrong.

packages/daemon/src/routes/skill-analytics.ts
```diff
@@ -0,0 +1,125 @@
...
+	const latencies = db
+		.prepare(`SELECT si.latency_ms FROM skill_invocations si WHERE ${where} ORDER BY si.latency_ms`)
+		.all(...params) as readonly { latency_ms: number }[];
+
+	const sorted = latencies.map((row) => row.latency_ms);
+
+	return {
+		totalCalls,
+		successRate: Math.round(successRate * 1000) / 1000,
+		topSkills,
```
PR-Reviewer-Ant, line 91: Warning: the new `/api/skills/analytics` route resolves scope but does not perform any explicit permission check before returning usage analytics. If the surrounding route stack does not already enforce an analytics/operator permission here, this expands read access to operational usage data in `team`/`hybrid` mode.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `d5a855a3`

[Automated Review] Found agent-scoping corruption in the new skill usage ledger and a likely auth gap on the new analytics endpoint.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The new write paths in `packages/daemon/src/daemon.ts` and `packages/daemon/src/scheduler/worker.ts` explicitly hardcode `agentId: "default"` when recording usage, which directly conflicts with the repo's multi-agent scoping invariant and will misattribute rows and `skill_meta` counters. I can prove that from this diff alone. I also found that `packages/daemon/src/routes/skill-analytics.ts` adds a new analytics endpoint with scope resolution but no explicit permission check; I could not inspect the surrounding auth middleware or the existing MCP analytics route in this environment, so that finding has slightly lower confidence.

packages/daemon/src/skill-invocations.ts
```diff
@@ -0,0 +1,87 @@
...
+		if (rows.length !== 1) return undefined;
+		return rows[0]?.agentId;
+	});
+}
+
+export function recordSkillInvocation(record: SkillInvocationRecord): void {
+	const skill = normalizeSkillName(record.skillName);
+	if (skill.length === 0) return;
+
+	const now = new Date().toISOString();
```
PR-Reviewer-Ant, line 29: Blocking: this silently skips recording whenever the skill owner cannot be resolved to exactly one installed `skill_meta` row. That means real usage disappears for uninstalled skills (which the spec explicitly says should still render with fallback metadata) and for multi-agent installs of the same skill name, because the scheduler path does not pass an `agentId`. In those cases the home card will undercount or omit real usage rather than showing historical rows.

packages/daemon/src/daemon.ts
```diff
@@ -7638,6 +7656,19 @@ app.post("/api/tasks/:id/run", async (c) => {
...
```
PR-Reviewer-Ant, line 7673: Blocking: manual task runs are attributing skill usage to `scoped.agentId` from the request, not to the task's own owner/scope. If an operator/admin triggers another agent's task, the invocation row and `skill_meta.use_count` update will be written under the caller scope instead of the task scope, which corrupts per-agent analytics. The code resolves request scope before loading the task and never derives analytics scope from the task record itself.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `f0842c85`

[Automated Review] Found two correctness issues in the new skill-usage tracking path that can make the overview analytics incomplete or misleading.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The primary issues are directly visible in the diff: `recordSkillInvocation()` drops writes unless exactly one installed owner can be resolved, and the manual task-run path records analytics using the caller scope captured before the task is loaded. I could not inspect the full scheduled-task schema or prior review thread state from the repo checkout, so confidence is not marked high.

packages/core/src/migrations/054-task-agent-scope.ts
```diff
@@ -0,0 +1,43 @@
+import type { MigrationDb } from "./index";
+
+export function up(db: MigrationDb): void {
```
PR-Reviewer-Ant, line 3: This PR claims it does not revive task-ownership scope expansion, but it does exactly that: legacy `scheduled_tasks` rows now gain an `agent_id`, and all task routes later enforce `WHERE ... agent_id = ?`. The backfill here only reassigns tasks when `skill_name` maps to exactly one installed agent; everything else silently stays `default`. On upgrade, any existing non-default user's tasks without a unique `skill_name` mapping become invisible/unreachable to that user. That is a breaking behavior change outside the stated scope of the PR and needs either a safe migration strategy or to be split out.

packages/daemon-rs/crates/signet-core/src/migrations.rs
```diff
@@ -231,9 +231,19 @@ static MIGRATIONS: &[Migration] = &[
         name: "memory-md-rolling-window-lineage",
         sql: include_str!("sql/037-memory-md-rolling-window-lineage.sql"),
     },
+    Migration {
```
PR-Reviewer-Ant, line 234: This updates Rust schema parity only, but the JS daemon also adds new runtime behavior: skill-invocation writes from scheduled/manual task runs and a new `/api/skills/analytics` read surface. The repo contract says daemon behavioral changes must also be reflected in `packages/daemon-rs/`. As written, shadow mode will diverge once these code paths are exercised, because the Rust side knows about the table but not the new behavior.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `3cef7a2f`

[Automated Review] Found two blocking issues: this PR introduces an unadvertised scheduled-task ownership/scoping change that can hide or orphan existing tasks, and it updates only Rust schema parity without implementing the matching Rust runtime behavior for the new analytics surface.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The scheduled-task scope change is directly visible in the migration and route diffs: `scheduled_tasks` now gains `agent_id`, every task endpoint filters on it, and the migration backfills only a narrow subset of legacy rows based on unique skill ownership. That is enough to show a breaking behavior change for pre-existing tasks. The Rust parity issue is also directly visible: only SQL/schema files were added under `packages/daemon-rs`, while the JS daemon gained new write paths and a new `/api/skills/analytics` route. I did not run the shadow daemon to capture a divergence trace, so parity impact is inferred from the changed modules rather than a runtime repro.

packages/daemon-rs/crates/signet-daemon/src/routes/scheduler.rs
```diff
@@ -152,6 +152,12 @@ pub async fn create(
                             body.skill_mode,
                         ],
                     )?;
+                    conn.execute(
```
PR-Reviewer-Ant, line 229: Blocking: new tasks always write `task_scope_hints.agent_id = 'default'` in the Rust daemon path. That means scheduled-task skill analytics will be misattributed for every non-default agent, which contradicts the PR's scoped analytics goal and breaks JS/Rust parity.

packages/daemon-rs/crates/signet-daemon/src/routes/skill_analytics.rs
```diff
@@ -0,0 +1,147 @@
...
+
+fn pct(sorted: &[i64], p: usize) -> i64 {
+    if sorted.is_empty() {
+        return 0;
+    }
+    let idx = ((p * sorted.len()).div_ceil(100)).saturating_sub(1);
+    *sorted.get(idx).unwrap_or(&0)
+}
+
+/// GET /api/skills/analytics
```
PR-Reviewer-Ant, line 37: Blocking: this endpoint trusts the caller-supplied `agent_id` directly instead of resolving it against request scope like the JS implementation does. On the Rust daemon path, a scoped caller can query another agent's analytics by passing a different `agent_id`, which is an agent-scoping leak.

packages/daemon-rs/crates/signet-daemon/src/routes/skill_analytics.rs
```diff
@@ -0,0 +1,147 @@
...
+pub async fn summary(
```
PR-Reviewer-Ant, line 38: Warning: `since` is not validated as ISO 8601 UTC here, even though the new API docs and JS route both enforce `400` for invalid timestamps. The Rust path will silently accept malformed values and return inconsistent behavior for the same API contract.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `8899dedf`

[Automated Review] Found two blocking parity/scoping issues in the Rust daemon path and one API contract regression.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The introduced Rust code in `routes/scheduler.rs` and `routes/skill_analytics.rs` directly shows incorrect agent attribution and missing request scoping/validation logic, while the PR description explicitly claims scoped analytics reads and Rust parity. I did not run the daemon, but these issues are visible from the added SQL and handler logic.

packages/daemon-rs/crates/signet-daemon/src/routes/skill_analytics.rs
```diff
@@ -0,0 +1,147 @@
...
+    State(state): State<Arc<AppState>>,
+    Query(params): Query<SkillAnalyticsQuery>,
+) -> impl IntoResponse {
+    let agent_id = params.agent_id.unwrap_or_else(|| "default".to_string());
+    let since = params.since;
```
PR-Reviewer-Ant, line 62: This Rust handler does not enforce the contract the PR description/docs claim for `/api/skills/analytics`. The JS route uses auth-aware scoped agent resolution and rejects malformed `since` with `400`, but this version accepts any `agent_id` query override and any `since` string. In team/hybrid mode that becomes an analytics data-leak / parity bug, because callers can query another agent's usage rows and malformed timestamps no longer fail fast.

packages/daemon-rs/crates/signet-daemon/src/routes/scheduler.rs
PR-Reviewer-Ant, line 229: New tasks created through the Rust daemon always write `task_scope_hints.agent_id = 'default'`. Later scheduled/manual usage attribution reads from this hint table, so skill invocation analytics and `skill_meta.use_count` updates will be recorded against the default agent even when the task belongs to some other agent. That breaks the repo's agent-scoping invariant and diverges from the JS implementation, which stores the resolved scoped agent id.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `af3cd7a3`

[Automated Review] Found two blocking parity/scoping issues in the new Rust analytics/task-scope paths.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The Rust route added in this diff reads raw `agent_id`/`since` query params directly and never applies the JS route's auth-scoped agent resolution or ISO-8601 validation. Separately, the Rust scheduler create path inserted here hardcodes `task_scope_hints.agent_id` to `default`, which deterministically misattributes later analytics writes for non-default agents.

packages/daemon/src/daemon.ts
```diff
@@ -7537,6 +7560,8 @@ app.delete("/api/tasks/:id", (c) => {
 // Trigger an immediate manual run
```
PR-Reviewer-Ant, line 7561: Blocking: the new scoped-agent handling here does not actually scope the task lookup or authorization. The route still reads the task by bare `id`, then later pulls `task_scope_hints.agent_id` for attribution. In team/hybrid mode, a caller with permission and another task's ID can trigger that task across agent boundaries and write analytics under the other agent's scope. This is a new cross-agent mutation/scoping leak introduced by the analytics attribution path.

packages/daemon-rs/crates/signet-daemon/src/main.rs
```diff
@@ -534,6 +534,7 @@ async fn main() -> anyhow::Result<()> {
             axum::routing::post(routes::scheduler::trigger),
         )
         .route("/api/tasks/{id}/runs", get(routes::scheduler::runs))
+        .route("/api/skills/analytics", get(routes::skill_analytics::summary))
```
PR-Reviewer-Ant, line 537: Blocking: this exposes `/api/skills/analytics` in the Rust daemon, but the PR never adds a Rust equivalent of `recordSkillInvocation` or any scheduler/manual-run write path to populate `skill_invocations` and update `skill_meta.use_count` / `last_used_at`. The JS daemon will produce data while the Rust daemon will report empty analytics forever, which diverges from both the PR description and the repo's daemon-rs parity rule.

packages/daemon/src/scheduler/worker.ts
```diff
@@ -316,4 +322,15 @@ export async function executeTask(
 		exitCode: result.exitCode,
```
PR-Reviewer-Ant, line 322: Warning: the new skill analytics ledger is only written from scheduled/manual task execution paths in this PR. I do not see any instrumentation for interactive skill invocations, so the dashboard's 'real usage' ranking is still limited to task-driven usage rather than overall skill usage. That does not fully match the PR description's broader claim.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `18e7702d`

[Automated Review] Found two blocking issues and one correctness gap in the new overview usage analytics implementation.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The cross-agent trigger path is directly visible in the JS daemon diff: the new `agent_id` handling is added, but `/api/tasks/:id/run` still fetches tasks by bare `id` and then attributes usage from `task_scope_hints`. The Rust parity gap is also explicit in the diff: new schema and read routes were added, but no Rust write path analogous to `recordSkillInvocation` was introduced. I did not run the daemon, so runtime behavior is inferred from the changed code paths rather than reproduced end-to-end.

packages/daemon-rs/crates/signet-daemon/src/routes/scheduler.rs
```diff
@@ -354,47 +435,89 @@ pub async fn delete(
...
+                    conn.query_row(sql, [&id], |r| {
+                        Ok((r.get::<_, Option<String>>(0)?, r.get::<_, String>(1)?))
+                    })
+                    .ok()
+                };
+
+                let Some((skill_name, task_agent)) = task else {
                     return Ok(serde_json::json!({"error": "not_found"}));
-                }
+                };
```
PR-Reviewer-Ant, line 490: This records a skill invocation with `success = true` and `latency_ms = 0` at manual-trigger time, before the task has actually run. That makes the new overview analytics lie for the Rust daemon, and it also diverges from the JS path, which records after execution completes. The invocation should be emitted from the actual execution path with the real outcome, not from the enqueue/trigger route.

packages/daemon/src/daemon.ts
```diff
@@ -7537,10 +7561,12 @@ app.delete("/api/tasks/:id", (c) => {
...
 app.post("/api/tasks/:id/run", async (c) => {
 	const taskId = c.req.param("id");
+	const scoped = resolveScopedAgentId(c, c.req.query("agent_id"));
+	if (scoped.error) return c.json({ error: scoped.error }, 403);
 
 	const task = getDbAccessor().withReadDb((db) =>
-		db.prepare("SELECT * FROM scheduled_tasks WHERE id = ?").get(taskId),
-	) as Record<string, unknown> | undefined;
```
PR-Reviewer-Ant, line 7543: The new trigger path now trusts `task_scope_hints` via `readScopedTask(...)`/`readTaskAgentId(...)`, but the task update route still lets `skillName` change without updating `task_scope_hints`. After editing a task to point at a different skill/agent, manual runs and analytics attribution can use the stale owner, causing incorrect `agent_id` usage data and, in scoped modes, false 404/forbidden behavior for the actual owner. The hint table needs to be kept in sync on task updates, not just creation/migration backfill.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.4 | commit: `067ce8fa`

[Automated Review] Found two blocking correctness issues in the new usage-attribution path.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The Rust manual-trigger path now writes a successful skill invocation before any task execution happens, which is directly visible in the diff. The JS daemon now relies on task_scope_hints for scoped reads and analytics attribution, while the existing task update path still allows skill changes without any matching hint maintenance; that stale-state bug is provable by reading the updated trigger/select logic together with the unchanged task update SQL.

packages/daemon/src/daemon.ts
NicholaiVogel: I’m going to draw a hard boundary here: I’m not taking further task-system design feedback in this PR.

This PR is about overview usage analytics, not about reopening task ownership, task hint lifecycle, or broader scheduler/task architecture. At this point this line of review is asking for adjacent task-system work, not reviewing the feature that’s actually in scope here.

So to be explicit:
- in scope: persisted overview usage analytics for MCP servers and task-driven skill usage
- out of scope: task ownership redesign, task hint lifecycle expansion, and broader scheduler/task-system architecture changes

If there is a concrete issue here that directly causes the overview analytics feature itself to return incorrect user-visible results within the intended scope of this PR, call out that specific regression. Otherwise, I’m not expanding #416 further in that direction.

PR #420, opened 2026-03-31
feat(forge): phase A–D harness work, LLM optimization phases, and UI/UX polish

aaf2tbz:
## Summary

This PR packages the full Forge A–D implementation sequence plus hardening/polish across reliability, UX, and Signet integration.

### Highlights

- Forge harness phases A–D completed (workspace/policy/service/guards)
- LLM optimization roadmap phases 1–3 completed with hardening
- TUI refresh (header/layout/theming/transparency), plus input/streaming quality-of-life fixes
- Voice backend migration from Whisper to a TranscriptionSuite-compatible path
- MCP + skills compatibility hardening for in-harness execution
- Session-end extraction reliability improvements

## Latest hardening update (Mar 31, 2026)

This PR now includes an additional full Forge hardening pass focused on Claude/Codex runtime behavior, stream stability, and safety:

### Claude/Codex response + stream hardening
- Added persistent CLI session resume handling:
  - Codex: captures `thread_id` and uses `codex exec resume <thread_id>`.
  - Claude: captures stream `session_id` and uses `--resume <session_id>`.
- Hardened event parsing and completion behavior:
  - handles trailing non-newline JSON at EOF,
  - maps CLI `is_error` and failed turn payloads into explicit Forge errors,
  - clamps large CLI tool outputs before rendering.
- Added provider-edge handling to prevent stuck turns after final assistant text.
- Ensured tool/process events from Codex command execution surface cleanly in Forge chat.

### Latency + UX hardening
- Reduced memory-recall wait on hot path with bounded timeout (`FORGE_RECALL_TIMEOUT_MS`, default 1000ms).
- Removed delayed drip buffering so streamed text renders immediately per frame.
- Reduced noisy status behavior:
  - hides repetitive “Turn plan” by default (opt-in via `FORGE_SHOW_TURN_PLAN=1`),
  - dedupes/suppresses duplicate transient statuses,
  - removes redundant animated “Responding ...” line.
- Status indicators for “Thinking/Recalling” now render as Signet-green diamonds with cleaner transient handling.

### `/forge-bypass` safety hardening
- Reworked `/forge-bypass` into explicit, confirmation-gated modes:
  - `status`, `off`, `once --yes`, `on --persist --yes`.
- Restricts bypass to supported CLI providers (claude-cli/codex-cli).
- Adds one-shot auto-disable behavior after turn completion/error.
- Prevents bypass state from persisting across restarts.

### 1) Harness + runtime reliability
- Added/finished Forge phase A–D implementation work.
- Hardened command policy/workspace boundary behavior and command routing.
- Added local MCP stdio registry loading/merging so slash-command discoverability and MCP availability stay resilient when daemon marketplace data is unavailable.

### 2) LLM performance and behavior
- Applied phase 1/2/3 optimization plan and hardening:
  - budgeting/planner controls,
  - adaptive memory/tool scheduling,
  - safer bounded compression/refresh behaviors,
  - timeout/fallback handling and telemetry guardrails.

### 3) Session extraction + transcript quality
- `session_end` hook now always submits unless transcript is truly empty.
- Removed short-transcript silent skip behavior.
- Transcript shaping is cleaner for extraction reliability (system/noise reduction + bounded length with truncation marker).
- Added explicit session-end submission logging for easier verification/troubleshooting.

### 4) Voice backend migration
- Removed Whisper path from Forge harness flow.
- Integrated a TranscriptionSuite-compatible backend path for cross-platform live dictation support.

### 5) TUI/UX polish
- Header redesigned into clearer grouped sections with stronger visual hierarchy.
- Theme updates:
  - signet-dark yellow accents -> green
  - signet-light yellow accents -> blue
  - transparency mode refinements and defaulting behavior
  - themed borders/contrast cleanup
- Launch warning prompt upgraded to left/right selector + Enter confirm.
- Agent label/name behavior fixed to use configured Signet agent name when available (fallback to Forge when disconnected).
- Input composer image attachment UX:
  - drag/drop and clipboard image attach support normalized,
  - inline tokens rendered as `[Image #N]`,
  - token-aware cursor behavior,
  - backspace-at-input-start removes attached images one-by-one,
  - send supports attachment-only prompts and clears attached images after submit.

### Checks run
- `cargo check --workspace` (packages/forge)
- `cargo test --workspace --lib` (packages/forge)
- smoke non-interactive runs:
  - codex-cli
  - claude-cli

### Local install verification
- Rebuilt/reinstalled local forge binary after changes:
  - `/Users/averyfelts/.config/signet/bin/forge`

## PR Readiness Checklist (Mandatory)

## Notes

This PR is intentionally broad because it captures the phased implementation + subsequent hardening passes and UX refinements requested during iterative testing.

Screenshot 2026-03-31 at 1 39 38 PM

### DIFF
--- a/packages/cli/src/features/forge.ts
+++ b/packages/cli/src/features/forge.ts
@@ -827,8 +827,8 @@ export async function restartForgeService(options: ForgeServiceOptions, deps: Fo
 
 export async function showForgeServiceStatus(options: ForgeServiceOptions, deps: ForgeDeps): Promise<void> {
 	const basePath = readForgeServicePath(options, deps);
-	const daemonRunning = await deps.isDaemonRunning();
 	const daemon = await deps.getDaemonStatus();
+	const daemonRunning = daemon.pid != null;
 	const payload = {
 		running: daemonRunning,
 		pid: daemon.pid,
--- a/packages/forge/crates/forge-agent/src/agent_loop.rs
+++ b/packages/forge/crates/forge-agent/src/agent_loop.rs
@@ -198,7 +198,7 @@ impl AgentLoop {
             let ollama = forge_provider::create_provider("ollama", "qwen3:4b", "ollama");
             match ollama {
                 Ok(p) => {
-                    info!("Ollama available — using qwen3:4b for context compaction");
+                    info!("Ollama endpoint configured for compaction (availability unverified)");
                     Some(Arc::from(p))
                 }
                 Err(_) => None,
@@ -519,11 +519,10 @@ impl AgentLoop {
             // Pre-sampling compaction: check token estimate BEFORE calling
             // the LLM. If we're approaching the context window limit, compact
             // first to avoid wasting an LLM call on an oversized prompt.
+            // Uses the already-adjusted estimate so the compaction gate is
+            // consistent with the turn plan's view of token pressure.
             {
-                let estimated = {
-                    let s = session.lock().await;
-                    ContextManager::estimate_tokens(&s.messages)
-                };
+                let estimated = estimated_messages_tokens;
                 if self.context_manager.should_compact_before_sampling(estimated) {
                     info!(
                         "Pre-sampling compaction: {} estimated tokens (~{}% of {})",
@@ -1388,8 +1387,17 @@ fn compute_available_injection_tokens(
     max_tokens: usize,
     estimated_message_tokens: usize,
 ) -> usize {
-    let soft_cap = max_tokens.saturating_mul(18) / 100;
-    let remaining = max_tokens.saturating_mul(78) / 100;
+    // When context window is unknown (0), use a sensible default so memory
+    // injection is not silently disabled — matches prior unconditional behavior.
+    const DEFAULT_CONTEXT_WINDOW: usize = 8_000;
+    let effective = if max_tokens == 0 {
+        warn!("Provider context window is 0 (unknown); defaulting to {DEFAULT_CONTEXT_WINDOW} for memory budget");
+        DEFAULT_CONTEXT_WINDOW
+    } else {
+        max_tokens
+    };
+    let soft_cap = effective.saturating_mul(18) / 100;
+    let remaining = effective.saturating_mul(78) / 100;
     let budget_after_messages = remaining.saturating_sub(estimated_message_tokens);
     budget_after_messages.min(soft_cap)
 }
--- a/packages/forge/crates/forge-agent/src/retry.rs
+++ b/packages/forge/crates/forge-agent/src/retry.rs
@@ -117,3 +117,88 @@ impl RetryState {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn transient_err() -> ForgeError {
+        ForgeError::Provider("HTTP 500 internal server error".to_string())
+    }
+
+    fn rate_limit_err() -> ForgeError {
+        ForgeError::Provider("429 rate limit exceeded".to_string())
+    }
+
+    fn overflow_err() -> ForgeError {
+        ForgeError::ContextOverflow { used: 200_000, limit: 128_000 }
+    }
+
+    fn client_err() -> ForgeError {
+        ForgeError::ApiKeyMissing("anthropic".to_string())
+    }
+
+    fn fatal_err() -> ForgeError {
+        ForgeError::Session("corrupted state".to_string())
+    }
+
+    #[test]
+    fn transient_retries_up_to_limit_then_fails() {
+        let mut state = RetryState::with_policy(RetryPolicy {
+            transient_retries: 2,
+            ..Default::default()
+        });
+        assert!(matches!(state.decide(&transient_err()), RetryDecision::Retry(_)));
+        assert!(matches!(state.decide(&transient_err()), RetryDecision::Retry(_)));
+        assert!(matches!(state.decide(&transient_err()), RetryDecision::Fail));
+    }
+
+    #[test]
+    fn exponential_backoff_doubles() {
+        let mut state = RetryState::with_policy(RetryPolicy {
+            transient_retries: 3,
+            transient_base: Duration::from_secs(1),
+            ..Default::default()
+        });
+        match state.decide(&transient_err()) {
+            RetryDecision::Retry(d) => assert_eq!(d, Duration::from_secs(1)),
+            other => panic!("expected Retry, got {:?}", std::mem::discriminant(&other)),
+        }
+        match state.decide(&transient_err()) {
+            RetryDecision::Retry(d) => assert_eq!(d, Duration::from_secs(2)),
+            other => panic!("expected Retry, got {:?}", std::mem::discriminant(&other)),
+        }
+        match state.decide(&transient_err()) {
+            RetryDecision::Retry(d) => assert_eq!(d, Duration::from_secs(4)),
+            other => panic!("expected Retry, got {:?}", std::mem::discriminant(&other)),
+        }
+    }
+
+    #[test]
+    fn overflow_compacts_once_then_fails() {
+        let mut state = RetryState::with_policy(RetryPolicy {
+            overflow_retries: 1,
+            ..Default::default()
+        });
+        assert!(matches!(state.decide(&overflow_err()), RetryDecision::CompactAndRetry));
+        assert!(matches!(state.decide(&overflow_err()), RetryDecision::Fail));
+    }
+
+    #[test]
+    fn client_and_fatal_never_retry() {
+        let mut state = RetryState::new();
+        assert!(matches!(state.decide(&client_err()), RetryDecision::Fail));
+        assert!(matches!(state.decide(&fatal_err()), RetryDecision::Fail));
+    }
+
+    #[test]
+    fn rate_limit_retries_up_to_limit_then_fails() {
+        let mut state = RetryState::with_policy(RetryPolicy {
+            rate_limit_retries: 2,
+            ..Default::default()
+        });
+        assert!(matches!(state.decide(&rate_limit_err()), RetryDecision::Retry(_)));
+        assert!(matches!(state.decide(&rate_limit_err()), RetryDecision::Retry(_)));
+        assert!(matches!(state.decide(&rate_limit_err()), RetryDecision::Fail));
+    }
+}
--- a/packages/forge/crates/forge-cli/src/main.rs
+++ b/packages/forge/crates/forge-cli/src/main.rs
@@ -825,7 +825,7 @@ fn select_provider(cli: &Cli, available: &[DiscoveredProvider]) -> Result<(Strin
     if let Some(model) = cli.model.clone() {
         let inferred = infer_provider_from_model(&model).to_string();
         let has_inferred = available.iter().any(|p| p.provider == inferred);
-        if has_inferred || inferred == "ollama" {
+        if has_inferred {
             return Ok((inferred, model));
         }
     }


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(forge): phase A–D harness work, LLM optimization phases, and UI/UX polish` (commit `20627418`) now and I'll follow up shortly with feedback.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -346,10 +722,45 @@ impl AgentLoop {
...
+                tc.name.hash(&mut hasher);
+                serde_json::to_string(&tc.input).unwrap_or_default().hash(&mut hasher);
+                let hash = hasher.finish();
+                if seen_hashes.insert(hash) {
+                    deduped_calls.push(tc);
+                } else {
+                    // Map duplicate tool call ID to the original's ID for result reuse
+                    if let Some(original_id) = dedup_map.get(&hash) {
+                        debug!("Deduplicating tool call: {} (duplicate of {})", tc.id, original_id);
+                    }
```
PR-Reviewer-Ant: **Blocking: duplicate tool calls receive no ToolResult — breaks API contract.**

When the LLM emits N tool calls where some have identical `(name, input)`, this code drops the duplicates from `deduped_calls` and logs a debug message, but never adds a `MessageContent::ToolResult` for their `tool_use_id`s to `tool_results_content`.

Anthropic (and every other provider) requires a strict 1-to-1 mapping: every `tool_use` block in the assistant message must have a corresponding `tool_result` in the following user message, matched by `tool_use_id`. Sending back fewer results than tool calls returns a 400 from the API and breaks the conversation entirely.

The `dedup_map` scaffolding is already built (original_id per hash), but the result-reuse step is missing. After executing `deduped_calls`, iterate over the *original* `tool_calls` list and for any call whose hash maps to an already-executed ID, copy that result under the duplicate's `tool_use_id`:

```rust
for tc in &tool_calls {
    if completed_ids.contains(&tc.id) { continue; }
    let hash = compute_hash(tc);
    if let Some(original_id) = dedup_map.get(&hash) {
        if let Some(original_result) = results_by_id.get(original_id) {
            tool_results_content.push(MessageContent::ToolResult {
                tool_use_id: tc.id.clone(),
                content: original_result.content.clone(),
                is_error: original_result.is_error,
            });
        }
    }
}
```

packages/forge/crates/forge-provider/src/anthropic.rs
```diff
@@ -34,34 +34,88 @@ impl AnthropicProvider {
...
+
+        non_system
+            .iter()
+            .enumerate()
+            .map(|(idx, m)| {
+                let should_cache = cache_indices.contains(&idx);
+
                 let content: Vec<Value> = m
                     .content
                     .iter()
```
PR-Reviewer-Ant, line 71: **Blocking: cache breakpoints at `len-4`/`len-2` from the end shift every turn — guaranteed cache misses.**

Anthropic's prompt cache is a *prefix cache*: a hit occurs only when the bytes from position 0 up to and including the breakpointed block are byte-for-byte identical to a previously cached call. For a hit, the breakpoint must sit at a *stable, fixed* position.

With `user_indices[len-4]` and `user_indices[len-2]`, the breakpoints slide one position deeper into the message array each turn. The prefix before the breakpoint is always longer than on the previous turn, so no cached entry ever matches — every call is a cache *write* with zero reads. The comment's "~90% cost reduction on cache hits" will not materialise; users will instead pay the cache-write surcharge on every request for no benefit.

The correct pattern is fixed breakpoints at stable content: the end of the system prompt, or the end of a large injected document that doesn't change turn-to-turn. One approach: find the last message index whose position has not changed since the previous call (i.e., fix a boundary at a "frozen" prefix and only advance it when compaction moves that boundary). At minimum, a breakpoint on the first user message is stable for the lifetime of the session.

packages/cli/src/features/forge.ts
```diff
@@ -751,3 +771,80 @@ export async function doctorForge(options: ForgeStatusOptions, deps: ForgeDeps):
...
 }
+
+function readForgeServicePath(options: ForgeServiceOptions, deps: ForgeDeps): string {
+	const raw = deps.extractPathOption(options.path);
+	return raw ? deps.normalizeAgentPath(raw) : deps.agentsDir;
+}
+
+export async function startForgeService(options: ForgeServiceOptions, deps: ForgeDeps): Promise<void> {
+	const basePath = readForgeServicePath(options, deps);
+	const running = await deps.isDaemonRunning();
```
PR-Reviewer-Ant, line 786: **Warning: `isDaemonRunning()` is path-agnostic; `startDaemon(basePath)` is path-specific.**

In `startForgeService`, `stopForgeService`, and `restartForgeService`, the running-state check calls `deps.isDaemonRunning()` with no path argument, while the actual start/stop operations pass `basePath`. In a multi-workspace setup (or if the default AGENTS_DIR differs from the path option), `isDaemonRunning()` may detect a daemon running for a *different* workspace and report "already running" when the target workspace daemon is not running — or vice versa. The check and the operation should be scoped to the same workspace.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -504,6 +1110,425 @@ impl AgentLoop {
...
+    let max_chars = env_max_chars
+        .map(|cap| adaptive_max_chars.min(cap))
+        .unwrap_or(adaptive_max_chars);
+    const DEFAULT_MAX_LINES: usize = 400;
+    const MIN_MAX_LINES: usize = 40;
+    const ABSOLUTE_MAX_LINES: usize = 4_000;
+    let max_lines = std::env::var("FORGE_TOOL_RESULT_MAX_LINES")
+        .ok()
+        .and_then(|v| v.parse::<usize>().ok())
+        .map(|v| v.clamp(MIN_MAX_LINES, ABSOLUTE_MAX_LINES))
```
PR-Reviewer-Ant, line 1285: **Warning: string-matching heuristics for `ErrorCategory` are overly broad.**

`is_transient` matches on bare `"500"`, `"502"`, `"connection"` etc. as substrings of a provider error *message*. A provider response body that happens to contain "500" (e.g., a transaction ID, HTTP example, or error description) will be incorrectly classified as transient and retried up to 3 times. Similarly `"connection"` is extremely common in error prose. The `Http` variant already has proper status-code access via `e.status()` — the `Provider(msg)` branch should require more specific patterns (e.g., `"HTTP 500"`, `"HTTP 502"`) or be limited to known provider error formats to avoid false-positive retries.

packages/forge/crates/forge-cli/src/main.rs
```diff
@@ -131,19 +143,69 @@ async fn main() -> Result<()> {
...
+                workspace.display()
+            )
+        })?;
+        info!(
+            "Agent workspace active: {} -> {}",
+            agent_name,
+            workspace.display()
+        );
+    }
+
```
PR-Reviewer-Ant, line 164: **Warning: `std::env::set_var` called from an active Tokio async runtime.**

`FORGE_WORKSPACE_ONLY`, `FORGE_ALLOWED_PATHS`, and `FORGE_ALLOWED_COMMANDS` are set via `std::env::set_var` inside `#[tokio::main]`, which has already spawned its thread pool. `set_var` is not thread-safe (it is unsound in Rust 1.81+ per the `unsafe` lint, and causes UB in multi-threaded contexts). These assignments should be moved to the earliest possible point in `main` before `tokio::main` initialises — or the values should be passed as explicit arguments rather than using environment variable side-channels.

packages/forge/crates/forge-agent/src/session.rs
```diff
@@ -48,17 +48,42 @@ impl Session {
...
+                continue;
+            }
+
+            let role = match m.role {
+                forge_core::Role::User => "User",
+                forge_core::Role::Assistant => "Assistant",
+                forge_core::Role::System => "System",
+            };
+
+            let mut text = m.text().replace("\r\n", "\n").replace('\r', "\n");
```
PR-Reviewer-Ant, line 68: **Nitpick: unreachable match arm after system-message filter.**

The `for m in &self.messages` loop skips system messages with `continue`, so the `forge_core::Role::System => "System"` arm in the subsequent `match m.role` block is dead code. Remove it or replace with `unreachable!()`.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -504,6 +1134,425 @@ impl AgentLoop {
...
+    let mut suggestions = suggest_tool_names(&tc.name, available_tools, 5);
+    if suggestions.is_empty() {
+        suggestions = available_tools.iter().take(5).cloned().collect();
+    }
+
+    let suggestion_text = if suggestions.is_empty() {
+        "No registered tools were available.".to_string()
+    } else {
+        format!("Try one of: {}", suggestions.join(", "))
+    };
```
PR-Reviewer-Ant, line 1222: **Panic on non-ASCII input.** `text.len()` returns the byte count, but `&text[..397]` indexes by bytes. If any multi-byte UTF-8 character (e.g. an accented letter, CJK character, or emoji) straddles byte position 397, Rust panics at runtime with 'byte index is not a char boundary'. Because this function is invoked whenever the active model changes mid-session, any user with non-ASCII content in their recent messages will crash the agent loop. Replace with char-safe slicing:
```rust
let clipped = if text.chars().count() > 400 {
    let cut: String = text.chars().take(397).collect();
    format!("{}...", cut)
} else {
    text
};
```

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -504,6 +1134,425 @@ impl AgentLoop {
...
+    let budget_after_messages = remaining.saturating_sub(estimated_message_tokens);
+    budget_after_messages.min(soft_cap)
+}
+
+fn context_tier(max_tokens: usize, estimated_message_tokens: usize) -> ContextTier {
+    let remaining = max_tokens
+        .saturating_mul(78)
+        .saturating_div(100)
+        .saturating_sub(estimated_message_tokens);
+    let ratio_pct = if max_tokens == 0 {
```
PR-Reviewer-Ant, line 1410: **Per-line truncation result is silently discarded.** When `per_line_clamped.lines().count() <= max_lines`, the `else` branch returns `content.to_string()` (the original, unmodified content) rather than `per_line_clamped`. This means: (a) any lines truncated by the per-line pass are discarded, and (b) `did_line_truncate` can be `true` while the returned `content` is fully intact — the `truncated` flag lies. Downstream telemetry (`tuning.record_tool_result(_, true)`) will then believe truncation occurred and tighten `recommended_tool_chars`, causing unnecessary context budget shrinkage. The accompanying test `truncation_flag_set_for_line_only_truncation` only asserts `out.truncated` and does not verify that `out.content` is actually shorter, so it does not catch this regression. Fix: change `content.to_string()` to `per_line_clamped` in the else branch.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -346,10 +722,46 @@ impl AgentLoop {
...
+            // Read-only tools are parallelized; write tools run sequentially.
+            // Identical tool calls (same name + input) are deduplicated.
             let mut tool_results_content = Vec::new();
+
+            // Deduplicate tool calls by (name, input) hash.
+            // duplicate_to_original maps each skipped call's ID → the original call's ID
+            // so we can copy results after execution to satisfy the API contract.
+            let mut seen_hashes = std::collections::HashSet::new();
+            let mut deduped_calls: Vec<&ToolCall> = Vec::new();
+            let mut hash_to_original: std::collections::HashMap<u64, String> = std::collections::HashMap::new();
```
PR-Reviewer-Ant, line 756: **Hash-only deduplication with no equality check.** `DefaultHasher` produces a 64-bit non-cryptographic hash with no collision guarantees, and its output is explicitly not stable across Rust releases. If two distinct tool calls `(name_A, input_A)` and `(name_B, input_B)` collide to the same u64, the second call is silently dropped and its tool_use ID gets a copied result from the wrong call — the LLM receives an incorrect tool result with no error signal. The probability is low per turn but non-zero over many sessions and large tool registries. Fix: store the full `(name, serialized_input)` pair in the seen-set for exact equality, using the hash only as a fast pre-filter:
```rust
let key = (tc.name.clone(), serde_json::to_string(&tc.input).unwrap_or_default());
if seen_keys.insert(key.clone()) { ... } else { ... }
```

packages/forge/crates/forge-cli/src/main.rs
```diff
@@ -113,6 +124,86 @@ async fn main() -> Result<()> {
...
+            anyhow::anyhow!(
+                "Failed to switch to agent workspace {}: {e}",
+                workspace.display()
+            )
+        })?;
+        info!(
+            "Agent workspace active: {} -> {}",
+            agent_name,
+            workspace.display()
+        );
```
PR-Reviewer-Ant, line 162: **Undocumented breaking behavior change: `FORGE_WORKSPACE_ONLY=1` is set even without `--agent`.** `selected_policy` is `AgentExecutionPolicy::default()` when no `--agent` flag is provided. `default()` leaves `workspace_only` as `None`, so `unwrap_or(true)` silently evaluates to `true` and sets `FORGE_WORKSPACE_ONLY=1` for every forge invocation regardless of agent selection. Users who previously ran bare `forge` with full filesystem access will have their tool calls silently restricted to the workspace directory after this update. This behavior change is not mentioned in the PR description or docs. Either default to `false` for non-agent invocations, or guard the set_var behind `if selected_agent.is_some()`.

docs/specs/planning/forge-llm-performance-roadmap.md
```diff
@@ -0,0 +1,83 @@
+# Forge LLM Performance Roadmap
```
PR-Reviewer-Ant, line 1: The AGENTS.md session protocol requires `INDEX.md` and `dependencies.yaml` to be kept in sync when adding spec work. A new planning spec is added here, but neither `docs/specs/INDEX.md` nor `docs/specs/dependencies.yaml` are visible in this diff. If they were updated, the PR description should reference it; if not, this is a spec drift violation per the recurring failure mode #4.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -504,6 +1156,425 @@ impl AgentLoop {
...
+    max_tokens: usize,
+    estimated_message_tokens: usize,
+) -> usize {
+    let soft_cap = max_tokens.saturating_mul(18) / 100;
+    let remaining = max_tokens.saturating_mul(78) / 100;
+    let budget_after_messages = remaining.saturating_sub(estimated_message_tokens);
+    budget_after_messages.min(soft_cap)
+}
+
+fn context_tier(max_tokens: usize, estimated_message_tokens: usize) -> ContextTier {
```
PR-Reviewer-Ant, line 1405: **Blocking: memory injection silently disabled for providers with unknown context window.**

`compute_available_injection_tokens` returns 0 when `max_tokens == 0` (both `soft_cap` and `remaining` saturate to 0). `compact_memory_context_by_tier` then hits its early-return guard `if memory_context.is_empty() || available_tokens == 0 { return String::new(); }`, silently dropping all recalled memories.

Before this PR, memory was injected unconditionally: `format!("{}

{}", self.system_prompt, memory_context)`. Any provider whose `context_window()` returns 0 (unknown/unset) now receives zero memory injection with no log or error — a silent behavioral regression.

Fix: when `max_tokens == 0`, either bypass the budget arithmetic and pass through the full context (matching prior behavior), or clamp to a sensible default (e.g. 8 000 tokens) and log a warning.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -173,54 +459,156 @@ impl AgentLoop {
...
+                        "Turn plan: {} (effort={}, max_tokens={}, tool_cap={})",
+                        turn_plan.label(),
+                        effective_effort.as_str(),
+                        turn_plan.max_tokens(),
+                        dynamic_tool_max_chars
+                    )))
+                    .await;
+            }
+
+            // Pre-sampling compaction: check token estimate BEFORE calling
```
PR-Reviewer-Ant, line 519: **Warning: pre-sampling compaction uses a different token estimate than the turn plan, causing inconsistent trigger points.**

Earlier in the same loop body (around the `choose_turn_plan` call), `estimated_messages_tokens` is computed as:
```rust
let raw = ContextManager::estimate_tokens(&s.messages);
let adjusted = adjust_estimate_for_model(raw, provider, model);
```
But the pre-sampling compaction block re-estimates independently:
```rust
let estimated = ContextManager::estimate_tokens(&s.messages); // raw only
if self.context_manager.should_compact_before_sampling(estimated) { … }
```
For providers like Gemini (`chars_per_token ≈ 4.2`), `adjust_estimate_for_model` scales the raw count by `4.0 / 4.2 ≈ 0.95`, so the adjusted estimate is *lower* than raw. The turn plan thus sees less pressure than the compaction gate does, meaning compaction fires before the turn plan would have promoted to `Short`. For Anthropic/Claude (`chars_per_token ≈ 3.7`), the factor is `~1.08`, so compaction fires *later* than the budget logic expects.

Fix: use the already-computed `estimated_messages_tokens` (the adjusted value) for the `should_compact_before_sampling` call.

packages/forge/crates/forge-cli/src/main.rs
```diff
@@ -597,6 +820,16 @@ fn select_provider(cli: &Cli, available: &[DiscoveredProvider]) -> Result<(Strin
         return Ok((provider.clone(), model));
     }
 
+    // Phase B routing: if a model is explicitly selected (CLI or agent profile)
+    // and provider is omitted, infer provider from model family.
+    if let Some(model) = cli.model.clone() {
```
PR-Reviewer-Ant, line 825: **Warning: Ollama is selected even when it is not in the available providers list.**

```rust
if has_inferred || inferred == "ollama" {
    return Ok((inferred, model));
}
```
The `inferred == "ollama"` arm bypasses the `has_inferred` guard entirely, so if a user specifies a model whose name maps to Ollama (e.g. `qwen3:4b`) and Ollama is not running or not discovered, the CLI returns `("ollama", model)` anyway. The subsequent provider call will then fail at the first LLM request rather than at startup, producing a confusing late error instead of a "no provider available" message.

Fix: Remove the `|| inferred == "ollama"` bypass or, if Ollama warrants special treatment (e.g. it is always assumed available locally), add an explicit comment and a connectivity pre-check.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -79,6 +191,20 @@ impl AgentLoop {
         mcp_clients: Vec<Arc<forge_mcp::McpStdioClient>>,
     ) -> Self {
         let context_window = provider.context_window();
+
+        // Try to use Ollama as a cheap compaction provider.
+        // Falls back to the primary model if Ollama isn't available.
+        let compaction_provider: Option<Arc<dyn Provider>> = {
```
PR-Reviewer-Ant, line 197: **Warning: "Ollama available" log fires on provider object creation, not connectivity.**

```rust
let ollama = forge_provider::create_provider("ollama", "qwen3:4b", "ollama");
match ollama {
    Ok(p) => {
        info!("Ollama available — using qwen3:4b for context compaction");
        Some(Arc::from(p))
    }
```
`create_provider` constructs a client struct; it does not probe the Ollama HTTP endpoint. The log message will appear even when Ollama is not installed or not running, misleading operators into thinking compaction will use the cheaper local model. The actual failure is deferred until compaction is needed, at which point the fallback to the primary provider silently takes over.

Fix: Either rename the log to `info!("Ollama endpoint configured for compaction (availability unverified)")` or add a lightweight health-check (e.g. `GET /api/tags`) before logging availability.

packages/forge/crates/forge-agent/src/retry.rs
```diff
@@ -0,0 +1,119 @@
+use forge_core::{ErrorCategory, ForgeError};
```
PR-Reviewer-Ant, line 1: **Warning: no unit tests for the retry module.**

`RetryState::decide` is a stateful, branching function (5 match arms, 3 independent counters, exponential backoff calculation with `saturating_pow`). The repo's testing philosophy explicitly requires tests for every new feature describing behavioral contracts, and the recurring PR failure pattern (#6 in AGENTS.md) flags missing regression tests. Given that retry logic directly impacts reliability under provider errors and rate limiting, this module needs at minimum:
- a test that verifies `Retry(delay)` is returned for transient errors up to `transient_retries` then `Fail`
- a test that verifies exponential backoff doubling
- a test that verifies `CompactAndRetry` fires exactly once before `Fail` on overflow
- a test that `Client` and `Fatal` errors are never retried

packages/cli/src/features/forge.ts
```diff
@@ -751,3 +771,84 @@ export async function doctorForge(options: ForgeStatusOptions, deps: ForgeDeps):
...
+	if (running) {
+		const stopped = await deps.stopDaemon(basePath);
+		if (!stopped) {
+			throw new Error("Failed to stop Forge service daemon.");
+		}
+	}
+	const started = await deps.startDaemon(basePath);
+	if (!started) {
```
PR-Reviewer-Ant, line 822: **Nitpick: TOCTOU between `isDaemonRunning()` and `getDaemonStatus()` in `showForgeServiceStatus`.**

Both are called in sequence with no synchronization. If the daemon stops between the two calls, the output shows `running: true` with `pid: null` (or vice versa on a start), producing a self-contradictory status display. Since the daemon is a singleton and the status command is read-only, this is low-severity but worth collapsing into a single `getDaemonStatus()` call whose `running` field drives the `running:` output line.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -346,10 +743,46 @@ impl AgentLoop {
...
+                } else if let Some(original_id) = hash_to_original.get(&hash) {
+                    debug!("Deduplicating tool call: {} (duplicate of {})", tc.id, original_id);
+                    duplicate_to_original.push((tc.id.clone(), original_id.clone()));
+                }
+            }
+
+            if deduped_calls.len() < tool_calls.len() {
+                info!(
+                    "Deduplicated {} → {} tool calls",
+                    tool_calls.len(),
```
PR-Reviewer-Ant, line 775: **DefaultHasher used for correctness-critical tool deduplication.** `std::collections::hash_map::DefaultHasher` provides no collision-resistance guarantee and its output is explicitly unstable across Rust versions and platforms. Two structurally distinct tool calls — e.g. `Read` with path `a/b` and `Glob` with pattern `a/b` — could collide, causing one to be silently dropped. Because each `ToolCall` already has a unique `id`, the safest fix is to deduplicate by comparing `(name, canonical_json_input)` strings directly (via a `HashSet<(String, String)>`) rather than hashing them with a non-cryptographic, non-stable hasher. If a hash is required for performance, use `FxHasher` or `AHasher` — both have better collision characteristics than `DefaultHasher`.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -504,6 +1155,434 @@ impl AgentLoop {
...
+            }
+            let head_chars = max_chars.saturating_mul(3) / 5;
+            let tail_chars = max_chars.saturating_sub(head_chars);
+
+            let head: String = memory_context.chars().take(head_chars).collect();
+            let tail_rev: String = memory_context.chars().rev().take(tail_chars).collect();
+            let tail: String = tail_rev.chars().rev().collect();
+            format!(
+                "{head}\n\n[adaptive compaction (warm tier): middle context omitted]\n\n{tail}"
+            )
```
PR-Reviewer-Ant, line 1470: **`context_tier` returns `Cold` when `max_tokens == 0`, but `compute_available_injection_tokens` defaults to 8,000 tokens for the same case.** The two functions handle the unknown-window case differently: `context_tier(0, _)` sets `ratio_pct = 0` and returns `ContextTier::Cold`, triggering aggressive salient-line filtering. `compute_available_injection_tokens(0, _)` logs a warning and uses `DEFAULT_CONTEXT_WINDOW = 8_000`, which may produce a non-zero budget. `prepare_memory_context_for_turn` calls both and passes the results to `compact_memory_context_by_tier`, so a session with an unknown context window silently loses most of its injected memory even when there is budget to spare. Fix: use a shared effective-window helper so both functions agree, or explicitly map the unknown case to `Hot` in `context_tier`.

packages/cli/src/features/forge.ts
```diff
@@ -751,3 +771,84 @@ export async function doctorForge(options: ForgeStatusOptions, deps: ForgeDeps):
...
+	console.log(chalk.green("✓ Forge service restarted"));
+}
+
+export async function showForgeServiceStatus(options: ForgeServiceOptions, deps: ForgeDeps): Promise<void> {
+	const basePath = readForgeServicePath(options, deps);
+	const daemon = await deps.getDaemonStatus();
+	const daemonRunning = daemon.pid != null;
+	const payload = {
+		running: daemonRunning,
+		pid: daemon.pid,
```
PR-Reviewer-Ant, line 834: **`showForgeServiceStatus` determines running state from `daemon.pid != null` instead of `daemon.running`.** The `ForgeDaemonStatus` type returned by `getDaemonStatus()` already has an authoritative `running: boolean` field. Using `pid != null` as a proxy silently reports `running: false` for any daemon managed as a system service that does not expose a PID (e.g., systemd socket-activated units). The JSON payload also sets `running: daemonRunning` derived from the PID check, so both the human-readable and JSON outputs can be wrong. Change to `const daemonRunning = daemon.running;`.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -504,6 +1155,434 @@ impl AgentLoop {
...
+         Previous model: {previous}\n\
+         Current model: {current}\n\
+         Preserve continuity with this recent context:\n{tail}"
+    )
+}
+
+fn invalid_tool_call_result(tc: &ToolCall, available_tools: &[String]) -> forge_core::ToolResult {
+    let mut suggestions = suggest_tool_names(&tc.name, available_tools, 5);
+    if suggestions.is_empty() {
+        suggestions = available_tools.iter().take(5).cloned().collect();
```
PR-Reviewer-Ant, line 1215: **`char_indices()` byte offsets compared against char-count limits throughout truncation functions.** `str::char_indices()` yields `(byte_offset, char)` pairs. The guards `*idx < max_chars` and `*idx < max_line_chars` compare a byte offset against what the callers treat as a character count (e.g., `max_chars = available_tokens * 4`). For ASCII-only content this is harmless, but for multi-byte UTF-8 (CJK, emoji, accented text) the cut fires earlier than intended, producing outputs shorter than the budget allows. All three truncation sites in `clamp_tool_content_with_dynamic_max`, `clamp_memory_context`, and the per-line loop should use `.char_indices().filter(|(_, ch)| …).nth(limit - 1)` or count characters explicitly.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -136,13 +403,21 @@ impl AgentLoop {
                 .await;
```
PR-Reviewer-Ant, line 403: **Last 6 session messages are embedded verbatim in the memory-recall query on model switch.** `build_model_switch_handoff` summarizes up to 6 recent messages (each clipped to 400 chars) and `recall_query` prepends this handoff to the user's prompt before it is sent to `hooks.prompt_submit`. If session content is sensitive (credentials, PII, proprietary code), it is now included in a daemon recall call and will appear in any daemon-side query logs or telemetry. Consider sending the handoff only as an in-context injection (it already is, a few lines later) rather than embedding it in the recall query string.

packages/forge/crates/forge-agent/src/agent_loop.rs
```diff
@@ -79,6 +191,20 @@ impl AgentLoop {
         mcp_clients: Vec<Arc<forge_mcp::McpStdioClient>>,
     ) -> Self {
         let context_window = provider.context_window();
+
+        // Try to use Ollama as a cheap compaction provider.
+        // Falls back to the primary model if Ollama isn't available.
+        let compaction_provider: Option<Arc<dyn Provider>> = {
+            let ollama = forge_provider::create_provider("ollama", "qwen3:4b", "ollama");
+            match ollama {
+                Ok(p) => {
```
PR-Reviewer-Ant, line 200: **Hardcoded `qwen3:4b` model for compaction provider with no user config path.** If Ollama is reachable but `qwen3:4b` has not been pulled, the first compaction attempt logs success at `create_provider` time, fails silently at `complete` time, and falls back to the expensive primary model — with no user-visible warning at startup. The `info!` message 'availability unverified' is too quiet for an operator to notice. At minimum, expose the compaction model as a config field in `agent.yaml` / `AgentExecutionPolicy` so users can override it, and emit a `warn!` on first fallback.

packages/cli/src/commands/forge.ts
```diff
@@ -10,10 +10,19 @@ interface ForgeInstallOptions {
 	yes?: boolean;
 }
```
PR-Reviewer-Ant, line 11: **`ForgeServiceOptions` is defined identically in both `commands/forge.ts` and `features/forge.ts`.** The repo conventions explicitly flag duplicate type/interface definitions as a code smell. The command-layer interface should import and re-export the canonical definition from `features/forge.ts`, or both should share a common types module.

docs/specs/planning/forge-llm-performance-roadmap.md
```diff
@@ -0,0 +1,83 @@
...
+  - dedupe/trim repeated context blocks where safe
+
+### Phase 2
+
+- Adaptive context tiers (hot/warm/cold by remaining budget)
+- Structured compaction artifacts over raw transcript chunks
+- Tool-result compression policy for verbose outputs
+
+### Phase 3
+
```
PR-Reviewer-Ant, line 32: **Stale branch reference checked into a planning spec.** `'Status: implemented in forge-agent on branch feat/forge-agent-loop-improvements'` names a feature branch. Once this PR merges, that branch reference is incorrect and misleads future readers about where to find the implementation. Replace with a link to this PR number or the relevant module path.

PR #273, opened 2026-03-22
fix: codex MCP config uses string command, not array

aaf2tbz:
## Summary

- **Bug**: Codex connector wrote `command = ['signet-mcp']` (TOML array) but Codex CLI expects `command = 'signet-mcp'` (string). This broke **all** `codex` CLI operations with: `invalid type: sequence, expected a string in mcp_servers.signet.command`
- **Fix**: `patchConfigToml` now writes `command` as a quoted string with a separate `args` array for Windows multi-part commands
- **README**: Updated Codex from "In progress | WIP" to "Supported | Hooks + MCP server"

### Verified
```
$ codex mcp list
Name    Command     Args  Env  Cwd  Status   Auth
signet  signet-mcp  -     -    -    enabled  Unsupported
```

## Test plan

- [x] `bun run build` passes
- [x] `codex mcp list` succeeds with fixed format
- [ ] `signet setup` → select Codex → verify config.toml format
- [ ] Run a Codex session → verify MCP tools are available

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #421, opened 2026-03-31
fix(codex): repair stale MCP config on setup instead of skipping

NicholaiVogel:
## Summary

- Fixes "invalid transport in 'mcp_servers.signet'" error reported by users after upgrading to v0.91.1
- `patchConfigToml()` now strips and re-adds the MCP section on every install instead of skipping when it already exists
- Users with stale array-format `command = ['signet-mcp']` (pre-#273) will be automatically repaired on next `signet setup`

## Root Cause

`patchConfigToml()` had a skip-if-exists guard (`if (content.includes("[mcp_servers.signet]")) return false`) that prevented it from ever fixing broken config entries. Users who installed before PR #273 had `command` as a TOML array instead of a string. Codex's Rust parser expects `Option<String>`, silently drops the array to `None`, finds neither `command` nor `url`, and throws "invalid transport."

## Test plan

- [x] 12 regression tests added covering fresh creation, stale config repair, idempotency, section preservation, and Windows paths
- [x] `bun test packages/connector-codex/` passes
- [x] `bun run build` passes

### DIFF
--- a/packages/connector-codex/src/config-toml.test.ts
+++ b/packages/connector-codex/src/config-toml.test.ts
@@ -9,7 +9,7 @@ import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { tmpdir } from "node:os";
-import { CodexConnector } from "./index.js";
+import { CodexConnector, buildMcpBlock } from "./index.js";
 
 class TempConnector extends CodexConnector {
 	constructor(private home: string) {
@@ -134,5 +134,37 @@ describe("CodexConnector.uninstall — config.toml cleanup", () => {
 		const content = readFileSync(configPath, "utf-8");
 		expect(content).toContain("[model]");
 		expect(content).not.toContain("[mcp_servers.signet]");
+	});
+});
+
+// buildMcpBlock is tested directly here because resolveSignetMcp() always
+// returns the non-Windows path on Linux, so Windows quoting can't be
+// exercised through install().
+describe("buildMcpBlock — TOML quoting", () => {
+	test("produces string command, not array", () => {
+		const block = buildMcpBlock({ command: "signet-mcp", args: [] });
+		expect(block).toContain("command = 'signet-mcp'");
+		expect(block).not.toContain("command = [");
+	});
+
+	test("Windows paths with backslashes are quoted correctly", () => {
+		const block = buildMcpBlock({
+			command: "C:\\Program Files\\node.exe",
+			args: ["C:\\signet\\mcp-stdio.js"],
+		});
+		// No single-quote in the path, so literal single-quote TOML strings are used
+		expect(block).toContain("command = 'C:\\Program Files\\node.exe'");
+		expect(block).toContain("args = ['C:\\signet\\mcp-stdio.js']");
+		expect(block).not.toContain("command = [");
+	});
+
+	test("omits args line when args is empty", () => {
+		const block = buildMcpBlock({ command: "signet-mcp", args: [] });
+		expect(block).not.toContain("args");
+	});
+
+	test("includes args line when args are present", () => {
+		const block = buildMcpBlock({ command: "node", args: ["mcp.js", "--port", "3000"] });
+		expect(block).toContain("args = ['mcp.js', '--port', '3000']");
 	});
 });
--- a/packages/connector-codex/src/config-toml.test.ts
+++ b/packages/connector-codex/src/config-toml.test.ts
@@ -135,6 +135,39 @@ describe("CodexConnector.uninstall — config.toml cleanup", () => {
 		expect(content).toContain("[model]");
 		expect(content).not.toContain("[mcp_servers.signet]");
 	});
+
+	test("handles multi-line TOML args without corrupting surrounding sections (regression: unpatchConfigToml)", async () => {
+		// A user who hand-edited args to multi-line form would have had
+		// continuation lines left in the file by the old section-end detection.
+		writeFileSync(
+			configPath,
+			[
+				"[other]",
+				"key = 'val'",
+				"",
+				"# Signet MCP server",
+				"[mcp_servers.signet]",
+				"command = 'signet-mcp'",
+				"args = [",
+				"  '--verbose'",
+				"]",
+				"",
+				"[after]",
+				"key = 'val'",
+				"",
+			].join("\n"),
+		);
+
+		const c = connector();
+		await c.uninstall();
+
+		const content = readFileSync(configPath, "utf-8");
+		expect(content).not.toContain("[mcp_servers.signet]");
+		// Continuation lines must not leak into the output
+		expect(content).not.toContain("--verbose");
+		expect(content).toContain("[other]");
+		expect(content).toContain("[after]");
+	});
 });
 
 // buildMcpBlock is tested directly here because resolveSignetMcp() always
--- a/packages/connector-codex/src/config-toml.test.ts
+++ b/packages/connector-codex/src/config-toml.test.ts
@@ -1,0 +1,209 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+// The TOML patching functions are module-private, so we re-implement the
+// same logic inline to test the behavioral contract. These tests validate
+// that the connector's config.toml handling produces correct output for
+// Codex's Rust TOML parser, which expects `command` as a plain string.
+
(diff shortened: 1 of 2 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(codex): repair stale MCP config on setup instead of skipping` (commit `d3e75b3f`) now and I'll follow up shortly with feedback.

packages/connector-codex/src/config-toml.test.ts
```diff
@@ -0,0 +1,209 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
```
PR-Reviewer-Ant, line 8: The test file re-declares every function (`tomlQuote`, `tomlInlineArray`, `buildMcpBlock`, `patchConfigToml`, `unpatchConfigToml`) inline and imports nothing from `index.ts`. These tests exercise only their own copies — no production code path is covered. A bug introduced exclusively in `index.ts` would pass CI undetected.

Per repo conventions ("Test the contract, not the implementation" / "Tests should survive a language rewrite"), the correct approach is to test through the public interface. `patchConfigToml` is reachable via `CodexConnector.install()` and `unpatchConfigToml` via `CodexConnector.uninstall()`. Integration-style tests calling those methods against a temp directory would cover the actual production code and satisfy the behavioral-contract requirement without duplicating implementation.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d3e75b3f`

[Automated Review - pr-reviewer bot] PR #421 correctly identifies and fixes the skip-if-exists guard that caused stale array-format MCP configs to persist. The core logic change in index.ts is sound. Two issues found: the test file does not cover the production code at all, and unpatchConfigToml can corrupt files that contain multi-line TOML array values.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both issues are directly provable from the diff. (1) The test file imports nothing from index.ts — every function is re-declared inline, so no production code path is exercised. (2) The unpatchConfigToml loop exits inSection on the first line that fails /^\s*\w+\s*=/, which means a multi-line TOML array like args = [\n  '--flag'\n] would leave the continuation lines in the output after stripping. The PR's own buildMcpBlock writes only inline single-line arrays, so the bug is dormant for Signet-generated configs, but any user who manually edited args into multi-line form would have their config corrupted on the next signet setup.


- packages/connector-codex/src/index.ts:196 - The `unpatchConfigToml` section-stripping loop exits `inSection` on the first line that does not match `/^\s*\w+\s*=/` and is not blank. This works for Signet's own inline-array format (`args = ['signet-mcp']`), but any pre-existing config.toml where `args` is written as a multi-line TOML array:
```toml
[mcp_servers.signet]
command = 'signet-mcp'
args = [
  '--flag'
]
```
would leave the continuation lines (`  '--flag'`, `]`) in `filtered` — corrupting the file on the next `signet setup`. The bug is dormant given that `buildMcpBlock` always emits single-line inline arrays, but a user who manually edited their config to multi-line style would have it corrupted. Consider detecting section end by the appearance of the next `[section]` header instead:
```typescript
if (line.match(/^\[/)) { inSection = false; }
```

packages/connector-codex/src/config-toml.test.ts
```diff
@@ -0,0 +1,138 @@
...
+		writeFileSync(configPath, '[model]\nname = "gpt-4o"\n');
+		const c = connector();
+		await c.install(tempHome);
+		await c.uninstall();
+
+		const content = readFileSync(configPath, "utf-8");
+		expect(content).toContain("[model]");
+		expect(content).not.toContain("[mcp_servers.signet]");
+	});
+});
```
PR-Reviewer-Ant, line 171: The Windows-path test (`handles Windows-style command with args`) was present in the previous revision and is absent from the rewrite. The new integration tests go through `connector().install()` which calls `resolveSignetMcp()` — but that function always returns the non-Windows path in a Linux test environment. Windows path quoting (backslash handling in `tomlQuote`) is now untested. Consider adding a test that directly verifies `tomlQuote` or `buildMcpBlock` behaviour with Windows-style paths, or explicitly note this as out of scope.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `37fd05ad`

[Automated Review - pr-reviewer bot] The test file has been properly rewritten to exercise production code via TempConnector. The previously-flagged test coverage issue is resolved. The multi-line TOML corruption bug in unpatchConfigToml remains unaddressed and no test was added for that scenario.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff shows the test file was fully rewritten to import CodexConnector and override getCodexHome() — the test coverage issue is provably resolved. The unpatchConfigToml function in index.ts is unchanged between d3e75b3f and 37fd05ad (only getCodexHome visibility was touched). The section-end detection bug is still present at the same lines in the current index.ts: when inSection=true and a non-key=value, non-empty line is encountered (e.g. a multi-line array continuation like `  '--flag'`), inSection is set false and the line is pushed to filtered, corrupting the output. No test for multi-line args was added in the new test suite.


- packages/connector-codex/src/index.ts:205 - The multi-line TOML array corruption bug flagged in the prior review remains unfixed. When `inSection=true` and a continuation line like `  '--flag'` is encountered (not matching `/^\s*\w+\s*=/` and not empty), `inSection` is set to false and that line is pushed into `filtered` — corrupting the file. The current code only stays in-section for lines matching `key = value` or blank lines; any other content (array continuation, inline comment after a closing `]` on its own line) exits early.

This is dormant for Signet-generated configs since `buildMcpBlock` always emits a single-line inline array. But a user who hand-edited their `args` to multi-line form would have their config corrupted on next `signet setup`.

The prior review suggested fixing section-end detection by checking for the next section header instead:
```typescript
if (line.match(/^\[/)) { inSection = false; }
```
This PR did not address this, and no test was added to cover the multi-line args case.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2a268d03`

[Automated Review - pr-reviewer bot] This incremental revision (37fd05ad → 2a268d03) adds a `buildMcpBlock` describe suite and exports the function, resolving both previously-open items about test coverage and Windows-path testing. One prior flagged issue remains: the multi-line TOML corruption bug in `unpatchConfigToml` is still unaddressed and no regression test was added for it.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff does not touch `unpatchConfigToml` in index.ts. The section-end detection logic at the `if (inSection)` block (lines ~196–200 of index.ts) still uses `line.match(/^\s*\w+\s*=/) || line.trim() === ''` to skip lines rather than checking for the next `[section]` header. A multi-line TOML array continuation line like `  '--flag'` or `]` would fail both predicates, flip `inSection = false`, and be pushed into `filtered` — corrupting the output file. This was explicitly flagged in both prior reviews (sha d3e75b3f and sha 37fd05ad) and the item remains [open] in the review state. The new test suite does not include a case for multi-line args. The fix and the missing test are directly provable from the current index.ts and config-toml.test.ts.


- packages/connector-codex/src/index.ts:196 - **[Re-raising: still open, no change in this diff]** `unpatchConfigToml`'s section-end detection exits `inSection` on the first line that neither matches `/^\s*\w+\s*=/` nor is blank. A manually-edited multi-line TOML array:
```toml
[mcp_servers.signet]
command = 'signet-mcp'
args = [
  '--flag'
]
```
would leave `  '--flag'` and `]` in `filtered`, corrupting the file on the next `signet setup`. The fix is to detect section end by the next `[header]` line instead:
```typescript
if (inSection) {
  if (line.match(/^\[/)) { inSection = false; }
  else { continue; }
}
```
A regression test covering this case is also missing from the new suite.

NicholaiVogel:
All three review items addressed:

1. **Test coverage** — tests were rewritten to exercise production code via a `TempConnector` subclass that overrides `getCodexHome()`. No implementation is duplicated.

2. **Windows path quoting** — `buildMcpBlock` is now exported; a `buildMcpBlock` describe suite covers backslash quoting directly since `resolveSignetMcp()` always returns the Linux path in CI.

3. **Multi-line TOML corruption in `unpatchConfigToml`** — fixed in d08652ac. Replaced the `key=value || blank` predicate with next-`[header]`-means-section-end logic:
   ```typescript
   if (inSection) {
     if (line.match(/^\[/)) inSection = false;
     else continue;
   }
   ```
   Regression test added covering a manually-edited multi-line `args` block with surrounding sections — confirms continuation lines no longer leak into the output.

PR #423, opened 2026-03-31
fix(openclaw): prevent double registration blocking gateway providers

### LINKED ISSUE #422: signet-memory-openclaw plugin blocks Telegram channel initialization on OpenClaw 2026.3.28
## Environment
- **OpenClaw:** 2026.3.28 (f9b1079)
- **Signet:** 0.91.1
- **signet-memory-openclaw:** 0.91.1
- **OS:** macOS 26.3.1 (arm64, Mac Mini M4)
- **Node:** 22.22.0
- **Ollama:** 0.19.0 (nomic-embed-text + nemotron-mini)

## Problem

When `signet-memory-openclaw` is enabled as a plugin in `openclaw.json`, the Telegram channel provider never starts. Gateway health reports `ok`, but Telegram is unavailable.

The plugin itself loads and registers successfully — logs show:
```
[plugins] signet-memory: registered (daemon: http://localhost:3850)
[plugins] signet-memory: registered (daemon: http://localhost:3850)
```

Note: the plugin registers **twice** (two log lines).

However, the Telegram provider never initializes — there is no `[telegram] starting provider` line in logs after plugin load.

## Configuration

```json
"plugins": {
  "allow": ["signet-memory-openclaw"],
  "load": {
    "paths": ["/Users/mars/.bun/install/global/node_modules/@signetai"]
  },
  "slots": {
    "memory": "signet-memory-openclaw"
  },
  "entries": {
    "signet-memory-openclaw": {
      "enabled": true
    }
  }
}
```

Signet daemon is running and healthy at localhost:3850.

## Workaround

Removing the plugin from `plugins.entries` and using only `hooks.internal.entries.signet-memory: { enabled: true }` allows Telegram to start normally. However, this disables auto-recall injection and memory tools.

## Steps to Reproduce

1. Install signetai 0.91.1 globally via bun
2. Run `signet setup` with OpenClaw workspace
3. Configure plugin in openclaw.json as shown above
4. Restart gateway: `openclaw gateway restart`
5. Telegram channel is unavailable; `[telegram] starting provider` never appears in logs

## Additional Notes

- Two unknown hooks are also logged:
  ```
  [gateway] [plugins] unknown typed hook "session:compact:before" ignored
  [gateway] [plugins] unknown typed hook "session:compact:after" ignored
  ```
- The double registration may indicate a race condition or duplicate register() call
- Tested on OpenClaw 2026.3.24 and 2026.3.28 — same behavior on both versions

NicholaiVogel:
## Summary

- Checks `api.registrationMode` and returns early on `"cli-metadata"` to skip tool/hook/service registration during the CLI metadata pass. OpenClaw calls `register()` twice by design (runtime + CLI metadata), and our plugin was running the full registration both times, producing duplicate tools and services that stalled the gateway before Telegram could initialize.
- Removes `session:compact:before` / `session:compact:after` hook registrations. These are internal OpenClaw events, not plugin API hooks. The plugin-facing equivalents (`before_compaction` / `after_compaction`) are already registered.
- Adds `PluginRegistrationMode` type to openclaw-types.ts stubs.

## Test plan

- [x] `bun run build` passes for openclaw adapter
- [x] 50/50 adapter tests pass (3 new regression tests added)
- [ ] Manual: confirm single "registered" log line when loaded by OpenClaw
- [ ] Manual: confirm Telegram channel initializes after plugin load

Closes #422

### DIFF
--- a/packages/adapters/openclaw/package.json
+++ b/packages/adapters/openclaw/package.json
@@ -21,7 +21,11 @@
       "dist/index.js"
     ]
   },
-  "clawdbot": {},
+  "clawdbot": {
+    "extensions": [
+      "dist/index.js"
+    ]
+  },
   "scripts": {
     "build": "bun build src/index.ts --outdir dist --target node && tsc --emitDeclarationOnly --declaration --outDir dist",
     "prepack": "bun run build",
--- a/packages/adapters/openclaw/package.json
+++ b/packages/adapters/openclaw/package.json
@@ -21,11 +21,7 @@
       "dist/index.js"
     ]
   },
-  "clawdbot": {
-    "extensions": [
-      "dist/index.js"
-    ]
-  },
+  "clawdbot": {},
   "scripts": {
     "build": "bun build src/index.ts --outdir dist --target node && tsc --emitDeclarationOnly --declaration --outDir dist",
     "prepack": "bun run build",
--- a/packages/adapters/openclaw/src/index.test.ts
+++ b/packages/adapters/openclaw/src/index.test.ts
@@ -22,7 +22,7 @@ mock.module("@signet/core", () => ({
 // Import after mock so the module picks up the stub.
 const signet = await import("./index");
 const signetPlugin = signet.default;
-const { memoryStore } = signet;
+const { memoryStore, _resetRegistration } = signet;
 
 type HookHandler = (event: Record<string, unknown>, ctx: unknown) => Promise<unknown> | unknown;
 type ToolRegistration = { name: string; label?: string; description?: string };
@@ -270,6 +270,7 @@ afterEach(async () => {
 	for (const service of registeredServices) {
 		await service.stop();
 	}
+	_resetRegistration();
 });
 
 describe("signet-memory-openclaw lifecycle hooks", () => {
@@ -483,14 +484,12 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		expect(result?.inject).not.toContain("daemon offline");
 	});
 
-	it("fires pre-compaction hooks once across both OpenClaw compaction event names", async () => {
+	it("fires pre-compaction hook and deduplicates identical calls", async () => {
 		const { api, hooks } = createMockApi();
 		signetPlugin.register(api);
 
 		const beforeCompaction = hooks.get("before_compaction");
-		const sessionCompactBefore = hooks.get("session:compact:before");
 		expect(beforeCompaction).toBeDefined();
-		expect(sessionCompactBefore).toBeDefined();
 
 		const event = { messageCount: 12, tokenCount: 240, compactingCount: 8 };
 		const ctx = {
@@ -500,7 +499,7 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		};
 
 		await beforeCompaction?.(event, ctx);
-		await sessionCompactBefore?.(event, ctx);
+		await beforeCompaction?.(event, ctx);
 
 		expect(getHits("/api/hooks/pre-compaction")).toBe(1);
 		expect(lastPreCompactionBody).toMatchObject({
@@ -597,9 +596,7 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		signetPlugin.register(api);
 
 		const afterCompaction = hooks.get("after_compaction");
-		const sessionCompactAfter = hooks.get("session:compact:after");
 		expect(afterCompaction).toBeDefined();
-		expect(sessionCompactAfter).toBeDefined();
 
 		const sessionFile = join(testDir, "session-after.jsonl");
 		writeFileSync(
@@ -623,7 +620,7 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		};
 
 		await afterCompaction?.(event, ctx);
-		await sessionCompactAfter?.(event, ctx);
+		await afterCompaction?.(event, ctx);
 
 		expect(getHits("/api/hooks/compaction-complete")).toBe(1);
 		expect(lastCompactionBody).toMatchObject({
@@ -723,14 +720,12 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		});
 	});
 
-	it("deduplicates duplicate compaction-complete writes even when session file visibility differs across hook aliases", async () => {
+	it("deduplicates duplicate compaction-complete writes for the same session", async () => {
 		const { api, hooks } = createMockApi();
 		signetPlugin.register(api);
 
 		const afterCompaction = hooks.get("after_compaction");
-		const sessionCompactAfter = hooks.get("session:compact:after");
 		expect(afterCompaction).toBeDefined();
-		expect(sessionCompactAfter).toBeDefined();
 
 		const sessionFile = join(testDir, "session-after-dedupe.jsonl");
 		writeFileSync(
@@ -750,7 +745,7 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 			{ summary: "Stable recovered summary.", sessionFile },
 			{ sessionKey: "session-after-dedupe", sessionFile, agentId: "agent-1" },
 		);
-		await sessionCompactAfter?.(
+		await afterCompaction?.(
 			{ summary: "Stable recovered summary." },
 			{ sessionKey: "session-after-dedupe", agentId: "agent-1" },
 		);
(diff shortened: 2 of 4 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(openclaw): prevent double registration blocking gateway providers` (commit `948e3b06`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `948e3b06`

[Automated Bot Review] PR #423 introduces a module-level double-registration guard, removes unrecognized `session:compact:before/after` hook registrations, and empties `clawdbot.extensions` in `package.json`. The core fix is sound. Two issues are worth discussing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The registration guard logic and test coverage are fully visible in the diff and are correct. Confidence is medium because `openclaw.plugin.json` and `clawdbot.plugin.json` — both listed in `package.json#files` — are not shown in the diff. If `clawdbot.plugin.json` independently declares an `extensions` entry pointing to `dist/index.js`, the `package.json` change is redundant (the guard catches it anyway), but that file's contents should be verified to confirm the stated root cause.

packages/adapters/openclaw/package.json
```diff
@@ -21,11 +21,7 @@
       "dist/index.js"
     ]
   },
-  "clawdbot": {
-    "extensions": [
-      "dist/index.js"
-    ]
-  },
+  "clawdbot": {},
```
PR-Reviewer-Ant: `clawdbot.plugin.json` is in the `files` list but not shown in this diff. If it carries its own `extensions: ["dist/index.js"]` entry, the `package.json` change here is incomplete as a preventive measure (though the module-level `registered` guard in `index.ts` would still catch the duplicate call and log a warning). Please confirm `clawdbot.plugin.json` does not re-list the extension, or update it in the same PR to stay consistent with the stated fix.

packages/adapters/openclaw/src/index.ts
```diff
@@ -2026,4 +2034,9 @@ const signetPlugin = {
 	},
 };
 
+/** @internal Test-only: reset the module-level registration guard. */
```
PR-Reviewer-Ant: `_resetRegistration` is exported as a named export from the production module, not just the test build. Anyone who imports this package can call it to silently disable the double-registration guard. Consider one of: (a) move it behind a `process.env.NODE_ENV === 'test'` conditional guard, (b) strip it at build time with a banner/define, or (c) accept the risk with a note that this is intentional (the `@internal` JSDoc helps, but does not prevent use). As a runtime plugin distributed on npm, the exposure is low but non-zero.

packages/adapters/openclaw/package.json
NicholaiVogel: Confirmed -- `clawdbot.plugin.json` contains only `id`, `kind`, `uiHints`, and `configSchema`. No `extensions` field. The double-load comes exclusively from the `package.json` fields, so this change is sufficient.

packages/adapters/openclaw/src/index.ts
NicholaiVogel: Good catch. Gated it behind `process.env.NODE_ENV === 'test'` so it's a no-op in production. Pushed in 46f1c99b.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `46f1c99b`

[Automated Bot Review] PR #423 — incremental diff (948e3b0 → 46f1c99) addresses the prior `_resetRegistration` concern via a `NODE_ENV === "test"` guard. One prior open item (`clawdbot.plugin.json`) remains unresolved.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single, self-contained change: wrapping `registered = false` in a `process.env.NODE_ENV === 'test'` guard. That change directly and completely addresses the prior reviewer concern about `_resetRegistration` being callable from production code. The remaining open item (`clawdbot.plugin.json`) is not affected by this delta — it was already flagged in the prior round and still has no diff evidence resolving it.

packages/adapters/openclaw/src/index.ts
```diff
@@ -2026,4 +2034,11 @@ const signetPlugin = {
 	},
 };
 
+/** @internal Test-only: reset the module-level registration guard. No-op in production. */
```
PR-Reviewer-Ant: Prior concern resolved. The `NODE_ENV === 'test'` guard makes `_resetRegistration` a no-op in production — calling it from outside the test suite can no longer bypass the double-registration guard. No further action needed on this item.

packages/adapters/openclaw/package.json
PR-Reviewer-Ant: `clawdbot.plugin.json` is listed in `files` but is absent from this diff. The PR empties `clawdbot.extensions` in `package.json`, but if `clawdbot.plugin.json` independently declares `"extensions": ["dist/index.js"]`, the stated root-cause prevention is incomplete. The module-level guard catches the duplicate load regardless, so this is not a correctness blocker — but the PR description's claim about preventing double-loading via the manifest change may be only partially true until that file is verified or also patched.

packages/adapters/openclaw/src/index.ts
```diff
@@ -1131,6 +1131,12 @@ const signetPlugin = {
 	configSchema: signetConfigSchema,
 
 	register(api: OpenClawPluginApi): void {
+		// OpenClaw calls register() twice: once for the full runtime pass and
```
PR-Reviewer-Ant, line 1140: **Warning — the new guard is strictly weaker than the one it replaces.**

The removed module-level `registered` flag was unconditional: any second call to `register()` was rejected regardless of `registrationMode`. The replacement only returns early when `api.registrationMode === 'cli-metadata'`. Because `registrationMode` is typed `readonly registrationMode?: PluginRegistrationMode` (optional), any call where the field is absent or `'full'` passes straight through. If OpenClaw ever calls twice with `registrationMode: 'full'` — or calls once from a version that doesn't set the field at all — the result is duplicate hooks, tools, timers, and services.

If the intent is that the mode-based check is *sufficient* because OpenClaw never double-calls with `'full'`, that assumption should be documented and ideally guarded with a defensive fallback (e.g. a scoped per-`api`-instance seen-check as a backstop).

packages/adapters/openclaw/src/index.test.ts
```diff
@@ -1520,3 +1516,34 @@ describe("signet-memory-openclaw lifecycle hooks", () => {
 		expect(refreshedNames.some((name) => name === "signet_server_a_beta_2")).toBeFalse();
 	});
```
PR-Reviewer-Ant, line 1518: **Warning — regression tests for the original #422 scenario were removed without a structural replacement.**

The previous test suite had three tests under `'double registration guard (#422)'` that directly verified the core invariant: a second `register()` call on a different API instance is a no-op. Those tests are deleted and replaced with `cli-metadata` mode tests, which only exercise the *new* fast-path. There is now no test that would catch a regression where two `full`-mode `register()` calls produce duplicate tools or services. Per repo conventions, every bug fix should include a test that would fail before the fix.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `65b5131e`

[Automated Bot Review] PR #423 incremental diff (46f1c99 → 65b5131) introduces a direct contradiction between the PR description and the `package.json` change, and weakens the double-registration protection in a way that could reintroduce the original #422 bug.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The contradiction is explicit in the diff: the PR description states 'Empties `clawdbot.extensions` in `package.json`' but the incremental diff shows `clawdbot.extensions: ["dist/index.js"]` being *added* (from `{}` to a populated entry). The guard regression is provable from the diff: the module-level `registered` flag and the three double-registration guard tests are removed, and the only remaining protection (`registrationMode === 'cli-metadata'`) is narrower and optional-typed. No cross-module context is needed — the risk is fully visible in the changed lines.

packages/adapters/openclaw/src/index.ts
NicholaiVogel: Good call. Restored the module-level `registered` flag as a defensive backstop alongside the `registrationMode` check. The mode check handles the documented OpenClaw double-call pattern, the module-level guard catches any scenario where `register()` is called twice with `"full"` or with the field absent. Both layers, belt and suspenders. Pushed in e2119acc.

packages/adapters/openclaw/src/index.test.ts
NicholaiVogel: Restored. The test suite now has both: cli-metadata mode tests (verify the documented OpenClaw pattern) and double full-mode registration tests (verify the backstop guard catches duplicate tools/hooks/services). 52 tests pass.

packages/adapters/openclaw/package.json
NicholaiVogel: Resolved. The `clawdbot.extensions` removal was reverted in 65b5131e after verifying OpenClaw only reads `openclaw.extensions` (`LEGACY_PROJECT_NAMES` is empty). The `_resetRegistration` export is back with its `NODE_ENV` guard.

PR #425, opened 2026-03-31
fix(pipeline): support Copilot models via OpenCode structured output

NicholaiVogel:
## Summary

- Use OpenCode's native `json_schema` structured output format for all pipeline LLM requests, which forces models to use the `StructuredOutput` tool and returns parsed JSON in `info.structured`
- Add `github-copilot/gpt-4o`, `gpt-4o-mini`, `o3-mini`, `claude-sonnet-4` to the OpenCode model registry
- Improve parsing robustness: strip leading non-JSON text, better diagnostic logging

Fixes parsing failures reported by a user when using free Copilot models through OpenCode as the extraction/synthesis provider. The root cause was that weaker models don't reliably produce clean JSON in text output, but OpenCode already supports structured output with auto-retry, which we weren't using.

## Test plan

- [x] `bun test packages/daemon/src/pipeline/provider.test.ts` (53 tests pass, 4 new)
- [x] `bun test packages/daemon/src/pipeline/extraction.test.ts` (23 tests pass, 6 new)
- [x] `bun run build` succeeds
- [ ] Manual: configure `extraction.provider: opencode` with `model: github-copilot/gpt-4o`, verify extraction works and `info.structured` is used

### DIFF
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -937,9 +937,9 @@ describe("createOpenCodeProvider", () => {
 		expect(JSON.parse(result).description).toBe("test skill");
 	});
 
-	it("generate() disables structured output on 400 and retries without format", async () => {
+	it("generate() disables structured output on 422 and retries without format", async () => {
 		let attempts = 0;
-		let lastBody: Record<string, unknown> = {};
+		const bodies: Record<string, unknown>[] = [];
 		mockFetch(async (url, init) => {
 			if (url.includes("/session") && !url.includes("/message")) {
 				return Response.json({
@@ -952,18 +952,50 @@ describe("createOpenCodeProvider", () => {
 				});
 			}
 			attempts++;
-			lastBody = JSON.parse(init?.body as string);
+			bodies.push(JSON.parse(init?.body as string));
 			if (attempts === 1) {
-				return new Response('{"error":"validation failed for format field"}', { status: 400 });
+				// 422 with "format" JSON key — signals structured output unsupported
+				return new Response('{"issues":[{"path":["format"],"message":"Unrecognized key"}]}', { status: 422 });
 			}
 			return Response.json(openCodeResponse("fallback works"));
 		});
 
 		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
-		const result = await provider.generate("test");
-		expect(result).toBe("fallback works");
-		// After disabling, subsequent calls should not include format
-		expect(lastBody.format).toBeUndefined();
+		const first = await provider.generate("test");
+		expect(first).toBe("fallback works");
+		// The retry within the same call should omit format
+		expect(bodies[1]?.format).toBeUndefined();
+
+		// A subsequent call should also omit format (structured output stays disabled)
+		await provider.generate("second call");
+		expect(bodies[2]?.format).toBeUndefined();
+	});
+
+	it("generate() does not disable structured output on an unrelated 400", async () => {
+		let attempts = 0;
+		mockFetch(async (url, init) => {
+			if (url.includes("/session") && !url.includes("/message")) {
+				return Response.json({
+					id: `ses_unrelated_${attempts}`,
+					slug: "test",
+					projectID: "p",
+					directory: "/tmp",
+					title: "test",
+					version: "1",
+				});
+			}
+			attempts++;
+			if (attempts === 1) {
+				// 400 with "format" word but not a structured-output rejection
+				return new Response('{"error":"Invalid request format: parts array is missing"}', { status: 400 });
+			}
+			return Response.json(openCodeResponse("ok"));
+		});
+
+		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
+		// Should throw, not silently disable structured output and retry
+		await expect(provider.generate("test")).rejects.toThrow(/OpenCode HTTP 400/);
+		expect(attempts).toBe(1);
 	});
 
 	it("generate() preserves error body on non-format 400", async () => {
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -706,6 +706,13 @@ describe("createOpenCodeProvider", () => {
 
 		expect(capturedBody.parts).toEqual([{ type: "text", text: "my prompt" }]);
 		expect(capturedBody.model).toEqual({ providerID: "google", modelID: "gemini-2.5-flash" });
+		// Structured output fields are included by default
+		expect(capturedBody.format).toEqual({
+			type: "json_schema",
+			schema: { type: "object", additionalProperties: true },
+			retryCount: 1,
+		});
+		expect(typeof capturedBody.system).toBe("string");
 	});
 
 	it("generate() joins multiple text parts", async () => {
@@ -865,6 +872,103 @@ describe("createOpenCodeProvider", () => {
 		expect(seenUrls).toContain("http://172.17.0.1:11434/api/generate");
 		const fallbackOptions = fallbackBody ? getObjectField(fallbackBody, "options") : undefined;
 		expect(fallbackOptions ? getNumberField(fallbackOptions, "num_ctx") : undefined).toBe(2048);
+	});
+
+	it("generate() prefers info.structured over text parts", async () => {
+		mockFetch(async (url) => {
+			if (url.includes("/session") && !url.includes("/message")) {
+				return Response.json({
+					id: "ses_structured",
+					slug: "test",
+					projectID: "p",
+					directory: "/tmp",
+					title: "test",
+					version: "1",
+				});
+			}
+			return Response.json({
+				info: {
+					role: "assistant",
+					id: "msg_s",
+					sessionID: "ses_structured",
+					cost: 0,
+					tokens: { input: 10, output: 5 },
+					structured: { facts: [{ content: "from structured", type: "fact", confidence: 0.9 }], entities: [] },
+				},
+				parts: [
+					{ type: "text", text: "ignore this text", id: "p1", sessionID: "ses_structured", messageID: "msg_s" },
+				],
+			});
+		});
+
+		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
+		const result = await provider.generate("test");
+		const parsed = JSON.parse(result);
+		expect(parsed.facts[0].content).toBe("from structured");
+	});
+
+	it("generate() returns info.structured as string when it is a string", async () => {
+		mockFetch(async (url) => {
+			if (url.includes("/session") && !url.includes("/message")) {
+				return Response.json({
+					id: "ses_str",
+					slug: "test",
+					projectID: "p",
+					directory: "/tmp",
+					title: "test",
+					version: "1",
+				});
+			}
+			return Response.json({
+				info: {
+					role: "assistant",
+					id: "msg_str",
+					sessionID: "ses_str",
+					cost: 0,
+					tokens: { input: 0, output: 0 },
+					structured: '{"description":"test skill","triggers":["run tests"],"tags":["testing"]}',
+				},
+				parts: [],
+			});
+		});
+
+		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
+		const result = await provider.generate("test");
+		expect(JSON.parse(result).description).toBe("test skill");
+	});
+
+	it("generate() disables structured output on 400 and retries without format", async () => {
+		let attempts = 0;
+		let lastBody: Record<string, unknown> = {};
+		mockFetch(async (url, init) => {
+			if (url.includes("/session") && !url.includes("/message")) {
+				return Response.json({
+					id: `ses_compat_${attempts}`,
+					slug: "test",
+					projectID: "p",
+					directory: "/tmp",
+					title: "test",
+					version: "1",
+				});
+			}
+			attempts++;
+			lastBody = JSON.parse(init?.body as string);
+			if (attempts === 1) {
+				return new Response('{"error":"validation failed for format field"}', { status: 400 });
+			}
+			return Response.json(openCodeResponse("fallback works"));
+		});
+
+		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
+		const result = await provider.generate("test");
+		expect(result).toBe("fallback works");
+		// After disabling, subsequent calls should not include format
+		expect(lastBody.format).toBeUndefined();
+	});
+
+	it("generate() parses github-copilot provider/model format", () => {
+		const provider = createOpenCodeProvider({ model: "github-copilot/gpt-4o" });
+		expect(provider.name).toBe("opencode:github-copilot/gpt-4o");
 	});
 });
 
--- a/packages/daemon/src/pipeline/provider.test.ts
+++ b/packages/daemon/src/pipeline/provider.test.ts
@@ -966,6 +966,25 @@ describe("createOpenCodeProvider", () => {
 		expect(lastBody.format).toBeUndefined();
 	});
 
+	it("generate() preserves error body on non-format 400", async () => {
+		mockFetch(async (url) => {
+			if (url.includes("/session") && !url.includes("/message")) {
+				return Response.json({
+					id: "ses_400",
+					slug: "test",
+					projectID: "p",
+					directory: "/tmp",
+					title: "test",
+					version: "1",
+				});
+			}
+			return new Response("bad request: missing required field", { status: 400 });
+		});
+
+		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
+		await expect(provider.generate("test")).rejects.toThrow(/bad request: missing required field/);
+	});
+
 	it("generate() parses github-copilot provider/model format", () => {
 		const provider = createOpenCodeProvider({ model: "github-copilot/gpt-4o" });
 		expect(provider.name).toBe("opencode:github-copilot/gpt-4o");
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -1915,15 +1915,10 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				// future body shape changes that happen to contain "format" elsewhere.
 				const isFormatRejection = (() => {
 					try {
-						const parsed = JSON.parse(consumedBody) as Record<string, unknown>;
+						const parsed: unknown = JSON.parse(consumedBody);
+						if (!isRecord(parsed)) return false;
 						const issues = Array.isArray(parsed.issues) ? parsed.issues : [];
-						return issues.some(
-							(i): boolean =>
-								typeof i === "object" &&
-								i !== null &&
-								Array.isArray((i as Record<string, unknown>).path) &&
-								((i as Record<string, unknown>).path as unknown[])[0] === "format",
-						);
+						return issues.some((i): boolean => isRecord(i) && Array.isArray(i.path) && i.path[0] === "format");
 					} catch {
 						return false;
 					}
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -1910,9 +1910,25 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 			// fall through to the throw path.
 			if (!res.ok && res.status === 422) {
 				consumedBody = await res.text().catch(() => "");
-				// Match Hono/Zod's path array notation: {"path":["format"],...}
-				// Avoids triggering on error messages that contain "format" as a value.
-				if (consumedBody.includes('["format"]')) {
+				// Parse the Hono/Zod rejection body to check whether "format" is the
+				// offending field. Unambiguous vs. substring matching and immune to
+				// future body shape changes that happen to contain "format" elsewhere.
+				const isFormatRejection = (() => {
+					try {
+						const parsed = JSON.parse(consumedBody) as Record<string, unknown>;
+						const issues = Array.isArray(parsed.issues) ? parsed.issues : [];
+						return issues.some(
+							(i): boolean =>
+								typeof i === "object" &&
+								i !== null &&
+								Array.isArray((i as Record<string, unknown>).path) &&
+								((i as Record<string, unknown>).path as unknown[])[0] === "format",
+						);
+					} catch {
+						return false;
+					}
+				})();
+				if (isFormatRejection) {
 					if (structuredOutputSupported) {
 						logger.info("pipeline", "OpenCode does not support structured output format, disabling", {
 							status: res.status,
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -1900,10 +1900,12 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 			let consumedBody: string | null = null;
 
 			// Older OpenCode versions may not support the format field.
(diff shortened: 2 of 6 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): support Copilot models via OpenCode structured output` (commit `a97f39b5`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,7 +1898,27 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				return pollForAssistantMessage(forSessionId);
```
PR-Reviewer-Ant, line 1899: **Body consumed before error handler — error detail silently lost for non-format 400/422 responses.**

The outer block reads `res.text()` unconditionally to check for format/schema/validation keywords:
```ts
const body = await res.text().catch(() => "");
if (body.includes("format") || ...) { ... }
// if condition is false, body stream is already consumed
```
When the keywords aren't present the condition exits without reassigning `res`, and the existing error path below does:
```ts
const body = await res.text().catch(() => "");   // returns "" — stream exhausted
throw new Error(`OpenCode HTTP ${res.status}: ${body.slice(0, 200)}`);
```
The thrown error becomes `OpenCode HTTP 400: ` with no detail, making non-format 400/422 failures invisible in logs and stack traces.

**Fix:** either store the body in a variable accessible to the outer scope, or only read the body when the keyword branch is actually entered:
```ts
let res = await postMessage(sid);
let consumedBody: string | null = null;
if (!res.ok && structuredOutputSupported && (res.status === 400 || res.status === 422)) {
    consumedBody = await res.text().catch(() => "");
    if (consumedBody.includes("format") || ...) { ... }
}
if (!res.ok) {
    const body = consumedBody ?? await res.text().catch(() => "");
    ...
}
```

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,7 +1898,27 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+				if (body.includes("format") || body.includes("schema") || body.includes("validation")) {
+					logger.info("pipeline", "OpenCode does not support structured output format, disabling", {
+						status: res.status,
+					});
+					structuredOutputSupported = false;
+					sessionId = null;
+					const retrySid = await getOrCreateSession();
+					res = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
+						method: "POST",
+						headers: { "Content-Type": "application/json" },
```
PR-Reviewer-Ant, line 1945: `buildMessageBody(prompt, true)` is called immediately after setting `structuredOutputSupported = false`. The call is correct because the guard inside `buildMessageBody` checks `structured && structuredOutputSupported`, so the format field is not emitted. But passing `true` here is misleading — a reader naturally expects that after disabling structured output the call would reflect that intent. Consider passing `false` or omitting the argument to make the intent explicit.

packages/daemon/src/pipeline/skill-enrichment.ts
```diff
@@ -109,7 +109,8 @@ export async function enrichSkillFrontmatter(
 		if (!result) {
 			logger.warn("pipeline", "Failed to parse skill enrichment output", {
 				skill: input.name,
-				preview: raw.slice(0, 200),
+				preview: raw.slice(0, 500),
+			length: raw.length,
```
PR-Reviewer-Ant: Indentation error on `length: raw.length` — it is not indented at the object-property level, breaking visual consistency with the surrounding object literal. This will likely trigger a lint warning.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a97f39b5`

[Automated Bot Review] PR #425 — one warning-level correctness issue found (response body consumed before error path), plus a misleading argument and a minor indentation error in a log call.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The body-consumption bug is directly traceable in the changed lines of provider.ts: the outer `if` block at the 400/422 path calls `res.text()` unconditionally before the keyword check, then the fall-through error handler calls `res.text()` again on the same (now-exhausted) stream. Both call sites are in the diff and the control flow is unambiguous without needing runtime traces.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `092a3528`

[Automated Bot Review] PR #425 — all three previously-open items appear resolved; one new warning-level issue found in the structured-output fallback heuristic.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The keyword-matching heuristic and the consumedBody flow are both fully visible in the provider.ts diff. The prior body-consumption bug fix is traceable through the consumedBody/null pattern. No runtime trace needed; the control flow is deterministic from the diff alone.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1896,32 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				return pollForAssistantMessage(forSessionId);
 			};
 
-			const res = await postMessage(sid);
+			let res = await postMessage(sid);
+			let consumedBody: string | null = null;
+
+			// Older OpenCode versions may not support the format field.
+			// If we get a validation error, disable structured output and retry.
```
PR-Reviewer-Ant: **Fragile keyword heuristic permanently disables structured output on false-positive matches.**

The check `consumedBody.includes("format") || consumedBody.includes("schema") || consumedBody.includes("validation")` will trigger on any 400/422 response body that contains these common English words — not just OpenCode structured-output rejection errors. Examples that would false-positive:

- `"Invalid request format: parts array is missing"` (unrelated formatting error)
- `"Schema validation failed for field 'parts'"` (a parts-level error, not a format-level one)
- `"Request validation error: model field required"` (model config error)

Because `structuredOutputSupported = false` is never reset within the provider lifetime, a single false-positive permanently degrades all subsequent extraction calls to free-text mode — silently, with no observable error surfaced to the caller. The test covers the happy path (`"validation failed for format field"`) and the clean non-match case (`"missing required field"`), but not ambiguous cases like the ones above.

Consider anchoring the check to a more specific pattern (e.g. `consumedBody.includes('"format"')` to check for a JSON key, or checking `res.status === 422` alone as the structured-output-unsupported signal since 400 is used for many unrelated errors).

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1896,32 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+			if (!res.ok && structuredOutputSupported && (res.status === 400 || res.status === 422)) {
+				consumedBody = await res.text().catch(() => "");
+				if (consumedBody.includes("format") || consumedBody.includes("schema") || consumedBody.includes("validation")) {
```
PR-Reviewer-Ant: **Concurrent call hazard with `sessionId = null` reset inside structured-output retry path.**

If two concurrent `generate()` calls share the same provider instance and both hit a 400/422 with format keywords, both will execute `structuredOutputSupported = false; sessionId = null` and both will call `getOrCreateSession()`, creating two orphaned sessions. The existing 404/410 retry logic has the same exposure, so this is not a new class of bug, but the new 400/422 path adds a second trigger. Since `structuredOutputSupported = false` is idempotent and sessions are cheap, this is a minor concern rather than a correctness failure — but worth noting if the provider is used in a high-concurrency pipeline.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -866,6 +873,122 @@ describe("createOpenCodeProvider", () => {
...
+					directory: "/tmp",
+					title: "test",
+					version: "1",
+				});
+			}
+			return new Response("bad request: missing required field", { status: 400 });
+		});
+
+		const provider = createOpenCodeProvider({ baseUrl: "http://localhost:9999" });
+		await expect(provider.generate("test")).rejects.toThrow(/bad request: missing required field/);
```
PR-Reviewer-Ant, line 1017: The comment `// After disabling, subsequent calls should not include format` is misleading. `lastBody` captures the body of the retry request within the *same* `generate()` call (attempt 2), not a second call to `generate()`. The described property — that a new top-level `generate()` invocation also omits `format` — is not actually tested here. Consider adding a second `await provider.generate(...)` call after the first and asserting its request body lacks `format`.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1896,34 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
 
-			const res = await postMessage(sid);
+			let res = await postMessage(sid);
+			let consumedBody: string | null = null;
+
+			// Older OpenCode versions may not support the format field.
+			// 422 is the Hono/Zod schema validation rejection for unknown fields.
+			// Only trigger on 422 to avoid false-positives from unrelated 400 errors
+			// (e.g. missing model, bad parts) that contain common words like "format".
+			if (!res.ok && structuredOutputSupported && res.status === 422) {
```
PR-Reviewer-Ant: **Concurrent call hazard (pre-existing, still unresolved):** Two concurrent `generate()` calls can both receive a 422 before either processes it. Call A processes first: sets `structuredOutputSupported = false`, resets `sessionId`, retries, and succeeds. Call B then checks `!res.ok && structuredOutputSupported && res.status === 422` — `structuredOutputSupported` is now `false`, so the condition is false, and Call B falls straight through to the `throw new Error("OpenCode HTTP 422: ...")` path. The result is that all concurrent callers except the first one to process the 422 will throw rather than being silently retried without structured output. Narrowing to 422-only (this diff) makes the trigger rarer, but when it does fire under concurrency the failure is more abrupt than before (throws instead of retrying). A simple fix: check the response status and body before reading `structuredOutputSupported`, or serialize the disable path with a flag-in-flight guard.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fac645e6`

[Automated Bot Review] PR #425 (incremental, sha fac645e) — the two remaining open items from the previous round are substantially addressed; one pre-existing warning-level race condition remains unresolved.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The 422-only gate and the '"format"' quoted-key check are both directly readable in the provider.ts diff. The concurrent hazard is traceable through the shared `structuredOutputSupported` flag and `sessionId` mutation path in the same function. No runtime trace is needed; the control flow is deterministic from the diff.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1896,38 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				return pollForAssistantMessage(forSessionId);
 			};
 
-			const res = await postMessage(sid);
+			let res = await postMessage(sid);
+			let consumedBody: string | null = null;
+
```
PR-Reviewer-Ant, line 1904: **Pre-existing concurrent session hazard — not worsened but still unresolved.** When two concurrent `generate()` calls both hit 422 with `'"format"'` in the body, both enter the `if (consumedBody.includes('"format"'))` branch. Call A sets `sessionId = null` and creates session S2 (setting `sessionId = S2`). Call B then calls `getOrCreateSession()`, finds `sessionId === S2`, and reuses it — so both retries POST to the same OpenCode session concurrently. OpenCode's session model does not guarantee message ordering across concurrent POSTs to the same session ID, which could cause either call to receive the other's response. The fix that was discussed (per-call session IDs, or a session-creation mutex) is still not present. This is a warning rather than a blocker because the scenario requires both rate-limiting AND concurrent extraction workers, and the worst outcome is one call getting an unrelated response (which the extraction parser would reject and return empty rather than corrupt data).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `6d51c7eb`

[Automated Bot Review — pr-reviewer] PR #425 incremental review (sha 6d51c7eb). The one substantive change in this diff — removing `structuredOutputSupported &&` from the 422 guard — correctly fixes the concurrent-caller regression flagged in the prior round (B would have skipped the retry path once A set the flag). Most previously-open items are now resolved. One pre-existing concurrent session hazard remains; it is not worsened by this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single logical change (removing the `structuredOutputSupported &&` guard from the outer 422 block). The downstream effects — concurrent B now enters the block regardless of flag state, `sessionId` mutation still shared — are traceable entirely within provider.ts lines 1896–1935. No runtime trace is needed; the control flow is deterministic from the diff.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1899,41 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+			// Older OpenCode versions may not support the format field.
+			// 422 is the Hono/Zod schema validation rejection for unknown fields.
+			// Check status/body before reading structuredOutputSupported so that
+			// concurrent callers both hitting 422 both retry correctly — if A sets
+			// structuredOutputSupported=false first, B must still retry rather than
+			// fall through to the throw path.
+			if (!res.ok && res.status === 422) {
+				consumedBody = await res.text().catch(() => "");
+				if (consumedBody.includes('"format"')) {
```
PR-Reviewer-Ant: **Residual: `'"format"'` keyword heuristic — still a string match, albeit tighter.** The check `consumedBody.includes('"format"')` now requires the quoted form, which is a meaningful improvement over the previous `includes('format')`. However, any 422 body whose JSON happens to contain the literal four characters `"format"` as a quoted string value (not key) — e.g. an error message like `{"detail":"unknown field \"format\""}` from a non-Hono server — would still trigger a permanent `structuredOutputSupported = false` for the lifetime of the provider instance. The risk is low in practice (OpenCode / Hono is the only server targeted here) but permanent disablement on a false positive is hard to recover from without restarting the daemon. A more defensive option would be to check for a path-qualified key pattern such as `"path":["format"]` or `"path":"format"`, which is what the Hono/Zod response actually emits and what the test asserts. Not blocking, but worth hardening if this provider is expected to work with other OpenCode-compatible servers.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fd072af7`

[Automated Bot Review — pr-reviewer] PR #425 incremental review (sha fd072af). The refactor of `getOrCreateSession` into `createSession` + `getOrCreateSession` correctly resolves the concurrent-null-reset race that was the main open blocker from prior rounds. Nearly all previously flagged items are now addressed. One warning-level concern on the keyword heuristic remains.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is narrow (refactoring getOrCreateSession, changing sessionId assignment in the 422 path). The concurrent hazard fix is directly traceable: both callers now receive their own local `retrySid` from `createSession()` and use it for the retry, so the shared `sessionId` last-write-wins behavior is deterministic and never produces a null/invalid state. The remaining `'"format"'` heuristic concern is verifiable from the literal string check in provider.ts and the two test cases that exercise it.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1899,43 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+
+			// Older OpenCode versions may not support the format field.
+			// 422 is the Hono/Zod schema validation rejection for unknown fields.
+			// Check status/body before reading structuredOutputSupported so that
+			// concurrent callers both hitting 422 both retry correctly — if A sets
+			// structuredOutputSupported=false first, B must still retry rather than
+			// fall through to the throw path.
+			if (!res.ok && res.status === 422) {
+				consumedBody = await res.text().catch(() => "");
+				// Match Hono/Zod's path array notation: {"path":["format"],...}
```
PR-Reviewer-Ant: The `'["format"]'` substring check is a meaningful improvement over the prior `'"format"'` — it now specifically targets Hono/Zod's `{"path":["format"],...}` schema-rejection shape. The remaining risk is that a future OpenCode version could emit a structurally unrelated 422 whose body happens to contain the literal `["format"]` (e.g. a validation error for a nested field path that includes `"format"` as an element). If that occurs, structured output would be permanently disabled for the lifetime of the provider instance with no way to recover short of restarting the daemon.

A JSON parse of the body (`JSON.parse(consumedBody).issues?.some(i => Array.isArray(i.path) && i.path[0] === 'format')`) would be unambiguous and equally cheap. That said, the current check is pragmatically tight and covers all known OpenCode versions, so this is a warning rather than a blocker.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `993a7740`

[Automated Bot Review — pr-reviewer] PR #425 incremental review (sha 993a774). The single change in this diff — tightening the 422 heuristic from `'"format"'` to `'["format"]'` — is a genuine improvement that substantially narrows the false-positive surface. Nearly all previously-flagged items are now addressed across this PR's series of commits. One warning-level concern on the remaining string heuristic is noted; no new issues introduced.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single-line logical change. The heuristic effect is directly traceable: `consumedBody.includes('["format"]')` will only match Hono/Zod's literal path array notation, whereas the prior `'"format"'` would have matched any error body containing the word format as a quoted string value. The remaining risk — a body like `{"path":["format"],"message":"..."}`  appearing in an unrelated 422 — is verifiable as astronomically unlikely from the OpenCode API surface, but cannot be proven zero without a full OpenCode API schema.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1899,59 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+			// Check status/body before reading structuredOutputSupported so that
+			// concurrent callers both hitting 422 both retry correctly — if A sets
+			// structuredOutputSupported=false first, B must still retry rather than
+			// fall through to the throw path.
+			if (!res.ok && res.status === 422) {
+				consumedBody = await res.text().catch(() => "");
+				// Parse the Hono/Zod rejection body to check whether "format" is the
+				// offending field. Unambiguous vs. substring matching and immune to
+				// future body shape changes that happen to contain "format" elsewhere.
+				const isFormatRejection = (() => {
```
PR-Reviewer-Ant, line 1916: Convention violation: `as Record<string, unknown>` and `(i as Record<string, unknown>)` are used inside `isFormatRejection`. The codebase convention is "no `as` — fix the types instead of asserting." The `isRecord` helper is already defined in this file and handles this pattern cleanly:

```ts
const parsed: unknown = JSON.parse(consumedBody);
if (!isRecord(parsed)) return false;
const issues = Array.isArray(parsed.issues) ? parsed.issues : [];
return issues.some(
  (i): boolean =>
    isRecord(i) &&
    Array.isArray(i.path) &&
    i.path[0] === "format",
);
```

This is not a correctness issue — the logic is right — but it introduces two `as` casts in a file that already has `isRecord` available for exactly this purpose.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1899,59 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+								Array.isArray((i as Record<string, unknown>).path) &&
+								((i as Record<string, unknown>).path as unknown[])[0] === "format",
+						);
+					} catch {
+						return false;
+					}
+				})();
+				if (isFormatRejection) {
+					if (structuredOutputSupported) {
+						logger.info("pipeline", "OpenCode does not support structured output format, disabling", {
```
PR-Reviewer-Ant, line 1928: Pre-existing shared-state write `sessionId = retrySid` in the 422 retry path: if two concurrent `generate()` calls both enter this branch, both call `createSession()` independently (correct — each gets its own local `retrySid` for the retry fetch) but both also overwrite `sessionId`. Last writer wins, causing one session to be abandoned on the server. The retry calls themselves are safe since each uses its local variable, and both sessions remain valid, so this is a resource-waste rather than a correctness issue. Noted in previous rounds as pre-existing and not worsened by this diff.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0cffc42e`

[Automated Bot Review — pr-reviewer] PR #425 final-commit review (sha 0cffc42e). The incremental change — replacing `consumedBody.includes('["format"]')` with proper JSON-parse structural detection — fully resolves the last open heuristic concern from prior rounds. All previously-open blocking items are now addressed. Two minor issues remain.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single logical change (13 lines replacing a substring check with a structural JSON parse). Correctness of the detection logic is directly verifiable from the diff: `JSON.parse(consumedBody)`, check for `parsed.issues` array, then `path[0] === 'format'`. Both issues flagged below are traceable to specific lines in the changed code without runtime context.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1867,10 +1899,54 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+						if (!isRecord(parsed)) return false;
+						const issues = Array.isArray(parsed.issues) ? parsed.issues : [];
+						return issues.some((i): boolean => isRecord(i) && Array.isArray(i.path) && i.path[0] === "format");
+					} catch {
+						return false;
+					}
+				})();
+				if (isFormatRejection) {
+					if (structuredOutputSupported) {
+						logger.info("pipeline", "OpenCode does not support structured output format, disabling", {
```
PR-Reviewer-Ant, line 1928: Pre-existing shared-state write: `sessionId = retrySid` after the 422 retry path. When two concurrent `generate()` calls both receive 422 and both call `createSession()` independently, each gets its own `retrySid` and uses it for the retry fetch — so both retries are safe. The race is benign: last writer wins on `sessionId`, but subsequent calls will simply reuse whichever session survived. No data corruption is possible because both callers already hold their respective `retrySid` local before the write. Noting for visibility only — this is not a regression introduced by this PR.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ff3c19f6`

[Automated Bot Review — pr-reviewer] PR #425 final-commit review (sha ff3c19f). The incremental change — removing `as Record<string, unknown>` casts inside `isFormatRejection` in favor of the existing `isRecord()` guard — directly resolves the last open convention violation flagged in prior rounds. No new issues introduced. All previously-blocking items are now addressed across this PR's commit series.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is 10 lines and affects only the `isFormatRejection` IIFE inside `sendMessage`. Correctness is directly verifiable: `JSON.parse(consumedBody)` now types the result as `unknown`, followed by an `isRecord(parsed)` guard before accessing `.issues`, and `isRecord(i)` before accessing `i.path`. This matches the codebase convention ('no `as` — use `unknown` with narrowing'). The one remaining comment-level concern — `sessionId = retrySid` shared-state write — is pre-existing and traceable to specific lines in the full file content; its safety is verifiable without runtime context.

PR #430, opened 2026-04-01
fix(pipeline): pass maxContextTokens to direct Ollama synthesis (#426)

### LINKED ISSUE #426: Chunked session summarization fails with direct Ollama synthesis because num_ctx is not set on the direct-ollama path
**Signet version:** 0.91.1
**Environment:** Windows, Node v25.5.0, Bun 1.3.11
**Related issues:** N/A

---

## Summary

Chunked session summarization fails consistently when memory.pipelineV2.synthesis.provider resolves to direct ollama for long transcripts. On long jobs, every chunk is skipped even though the same model can summarize successfully when given a larger explicit context window. This matters because session summaries and downstream memory synthesis degrade or fail for larger coding sessions.

---

## Observed Behavior

Long transcript jobs enter chunked summarization, then each chunk logs Chunk summarization failed, skipping. The affected jobs retry and eventually become dead or remain processing.

```
{"timestamp":"2026-03-31T22:22:13.061Z","level":"info","category":"summary-worker","message":"Processing session summary","data":{"jobId":"9c09b353-22a2-40cc-9446-a162fc08cf39","harness":"claude-code","attempt":1,"sessionKey":"ab400917-d215-4c14-a6fa-6e778c5513fd","project":"REDACTED"}}
{"timestamp":"2026-03-31T22:22:23.826Z","level":"warn","category":"summary-worker","message":"Chunk summarization failed, skipping","data":{"chunk":1,"total":5}}
{"timestamp":"2026-03-31T22:23:38.274Z","level":"warn","category":"summary-worker","message":"Chunk summarization failed, skipping","data":{"chunk":2,"total":5}}
{"timestamp":"2026-03-31T22:24:16.001Z","level":"warn","category":"summary-worker","message":"Chunk summarization failed, skipping","data":{"chunk":3,"total":5}}
{"timestamp":"2026-03-31T22:24:45.205Z","level":"warn","category":"summary-worker","message":"Chunk summarization failed, skipping","data":{"chunk":4,"total":5}}
{"timestamp":"2026-03-31T22:25:16.480Z","level":"warn","category":"summary-worker","message":"Chunk summarization failed, skipping","data":{"chunk":5,"total":5}}
```

Config logs show synthesis is using direct Ollama:

```
{"timestamp":"2026-03-31T01:56:46.111Z","level":"info","category":"config","message":"Synthesis provider","data":{"configured":null,"resolved":"ollama","effective":"ollama","endpoint":"http://127.0.0.1:11434"}}
{"timestamp":"2026-03-31T01:56:46.112Z","level":"info","category":"synthesis","message":"Synthesis worker started","data":{"provider":"ollama","model":"qwen3:4b-instruct-2507-q4_K_M","idleGapMinutes":15}}
```

Current summary_jobs state for the failing transcript:

```json
[
  {
    "id": "fe4fd610-4f0d-4867-9070-591db7a88d47",
    "status": "processing",
    "attempts": 3,
    "max_attempts": 3,
    "len": 100012
  },
  {
    "id": "9c09b353-22a2-40cc-9446-a162fc08cf39",
    "status": "dead",
    "attempts": 3,
    "max_attempts": 3,
    "len": 100012
  },
  {
    "id": "e251902f-c1fb-4ed7-92e9-d158918a0574",
    "status": "completed",
    "attempts": 1,
    "max_attempts": 3,
    "len": 49008
  }
]
```

Manual replay of the first failing chunk against the same Ollama model without explicit num_ctx returned non-JSON and showed a 4096 token prompt budget:

```json
{
  "ok": true,
  "responsePreview": "Got it. Let's build this step by step — **first file: `WizardInput.cs`** ...",
  "thinkingPreview": "",
  "doneReason": "stop",
  "promptEvalCount": 4096,
  "evalCount": 685
}
```

Manual replay of the exact same chunk with options.num_ctx = 8192 returned valid JSON:

```json
{
  "ok": true,
  "responsePreview": "{\n  \"summary\": \"The session focused on refining the architectural scope of Ticket 15: ItemCollectionEvaluator ...",
  "promptEvalCount": 5089,
  "evalCount": 966
}
```
---

## Expected Behavior

Long transcript chunk summarization should succeed under direct Ollama synthesis, or at minimum adapt chunk size/context settings so that the model retains the prompt instructions and returns parseable JSON. The summary worker should not silently skip every chunk for a transcript that fits within the configured chunking strategy.

---

## Steps to Reproduce

1. Configure Signet so memory.pipelineV2.synthesis.provider resolves to direct ollama.
2. Use an Ollama model such as qwen3:4b with default runtime context behavior.
3. Create a long session transcript around 100k characters so Signet enters chunked summarization.
4. Let the summary worker process the job.
5. Observe that each chunk logs Chunk summarization failed, skipping and the job eventually retries/fails.

---

## Root Cause Analysis

The direct-ollama synthesis path appears not to pass an explicit num_ctx to Ollama, while chunk sizing assumes a larger usable context budget. For large chunk prompts, Ollama appears to fall back to a 4096 prompt context, which truncates the front of the prompt. That likely removes the Return ONLY a JSON object instruction, so the model responds with generic assistant prose and `parseLlmResponse()` returns `null`.

Relevant code behavior observed in the installed package:

- `resolveSummaryProvider()` computes `ollamaFallbackMaxContextTokens`
- fallback Ollama creation includes `maxContextTokens`
- direct provider: ollama synthesis creation appears to omit `maxContextTokens` unless using the fallback branch

This would explain why:

- smaller jobs succeed
- larger chunked jobs fail
- manual replay succeeds once num_ctx is explicitly set

---

## Suggested Fix

```ts
// Ensure direct-ollama synthesis also passes maxContextTokens
return createOllamaProvider({
  ...effectiveSynthesisModel ? { model: effectiveSynthesisModel } : {},
  baseUrl: synthesisOllamaFallbackBaseUrl,
  defaultTimeoutMs: memoryCfg.pipelineV2.synthesis.timeout,
  maxContextTokens: ollamaFallbackMaxContextTokens
});
```
Also helpful:

- log parse failures with a short redacted response preview
- log prompt_eval_count when available
- optionally adapt chunk size based on provider context budget

---

## Impact

Large session summaries can fail entirely under direct Ollama synthesis. That degrades:

- session summary generation
- downstream fact extraction from summaries
- memory synthesis quality
- reliability of long-session capture in local-only setups using Ollama

This primarily affects users running Signet with local Ollama synthesis on longer coding sessions.

---

## Workaround

Create an Ollama model alias with a larger default num_ctx and point Signet synthesis to that alias. For example, a local alias based on the same model with PARAMETER num_ctx 8192 avoids the failure in testing.

---

## Environment Details

Signet: 0.91.1
OS: Windows
Node: v25.5.0
Bun: 1.3.11
DB: SQLite
Date observed: 2026-03-31

NicholaiVogel:
## Summary

- Pass `maxContextTokens` (default 8192) to the direct Ollama provider in `resolveSummaryProvider()`, fixing chunked session summarization failures when Ollama defaults to 4096 context tokens
- Add `responsePreview` to chunk failure warning logs for easier diagnosis of future parse failures
- Add regression test verifying direct Ollama path includes `num_ctx` in API requests

Closes #426

## Test plan

- [x] Existing summary-worker tests pass (20/20)
- [x] New regression test verifies `num_ctx` is sent for direct Ollama path
- [x] Typecheck passes
- [ ] Manual: configure direct Ollama synthesis, run a long session (~100k chars), verify chunked summarization succeeds

### DIFF
--- a/packages/daemon-rs/crates/signet-pipeline/src/provider.rs
+++ b/packages/daemon-rs/crates/signet-pipeline/src/provider.rs
@@ -554,6 +554,11 @@ const DEFAULT_TIMEOUT_MS: u64 = 90_000;
 pub const DEFAULT_OLLAMA_MAX_CONTEXT_TOKENS: u32 = 8192;
 
 /// Resolve the Ollama context window size from env or default.
+///
+/// Reads `SIGNET_OLLAMA_FALLBACK_MAX_CTX` (kept for backwards compatibility).
+/// Despite the `FALLBACK` label, this value applies to **all** Ollama summary
+/// paths — both the degraded-fallback case and an explicitly-configured
+/// `synthesis.provider = ollama` deployment.
 pub fn resolve_ollama_max_context_tokens() -> u32 {
     std::env::var("SIGNET_OLLAMA_FALLBACK_MAX_CTX")
         .ok()
--- a/packages/daemon-rs/crates/signet-pipeline/src/provider.rs
+++ b/packages/daemon-rs/crates/signet-pipeline/src/provider.rs
@@ -115,6 +115,7 @@ pub struct OllamaLlmProvider {
     base_url: String,
     model: String,
     default_timeout_ms: u64,
+    max_context_tokens: Option<u32>,
     health: Mutex<HealthTracker>,
 }
 
@@ -131,6 +132,8 @@ struct OllamaGenRequest<'a> {
 struct OllamaGenOptions {
     #[serde(skip_serializing_if = "Option::is_none")]
     num_predict: Option<u32>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    num_ctx: Option<u32>,
 }
 
 #[derive(Deserialize)]
@@ -143,6 +146,10 @@ struct OllamaGenResponse {
 
 impl OllamaLlmProvider {
     pub fn new(base_url: &str, model: &str, timeout_ms: u64) -> Self {
+        Self::with_context(base_url, model, timeout_ms, None)
+    }
+
+    pub fn with_context(base_url: &str, model: &str, timeout_ms: u64, max_context_tokens: Option<u32>) -> Self {
         let client = reqwest::Client::builder()
             .timeout(Duration::from_millis(timeout_ms.max(5000)))
             .build()
@@ -153,6 +160,7 @@ impl OllamaLlmProvider {
             base_url: base_url.trim_end_matches('/').to_string(),
             model: model.to_string(),
             default_timeout_ms: timeout_ms,
+            max_context_tokens,
             health: Mutex::new(HealthTracker::new()),
         }
     }
@@ -169,9 +177,14 @@ impl OllamaLlmProvider {
         let start = Instant::now();
         let timeout = Duration::from_millis(opts.timeout_ms.unwrap_or(self.default_timeout_ms));
 
-        let options = opts.max_tokens.map(|n| OllamaGenOptions {
-            num_predict: Some(n),
-        });
+        let options = if opts.max_tokens.is_some() || self.max_context_tokens.is_some() {
+            Some(OllamaGenOptions {
+                num_predict: opts.max_tokens,
+                num_ctx: self.max_context_tokens,
+            })
+        } else {
+            None
+        };
 
         let body = OllamaGenRequest {
             model: &self.model,
@@ -538,6 +551,16 @@ impl HealthTracker {
 
 const DEFAULT_OLLAMA_URL: &str = "http://localhost:11434";
 const DEFAULT_TIMEOUT_MS: u64 = 90_000;
+pub const DEFAULT_OLLAMA_MAX_CONTEXT_TOKENS: u32 = 8192;
+
+/// Resolve the Ollama context window size from env or default.
+pub fn resolve_ollama_max_context_tokens() -> u32 {
+    std::env::var("SIGNET_OLLAMA_FALLBACK_MAX_CTX")
+        .ok()
+        .and_then(|s| s.trim().parse::<u32>().ok())
+        .filter(|&n| n > 0)
+        .unwrap_or(DEFAULT_OLLAMA_MAX_CONTEXT_TOKENS)
+}
 
 /// LLM provider configuration.
 #[derive(Debug, Clone)]
@@ -547,6 +570,7 @@ pub struct LlmProviderConfig {
     pub base_url: Option<String>,
     pub api_key: Option<String>,
     pub timeout_ms: Option<u64>,
+    pub max_context_tokens: Option<u32>,
 }
 
 /// Create an LLM provider from config.
@@ -561,7 +585,7 @@ pub fn from_config(cfg: &LlmProviderConfig) -> Arc<dyn LlmProvider> {
                 .filter(|s| !s.is_empty())
                 .unwrap_or(DEFAULT_OLLAMA_URL);
             info!(provider = "ollama", model = %cfg.model, url, timeout_ms = timeout, "LLM provider initialized");
-            Arc::new(OllamaLlmProvider::new(url, &cfg.model, timeout))
+            Arc::new(OllamaLlmProvider::with_context(url, &cfg.model, timeout, cfg.max_context_tokens))
         }
         "anthropic" => {
             let key = cfg.api_key.as_deref().unwrap_or("");
@@ -578,7 +602,7 @@ pub fn from_config(cfg: &LlmProviderConfig) -> Arc<dyn LlmProvider> {
                 .as_deref()
                 .filter(|s| !s.is_empty())
                 .unwrap_or(DEFAULT_OLLAMA_URL);
-            Arc::new(OllamaLlmProvider::new(url, &cfg.model, timeout))
+            Arc::new(OllamaLlmProvider::with_context(url, &cfg.model, timeout, cfg.max_context_tokens))
         }
     }
 }
--- a/packages/daemon/src/pipeline/summary-worker.test.ts
+++ b/packages/daemon/src/pipeline/summary-worker.test.ts
@@ -542,11 +542,12 @@ describe("resolveSummaryProvider", () => {
 		// Intercept fetch to verify num_ctx is included in the request
 		const original = globalThis.fetch;
 		let captured: unknown = null;
-		globalThis.fetch = (async (_url: string | URL | Request, init?: RequestInit) => {
+		const mockFetch: typeof fetch = async (_url, init) => {
 			if (typeof init?.body !== "string") throw new Error(`Expected string body, got: ${typeof init?.body}`);
 			captured = JSON.parse(init.body);
 			return new Response(JSON.stringify({ response: "{}", done: true }), { status: 200 });
-		}) as typeof fetch;
+		};
+		globalThis.fetch = mockFetch;
 
 		try {
 			await provider.generate("test prompt");
--- a/packages/daemon/src/pipeline/summary-worker.test.ts
+++ b/packages/daemon/src/pipeline/summary-worker.test.ts
@@ -556,11 +556,12 @@ describe("resolveSummaryProvider", () => {
 
 		expect(captured).not.toBeNull();
 		if (typeof captured !== "object" || captured === null) throw new Error("Expected object body");
-		const body = captured as Record<string, unknown>;
-		const options = body.options;
+		if (!("options" in captured)) throw new Error("Expected options field");
+		const { options } = captured;
 		if (typeof options !== "object" || options === null) throw new Error("Expected options object");
-		const opts = options as Record<string, unknown>;
-		expect(typeof opts.num_ctx).toBe("number");
-		expect(opts.num_ctx).toBeGreaterThanOrEqual(8192);
+		if (!("num_ctx" in options)) throw new Error("Expected num_ctx field");
+		const { num_ctx: numCtx } = options;
+		expect(typeof numCtx).toBe("number");
+		expect(numCtx).toBeGreaterThanOrEqual(8192);
 	});
 });
--- a/packages/daemon/src/pipeline/summary-worker.test.ts
+++ b/packages/daemon/src/pipeline/summary-worker.test.ts
@@ -541,9 +541,10 @@ describe("resolveSummaryProvider", () => {
 
 		// Intercept fetch to verify num_ctx is included in the request
 		const original = globalThis.fetch;
-		let captured: Record<string, unknown> | null = null;
+		let captured: unknown = null;
 		globalThis.fetch = (async (_url: string | URL | Request, init?: RequestInit) => {
-			captured = JSON.parse(init?.body as string) as Record<string, unknown>;
+			if (typeof init?.body !== "string") throw new Error(`Expected string body, got: ${typeof init?.body}`);
+			captured = JSON.parse(init.body);
 			return new Response(JSON.stringify({ response: "{}", done: true }), { status: 200 });
 		}) as typeof fetch;
 
@@ -554,9 +555,12 @@ describe("resolveSummaryProvider", () => {
 		}
 
 		expect(captured).not.toBeNull();
-		const options = captured?.options as Record<string, unknown> | undefined;
-		expect(options).toBeDefined();
-		expect(typeof options?.num_ctx).toBe("number");
-		expect((options?.num_ctx as number) > 0).toBe(true);
+		if (typeof captured !== "object" || captured === null) throw new Error("Expected object body");
+		const body = captured as Record<string, unknown>;
+		const options = body.options;
+		if (typeof options !== "object" || options === null) throw new Error("Expected options object");
+		const opts = options as Record<string, unknown>;
+		expect(typeof opts.num_ctx).toBe("number");
+		expect(opts.num_ctx).toBeGreaterThanOrEqual(8192);
 	});
 });
--- a/packages/daemon/src/pipeline/summary-worker.test.ts
+++ b/packages/daemon/src/pipeline/summary-worker.test.ts
@@ -528,4 +528,35 @@ describe("resolveSummaryProvider", () => {
 		const provider = await resolveSummaryProvider(loadMemoryConfig(dir));
 		expect(provider.name).toBe("ollama:qwen3.5:4b");
 	});
+
+	it("direct ollama provider sends num_ctx so chunks are not truncated (#426)", async () => {
+		const dir = makeAgentsDir(`memory:
+  pipelineV2:
+    extractionProvider: ollama
+    extractionModel: qwen3:4b
+`);
+
+		const provider = await resolveSummaryProvider(loadMemoryConfig(dir));
+		expect(provider.name).toBe("ollama:qwen3:4b");
+
+		// Intercept fetch to verify num_ctx is included in the request
+		const original = globalThis.fetch;
+		let captured: Record<string, unknown> | null = null;
+		globalThis.fetch = (async (_url: string | URL | Request, init?: RequestInit) => {
+			captured = JSON.parse(init?.body as string) as Record<string, unknown>;
+			return new Response(JSON.stringify({ response: "{}", done: true }), { status: 200 });
+		}) as typeof fetch;
+
+		try {
+			await provider.generate("test prompt");
+		} finally {
+			globalThis.fetch = original;
+		}
+
+		expect(captured).not.toBeNull();
+		const options = captured?.options as Record<string, unknown> | undefined;
+		expect(options).toBeDefined();
+		expect(typeof options?.num_ctx).toBe("number");
+		expect((options?.num_ctx as number) > 0).toBe(true);
+	});
 });
--- a/packages/daemon/src/pipeline/summary-worker.ts
+++ b/packages/daemon/src/pipeline/summary-worker.ts
@@ -1338,11 +1338,11 @@ export async function resolveSummaryProvider(cfg: ReturnType<typeof loadMemoryCo
 	const model = cfg.pipelineV2.synthesis.model;
 	const timeout = cfg.pipelineV2.synthesis.timeout;
 	const endpoint = cfg.pipelineV2.synthesis.endpoint;
-	const ollamaFallbackMaxContextTokens = resolveDefaultOllamaFallbackMaxContextTokens();
+	const ollamaMaxContextTokens = resolveDefaultOllamaFallbackMaxContextTokens();
 	const fallback = () =>
 		createOllamaProvider({
 			defaultTimeoutMs: timeout,
-			maxContextTokens: ollamaFallbackMaxContextTokens,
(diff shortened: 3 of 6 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): pass maxContextTokens to direct Ollama synthesis (#426)` (commit `85cfcecd`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1410,12 +1411,11 @@ export async function resolveSummaryProvider(cfg: ReturnType<typeof loadMemoryCo
...
 		default:
-			// Intentionally omit maxContextTokens here. When Ollama is explicitly
-			// configured (not fallback), users control context via model config.
 			return createOllamaProvider({
 				...(typeof model === "string" && model.trim().length > 0 ? { model } : {}),
 				...(endpoint ? { baseUrl: endpoint } : {}),
 				defaultTimeoutMs: timeout,
+				maxContextTokens: ollamaFallbackMaxContextTokens,
 			});
 	}
```
PR-Reviewer-Ant, line 1420: **Rust parity rule not satisfied.** AGENTS.md states: 'Any behavioral change made to @signet/daemon must also be reflected in packages/daemon-rs/.' This commit changes the observable behavior of `resolveSummaryProvider()` — the direct Ollama path now sends `num_ctx` where it previously did not. The daemon-rs shadow rewrite must receive the equivalent change. No daemon-rs file appears in this diff.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1410,12 +1411,11 @@ export async function resolveSummaryProvider(cfg: ReturnType<typeof loadMemoryCo
 				defaultTimeoutMs: timeout,
```
PR-Reviewer-Ant, line 1411: The variable `ollamaFallbackMaxContextTokens` was named to describe the fallback path and is now reused verbatim for the explicit/direct Ollama path. The name now actively misleads: a reader of the `default` branch sees 'fallback' in a non-fallback context. Consider renaming to `ollamaMaxContextTokens` (or similar) at the declaration site to prevent future confusion about which paths it applies to.

packages/daemon/src/pipeline/summary-worker.test.ts
```diff
@@ -528,4 +528,35 @@ describe("resolveSummaryProvider", () => {
...
+		expect(provider.name).toBe("ollama:qwen3:4b");
+
+		// Intercept fetch to verify num_ctx is included in the request
+		const original = globalThis.fetch;
+		let captured: Record<string, unknown> | null = null;
+		globalThis.fetch = (async (_url: string | URL | Request, init?: RequestInit) => {
+			captured = JSON.parse(init?.body as string) as Record<string, unknown>;
+			return new Response(JSON.stringify({ response: "{}", done: true }), { status: 200 });
+		}) as typeof fetch;
+
```
PR-Reviewer-Ant, line 551: The test uses `as string`, `as Record<string, unknown>`, and `as number` — all of which violate the repo's explicit 'no `as`' convention. The `init?.body` path in particular: if the provider changes its request encoding (e.g., to a `ReadableStream`), the `as string` cast will silently produce `[object ReadableStream]` and `JSON.parse` will throw, making the failure mode confusing. Use a type guard (`typeof init?.body === 'string'`) and fail the test explicitly if the body is not a plain string.

packages/daemon/src/pipeline/summary-worker.test.ts
```diff
@@ -528,4 +528,35 @@ describe("resolveSummaryProvider", () => {
...
+		} finally {
+			globalThis.fetch = original;
+		}
+
+		expect(captured).not.toBeNull();
+		const options = captured?.options as Record<string, unknown> | undefined;
+		expect(options).toBeDefined();
+		expect(typeof options?.num_ctx).toBe("number");
+		expect((options?.num_ctx as number) > 0).toBe(true);
+	});
```
PR-Reviewer-Ant, line 567: The assertion `expect((options?.num_ctx as number) > 0).toBe(true)` checks only that `num_ctx` is a positive number. It does not verify the specific value that `resolveDefaultOllamaFallbackMaxContextTokens()` returns (described as 8192 in the PR). A tighter assertion — e.g., `expect(options?.num_ctx).toBeGreaterThanOrEqual(8192)` — would catch a future regression where the function returns 1 or some placeholder value.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `85cfcecd`

[Automated Bot Review] One blocking issue (Rust parity rule not satisfied), one warning on naming drift, and minor test convention violations.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The behavioral change in summary-worker.ts is clear from the diff. The blocking Rust parity concern is grounded in the explicit AGENTS.md rule ('Any behavioral change made to @signet/daemon must also be reflected in packages/daemon-rs/') and the absence of any daemon-rs changes in the diff. I cannot confirm whether daemon-rs has an equivalent resolveSummaryProvider that also needs patching, but the rule is unconditional and no daemon-rs file appears in the diff.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `63d3b450`

[Automated Bot Review] Prior blocking Rust parity issue is now resolved. Two residual `as` casts remain in the test file after previous review feedback.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is complete and unambiguous. The Rust summary worker at main.rs:948 now passes `max_context_tokens: Some(resolve_ollama_max_context_tokens())`, satisfying the parity rule that was blocking the prior review. The remaining issues are directly visible in the test diff at lines 558 and 561.

packages/daemon/src/pipeline/summary-worker.test.ts
```diff
@@ -528,4 +528,39 @@ describe("resolveSummaryProvider", () => {
...
+		}) as typeof fetch;
+
+		try {
+			await provider.generate("test prompt");
+		} finally {
+			globalThis.fetch = original;
+		}
+
+		expect(captured).not.toBeNull();
+		if (typeof captured !== "object" || captured === null) throw new Error("Expected object body");
```
PR-Reviewer-Ant, line 559: Two `as` casts remain after the explicit type-narrowing guards: `const body = captured as Record<string, unknown>` and `const opts = options as Record<string, unknown>` (line 561). The repo convention is 'no `as` — fix the types instead of asserting'. After `typeof x !== "object" || x === null` guards, TypeScript narrows to `object & {}` but not `Record<string, unknown>`. The idiomatic fix without `as` is to index through `Object.prototype.hasOwnProperty` narrowing or restructure the check to assign directly from `JSON.parse` with an explicit type parameter: `const body = JSON.parse(init.body) as never` is still `as`, but `JSON.parse(init.body) satisfies Record<string, unknown>` (TS 4.9+) or a type predicate helper avoids it entirely. Given the prior review already flagged these and partial progress was made (the `init?.body as string` and `options?.num_ctx as number` casts were cleaned up), the remaining two should be addressed for full compliance.

packages/daemon-rs/crates/signet-pipeline/src/provider.rs
```diff
@@ -538,6 +551,16 @@ impl HealthTracker {
 
 const DEFAULT_OLLAMA_URL: &str = "http://localhost:11434";
 const DEFAULT_TIMEOUT_MS: u64 = 90_000;
+pub const DEFAULT_OLLAMA_MAX_CONTEXT_TOKENS: u32 = 8192;
```
PR-Reviewer-Ant, line 554: The public function `resolve_ollama_max_context_tokens()` reads `SIGNET_OLLAMA_FALLBACK_MAX_CTX`. With this PR, that env var now controls both the fallback path *and* the explicitly-configured direct Ollama path (summary worker). The `FALLBACK` label in the env var name is now misleading: an operator who set `SIGNET_OLLAMA_FALLBACK_MAX_CTX=4096` to intentionally limit the fallback will inadvertently also cap their direct Ollama summary path at 4096, reproducing the truncation bug they may not be aware of. The TS counterpart `resolveDefaultOllamaFallbackMaxContextTokens()` has the same issue. Consider documenting that this env var applies to all Ollama paths, or introduce a distinct variable. Not blocking, but worth a docs/changelog note.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `78e1990a`

[Automated Bot Review] Incremental diff correctly resolves all five previously-open items. All `as` casts have been replaced with proper `in`-operator narrowing, the assertion now checks `>= 8192`, Rust parity is satisfied (summary worker at main.rs:948 passes `Some(resolve_ollama_max_context_tokens())`), and the TS variable rename (`ollamaMaxContextTokens`) is in place. One residual naming concern about the env var.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is small and unambiguous (7 lines changed in one test file). All prior open items are verifiable against the full file contents provided: the `as` casts are gone, the `>= 8192` assertion is explicit, Rust parity is confirmed at main.rs:948 and provider.rs:551-560, and the TS variable is renamed at summary-worker.ts:1341. The one remaining concern is directly readable in both the Rust and TS source.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): pass maxContextTokens to direct Ollama synthesis (#426)` (commit `3f9e9c2c`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3f9e9c2c`

You've hit your limit · resets Apr 2, 11pm (America/New_York)

packages/daemon/src/pipeline/summary-worker.test.ts
```diff
@@ -528,4 +528,40 @@ describe("resolveSummaryProvider", () => {
...
+`);
+
+		const provider = await resolveSummaryProvider(loadMemoryConfig(dir));
+		expect(provider.name).toBe("ollama:qwen3:4b");
+
+		// Intercept fetch to verify num_ctx is included in the request
+		const original = globalThis.fetch;
+		let captured: unknown = null;
+		globalThis.fetch = (async (_url: string | URL | Request, init?: RequestInit) => {
+			if (typeof init?.body !== "string") throw new Error(`Expected string body, got: ${typeof init?.body}`);
```
PR-Reviewer-Ant, line 546: `globalThis.fetch = (...) as typeof fetch` — the sole remaining `as` cast after the prior-round fixes. For a test-only global monkey-patch this is practically unavoidable in TypeScript, so it is a nitpick rather than a blocker, but it is technically a violation of the repo's 'no `as`' rule. One alternative is `Object.defineProperty(globalThis, 'fetch', { value: ..., writable: true })` with proper typing, though that adds noise for minimal gain.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3f9e9c2c`

[Automated Bot Review] One remaining warning (CHANGELOG documentation regression in incremental diff) and one residual nitpick (`as typeof fetch` cast in test). All prior blocking items have been resolved.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single CHANGELOG hunk. Prior open items are verifiable against the full file contents: `as` casts in the test are gone (replaced with `in`-operator narrowing), the `>= 8192` assertion is explicit, Rust parity is confirmed at main.rs:948 (`max_context_tokens: Some(resolve_ollama_max_context_tokens())`), and the TS variable is renamed to `ollamaMaxContextTokens` at summary-worker.ts:1341. The only new content in this incremental diff is the removal of the [Unreleased] CHANGELOG section.


- CHANGELOG.md:2 - The incremental diff removes the [Unreleased] section without moving its content into the [0.91.3] entry (which immediately follows). That section contained two user-facing notes: (1) the fix description for #426, and (2) an explicit behavior-change warning that `SIGNET_OLLAMA_FALLBACK_MAX_CTX` now controls **all** Ollama summary paths, not just the fallback. If these are not captured in [0.91.3], users upgrading will have no documented signal about the expanded env-var scope. AGENTS.md explicitly lists 'Doc and spec drift' as a top recurring failure mode and requires behavior changes to be documented in the same PR. The [0.91.3] section should include both the fix note and the `SIGNET_OLLAMA_FALLBACK_MAX_CTX` behavior-change callout.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `46ada4a1`

You've hit your limit · resets Apr 2, 11pm (America/New_York)

packages/daemon-rs/crates/signet-pipeline/src/provider.rs
PR-Reviewer-Ant, line 554: The env var `SIGNET_OLLAMA_FALLBACK_MAX_CTX` is named for the fallback path but now controls context window size for both the Ollama fallback *and* the explicitly-configured direct Ollama summary path (see main.rs:948). An operator who reads the env var name will assume it only applies to the degraded-fallback case and may not realize it also caps direct Ollama sessions. This won't cause a runtime bug today, but it creates a doc/naming mismatch that will bite whoever has to tune context size for an explicit Ollama deployment. Consider either renaming to `SIGNET_OLLAMA_MAX_CTX` (or `SIGNET_OLLAMA_CONTEXT_TOKENS`) with a migration note, or documenting the expanded scope in a comment alongside the var definition.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `46ada4a1`

[Automated Bot Review] Incremental diff (one hunk, 5 lines) resolves the last open `as typeof fetch` cast by declaring a properly-typed const before assignment. All prior blocking and warning items appear resolved in the full file view. One residual naming warning remains from the previous round.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single 5-line change in summary-worker.test.ts. All previously-open items are verifiable against the full file contents provided: the `as typeof fetch` cast is gone (replaced by typed const `mockFetch: typeof fetch`), all `as` casts in assertions are replaced with `in`-operator narrowing and `typeof` checks (lines 550–562 of the test file), the assertion uses `toBeGreaterThanOrEqual(8192)` (line 562), the TS variable is `ollamaMaxContextTokens` (summary-worker.ts:1341), and Rust parity is confirmed (main.rs:948 passes `max_context_tokens: Some(resolve_ollama_max_context_tokens())`). The one remaining concern — the env var name — is directly readable in provider.rs:554.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `23b39949`

You've hit your limit · resets Apr 2, 11pm (America/New_York)

packages/daemon/src/pipeline/summary-worker.test.ts
```diff
@@ -528,4 +528,41 @@ describe("resolveSummaryProvider", () => {
...
+
+		expect(captured).not.toBeNull();
+		if (typeof captured !== "object" || captured === null) throw new Error("Expected object body");
+		if (!("options" in captured)) throw new Error("Expected options field");
+		const { options } = captured;
+		if (typeof options !== "object" || options === null) throw new Error("Expected options object");
+		if (!("num_ctx" in options)) throw new Error("Expected num_ctx field");
+		const { num_ctx: numCtx } = options;
+		expect(typeof numCtx).toBe("number");
+		expect(numCtx).toBeGreaterThanOrEqual(8192);
```
PR-Reviewer-Ant, line 566: `expect(numCtx).toBeGreaterThanOrEqual(8192)` assumes the default or any configured value is ≥ 8192. If `SIGNET_OLLAMA_FALLBACK_MAX_CTX` is set to a value below 8192 in a test environment (e.g. a low-resource runner), the test fails even though the fix is correct. Consider asserting `numCtx === resolveDefaultOllamaFallbackMaxContextTokens()` (importing the function directly) or at minimum `expect(numCtx).toBeGreaterThan(0)` to decouple correctness from the default value assumption.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `23b39949`

[Automated Review — pr-reviewer] PR #430 correctly fixes the core issue (#426): `maxContextTokens` is now passed to the direct Ollama provider in `resolveSummaryProvider()`'s `default` case, and the regression test validates that `num_ctx` is included in the outbound request. Two concerns are noted — one warning about a pre-existing `break` bug whose scope is overstated by new documentation added in this PR, and one nitpick on a fragile test assertion.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The fix logic and regression test are fully visible in the diff. The `case 'ollama': break` issue is directly provable from the post-patch `summary-worker.ts` full content. `loadMemoryConfig` is not in the diff, so I cannot confirm what `synthesis.provider` defaults to; however, the pre-existing passing test 'falls back to resolved extraction config when synthesis is absent' strongly implies the default hits the `default` case, not `case 'ollama'`, confirming the break-path bug is dormant under normal configs.


- packages/daemon/src/pipeline/summary-worker.ts:1351 - Pre-existing logic bug whose coverage is overstated by new docs in this PR: `case 'ollama': break` does **not** fall through to `default` in JavaScript — `break` exits the switch. The comment `// Note: falls through to default` is incorrect. If a user explicitly sets `synthesis.provider: ollama`, `resolveSummaryProvider` silently returns `undefined` (async function falls off the end of the switch with no subsequent return statement), which will throw a `TypeError` when the caller invokes `provider.generate(...)`.

This bug is not introduced here. However, the new docstring added to `provider.ts` (line ~305) — and mirrored verbatim in `signet-pipeline/src/provider.rs` — states this `num_ctx` fix applies to *"an explicitly-configured `synthesis.provider = ollama` deployment"*. That claim is true for the Rust path (`from_config` handles the provider type uniformly) but **false for the TypeScript path** due to this break. The PR introduces documentation that overstates its own fix coverage.

PR #436, opened 2026-04-01
fix(daemon): resolve ReferenceError in /api/hooks/recall endpoint

ddasgupta4:
## Problem

The `/api/hooks/recall` route in the daemon references an undefined `cfg` variable when calling `hybridRecall()`, causing a `ReferenceError` on every request:

```
ReferenceError: cfg is not defined
    at <anonymous> (/opt/homebrew/lib/node_modules/signetai/dist/daemon.js:118382:13)
```

This returns HTTP 500 with `{"error": "Hook execution failed"}` for all recall hook requests.

## Impact

**Automatic memory recall injection is completely broken for OpenClaw plugin users.** The plugin's `before_prompt_build` hook calls `/api/hooks/recall` to inject relevant memories into agent context before each response. With this bug, every auto-recall returns 0 results, and the plugin falls back to a temporal-fallback engine that produces low-quality results (often searching for its own XML tags).

Direct recall via `/api/memory/recall` is unaffected — only the hooks pipeline path is broken.

## Root Cause

Every other route in `daemon.ts` that calls `hybridRecall()` creates a local `cfg` via `const cfg = loadMemoryConfig(AGENTS_DIR)`. The `/api/hooks/recall` route (line ~6092) was missing this declaration, referencing an out-of-scope `cfg` variable.

## Fix

Added `const cfg = loadMemoryConfig(AGENTS_DIR)` in the recall hook handler and passed it to `hybridRecall()`.

## Rust Parity

Verified the daemon-rs recall route uses direct SQL via `state.pool.read()` (no `hybridRecall(cfg, ...)` pattern), so the same ReferenceError does not exist. Applied search config (`top_k`, `min_score`) from `state.config.manifest.search` for behavioral parity. Documented config-freshness divergence (Rust reads startup state vs TS reads disk per-request) inline.

## Tests

Added `hooks-recall.test.ts` with 3 tests:
- Verifies 200 response on valid request (regression — no ReferenceError/500)
- Verifies 400 on missing `harness`
- Verifies 400 on missing `query`

All 3 pass. No new lint or typecheck errors introduced.

## PR Readiness Checklist

- [x] AI assistance disclosed


PR-Reviewer-Ant:
Hi @ddasgupta4 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): resolve ReferenceError in /api/hooks/recall endpoint` (commit `cc66216e`) now and I'll follow up shortly with feedback.

packages/daemon/src/hooks-recall.test.ts
```diff
@@ -0,0 +1,95 @@
...
+	afterAll(() => {
+		if (prev === undefined) {
+			process.env.SIGNET_PATH = undefined;
+		} else {
+			process.env.SIGNET_PATH = prev;
+		}
+		try {
+			rmSync(dir, { recursive: true, force: true });
+		} catch {
+			// cleanup best-effort
```
PR-Reviewer-Ant, line 39: Setting `process.env.SIGNET_PATH = undefined` does **not** unset the variable — in Bun/Node it coerces `undefined` to the string `"undefined"`, leaving `SIGNET_PATH="undefined"` in the environment for any tests that run later in the same process. Use `delete process.env.SIGNET_PATH` instead.

packages/daemon/src/daemon.ts
```diff
@@ -6115,6 +6115,7 @@ app.post("/api/hooks/recall", async (c) => {
 			sessionKey: body.sessionKey,
 		});
 		const agentScope = getAgentScope(agentId);
+		const memoryCfg = loadMemoryConfig(AGENTS_DIR);
```
PR-Reviewer-Ant: AGENTS.md mandates a Rust parity rule: any behavioural change to `@signet/daemon` must be reflected in `packages/daemon-rs/`. The same `cfg` bug likely exists in the shadow rewrite. The PR description doesn't mention checking or updating `packages/daemon-rs/`.

packages/daemon/src/hooks-recall.test.ts
```diff
@@ -0,0 +1,95 @@
...
+		}
+	});
+
+	it("does not throw ReferenceError for cfg", async () => {
+		const resp = await app.request("/api/hooks/recall", {
+			method: "POST",
+			headers: { "Content-Type": "application/json" },
+			body: JSON.stringify({
```
PR-Reviewer-Ant: The `it('does not throw ReferenceError…')` test and the `it('returns 200…')` test at line 63 send an identical request and assert overlapping conditions (not-500 vs exactly-200). They can be merged into one test, or the first can be removed — the 200-assertion already covers the ReferenceError regression.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant: Convention note: repo style prefers single-word identifiers (`cfg`) over two-word compounds (`memoryCfg`). The PR description says this was chosen to avoid shadowing — if there is no outer `cfg` in scope at this call site, a simple `cfg` matches existing patterns and the style guide.

packages/daemon/src/daemon.ts
```diff
@@ -6115,6 +6115,7 @@ app.post("/api/hooks/recall", async (c) => {
 			sessionKey: body.sessionKey,
 		});
 		const agentScope = getAgentScope(agentId);
+		const cfg = loadMemoryConfig(AGENTS_DIR);
```
PR-Reviewer-Ant, line 6118: AGENTS.md mandates: 'Any behavioral change made to @signet/daemon must also be reflected in `packages/daemon-rs/`.' This bug fix is a behavioral change (the endpoint was crashing; now it works). The incremental diff contains zero changes to `packages/daemon-rs/`, and this was flagged as an open item in the prior review round. The shadow daemon runs in parallel and logs divergences — leaving the Rust path with the same `cfg` reference bug means the shadow divergence log will continue to record crashes on this endpoint, defeating the purpose of the shadow proxy.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1709,14 +1709,26 @@ pub async fn recall(
         return conflict_response(claimed_by);
     }
 
-    let limit = body.limit.unwrap_or(10).min(50);
+    // Load search config — parity with TS `const cfg = loadMemoryConfig(AGENTS_DIR)`.
+    // Previously the handler used a hardcoded default of 10 and applied no min_score
+    // filter; the TS endpoint was updated to pass cfg through hybridRecall which
```
PR-Reviewer-Ant, line 1716: Config freshness parity: the TS handler calls `loadMemoryConfig(AGENTS_DIR)` on every request, which reads from disk each time. The Rust handler reads `state.config.manifest.search` from shared app state, which is likely populated at daemon startup. If `state.config` is not live-reloaded on `agent.yaml` changes, search config changes (e.g. updated `top_k` or `min_score`) will silently take no effect in the Rust daemon until restart, while the TS daemon picks them up immediately. This is a behavioural divergence from the stated parity goal. Please confirm whether `state.config` is refreshed on file changes or document the divergence.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1751,7 +1763,17 @@ pub async fn recall(
...
                     match rows {
-                        Ok(rows) => rows.filter_map(|r| r.ok()).collect::<Vec<_>>(),
+                        Ok(rows) => rows
+                            .filter_map(|r| r.ok())
+                            // Apply min_score filter (importance is the FTS-path
+                            // proxy for relevance score, matching TS cfg.search.min_score).
+                            .filter(|m| {
+                                m.get("importance")
+                                    .and_then(|v| v.as_f64())
+                                    .map(|imp| imp >= min_score)
```
PR-Reviewer-Ant, line 1775: `unwrap_or(true)` on the importance filter means any memory row whose `importance` field is missing or non-numeric will always pass the `min_score` gate. This may be intentional (don't silently discard unscored memories), but it diverges from TS behaviour if `hybridRecall` excludes such rows. Worth a comment explaining the intent.

packages/daemon/src/hooks-recall.test.ts
ddasgupta4: Fixed in f18939f7 — switched to `delete process.env.SIGNET_PATH`. Good catch on the string coercion.

packages/daemon/src/daemon.ts
ddasgupta4: Checked — daemon-rs recall route (`packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs:1682`) uses direct SQL via `state.pool.read()`, not a `hybridRecall(cfg, ...)` pattern. There is no `cfg` variable to be undefined. Addressed Rust parity in 941bd272 by applying search config (top_k, min_score) from `state.config.manifest.search`.

packages/daemon/src/hooks-recall.test.ts
ddasgupta4: Merged into a single test in f18939f7 — the 200 assertion now also checks `body.error !== "Hook execution failed"` to cover the regression case.

packages/daemon/src/daemon.ts
ddasgupta4: Renamed to `cfg` in f18939f7. No outer scope shadowing at this call site.

packages/daemon/src/daemon.ts
ddasgupta4: Addressed in 941bd272 — the Rust handler now loads `SearchConfig` from `state.config.manifest.search` (with defaults as fallback) and applies `top_k` and `min_score` to match the TS behavior of calling `loadMemoryConfig()` before `hybridRecall()`.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
ddasgupta4: Good observation. `state.config` in daemon-rs is populated at startup and not live-reloaded on `agent.yaml` changes — so yes, search config changes require a restart in the Rust daemon while TS picks them up immediately. This is a pre-existing divergence that predates this PR. I can add an inline comment documenting it, but fixing live config reload in the Rust daemon feels out of scope for a one-line recall bug fix. Happy to file a follow-up issue if preferred.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
ddasgupta4: The `unwrap_or(true)` is intentional — memories without importance scores (e.g. older records before importance was tracked) should not be silently excluded from recall results. The TS `hybridRecall` similarly includes unscored memories in results. Added a clarifying inline comment in 941bd272.

packages/daemon/src/hooks-recall.test.ts
```diff
@@ -0,0 +1,80 @@
+import { afterAll, beforeAll, describe, expect, it } from "bun:test";
```
PR-Reviewer-Ant, line 1: Nitpick: PR description still says '4 tests' but the file now has 3 (the duplicate no-500/200 tests were correctly consolidated per prior review feedback). The description is stale but the tests themselves are correct.

ddasgupta4: Updated PR description — now correctly says 3 tests. Thanks for the catch.

NicholaiVogel:
@ddasgupta4 nice job, thank you for taking time to fix this! merged

PR #443, opened 2026-04-02
fix: enforce 5000-token MEMORY.md budget

NicholaiVogel:
## Summary
Enforce a real 5000-token budget for `MEMORY.md` so synthesized working memory stays bounded and truncates overflow from the tail instead of growing without limit.

## Context
`MEMORY.md` is the prompt-facing global head for working memory, so unbounded growth increases prompt pressure and can crowd out newer context. The existing write path did not apply a tokenizer-backed budget at write time, which meant large synthesized heads could exceed the intended size cap.

## Changes
- added `js-tiktoken` to the daemon package for real token counting
- capped rendered `MEMORY.md` output at 5000 tokens using `cl100k_base`
- reserved space for the generated header comment before truncating body content
- truncated overflow from the end of the synthesized body
- stored the truncated body in head state so the DB lease/hash/revision state matches what is actually written to disk
- added regression coverage for both preserved short content and oversized tail truncation

The write path now projects the final file before commit, counts header tokens plus body tokens, and trims the body to fit the remaining token budget. This keeps the on-disk file and the persisted `memory_md_heads.content` value aligned, which avoids hash/revision drift when the projection has been truncated.

## Use Cases
- large synthesis runs no longer produce an oversized `MEMORY.md`
- the most recent/highest-priority content at the start of the head is preserved while older tail content is dropped first
- prompt injection stays bounded without relying on rough character-based estimates

## Testing
```bash
cd packages/daemon
bun test src/memory-head.test.ts src/pipeline/synthesis-worker.test.ts
bun run build
```

## Links
- None


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: enforce 5000-token MEMORY.md budget` (commit `46ec2103`) now and I'll follow up shortly with feedback.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: enforce 5000-token MEMORY.md budget` (commit `46ec2103`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `46ec2103`

packages/daemon/src/memory-head.ts
```diff
@@ -2,10 +2,15 @@ import { createHash, randomUUID } from "node:crypto";
...
 import { homedir } from "node:os";
 import { join } from "node:path";
+import { Tiktoken } from "js-tiktoken/lite";
+import cl100k_base from "js-tiktoken/ranks/cl100k_base";
 import { getDbAccessor } from "./db-accessor";
 import { countChanges } from "./db-helpers";
 import { loadMemoryConfig } from "./memory-config";
 
+const MEMORY_HEAD_MAX_TOKENS = 5000;
+const headTok = new Tiktoken(cl100k_base);
```
PR-Reviewer-Ant, line 12: AGENTS.md has an explicit Rust parity rule: 'Any behavioral change made to @signet/daemon must also be reflected in packages/daemon-rs/.' This PR adds token-budget enforcement to the MEMORY.md write path — a behavioral change — with no corresponding change in daemon-rs. The shadow proxy running in parallel will log every truncated write as a divergence until daemon-rs is updated. This needs to be addressed in the same PR.

packages/daemon/src/memory-head.test.ts
```diff
@@ -0,0 +1,72 @@
...
+	});
+
+	afterAll(() => {
+		rmSync(agentsDir, { recursive: true, force: true });
+		if (prevSignetPath === undefined) {
+			process.env.SIGNET_PATH = undefined;
+			return;
+		}
+		process.env.SIGNET_PATH = prevSignetPath;
+	});
```
PR-Reviewer-Ant, line 37: In Node.js and Bun, `process.env.SIGNET_PATH = undefined` coerces to the string `"undefined"` rather than removing the key, leaving SIGNET_PATH set to the string "undefined" for any tests that run after this suite in the same process. Use `delete process.env.SIGNET_PATH` instead.

packages/daemon/src/memory-head.test.ts
```diff
@@ -0,0 +1,72 @@
+import { afterAll, beforeAll, beforeEach, describe, expect, it } from "bun:test";
+import { mkdirSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { Tiktoken } from "js-tiktoken/lite";
+import cl100k_base from "js-tiktoken/ranks/cl100k_base";
+import { writeMemoryHead } from "./memory-head";
+
+const MEMORY_HEAD_MAX_TOKENS = 5000;
```
PR-Reviewer-Ant: `MEMORY_HEAD_MAX_TOKENS = 5000` is redefined here verbatim rather than imported from the source module. The repo convention explicitly prohibits duplicating constants across files. If the budget is later changed in memory-head.ts, the test will silently go stale. Export the constant from memory-head.ts and import it in the test.

bun.lock
```diff
@@ -27,7 +27,7 @@
     },
     "packages/cli": {
```
PR-Reviewer-Ant, line 28: All 15+ packages jump from version 0.86.3 to 0.91.4 — a 5 minor-version bump — for a PR filed as `fix:`. Per the repo's conventional-commits rule, `fix:` bumps the patch version only; `feat:` bumps minor. A 5-minor jump suggests this bun.lock was generated from a different branch state or the version sync included unrelated changes not present in this diff. Verify the version bump is intentional and that no unrelated commits were folded in.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `46ec2103`

[Automated Review — pr-reviewer bot] PR #443 correctly enforces a 5000-token MEMORY.md budget and aligns the DB hash with the truncated body. Three issues require attention before merge: a missing daemon-rs parity update (required by AGENTS.md), an env-var cleanup bug in the test teardown, and a duplicated constant that violates the repo's single-source-of-truth rule.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All three blocking/warning findings are directly provable from the changed files: (1) no diff exists for packages/daemon-rs/ while AGENTS.md explicitly mandates parity — this is checkable from the diff alone; (2) the process.env assignment to undefined is visible at memory-head.test.ts:38; (3) the constant duplication is visible at memory-head.test.ts:9 vs memory-head.ts:11. The decode return-type concern was ruled out after verifying js-tiktoken returns string, not Uint8Array.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e0ca3880`

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -1975,6 +2012,55 @@ mod tests {
         fs::remove_dir_all(root).ok();
```
PR-Reviewer-Ant, line 2012: The regression test `projection_truncates_oversized_memory_md_to_token_budget` does not actually exercise the truncation path.

`build_ledger` emits only the `memory_sentence` field (≈20 tokens) per session, never the full artifact body. The ~6000-token body ending with `"TRUNCATE-ME-TAIL"` is stored in `memory_artifacts.content` but is never rendered verbatim into the projection. Additionally, the `memories`, `session_summaries`, and `memory_thread_heads` tables are empty in this test, so the full rendered projection is on the order of 100–200 tokens — far below the 5000-token budget. As a result:
- `assert!(!rendered.content.contains(tail))` is trivially true (tail was never a candidate to appear)
- `assert!(tok.encode_ordinary(&file).len() <= MEMORY_MD_MAX_TOKENS)` passes without truncation ever running

Per the repo's testing philosophy ("Every test should encode *what must be true*… A test that would break if you rewrote the module… is testing plumbing, not theory"), this test provides no regression coverage for the Rust truncation path. To fix: call `truncate_memory_projection` directly with a string of known token count (>5000), assert the returned string is ≤5000 tokens, and assert specific text from beyond the cutoff is absent. Alternatively, populate `memories` or `memory_thread_heads` with enough items to push the rendered projection over 5000 tokens.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e0ca3880`

[Automated Review — pr-reviewer bot] PR #443 (incremental: 46ec2103→e0ca3880) has addressed all four previously open items. One new warning: the Rust regression test does not actually exercise the truncation code path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The data flow from `write_summary_artifact` → `build_ledger` → rendered projection is fully visible in the provided `memory_lineage.rs` content. `build_ledger` produces per-session lines using only `session.sentence` (the short `memory_sentence` field from the artifact row). The full body text (`content`) is passed to `coerce_sentence` only for validation/fallback logic and is never emitted verbatim into the projection. 'TRUNCATE-ME-TAIL' lives in the full body and will never appear in rendered output regardless of whether truncation runs. Additionally, the `memories`, `session_summaries`, and `memory_thread_heads` tables are empty in this test, so the rendered projection is ~100–200 tokens — well under 5000. The `tok.encode_ordinary(&file).len() <= MEMORY_MD_MAX_TOKENS` assertion is therefore satisfied without the truncation path ever executing. All prior open items (parity, env cleanup, constant duplication, version bump) are confirmed resolved in the incremental diff.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8a3c600e`

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -1975,6 +2012,56 @@ mod tests {
...
     }
 
+    #[test]
```
PR-Reviewer-Ant, line 2015: Previously flagged concern RESOLVED. The old test seeded data via `write_summary_artifact`, whose body only surfaces as a ~20-token `memory_sentence` in the ledger, making the pre-truncation render ~100 tokens and the truncation path dead. The new test seeds the `memories` table directly so each row appears verbatim in both the Global Head and Durable Notes sections (8 rows × ~500 tokens × 2 = ~8000 tokens). The pre-condition assertion `tok.encode_ordinary(&pre.content).len() > MEMORY_MD_MAX_TOKENS` is a load-bearing guard that makes the test self-validating — if the setup ever produces a render below budget, the test will fail before reaching the budget assertion, preventing silent regression.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8a3c600e`

[Automated Review — pr-reviewer bot] PR #443 (incremental: e0ca3880→8a3c600e) resolves the last open blocking concern. The Rust regression test now correctly exercises the truncation code path. Three of the four remaining open items are confirmed resolved in the current file state; the version-bump concern from prior reviews is still present but not touched by this incremental diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single-function rewrite of the Rust test. The data flow through render_memory_projection is fully visible in the provided memory_lineage.rs content: the memories table rows are emitted verbatim in both the Global Head and Durable Notes sections, so 8 rows × ~500 tokens × 2 appearances ≈ 8000 tokens exceeds the 5000-token budget. The pre-condition assertion (pre.content tokens > MEMORY_MD_MAX_TOKENS) structurally guarantees truncation executes before the budget assertion is checked. All other previously flagged items are verifiable from the full file contents provided as context.


- bun.lock:28 - Carried forward from prior open review state (no new diff evidence): package.json still shows `"version": "0.91.4"` for a PR filed as `fix:`. Per the repo's conventional-commits rule, `fix:` bumps patch only. The bun.lock incremental diff in this commit is empty — no new evidence to re-evaluate — but the concern remains unresolved. If the 0.91.4 version predates this branch, this is a non-issue; if it was bumped as part of this PR's commit history, the commit type should be `feat:` or the version should be reverted to a patch increment.

NicholaiVogel:
Re: version-bump concern on bun.lock — confirming this is a non-issue. Version `0.91.4` was released to `main` before this branch was cut (`55995fac chore: release 0.91.4`). The workspace entries in `bun.lock` had a stale `0.86.3` snapshot that predates this PR. This branch only adds the `js-tiktoken` entry to `bun.lock` and does not bump any workspace versions.

PR #442, opened 2026-04-02
feat: dreaming memory consolidation — periodic LLM-driven knowledge graph refinement

NicholaiVogel:
## Summary

Implements periodic LLM-driven knowledge graph consolidation ("dreaming"), where a background worker uses a smart model to review accumulated session summaries and the entity graph, then applies structured mutations to merge duplicates, prune junk, update stale attributes, and integrate new learnings.

## Context

The knowledge graph accumulates entities and attributes through the extraction pipeline during sessions, but over time this leads to duplicate entities (e.g., "TypeScript" vs "Typescript"), stale attributes, junk fragments from markdown artifacts, and missed cross-entity relationships. The dreaming spec (`docs/specs/approved/dreaming-memory-consolidation.md`) defines a periodic consolidation pass that uses a reasoning-capable model to review and clean the graph between sessions.

### Core Agent (`packages/daemon/src/pipeline/dreaming.ts`)
- Prompt builder that assembles identity files, working memory, session summaries, and the full entity graph into a structured prompt
- JSON mutation parser with code-fence stripping and graceful error handling
- 8 mutation operations: `create_entity`, `merge_entities`, `delete_entity`, `update_aspect`, `delete_aspect`, `supersede_attribute`, `create_attribute`, `delete_attribute`
- Safety guards: pinned entities cannot be deleted, constraint attributes cannot be superseded
- All mutations applied atomically in a single write transaction

### Trigger Mechanism
- Token accumulation hook in `summary-worker.ts` -- each completed session summary adds its token count to `dreaming_state`
- Threshold check: when `tokensSinceLastPass >= tokenThreshold` (default 500K), a pass triggers
- First-run backfill: with `backfillOnFirstRun: true`, the first pass uses `compact` mode to clean the existing graph

### Infrastructure
- **Migration 055**: `dreaming_state` (token counter, last pass tracking) and `dreaming_passes` (pass history with status/mutations/errors)
- **Config**: `DreamingConfig` type with `tokenThreshold`, `backfillOnFirstRun`, `maxInputTokens`, `maxOutputTokens`, `timeout`; defaults in `memory-config.ts`; configurable via `agent.yaml > memory.dreaming`
- **Worker**: `dreaming-worker.ts` polls every 5 min, supports manual trigger via API
- **Provider**: uses the synthesis provider (shared with dependency synthesis) — all synthesis backends supported (`claude-code`, `codex`, `opencode`, `anthropic`, `openrouter`, `ollama`); no separate dreaming-specific provider config

[...]

PR #444, opened 2026-04-02
feat(connector-forge): add ForgeCode connector

NicholaiVogel:
## Summary

Add `@signet/connector-forge` to integrate Signet identity, memory (via MCP), and skills into the ForgeCode harness.

## Context

ForgeCode is a Rust-based AI coding agent that reads global instructions from `~/forge/AGENTS.md`, loads skills from `~/forge/skills/`, and discovers MCP servers from `~/forge/.mcp.json`. Unlike Claude Code or Codex, ForgeCode has no external hook/callback system -- its lifecycle hooks are compiled-in Rust with no subprocess, HTTP, or plugin extension points. This means the integration surface is limited to static file placement and MCP server registration, which is what this connector implements.

Previously the CLI's `configureHarnessHooks` had a no-op stub for the `"forge"` case. This PR replaces that with a real connector.

## Changes

- **`packages/connector-forge/src/index.ts`** -- `ForgeConnector` class extending `BaseConnector` with `install()`, `uninstall()`, and `isInstalled()`:
  - Generates `~/forge/AGENTS.md` with Signet marker header and full identity extras (SOUL, IDENTITY, USER, MEMORY)
  - Registers a `signet` MCP server entry in `~/forge/.mcp.json` (preserves unrelated entries)
  - Symlinks `~/.agents/skills` into `~/forge/skills/`
  - Validates identity before install; fails fast with a clear message
  - Uninstall removes only Signet-managed artifacts (checks marker, preserves user config)

- **`packages/cli/src/cli.ts`** -- Wired the `"forge"` case in `configureHarnessHooks` to call `ForgeConnector.install()`

- **`package.json` / `packages/cli/package.json` / `bun.lock`** -- Added `@signet/connector-forge` as workspace dependency and to the `build:deps` filter

- **`docs/specs/planning/connector-forgecode.md`** -- Planning spec documenting the integration strategy, ForgeCode's architecture constraints, and future opportunities

- **`docs/specs/INDEX.md` / `docs/specs/dependencies.yaml`** -- Registered the spec with hard dependency on `signet-runtime` and 4 success criteria

ForgeCode's lifecycle hooks (`Start`, `End`, `Request`, `Response`, `ToolcallStart`, `ToolcallEnd`) are Rust-native and composed at compile time in `ForgeApp::chat()`. There is no external hook registration mechanism. This means the connector operates in a **pull-only model** -- the agent must explicitly call Signet MCP tools rather than receiving automatic memory injection/extraction. This is an architectural limitation of the harness, not a gap in the connector.

The connector reuses `BaseConnector` utilities: `stripLegacySignetBlock()`, `symlinkSkills()`, `generateHeader()`, and `composeIdentityExtras()` -- matching the pattern used by the OpenCode connector.

## Testing

```bash
# Run connector tests (5 tests)
bun test packages/connector-forge/index.test.ts

# Tests cover:
# - Full install flow (AGENTS.md, MCP config, skills symlink)
# - Uninstall preserving unrelated MCP servers
# - Graceful failure on invalid JSON config
# - Build wiring verification (build:deps filter, CLI dependency)
```

```bash
# Build verification
bun run build        # Connector builds as part of build:deps
bun run typecheck    # Clean
```

## Links

- Planning spec: `docs/specs/planning/connector-forgecode.md`
- ForgeCode reference: `references/forgecode/`
- ForgeCode env/paths: `references/forgecode/crates/forge_domain/src/env.rs`


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(connector-forge): add ForgeCode connector` (commit `7a146c65`) now and I'll follow up shortly with feedback.

BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(connector-forge): add ForgeCode connector` (commit `7a146c65`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7a146c65`

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,241 @@
...
+		const filesRemoved: string[] = [];
+		const configsPatched: string[] = [];
+
+		const agentsPath = this.getAgentsPath();
+		if (existsSync(agentsPath)) {
+			const content = readFileSync(agentsPath, "utf-8");
+			if (content.includes(SIGNET_FORGE_MARKER)) {
+				rmSync(agentsPath, { force: true });
+				filesRemoved.push(agentsPath);
+			}
```
PR-Reviewer-Ant, line 198: **uninstall() does not remove skill symlinks created during install.** `install()` calls `this.symlinkSkills(skillsSource, this.getSkillsPath())`, which creates per-skill symlinks under `~/forge/skills/`. `uninstall()` only removes AGENTS.md and the signet MCP entry. After uninstall, `~/forge/skills/<name>` symlinks remain pointing at `~/.agents/skills/<name>`. If the workspace is later deleted, ForgeCode will encounter broken symlinks. The spec's own validation criteria say the connector should clean up managed artifacts on uninstall. The symlink directory (or individual managed symlinks) should be removed here, matching how `install` tracks what it created.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,241 @@
...
+		const strippedAgentsPath = this.stripLegacySignetBlock(expandedBasePath);
+		if (strippedAgentsPath !== null) {
+			filesWritten.push(strippedAgentsPath);
+		}
+
+		const forgeHome = this.getForgeHome();
+		mkdirSync(forgeHome, { recursive: true });
+
+		const agentsPath = this.generateAgentsMd(expandedBasePath);
+		filesWritten.push(agentsPath);
```
PR-Reviewer-Ant: **`stripLegacySignetBlock(expandedBasePath)` passes a directory, not a file path.** Without BaseConnector source it cannot be confirmed, but every other connector call in this codebase pattern passes a specific target file path to `stripLegacySignetBlock`. Here `expandedBasePath` is a directory like `~/.agents`. If the base implementation expects a file path to strip inline, this call is a no-op at best or silently corrupts the wrong file at worst. Please confirm against the OpenCode connector's usage of this method.

packages/connector-forge/index.test.ts
```diff
@@ -0,0 +1,178 @@
...
+});
+
+afterEach(() => {
+	if (typeof originalHome === "string") {
+		process.env.HOME = originalHome;
+	} else {
+		process.env.HOME = undefined;
+	}
+	if (tmpRoot) {
+		rmSync(tmpRoot, { recursive: true, force: true });
```
PR-Reviewer-Ant, line 47: **`process.env.HOME = undefined` sets HOME to the string `"undefined"`, not unset it.** In both Node.js and Bun, assigning `undefined` to an env var converts it to the string `"undefined"`. The correct idiom is `delete process.env.HOME`. As written, if `originalHome` was `undefined` at process start, any test infrastructure running after the suite (e.g. other test files in the same worker) sees HOME as `"undefined"`. Change to `delete process.env.HOME` for the `else` branch.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,241 @@
...
+	}
+
+	getConfigPath(): string {
+		return this.getMcpConfigPath();
+	}
+
+	async install(basePath: string): Promise<InstallResult> {
+		const filesWritten: string[] = [];
+		const configsPatched: string[] = [];
+		const expandedBasePath = expandHome(basePath || join(getHomeDir(), ".agents"));
```
PR-Reviewer-Ant, line 122: **`args: []` is always emitted in the MCP server entry, even on non-Windows where it is always empty.** Some MCP clients treat an explicit empty `args` array differently from a missing `args` key. On non-Windows the resolved `args` is always `[]` (from `resolveSignetMcp`). Consider omitting the key when empty: `...(mcp.args.length > 0 ? { args: mcp.args } : {})`. This matches typical MCP config conventions used elsewhere in the codebase.

package.json
```diff
@@ -1,51 +1,46 @@
 {
```
PR-Reviewer-Ant, line 1: **Root `package.json` and `packages/cli/package.json` were fully reformatted (2-space → tabs) adding ~80 lines of diff noise.** The functional change is only the `@signet/connector-forge` additions. The mass-reformat makes it harder to audit the actual logic changes and will cause spurious merge conflicts on in-flight branches touching these files. Recommend reverting to original indentation and applying only the minimal additive changes.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7a146c65`

[Automated Bot Review — pr-reviewer] PR #444 adds @signet/connector-forge. The connector logic is largely sound and matches codebase patterns, but uninstall leaves managed skill symlinks behind (functional gap vs stated behavior), a test teardown assigns undefined to an env var incorrectly, and a cross-module call cannot be verified without BaseConnector source.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The uninstall skill-cleanup gap and the env-var teardown bug are directly readable in the diff. The stripLegacySignetBlock argument type concern requires BaseConnector source (packages/connector-base/src/index.ts is not in the diff) to resolve — the call passes a directory path, and whether that is the correct calling convention for this method is ambiguous from the diff alone.

NicholaiVogel:
Addressed all 5 review comments in `58925099`:

1. **Skill symlink cleanup on uninstall** — Added `removeSkillSymlinks()` that iterates `~/forge/skills/`, removes only symlink entries (preserves user-created directories), and cleans up the empty directory. Test updated to verify.

2. **`stripLegacySignetBlock()` usage** — This is actually correct. `BaseConnector.stripLegacySignetBlock()` accepts a directory path (`basePath`) and internally joins `AGENTS.md` (`connector-base/src/index.ts:107-108`). Every other connector calls it identically (claude-code:111, codex:242, opencode:245, openclaw:338, oh-my-pi:162). No change needed.

3. **`process.env.HOME = undefined`** — Reverted to `delete process.env.HOME` with a `biome-ignore` suppression explaining why `delete` is required (env vars coerce `undefined` to the string `"undefined"`).

4. **Empty `args` array** — MCP entry now conditionally spreads `args` only when non-empty. Test updated to assert `args` is absent on non-Windows.

5. **Package.json formatting noise** — Restored both `package.json` and `packages/cli/package.json` to original formatting with only the minimal additive changes (one line each).

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `58925099`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `58925099`

[Automated Bot Review — pr-reviewer] Re-review of #444 at 58925099. All five previously flagged items are resolved. Two new findings remain: the connector-forgecode spec was not promoted from `planning` to `approved`/`complete` despite shipping implementation (violates mandatory AGENTS.md session protocol and PR readiness checklist), and `removeSkillSymlinks()` removes every symlink in `~/forge/skills/` regardless of origin, not only Signet-managed ones.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The spec-drift finding is directly readable in docs/specs/dependencies.yaml (status: planning unchanged) and docs/specs/INDEX.md. The removeSkillSymlinks scope issue is directly readable in packages/connector-forge/src/index.ts lines 220–234. BaseConnector source (packages/connector-base/src/index.ts) is still not in the diff, so the stripLegacySignetBlock and symlinkSkills calling conventions remain unverifiable — the author's rationale is consistent with the call-site patterns visible elsewhere in cli.ts, so no action needed there.

docs/specs/dependencies.yaml
```diff
@@ -331,6 +331,22 @@ specs:
...
+    path: docs/specs/planning/connector-forgecode.md
+    hard_depends_on:
+      - signet-runtime
+    soft_depends_on: []
+    blocks: []
+    informed_by:
+      - docs/research/technical/RESEARCH-REFERENCE-REPOS.md
+    success_criteria:
+      - "ForgeCode reads a Signet-managed global AGENTS.md from `~/forge/AGENTS.md` that preserves Signet identity context"
+      - "ForgeCode loads Signet skills from `~/forge/skills` without copying over user-owned directories"
```
PR-Reviewer-Ant, line 345: The `connector-forgecode` spec remains at `status: planning` while this PR ships the full implementation. Per the mandatory session protocol in AGENTS.md (step 3: confirm spec status before coding; step 4: keep INDEX.md and dependencies.yaml in sync), and the PR readiness checklist (Spec alignment validated), the status should be promoted to at least `approved` — ideally `complete` given that the implementation is live and all five validation criteria in the spec are satisfied by this PR. This is one of the explicitly listed recurring PR failure modes (#4 doc and spec drift).

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,273 @@
...
+			return "signet" in servers;
+		} catch {
+			return false;
+		}
+	}
+
+	private removeSkillSymlinks(filesRemoved: string[]): void {
+		const skillsDir = this.getSkillsPath();
+		if (!existsSync(skillsDir)) return;
+
```
PR-Reviewer-Ant, line 270: `removeSkillSymlinks()` removes every symlink found in `~/forge/skills/`, regardless of whether it was created by Signet. If a user manually created a symlink in `~/forge/skills/` pointing to an unrelated skill, it would be silently deleted on uninstall. The spec says 'Existing non-Signet MCP entries and real skill directories remain untouched' and 'keeps user-owned non-symlinked skill directories intact', but user-owned symlinks are not protected. A targeted fix would resolve only symlinks whose target resolves to a path under `skillsSource` (the Signet workspace skills dir). The current test only verifies that a Signet-created symlink is removed; it does not verify that a pre-existing non-Signet symlink is preserved after uninstall.

NicholaiVogel:
Second round of feedback addressed in `092b00f4`:

1. **`removeSkillSymlinks()` too broad** — Now scoped to Signet-owned symlinks only. Reads `SIGNET_PATH` from the MCP config `env` block via `extractSignetPath()`, resolves each symlink target with `readlinkSync`, and skips any that don't point into `SIGNET_PATH/skills`. User-created symlinks (e.g. to a personal toolbox) are never touched. Test added to verify preservation.

2. **Spec not promoted** — `connector-forgecode` moved from `docs/specs/planning/` → `docs/specs/complete/` (git rename, not copy). `dependencies.yaml` path and status updated. Entry added to the INDEX registry table with `complete` status.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `092b00f4`

docs/specs/complete/connector-forgecode.md
```diff
@@ -0,0 +1,143 @@
+---
+title: "Connector: ForgeCode"
+id: connector-forgecode
```
PR-Reviewer-Ant, line 3: The spec file frontmatter still reads `status: planning` (and the body on line 12 repeats `Status: planning`), but the file lives in `docs/specs/complete/`, `docs/specs/dependencies.yaml` records it as `status: complete`, and `docs/specs/INDEX.md` lists it as `complete`. The prior review flagged that the spec wasn't promoted; the registries were updated in this commit but the spec file content wasn't, introducing a new internal inconsistency. Both `status: planning` lines in the frontmatter and body should be updated to `complete`.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,303 @@
...
+		if (!isJsonObject(env)) return null;
+		const path = env.SIGNET_PATH;
+		return typeof path === "string" && path.length > 0 ? path : null;
+	}
+
+	private removeSkillSymlinks(filesRemoved: string[], signetPath: string | null): void {
+		const skillsDir = this.getSkillsPath();
+		if (!existsSync(skillsDir)) return;
+
+		// Without a known source we cannot safely scope removal — skip.
```
PR-Reviewer-Ant: `resolved.startsWith(signetSkillsSource)` uses bare string prefix matching without a trailing separator. If `signetSkillsSource` is `/home/user/.agents/skills`, a symlink pointing to `/home/user/.agents/skills-backup/test-skill` would satisfy the check and be incorrectly deleted during uninstall. Change to `resolved.startsWith(signetSkillsSource + "/") || resolved === signetSkillsSource` (or normalize with `path.sep`) to ensure only paths that are genuinely within the skills directory are matched.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `092b00f4`

[Automated Bot Review — pr-reviewer] PR #444 at 092b00f4. The removeSkillSymlinks origin-scoping and test teardown fixes from prior reviews are confirmed present. Two issues remain: the spec file frontmatter still says `planning` while the registries now say `complete` (a new inconsistency introduced by partially addressing the prior finding), and the symlink-origin check uses bare `startsWith` without a trailing separator, which can produce false positives.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly readable in the diff: the spec file frontmatter `status: planning` vs registry `status: complete` is explicit in docs/specs/complete/connector-forgecode.md vs docs/specs/dependencies.yaml and docs/specs/INDEX.md. The startsWith prefix check is explicit at packages/connector-forge/src/index.ts lines 242–243 and the test at index.test.ts line 148 does not exercise the sibling-prefix edge case. No cross-module context is needed to evaluate either finding.

NicholaiVogel:
Third round addressed in `ce22223d`:

1. **`startsWith` path-prefix collision** — Changed to `resolved.startsWith(\`${signetSkillsSource}/\`) || resolved === signetSkillsSource`. A path like `/home/user/.agents/skills-backup/foo` no longer satisfies the check and won't be deleted on uninstall.

2. **Spec file frontmatter inconsistency** — Updated both `status: planning` (YAML frontmatter, line 4) and `- Status: \`planning\`` (body metadata, line 23) in `docs/specs/complete/connector-forgecode.md` to `complete`, consistent with `dependencies.yaml` and `INDEX.md`.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ce22223d`

packages/cli/src/cli.ts
```diff
@@ -297,8 +298,11 @@ async function configureHarnessHooks(
 			break;
 		}
 		case "forge": {
-			// Forge is a first-party native harness managed as its own product.
-			// There are no hook/config patches to apply from the Signet CLI.
+			const connector = new ForgeConnector();
```
PR-Reviewer-Ant, line 302: **Pattern inconsistency: Forge is the only connector that throws on install failure.** Every other case in `configureHarnessHooks` (`claude-code`, `codex`, `opencode`, `oh-my-pi`, `openclaw`) calls `await connector.install(basePath)` and ignores the return value entirely. The Forge case checks `result.success` and throws `new Error(result.message)` on failure. Concretely: if a user has a malformed `~/forge/.mcp.json`, setup will abort with a thrown error for Forge but silently continue for any other harness. The behavior may be intentional (Forge fails loudly), but it diverges from the established contract and will surprise callers that wrap `configureHarnessHooks` in a try/catch expecting only exceptions for truly exceptional failures. At minimum the divergence should be documented in a comment near the case.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,303 @@
...
+	return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+function readJsonObject(path: string): JsonObject {
+	if (!existsSync(path)) {
+		return {};
+	}
+
+	const raw = readFileSync(path, "utf-8");
+	const parsed: unknown = JSON.parse(raw);
```
PR-Reviewer-Ant, line 36: **Windows MCP path resolution is fragile.** `process.argv[1]` is the CLI entry script (`…/dist/cli.js`), not the connector package. `join(entry, '..', '..', 'bin', 'mcp-stdio.js')` resolves to `packages/cli/bin/mcp-stdio.js` relative to the CLI dist output — not the connector's own `bin/` directory. If that path happens to exist in the installed layout, it may work. If it doesn't exist, the code silently falls back to `'signet-mcp'`, which on Windows is unlikely to be on `PATH` after a standard npm/bun global install (there's no `.cmd` shim created for `signet-mcp`). This means the MCP server entry written to `~/forge/.mcp.json` will silently fail at ForgeCode startup on Windows. The path derivation should be verified against the published package layout, or the Windows fallback should register the `.cmd` shim path instead.

docs/specs/complete/connector-forgecode.md
```diff
@@ -0,0 +1,143 @@
...
+status: complete
+informed_by:
+  - docs/research/technical/RESEARCH-REFERENCE-REPOS.md
+section: "Connectors"
+depends_on:
+  - "signet-runtime"
```
PR-Reviewer-Ant, line 9: `draft_quality: "implementation-aligned planning stub"` is inconsistent with `status: complete` and the spec being placed under `docs/specs/complete/`. Future readers using this metadata to filter spec maturity will see a spec marked `complete` but described as a "stub". Consider updating `draft_quality` to reflect its actual state (e.g., `implementation-aligned` or removing the field).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ce22223d`

[Automated Bot Review — pr-reviewer] PR #444 (@signet/connector-forge) is in good shape. Both findings from the prior review (spec status not promoted; `removeSkillSymlinks` removing all symlinks regardless of origin) are resolved in the current diff. Three new items remain: a minor inconsistency in how `configureHarnessHooks` handles install failures for the Forge harness vs. all other harnesses, a fragile Windows MCP-path derivation, and a stale `draft_quality` label in the spec.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The three findings are directly readable in the changed files. The BaseConnector source (`packages/connector-base/src/index.ts`) is not in the diff, so `stripLegacySignetBlock`, `symlinkSkills`, `generateHeader`, and `composeIdentityExtras` can only be validated by their call-sites and naming conventions, not their implementations. Prior reviews flagged these as unknowns and no new evidence changes that assessment.

NicholaiVogel:
Fourth round addressed in `07f75626`:

1. **`configureHarnessHooks` pattern inconsistency** — Removed the `result.success` check and `throw`. Forge now uses the same fire-and-forget pattern as every other connector: `await connector.install(basePath)` with no return value inspection.

2. **Windows MCP path fragility** — Switched from `bin/mcp-stdio.js` to `dist/mcp-stdio.js` to match the Claude Code / OpenCode convention (the established path for the bundled mcp-stdio entry point). Added `console.warn` on fallback when the path can't be resolved, matching the error-surfacing pattern in those connectors.

3. **`draft_quality` inconsistency** — Updated from `"implementation-aligned planning stub derived from reference-repo comparison"` to `"implementation-aligned"`, consistent with the spec being in `complete` status.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `07f75626`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `07f75626`

[Automated Bot Review — pr-reviewer] PR #444 at 07f75626. Both previously flagged blockers are resolved: `connector-forgecode` spec is now marked `complete` in INDEX.md and dependencies.yaml, and `removeSkillSymlinks` is now scoped to symlinks that point into the Signet workspace. One new warning and one nitpick remain.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - Full connector source (src/index.ts), tests, and spec files are in the diff. The Windows path issue is directly provable from the string comparisons at src/index.ts:226–228. BaseConnector source (packages/connector-base/src/index.ts) is not in the diff — `symlinkSkills`, `stripLegacySignetBlock`, `composeIdentityExtras`, `generateHeader`, and `atomicWriteJson` calling conventions are inferred from call-site patterns and test assertions rather than verified from source.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,310 @@
...
+				const content = readFileSync(this.getAgentsPath(), "utf-8");
+				if (content.includes(SIGNET_FORGE_MARKER)) {
+					return true;
+				}
+			} catch {
+				// Ignore unreadable AGENTS.md and fall through to MCP config check.
+			}
+		}
+
+		try {
```
PR-Reviewer-Ant, line 247: Windows path compatibility gap in skill-symlink origin check. `linkTarget.startsWith("/")` does not match Windows absolute paths (e.g. `C:\...`), and the subsequent `resolved.startsWith(\`${signetSkillsSource}/\`)` uses a forward slash that won't match Windows path separators. On Windows every symlink silently fails the origin check and is skipped, leaving stale symlinks behind after uninstall. Fix: use `path.isAbsolute(linkTarget)` for the absolute-path branch and `path.resolve()` + a separator-aware prefix check (or `path.relative()` + check that the relative path does not start with `..`) for the comparison. Low real-world impact since ForgeCode is Unix/macOS-first, but `resolveSignetMcp` already handles Windows specially, so the asymmetry is worth closing.

docs/specs/complete/connector-forgecode.md
```diff
@@ -0,0 +1,143 @@
+---
```
PR-Reviewer-Ant, line 1: PR description says 'Planning spec: `docs/specs/planning/connector-forgecode.md`' but the spec file shipped in this PR lives at `docs/specs/complete/connector-forgecode.md`. Stale PR description text only — no code or index impact.

NicholaiVogel:
Fifth round addressed in `84302eb9`:

1. **Windows path compatibility in `removeSkillSymlinks()`** — Two fixes:
   - `linkTarget.startsWith("/")` → `isAbsolute(linkTarget)`: correctly handles Windows absolute paths like `C:\...`
   - Dropped the `startsWith(signetSkillsSource + "/")` check entirely in favour of `path.relative()`: compute `rel = relative(signetSkillsSource, resolved)` and skip when `rel.startsWith("..")` or `isAbsolute(rel)`. This is cross-platform, handles same-prefix sibling directories, and handles different Windows drive letter roots.
   - Added `isAbsolute`, `relative`, `resolve` to the `node:path` import.

2. **Stale PR description** — Updated the two references to `docs/specs/planning/connector-forgecode.md` → `docs/specs/complete/connector-forgecode.md`. Also corrected test count (5 → 6) and description to mention user-owned symlink preservation.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `84302eb9`

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,315 @@
...
+	private removeSkillSymlinks(filesRemoved: string[], signetPath: string | null): void {
+		const skillsDir = this.getSkillsPath();
+		if (!existsSync(skillsDir)) return;
+
+		// Without a known source we cannot safely scope removal — skip.
+		if (signetPath === null) return;
+		const signetSkillsSource = join(signetPath, "skills");
+
+		try {
+			for (const entry of readdirSync(skillsDir)) {
```
PR-Reviewer-Ant, line 279: `rel.startsWith("..")` is a correct-direction check but too broad: `path.relative()` can return a value like `"..skillname"` when the resolved target lives directly inside `signetSkillsSource` but its directory entry happens to begin with the characters `..`. In that case `startsWith("..")` is `true` and the symlink is incorrectly preserved during uninstall.

The safe fix is to match only the canonical parent-escape prefixes:
```ts
if (rel === ".." || rel.startsWith(".." + sep) || isAbsolute(rel)) continue;
```
where `sep` is `path.sep`. On POSIX, `"../"` is the only escape prefix; on Windows it would be `"..\\"`. The current code without a separator guard would silently leave behind a symlink named `"..someSkill"` if one ever existed in the Signet workspace.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `84302eb9`

[Automated Bot Review — pr-reviewer] PR #444: ForgeCode connector implementation. One prior warning-level finding remains unresolved; prior blocking finding (spec frontmatter status) is confirmed resolved in this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The `startsWith` check is directly readable in the diff at the identified line in removeSkillSymlinks. The spec frontmatter now reads `status: complete` in the new `docs/specs/complete/connector-forgecode.md` file, confirming that prior finding is resolved. No runtime context is needed to verify either claim.

NicholaiVogel:
Sixth round addressed in `085c2c92`:

**`rel.startsWith("..")` too broad** — A skill directory literally named `"..something"` would produce `rel = "..something"` from `path.relative()`, which starts with `".."` but is actually *inside* `signetSkillsSource` and should be unlinked on uninstall. Fixed by checking only canonical parent-escape prefixes:

```ts
if (rel === ".." || rel.startsWith(`..${sep}`) || isAbsolute(rel)) continue;
```

- `rel === ".."` — resolved is the parent directory itself
- `rel.startsWith(`..${sep}`)` — resolved escapes via `../` (or `..\` on Windows)
- `isAbsolute(rel)` — cross-drive Windows path (different root)

A skill named `"..hidden"` produces `rel = "..hidden"`, which passes all three guards and is correctly unlinked. Added `sep` to the `node:path` import.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `085c2c92`

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `085c2c92`

[Automated Bot Review — pr-reviewer] PR #444 (@signet/connector-forge) is structurally sound. The two regressions called out in the prior review (spec status not promoted to `complete`; `removeSkillSymlinks` removing symlinks regardless of origin) are both resolved in this diff. Two prior-review open items remain unaddressed, and a test-coverage gap is newly visible.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The full connector source, test suite, and CLI integration are visible. The BaseConnector methods (`stripLegacySignetBlock`, `symlinkSkills`, `generateHeader`, `composeIdentityExtras`) are not in the diff and their behavior is inferred only from test assertions. The prior review's third open item was truncated mid-sentence and cannot be confirmed resolved or still present.

packages/cli/src/cli.ts
```diff
@@ -297,8 +298,8 @@ async function configureHarnessHooks(
 			break;
 		}
```
PR-Reviewer-Ant, line 300: `ForgeConnector.install()` returns a structured `{ success: false, message: '...' }` for two distinct failure modes — missing identity and unreadable MCP config — with informative human-readable messages. The `configureHarnessHooks` case discards the return value, so users who hit either failure path during `signet setup` receive no diagnostic output. Every other harness follows the same discard pattern, but Forge is the first connector with meaningful `success: false` payloads, making this a practical gap unique to this PR. Surfacing `result.message` (via `console.error` or by returning it for the caller to log) would fix it.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,317 @@
...
+	// argv[1] is e.g. .../signetai/bin/signet.js; mcp-stdio.js lives at
+	// .../signetai/dist/mcp-stdio.js (matches Claude Code / OpenCode pattern).
+	const cliEntry = process.argv[1] || "";
+	const mcpJs = join(cliEntry, "..", "..", "dist", "mcp-stdio.js");
+	if (existsSync(mcpJs)) {
+		return { command: process.execPath, args: [mcpJs] };
+	}
+
+	console.warn(
+		`[signet] Warning: could not resolve mcp-stdio.js from argv[1]="${cliEntry}". MCP server config will use "signet-mcp" which may fail on Windows without shell:true.`,
```
PR-Reviewer-Ant, line 79: On Windows the absolute path to `mcp-stdio.js` is resolved from `process.argv[1]` at install time and written verbatim into `~/forge/.mcp.json`. If signet is updated or relocated after the initial setup, the baked-in path goes stale and ForgeCode silently fails to spawn the MCP server. The `console.warn` at line 82 fires at install time only — there is no runtime re-validation. Consider documenting that `signet setup` must be re-run after signet upgrades, or store a PATH-resolved command (`signet-mcp`) on Windows as a fallback at the cost of requiring it on PATH.

packages/connector-forge/index.test.ts
```diff
@@ -0,0 +1,220 @@
...
+	});
+
+	it("wires the CLI to depend on the Forge connector package", () => {
+		const dir = dirname(fileURLToPath(import.meta.url));
+		const path = resolve(dir, "..", "cli", "package.json");
+		const text = readFileSync(path, "utf-8");
+
+		expect(text).toContain('"@signet/connector-forge": "workspace:*"');
+	});
+});
```
PR-Reviewer-Ant, line 233: The test suite exercises the invalid-MCP-JSON failure path but not the `hasValidIdentity → false` path. Per the repo's recurring-failures checklist, each guarded failure mode should have a test that would fail before the guard existed. A test calling `connector.install(emptyDir)` and asserting `result.success === false` with the expected message string would close this gap and match the pattern of the existing MCP-error test.

NicholaiVogel:
Seventh round addressed in `bf869a7e`:

1. **Silent install failures in CLI** — The `"forge"` case now captures the return value and emits a `chalk.yellow` warning when `result.success === false`. Forge is the first connector with meaningful failure payloads (missing identity, malformed MCP JSON), so users running `signet setup` now see a diagnostic instead of a silent no-op. No throw — consistent with the fire-and-forget model of other connectors.

2. **Windows baked-in path staleness** — Added a comment to `resolveSignetMcp()` documenting the known limitation: the absolute `mcp-stdio.js` path is written to `~/forge/.mcp.json` at install time and goes stale if signet is updated or relocated. Mitigation: re-running `signet setup` resolves the path afresh and overwrites the stale entry. A runtime re-validation mechanism is out of scope for this connector.

3. **Missing `hasValidIdentity → false` test** — Added a test that calls `connector.install()` with a basePath containing no identity files, asserts `result.success === false` and that the message contains `"No valid Signet identity found at"` plus the path. Test count is now 7.

4. **Import ordering** — Fixed `expandHome` import ordering in `cli.ts` (alphabetical, pre-existing lint issue surfaced by the forge case change).

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `bf869a7e`

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,323 @@
...
+				rmSync(agentsPath, { force: true });
+				filesRemoved.push(agentsPath);
+			}
+		}
+
+		// Read MCP config first so we can derive SIGNET_PATH before patching it.
+		const mcpPath = this.getMcpConfigPath();
+		let config: JsonObject = {};
+		if (existsSync(mcpPath)) {
+			try {
```
PR-Reviewer-Ant, line 205: **Mixed exception contract — uncaught throws after the validated-failure early returns.**

`install()` uses a result-type pattern (`{ success: false, message }`) for the two guarded failures (missing identity, invalid JSON config). But the code that follows — `mkdirSync(forgeHome)`, `generateAgentsMd()` (calls `readFileSync` + `writeFileSync`), `atomicWriteJson()`, and `this.symlinkSkills()` — can all throw (permission denied, disk full, ENOENT race, etc.) and those exceptions propagate uncaught.

The caller in `cli.ts` does:
```ts
const result = await connector.install(basePath);
if (!result.success) { console.warn(...) }
```
There is no `try/catch` around the call. An uncaught throw here crashes the setup flow entirely, leaving AGENTS.md and `.mcp.json` already written (partially installed state) with no cleanup and no user-friendly message.

Fix: either wrap the post-guard body in a `try/catch` that returns `{ success: false, message }` on any error, or add a `try/catch` in the `cli.ts` caller. The former keeps the contract consistent with the design intent.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,323 @@
...
+		try {
+			const config = readJsonObject(this.getMcpConfigPath());
+			const servers = readMcpServers(config);
+			return "signet" in servers;
+		} catch {
+			return false;
+		}
+	}
+
+	private extractSignetPath(config: JsonObject): string | null {
```
PR-Reviewer-Ant, line 256: **`removeSkillSymlinks` silently skips cleanup when `signetPath` is null.**

```ts
// Without a known source we cannot safely scope removal — skip.
if (signetPath === null) return;
```

`signetPath` comes from `extractSignetPath(config)`, which reads `SIGNET_PATH` out of the MCP config. If the user manually edited `.mcp.json` after install (removed `SIGNET_PATH`, renamed the key, etc.) or if a partial uninstall already stripped the MCP entry, `signetPath` is `null` and ALL skill symlinks in `~/forge/skills/` are left behind silently. No warning is logged, nothing is added to `filesRemoved`.

This is a data-integrity issue: `uninstall()` claims to clean up managed artifacts but leaves orphaned symlinks without any indication. At minimum, log a warning when skipping. Optionally: fall back to removing symlinks whose targets resolve to anywhere inside the Signet skills dir pattern (`~/.agents/skills/` or the default) so cleanup succeeds in the common case even without an exact path.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,323 @@
...
+			};
+		}
+
+		const strippedAgentsPath = this.stripLegacySignetBlock(expandedBasePath);
+		if (strippedAgentsPath !== null) {
+			filesWritten.push(strippedAgentsPath);
+		}
+
+		const forgeHome = this.getForgeHome();
+		mkdirSync(forgeHome, { recursive: true });
```
PR-Reviewer-Ant: Both early-return failure paths omit `configsPatched` from the returned object:
```ts
return {
    success: false,
    message: `No valid Signet identity found at ${expandedBasePath}`,
    filesWritten,
    // configsPatched missing
};
```
If `InstallResult.configsPatched` is required (not optional), any caller doing `result.configsPatched.forEach(...)` will throw a runtime error on these paths. TypeScript catches this at build time only if the field is non-optional. Worth confirming the type, or explicitly including `configsPatched: []` for clarity.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `bf869a7e`

[Automated Review — pr-reviewer] PR #444 adds `@signet/connector-forge`. The overall structure is solid and follows the BaseConnector pattern. Two issues stand out: a mixed exception/result-type contract in `install()` that can crash the setup flow, and a silent skill-cleanup gap in `uninstall()`.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The exception-contract issue is directly visible in the diff at the `symlinkSkills`, `generateAgentsMd`, and `mkdirSync` call sites. The confidence is medium rather than high because `BaseConnector.symlinkSkills` is not in this diff — if it already swallows all errors internally (as some connectors' base helpers do), the blocking finding would downgrade to a warning. The silent-skip finding in `removeSkillSymlinks` is unambiguous from the diff alone.

NicholaiVogel:
Eighth round addressed in `4bb8c8f4`:

1. **Mixed exception contract in `install()`** — Wrapped the write phase (`stripLegacySignetBlock`, `mkdirSync`, `generateAgentsMd`, `atomicWriteJson`, `symlinkSkills`) in a try-catch that returns `{ success: false, message, filesWritten, configsPatched }` on any infrastructure error (permission denied, disk full, ENOENT race, etc.). The contract is now consistent across all failure paths — no unhandled throws propagate out of `install()`.

2. **`configsPatched` missing from early returns** — Added `configsPatched: []` to both `success: false` early-return paths (missing identity, invalid JSON config). The field is always present on all return paths, matching the success case and avoiding `undefined` dereferences in callers.

3. **`removeSkillSymlinks` null fallback** — Replaced the `if (signetPath === null) return` early exit with a fallback to the default workspace: `const source = signetPath ?? join(getHomeDir(), ".agents")`. If the user manually edited `.mcp.json` to remove `SIGNET_PATH` (or a partial uninstall already stripped the entry), skill symlinks pointing at `~/.agents/skills` are still cleaned up using the well-known default path rather than being silently left behind.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4bb8c8f4`

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,341 @@
...
+		const mcpPath = this.getMcpConfigPath();
+		let config: JsonObject = {};
+		if (existsSync(mcpPath)) {
+			try {
+				config = readJsonObject(mcpPath);
+			} catch {
+				// Leave unreadable user config untouched.
+			}
+		}
+
```
PR-Reviewer-Ant, line 211: **Bug: `readFileSync` in `uninstall()` is not inside a try/catch.** Every other risky filesystem operation in this method (`readJsonObject`, `atomicWriteJson`, `removeSkillSymlinks`) is wrapped in a `try/catch` that leaves the user's config untouched on error. The `readFileSync(agentsPath)` call here is not protected. A race condition (file deleted between `existsSync` and `readFileSync`) or a permission error will throw an unhandled exception and reject the returned Promise, which can cause an unhandled rejection in the CLI if the caller doesn't catch it. Wrap lines 207-215 in a `try/catch` to match the rest of the function.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,341 @@
...
+		// disk full, ENOENT race, etc.) are surfaced as result-type failures
+		// rather than thrown exceptions, keeping the contract consistent.
+		try {
+			const strippedAgentsPath = this.stripLegacySignetBlock(expandedBasePath);
+			if (strippedAgentsPath !== null) {
+				filesWritten.push(strippedAgentsPath);
+			}
+
+			const forgeHome = this.getForgeHome();
+			mkdirSync(forgeHome, { recursive: true });
```
PR-Reviewer-Ant, line 157: **Data-integrity gap: `stripLegacySignetBlock` modifies the source workspace AGENTS.md before the write phase completes.** `stripLegacySignetBlock(expandedBasePath)` is the first call in the write-phase `try` block, so it mutates the user's source file (in `basePath`) before `generateAgentsMd` and `atomicWriteJson` have run. If either of those subsequent steps throws (e.g. disk full, permission denied on `~/forge/`), `install()` returns `{ success: false }`, but the source AGENTS.md has already been stripped. There is no rollback. The caller in `cli.ts` only logs a warning and does not attempt to restore the source. The source modification should either be moved after the target write succeeds, or a backup/restore step should guard it.

packages/connector-forge/src/index.ts
```diff
@@ -0,0 +1,341 @@
...
+				const linkTarget = readlinkSync(target);
+				const resolved = isAbsolute(linkTarget) ? linkTarget : resolve(skillsDir, linkTarget);
+				const rel = relative(signetSkillsSource, resolved);
+				// Skip when resolved is outside signetSkillsSource. Use the canonical
+				// parent-escape prefix (rel === ".." or starts with "../") rather than
+				// bare startsWith("..") which would incorrectly exclude skill names
+				// that happen to begin with the characters "..".
+				if (rel === ".." || rel.startsWith(`..${sep}`) || isAbsolute(rel)) continue;
+				unlinkSync(target);
+				filesRemoved.push(target);
```
PR-Reviewer-Ant, line 295: **Unconditional overwrite of `~/forge/AGENTS.md` without checking for existing user content.** `generateAgentsMd` calls `writeFileSync(targetPath, ...)` without first checking whether the target already exists and contains non-Signet content. The `uninstall()` method is careful to check `content.includes(SIGNET_FORGE_MARKER)` before removing the file, but `install()` makes no analogous check before overwriting. If a user has hand-written content in `~/forge/AGENTS.md` before running `signet setup --harness forge`, it is silently destroyed. At minimum, log a warning when overwriting an existing file that lacks the Signet marker. If other connectors already follow the same unconditional-overwrite pattern, this can be closed as consistent.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4bb8c8f4`

[Automated Review — pr-reviewer bot] PR #444 adds @signet/connector-forge. The implementation matches the stated goals (AGENTS.md generation, MCP registration, skills symlink). Three issues found: one real bug in uninstall(), one data-integrity gap in install(), and one unconditional overwrite that is inconsistent with the marker-guarded uninstall.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The unprotected readFileSync in uninstall() is directly visible in the diff. The stripLegacySignetBlock partial-write concern depends on the BaseConnector implementation in @signet/connector-base, which is not in this diff — the risk assessment is based on the method name and call site ordering. The unconditional overwrite concern could be dismissed if other connectors follow the same pattern, which is also not visible in this diff.

NicholaiVogel:
Folded in the daemon schema fix from #449 (`a32b6578`):

**`fix(daemon): remove propertyNames from MCP tool schemas for OpenAI compat`**

`z.record(z.string(), ValueType)` emits `propertyNames: { type: "string" }` in the Zod JSON Schema output. Claude accepts this non-standard keyword; OpenAI rejects it with a hard 400 (`"Invalid schema for function 'mcp_signet_tool_mcp_server_call'"`). This broke ForgeCode entirely when using Codex/GPT models — which is precisely the scenario this connector enables.

Fix: drop the explicit key-type arg on all 4 affected `z.record()` calls in `packages/daemon/src/mcp/tools.ts` (`memory_feedback.ratings/paths/rewards`, `mcp_server_call.args`).

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a32b6578`

NicholaiVogel:
Addressed both pre-merge items in `543c6954`:

1. **`secret_exec` schema** — Fixed `secrets: z.record(z.string(), z.string())` which was missed in the previous commit. Now `z.object({}).catchall(z.string())`.

2. **Regression test + corrected approach** — The previous fix (`z.record(ValueType)` single-arg) turned out to be equally broken:

   | Form | Behavior with MCP SDK 1.26.0 + Zod v3 |
   |---|---|
   | `z.record(z.string(), V)` | Serializes OK, emits `propertyNames` (OpenAI rejects) |
   | `z.record(V)` | Crashes `tools/list` with `_zod undefined` (Zod v3/v4 mismatch) |
   | `z.object({}).catchall(V)` | Serializes cleanly, emits `additionalProperties` only ✓ |

   The regression test (`createMcpServer > schema compatibility > no tool schema emits propertyNames`) uses `InMemoryTransport` + `Client.listTools()` to exercise the actual serialization path the MCP server uses when ForgeCode initializes — not just the schema definition. All 5 affected fields (`memory_feedback.ratings/paths/rewards`, `mcp_server_call.args`, `secret_exec.secrets`) now use `catchall`. 24/24 tests pass.

Also updated the installed `~/.agents/` bundle with the correct patches so the fix is live immediately without waiting for publish.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `543c6954`

NicholaiVogel:
Merge conflict with `docs/specs/INDEX.md` resolved in `c8a0b28d`. Merged `origin/main` (which had concurrent spec index updates from #442) into the branch — took main's new entries and kept the `CFC` node + edges we added. No functional changes.

PR #445, opened 2026-04-02
fix(cli): use wall-clock deadline for daemon startup poll and exit non-zero on failure

### LINKED ISSUE #429: signet daemon start reports 'Failed to start daemon' on successful start (poll timeout too short)
# Bug Report: `signet daemon start` reports "Failed to start daemon" on successful start

**Signet version:** 0.86.3 (present since at least v0.77.5, likely much earlier)
**Platform:** Linux (Ubuntu 24.04, kernel 6.8.0-106-generic x86_64)
**Node:** v25.6.0 | **Bun:** 1.3.11
**Harness:** OpenClaw
**Date observed:** 2026-03-30 through 2026-03-31

---

## Summary

`signet daemon start` (and `signet start`) always print `✖ Failed to start daemon` — even when the daemon starts successfully. The error message is a false negative: the bun process spawns correctly, the health endpoint eventually responds, and `signet status` confirms the daemon is running. The root cause is a hardcoded 5-second poll timeout in `startDaemon()` that is shorter than the actual startup time on many systems (~8+ seconds observed on this machine).

---

## Environment

```
OS:       Ubuntu 24.04 LTS (Linux 6.8.0-106-generic x86_64)
signet:   v0.86.3
bun:      1.3.11
hardware: VM — 4 vCPU, 8GB RAM (OpenClaw gateway)
install:  npm global (/home/linuxbrew/.linuxbrew/lib/node_modules/signetai)
```

---

## Steps to Reproduce

1. Ensure the daemon is not running: `signet daemon stop`
2. Run: `signet daemon start` (or `signet start`)
3. Observe output

**Expected:**
```
◈ signet v0.86.3
  own your agent. bring it anywhere.

- Starting daemon...
✔ Daemon started
```

**Actual:**
```
◈ signet v0.86.3
  own your agent. bring it anywhere.

- Starting daemon...
✖ Failed to start daemon
```

4. Wait ~3–5 seconds after the failure message, then run `signet status`
5. Observe: daemon is running, health endpoint responds, memories accessible

---

### Source location

The bug is in `startDaemon()` in `cli/src/features/daemon.ts` (compiled to `dist/cli.js` at line ~122264 in v0.86.3).

### The poll loop

After spawning the daemon process, `startDaemon()` polls the health endpoint with a fixed loop:

```js
// dist/cli.js ~122329
for (let i = 0; i < 20; i += 1) {
  await sleep2(250);             // 250ms per iteration
  if (await isDaemonRunning()) { // HTTP GET /health with 1200ms timeout
    return true;
  }
}
return false;  // ← returns false = "failed"
```

**Total poll budget: 20 × 250ms = 5000ms maximum.**

### The health check

`isDaemonRunning()` calls `isDaemonHealthyAt()` which does:

```js
const response = await fetch(`${baseUrl}/health`, {
  signal: AbortSignal.timeout(1200)  // 1.2s per health probe
});
return response.ok;
```

Each poll iteration can take up to **1.2s** if the connection is refused (timeout fires). With a refused connection (daemon not yet bound), each iteration takes ~250ms + ~1200ms = **up to ~1450ms**. Under worst-case timing, the poll loop exhausts in as few as **3–4 iterations** rather than 20.

### Measured startup time

Empirical measurement on this machine (Ubuntu 24.04 VM):

```
poll 1  (250ms):   connection refused
poll 2  (596ms):   connection refused
...
poll 20 (5115ms):  connection refused  ← startDaemon() returns false HERE
poll 21 (5366ms):  connection refused
poll 22 (5617ms):  connection refused
...
poll 32 (8078ms):  HTTP 200 ✔ READY   ← daemon actually ready ~3s after "failure"
```

**The daemon starts successfully at ~8 seconds. The CLI gives up at ~5 seconds.**

### Why the exit code is 0

`doStart()` does not propagate a non-zero exit code when `startDaemon()` returns false:

```js
async function doStart(options, deps) {
  ...
  const started = await deps.startDaemon(basePath);
  if (started) {
    spinner.succeed("Daemon started");
    ...
    return;
  }
  spinner.fail("Failed to start daemon");
  // ← no process.exit(1), no throw — function returns normally
}
```

This is why the CLI exits with code 0 despite displaying a failure message — the error path doesn't set a non-zero exit code. Scripts and orchestrators that rely on exit code to determine success will see code 0 and think everything is fine; users reading the output will think it failed.

### Why this affects many machines

The 1.2-second `AbortSignal.timeout` on each health probe combines with the 250ms `sleep2` per iteration. On a machine where each iteration costs close to its maximum (connection refused triggers the full 1.2s timeout), the 20-iteration budget is exhausted in:

`20 × (250ms + ~1200ms) ≈ 29 seconds`

But on a machine where connections are refused instantly (TCP RST), iterations cost only ~250ms each, consuming the budget in exactly 5 seconds — before a daemon that needs 6–10 seconds to start is ready.

The actual behavior depends on how fast the OS rejects connections to an unbound port, which varies by kernel, network stack tuning, and load.

---

## Impact

- **Severity:** Medium — cosmetic but operationally confusing
- **User impact:** Every user who runs `signet daemon start` on a system where startup takes >5 seconds sees a false failure. Since `signet daemon start` is also called internally by other commands (`doRestart`, `launchDashboard`, setup flows), this failure message appears in many common workflows.
- **Automation impact:** Scripts using `signet daemon start` get exit code 0 with a failure message — misleading but not blocking (daemon starts anyway). However, any script that parses output for `✖` to detect failure will incorrectly trigger error handling.
- **Workaround:** Users must either ignore the message and wait, or run `signet status` to confirm the daemon is actually running.

---

### Option 1: Increase poll count (minimal change)

Increase iterations from 20 to 60 (15 seconds total at 250ms/poll, ignoring health probe timeout):

```diff
- for (let i = 0; i < 20; i += 1) {
+ for (let i = 0; i < 60; i += 1) {
```

**Pros:** One-line fix. **Cons:** Doesn't account for health probe timeout stacking.

### Option 2: Track wall-clock time instead of iteration count (recommended)

Replace the fixed loop with a deadline-based loop:

```js
const deadline = Date.now() + 15_000; // 15-second wall-clock budget
while (Date.now() < deadline) {
  await sleep2(250);
  if (await isDaemonRunning()) {
    return true;
  }
}
return false;
```

**Pros:** Budget is always ~15 real seconds regardless of how long each health probe takes. Handles both fast-reject and slow-timeout cases correctly.

### Option 3: Fix exit code (independent, should also be done)

Regardless of poll strategy, `doStart()` should exit with a non-zero code on failure:

```diff
  spinner.fail("Failed to start daemon");
+ process.exit(1);
```

---

## Workaround

Until fixed, users can ignore the `✖ Failed to start daemon` message and run `signet status` or wait ~10 seconds before using the CLI. The daemon will be running.

---

## Additional Notes

- The `signet start` alias (`startAlias` command) calls `doStart()` through the same code path and exhibits the same behavior.
- `signet daemon restart` also calls `startDaemon()` and will show `✖ Failed to restart daemon` for the same reason.
- The `launchDashboard` function (used by `signet open` or similar) checks `started` and calls `process.exit(1)` on failure — so the dashboard launch path does exit non-zero, but the plain `signet daemon start` path does not.
- This issue is tracked internally as issue #331 (referenced in OpenClaw's HEARTBEAT.md watch list).

NicholaiVogel:
## Summary

Fix false "Failed to start daemon" message caused by a poll timeout that was shorter than actual daemon startup time, and add missing non-zero exit codes on all daemon lifecycle failure paths.

## Context

`signet daemon start` (and `signet start`) always reports "Failed to start daemon" on systems where the daemon takes more than ~5 seconds to start — even though the daemon starts successfully. The root cause is a hardcoded 20-iteration poll loop (20 x 250ms = 5s nominal budget) in `startDaemon()`, but each `isDaemonRunning()` call includes a 1.2s `AbortSignal.timeout` on the health probe. Under connection-refused conditions, each iteration can consume up to ~1.45s, exhausting the loop in as few as 3-4 polls — well before many systems finish starting (~8s observed on Ubuntu 24.04 VM).

Additionally, `doStart()`, `doStop()`, and `doRestart()` all printed failure messages but returned with exit code 0, misleading scripts and users.

Reported in #429 (tracked internally as #331).

### 1. Wall-clock deadline for startup polling (`packages/cli/src/lib/runtime.ts`)

Replaced the fixed `for (let i = 0; i < 20; ...)` loop with a deadline-based `while (Date.now() < deadline)` loop using a 15-second wall-clock budget. This guarantees ~15 real seconds of polling regardless of how long each health probe takes, handling both fast-reject (TCP RST) and slow-timeout scenarios correctly.

### 2. Non-zero exit codes on failure (`packages/cli/src/features/daemon.ts`)

Added `process.exit(1)` to four failure paths that previously returned normally:

| Function | Failure message |
|---|---|
| `doStart` | "Failed to start daemon" |
| `doStop` | "Failed to stop daemon" |
| `doRestart` | "Failed to stop daemon" (during restart) |
| `doRestart` | "Failed to restart daemon" |

`launchDashboard` already had `process.exit(1)` — no change needed.

## Testing

All 6 existing daemon tests pass:

```bash
bun test packages/cli/src/features/daemon.test.ts
```

Manual verification:
1. `signet daemon stop`
2. `signet daemon start` — should no longer show false failure on systems with >5s startup
3. On actual failure (e.g. port conflict), CLI now exits with code 1

## Links

- Closes #429

### DIFF
--- a/packages/cli/src/features/daemon.ts
+++ b/packages/cli/src/features/daemon.ts
@@ -205,6 +205,7 @@ export async function doStart(options: PathOptions, deps: Deps): Promise<void> {
 	}
 
 	spinner.fail("Failed to start daemon");
+	process.exit(1);
 }
 
 export async function doStop(options: PathOptions, deps: Deps): Promise<void> {
@@ -225,6 +226,7 @@ export async function doStop(options: PathOptions, deps: Deps): Promise<void> {
 	}
 
 	spinner.fail("Failed to stop daemon");
+	process.exit(1);
 }
 
 export async function doRestart(options: RestartOptions, deps: Deps): Promise<void> {
@@ -238,7 +240,7 @@ export async function doRestart(options: RestartOptions, deps: Deps): Promise<vo
 		const stopped = await deps.stopDaemon(basePath);
 		if (!stopped) {
 			spinner.fail("Failed to stop daemon");
-			return;
+			process.exit(1);
 		}
 		await deps.sleep(500);
 	}
@@ -253,7 +255,7 @@ export async function doRestart(options: RestartOptions, deps: Deps): Promise<vo
 		}
 	} else {
 		spinner.fail("Failed to restart daemon");
-		return;
+		process.exit(1);
 	}
 
 	if (options.openclaw === false || !isOpenClawDetected()) {
--- a/packages/cli/src/lib/runtime.ts
+++ b/packages/cli/src/lib/runtime.ts
@@ -454,13 +454,27 @@ export async function startDaemon(agentsDir: string = AGENTS_DIR): Promise<boole
 		}
 	});
 
+	// Track process exit so the poll loop can short-circuit on fast failures
+	// (port conflict, missing binary, bad config) rather than waiting the
+	// full deadline.
+	let procExited = false;
+	proc.on("exit", () => {
+		procExited = true;
+	});
+
 	proc.unref();
 	if (stderrFd !== null) {
 		closeSync(stderrFd);
 	}
 
-	for (let i = 0; i < 20; i += 1) {
+	// Use wall-clock deadline instead of iteration count so the budget
+	// is always ~15 real seconds regardless of how long each health
+	// probe takes (connection-refused can stall up to 1.2s per probe).
+	// If the spawned process exits early (fast failure), break immediately.
+	const deadline = Date.now() + 15_000;
+	while (Date.now() < deadline) {
 		await sleep(250);
+		if (procExited) break;
 		if (await isDaemonRunning()) {
 			return true;
 		}
--- a/packages/cli/src/features/daemon.test.ts
+++ b/packages/cli/src/features/daemon.test.ts
@@ -1,5 +1,5 @@
-import { describe, expect, it } from "bun:test";
-import { doRestart, doStop, requestPipelinePauseApi, summarizePipelineToggle } from "./daemon.js";
+import { afterEach, describe, expect, it, spyOn } from "bun:test";
+import { doRestart, doStart, doStop, requestPipelinePauseApi, summarizePipelineToggle } from "./daemon.js";
 
 describe("requestPipelinePauseApi", () => {
 	it("uses the live daemon pause endpoint when available", async () => {
@@ -119,5 +119,64 @@ describe("daemon lifecycle recovery", () => {
 		await doStop({}, deps);
 
 		expect(stopped).toBe(true);
+	});
+});
+
+// Regression: #429 — failure paths must exit non-zero
+describe("daemon exit codes on failure", () => {
+	let exitSpy: ReturnType<typeof spyOn>;
+
+	afterEach(() => {
+		exitSpy?.mockRestore();
+	});
+
+	it("doStart exits with code 1 when startDaemon returns false", async () => {
+		exitSpy = spyOn(process, "exit").mockImplementation(() => {
+			throw new Error("EXIT_1");
+		});
+
+		const deps = makeDeps({ startDaemon: async () => false });
+
+		await expect(doStart({}, deps)).rejects.toThrow("EXIT_1");
+		expect(exitSpy).toHaveBeenCalledWith(1);
+	});
+
+	it("doStop exits with code 1 when stopDaemon returns false", async () => {
+		exitSpy = spyOn(process, "exit").mockImplementation(() => {
+			throw new Error("EXIT_1");
+		});
+
+		const deps = makeDeps({
+			isDaemonRunning: async () => true,
+			stopDaemon: async () => false,
+		});
+
+		await expect(doStop({}, deps)).rejects.toThrow("EXIT_1");
+		expect(exitSpy).toHaveBeenCalledWith(1);
+	});
+
+	it("doRestart exits with code 1 when startDaemon returns false", async () => {
+		exitSpy = spyOn(process, "exit").mockImplementation(() => {
+			throw new Error("EXIT_1");
+		});
+
+		const deps = makeDeps({ startDaemon: async () => false });
+
+		await expect(doRestart({ openclaw: false }, deps)).rejects.toThrow("EXIT_1");
+		expect(exitSpy).toHaveBeenCalledWith(1);
+	});
+
+	it("doRestart exits with code 1 when stopDaemon returns false during restart", async () => {
+		exitSpy = spyOn(process, "exit").mockImplementation(() => {
+			throw new Error("EXIT_1");
+		});
+
+		const deps = makeDeps({
+			isDaemonRunning: async () => true,
+			stopDaemon: async () => false,
+		});
+
+		await expect(doRestart({ openclaw: false }, deps)).rejects.toThrow("EXIT_1");
+		expect(exitSpy).toHaveBeenCalledWith(1);
 	});
 });


BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): use wall-clock deadline for daemon startup poll and exit non-zero on failure` (commit `9f95f8d6`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9f95f8d6`

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): use wall-clock deadline for daemon startup poll and exit non-zero on failure` (commit `9f95f8d6`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3fdd73c8`

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): use wall-clock deadline for daemon startup poll and exit non-zero on failure` (commit `3fdd73c8`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9f95f8d6`

The background fetch hit a `gh` auth wall — no additional context retrieved. The review above is complete and based on the full diff and file contents provided in the prompt.

packages/cli/src/lib/runtime.ts
```diff
@@ -459,7 +459,11 @@ export async function startDaemon(agentsDir: string = AGENTS_DIR): Promise<boole
 		closeSync(stderrFd);
 	}
 
-	for (let i = 0; i < 20; i += 1) {
+	// Use wall-clock deadline instead of iteration count so the budget
+	// is always ~15 real seconds regardless of how long each health
```
PR-Reviewer-Ant, line 471: **UX regression on fast-fail**: The 15s wall-clock budget solves the slow-start false-negative, but it also means a daemon that fails immediately (port conflict, missing binary, bad config) will make the user wait the full ~15 seconds before seeing the error — up from ~3-4s with the old iteration loop under connection-refused conditions. The PR description acknowledges the tradeoff implicitly but it's worth deciding whether 15s is the right number. A process-exit sentinel (check if the spawned `proc` has already exited via an `'exit'` event listener set before the poll loop) would bound the fast-fail wait to milliseconds without changing the slow-start behaviour.

packages/cli/src/features/daemon.ts
```diff
@@ -205,6 +205,7 @@ export async function doStart(options: PathOptions, deps: Deps): Promise<void> {
 	}
```
PR-Reviewer-Ant, line 205: **Inconsistency with `togglePipelinePause` failure paths**: The PR correctly adds `process.exit(1)` to `doStart`/`doStop`/`doRestart`, but the analogous `stopDaemon`/`startDaemon` failure branches inside `togglePipelinePause` (the `signet pause`/`signet resume` commands) still `return` with exit code 0. If the intent is that all daemon lifecycle failures signal non-zero to callers/scripts, these paths are inconsistent. If the partial-success semantics there (config written, daemon restart failed) are intentional, a comment explaining why they're exempt would prevent future confusion.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3fdd73c8`

[Automated Review — pr-reviewer bot] PR #445 addresses two distinct problems: a wall-clock deadline replacing a fixed iteration poll in `startDaemon`, and missing `process.exit(1)` on failure paths in the daemon lifecycle commands. The implementation is correct and the regression tests cover the stated failure cases. Two minor concerns are worth discussing before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both changed files are fully visible in the diff and the complete file contents are provided. The fix to the polling loop is directly verifiable from `runtime.ts:459-470` (old iteration count) and the new `deadline` pattern. The `process.exit(1)` additions are directly provable from the four diff hunks in `daemon.ts`. The test assertions are unambiguous.

packages/cli/src/lib/runtime.ts
NicholaiVogel: Good catch. Fixed in the latest push — added a `procExited` flag set by a `proc.on("exit", ...)` listener before `proc.unref()`. The poll loop now breaks immediately on the first 250ms tick after the process dies, so fast-fail cases (port conflict, missing binary) exit in ~250ms rather than the full 15s. The 15s deadline still applies for the slow-start case it was introduced for.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3c32d539`

packages/cli/src/features/daemon.ts
NicholaiVogel: Intentional — the semantics are different here. In `togglePipelinePause`, the config file has already been written to disk before the daemon restart is attempted, so the user's primary operation (pause/resume the pipeline) has been completed. A daemon restart failure at that point is a secondary concern. The spinner message reflects this: `"Config updated, but daemon failed to restart"` vs. `"Failed to start daemon"`. Exiting non-zero there would misrepresent the outcome — the user can always follow up with `signet daemon restart`. The `doStart`/`doStop`/`doRestart` commands are purely lifecycle operations, so failure there is total and warrants exit 1.

PR #446, opened 2026-04-02
fix(cli): auto-migrate legacy-only OpenClaw runtime

### LINKED ISSUE #434: feat(cli): auto-migrate legacy OpenClaw hook users to plugin runtime path
## Summary

Users who ran `signet setup` on older versions (before the plugin adapter existed) are permanently stuck on the legacy hook path with no automatic migration. `signet sync` checks `getConfiguredRuntimePath()` and skips plugin installation when it returns `"legacy"`.

## Problem

1. Early adopter runs `signet setup` on e.g. v0.76 — only legacy hooks exist, config gets `hooks.internal.entries.signet-memory.enabled: true`
2. User upgrades to v0.91+ which ships the plugin adapter
3. `signet sync` runs `ensureOpenClawPluginPackage()` but it bails immediately because `getConfiguredRuntimePath() !== 'plugin'`
4. User has no idea they're missing full lifecycle capture (session-start, session-end, auto-recall, etc.)
5. Only way to fix: manually run `signet setup --harness openclaw --openclaw-runtime-path plugin` — but users don't know they need to

## Impact on affected user

- **149 OpenClaw sessions (194MB)** over ~6 weeks went uncaptured
- 0 conversations recorded from the primary harness
- Only manual `/remember` commands worked (legacy path)
- Claude Code sessions captured fine (4 lifecycle hooks) while OpenClaw sessions were silently dropped

## Suggested fix

`signet sync` (or a one-time migration in `signet setup`) should detect:
1. OpenClaw is a configured harness
2. Legacy hook is enabled but plugin is not configured
3. Prompt or auto-migrate to plugin path

A `signet doctor` or `signet health` check that warns about this would also help.

## Environment

- Original setup: ~v0.76
- Current: v0.91.3
- macOS, OpenClaw as primary harness

NicholaiVogel:
## Summary
- auto-migrate legacy-only OpenClaw configs to the plugin runtime path during `signet sync`
- surface legacy-only vs dual-runtime OpenClaw states in status/doctor
- add regression tests for connector runtime detection, doctor warnings, and sync migration
- close the remaining review gaps around OpenClaw runtime aggregation and test env cleanup
- add the required planning stub and docs updates for the incident guardrail

## Changes
- aggregate OpenClaw runtime detection across all discovered config files so mixed plugin/legacy installs now report `dual`
- restore `OPENCLAW_CONFIG_PATH` test cleanup with `delete` semantics and document the required Biome suppression
- add a regression test covering split-config dual runtime detection across `OPENCLAW_CONFIG_PATH` and `CLAWDBOT_CONFIG_PATH`

## Type
- [x] `fix` — bug fix
- [x] `test` — test coverage

## Packages affected
- [x] `@signet/connector-*`
- [x] Other: docs/spec metadata already included in this PR

## Screenshots

N/A

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing
- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Commands run:
- `bun run --filter '@signet/connector-openclaw' build`
- `bun test packages/connector-openclaw/test/index.test.ts packages/cli/src/features/health.test.ts packages/cli/src/features/sync.test.ts`
- `bun run typecheck`
- `bun run lint` *(fails on pre-existing repo-wide Biome diagnostics outside this PR)*
- `bunx biome check packages/cli/src/features/health.test.ts packages/connector-openclaw/src/index.ts packages/connector-openclaw/test/index.test.ts`

## AI disclosure

## Notes
- `bun run lint` was re-run and still reports long-standing repo-wide Biome diagnostics outside this PR's diff (currently 1863 errors / 173 warnings); the touched files pass targeted Biome checks locally.
- No admin or mutation endpoints were changed in this PR.
- No database migrations were changed in this PR.

Closes #434


BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): auto-migrate legacy-only OpenClaw runtime` (commit `2b88256b`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2b88256b`

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): auto-migrate legacy-only OpenClaw runtime` (commit `2b88256b`) now and I'll follow up shortly with feedback.

packages/cli/src/features/health.ts
```diff
@@ -1,6 +1,4 @@
-import { existsSync, readFileSync } from "node:fs";
-import { homedir } from "node:os";
-import { join } from "node:path";
+import { OpenClawConnector, type OpenClawRuntimeState } from "@signet/connector-openclaw";
```
PR-Reviewer-Ant, line 1: **Blocking — missing import causes compile error.** The hunk removes `import { join } from 'node:path'` along with the other three node imports, but `join` is still used further down the file at `Database(join(basePath, 'memory', 'memories.db'), { readonly: true })`. `existsSync`, `readFileSync`, and `homedir` were only referenced inside the now-deleted `detectOpenClawDualSystem()`, so removing them is correct. `join` was also used outside that function and must be kept. Add back `import { join } from 'node:path'`.

packages/cli/src/features/health.test.ts
```diff
@@ -296,6 +296,112 @@ describe("status report openclaw backup risk", () => {
...
 
+describe("status report openclaw runtime", () => {
+	it("reports legacy-only runtime when only the hook path is enabled", async () => {
+		const root = mkdtempSync(join(tmpdir(), "health-runtime-"));
+		const workspace = join(root, "agents");
+		try {
+			process.env.HOME = root;
+			mkdirSync(workspace, { recursive: true });
+			const cfgPath = join(root, "openclaw.json");
+			writeFileSync(
```
PR-Reviewer-Ant, line 316: **Warning — `process.env.HOME` is never restored after these tests.** All three new tests in `describe('status report openclaw runtime', ...)` set `process.env.HOME = root`, but the `finally` blocks only delete the temp directory; they do not reset `HOME`. The shared `afterEach` only restores `OPENCLAW_CONFIG_PATH`. After the temp dir is deleted, `HOME` still points to a non-existent path, which will corrupt any subsequent test that calls `homedir()`. Compare with `sync.test.ts` (line 12–19 of that file) where `originalHome` is captured and restored in `afterEach`. Apply the same pattern here.

packages/cli/src/features/health.ts
```diff
@@ -68,49 +66,11 @@ interface StatusReport {
 		readonly origin: string | null;
```
PR-Reviewer-Ant, line 67: **Warning — breaking change to `signet status --json` output.** The `StatusReport` interface replaces `openclawDualSystem: boolean` with `openclawRuntime: OpenClawRuntimeState` (a string union). Any external script, CI check, or the daemon-rs shadow (which the AGENTS.md Rust parity rule requires to mirror behavioral changes) that reads `report.openclawDualSystem` from JSON output will silently receive `undefined`. The daemon-rs parity note in AGENTS.md is worth a callout here — if the shadow path serializes status output it will need updating too.

packages/cli/src/features/health.ts
```diff
@@ -141,6 +101,7 @@ export async function getStatusReport(basePath: string, deps: StatusDeps): Promi
 	const git = getGitRemoteState(basePath);
```
PR-Reviewer-Ant, line 103: **Warning — possible regression for clawdbot users.** The removed `detectOpenClawDualSystem()` explicitly checked `process.env.CLAWDBOT_CONFIG_PATH` as a candidate path alongside `OPENCLAW_CONFIG_PATH`. The replacement `new OpenClawConnector().getRuntimeState()` delegates to `getDiscoveredConfigPaths()`, which is not visible in this diff. If `getDiscoveredConfigPaths()` does not honour `CLAWDBOT_CONFIG_PATH`, clawdbot users will no longer get legacy/dual detection in `status` or `doctor`. Verify that `getDiscoveredConfigPaths()` handles this env var, or add it to the discovered paths explicitly.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2b88256b`

[Automated Review — pr-reviewer bot] PR #446 adds auto-migration for legacy-only OpenClaw configs during `signet sync`, surfaces the degraded state in `status`/`doctor`, and ships regression tests. One blocking bug found (missing `join` import), two warnings around test pollution and a breaking JSON API field rename.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The missing `join` import is directly provable from the diff: line 1 of the health.ts hunk removes all three node imports including `import { join } from 'node:path'`, but the full file content (provided in context) still calls `Database(join(basePath, 'memory', 'memories.db'), ...)`. None of the remaining imports re-export `join`. The CLAWDBOT_CONFIG_PATH regression requires reading `getDiscoveredConfigPaths()` in full connector source, which is not included in this diff.

NicholaiVogel:
addressed the review feedback in `7c6c11e`.

- added back the missing `join` import in `packages/cli/src/features/health.ts`
- restored `process.env.HOME` in `packages/cli/src/features/health.test.ts` so the new tests do not leak env state
- kept the new `openclawRuntime` field, but restored `openclawDualSystem` in the status JSON for backward compatibility
- verified the clawdbot concern: `OpenClawConnector.getConfigCandidates()` still honors both `CLAWDBOT_CONFIG_PATH` and `CLAWDBOT_STATE_DIR`, so the switch to `getRuntimeState()` does not drop clawdbot detection

re-ran:
- `bun test packages/connector-openclaw/test/index.test.ts packages/cli/src/features/health.test.ts packages/cli/src/features/sync.test.ts`
- `bun run typecheck`

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7c6c11eb`

packages/cli/src/features/health.test.ts
```diff
@@ -3,11 +3,18 @@ import { spawnSync } from "node:child_process";
...
+const originalHome = process.env.HOME;
 const originalOpenClawConfig = process.env.OPENCLAW_CONFIG_PATH;
 
 afterEach(() => {
+	if (originalHome === undefined) {
+		delete process.env.HOME;
+	} else {
+		process.env.HOME = originalHome;
+	}
+
```
PR-Reviewer-Ant, line 18: The `OPENCLAW_CONFIG_PATH` cleanup still uses `process.env.OPENCLAW_CONFIG_PATH = undefined` (which stores the literal string `"undefined"` in the environment) instead of `delete process.env.OPENCLAW_CONFIG_PATH`. The new `sync.test.ts` added in this same PR correctly uses `delete`. If `originalOpenClawConfig` was unset and a test sets `OPENCLAW_CONFIG_PATH` to a real temp path, after cleanup subsequent tests see `OPENCLAW_CONFIG_PATH = "undefined"`. The connector skips non-existent paths gracefully, so this is unlikely to cause test failures today, but it leaves the variable polluted and is inconsistent with the pattern the author demonstrably knows (sync.test.ts:16). Worth fixing to match the new file.

packages/connector-openclaw/src/index.ts
```diff
@@ -691,12 +693,18 @@ export class OpenClawConnector extends BaseConnector {
 				const pluginSlot =
```
PR-Reviewer-Ant, line 694: `getRuntimeState()` returns `"dual"` only when both plugin and legacy are enabled **within the same config file**. If a user has plugin enabled in `~/.openclaw/openclaw.json` and a legacy hook in `~/.clawdbot/clawdbot.json`, the method returns `"plugin"` (early-return on the first file with plugin) and never flags dual state. The same per-file scope limitation existed in the removed `detectOpenClawDualSystem()`, so this is not a regression from this PR — but the new `"dual"` return type implies cross-config detection that does not actually happen. Consider a brief comment noting the per-file scope so future callers don't assume cross-file dual detection.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7c6c11eb`

[Automated Review — pr-reviewer] PR #446 adds legacy→plugin migration for OpenClaw during `signet sync`, exposes `OpenClawRuntimeState` across status/doctor, and adds regression tests. Implementation achieves the stated goals. One warning-severity finding; no blockers.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The runtime-state detection logic, migration wiring, and test cleanup behavior are all directly readable from the diff and provided file contents. The specific cleanup inconsistency is explicit at health.test.ts afterEach vs sync.test.ts afterEach. No missing runtime traces or cross-module context needed to evaluate the changed paths.

packages/connector-openclaw/src/index.ts
```diff
@@ -691,22 +694,37 @@ export class OpenClawConnector extends BaseConnector {
...
```
PR-Reviewer-Ant, line 694: **Behavioral regression vs. the old `detectOpenClawDualSystem()`: cross-file dual detection now fires where it didn't before.**

The old per-file implementation set `hookEnabled && pluginEnabled` only when BOTH were in the same config file. The new logic accumulates `sawPlugin` and `sawLegacy` across ALL discovered configs (openclaw.json, clawdbot.json, moltbot.json, …) and returns `"dual"` if the two flags surface in different files.

Concrete failure mode: a user who successfully migrated openclaw to the plugin path still has a stale `clawdbot.json` with the legacy hook enabled. `getRuntimeState()` returns `"dual"` → doctor shows an ERROR → `sync.ts`'s `if (state === "legacy")` branch never fires → migration is blocked and the user is stuck with a misleading dual-conflict error.

The cross-file test (`"detects dual runtime state across multiple discovered configs"`) confirms this is intentional, but the consequences for migration eligibility make it worth an explicit decision. If separate OpenClaw-family products (openclaw vs. clawdbot) are treated as independent runtimes, cross-file dual detection is a false positive. If they're treated as one shared namespace, the behavior is correct but the migration path needs to handle the cross-file cleanup rather than simply refusing to proceed.

packages/cli/src/features/sync.ts
```diff
@@ -119,7 +123,20 @@ async function syncHarnessHooks(basePath: string, deps: Deps): Promise<number> {
 	let synced = 0;
 	for (const harness of detectHarnesses()) {
 		try {
-			await deps.configureHarnessHooks(harness, basePath);
+			let runtimePath: "plugin" | "legacy" | undefined;
```
PR-Reviewer-Ant, line 126: **Integration gap: the production `configureHarnessHooks` implementation is not in this diff, so the end-to-end migration is unverifiable.**

The `sync.test.ts` mock manually calls `new OpenClawConnector().install(path, { runtimePath: options?.openclawRuntimePath })` itself — it validates the TYPE CONTRACT (that sync passes `{ openclawRuntimePath: "plugin" }`) but not that the real CLI wiring propagates the option into `OpenClawConnector.install()`.

TypeScript structural subtyping means `bun run typecheck` will PASS even if the real caller's `configureHarnessHooks` has the old two-argument signature and simply ignores `options`. The PR description's claim that `signet sync` now migrates legacy configs would then be false — the option would be dropped silently at the injection boundary.

Please confirm (or include in the diff) the code that wires `configureHarnessHooks` in the actual `signet sync` CLI command, and that it threads `openclawRuntimePath` through to `OpenClawConnector.install()`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `47d40c36`

[Automated Bot Review] PR #446 introduces a behavioral change in dual-runtime detection (cross-file vs. per-file) and has an unverifiable integration gap where the production `configureHarnessHooks` implementation is not shown in the diff, meaning the migration could silently do nothing for real users.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The cross-file dual detection behavioral change is directly provable from the diff (`sawPlugin ||= plugin; sawLegacy ||= legacy` across all discovered configs). The integration gap is a structural concern: the real CLI caller that provides `configureHarnessHooks` is not in this diff. TypeScript's callback parameter compatibility rules mean `bun run typecheck` will pass even if the real caller drops the third `options?` argument entirely, leaving the migration silently inoperative in production. Without seeing the CLI command that wires up `configureHarnessHooks`, that path cannot be confirmed.

packages/cli/src/features/sync.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import { OpenClawConnector } from "@signet/connector-openclaw";
+import { syncTemplates } from "./sync.js";
+
+const originalHome = process.env.HOME;
+const originalOpenClawConfig = process.env.OPENCLAW_CONFIG_PATH;
+
+afterEach(() => {
+	if (originalHome === undefined) {
+		delete process.env.HOME;
+	} else {
```
PR-Reviewer-Ant, line 16: `delete process.env.HOME` and `delete process.env.OPENCLAW_CONFIG_PATH` in `afterEach` are missing the `// biome-ignore lint/performance/noDelete:` comments added in the equivalent `health.test.ts` teardown. This is inconsistent and will cause a lint failure if Biome's `noDelete` rule is enforced in CI.

packages/cli/src/features/sync.test.ts
```diff
@@ -0,0 +1,97 @@
...
+import { join } from "node:path";
+import { OpenClawConnector } from "@signet/connector-openclaw";
+import { syncTemplates } from "./sync.js";
+
+const originalHome = process.env.HOME;
+const originalOpenClawConfig = process.env.OPENCLAW_CONFIG_PATH;
+
+afterEach(() => {
+	if (originalHome === undefined) {
+		delete process.env.HOME;
```
PR-Reviewer-Ant, line 15: Missing `biome-ignore lint/performance/noDelete` comment. This PR explicitly added that suppression to `health.test.ts` (lines 20 and 27 of the changed file) for the identical pattern, citing the reason 'assigning undefined to process.env stores the string "undefined"'. The new `sync.test.ts` uses the same `delete process.env.HOME` and `delete process.env.OPENCLAW_CONFIG_PATH` pattern (lines 13 and 18) without those comments. `packages/cli/src/features/sync.test.ts` was also not included in the `bunx biome check` invocation listed in the PR, and `bun run lint` is unchecked in the PR readiness checklist. This will fail the targeted Biome check required by AGENTS.md.

packages/connector-openclaw/test/index.test.ts
```diff
@@ -361,6 +361,89 @@ describe("OpenClawConnector config patching", () => {
...
+		process.env.OPENCLAW_CONFIG_PATH = openClawConfigPath;
+		process.env.CLAWDBOT_CONFIG_PATH = clawdbotConfigPath;
+
+		const connector = new OpenClawConnector();
+		expect(connector.getRuntimeState()).toBe("dual");
+		expect(connector.getConfiguredRuntimePath()).toBe("plugin");
+	});
+
 	it("adds signet memory plugin to plugins.allow during plugin install", async () => {
 		const configPath = join(tmpRoot, "openclaw.json");
```
PR-Reviewer-Ant, line 472: `process.env.CLAWDBOT_CONFIG_PATH` is set inside the 'detects dual runtime state across multiple discovered configs' test but is never restored. There is no `delete process.env.CLAWDBOT_CONFIG_PATH` in a `finally` block or `afterEach`. If this test runs before other tests that call `getDiscoveredConfigPaths()`, the leaked env var will cause those tests to pick up the temp config file (which is cleaned up by `rmSync` in the `finally` block), resulting in a file-not-found skip or a spurious detection depending on timing.

packages/cli/src/features/sync.ts
```diff
@@ -119,7 +123,20 @@ async function syncHarnessHooks(basePath: string, deps: Deps): Promise<number> {
...
+			if (harness === "openclaw") {
+				const state = new OpenClawConnector().getRuntimeState();
```
PR-Reviewer-Ant, line 128: The migration log message fires and `configureHarnessHooks` is called with `{ openclawRuntimePath: 'plugin' }`, but the actual production binding of `configureHarnessHooks` is not in this diff. The regression test in `sync.test.ts` uses a hand-written mock that correctly passes `options?.openclawRuntimePath` through to `OpenClawConnector().install()`. If the real CLI wiring ignores the third argument (e.g., the production `configureHarnessHooks` only accepts two parameters and discards `options`), the migration message will appear in `signet sync` output but the config will not be patched. Recommend confirming the production caller reads `openclawRuntimePath` from `options`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9946d4f4`

[Automated Review — pr-reviewer bot] PR #446 introduces OpenClaw legacy→plugin auto-migration in `signet sync`, surfaces runtime state in status/doctor, adds regression tests, and updates spec/doc metadata. One blocking issue found: the new `sync.test.ts` omits Biome suppression comments that were explicitly added to `health.test.ts` in the same PR. Two warnings around test env cleanup and a cross-module wiring gap.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The blocking issue is directly verifiable from the diff: `health.test.ts` adds `biome-ignore lint/performance/noDelete` comments on every `delete process.env.*` statement, the PR's own command log runs Biome against `health.test.ts` explicitly, but the new `sync.test.ts` file (not included in that Biome invocation) omits those same suppression comments on identical `delete` statements. The env-leak issue in `connector-openclaw/test/index.test.ts` is visible in the added lines. The production-wiring gap is flagged as a warning because the actual `configureHarnessHooks` binding is not in the diff.

NicholaiVogel:
Addressed the remaining review feedback in `9946d4f`.

- aggregate OpenClaw runtime detection across all discovered config files so split plugin/legacy installs now report `dual`
- add a regression test covering the split-config `OPENCLAW_CONFIG_PATH` + `CLAWDBOT_CONFIG_PATH` case
- restore `OPENCLAW_CONFIG_PATH` cleanup in `health.test.ts` using `delete` semantics so the env var does not leak as the literal string `"undefined"`
- rebase the branch onto the latest `main`

Local verification run:
- `bun run --filter '@signet/connector-openclaw' build`
- `bun test packages/connector-openclaw/test/index.test.ts packages/cli/src/features/health.test.ts packages/cli/src/features/sync.test.ts`
- `bun run typecheck`
- `bunx biome check packages/cli/src/features/health.test.ts packages/connector-openclaw/src/index.ts packages/connector-openclaw/test/index.test.ts`

The PR readiness checklist is now populated and the required checklist check is passing. Remaining GitHub checks are still running.

NicholaiVogel:
Addressed the latest review follow-up in `b5cf2482`.

- document in the production `configureHarnessHooks` path that `sync.ts` can force the OpenClaw plugin migration via `openclawRuntimePath`, while still falling back to discovered runtime when no override is supplied
- mirror the required `process.env` cleanup pattern in `sync.test.ts` for both `OPENCLAW_CONFIG_PATH` and `CLAWDBOT_CONFIG_PATH`, using `delete` semantics plus the targeted Biome suppression

Re-ran local verification:
- `bun test packages/cli/src/features/sync.test.ts packages/connector-openclaw/test/index.test.ts packages/cli/src/features/health.test.ts`
- `bunx biome check packages/cli/src/features/sync.test.ts packages/cli/src/cli.ts packages/connector-openclaw/test/index.test.ts packages/cli/src/features/health.test.ts`
- `bun run typecheck`

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b5cf2482`

```pr-review-json
{}
```

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b5cf2482`

[Automated Bot Review] PR #446 introduces OpenClaw legacy-to-plugin runtime migration. No blocking bugs or security issues found. A few observations worth discussing.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The full diff is present along with the changed file contents for all affected modules. The logic in getRuntimeState(), getConfiguredRuntimePath(), syncHarnessHooks(), and the test assertions are all directly inspectable and verifiable without additional runtime trace or missing module context.

packages/connector-openclaw/src/index.ts
PR-Reviewer-Ant, line 694: The aggregation loop short-circuits to 'dual' as soon as both flags are set, but the old getConfiguredRuntimePath() returned 'plugin' early on the first config with a plugin entry. The new behaviour is correct. One edge case to verify: if a config file has pluginSlot (slots.memory === 'signet-memory-openclaw') but its plugin entry is NOT in entries (i.e. pluginEntry is false and the entry for 'signet-memory-openclaw' is absent), `plugin` is still set to true via `pluginSlot`. That mirrors old behaviour. Confirm the intent is to treat a slot assignment alone (without a matching enabled entry) as an active plugin path — the test fixtures all include both the slot and the entry, so this path is untested.

packages/cli/src/features/sync.ts
```diff
@@ -119,7 +123,20 @@ async function syncHarnessHooks(basePath: string, deps: Deps): Promise<number> {
 	let synced = 0;
```
PR-Reviewer-Ant, line 123: When getRuntimeState() returns 'dual', runtimePath stays undefined, configureHarnessHooks is called with undefined options, and cli.ts falls back to getConfiguredRuntimePath() which maps 'dual' → 'plugin'. The dual state is already an error-level doctor finding, but sync silently re-installs plugin without disabling the residual legacy hook or logging a migration message. If the intent is that sync should also clean up dual-state configs (not just legacy-only), this path is missing. If dual is intentionally out of scope for auto-migration, a brief comment here would make that clear.

packages/cli/src/features/sync.test.ts
```diff
@@ -0,0 +1,107 @@
+import { afterEach, describe, expect, it, mock } from "bun:test";
```
PR-Reviewer-Ant, line 1: The test file is new and well-structured, but bun run lint is unchecked in the PR checklist. The PR notes that targeted Biome checks pass for touched files; however, the repo convention is that lint must pass repo-wide before merge. The PR explicitly acknowledges this deviation, but it should be resolved or a concrete plan noted for follow-up.

packages/connector-openclaw/test/index.test.ts
```diff
@@ -361,6 +361,89 @@ describe("OpenClawConnector config patching", () => {
 		expect(connector.getConfiguredRuntimePath()).toBe("legacy");
```
PR-Reviewer-Ant, line 385: The new tests cover dual (single-config) and dual (cross-config) states but there is no new test for getRuntimeState() returning 'plugin' in isolation (plugin entry present, no legacy hook). The case is implicitly covered by the existing getConfiguredRuntimePath() tests that predate this PR, which now go through getRuntimeState(), but an explicit getRuntimeState() === 'plugin' assertion would complete the suite and guard the new aggregation path directly.

docs/specs/planning/openclaw-legacy-plugin-migration.md
```diff
@@ -0,0 +1,38 @@
+---
```
PR-Reviewer-Ant, line 1: The spec frontmatter has informed_by: [] while the INDEX.md row also shows no informed_by. The guardrail loop convention asks for research provenance on new specs. This is an incident-driven planning stub so the empty list may be acceptable, but if any prior research (e.g. openclaw-hardening or openclaw-workspace-protection research docs) informed the decision to auto-migrate, linking them here would help future maintainers.

NicholaiVogel:
Addressed the newest review observations in `0a20a2eb`.

- documented the intentional `sync` behavior for dual-runtime OpenClaw installs: only legacy-only configs are auto-migrated, while dual-state installs stay visible for manual cleanup
- added a regression test proving `sync` leaves dual-runtime installs on the discovered path instead of forcing another migration
- expanded `OpenClawConnector.getRuntimeState()` coverage with explicit plugin-only assertions, including the slot-without-entry edge case
- added research provenance to the planning stub frontmatter for the incident guardrail

Verification:
- `bun test packages/cli/src/features/sync.test.ts packages/connector-openclaw/test/index.test.ts packages/cli/src/features/health.test.ts`
- `bunx biome check packages/cli/src/features/sync.ts packages/cli/src/features/sync.test.ts packages/connector-openclaw/test/index.test.ts packages/cli/src/cli.ts`
- `bun run typecheck`

Note: `bun scripts/spec-deps-check.ts` still fails on a pre-existing unrelated `recall-confidence-gate` metadata issue outside this PR.

docs/specs/dependencies.yaml
```diff
@@ -238,6 +238,21 @@ specs:
...
+    soft_depends_on: []
+    blocks: []
+    informed_by: []
+    success_criteria:
+      - "Legacy-only OpenClaw configs are auto-migrated to the plugin runtime path during `signet sync`"
+      - "`signet doctor` flags legacy-only OpenClaw runtime state with a remediation command"
+      - "Regression tests cover the upgrade path from legacy-only config to plugin runtime activation"
+    decision: null
+
   - id: dogfood-hardening-2026-03-29
```
PR-Reviewer-Ant, line 258: `informed_by: []` here does not match the spec file frontmatter, which lists two research docs:
```
  - docs/research/technical/notebook-dump-2026-02-25.md
  - docs/research/technical/RESEARCH-OPENCLAW-WORKSPACE-PROTECTION.md
```
Per repo conventions ("Doc/spec drift" failure mode and `bun scripts/spec-deps-check.ts` parity rules), `dependencies.yaml` should carry the same `informed_by` entries as the spec. The validator does not currently enforce this field, but it is the stated source of truth for research provenance.

packages/cli/src/features/sync.ts
```diff
@@ -119,7 +123,26 @@ async function syncHarnessHooks(basePath: string, deps: Deps): Promise<number> {
...
 	for (const harness of detectHarnesses()) {
 		try {
-			await deps.configureHarnessHooks(harness, basePath);
+			let runtimePath: "plugin" | "legacy" | undefined;
+			if (harness === "openclaw") {
+				const state = new OpenClawConnector().getRuntimeState();
```
PR-Reviewer-Ant, line 128: The migration log ("↺ OpenClaw legacy-only config detected, migrating…") is emitted before `configureHarnessHooks` is awaited. If `configureHarnessHooks` throws, the catch block logs an error but the migration announcement has already printed — giving the impression the migration succeeded. The surrounding try/catch is existing code, so this is a pre-existing pattern, but worth noting since this is the first user-visible migration message tied to a mutating operation.

packages/cli/src/features/sync.test.ts
```diff
@@ -0,0 +1,162 @@
...
```
PR-Reviewer-Ant, line 1: `bun run lint` is unchecked in the PR checklist. The author notes only repo-wide pre-existing Biome diagnostics affect the full-repo run, and touched files pass targeted checks. Per AGENTS.md: "Run lint and remove dead vars/imports before review." If any of the new `biome-ignore` suppression comments (health.test.ts, sync.test.ts) masked a real lint issue rather than just the known `noDelete` performance rule, it would not surface. Recommend verifying with `bunx biome check packages/cli/src/features/sync.test.ts packages/connector-openclaw/src/index.ts` before merge.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0a20a2eb`

[Automated Review — pr-reviewer bot] This PR correctly implements legacy-only OpenClaw runtime auto-migration during `signet sync`, aggregated cross-config dual detection, and accompanying doctor/status warnings. Two non-blocking issues found: a `dependencies.yaml` ↔ spec frontmatter parity drift, and an unchecked lint gate in the PR checklist.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The core logic changes are fully visible in the diff. The `getRuntimeState()` aggregation (sawPlugin/sawLegacy accumulation across all discovered configs) is provably correct and replaces the early-return that missed cross-file dual detection. The `getConfiguredRuntimePath()` wrapper correctly maps `dual` → `plugin` for the existing call site in cli.ts:264. Test coverage matches the claimed behavior. The two findings are directly evidenced by lines in the diff.

NicholaiVogel:
Addressed the latest feedback in `8be43b69`.

- synced `docs/specs/dependencies.yaml` with the planning spec so `openclaw-legacy-plugin-migration` now carries the same `informed_by` research provenance as the spec frontmatter
- re-ran `bun run lint` and updated the PR body with the concrete repo-wide failure note (`1863` errors / `173` warnings outside this PR)

Verification:
- `bun run lint` *(still fails on pre-existing repo-wide Biome diagnostics outside this PR)*
- `bun scripts/spec-deps-check.ts` *(still fails on unrelated pre-existing spec metadata issues: `recall-confidence-gate`, `engram-informed-predictor-track`)*

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8be43b69`

Automated review (pr-reviewer bot): PR #446 correctly implements OpenClaw legacy-to-plugin auto-migration in `signet sync`, surfaces the degraded state in `doctor`/`status`, and aggregates runtime detection across all discovered config files. One behavior/comment mismatch was found in the dual-state sync path; everything else is logically sound.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The dual-state comment mismatch is directly provable from the diff: sync.ts line ~138 says dual-state installs are 'left for manual cleanup,' but when configureHarnessHooks is called without openclawRuntimePath for a dual-state install, cli.ts falls through to connector.getConfiguredRuntimePath() returning 'plugin', and connector.install() with runtimePath 'plugin' sets hooks.internal.entries['signet-memory'].enabled = false — confirmed by the sync migration test at sync.test.ts lines 110-116 which asserts exactly this behavior for the legacy case. The core getRuntimeState() aggregation logic is verifiable directly from the diff in index.ts lines 694-707.

packages/cli/src/features/sync.ts
```diff
@@ -119,7 +123,26 @@ async function syncHarnessHooks(basePath: string, deps: Deps): Promise<number> {
...
+				if (state === "legacy") {
+					runtimePath = "plugin";
+					console.log(
+						chalk.yellow(
+							"  ↺ OpenClaw legacy-only config detected, migrating to the plugin runtime path for full lifecycle capture",
+						),
+					);
+				}
+				// Leave dual-state installs visible in doctor/status for manual cleanup.
```
PR-Reviewer-Ant, line 137: The comment says 'Leave dual-state installs visible in doctor/status for manual cleanup' and 'sync only self-heals legacy-only configs and should not silently remove hooks.' But when the runtime state is 'dual', runtimePath stays undefined and configureHarnessHooks is called without options. In the real configureHarnessHooks (cli.ts), this falls through to `connector.getConfiguredRuntimePath() ?? 'plugin'` = 'plugin' for dual-state, and `connector.install` with `runtimePath: 'plugin'` actively sets the legacy hook's `enabled` to `false`. The sync test for the migration path (sync.test.ts line 112) asserts this exact behavior. The result is that a sync run on a dual-state install silently resolves the dual state by disabling the legacy hook — contradicting the stated intent of leaving it for manual cleanup. If the intent is truly to leave dual-state alone, `configureHarnessHooks` should be skipped entirely for `state === 'dual'`, or the comment should be corrected to reflect that sync does resolve dual-state.

packages/connector-openclaw/src/index.ts
PR-Reviewer-Ant, line 694: The new `getRuntimeState()` correctly aggregates plugin/legacy signals across all discovered config paths before the old `getConfiguredRuntimePath()` returned early on the first plugin match. One edge case to be aware of: a config file where `plugins.slots.memory === 'signet-memory'` (the generic slot value, not 'signet-memory-openclaw') is treated as plugin-active. The old code had the same behavior. If a user has that generic slot set but doesn't intend full plugin activation, they'd see 'plugin' or 'dual' state incorrectly. This is a pre-existing limitation inherited from the old code, not introduced here — flagging for awareness.

PR #447, opened 2026-04-02
[codex] Bound MEMORY.md projection and filter temp sessions

NicholaiVogel:
## what changed

This PR hardens the deterministic `MEMORY.md` projection so it stays bounded and readable even when lineage artifacts pile up.

- adds real token-budget enforcement to `packages/daemon/src/memory-lineage.ts` using `js-tiktoken`
- preserves high-signal sections first, then clips oldest ledger rows with an explicit notice
- filters temp and synthetic/test sessions out of projection surfaces
- blocks new temp-session pollution at transcript, summary, fact, DAG, and compaction write paths
- adds a cleanup helper for canonical noise sessions
- updates the rolling-lineage spec contract and dependency/index docs
- adds regression tests for projection budgeting, temp-session cleanup, and noise detection

## why

`MEMORY.md` was blowing up because the deterministic renderer had no real projection budget, and `/tmp/signetai` junk sessions were still being written into canonical lineage. That let low-value artifacts dominate the ledger and push the useful head sections out of view.

## impact

- `MEMORY.md` now stays within a fixed token budget
- useful head sections remain visible under pressure
- temp/test sessions no longer show up in projected working memory
- future temp-session artifacts are rejected before they can poison the visible memory surfaces

## root cause

The previous token limiting lived on an older synthesis config path, not on the deterministic projection path that actually renders `MEMORY.md`. At the same time, temp-session artifacts were still allowed through several ingest and write surfaces, so junk lineage accumulated and kept resurfacing.

## validation

- `bun test packages/daemon/src/memory-head.test.ts packages/daemon/src/memory-lineage.test.ts packages/daemon/src/session-noise.test.ts packages/daemon/src/pipeline/summary-worker.test.ts`
- manually cleaned the live workspace junk lineage and regenerated `~/.agents/MEMORY.md` to confirm `/tmp/signetai` no longer appears and clipping is explicit when needed


BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the changes in `[codex] Bound MEMORY.md projection and filter temp sessions` (commit `57ba0f1e`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `57ba0f1e`

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the changes in `[codex] Bound MEMORY.md projection and filter temp sessions` (commit `57ba0f1e`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `57ba0f1e`

[Automated Review — pr-reviewer bot] PR #447 adds temp-session noise filtering and token-budget enforcement to the deterministic MEMORY.md renderer. The implementation is coherent and the spec/dependency updates are mostly correct, but there are three findings worth addressing: a stale frontmatter success criterion in the spec file, a quadratic tokenization loop under high session volume, and a missing `package.json` diff for the new `js-tiktoken` dependency.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All three findings are directly provable from the changed files. The stale criterion is visible in the `memory-md-rolling-window-lineage.md` frontmatter (unchanged fourth bullet) vs the updated `dependencies.yaml` entry. The O(n²) loop is the `while (kept.length > 0) { … kept = kept.slice(0, -1) }` pattern in `renderLedgerSection`. The missing `package.json` change is an absence of any diff for `packages/daemon/package.json` despite two new top-level imports of `js-tiktoken/lite` and `js-tiktoken/ranks/cl100k_base`.


- docs/specs/complete/memory-md-rolling-window-lineage.md:9 - The spec frontmatter `success_criteria` still contains the old "no clipping" contract:

```
"MEMORY.md renders a strict rolling 30-day UTC ledger with one high-signal sentence per in-window session and no clipping"
```

`dependencies.yaml` was correctly updated to the new projectable/clipping language, but the spec file's own frontmatter was not. The spec-deps-check CI script may not catch intra-file drift, but any reader consulting the spec directly will see a conflicting invariant. AGENTS.md failure mode #4 applies. The fourth criterion should be updated to match the `dependencies.yaml` wording.

packages/daemon/src/memory-lineage.ts
```diff
@@ -1075,16 +1148,70 @@ function buildLedger(agentId: string): { markdown: string; refs: ReadonlyArray<s
...
+): { readonly block: string; readonly refs: ReadonlyArray<string>; readonly count: number } {
+	let kept = [...sessions];
+	while (kept.length > 0) {
+		const clipped = sessions.length - kept.length;
+		const lines =
+			clipped > 0
+				? [
+						`- older ledger rows clipped: kept ${kept.length} of ${sessions.length} in-window sessions within projection budget.`,
+						"",
+						...renderLedgerRows(kept),
```
PR-Reviewer-Ant, line 1175: O(n²) tokenization in `renderLedgerSection`. The loop removes one session per iteration (`kept = kept.slice(0, -1)`) and calls `fitsBudget` — which encodes the full block via `projTok.encode` — on every step. With the 220-session test case that is up to 220 full re-encodes per `renderMemoryProjection` call. Under pathological input (500+ sessions over 30 days) this compounds further.

A binary search over `sessions.length` would reduce encode calls to O(log n) with identical correctness: binary-search the largest prefix that fits the budget, then emit one clipping notice. Alternatively, pre-compute per-row token counts and use a running sum.

packages/daemon/src/memory-lineage.ts
```diff
@@ -2,10 +2,14 @@ import { createHash } from "node:crypto";
 import { existsSync, mkdirSync, readFileSync, readdirSync, renameSync, rmSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { basename, join } from "node:path";
+import { Tiktoken } from "js-tiktoken/lite";
```
PR-Reviewer-Ant, line 5: `js-tiktoken/lite` and `js-tiktoken/ranks/cl100k_base` are imported here and in the test file, but the diff contains no change to `packages/daemon/package.json`. If `js-tiktoken` is not already listed as a direct dependency of `@signet/daemon`, the build will fail for clean installs and in CI. Please confirm the dependency is explicit in the package manifest (not just present in `bun.lock` from a transitive pull).

packages/daemon/src/daemon.ts
```diff
@@ -6214,36 +6215,44 @@ app.post("/api/hooks/compaction-complete", async (c) => {
...
+			sessionKey: body.sessionKey ?? null,
+			sessionId,
+			harness: body.harness,
+		});
+		const summaryId = noise ? null : crypto.randomUUID();
+		if (!noise && summaryId) {
+			getDbAccessor().withWriteTx((db) => {
+				db.prepare(
+					`INSERT INTO memories (
+						id, content, type, importance, source_id, source_type,
```
PR-Reviewer-Ant, line 6238: `summaryId` is set to `null` when `noise` is true, then the write block is guarded by `if (!noise && summaryId)`. The second condition is always true when `!noise` (a UUID is always truthy), making `summaryId` check redundant. No functional impact, but the dual-condition reads as if `summaryId` could be null when `!noise`, which it cannot by construction.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `00548421`

packages/daemon/src/memory-lineage.ts
```diff
@@ -1075,16 +1148,92 @@ function buildLedger(agentId: string): { markdown: string; refs: ReadonlyArray<s
...
+		const kept = sessions.slice(0, best);
+		const block = renderCount(best);
+		const refs = kept
+			.map((session) => session.manifestPath)
+			.filter((path): path is string => typeof path === "string");
+		return {
+			block,
+			refs,
+			count: kept.length,
+		};
```
PR-Reviewer-Ant, line 1222: **[Blocking] Daemon-rs parity not updated.** AGENTS.md hard-requires that any behavioral change to `@signet/daemon` be reflected in `packages/daemon-rs/`. This PR gates compaction writes, transcript artifact writes, and `insertSummaryFacts`/`writeSummaryToDAG` behind `isNoiseSession`, but `packages/daemon-rs/` has no corresponding changes. The Rust shadow daemon will continue writing noise artifacts for temp sessions, causing persistent divergence entries in `shadow-divergences.jsonl` and defeating the cleanup goal for any installation using the Rust path.

packages/daemon/src/memory-lineage.ts
```diff
@@ -1075,16 +1148,92 @@ function buildLedger(agentId: string): { markdown: string; refs: ReadonlyArray<s
 			`- ${session.membershipTs} | session=${session.sessionKey ?? session.sessionId} | project=${session.project ?? "none"} | ${session.sentence} ${links.join(" ")}`.trim(),
 		);
 	}
+	return lines;
+}
+
+function renderLedgerSection(
+	sessions: ReadonlyArray<LedgerSession>,
+	base: ReadonlyArray<string>,
+): { readonly block: string; readonly refs: ReadonlyArray<string>; readonly count: number } {
```
PR-Reviewer-Ant, line 1166: **Binary search assumes monotonicity but `renderCount` is not monotone at the `sessions.length` boundary.** When `count === sessions.length`, the clipping notice is omitted; when `count === sessions.length - 1`, it is added. If the clipping notice adds more tokens than the dropped row (possible for very short session rows, ~20-token notice vs <20-token row), then `renderCount(sessions.length)` fits the budget but `renderCount(sessions.length - 1)` does not. The binary search, finding `sessions.length - 1` over-budget, will terminate with a sub-optimal count and miss that `sessions.length` itself fits. In practice session rows are typically 50+ tokens so this is unlikely to trigger, but it is a latent correctness gap worth a targeted comment in the code.

packages/daemon/src/memory-lineage.ts
```diff
@@ -1208,3 +1360,42 @@ export function removeCanonicalSession(agentId: string, sessionToken: string, re
 		db.prepare("DELETE FROM memory_artifacts WHERE agent_id = ? AND session_token = ?").run(agentId, sessionToken);
```
PR-Reviewer-Ant, line 1380: **`purgeCanonicalNoiseSessions` is implemented but has no production invocation path in this PR.** The function is tested and exported, but neither a daemon API endpoint, a CLI command, nor a startup sweep calls it. Existing workspace noise accumulated before this PR is only cleaned if someone calls this function directly. If the intent is to defer the trigger to a follow-up PR, a brief comment on the export noting the expected caller would prevent it from being overlooked.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `00548421`

[Automated Bot Review — pr-reviewer] PR #447 adds token-budget enforcement and temp/synthetic session filtering to the MEMORY.md deterministic projection. The core logic is sound and the spec/dependency updates are consistent. One blocking gap and three non-blocking issues found.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The daemon-rs parity gap is directly provable from the diff: `packages/daemon-rs/` has no changes, yet `daemon.ts`, `hooks.ts`, and `summary-worker.ts` all add new noise-gated write paths. AGENTS.md explicitly states 'Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`.' The binary search non-monotonicity is derived from reading `renderLedgerSection` logic and the clipping-notice token delta. The misleading log string is directly visible in the `daemon.ts` diff.


- packages/daemon/src/daemon.ts:6316 - **Misleading log message for skipped noise sessions.** `logger.info("hooks", "Compaction summary saved", ...)` fires unconditionally, including when `noise === true` and nothing was written. During debugging, this makes noise-skipped compactions indistinguishable from real writes in the log. The `memoryId: summaryId ?? "skipped-temp-session"` field partially signals this, but the top-level message should branch: e.g. `"Compaction summary saved"` vs `"Compaction summary skipped (noise session)"`.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c2d94a63`

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -21,6 +21,7 @@ const LEDGER_HEADING: &str = "Session Ledger (Last 30 Days)";
 const MEMORY_MD_MAX_TOKENS: usize = 5000;
```
PR-Reviewer-Ant: **Blocking — Rust parity gap: token-budget clipping not implemented.**

`MEMORY_MD_MAX_TOKENS` is defined but is never used to clip ledger rows in the Rust `render_memory_projection` path. The TS side has a full binary-search `renderLedgerSection` that prunes oldest rows when the projection exceeds `MEMORY_PROJECTION_MAX_TOKENS`. The Rust render path gets noise filtering and the purge call, but the clipping logic is absent.

AGENTS.md (Rust parity rule): *"Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`... shadow proxy runs both in parallel and logs divergences."* Without clipping parity, the Rust daemon will still produce unbounded `MEMORY.md` output under pressure, defeating the stated goal for users running the Rust path. Additionally, the TS value is derived dynamically (`Math.max(512, MEMORY_HEAD_MAX_TOKENS - 256)`) while Rust hardcodes `5000`; these need to agree or at least be documented as intentionally different.

packages/daemon/src/memory-lineage.ts
```diff
@@ -1117,12 +1281,12 @@ export function renderMemoryProjection(agentId = "default"): {
 	fileCount: number;
 	indexBlock: string;
```
PR-Reviewer-Ant, line 1290: **Warning — destructive side-effect embedded in the render function.**

`purgeCanonicalNoiseSessions` is called unconditionally at the top of `renderMemoryProjection`. This means every call to render MEMORY.md — including during tests, health checks, or any background refresh — now silently deletes files and DB rows. The render function is no longer read-safe.

Two concrete risks:
1. If `isNoiseSession` contains a bug (e.g. a false-positive on a real session), the error is irreversible and happens on every render cycle, not just once.
2. Callers that invoke `renderMemoryProjection` expecting a pure read will get unexpected write I/O and tombstone entries as a side-effect, complicating debugging and making the function hard to call in read-only contexts.

The purge is better triggered explicitly on a maintenance schedule (e.g., once per daemon restart or via a dedicated cleanup endpoint) rather than as a side-effect of every projection render. The same concern applies to the equivalent call at the top of `render_memory_projection` in `memory_lineage.rs`.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1082,6 +1092,15 @@ export function insertSummaryFacts(
 	job: Pick<SummaryJobRow, "harness" | "project" | "session_key" | "agent_id">,
```
PR-Reviewer-Ant: **Warning — `insertSummaryFacts` noise check omits `sessionId`.**

`insertSummaryFacts` calls `isNoiseSession` without passing `sessionId`:
```ts
isNoiseSession({
  project: job.project,
  sessionKey: job.session_key,
  harness: job.harness,
  // sessionId absent
})
```
The sibling guard in `writeSummaryToDAG` correctly passes `sessionId: job.session_id ?? job.id`. When `project` is null/empty, `isNoiseSession` falls through to the synthetic-prefix check on `sessionKey` and `sessionId`. A session with a synthetic `session_id` but a non-synthetic `session_key` and no project would be caught by `writeSummaryToDAG` but pass through `insertSummaryFacts`, persisting facts that should be suppressed. Add `sessionId: job.session_id ?? job.id` to the `insertSummaryFacts` guard to match `writeSummaryToDAG`.

packages/daemon/src/hooks.ts
```diff
@@ -2576,17 +2577,26 @@ export function handleSessionEnd(req: SessionEndRequest): SessionEndResponse {
```
PR-Reviewer-Ant, line 2633: **Nitpick — LLM extraction still runs for noise sessions.**

The summary job is queued (`counts.0 === 1` asserted in the regression test) even for noise sessions. `processJob` in `summary-worker.ts` correctly discards the results, but the full LLM extraction round-trip (transcript → summary → facts) still executes. If `/tmp/signetai` is active (e.g. a CI runner that keeps firing session-end hooks), this silently consumes LLM quota. Consider gating the job enqueue behind an `isNoiseSession` check in `handleSessionEnd`, or at minimum logging a `debug`-level notice when a job is intentionally discarded by `processJob`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c2d94a63`

[Automated Review — pr-reviewer] PR #447 adds temp-session filtering and token-budget clipping to the MEMORY.md projection path. Two issues warrant attention before merge: a Rust parity gap on the clipping implementation (required by AGENTS.md) and a destructive side-effect embedded in the render path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly traceable in the diff. The Rust constant `MEMORY_MD_MAX_TOKENS` is defined at line 21 of `memory_lineage.rs` but the binary-search clipping block that uses it in the TS is absent from the Rust `render_memory_projection` path — visible by comparing the two render functions. The purge side-effect is explicit at the first line of both `renderMemoryProjection` (TS) and `render_memory_projection` (Rust).

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8b083eeb`

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `11beedff`

You've hit your limit · resets 12am (America/New_York)

packages/daemon/src/memory-lineage.ts
```diff
@@ -19,8 +23,12 @@ const SANITIZER_VERSION = "sanitize_transcript_v1";
 const SENTENCE_VERSION = "memory_sentence_v1";
 const LEDGER_HEADING = "Session Ledger (Last 30 Days)";
 const LOW_SIGNAL_SENTENCES = new Set(["Investigated issue.", "Worked on task.", "Reviewed code."]);
+const PROJECTION_HEADROOM_TOKENS = 256;
+export const MEMORY_PROJECTION_MAX_TOKENS = Math.max(512, MEMORY_HEAD_MAX_TOKENS - PROJECTION_HEADROOM_TOKENS);
+export const NOISE_PURGE_REASON = "automatic projection cleanup for temp/test sessions";
 
 const BASE32 = "abcdefghijklmnopqrstuvwxyz234567";
```
PR-Reviewer-Ant, line 31: `const projTok = new Tiktoken(cl100k_base)` is evaluated at module load time. This is a non-trivial side effect — it processes the entire cl100k_base BPE table and allocates a large typed array. AGENTS.md explicitly requires effect-free module scope. The Rust counterpart correctly uses `OnceLock` / a `memory_md_tokenizer()` accessor for lazy init. The TS side should do the same: either a module-level `OnceLock`-style pattern (`let projTok: Tiktoken | null = null; function getTok() { ... }`) or, at minimum, a lazily-initialised closure. As-is this adds measurable cold-start cost to any import of `memory-lineage`.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1079,9 +1089,19 @@ async function scoreContinuity(
 
 export function insertSummaryFacts(
 	accessor: DbAccessor,
-	job: Pick<SummaryJobRow, "harness" | "project" | "session_key" | "agent_id">,
+	job: Pick<SummaryJobRow, "harness" | "project" | "session_key" | "session_id" | "id" | "agent_id">,
```
PR-Reviewer-Ant: `insertSummaryFacts` is an exported function whose `job` Pick was widened from `{ harness, project, session_key, agent_id }` to also require `session_id` and `id`. This is a breaking change to the public API of the module. The two test call sites are updated, and the primary production caller (`processJob`) passes a full `SummaryJobRow` so it is fine. However, if any other package or integration test constructs a partial job object for this function and is not in the diff, the build will fail. Worth confirming no other partial-pick callers exist before merge.

packages/daemon/src/hooks.ts
```diff
@@ -3311,7 +3340,9 @@ export function handleSynthesisRequest(
 	void _sinceTimestamp;
 	void _maxTokens;
```
PR-Reviewer-Ant, line 3399: `purgeCanonicalNoiseSessions` is now called on every synthesis request (`handleSynthesisRequest`). Each call does a full `SELECT … FROM memory_artifacts WHERE agent_id = ?` scan and potentially issues tombstone writes. For agents with large artifact tables this is O(N) on every context refresh. The Rust side anchors the purge inside `write_memory_projection`, which is called at every session-end and compaction, giving it similar frequency. Consider rate-limiting the purge (e.g., at most once per N seconds or once per daemon restart) or moving it to a background maintenance pass, rather than running it synchronously on the hot path of every synthesis call.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -941,6 +978,85 @@ pub fn reindex_memory_artifacts(
...
+pub fn purge_canonical_noise_sessions(
+    conn: &Connection,
+    root: &Path,
+    agent_id: &str,
+    reason: &str,
+) -> Result<usize, String> {
+    let mut stmt = conn
+        .prepare(
+            "SELECT DISTINCT session_token, session_id, session_key, project, harness
+             FROM memory_artifacts
```
PR-Reviewer-Ant, line 1047: Minor parity difference vs TS: `purge_canonical_noise_sessions` uses `SELECT DISTINCT session_token, session_id, session_key, project, harness` which deduplicates on the combination of all five columns. If one session token has artifacts with different `session_key` or `project` values (e.g. a NULL vs a real value across transcript/summary rows), the same `session_token` can appear multiple times in `rows`. The TS implementation uses a `seen: Set<string>` keyed on `session_token` to prevent this. The outcome is not incorrect — the tombstone upsert is idempotent and the DELETE is a no-op after the first — but it causes redundant work and the parity gap could mask future divergences. Consider adding a `GROUP BY session_token` or a Rust-side `seen` HashSet.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `11beedff`

[Automated Review — pr-reviewer bot] PR #447 introduces noise-session filtering and fixed-budget ledger projection. The logic is sound and tests are comprehensive. Three items warrant discussion before merge: a module-scope side effect that violates the repo's style contract, a breaking change to an exported function signature, and a per-call purge scan that runs on every synthesis request.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The core logic in session-noise.ts, memory-lineage.ts, and the Rust mirror is fully visible and consistent. The binary search monotonicity argument and the spec contract change are verifiable from the diff. The medium confidence comes from one missing artifact: the full list of callers of `insertSummaryFacts` in summary-worker.ts. The signature Pick type was widened to include `session_id` and `id`; if any call site outside the diff does not supply those fields the build fails silently in CI. The diff shows two test sites updated but does not show whether `processJob` (the primary non-test caller visible in the diff at ~L541) passes a full `SummaryJobRow` — it does, but we cannot rule out other partial-pick callers in the wider codebase.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -2109,6 +2098,12 @@ pub async fn compaction_complete(
...
     let project: Option<String> = meta["project"].as_str().map(str::to_string);
+    let noise = is_noise_session(
+        project.as_deref(),
+        Some(session_id.as_str()),
+        session_key.as_deref(),
+        Some(harness),
+    );
 
     // Step 2: DB ingest. Artifact is committed above so this is recoverable
     // on failure. idempotency_key=session_id ensures retries don't create
```
PR-Reviewer-Ant: **Blocking — JS/Rust parity gap: compaction artifact file is still written for noise sessions in the Rust daemon.**

The `noise` check is inserted at Step 2 (DB ingest). The comment directly above — "Artifact is committed above so this is recoverable on failure" — confirms that the Step 1 artifact file write (`write_compaction_artifact` equivalent) executes unconditionally before this block, so noise sessions still produce compaction artifact files on disk.

In the JS daemon (`daemon.ts`), the `writeCompactionArtifact` call is inside `if (!noise) { ... }` and is fully skipped for noise sessions.

This violates AGENTS.md's Rust parity rule: "Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`." The shadow divergence log will flag every noise-session compaction as a divergence. The artifact files also accumulate on disk with no corresponding DB record, creating a permanent orphan state that `purge_canonical_noise_sessions` cannot clean (it queries `memory_artifacts`, which is empty for these sessions).

Fix: move the noise check to before the Step 1 artifact write, or gate the artifact write call on `!noise`.

packages/daemon/src/hooks.ts
```diff
@@ -2657,6 +2669,23 @@ export function handleSessionEnd(req: SessionEndRequest): SessionEndResponse {
 	// Queue for async processing by the summary worker instead of
```
PR-Reviewer-Ant, line 2723: **Blocking — JS/Rust parity gap: session state not cleaned up before noise early-return.**

The JS early-return for noise sessions does not clean up session, continuity, or dedup state. The Rust `session_end` explicitly calls `state.sessions.release`, `state.continuity.clear`, `state.dedup.clear_session_start`, and `state.dedup.clear` before its equivalent return. Any cleanup that the JS path would have executed after this point (dedup flushing, continuity clearing, sessions release) is silently skipped.

This can leave stale entries in the in-memory session tracker, dedup map, and continuity state for noise sessions, which may cause false "session already active" conflicts or dedup misses for subsequent real sessions that reuse the same session key.

Fix: add the equivalent state cleanup calls (matching what the Rust path already does) before the early return.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -941,6 +979,101 @@ pub fn reindex_memory_artifacts(
     Ok(())
 }
```
PR-Reviewer-Ant, line 981: **Warning — `PROJECTION_PURGE_SEEN` static has no reset path for Rust tests.**

The JS side exports `resetProjectionPurgeState()` and tests call it in `beforeEach`/`afterEach` to ensure each test starts with a clean purge slate. The Rust `PROJECTION_PURGE_SEEN` is a process-global `OnceLock<Mutex<HashSet>>` with no exported reset. The `projection_clips_older_ledger_rows_within_budget` test calls `write_memory_projection` which triggers `should_purge_projection_noise`. If `temp_root("projection-ledger-budget")` produces a deterministic (fixed) path and the test binary is re-executed in the same process (e.g. via `cargo test -- --test-threads 1` repeated runs or if a test framework reruns failures), the purge will already be marked seen and will not run, causing the `assert!(!rendered.content.contains("/tmp/signetai"))` assertion to fail.

Consider exposing a `#[cfg(test)] pub fn reset_projection_purge_seen()` or using test-unique paths that cannot collide across runs.

packages/daemon/src/memory-lineage.ts
```diff
@@ -860,13 +900,16 @@ function readThreadHeads(agentId: string): ReadonlyArray<{
 	readonly latest_at: string;
 	readonly sample: string;
 	readonly node_id: string;
+	readonly project: string | null;
```
PR-Reviewer-Ant, line 904: **Warning — `memory_thread_heads` query now selects `project`, `session_key`, `harness` with no migration shown.**

The pre-PR query selected 5 columns (`label, source_type, latest_at, sample, node_id`). This PR expands it to 8 by adding `project`, `session_key`, `harness`. The diff does not include a migration adding these columns to `memory_thread_heads`. If these columns were added in a prior migration that is not visible here, this is fine. If they are new, any user upgrading without the migration will silently receive `[]` from `readThreadHeads` (the try/catch swallows the schema error), making the Thread Heads (Tier 2) section of MEMORY.md permanently empty until a manual migration or fresh install.

The identical expansion in `packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs` carries the same risk.

Please confirm whether a migration for these columns already exists (e.g. migration 054 or earlier), or add one.

docs/specs/complete/memory-md-rolling-window-lineage.md
```diff
@@ -201,7 +201,10 @@ At render time `t_now` in UTC:
 - `membership_ts = ended_at` when present, else `captured_at`
```
PR-Reviewer-Ant, line 201: **Warning — spec validation test list not updated to match new clipping behavior.**

The success criteria and the prose contract were updated correctly. However, the "Validation and regression tests" section (item 1) still reads: "1,500-session window test (50/day x 30) yields 1,500 ledger rows." The corresponding code test was renamed to "projection clips older in-window sessions once the ledger exceeds the projection budget" and now asserts clipping occurs and count < 1500. The spec document and the implementation are now inconsistent. Per AGENTS.md recurring failure mode #4, doc/spec drift must be fixed in the same PR.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8df4d4a9`

[Automated Review — pr-reviewer bot] PR #447 adds noise-session filtering and token-budget clipping to MEMORY.md projection. The feature logic is correct and well-tested, but there are two concrete JS/Rust parity gaps that introduce inconsistent runtime behavior, plus two weaker warnings around test isolation and schema drift.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both blocking issues are directly provable from the diff. (1) The Rust compaction_complete handler comment 'Artifact is committed above so this is recoverable' sits immediately above the noise check insertion, confirming the Step 1 artifact write is unconditional while JS places it inside `if (!noise)`. (2) The JS early-return at hooks.ts ~line 2680 precedes the enqueueSummaryJob call and any later cleanup, while the Rust session_end explicitly calls sessions.release, continuity.clear, dedup.clear_session_start, and dedup.clear before its equivalent return. Both gaps are observable from the diff alone.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1351,11 +1319,18 @@ pub async fn session_end(
...
+        project.as_deref(),
+        Some(session_id.as_str()),
+        Some(session_key.as_str()),
+        Some(harness),
+    );
+
     // Canonical artifact always written before pipeline gates — it is the
     // lineage source of truth regardless of pipeline_enabled or shadow_mode.
     // This matches compaction_complete, which also writes artifacts unconditionally
     // so manifests and backlinks never reference transcripts that don't exist.
```
PR-Reviewer-Ant, line 1362: **Bug — orphaned `session_transcripts` rows for noise sessions.**

The noise early-return at line ~1332 calls `state.sessions.release`, `state.continuity.clear`, and `state.dedup.clear*`, then returns `200 OK` — but it never executes the `DELETE FROM session_transcripts WHERE session_key = ?1 AND agent_id = ?2` that appears later in the non-noise pool write. Any noise session that had already written a `session_transcripts` row (e.g., a `/tmp/signetai` session that fired a mid-session checkpoint) will leave that row behind indefinitely. `purge_canonical_noise_sessions` only touches `memory_artifacts`, not `session_transcripts`, so the rows are unreclaimed. At scale this silently inflates the table.

Fix: issue the `DELETE FROM session_transcripts` inside the noise early-return block, or add `session_transcripts` to the purge function's scope.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -941,6 +979,111 @@ pub fn reindex_memory_artifacts(
...
+    format!("{}\u{0}{agent_id}", root.display())
+}
+
+fn should_purge_projection_noise(root: &Path, agent_id: &str) -> Result<bool, String> {
+    let seen = PROJECTION_PURGE_SEEN.get_or_init(|| Mutex::new(HashSet::new()));
+    let mut guard = seen
+        .lock()
+        .map_err(|_| "projection purge state lock poisoned".to_string())?;
+    Ok(guard.insert(projection_purge_key(root, agent_id)))
+}
```
PR-Reviewer-Ant, line 992: **Bug — concurrent Rust test interference via process-global `PROJECTION_PURGE_SEEN`.**

`reset_projection_purge_state()` calls `guard.clear()` on the process-global `Mutex<HashSet>`. Rust tests run in parallel threads by default. If two tests share the same process (they do) and both call `reset_projection_purge_state()` while another test is mid-execution — specifically between the `should_purge_projection_noise` insert and the `purge_canonical_noise_sessions` call — the second test's clear can silently allow the purge to re-run in the first test's workspace, or suppress it entirely.

The unique temp dirs make this rare in practice, but any future test that reuses `agent_id = "default"` with a deterministic workspace path would be reliably affected. Fix with `#[serial]` (via `serial_test` crate) on affected tests, or replace the global with a per-test injected state.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -941,6 +979,111 @@ pub fn reindex_memory_artifacts(
...
+    let mut guard = seen
+        .lock()
+        .map_err(|_| "projection purge state lock poisoned".to_string())?;
+    guard.clear();
+    Ok(())
+}
+
+pub fn purge_canonical_noise_sessions(
+    conn: &Connection,
+    root: &Path,
```
PR-Reviewer-Ant, line 1040: **Fragile — `purge_canonical_noise_sessions` classifies by first-row metadata for a `session_token`.**

The query uses `SELECT DISTINCT` on all five columns, which returns one row per unique `(session_token, session_id, session_key, project, harness)` tuple — not one row per `session_token`. The Rust-level `seen` set deduplicates by `session_token`, so the first row in SQL result order determines whether the whole session is treated as noise. Since SQL provides no ordering guarantee without `ORDER BY`, if a single `session_token` has rows with a `/tmp/` project AND a non-`/tmp/` project (possible after a partial in-flight write), classification is undefined.

For the `/tmp/`-project primary use case this is harmless (project is consistent per session), but for the synthetic-prefix path (`session_key = "test-..."`) a legitimate production session whose `session_key` happens to share a token with a legacy test row could be wrongly purged. Fix: `SELECT DISTINCT session_token, MIN(project) as project, ... GROUP BY session_token` to pick a deterministic row, or add `ORDER BY session_token, captured_at DESC` and let the Rust dedup take the newest row.

packages/daemon/src/memory-lineage.ts
```diff
@@ -1208,3 +1382,59 @@ export function removeCanonicalSession(agentId: string, sessionToken: string, re
 		db.prepare("DELETE FROM memory_artifacts WHERE agent_id = ? AND session_token = ?").run(agentId, sessionToken);
```
PR-Reviewer-Ant, line 1380: **Same fragile purge classification pattern as the Rust side (see `purge_canonical_noise_sessions`).**

The TS `purgeCanonicalNoiseSessions` reads all rows without `ORDER BY`, then deduplicates by `session_token` using a `Set`. The first row for a given `session_token` decides noise classification. Since `memory_artifacts` has no guaranteed row order without explicit ordering, the result is non-deterministic for tokens with mixed project metadata. Recommend adding `ORDER BY session_token, captured_at DESC` to the query so the most-recent row wins.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5335ae36`

[Automated Review — pr-reviewer bot] PR #447 adds temp/test-session noise filtering and token-budget enforcement for MEMORY.md projection. The core logic is sound but there are two medium-confidence bugs and one fragile pattern worth addressing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The session_transcripts leak (item 1) is directly visible in the Rust hooks.rs diff: the noise early-return block clears in-memory state but the SQL DELETE FROM session_transcripts that runs in the non-noise path is unreachable. The ordering-dependent purge classification (item 2) is directly visible in the query + iteration pattern in both language implementations. The Rust parallel-test race (item 3) depends on Rust's default test execution model, which is well-established. The one unknown is whether session_transcripts rows are cleaned elsewhere in the daemon startup flow — that path is not in the diff.


- packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs:1327 - **Minor — `SELECT DISTINCT` on all five columns in `purge_canonical_noise_sessions` does not deduplicate by `session_token` alone.**

The intent is to visit each logical session once, but `DISTINCT (session_token, session_id, session_key, project, harness)` still emits multiple rows per `session_token` when other fields differ. The Rust-level `seen` set is the real dedup. The `DISTINCT` adds query overhead without providing the intended guarantee. Consider `SELECT DISTINCT session_token, session_id, session_key, project, harness FROM ... GROUP BY session_token ORDER BY captured_at DESC` or remove the misleading `DISTINCT`.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -941,6 +979,133 @@ pub fn reindex_memory_artifacts(
...
+            .or_default()
+            .push(row);
+    }
+
+    let mut removed = 0usize;
+    for (session_token, group) in groups {
+        if !is_noise_artifact_group(&group) {
+            continue;
+        }
+        let mut path_stmt = conn
```
PR-Reviewer-Ant, line 1080: **[Blocking — daemon-rs parity divergence]** `purge_canonical_noise_sessions` explicitly calls `fs::remove_file` for each artifact path, but the TypeScript counterpart (`purgeCanonicalNoiseSessions` → `removeCanonicalSession`) only performs DB operations and never touches the filesystem. The AGENTS.md parity rule states: *"Any behavioral change made to `@signet/daemon` must also be reflected in `packages/daemon-rs/`"*. Here the divergence is the other way around: Rust deletes files on disk, TS leaves them as orphaned markdown. The orphaned files do not surface in `MEMORY.md` (noise filtering happens at render time), but they accumulate indefinitely as storage leaks on the TS daemon. Either add file deletion to the TS `removeCanonicalSession`, or remove the `fs::remove_file` loop from Rust and rely solely on the tombstone + DB delete (consistent with the TS approach). Whichever direction is chosen must be reflected in both daemons and their tests.

PR-Reviewer-Ant, line 1080: **[Warning — crash-consistency window]** Inside `purge_canonical_noise_sessions`, the order is: (1) `fs::remove_file` for all paths, (2) `INSERT INTO memory_artifact_tombstones`, (3) `DELETE FROM memory_artifacts`. If the process crashes after step 1 but before step 2, the artifact file is gone from disk but the DB row still refers to it and no tombstone exists. A subsequent re-index would try to read a file that no longer exists. The tombstone `ON CONFLICT … DO UPDATE` makes step 2 idempotent on retry, but the crash window is only closed if step 2 runs before step 1. Recommended fix: perform the DB writes (tombstone + row delete) inside a transaction first, then delete the file. The `let _ = fs::remove_file(...)` already ignores IO errors, so a stale file left by a failed cleanup is the safer failure mode.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -941,6 +979,133 @@ pub fn reindex_memory_artifacts(
...
+}
+
+fn should_purge_projection_noise(root: &Path, agent_id: &str) -> Result<bool, String> {
+    let seen = PROJECTION_PURGE_SEEN.get_or_init(|| Mutex::new(HashSet::new()));
+    let mut guard = seen
+        .lock()
+        .map_err(|_| "projection purge state lock poisoned".to_string())?;
+    Ok(guard.insert(projection_purge_key(root, agent_id)))
+}
+
```
PR-Reviewer-Ant, line 993: **[Warning — no test reset for `PROJECTION_PURGE_SEEN`]** `PROJECTION_PURGE_SEEN` is a process-lifetime `OnceLock<Mutex<HashSet<String>>>`. The TS side added `resetProjectionPurgeState()` specifically to let tests clear this state between runs (`beforeEach`/`afterEach` in `hooks.test.ts`). The Rust tests avoid collision by using unique `temp_root` prefixes per test (so the `(root, agent_id)` key differs), but if a future test reuses the same prefix or runs `write_memory_projection` on the same path, the purge will silently not fire. Adding a `#[cfg(test)]` reset function (or making `PROJECTION_PURGE_SEEN` a thread-local for tests) would match TS behavior and make the invariant explicit.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d4778f7c`

[Automated Review — pr-reviewer bot] PR #447 introduces correct and well-structured noise-session gating and token-budget clipping. Two concrete issues found: a Rust/TS parity divergence on disk cleanup (violates the stated daemon-rs parity invariant) and a crash-consistency window in the Rust purge path where files are deleted before the tombstone is committed. One warning for missing Rust test isolation for the `PROJECTION_PURGE_SEEN` static.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The parity divergence is proven by the diff itself: Rust `purge_canonical_noise_sessions` (memory_lineage.rs ~line 1492) calls `fs::remove_file` for each artifact path before committing the tombstone; TS `purgeCanonicalNoiseSessions` delegates to the pre-existing `removeCanonicalSession`, which per the Changed File Contents only executes two DB statements (tombstone insert + `DELETE FROM memory_artifacts`) and never touches the filesystem. The crash-consistency window is directly readable in the Rust diff ordering. The missing Rust purge-state reset is directly observable from the absence of any equivalent to `resetProjectionPurgeState()` in the Rust test module.


- packages/daemon/src/memory-lineage.ts:1262 - **[Nitpick]** `renderLedgerSection` binary-searches over `[1, sessions.length]`. The full-length case (`count = sessions.length`) produces `renderCount` output with no clipping notice — identical to the pre-check at the top of the function. If `fitsBudget` is deterministic (it is), the pre-check already returns for this case, so the binary search never sees it succeed. The upper bound could be `sessions.length - 1` to make this explicit and save one tokenization call, but this is benign.

packages/daemon/src/session-noise.ts
```diff
@@ -0,0 +1,35 @@
+export interface SessionNoiseSeed {
```
PR-Reviewer-Ant, line 1: This file is imported by hooks.ts, daemon.ts, memory-lineage.ts, and summary-worker.ts as the sole source of `isNoiseSession` and `isTempProject`, and is listed explicitly in the PR test command (`session-noise.test.ts`), but neither the implementation nor the tests appear in the diff. The Rust equivalent (`is_noise_session` in `memory_lineage.rs`) is visible and can be read, but parity with this TS module cannot be verified. All noise-gating decisions in the TS daemon depend on it. This file must be included in the diff before merge.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -1450,36 +1769,28 @@ pub fn render_memory_projection(
     } else {
```
PR-Reviewer-Ant, line 1792: `.take(8)` is added to the global head memory iterator in `render_memory_projection` but is not described anywhere in the PR. The spec states 'high-signal head sections stay intact' under budget pressure, yet this hard-caps the Tier 1 global head at 8 items unconditionally regardless of budget. If the SQL query already limits the result set to 8, the `.take(8)` is redundant. If it does not, this silently drops memories beyond position 8. Either way it is undocumented. Check the corresponding TS path in `memory-lineage.ts` — if it does not apply the same cap, this is a daemon-rs/TS parity divergence that will produce different `MEMORY.md` output between runtimes.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1340,6 +1323,21 @@ pub async fn session_end(
```
PR-Reviewer-Ant, line 1323: The new empty-transcript early-return block now calls `delete_session_transcript` (DELETE FROM session_transcripts WHERE session_key = ? AND agent_id = ?) before returning. This deletes ALL session_transcripts rows for the session, not just a hypothetical final-transcript row. If a session accumulated checkpoint-transcript rows via `upsert_session_transcript` during the session and then ended with an empty final transcript (e.g., a crash or reset), all checkpoint data is now deleted rather than preserved for the transcript-fallback retrieval path described in the lossless-working-memory specs. Before this PR the empty-transcript path returned early without touching `session_transcripts`. Confirm this is intentional and that the lossless-transcript-fallback path does not need checkpoint rows to survive empty-transcript session endings.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -97,6 +101,40 @@ pub struct SummaryArtifactInput {
...
+        let trimmed = value.trim();
+        trimmed.starts_with("/tmp/") || trimmed.starts_with("/private/tmp/")
+    })
+}
+
+fn has_synthetic_prefix(value: Option<&str>) -> bool {
+    value.is_some_and(|raw| {
+        let lower = raw.trim().to_ascii_lowercase();
+        ["test-", "spec-", "fixture-", "synthetic-", "tmp-"]
+            .iter()
```
PR-Reviewer-Ant, line 115: `has_synthetic_prefix` matches session keys/IDs that start with `test-`, `spec-`, `fixture-`, `synthetic-`, or `tmp-`. This only fires when `project` is None or empty (the real-project short-circuit above it returns `false`). However, legitimate long-running sessions with no configured project and session keys that happen to start with these prefixes (e.g., a CI harness that names sessions `test-suite-run-42`, or an agent whose key is `spec-review-agent`) would be silently dropped from all projection surfaces, artifact writes, and fact ingest with no warning to the caller. The 200 OK response and `queued: false` make the data loss invisible. Consider logging a warning when the synthetic-prefix path is hit (as distinct from the /tmp path, which is unambiguous), so operators can detect false positives.

packages/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -1333,7 +1545,100 @@ fn build_ledger(conn: &Connection, agent_id: &str) -> Result<(String, Vec<String
     if sessions.is_empty() {
```
PR-Reviewer-Ant, line 1568: In `render_ledger_section`, the comment correctly notes that `renderCount` is not monotone at the full-count boundary (unclipped output is smaller than clipped output by the length of the clipping notice). The code handles this by checking the unclipped case first, then binary-searching only clipped prefixes. However, `renderCount(0)` when `sessions.len() > 0` produces a notice-only block (no rows), and `renderCount(1)` produces notice + 1 row. The binary search initialises `low = 1` and `high = sessions.len()` — correct. But the loop condition `while low <= high` means when `best` stays 0 and `sessions.len() > 0`, the final fallback path is taken, which returns a notice saying 0 of N kept. Verify the edge case where even a single session row + clipping notice exceeds budget: `best = 0`, `sessions.len() = 1` should produce the '0 of 1' notice. The TS equivalent initialises `let best = 0` with `low = 1, high = sessions.length` — same structure. This appears correct but deserves an explicit test for the single-session-over-budget case.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1193,6 +1218,16 @@ export function insertSummaryFacts(
 // ---------------------------------------------------------------------------
```
PR-Reviewer-Ant, line 1218: `writeSummaryToDAG` now contains its own `isNoiseSession` guard at the top of the function, and the call site in `processJob` also wraps the call in `if (!noise)`. The double-gating is harmless but the internal guard makes `writeSummaryToDAG` a dead-code path when called from the noise branch — which should never happen given the outer guard. Remove one of the two checks to reduce confusion about which is authoritative.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7a1b70aa`

[Automated Review — pr-reviewer] PR #447 adds noise-session filtering and token-budget projection clipping across both the TS and Rust daemon paths. The diff is explicitly truncated at `summary-worker.ts` and `session-noise.ts`/`session-noise.test.ts` — the module that implements `isNoiseSession`/`isTempProject` and is imported by every changed file — is entirely absent from the shown diff. Without it the core gating logic cannot be audited. Within the visible diff, three substantive issues were found: an undocumented hard cap of 8 memories on the Rust global head section that diverges from the TS path; the new empty-transcript handler now deletes all `session_transcripts` rows for a session (including checkpoint rows) rather than skipping them; and binary search in `render_ledger_section` is non-monotone at the full-count boundary in a subtle way that is partially worked around but worth double-checking.

**Confidence:** Low [missing_cross_module_context] - The diff ends with '[diff truncated]' inside `summary-worker.ts`. The module `packages/daemon/src/session-noise.ts` is newly imported by every changed TS file (hooks.ts, daemon.ts, summary-worker.ts, memory-lineage.ts) and is referenced in the PR test command as `session-noise.test.ts`, but neither file appears in the diff. All noise-gating decisions in the TS daemon flow through this module. Without it, whether `isNoiseSession` implements the same contract as the Rust `is_noise_session` (including the temp-project path prefix, synthetic-prefix logic, and harness=='test' branch) cannot be verified. This is the primary implementation gap for the PR's stated goal.

NicholaiVogel:
Addressed the remaining feedback in dc932ccd:

**Crash-consistency fix (Rust `purge_canonical_noise_sessions`)**
The tombstone INSERT and `memory_artifacts` DELETE are now wrapped in a single explicit transaction, matching the TS path which uses `withWriteTx`. Previously the two statements ran independently — a crash between them would leave the DB inconsistent (tombstone present, artifacts still live).

**Test fix: noise session false negative**
The `creates fresh canonical artifacts when distinct session-end events reuse the same sessionKey` test was using `cwd: "/tmp/signetai"`, which `isTempProject()` correctly identifies as a temp path and marks the session as noise — so `handleSessionEnd` returned `{ queued: false }` and the test failed. Changed to `cwd: "/home/user/signetai"` to match the intent of the test (verifying artifact deduplication across real sessions sharing a key).

All 118 TS hooks tests pass. Rust `signet-pipeline` and `signet-daemon` test suites clean.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `dc932ccd`

The background agent confirms: the PR is open on `main`, and a prior automated review round flagged the crash-consistency bug (non-atomic tombstone+artifact delete). That was fixed in the latest commit `dc932ccd` — which is the state my review analyzed. My findings above are new issues not previously surfaced.

No changes to the review output are needed.

Issue #448, opened 2026-04-02
Multi-agent memory extraction and MEMORY.md generation not working for sub-agents


NicholaiVogel:
**Signet version:** 0.86.3
**Environment:** macOS (Apple Silicon), Node v22, Bun
**Related issues:** N/A

---

## Summary

Sub-agents (dot, miles) registered via `signet agent add` have session transcripts being ingested but the memory extraction pipeline produces almost no memories for them, and `MEMORY.md` is never generated for their workspaces. The default agent (Rose) works correctly with 4,788 memories and a fully maintained `MEMORY.md`. This effectively means sub-agents have no working memory, which defeats the purpose of multi-agent deployment.

---

## Observed Behavior

**Memory counts by agent:**
```
default|global|4788
dot|global|1
dot|private|1
main|global|6
miles|global|9
```

Dot has 2 memories (both manual test entries from scoping tests). Miles has 9 (mostly trivial session-start observations). Meanwhile the extraction pipeline has ingested substantial transcript data:

```
Session transcripts by agent_id:
default: 3
dot: 8
miles: 62
```

Miles has 62 session transcripts but only 9 extracted memories — the pipeline is running but barely producing output.

**MEMORY.md generation:**
```sql
SELECT agent_id FROM memory_md_heads;
-- returns only: default
```

No `memory_md_heads` row exists for dot or miles. Their `workspace/AGENTS.md` files exist and are recomposed, but there is no `MEMORY.md` in their workspace directories. The MEMORY.md writer only targets the default agent.

**Additional config inconsistency:**
`agent.yaml` roster only lists miles, not dot — even though dot exists in the `agents` DB table and has session transcripts:
```yaml
agents:
  roster:
    - {"name":"miles","read_policy":"shared"}
```

---

## Expected Behavior

1. The extraction pipeline should produce memories for sub-agents at a comparable rate to the default agent, given equivalent transcript volume.
2. `memory_md_heads` should have rows for each registered agent.
3. Each sub-agent's workspace directory should contain a generated `MEMORY.md` (e.g., `~/.agents/agents/dot/workspace/MEMORY.md`).
4. `agent.yaml` roster should stay in sync with the `agents` DB table (or the DB should be authoritative and the roster should be informational).

---

## Steps to Reproduce

1. Register sub-agents: `signet agent add dot --memory shared` and `signet agent add miles --memory shared`
2. Run sessions through OpenClaw with those agents (Telegram DMs, group chats)
[...]

PR #455, opened 2026-04-03
fix: scope sub-agent memory dedupe and synthesis

NicholaiVogel:
## Summary
Fix the multi-agent memory pipeline so sub-agents can persist their own extracted facts, retain their own synthesized MEMORY head state, and survive CLI/daemon roster round-trips without policy drift.

## Context
Issue #448 reported that OpenClaw sub-agents were ending sessions with almost no retained memories and no per-agent MEMORY head state, even though the default agent accumulated plenty of memory data.

The main failure modes here were:
- duplicate suppression comparing new sub-agent facts against all memories instead of the current agent scope,
- synthesis writes rendering with an agent scope but recording the resulting MEMORY head as `default`,
- CLI roster persistence using a different shape than daemon startup sync expected.

## Changes
- Scope summary-worker duplicate detection by `agent_id`
- Pass the requested `agentId` through scheduled/forced MEMORY synthesis writes
- Add shared roster normalization/build helpers in core
- Update CLI agent roster writes to use canonical nested `memory.read_policy` entries
- Update daemon roster sync to normalize both canonical and legacy roster entries
- Switch shared YAML utilities to the repo YAML serializer/parser so nested roster config round-trips correctly
- Add regression tests for dedupe, synthesis scoping, and roster serialization

Summary fact insertion now calls the duplicate checker with the job's `agent_id`, and the duplicate checker filters candidate matches to the same memory owner before overlap comparison.

Synthesis already rendered with the requested agent scope, but the write path dropped that scope. The write helper now receives the same `agentId`, which keeps `memory_md_heads` updates aligned with the agent being synthesized.

Roster handling is now centralized in core so the CLI and daemon share the same normalization rules for canonical nested entries and legacy flat entries. That preserves backward compatibility while making new writes consistent.

## Use Cases
- A sub-agent can save a fact even if the default agent already has the same fact.
- Forced synthesis after a sub-agent session updates that sub-agent's MEMORY head instead of overwriting `default` state.
- `signet agent add --memory group --group writers` survives daemon restarts without silently degrading back to isolated behavior.

## Testing
```bash
bun test packages/daemon/src/pipeline/summary-worker.test.ts packages/daemon/src/pipeline/synthesis-worker.test.ts packages/core/src/__tests__/identity.test.ts packages/cli/src/commands/agent.test.ts
```

Targeted regression tests pass locally.

I also attempted package-level TypeScript checks with:
- `bunx tsc --noEmit` in `packages/daemon`
- `bunx tsc --noEmit` in `packages/cli`

Those still report unrelated pre-existing workspace/typecheck issues outside this PR.

## Links
- Related issues: #448

### DIFF
--- a/packages/cli/src/commands/agent.test.ts
+++ b/packages/cli/src/commands/agent.test.ts
@@ -1,0 +1,71 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { existsSync, mkdtempSync, readFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { Command } from "commander";
+import { registerAgentCommands } from "./agent";
+
+const prevLog = console.log;
+const tempDirs: string[] = [];
+
+afterEach(() => {
+	console.log = prevLog;
+	while (tempDirs.length > 0) {
+		const dir = tempDirs.pop();
+		if (!dir) continue;
+		rmSync(dir, { recursive: true, force: true });
+	}
+});
+
+describe("registerAgentCommands", () => {
+	test("agent add writes canonical nested roster memory config", async () => {
+		const dir = mkdtempSync(join(tmpdir(), "signet-agent-command-"));
+		tempDirs.push(dir);
+		console.log = () => {};
+
+		const program = new Command();
+		registerAgentCommands(program, {
+			AGENTS_DIR: dir,
+			fetchFromDaemon: async () => null,
+		});
+
+		await program.parseAsync(["node", "test", "agent", "add", "writer", "--memory", "group", "--group", "writers"]);
+
+		const yaml = readFileSync(join(dir, "agent.yaml"), "utf-8");
+		expect(yaml).toContain("- name: writer");
+		expect(yaml).toContain("type: group");
+		expect(yaml).toContain("group: writers");
+		expect(existsSync(join(dir, "agents", "writer", "SOUL.md"))).toBe(true);
+		expect(existsSync(join(dir, "agents", "writer", "IDENTITY.md"))).toBe(true);
+	});
+
+	test("agent list reads canonical nested roster memory config while daemon is offline", async () => {
+		const dir = mkdtempSync(join(tmpdir(), "signet-agent-command-"));
+		tempDirs.push(dir);
+		console.log = () => {};
+
+		const addProgram = new Command();
+		registerAgentCommands(addProgram, {
+			AGENTS_DIR: dir,
+			fetchFromDaemon: async () => null,
+		});
+		await addProgram.parseAsync(["node", "test", "agent", "add", "writer", "--memory", "group", "--group", "writers"]);
+
+		const lines: string[] = [];
+		console.log = (line?: unknown) => {
+			lines.push(String(line ?? ""));
+		};
+		const listProgram = new Command();
+		registerAgentCommands(listProgram, {
+			AGENTS_DIR: dir,
+			fetchFromDaemon: async () => null,
+		});
+
+		await listProgram.parseAsync(["node", "test", "agent", "list"]);
+
+		expect(lines.some((line) => line.includes("Daemon offline"))).toBe(true);
+		expect(lines.some((line) => line.includes("writer") && line.includes("group") && line.includes("writers"))).toBe(
+			true,
+		);
+	});
+});
--- a/packages/core/src/agents.ts
+++ b/packages/core/src/agents.ts
@@ -7,7 +7,15 @@
 
 import { existsSync, mkdirSync, readdirSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
-import type { AgentDefinition } from "./types";
+import type { AgentDefinition, ReadPolicy } from "./types";
+
+export type AgentRosterReadPolicy = "isolated" | "shared" | "group";
+
+export interface NormalizedAgentRosterEntry {
+	readonly name: string;
+	readonly readPolicy: AgentRosterReadPolicy;
+	readonly policyGroup: string | null;
+}
 
 /** Identity files that inherit from agent subdir (fall back to root). */
 const AGENT_SPECIFIC = new Set(["SOUL.md", "IDENTITY.md"]);
@@ -69,6 +77,54 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
 	}
 
 	return result;
+}
+
+function normalizeReadPolicy(readPolicy: unknown, policyGroup: unknown): {
+	readonly readPolicy: AgentRosterReadPolicy;
+	readonly policyGroup: string | null;
+} {
+	if (readPolicy === "shared") return { readPolicy: "shared", policyGroup: null };
+	if (readPolicy === "isolated") return { readPolicy: "isolated", policyGroup: null };
+	if (
+		typeof readPolicy === "object" &&
+		readPolicy !== null &&
+		(readPolicy as { type?: unknown }).type === "group" &&
+		typeof (readPolicy as { group?: unknown }).group === "string"
+	) {
+		return {
+			readPolicy: "group",
+			policyGroup: (readPolicy as { group: string }).group,
+		};
+	}
+	if (readPolicy === "group" && typeof policyGroup === "string") {
+		return { readPolicy: "group", policyGroup };
+	}
+	return { readPolicy: "isolated", policyGroup: null };
+}
+
+export function normalizeAgentRosterEntry(entry: unknown): NormalizedAgentRosterEntry | null {
+	if (typeof entry !== "object" || entry === null) return null;
+	const record = entry as Record<string, unknown>;
+	if (typeof record.name !== "string" || record.name.length === 0) return null;
+	const memory =
+		typeof record.memory === "object" && record.memory !== null ? (record.memory as Record<string, unknown>) : null;
+	const policy = normalizeReadPolicy(memory?.read_policy ?? record.read_policy, record.policy_group);
+	return {
+		name: record.name,
+		readPolicy: policy.readPolicy,
+		policyGroup: policy.policyGroup,
+	};
+}
+
+export function buildAgentMemoryConfig(
+	readPolicy: AgentRosterReadPolicy,
+	policyGroup: string | null,
+): { readonly read_policy?: ReadPolicy } {
+	if (readPolicy === "shared") return { read_policy: "shared" };
+	if (readPolicy === "group" && typeof policyGroup === "string" && policyGroup.length > 0) {
+		return { read_policy: { type: "group", group: policyGroup } };
+	}
+	return { read_policy: "isolated" };
 }
 
 /**
--- a/packages/core/src/agents.ts
+++ b/packages/core/src/agents.ts
@@ -111,7 +111,10 @@ export function normalizeAgentRosterEntry(entry: unknown): NormalizedAgentRoster
 	if (typeof record.name !== "string" || record.name.length === 0) return null;
 	const memory =
 		typeof record.memory === "object" && record.memory !== null ? (record.memory as Record<string, unknown>) : null;
-	const policy = normalizeReadPolicy(memory?.read_policy ?? record.read_policy, record.policy_group);
+	const policy = normalizeReadPolicy(
+		memory?.read_policy ?? record.read_policy,
+		memory?.policy_group ?? record.policy_group,
+	);
 	return {
 		name: record.name,
 		readPolicy: policy.readPolicy,
--- a/packages/core/src/agents.ts
+++ b/packages/core/src/agents.ts
@@ -79,7 +79,10 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
 	return result;
 }
 
-function normalizeReadPolicy(readPolicy: unknown, policyGroup: unknown): {
+function normalizeReadPolicy(
+	readPolicy: unknown,
+	policyGroup: unknown,
+): {
 	readonly readPolicy: AgentRosterReadPolicy;
 	readonly policyGroup: string | null;
 } {
@@ -119,7 +122,7 @@ export function normalizeAgentRosterEntry(entry: unknown): NormalizedAgentRoster
 export function buildAgentMemoryConfig(
 	readPolicy: AgentRosterReadPolicy,
 	policyGroup: string | null,
-): { readonly read_policy?: ReadPolicy } {
+): { readonly read_policy: ReadPolicy } {
 	if (readPolicy === "shared") return { read_policy: "shared" };
 	if (readPolicy === "group" && typeof policyGroup === "string" && policyGroup.length > 0) {
 		return { read_policy: { type: "group", group: policyGroup } };
--- a/packages/core/src/agents.ts
+++ b/packages/core/src/agents.ts
@@ -130,6 +130,9 @@ export function buildAgentMemoryConfig(
 	if (readPolicy === "group" && typeof policyGroup === "string" && policyGroup.length > 0) {
 		return { read_policy: { type: "group", group: policyGroup } };
 	}
+	// CLI validation already rejects `group` without a concrete group name.
+	// Keep the fallback isolated here so malformed callers fail closed instead
+	// of widening access implicitly.
 	return { read_policy: "isolated" };
 }
 
--- a/packages/core/src/migrations/056-agent-scoped-content-hash.ts
+++ b/packages/core/src/migrations/056-agent-scoped-content-hash.ts
@@ -1,0 +1,32 @@
+import type { MigrationDb } from "./index";
+
+function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
+	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<Record<string, unknown>>;
+	if (cols.some((col) => col.name === column)) return;
+	db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${definition}`);
+}
+
+/**
+ * Migration 056: Agent-aware content hash dedupe
+ *
+ * Summary facts and extracted memories may legitimately share identical
+ * content across different agent scopes. Tighten the unique partial index so
+ * duplicate content_hash values are only rejected within the same
+ * (agent_id, scope) tuple.
+ */
+export function up(db: MigrationDb): void {
+	// Defensive repair for stamped-but-partial databases.
+	addColumnIfMissing(db, "memories", "agent_id", "TEXT DEFAULT 'default'");
+	addColumnIfMissing(db, "memories", "scope", "TEXT");
+
+	db.exec("DROP INDEX IF EXISTS idx_memories_content_hash_unique");
+	db.exec(`
+		CREATE UNIQUE INDEX idx_memories_content_hash_unique
+		ON memories(
+			content_hash,
+			COALESCE(NULLIF(agent_id, ''), 'default'),
+			COALESCE(scope, '__NULL__')
+		)
+		WHERE content_hash IS NOT NULL AND is_deleted = 0
+	`);
+}
--- a/packages/core/src/migrations/056-agent-scoped-content-hash.ts
+++ b/packages/core/src/migrations/056-agent-scoped-content-hash.ts
@@ -1,6 +1,7 @@
 import type { MigrationDb } from "./index";
 
 function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
+	// Internal migration helper: callers must pass trusted SQL identifier/type literals.
 	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<Record<string, unknown>>;
 	if (cols.some((col) => col.name === column)) return;
 	db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${definition}`);
--- a/packages/core/src/migrations/056-agent-scoped-content-hash.ts
+++ b/packages/core/src/migrations/056-agent-scoped-content-hash.ts
@@ -1,10 +1,10 @@
 import type { MigrationDb } from "./index";
 
-function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
-	// Internal migration helper: callers must pass trusted SQL identifier/type literals.
-	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<Record<string, unknown>>;
-	if (cols.some((col) => col.name === column)) return;
-	db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${definition}`);
+function ensureMemoriesScopeColumns(db: MigrationDb): void {
+	const cols = db.prepare("PRAGMA table_info(memories)").all() as ReadonlyArray<Record<string, unknown>>;
+	const names = new Set(cols.map((col) => col.name).filter((name): name is string => typeof name === "string"));
+	if (!names.has("agent_id")) db.exec("ALTER TABLE memories ADD COLUMN agent_id TEXT DEFAULT 'default'");
+	if (!names.has("scope")) db.exec("ALTER TABLE memories ADD COLUMN scope TEXT");
 }
 
 /**
@@ -17,8 +17,7 @@ function addColumnIfMissing(db: MigrationDb, table: string, column: string, defi
  */
 export function up(db: MigrationDb): void {
 	// Defensive repair for stamped-but-partial databases.
-	addColumnIfMissing(db, "memories", "agent_id", "TEXT DEFAULT 'default'");
-	addColumnIfMissing(db, "memories", "scope", "TEXT");
+	ensureMemoriesScopeColumns(db);
 
 	db.exec("DROP INDEX IF EXISTS idx_memories_content_hash_unique");
 	db.exec(`
--- a/packages/core/src/yaml.ts
+++ b/packages/core/src/yaml.ts
@@ -14,6 +14,9 @@ export function parseSimpleYaml(text: string): Record<string, unknown> {
 
 /**
  * Format a JavaScript object as YAML.
+ *
+ * `_indent` is retained for internal call-site compatibility, but the
+ * shared YAML library always emits 2-space indentation here.
  */
 export function formatYaml(obj: Record<string, unknown>, _indent = 0): string {
 	return YAML.stringify(obj, {
--- a/packages/core/src/yaml.ts
+++ b/packages/core/src/yaml.ts
@@ -6,10 +6,17 @@ import YAML from "yaml";
 
 /**
  * Parse a YAML string into a JavaScript object.
+ *
+ * Malformed user-owned YAML should degrade to an empty object instead of
+ * propagating parser exceptions into daemon or CLI startup.
  */
 export function parseSimpleYaml(text: string): Record<string, unknown> {
-	const parsed = YAML.parse(text);
-	return typeof parsed === "object" && parsed !== null ? (parsed as Record<string, unknown>) : {};
+	try {
+		const parsed = YAML.parse(text);
+		return typeof parsed === "object" && parsed !== null ? (parsed as Record<string, unknown>) : {};
+	} catch {
+		return {};
+	}
 }
 
 /**
--- a/packages/core/src/yaml.ts
+++ b/packages/core/src/yaml.ts
@@ -1,152 +1,23 @@
 /**
  * @signet/core - YAML utilities
- *
- * Simple YAML parser and formatter for flat/shallow config files.
- * Handles the common patterns used in Signet config files (agent.yaml, config.yaml)
- * without requiring a full YAML library dependency.
- *
- * Limitations:
- * - Supports up to 3 levels of nesting
- * - No support for YAML anchors, aliases, or complex types
- * - Arrays are formatted in block style only
  */
+
+import YAML from "yaml";
 
 /**
- * Parse a simple YAML string into a JavaScript object.
- *
- * Supports:
- * - Comments (line and inline)
- * - Key: value pairs
- * - Up to 3 levels of nesting
- * - Basic type coercion (strings, numbers, booleans)
- * - Quoted strings (strips quotes)
- * - Multiline values with | indicator
+ * Parse a YAML string into a JavaScript object.
  */
 export function parseSimpleYaml(text: string): Record<string, unknown> {
-	const result: Record<string, unknown> = {};
-	const lines = text.split("\n");
-
-	// Stack tracks current nesting context
-	// Each entry: { indent: number, obj: the object at that level, key?: parent key }
-	const stack: Array<{
-		indent: number;
-		obj: Record<string, unknown>;
-		key?: string;
-	}> = [{ indent: -1, obj: result }];
-
-	for (const line of lines) {
-		// Skip empty lines and comments
-		const trimmedLine = line.trim();
-		if (!trimmedLine || trimmedLine.startsWith("#")) continue;
-
-		// Skip YAML document markers
-		if (trimmedLine === "---" || trimmedLine === "...") continue;
-
-		const indent = line.search(/\S/);
-
-		// Pop stack to correct level for this indent
-		while (stack.length > 1 && stack[stack.length - 1].indent >= indent) {
-			stack.pop();
-		}
-
-		const colonIdx = trimmedLine.indexOf(":");
-		if (colonIdx === -1) continue;
-
-		const key = trimmedLine.slice(0, colonIdx).trim();
-		let value = trimmedLine.slice(colonIdx + 1).trim();
-
-		const parent = stack[stack.length - 1].obj;
-
-		// Handle different value cases
-		if (value === "" || value === "|") {
-			// Nested object or multiline start
-			parent[key] = {};
-			stack.push({ indent, obj: parent[key] as Record<string, unknown>, key });
-		} else if (value.startsWith("|")) {
-			// Multiline string - this simplified parser just uses empty string
-			// A full implementation would collect following indented lines
-			parent[key] = "";
-		} else {
-			// Simple value - apply type coercion
-			parent[key] = coerceYamlValue(value);
-		}
-	}
-
-	return result;
+	const parsed = YAML.parse(text);
+	return typeof parsed === "object" && parsed !== null ? (parsed as Record<string, unknown>) : {};
 }
 
 /**
  * Format a JavaScript object as YAML.
- *
- * Supports:
- * - Nested objects
- * - Arrays (block style)
- * - Primitive values (string, number, boolean, null)
  */
-export function formatYaml(obj: Record<string, unknown>, indent = 0): string {
-	const pad = "  ".repeat(indent);
-	let result = "";
-
-	for (const [key, value] of Object.entries(obj)) {
-		if (value === undefined) continue;
-
-		if (Array.isArray(value)) {
-			result += `${pad}${key}:\n`;
-			for (const item of value) {
-				if (typeof item === "object" && item !== null) {
-					result += `${pad}  - ${JSON.stringify(item)}\n`;
-				} else {
-					result += `${pad}  - ${item}\n`;
-				}
-			}
-		} else if (typeof value === "object" && value !== null) {
-			result += `${pad}${key}:\n`;
-			result += formatYaml(value as Record<string, unknown>, indent + 1);
-		} else if (value === null) {
-			result += `${pad}${key}: null\n`;
-		} else if (typeof value === "string") {
-			if (
-				value.includes(":") ||
-				value.includes("#") ||
-				value.includes("\n") ||
-				value.includes('"') ||
-				value.includes("\\") ||
-				value.includes("\t") ||
-				/^\d/.test(value)
-			) {
-				const escaped = value.replace(/\\/g, "\\\\").replace(/"/g, '\\"').replace(/\n/g, "\\n").replace(/\t/g, "\\t");
-				result += `${pad}${key}: "${escaped}"\n`;
-			} else {
-				result += `${pad}${key}: ${value}\n`;
-			}
-		} else {
-			result += `${pad}${key}: ${value}\n`;
-		}
-	}
-
-	return result;
-}
-
-/**
- * Coerce a YAML value string to the appropriate JavaScript type.
- */
-function coerceYamlValue(value: string): unknown {
-	// Strip surrounding quotes
-	const unquoted = value.replace(/^["']|["']$/g, "");
-
-	// Boolean
-	if (unquoted === "true") return true;
-	if (unquoted === "false") return false;
-
-	// Null
-	if (unquoted === "null" || unquoted === "~") return null;
-
-	// Integer
-	if (/^-?\d+$/.test(unquoted)) return parseInt(unquoted, 10);
-
-	// Float
-	if (/^-?\d+\.\d+$/.test(unquoted)) return parseFloat(unquoted);
-
-	// String (return unquoted version)
-	return unquoted;
+export function formatYaml(obj: Record<string, unknown>, _indent = 0): string {
+	return YAML.stringify(obj, {
+		indent: 2,
+		simpleKeys: true,
+	});
 }
--- a/packages/daemon-rs/contracts/schema.sql
+++ b/packages/daemon-rs/contracts/schema.sql
@@ -67,7 +67,7 @@ CREATE TABLE "memories" (
   last_accessed   TEXT,
   access_count    INTEGER DEFAULT 0,
   pinned          INTEGER DEFAULT 0
-, content_hash TEXT, normalized_content TEXT, is_deleted INTEGER DEFAULT 0, deleted_at TEXT, extraction_status TEXT DEFAULT 'none', embedding_model TEXT, extraction_model TEXT, update_count INTEGER DEFAULT 0, idempotency_key TEXT, runtime_path TEXT, source_path TEXT, source_section TEXT, agent_id TEXT DEFAULT 'default', visibility TEXT DEFAULT 'global');
+, content_hash TEXT, normalized_content TEXT, is_deleted INTEGER DEFAULT 0, deleted_at TEXT, extraction_status TEXT DEFAULT 'none', embedding_model TEXT, extraction_model TEXT, update_count INTEGER DEFAULT 0, idempotency_key TEXT, runtime_path TEXT, source_path TEXT, source_section TEXT, scope TEXT, agent_id TEXT DEFAULT 'default', visibility TEXT DEFAULT 'global');
 CREATE INDEX idx_memories_type ON memories(type);
 CREATE INDEX idx_memories_category ON memories(category);
 CREATE INDEX idx_memories_source ON memories(source_type, source_id);
--- a/packages/daemon-rs/contracts/schema.sql
+++ b/packages/daemon-rs/contracts/schema.sql
@@ -163,7 +163,7 @@ CREATE INDEX idx_relations_target
 CREATE INDEX idx_memory_entity_mentions_entity
 			ON memory_entity_mentions(entity_id);
 CREATE UNIQUE INDEX idx_memories_content_hash_unique
-			ON memories(content_hash)
+			ON memories(content_hash, COALESCE(NULLIF(agent_id, ''), 'default'), COALESCE(scope, '__NULL__'))
 			WHERE content_hash IS NOT NULL AND is_deleted = 0
 	;
 CREATE INDEX idx_memories_deleted_at
--- a/packages/daemon/src/hooks.test.ts
+++ b/packages/daemon/src/hooks.test.ts
@@ -50,7 +50,7 @@ describe("writeMemoryMd", () => {
 	afterAll(() => {
 		rmSync(agentsDir, { recursive: true, force: true });
 		if (previousSignetPath === undefined) {
-			process.env.SIGNET_PATH = undefined;
+			Reflect.deleteProperty(process.env, "SIGNET_PATH");
 			return;
 		}
 		process.env.SIGNET_PATH = previousSignetPath;
--- a/packages/daemon/src/hooks.test.ts
+++ b/packages/daemon/src/hooks.test.ts
@@ -1,10 +1,15 @@
-import { describe, expect, it } from "bun:test";
+import { afterAll, afterEach, beforeAll, beforeEach, describe, expect, it } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { closeDbAccessor, getDbAccessor, initDbAccessor } from "./db-accessor";
 import {
 	buildSignetSystemPrompt,
 	normalizeCodexTranscript,
 	normalizeJsonConversationTranscript,
 	normalizeSessionTranscript,
 	queryAnchorsMissingFromRecall,
+	writeMemoryMd,
 } from "./hooks";
 
 describe("buildSignetSystemPrompt", () => {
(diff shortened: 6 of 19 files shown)


BusyBee3333:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: scope sub-agent memory dedupe and synthesis` (commit `bc0bc5e2`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: scope sub-agent memory dedupe and synthesis` (commit `bc0bc5e2`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/synthesis-worker.ts
```diff
@@ -218,7 +218,10 @@ async function runSynthesisWithDeps(
 		const finalText = synthesisData.prompt.trimEnd();
 
 		// Write MEMORY.md via shared helper (handles backup)
-		const writeResult = deps.writeMemoryMd(finalText, { owner: "synthesis-worker" });
+		const writeResult = deps.writeMemoryMd(finalText, {
```
PR-Reviewer-Ant, line 222: `agentId` is now passed to `deps.writeMemoryMd(finalText, { agentId, owner: 'synthesis-worker' })`, but the hooks.ts diff in this PR only changes `isDuplicate` — there is no corresponding change to `writeMemoryMd` to accept or act on `agentId`. In JavaScript an extra object property is silently ignored, so the actual MEMORY.md write and any `memory_md_heads` update will still use the default agent path. The synthesis scoping fix is wired at the call site but not implemented in the write helper. The synthesis tests mock `writeMemoryMd`, so they validate the call shape but not the underlying write behavior. This leaves the second failure mode from issue #448 (synthesis head recorded as `default` after a sub-agent session) unfixed at runtime.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,23 @@
...
- *
- * Simple YAML parser and formatter for flat/shallow config files.
- * Handles the common patterns used in Signet config files (agent.yaml, config.yaml)
- * without requiring a full YAML library dependency.
- *
- * Limitations:
- * - Supports up to 3 levels of nesting
- * - No support for YAML anchors, aliases, or complex types
- * - Arrays are formatted in block style only
  */
```
PR-Reviewer-Ant, line 3: `import YAML from "yaml"` is added here, but there is no corresponding change to `packages/core/package.json` in this diff. If `yaml` is not already a direct dependency of `@signet/core`, the import will fail at runtime in any environment that resolves package boundaries strictly (including `bun build` and consumers of the published package). Please confirm `yaml` is in `packages/core/package.json#dependencies`, or add it in this PR.

packages/daemon/src/pipeline/summary-worker.test.ts
```diff
@@ -114,6 +114,28 @@ describe("insertSummaryFacts", () => {
 		expect(row?.updated_by).toBe(SUMMARY_WORKER_UPDATED_BY);
 	});
 
+	it("scopes duplicate detection to the fact owner's agent", () => {
```
PR-Reviewer-Ant, line 151: The new scoped-deduplication test inserts the existing `default` memory directly into the `memories` table but not into `memories_fts`. If FTS triggers are not active on the in-memory test database (which is typical for `:memory:` + `runMigrations` without explicit trigger setup), `isDuplicate` will return `false` for any content because the FTS MATCH query will produce zero rows. The test then passes (`saved === 1`) vacuously — not because agent scoping works, but because the FTS table has no data at all. Consider either populating `memories_fts` in the test fixture or asserting that a same-agent duplicate is actually suppressed (insert a memory for `agent-a` first, then verify a second identical insert returns `0`).

packages/core/src/agents.ts
```diff
@@ -71,6 +79,54 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
...
+	}
+	return { readPolicy: "isolated", policyGroup: null };
+}
+
+export function normalizeAgentRosterEntry(entry: unknown): NormalizedAgentRosterEntry | null {
+	if (typeof entry !== "object" || entry === null) return null;
+	const record = entry as Record<string, unknown>;
+	if (typeof record.name !== "string" || record.name.length === 0) return null;
+	const memory =
+		typeof record.memory === "object" && record.memory !== null ? (record.memory as Record<string, unknown>) : null;
```
PR-Reviewer-Ant, line 113: `normalizeReadPolicy` handles the legacy group case (`read_policy: "group"`, `policy_group: "writers"`) at line ~110, but there is no regression test for this specific combination. The existing tests cover `shared` (legacy flat) and the canonical nested `{ type: "group", group: "writers" }` form, but a legacy `{ name: "x", read_policy: "group", policy_group: "writers" }` entry that may exist in production `agent.yaml` files produced by older CLI versions is not exercised. A missing test for this case means a future refactor of `normalizeReadPolicy` could silently regress it.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,23 @@
...
-		}
-	}
-
-	return result;
+	const parsed = YAML.parse(text);
+	return typeof parsed === "object" && parsed !== null ? (parsed as Record<string, unknown>) : {};
 }
 
 /**
  * Format a JavaScript object as YAML.
```
PR-Reviewer-Ant, line 23: The `_indent` parameter is kept in the signature but silently ignored — `YAML.stringify` always uses `indent: 2`. Any internal caller that previously relied on the custom indentation level (even if only as a stylistic convention) will now get 2-space output without warning. This is unlikely to cause correctness issues since the parameter was always internal, but renaming it to `_indent` and documenting the no-op would be cleaner.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `bc0bc5e2`

[Automated Review — pr-reviewer bot] This PR fixes three real sub-agent memory bugs (dedupe scoping, synthesis head alignment, roster round-trip). The dedupe and roster changes are sound; however, the synthesis-scoping fix appears incomplete as written, and the yaml library dependency is not confirmed added to core's package.json.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The incomplete synthesis-write scoping is directly provable from the diff: synthesis-worker.ts:221 passes `agentId` to `writeMemoryMd`, but the hooks.ts diff only shows `isDuplicate` being changed — `writeMemoryMd` is not updated. The test mocks `writeMemoryMd` so it cannot catch this gap. The yaml dependency concern requires seeing packages/core/package.json (not in the diff) to confirm; if yaml was not already a direct dep the build breaks.

packages/daemon/src/pipeline/synthesis-worker.ts
NicholaiVogel: Wired this through and pushed the follow-up commit.

- `writeMemoryMd` now accepts/forwards `agentId`: `packages/daemon/src/hooks.ts:3295-3308`
- the underlying head write persists under the requested agent scope: `packages/daemon/src/memory-head.ts:154-205`
- added a direct regression covering the shared helper path (not just the synthesis mock): `packages/daemon/src/hooks.test.ts:29-85`

packages/core/src/yaml.ts
NicholaiVogel: Confirmed and pushed. `yaml` is already a direct dependency in core: `packages/core/package.json:29-32`.

packages/daemon/src/pipeline/summary-worker.test.ts
NicholaiVogel: Good catch. I tightened the regression so it now proves both sides explicitly: same-agent duplicates are suppressed and cross-agent inserts still succeed: `packages/daemon/src/pipeline/summary-worker.test.ts:117-145`.

packages/core/src/agents.ts
NicholaiVogel: Added the missing legacy flat `group` regression case and pushed it: `packages/core/src/__tests__/identity.test.ts:122-128`.

packages/core/src/yaml.ts
NicholaiVogel: Agreed. I renamed the retained compatibility parameter to `_indent` and documented that it is an intentional no-op with the shared YAML library: `packages/core/src/yaml.ts:15-25`.

packages/daemon/src/memory-head.test.ts
```diff
@@ -23,14 +24,19 @@ describe("writeMemoryHead", () => {
...
+	afterEach(() => {
+		closeDbAccessor();
+	});
+
 	afterAll(() => {
 		rmSync(agentsDir, { recursive: true, force: true });
 		if (prevSignetPath === undefined) {
-			delete process.env.SIGNET_PATH;
+			process.env.SIGNET_PATH = undefined;
 			return;
```
PR-Reviewer-Ant, line 40: **Bug: assigning `undefined` to `process.env.KEY` sets it to the string `"undefined"`, not removing it.**

In both Bun and Node.js, `process.env` values are coerced to strings. After teardown the env var will be `SIGNET_PATH="undefined"` instead of unset, which can cause any subsequent test that reads `process.env.SIGNET_PATH` to receive a non-empty, structurally invalid path.

Fix:
```ts
if (prevSignetPath === undefined) {
  delete process.env.SIGNET_PATH;  // ← correct
  return;
}
```
The same issue exists in `hooks.test.ts` at the corresponding `afterAll` teardown.

packages/daemon/src/hooks.test.ts
```diff
@@ -21,6 +26,64 @@ describe("buildSignetSystemPrompt", () => {
...
+		closeDbAccessor();
+	});
+
+	afterAll(() => {
+		rmSync(agentsDir, { recursive: true, force: true });
+		if (previousSignetPath === undefined) {
+			process.env.SIGNET_PATH = undefined;
+			return;
+		}
+		process.env.SIGNET_PATH = previousSignetPath;
```
PR-Reviewer-Ant, line 56: Same `process.env.SIGNET_PATH = undefined` teardown bug as in `memory-head.test.ts`. Should be `delete process.env.SIGNET_PATH` to fully unset the variable instead of leaving it set to the string `"undefined"`.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,26 @@
...
- *
- * Simple YAML parser and formatter for flat/shallow config files.
- * Handles the common patterns used in Signet config files (agent.yaml, config.yaml)
- * without requiring a full YAML library dependency.
- *
- * Limitations:
- * - Supports up to 3 levels of nesting
- * - No support for YAML anchors, aliases, or complex types
- * - Arrays are formatted in block style only
  */
```
PR-Reviewer-Ant, line 3: **Unverified external dependency.** `import YAML from "yaml"` was added but the `package.json` for `@signet/core` is not in the diff. If `yaml` is not already listed as a dependency, this will fail at install/build time. Please confirm `"yaml"` appears in `packages/core/package.json` dependencies (not just devDependencies) and that the version constraint is compatible with the usage here.

packages/daemon/src/hooks.ts
```diff
@@ -516,7 +516,7 @@ function getSessionGapSummary(): string | undefined {
 }
 
 /** Check if content overlaps 70%+ with existing memories via FTS */
-export function isDuplicate(db: Database, content: string): boolean {
+export function isDuplicate(db: Database, content: string, agentId = "default"): boolean {
```
PR-Reviewer-Ant: **Scope of `isDuplicate` default-parameter change.** The new signature `isDuplicate(db, content, agentId = "default")` is correct for the summary-worker call site, but any other existing caller in the daemon that does not pass `agentId` will now silently check duplicates only within the `"default"` agent's memories. The diff does not show all call sites; if `isDuplicate` is called elsewhere (e.g., from the hooks pipeline outside summary-worker), those callers should be audited and updated to pass the correct `agentId`.

packages/daemon-rs/contracts/schema.sql
```diff
@@ -163,7 +163,7 @@ CREATE INDEX idx_relations_target
 CREATE INDEX idx_memory_entity_mentions_entity
 			ON memory_entity_mentions(entity_id);
 CREATE UNIQUE INDEX idx_memories_content_hash_unique
-			ON memories(content_hash)
+			ON memories(content_hash, COALESCE(NULLIF(agent_id, ''), 'default'), COALESCE(scope, '__NULL__'))
```
PR-Reviewer-Ant, line 166: The updated index now references the `scope` column (`COALESCE(scope, '__NULL__')`), but the `memories` table CREATE TABLE statement visible in this file does not include a `scope` column. Migration 033 adds it at runtime, but if daemon-rs parity tests build the schema from this SQL file directly, the index creation will fail with "no such column: scope". The `scope` column should either be present in the CREATE TABLE definition or added via an ALTER TABLE statement before this index in the schema.sql file.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `dbb9e766`

[Automated Review — pr-reviewer bot] PR #455 fixes three distinct sub-agent memory bugs: duplicate suppression scope, synthesis MEMORY-head scope, and roster round-trip shape. The implementation is largely correct and well-tested, but two concrete defects were found: a test-teardown env-var bug that can pollute subsequent tests, and an unverified external dependency introduction.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The test-teardown defect is directly visible at the changed lines in memory-head.test.ts and hooks.test.ts. The yaml-package import is explicit in the yaml.ts diff; the only ambiguity is whether it was already listed in @signet/core's package.json, which is not included in the diff. The isDuplicate default-parameter scope change is readable from the hooks.ts diff combined with the lack of any other call-site update in the diff.

packages/daemon/src/memory-head.test.ts
NicholaiVogel: Fixed in the rebased branch and force-pushed. The teardown now removes the variable instead of assigning the string `"undefined"`: `packages/daemon/src/memory-head.test.ts:36-43`.

packages/daemon/src/hooks.test.ts
NicholaiVogel: Fixed here as well in the rebased branch: teardown now fully unsets `SIGNET_PATH` with `Reflect.deleteProperty(...)`: `packages/daemon/src/hooks.test.ts:50-57`.

packages/core/src/yaml.ts
NicholaiVogel: Confirmed on the pushed branch: `yaml` is already a direct runtime dependency of `@signet/core` at `packages/core/package.json:29-32`, and the current usage remains the same `YAML.parse`/`YAML.stringify` calls in `packages/core/src/yaml.ts:5-25`.

packages/daemon/src/hooks.ts
NicholaiVogel: Addressed in the rebased branch by removing the default entirely. `isDuplicate` now requires an explicit `agentId`: `packages/daemon/src/hooks.ts:519-539`. The only daemon call site is the summary worker, which passes `job.agent_id`: `packages/daemon/src/pipeline/summary-worker.ts:1160-1163`.

packages/daemon-rs/contracts/schema.sql
NicholaiVogel: Confirmed on the rebased branch: the contract schema already includes `scope` in the `memories` table definition at `packages/daemon-rs/contracts/schema.sql:47-70`, and the scoped unique index is created against that same column at `packages/daemon-rs/contracts/schema.sql:165-167`.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,26 @@
...
- * Parse a simple YAML string into a JavaScript object.
- *
- * Supports:
- * - Comments (line and inline)
- * - Key: value pairs
- * - Up to 3 levels of nesting
- * - Basic type coercion (strings, numbers, booleans)
- * - Quoted strings (strips quotes)
- * - Multiline values with | indicator
+ * Parse a YAML string into a JavaScript object.
```
PR-Reviewer-Ant, line 8: `import YAML from "yaml"` adds a new runtime dependency, but `packages/core/package.json` is not in this diff. The original file's top-level comment explicitly stated the custom parser was written to avoid a full YAML library dependency. If `yaml` is not listed in `packages/core/package.json` `dependencies`, the daemon and CLI will fail at require-time on a clean install. Please confirm the dependency is declared.

packages/daemon/src/pipeline/synthesis-worker.test.ts
```diff
@@ -119,9 +121,12 @@ describe("synthesis-worker", () => {
 				reason: undefined,
```
PR-Reviewer-Ant, line 121: The two existing default-agent synthesis tests were updated to use `expect.objectContaining({ owner: "synthesis-worker" })`, which no longer asserts the value of `agentId`. If `agentId` is passed as `undefined` or omitted for the default agent, `writeMemoryMd` would silently fall back to whatever its own default is, potentially writing the head under the wrong scope. Adding `agentId: "default"` (or whatever the expected value is) to these assertions would close the gap and match the rigor of the new agent-a assertion at line 236.

packages/core/src/agents.ts
```diff
@@ -71,6 +79,54 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
...
+	const policy = normalizeReadPolicy(memory?.read_policy ?? record.read_policy, record.policy_group);
+	return {
+		name: record.name,
+		readPolicy: policy.readPolicy,
+		policyGroup: policy.policyGroup,
+	};
+}
+
```
PR-Reviewer-Ant, line 124: `buildAgentMemoryConfig` is typed to return `{ readonly read_policy?: ReadPolicy }` (optional field), but every code path returns a `read_policy`. The optional `?` is misleading and could cause callers to add unnecessary null-guards. Return type should be `{ readonly read_policy: ReadPolicy }`.

packages/core/src/migrations/056-agent-scoped-content-hash.ts
```diff
@@ -0,0 +1,32 @@
+import type { MigrationDb } from "./index";
+
+function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
+	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<Record<string, unknown>>;
+	if (cols.some((col) => col.name === column)) return;
```
PR-Reviewer-Ant: `PRAGMA table_info(${table})` and `ALTER TABLE ${table} ADD COLUMN ${column} ${definition}` interpolate all three parameters directly into SQL strings. Safe as currently called (only literal string constants are passed), but the function signature accepts arbitrary `string` values with no guard. Adding a comment noting this is an internal helper whose arguments must be literals would prevent misuse if the function is ever reused.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `efae7ef0`

[Automated Review — pr-reviewer bot] PR #455 fixes sub-agent memory dedupe scoping, synthesis MEMORY-head attribution, and roster round-trip consistency. The implementation is mostly correct and well-tested. Two items warrant discussion before merge: a dependency addition not visible in the diff and a weak test assertion that could mask a regression in the default-agent synthesis path.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The core logic changes (isDuplicate scope, synthesis agentId threading, normalizeAgentRosterEntry) are all visible and traceable in the diff. The missing context is: (1) packages/core/package.json is not in the diff, so whether the `yaml` package was already declared as a dependency cannot be confirmed; (2) the daemon's /api/agents POST handler is not in the diff, so the CLI→daemon POST format mismatch cannot be fully verified.


- packages/cli/src/commands/agent.ts:159 - The `POST /api/agents` body still uses the legacy flat `read_policy` / `policy_group` shape, while the `addToRoster` write path (line 54) now uses the canonical nested `memory: { read_policy: ... }` shape. If the daemon's `/api/agents` handler is ever updated to expect canonical shape, this divergence will silently drop group/shared policy for daemon-registered agents. Consider aligning the POST body with the new canonical format, or add a comment noting the deliberate mismatch.

packages/core/src/yaml.ts
NicholaiVogel: Confirmed on the pushed branch: `yaml` is already declared as a direct runtime dependency in `packages/core/package.json:29-32`.

packages/daemon/src/pipeline/synthesis-worker.test.ts
NicholaiVogel: Good catch. I tightened the two default-agent assertions to require `agentId: "default"` in `packages/daemon/src/pipeline/synthesis-worker.test.ts:124-130` and `packages/daemon/src/pipeline/synthesis-worker.test.ts:211-217`, and fixed the worker to normalize and forward the explicit default scope through both render and write paths in `packages/daemon/src/pipeline/synthesis-worker.ts:189-225`.

packages/core/src/agents.ts
NicholaiVogel: Updated the signature so the return type matches the actual contract: `packages/core/src/agents.ts:122-130`.

packages/core/src/migrations/056-agent-scoped-content-hash.ts
NicholaiVogel: Added that guardrail comment to the internal helper so the literal-only assumption is explicit: `packages/core/src/migrations/056-agent-scoped-content-hash.ts:3-7`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e10973bb`

Automated bot review: Found one security concern in the migration helper and one behavioral gap in the duplicate-detection scoping logic.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The SQL injection risk in addColumnIfMissing is directly visible in the migration file. The behavioral gap in normalizeReadPolicy for the legacy 'group' policy is verifiable by reading the function against the test that exercises that path.

packages/core/src/migrations/056-agent-scoped-content-hash.ts
```diff
@@ -0,0 +1,33 @@
+import type { MigrationDb } from "./index";
+
+function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
+	// Internal migration helper: callers must pass trusted SQL identifier/type literals.
+	const cols = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<Record<string, unknown>>;
+	if (cols.some((col) => col.name === column)) return;
```
PR-Reviewer-Ant: **SQL injection via unsanitized table/column names.** `addColumnIfMissing` interpolates `table`, `column`, and `definition` directly into SQL strings — both in the `PRAGMA table_info(${table})` query and the `ALTER TABLE` statement. The inline comment says "callers must pass trusted SQL identifier/type literals", but this relies on every future call site following that convention. Migrations are long-lived and copied forward; the comment is not a safe guardrail. At minimum, validate that `table` and `column` match `[A-Za-z_][A-Za-z0-9_]*` before interpolation, or use a column-name allowlist. In the current call sites the values are hardcoded literals so the risk is low now, but the pattern is dangerous as written.

packages/core/src/agents.ts
```diff
@@ -71,6 +79,57 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
...
+	if (readPolicy === "group" && typeof policyGroup === "string") {
+		return { readPolicy: "group", policyGroup };
+	}
+	return { readPolicy: "isolated", policyGroup: null };
+}
+
+export function normalizeAgentRosterEntry(entry: unknown): NormalizedAgentRosterEntry | null {
+	if (typeof entry !== "object" || entry === null) return null;
+	const record = entry as Record<string, unknown>;
+	if (typeof record.name !== "string" || record.name.length === 0) return null;
```
PR-Reviewer-Ant, line 111: **Legacy `group` policy loses `policyGroup` when `memory` block is present but `read_policy` is the flat string `"group"`.** In `normalizeReadPolicy`, the object-shape branch (lines 95–103) is checked before the flat-string `"group"` branch (line 104). If a roster entry is written as `{ name, memory: { read_policy: "group" } }` without a nested object — i.e., the canonical block is present but the policy value is still a flat string — the function reaches the flat-string branch with `policyGroup = memory.policy_group` which is `undefined`, not the top-level `record.policy_group`. The function then falls through to the `isolated` default. This edge case is not covered by the new tests (which only test `{ memory: { read_policy: { type: "group", ... } } }` and the bare legacy flat form). It may never appear in practice given `buildAgentMemoryConfig` always emits the nested object shape, but the normalizer claims backward compatibility and this gap could silently downgrade a hand-edited or third-party-written roster entry.

packages/core/src/migrations/056-agent-scoped-content-hash.ts
NicholaiVogel: I removed the generic interpolated helper and replaced it with a fixed `memories`-only repair path, so there are no caller-controlled identifiers in the migration anymore: `packages/core/src/migrations/056-agent-scoped-content-hash.ts:3-20`.

packages/core/src/agents.ts
NicholaiVogel: Fixed and pushed. `normalizeAgentRosterEntry` now preserves `memory.policy_group` when the `memory` block still uses the legacy flat `read_policy: "group"` form: `packages/core/src/agents.ts:108-123`. I also added a regression covering that mixed-shape case: `packages/core/src/__tests__/identity.test.ts:130-138`.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1159,7 +1159,7 @@ export function insertSummaryFacts(
 
 			const importance = Math.min(item.importance ?? 0.3, 0.5);
 
-			if (isDuplicate(db as unknown as Database, item.content)) continue;
+			if (isDuplicate(db as unknown as Database, item.content, job.agent_id)) continue;
```
PR-Reviewer-Ant: **Blocking: agentId passed raw — dedup silently disabled when agent_id is absent.**

The existing test `'writes summary facts with updated_by metadata'` calls `insertSummaryFacts` without `agent_id`, which means the second parameter type has `agent_id` as optional. At the changed line, `job.agent_id` is therefore `string | undefined`. When it is `undefined`, bun:sqlite binds NULL and the new FTS WHERE clause `AND m.agent_id = ?` evaluates to `AND m.agent_id = NULL` — always false in SQL — so no candidates are returned and `isDuplicate` returns `false` unconditionally. Dedup is silently disabled for any session without an explicit `agent_id`.

The synthesis-worker already has a pattern for this: `const scopeAgentId = normalizeAgentId(agentId)`. Apply the same here:
```ts
if (isDuplicate(db as unknown as Database, item.content, job.agent_id ?? "default")) continue;
```
Or thread `normalizeAgentId` into the job before the loop.

packages/core/src/yaml.ts
PR-Reviewer-Ant, line 3: **Blocking (unresolved from prior reviews): `yaml` package not confirmed as a declared dependency.**

`import YAML from "yaml"` is added here but `packages/core/package.json` is not in the diff across any of the four review rounds. If `yaml` is not already listed as a direct dependency of `@signet/core`, the build fails at runtime with a module-not-found error. Please confirm the package is declared or add it.

packages/core/src/agents.ts
```diff
@@ -71,6 +79,60 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
...
+	const memory =
+		typeof record.memory === "object" && record.memory !== null ? (record.memory as Record<string, unknown>) : null;
```
PR-Reviewer-Ant, line 113: **Nitpick: silent `isolated` fallback when `readPolicy === 'group'` but `policyGroup` is null/empty.**

`buildAgentMemoryConfig('group', null)` returns `{ read_policy: 'isolated' }` without any log or error. The CLI validates `--group` before calling this, so the case is prevented in practice. But if this helper is called from other paths in the future, the silent downgrade could mask misconfiguration. A comment or a console.warn would make the invariant explicit.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a63929f4`

[Automated Review — pr-reviewer bot] PR #455 fixes three real sub-agent memory bugs. One blocking behavioral regression was introduced in the dedupe path, one unresolved external dependency needs confirmation, and one previously flagged item (POST body shape) remains a latent consistency gap.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The agentId fallback gap in isDuplicate is directly provable: the pre-existing test at summary-worker.test.ts (first describe block, 'writes summary facts with updated_by metadata') calls insertSummaryFacts without agent_id, confirming the parameter is optional at the call-site type level. The changed line summary-worker.ts passes job.agent_id raw — if that field is undefined, bun:sqlite binds NULL and the WHERE clause AND m.agent_id = ? is always false, silently disabling dedup for that path. The yaml import is explicit in the yaml.ts diff; absence of packages/core/package.json from the diff is the only ambiguity. All other findings are directly readable from the changed files.


- packages/cli/src/commands/agent.ts:159 - **Warning: `POST /api/agents` body still uses the legacy flat shape while `addToRoster` now writes canonical nested shape.**

The `addToRoster` call (line 54) writes `memory: { read_policy: { type: 'group', group: '...' } }` to `agent.yaml`. The daemon `POST /api/agents` body (unchanged) still sends `{ name, read_policy, policy_group }`. The daemon's `syncAgentRoster` now correctly uses `normalizeAgentRosterEntry` which handles both formats, so there is no active breakage. However, if the `/api/agents` POST handler is ever updated to expect canonical shape, group/shared policy will silently be interpreted as `isolated`. A comment noting the deliberate format divergence would prevent a future regression.

packages/daemon/src/pipeline/summary-worker.ts
NicholaiVogel: Fixed and pushed in `5953554b`. `insertSummaryFacts` now normalizes an empty/missing agent scope to `"default"` before duplicate checks and inserts, so the scoped FTS filter cannot collapse to `agent_id = NULL`: `packages/daemon/src/pipeline/summary-worker.ts:1137-1189`. I also tightened the regression to prove the default-scope write explicitly: `packages/daemon/src/pipeline/summary-worker.test.ts:84-123`.

packages/core/src/yaml.ts
NicholaiVogel: Confirmed again on the current branch: `yaml` is already a direct runtime dependency of `@signet/core` at `packages/core/package.json:29-32`, and this PR continues to use it via the same runtime import in `packages/core/src/yaml.ts:5-25`.

packages/core/src/agents.ts
NicholaiVogel: Added the requested invariant comment and pushed it in `5953554b`. The helper still fails closed to `isolated`, but the fallback is now documented as intentional because CLI validation already rejects `group` without a concrete group name: `packages/core/src/agents.ts:125-136`.

packages/daemon/src/hooks.ts
```diff
@@ -517,7 +517,7 @@ function getSessionGapSummary(): string | undefined {
 }
```
PR-Reviewer-Ant, line 517: `isDuplicate` is an exported function and its signature changed from `(db, content)` to `(db, content, agentId)`. The diff only shows `summary-worker.ts` as a caller. Any other call sites in the daemon (or in packages that import from hooks.ts) are now broken at the TypeScript level. The PR description notes pre-existing typecheck errors that could mask new ones. Please confirm `grep -r 'isDuplicate' packages/` shows no other callers, or update them.

packages/daemon/src/pipeline/synthesis-worker.ts
```diff
@@ -191,18 +191,19 @@ async function runSynthesisWithDeps(
 	config: PipelineSynthesisConfig,
 	agentId?: string,
 ): Promise<SynthesisResult> {
+	const scopeAgentId = normalizeAgentId(agentId);
```
PR-Reviewer-Ant, line 194: `normalizeAgentId(agentId)` is called but is not defined in the diff and is not shown in any import change. If this function does not already exist in synthesis-worker.ts or its imports, this is a compile error. Please verify it is present and handles `null`/`undefined` → `"default"` normalization correctly.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,26 @@
...
- *
- * Simple YAML parser and formatter for flat/shallow config files.
- * Handles the common patterns used in Signet config files (agent.yaml, config.yaml)
- * without requiring a full YAML library dependency.
- *
- * Limitations:
- * - Supports up to 3 levels of nesting
- * - No support for YAML anchors, aliases, or complex types
- * - Arrays are formatted in block style only
  */
```
PR-Reviewer-Ant, line 4: The `yaml` npm package is now imported here. The diff does not include `packages/core/package.json`, so it is unclear whether `yaml` is already a declared dependency. If it is not, this will fail at build time. Please confirm `packages/core/package.json` lists `yaml` as a dependency.

packages/core/src/agents.ts
```diff
@@ -71,6 +79,63 @@ export function getAgentIdentityFiles(name: string, agentsDir: string): Record<s
...
+
+export function buildAgentMemoryConfig(
+	readPolicy: AgentRosterReadPolicy,
+	policyGroup: string | null,
+): { readonly read_policy: ReadPolicy } {
+	if (readPolicy === "shared") return { read_policy: "shared" };
+	if (readPolicy === "group" && typeof policyGroup === "string" && policyGroup.length > 0) {
+		return { read_policy: { type: "group", group: policyGroup } };
+	}
+	// CLI validation already rejects `group` without a concrete group name.
```
PR-Reviewer-Ant, line 133: `buildAgentMemoryConfig("group", null)` silently falls back to `{ read_policy: "isolated" }` — the fail-closed behavior is correct per the comment, but no test covers this case. The comment explains why this is intentional, but a regression test would prevent a future reader from removing the fallback thinking it's dead code.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5953554b`

[Automated Bot Review] PR #455 — fix: scope sub-agent memory dedupe and synthesis. Core logic is sound and the regression tests are well-targeted. Three cross-module concerns require human verification before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The three flagged issues each require context outside the diff: (1) isDuplicate callers — only summary-worker.ts is shown; other daemon-side callers aren't visible. (2) normalizeAgentId in synthesis-worker.ts — not defined in the diff, must already exist in the file. (3) yaml package in packages/core/package.json — not shown in the diff. The core logic changes (agent-scoped FTS JOIN, synthesis agentId threading, roster normalization) are correct and provable from this diff alone.

packages/daemon/src/hooks.ts
NicholaiVogel: Confirmed and pushed in `aecf4893`. The only runtime caller remains the summary worker at `packages/daemon/src/pipeline/summary-worker.ts:1161-1180`, and I updated the remaining legacy test callers to pass the explicit scope as well: `packages/daemon/test/hooks.test.ts:562-607`.

packages/daemon/src/pipeline/synthesis-worker.ts
NicholaiVogel: Verified on the current branch: `normalizeAgentId` is already defined in the worker and normalizes blank/undefined values to `"default"`: `packages/daemon/src/pipeline/synthesis-worker.ts:41-44`. The scheduled synthesis path uses that normalized value for both render and write flows: `packages/daemon/src/pipeline/synthesis-worker.ts:189-225`.

packages/core/src/yaml.ts
NicholaiVogel: Confirmed again on the pushed branch: `yaml` is already declared as a direct dependency of `@signet/core` in `packages/core/package.json:29-32`, and the runtime import remains in `packages/core/src/yaml.ts:5-25`.

packages/core/src/agents.ts
NicholaiVogel: Added the requested regression and pushed it in `aecf4893`. The fail-closed fallback is now covered directly by `buildAgentMemoryConfig("group", null)`: `packages/core/src/__tests__/identity.test.ts:140-150`, alongside the documented invariant in `packages/core/src/agents.ts:125-136`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `aecf4893`

[Automated Review — pr-reviewer] PR #455 adds agent-scoped memory deduplication, routes synthesis writes to the correct agent, and centralizes roster normalization. The core logic is correct and well-tested. Two issues worth discussing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The duplicate-detection join, migration DDL, and roster normalization logic are all directly readable in the diff and provably correct. The two flagged issues require seeing packages/core/package.json (not in the diff) and the full set of parseSimpleYaml call sites in daemon.ts (only partial daemon.ts is shown).

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,26 @@
...
 
+import YAML from "yaml";
+
 /**
```
PR-Reviewer-Ant, line 7: `import YAML from "yaml"` requires the `yaml` package to be listed in `packages/core/package.json`. The diff contains no package.json change. If `yaml` is not already a direct or workspace-hoisted dependency of `@signet/core`, this will throw at import time and break every consumer of the package (daemon startup, CLI startup). Please confirm the package is declared, or add it to `packages/core/package.json`.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,26 @@
...
-		} else {
-			// Simple value - apply type coercion
-			parent[key] = coerceYamlValue(value);
-		}
-	}
-
-	return result;
+	const parsed = YAML.parse(text);
+	return typeof parsed === "object" && parsed !== null ? (parsed as Record<string, unknown>) : {};
 }
```
PR-Reviewer-Ant, line 20: Behavioral change: `YAML.parse` throws a `YAMLParseError` on malformed input, whereas the old hand-rolled parser silently returned a partial result. `readYaml()` in `agent.ts` is protected with a try/catch, but any other call-site in daemon.ts or elsewhere that calls `parseSimpleYaml` directly without error handling will now propagate an exception instead of degrading gracefully. If the daemon calls `parseSimpleYaml` on a user-owned file (e.g. `agent.yaml`) without a catch, a malformed file could crash the startup path. Audit direct call-sites outside agent.ts and wrap them, or add the try/catch inside `parseSimpleYaml` itself.

packages/core/src/yaml.ts
NicholaiVogel: Confirmed on the pushed branch: `yaml` is already a direct dependency of `@signet/core` in `packages/core/package.json:29-32`, and this PR continues to import it from `packages/core/src/yaml.ts:5-33`.

packages/core/src/yaml.ts
NicholaiVogel: Good catch. I restored fail-soft parsing in `28180d68`: `parseSimpleYaml` now catches malformed YAML and degrades to `{}` instead of propagating parser exceptions into daemon/CLI startup: `packages/core/src/yaml.ts:7-20`. I also added a regression covering malformed input directly: `packages/core/src/__tests__/identity.test.ts:101-105`.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,33 @@
...
- * without requiring a full YAML library dependency.
- *
- * Limitations:
- * - Supports up to 3 levels of nesting
- * - No support for YAML anchors, aliases, or complex types
- * - Arrays are formatted in block style only
  */
 
+import YAML from "yaml";
+
```
PR-Reviewer-Ant, line 6: `import YAML from 'yaml'` is a new hard dependency on the `yaml` package. The diff does not include any `package.json` change for `packages/core`. If `yaml` is not already listed as a dependency in `packages/core/package.json`, this import will fail at runtime (module not found). Worth confirming the package is present before merge.

packages/daemon/src/hooks.ts
PR-Reviewer-Ant, line 517: `isDuplicate` is an exported function and its signature now requires a third `agentId: string` argument. The diff updates all callers visible in the diff (`summary-worker.ts`, `test/hooks.test.ts`). However, since the function is exported, other consumers outside the diffed files are possible. The PR notes that `bunx tsc --noEmit` reported pre-existing workspace issues — if those masked new arity errors, a silently broken caller could slip through. Worth a targeted grep for `isDuplicate(` across the daemon package to confirm no callsite was missed.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1140,6 +1140,7 @@ export function insertSummaryFacts(
 	facts: ReadonlyArray<LlmSummaryResult["facts"][number]>,
 ): number {
 	const now = new Date().toISOString();
+	const agentId = job.agent_id.length > 0 ? job.agent_id : "default";
```
PR-Reviewer-Ant: `job.agent_id.length > 0` will throw `TypeError: Cannot read properties of null` if `agent_id` is null at runtime, even though the TypeScript type says `string`. `summary_jobs.agent_id` has `DEFAULT 'default'`, so existing rows are safe, but a belt-and-suspenders `(job.agent_id ?? '').length > 0` would be more defensive here.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `28180d68`

[Automated Review — pr-reviewer bot] PR #455 scopes sub-agent memory dedupe and synthesis writes correctly. No blocking issues found; three items worth discussing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The core behavioral changes in hooks.ts, summary-worker.ts, synthesis-worker.ts, agents.ts, and the migration are fully visible and verifiable. Two gaps: (1) memory-head.ts and writeMemoryMd full implementations are not in the diff — the tests claim the agentId forwarding works, but the implementation path is inferred, not read; (2) packages/core/package.json is not in the diff, so the new `import YAML from 'yaml'` in yaml.ts cannot be confirmed to have a matching dependency entry.

packages/core/src/yaml.ts
NicholaiVogel: Confirmed on the current branch: `yaml` is already a direct dependency of `@signet/core` in `packages/core/package.json:29-32`.

packages/daemon/src/hooks.ts
NicholaiVogel: I did the targeted caller check before pushing `f62ca633`. The exported definition is at `packages/daemon/src/hooks.ts:520-559`, and the only in-repo call sites are the summary-worker runtime path at `packages/daemon/src/pipeline/summary-worker.ts:1161-1163` plus the dedicated hooks tests at `packages/daemon/test/hooks.test.ts:562-607`.

packages/daemon/src/pipeline/summary-worker.ts
NicholaiVogel: Agreed. I hardened that path in `f62ca633` so runtime nulls fail closed to the default scope instead of throwing: `packages/daemon/src/pipeline/summary-worker.ts:1142-1163`. I also added a regression that passes a null `agent_id` through the insert path and asserts the stored fact lands under `default`: `packages/daemon/src/pipeline/summary-worker.test.ts:125-149`.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,33 @@
...
- *
- * Simple YAML parser and formatter for flat/shallow config files.
- * Handles the common patterns used in Signet config files (agent.yaml, config.yaml)
- * without requiring a full YAML library dependency.
- *
- * Limitations:
- * - Supports up to 3 levels of nesting
- * - No support for YAML anchors, aliases, or complex types
- * - Arrays are formatted in block style only
  */
```
PR-Reviewer-Ant, line 4: `import YAML from "yaml"` is a new dependency on the `yaml` npm package, but `packages/core/package.json` is not in this diff. If `yaml` is not already an explicit dependency of `@signet/core`, this import fails at runtime and breaks daemon and CLI startup for all users. Please confirm the package is declared in the core package manifest (or add it).

packages/daemon/src/hooks.ts
```diff
@@ -527,8 +527,16 @@ export function isDuplicate(db: Database, content: string): boolean {
 	try {
 		const ftsQuery = words.slice(0, 10).join(" OR ");
 		const rows = db
-			.prepare("SELECT content FROM memories_fts WHERE memories_fts MATCH ? LIMIT 10")
-			.all(ftsQuery) as Array<{ content: string }>;
+			.prepare(
```
PR-Reviewer-Ant, line 530: The FTS duplicate query filters with `AND m.agent_id = ?` using the already-normalized `agentId` string (always 'default' or a non-empty name). The DB-level unique index in migration 056 uses `COALESCE(NULLIF(agent_id, ''), 'default')`, which additionally normalizes empty-string agent_id rows to 'default'. Any row with `agent_id = ''` (possible on very old databases before migration 043 backfilled the column) would be caught by the unique index but missed by this FTS check, allowing a duplicate insert attempt that the index then rejects with a constraint error rather than being silently skipped. The write path in `insertSummaryFacts` already normalizes to 'default' before writing, so new rows will never have an empty agent_id — but existing rows could. Low risk, but worth documenting or adding a `OR m.agent_id = ''` arm to match the index semantics exactly.

packages/core/src/yaml.ts
```diff
@@ -1,152 +1,33 @@
...
-		}
+	try {
+		const parsed = YAML.parse(text);
+		return typeof parsed === "object" && parsed !== null ? (parsed as Record<string, unknown>) : {};
+	} catch {
+		return {};
 	}
-
-	return result;
 }
```
PR-Reviewer-Ant, line 20: The `_indent` parameter is silently ignored; `YAML.stringify` always uses 2-space indentation. The old `formatYaml` was recursive and accepted an indent offset to support nested calls. Any internal call site that relied on a non-zero indent (e.g., recursive depth) will now produce different output. From the diff the function is only called at depth 0 (`writeYaml(dir, cfg)`), so this is not a current bug, but the signature change is a silent behavioral difference that should be noted in the function's JSDoc or removed from the signature entirely to avoid confusion.

packages/cli/src/commands/agent.test.ts
```diff
@@ -0,0 +1,71 @@
...
+		const dir = mkdtempSync(join(tmpdir(), "signet-agent-command-"));
+		tempDirs.push(dir);
+		console.log = () => {};
+
+		const program = new Command();
+		registerAgentCommands(program, {
+			AGENTS_DIR: dir,
+			fetchFromDaemon: async () => null,
+		});
+
```
PR-Reviewer-Ant, line 31: This line exceeds the 120-character hard limit defined in AGENTS.md. Not a correctness issue but violates the repo's hard line-length rule; the Biome check may flag it.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f62ca633`

[Automated Review — pr-reviewer bot] PR #455 fixes sub-agent memory scoping. Core logic is sound and well-tested. Two issues worth flagging: a missing-dependency risk on the yaml package and a minor normalization inconsistency between the application-level FTS duplicate check and the DB-level unique index.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff is complete for all changed files. The yaml dependency concern cannot be confirmed without packages/core/package.json, which is not in the diff. The FTS/index normalization gap is provable from hooks.ts:530 and 056-agent-scoped-content-hash.ts:24–27.

PR #457, opened 2026-04-03

aaf2tbz:
## Summary

Split `packages/daemon/src/daemon.ts` from **12,690 → 2,495 lines** (80% reduction) by extracting route handlers into 13 dedicated modules under `routes/`.

Primary goal remains structural: `daemon.ts` now retains app setup, middleware, permission guards, file watcher, pipeline runtime startup/shutdown, cleanup handlers, and `main()`. Follow-up commits on this PR also fixed a few extraction regressions discovered in review, without changing intended daemon behavior.

## Changes

- `routes/state.ts` now owns shared mutable runtime state used by extracted route modules, with setters used by `daemon.ts` during startup/shutdown
- Restored extracted auth endpoints/guards that still belong in `daemon.ts` (`/api/auth/whoami`, `/api/auth/token`)
- Moved session cleanup startup back into `main()` and removed import-time cleanup side effects from extracted modules
- Made git shutdown cleanup explicitly drain the auto-commit debounce managed by `routes/git-sync.ts`
- Added a regression test proving importing `daemon.ts` for route registration does not start session cleanup
- Fixed `update-route.test.ts` structural path after extraction

|---|---|---|
| `memory-routes.ts` | 2,574 | `/api/memory/*`, `/memory/*`, embeddings, documents |
| `hooks-routes.ts` | 1,183 | `/api/hooks/*`, cross-agent, synthesis |
| `utils.ts` | 1,107 | Shared parsers, validators, query builders |
| `git-sync.ts` | 729 | Git pull/push/sync, auto-commit, credential resolution |
| `misc-routes.ts` | 806 | Logs, config, identity, agents, update, tasks |
| `repair-routes.ts` | 665 | `/api/repair/*`, `/api/troubleshoot/*` |
| `pipeline-routes.ts` | 534 | `/api/status`, pipeline, dream, diagnostics |
| `knowledge-routes.ts` | 498 | `/api/knowledge/*` |
| `telemetry-routes.ts` | 467 | Analytics, telemetry, timeline, checkpoints |
| `state.ts` | 442 | Shared runtime state, config utilities, network helpers |
| `connectors-routes.ts` | 442 | `/api/connectors/*`, harnesses |
| `session-routes.ts` | 296 | `/api/sessions/*` |
| `secrets-routes.ts` | 241 | `/api/secrets/*` |

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Security checks applied to admin/mutation endpoints — `/api/auth/token` guard restored in `daemon.ts`; other admin guards preserved
- [x] Docs updated for API/spec/status changes — no API/spec/schema change beyond restoring extracted auth wiring
- [x] Regression tests added for each bug fix — added import-side-effect regression coverage for session cleanup startup
- [x] Lint/typecheck/tests pass locally — touched-file Biome passes; focused daemon tests pass (see Testing)

N/A — no migrations touched.

## Testing

- [x] `bun test packages/daemon/src/daemon-refactor.test.ts packages/daemon/src/auth/auth.test.ts packages/daemon/src/session-api.test.ts packages/daemon/src/update-route.test.ts`
- [x] `bunx biome check packages/daemon/src/daemon.ts packages/daemon/src/routes/state.ts packages/daemon/src/routes/hooks-routes.ts packages/daemon/src/routes/git-sync.ts packages/daemon/src/session-tracker.ts packages/daemon/src/daemon-refactor.test.ts`
- [ ] `bun run typecheck`
- [ ] N/A

Notes:
- `bun run --filter @signet/daemon build:types` still fails on many pre-existing workspace/type issues outside this patch, so it is not a reliable signal for this PR.

## AI disclosure

## Notes

Shared mutable daemon runtime state is now centralized in `routes/state.ts` and updated through explicit setters from `daemon.ts`, so extracted route modules observe live startup/shutdown state without reintroducing module-scope side effects.

### DIFF
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -4,21 +4,11 @@
  * Background service for memory, API, and dashboard hosting
  */
 
-import type { Database } from "bun:sqlite";
-import { type ChildProcess, execFileSync, spawn } from "node:child_process";
+import type { ChildProcess } from "node:child_process";
+import { execFileSync, spawn } from "node:child_process";
 import { createHash } from "node:crypto";
-import { copyFileSync } from "node:fs";
-import {
-	appendFileSync,
-	existsSync,
-	mkdirSync,
-	readFileSync,
-	readdirSync,
-	realpathSync,
-	statSync,
-	unlinkSync,
-	writeFileSync,
-} from "node:fs";
+import { appendFileSync, copyFileSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync, readdirSync, statSync, unlinkSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { basename, dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
@@ -26,119 +16,45 @@ import { createAdaptorServer, type serve } from "@hono/node-server";
 import { serveStatic } from "@hono/node-server/serve-static";
 import {
 	type AgentDefinition,
-	CONNECTOR_PROVIDERS,
-	type ConnectorConfig,
-	SIGNET_GIT_PROTECTED_PATHS,
-	type SyncCursor,
 	buildArchitectureDoc,
-	findSignetForgeBinary,
-	keywordSearch,
-	mergeSignetGitignoreEntries,
 	networkModeFromBindHost,
 	parseSimpleYaml,
-	readNetworkMode,
-	readPipelinePauseState,
-	resolveNetworkBinding,
-	resolveSignetForgeManagedPath,
-	setPipelinePaused,
 	stripSignetBlock,
-	vectorSearch,
 } from "@signet/core";
 import { watch } from "chokidar";
 import { type Context, Hono } from "hono";
 import { cors } from "hono/cors";
-import { logger as honoLogger } from "hono/logger";
-import { getAgentScope, resolveAgentId } from "./agent-id";
-import { type AnalyticsCollector, type ErrorStage, createAnalyticsCollector } from "./analytics";
+import { resolveAgentId } from "./agent-id";
+import { type AnalyticsCollector, createAnalyticsCollector } from "./analytics";
 import {
 	type AuthConfig,
 	AuthRateLimiter,
-	type TokenRole,
-	type TokenScope,
 	checkScope,
 	createAuthMiddleware,
-	createToken,
 	loadOrCreateSecret,
 	parseAuthConfig,
 	requirePermission,
 	requireRateLimit,
-	requireScope,
 } from "./auth";
 import { bindWithRetry } from "./bind-with-retry";
 import { migrateConfig } from "./config-migration";
-import { createFilesystemConnector } from "./connectors/filesystem";
-import {
-	getConnector,
-	listConnectors,
-	registerConnector,
-	removeConnector,
-	updateConnectorStatus,
-	updateCursor,
-} from "./connectors/registry";
+import { listConnectors } from "./connectors/registry";
 import { normalizeAndHashContent } from "./content-normalization";
-import {
-	type AgentMessage,
-	type AgentMessageType,
-	clearAllPresence,
-	createAgentMessage,
-	getAgentPresenceForSession,
-	isMessageVisibleToAgent,
-	listAgentMessages,
-	listAgentPresence,
-	relayMessageViaAcp,
-	removeAgentPresence,
-	subscribeCrossAgentEvents,
-	touchAgentPresence,
-	upsertAgentPresence,
-} from "./cross-agent";
+import { clearAllPresence } from "./cross-agent";
 import { closeDbAccessor, getDbAccessor, initDbAccessor } from "./db-accessor";
-import { syncVecDeleteBySourceId, syncVecInsert, vectorToBlob } from "./db-helpers";
-import {
-	type DiagnosticsReport,
-	type PredictorHealthParams,
-	createProviderTracker,
-	getDiagnostics,
-	getPredictorHealth,
-} from "./diagnostics";
-import {
-	fetchEmbedding,
-	resolveEmbeddingApiKey,
-	resolveEmbeddingBaseUrl,
-	resolveOllamaUrl,
-	setNativeFallbackToOllama,
-} from "./embedding-fetch";
-import { buildEmbeddingHealth } from "./embedding-health";
+import { syncVecDeleteBySourceId, syncVecInsert } from "./db-helpers";
+import { type DiagnosticsReport, createProviderTracker } from "./diagnostics";
+import { fetchEmbedding, setNativeFallbackToOllama } from "./embedding-fetch";
 import { type EmbeddingTrackerHandle, startEmbeddingTracker } from "./embedding-tracker";
 import { getAllFeatureFlags, initFeatureFlags } from "./feature-flags";
 import { writeFileIfChangedAsync } from "./file-sync";
-import { walkImpact } from "./graph-impact";
 import { syncAgentWorkspaces } from "./identity-sync";
-import { linkMemoryToEntities } from "./inline-entity-linker";
-import {
-	getAttributesForAspectFiltered,
-	getEntityAspectsWithCounts,
-	getEntityDependenciesDetailed,
-	getEntityHealth,
-	getKnowledgeEntityDetail,
-	getKnowledgeGraphForConstellation,
-	getKnowledgeStats,
-	getPinnedEntities,
-	listKnowledgeEntities,
-	pinEntity,
-	resolveNamedEntity,
-	unpinEntity,
-} from "./knowledge-graph";
 import { closeLlmProvider, getLlmProvider, initLlmProvider } from "./llm";
-import { type LogCategory, type LogEntry, logger } from "./logger";
+import { type LogEntry, logger } from "./logger";
 import { type EmbeddingConfig, type ResolvedMemoryConfig, loadMemoryConfig } from "./memory-config";
-import { type RecallParams, hybridRecall } from "./memory-search";
-import { buildMemoryTimeline } from "./memory-timeline";
-import { ONEPASSWORD_SERVICE_ACCOUNT_SECRET, importOnePasswordSecrets, listOnePasswordVaults } from "./onepassword.js";
-import { recordPathFeedback } from "./path-feedback";
 import {
 	DEFAULT_RETENTION,
 	enqueueDocumentIngestJob,
-	enqueueExtractionJob,
 	ensureRetentionWorker,
 	getDreamingPasses,
 	getDreamingState,
@@ -152,16 +68,8 @@ import {
 	stopPipeline,
 } from "./pipeline";
 import { AlreadyRunningError, type DreamingWorkerHandle, startDreamingWorker } from "./pipeline/dreaming-worker";
-import { getFeedbackTelemetry } from "./pipeline/aspect-feedback";
-import { clusterEntities } from "./pipeline/community-detection";
-import { deadLetterExtractionJob, deadLetterPendingExtractionJobs } from "./pipeline/extraction-fallback";
-import { getGraphBoostIds } from "./pipeline/graph-search";
-import {
-	getTraversalStatus,
-	invalidateTraversalCache,
-	resolveFocalEntities,
-	traverseKnowledgeGraph,
-} from "./pipeline/graph-traversal";
+import { deadLetterPendingExtractionJobs } from "./pipeline/extraction-fallback";
+import { getTraversalStatus, invalidateTraversalCache } from "./pipeline/graph-traversal";
 import {
 	getAvailableModels,
 	getModelsByProvider,
@@ -183,79 +91,62 @@ import {
 	stopOpenCodeServer,
 } from "./pipeline/provider";
 import { resolveRuntimeModel } from "./pipeline/provider-resolution";
-import { type RerankCandidate, noopReranker, rerank } from "./pipeline/reranker";
-import { createEmbeddingReranker } from "./pipeline/reranker-embedding";
-import { type PredictorClient, createPredictorClient, resolvePredictorCheckpointPath } from "./predictor-client";
+import { startReconciler } from "./pipeline/skill-reconciler";
+import { type PredictorClient, createPredictorClient } from "./predictor-client";
 import { detectDrift } from "./predictor-comparison";
-import {
-	getComparisonsByEntity,
-	getComparisonsByProject,
-	listComparisons,
-	listTrainingRuns,
-} from "./predictor-comparisons";
 import { getPredictorState } from "./predictor-state";
+import { type RepairContext, backfillSkippedSessions, structuralBackfill } from "./repair-actions";
+import { shouldEnforceScope } from "./request-scope";
 import {
-	type RepairContext,
-	type RepairResult,
-	backfillSkippedSessions,
-	checkFtsConsistency,
-	cleanOrphanedEmbeddings,
-	createRateLimiter,
-	deduplicateMemories,
-	findDeadMemories,
-	forgetDeadMemories,
-	getDedupStats,
-	getEmbeddingGapStats,
-	pruneChunkGroupEntities,
-	pruneSingletonExtractedEntities,
-	reclassifyEntities,
-	reembedMissingMemories,
-	releaseStaleLeases,
-	requeueDeadJobs,
-	resyncVectorIndex,
-	structuralBackfill,
-	triggerRetentionSweep,
-} from "./repair-actions";
-import {
-	resolveScopedAgent,
-	resolveScopedProject as resolveScopedProjectForClaims,
-	shouldEnforceScope,
-} from "./request-scope";
-import {
-	CRON_PRESETS,
-	computeNextRun,
-	isHarnessAvailable,
-	resolveSkillPrompt,
-	startSchedulerWorker,
-	validateCron,
-} from "./scheduler";
-import { emitTaskStream, getTaskStreamSnapshot, subscribeTaskStream } from "./scheduler/task-stream";
-import { deleteSecret, execWithSecrets, getSecret, hasSecret, listSecrets, putSecret } from "./secrets.js";
-import {
-	flushPendingCheckpoints,
-	getCheckpointsByProject,
-	getCheckpointsBySession,
-	initCheckpointFlush,
-	pruneCheckpoints,
-	redactCheckpointRow,
-} from "./session-checkpoints";
-import { parseFeedback, recordAgentFeedback } from "./session-memories";
+	AGENTS_DIR,
+	ALLOWED_ORIGINS,
+	BIND_HOST,
+	CURRENT_VERSION,
+	DAEMON_DIR,
+	HOST,
+	INTERNAL_SELF_HOST,
+	LOG_DIR,
+	MEMORY_DB,
+	NETWORK_MODE,
+	PID_FILE,
+	PORT,
+	analyticsCollector,
+	authCrossAgentMessageLimiter,
+	bindAbort,
+	invalidateDiagnosticsCache,
+	isAllowedOrigin,
+	isManagedOpenCodeLocalEndpoint,
+	normalizeRuntimeBaseUrl,
+	providerRuntimeResolution,
+	providerTracker,
+	queueExtractionJob,
+	readEnvTrimmed,
+	redactUrlForLogs,
+	repairLimiter,
+} from "./routes/state.js";
+import { isHarnessAvailable } from "./scheduler";
+import { startSchedulerWorker } from "./scheduler/index.js";
+import { getSecret, hasSecret } from "./secrets.js";
+import { flushPendingCheckpoints, initCheckpointFlush, pruneCheckpoints } from "./session-checkpoints";
+import { releaseAllSessions, startSessionCleanup, stopSessionCleanup } from "./session-tracker";
 import { createSingleFlightRunner } from "./single-flight-runner";
 import { closeSynthesisProvider, initSynthesisProvider } from "./synthesis-llm";
-import { readScopedTask, readTaskAgentId } from "./task-scope";
 import { type TelemetryCollector, type TelemetryEventType, createTelemetryCollector } from "./telemetry";
-import { expandTemporalNode } from "./temporal-expand";
-import { upsertThreadHead } from "./thread-heads";
-import { type TimelineSources, buildTimeline } from "./timeline";
-import {
-	type MutationContext,
-	txFinalizeAccessAndHistory,
-	txForgetMemory,
-	txIngestEnvelope,
-	txModifyMemory,
-	txRecoverMemory,
-} from "./transactions";
-import { cacheProjection, computeProjection, computeProjectionForQuery, getCachedProjection } from "./umap-projection";
+import { closeWidgetProvider, initWidgetProvider } from "./widget-llm";
+
+import { mountMcpRoute } from "./mcp/route.js";
+import { registerConnectorRoutes } from "./routes/connectors-routes.js";
+import { getGitStatus, gitSync, scheduleAutoCommit, startGitSyncTimer, stopGitSyncTimer } from "./routes/git-sync.js";
+import { registerHooksRoutes } from "./routes/hooks-routes.js";
+import { registerKnowledgeRoutes } from "./routes/knowledge-routes.js";
+import { registerMemoryRoutes } from "./routes/memory-routes.js";
+import { registerMiscRoutes } from "./routes/misc-routes.js";
+import { registerPipelineRoutes } from "./routes/pipeline-routes.js";
+import { registerRepairRoutes } from "./routes/repair-routes.js";
+import { registerSecretRoutes } from "./routes/secrets-routes.js";
+import { registerSessionRoutes } from "./routes/session-routes.js";
+import { registerTelemetryRoutes } from "./routes/telemetry-routes.js";
+import { blobToVector, chunkBySentence, getConfiguredProviderHints, parseOptionalString } from "./routes/utils.js";
 import {
 	MAX_UPDATE_INTERVAL_SECONDS,
 	MIN_UPDATE_INTERVAL_SECONDS,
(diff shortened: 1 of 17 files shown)


BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `50c3b398`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `50c3b398`) now and I'll follow up shortly with feedback.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `77a689bd`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `77a689bd`) now and I'll follow up shortly with feedback.

packages/daemon/src/routes/git-sync.ts
```diff
@@ -0,0 +1,729 @@
...
+	if (!remoteUrl) return false;
+
+	if (remoteUrl.startsWith("git@")) return true;
+
+	if (remoteUrl.startsWith("https://")) {
+		const creds = await getCredentialHelperToken(remoteUrl, AGENTS_DIR);
+		if (creds) return true;
+	}
+
+	const isGitHub = remoteUrl.includes("github.com");
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — replaced string includes with URL hostname parsing.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
-const PROJECTION_ERROR_TTL_MS = 30_000;
-let hasMemoriesSessionIdColumnCache: boolean | null = null;
-
-// Auth state — initialized lazily in main(), but middleware reads from here
+const OPENCLAW_STALE_MS = 10 * 60 * 1000;
 let authConfig: AuthConfig = parseAuthConfig(undefined, AGENTS_DIR);
 let authSecret: Buffer | null = null;
 let authForgetLimiter = new AuthRateLimiter(60_000, 30);
 let authModifyLimiter = new AuthRateLimiter(60_000, 60);
 let authBatchForgetLimiter = new AuthRateLimiter(60_000, 5);
```
PR-Reviewer-Ant, line 178: **Blocking — initialization order bug.** `startSessionCleanup()` is called at module-evaluation scope (immediately after `registerMiscRoutes(app)`), well before `main()` runs and before `initDbAccessor()` is called. If `startSessionCleanup()` calls `getDbAccessor()` internally (which session cleanup logic almost certainly does), this will throw or silently no-op on an uninitialized accessor on every daemon start. The PR description says this import was 'fixed as missing during extraction' — the correct fix is to call `startSessionCleanup()` from inside `main()`, after `initDbAccessor()`, not at module scope.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +91,62 @@ import {
 	stopOpenCodeServer,
```
PR-Reviewer-Ant, line 93: **Warning — auth routes not accounted for in diff.** The original daemon.ts contained `GET /api/auth/whoami` and `POST /api/auth/token` (with its `app.use` permission guard). These are removed from daemon.ts in this PR, but none of the 13 extracted module descriptions mention auth, and the `misc-routes.ts` file content is not provided in the diff. If these routes were not carried into a route module, token creation and whoami are silently broken — a behavioral regression despite the 'no behavior changes' claim. Please confirm which module holds these handlers.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
-	return hasMemoriesSessionIdColumnCache;
-}
+let shuttingDown = false;
+let httpServer: ReturnType<typeof createAdaptorServer> | null = null;
+let dreamingWorkerHandle: DreamingWorkerHandle | null = null;
+let shadowProcess: ChildProcess | null = null;
+let predictorClientRef: PredictorClient | null = null;
+let embeddingTrackerHandle: EmbeddingTrackerHandle | null = null;
+let skillReconcilerHandle: ReturnType<typeof startReconciler> | null = null;
+let structuralBackfillTimer: ReturnType<typeof setTimeout> | null = null;
```
PR-Reviewer-Ant, line 189: **Warning — dead timer state in cleanup path.** `commitPending` and `commitTimer` are declared as module-level `let` variables in daemon.ts and the `cleanup()` function clears them, but the actual auto-commit debounce scheduling is now handled by `scheduleAutoCommit()` imported from `git-sync.ts`, which manages its own internal timer state. The `commitTimer` in daemon.ts is never set — clearing it in `cleanup()` is a no-op. The debounced commit timer inside `git-sync.ts` must be explicitly cancelled on shutdown (either by `stopGitSyncTimer()` covering it, or by a new export). If not, a pending commit could fire after shutdown begins. Verify `stopGitSyncTimer()` cancels both the periodic sync timer and the debounce timer.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
 	stopUpdateTimer,
```
PR-Reviewer-Ant, line 171: **Nitpick — weakened type on exported function.** `recordPredictorLatency` was previously typed as `(operation: 'predictor_score' | 'predictor_train', ms: number)` — a closed union matching the known operation keys. The new signature is `(operation: string, durationMs: number)`, which accepts arbitrary strings and can silently record meaningless analytics keys. The codebase convention is to prefer discriminated unions over `string`. Consider keeping the union type or at least extracting it as a named type.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `50c3b398`

[Automated Review — pr-reviewer bot] This is a large structural refactor (80% LOC reduction in daemon.ts). Three correctness issues were found: one blocking initialization-order bug, one unverifiable route-drop risk, and one shutdown timer leak. No security regressions were introduced by the restructuring itself.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The initialization-order bug is directly provable from the changed file contents: `startSessionCleanup()` appears at module-evaluation scope in the new daemon.ts (after the `registerMiscRoutes` call, outside any function), while `initDbAccessor()` is called inside `main()`. The auth-route gap is derivable from the diff showing removal of `/api/auth/whoami` and `/api/auth/token` from daemon.ts with no corresponding addition visible in the provided changed-file contents. The `commitTimer` dead-state issue is visible from the new daemon.ts cleanup() function referencing a timer variable (`commitTimer`) that is never set within daemon.ts itself, while `scheduleAutoCommit` is delegated to git-sync.ts.

packages/daemon/src/routes/git-sync.ts
```diff
@@ -0,0 +1,729 @@
...
+					method: "credential-helper",
+					authUrl: buildAuthUrlFromCreds(remoteUrl, creds),
+				};
+			}
+		} catch {
+			/* ignore */
+		}
+	}
+
+	const isGitHub = remoteUrl.includes("github.com") || remoteUrl.includes("github.com:");
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,534 @@
...
+			return c.json({
+				enabled: true,
+				alive: false,
+				crashCount: 0,
+				crashDisabled: false,
+				status: null,
+			});
+		}
+
+		const status = await client.status();
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — changed if (client === null) to if (client == null) to handle both null and undefined.

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,534 @@
...
+		return c.json({
+			enabled: true,
+			alive: client.isAlive(),
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — nullish guard (== null) now covers undefined, so client.isAlive() is only reached when client is non-nullish.

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,534 @@
...
+			crashCount: client.crashCount,
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — same nullish guard covers client.crashCount access.

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,534 @@
...
+			crashDisabled: client.crashDisabled,
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — if (client === null) changed to if (client == null). Optional chaining not needed since the guard fully covers the nullish case.

packages/daemon/src/routes/hooks-routes.ts
```diff
@@ -0,0 +1,1183 @@
...
+					agentId: scopedAgent.agentId,
+					owner: "api-hooks-synthesis-complete",
+				});
+				if (!result.ok) {
+					const status = result.code === "busy" ? 409 : 400;
+					return c.json({ error: result.error }, status);
+				}
+				logger.info("hooks", "MEMORY.md synthesized");
+			} finally {
+				if (worker && lockToken !== null) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — removed redundant lockToken !== null check. After the early return at line 1103, TypeScript narrows lockToken to number, making the comparison always true.

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,534 @@
...
+		}
+
+		const contentType = c.req.header("content-type") ?? "";
+		let mode: "compact" | "incremental" = "incremental";
+		if (contentType.includes("application/json")) {
+			const raw: unknown = await c.req.json().catch(() => null);
+			if (raw === null) {
+				return c.json({ error: "Malformed JSON body" }, 400);
+			}
+			if (typeof raw === "object" && raw !== null && "mode" in raw && raw.mode === "compact") {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — cast raw to { mode?: unknown } after narrowing to object, removing the direct property access on the broad object type.

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,534 @@
...
+import { STATUS_CACHE_TTL, cachedEmbeddingStatus, statusCacheTime } from "./utils.js";
+
+const pipelineAdminGuard = async (c: Context, next: () => Promise<void>): Promise<Response | undefined> => {
+	const perm = requirePermission("admin", authConfig);
+	const rate = requireRateLimit("admin", authAdminLimiter, authConfig);
+	return perm(c, async () => rate(c, next));
+};
+
+async function togglePipelinePause(c: Context, paused: boolean): Promise<Response> {
+	if (shuttingDown) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — removed the dead if (shuttingDown) branch and its import. shuttingDown is const false in state.ts so the branch was unreachable.

packages/daemon/src/routes/telemetry-routes.ts
```diff
@@ -0,0 +1,467 @@
...
+
+		return c.json({
+			...result,
+			checkpoint_path: checkpointPath,
+			checkpoint_saved: checkpointSaved,
+		});
+	});
+
+	app.get("/api/telemetry/events", (c) => {
+		if (!telemetryRef) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — inverted conditional to if (telemetryRef) { ... } return fallback pattern.

packages/daemon/src/routes/telemetry-routes.ts
```diff
@@ -0,0 +1,467 @@
...
+		const event = c.req.query("event") as TelemetryEventType | undefined;
+		const since = c.req.query("since");
+		const until = c.req.query("until");
+		const limit = Number.parseInt(c.req.query("limit") ?? "100", 10);
+		const events = telemetryRef.query({ event, since, until, limit });
+		return c.json({ events, enabled: true });
+	});
+
+	app.get("/api/telemetry/stats", (c) => {
+		if (!telemetryRef) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in 77a689bd — restructured /api/telemetry/stats with the same inverted conditional. Defensive guard preserved for runtime safety.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
-		embeddings: [],
-		count: 0,
-		total: 0,
-		limit,
-		offset,
-		hasMore: false,
-		error,
-	};
-}
+app.use("/api/repair/*", async (c, next) => {
```
PR-Reviewer-Ant, line 512: **startSessionCleanup() is called at module scope, before main() initializes the DB.** This runs when the ES module is first imported — before `initDbAccessor(MEMORY_DB, ...)` is called in `main()`. If `startSessionCleanup` issues any DB queries on startup (very likely for a session cleanup worker), it will call `getDbAccessor()` on an uninitialized accessor and either crash or silently no-op. Move this call inside `main()` after `initDbAccessor()`, mirroring how `initCheckpointFlush`, `startSchedulerWorker`, and other lifecycle functions are called.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
+});
+app.use("/api/hook/remember", async (c, next) => {
+	return requirePermission("remember", authConfig)(c, next);
+});
+app.use("/api/memory/recall", async (c, next) => {
+	return requirePermission("recall", authConfig)(c, next);
+});
+app.use("/api/memory/search", async (c, next) => {
+	return requirePermission("recall", authConfig)(c, next);
+});
```
PR-Reviewer-Ant, line 400: **The `/api/auth/token` admin permission guard is missing.** The original daemon.ts registered:
```ts
app.use("/api/auth/token", async (c, next) => {
  const perm = requirePermission("admin", authConfig);
  const rate = requireRateLimit("admin", authAdminLimiter, authConfig);
  await perm(c, async () => { await rate(c, next); });
});
```
This guard is absent from the new daemon.ts permission guards section. The PR description states 'all permission guards remain in daemon.ts unchanged', but this one does not. If it was not re-added in `misc-routes.ts` (which is not shown in this diff), the token creation endpoint loses both its admin permission check and its rate limit — allowing any authenticated caller to mint tokens. Verify it exists in `misc-routes.ts` and confirm the `authAdminLimiter` reference works correctly across module boundaries given that `authAdminLimiter` is a module-level `let` in daemon.ts.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
+// ============================================================================
+// Hono App
+// ============================================================================
 
-	// Copy agent.yaml so the shadow daemon uses real config (auth, pipeline flags, etc.)
-	// rather than built-in defaults.
-	const mainCfg = join(agentsDir, "agent.yaml");
-	const shadowCfg = join(shadowRoot, "agent.yaml");
-	if (existsSync(mainCfg)) copyFileSync(mainCfg, shadowCfg);
+export const app = new Hono();
```
PR-Reviewer-Ant, line 206: `recordPredictorLatency` changed its `operation` parameter type from `"predictor_score" | "predictor_train"` to `string`. If `analyticsCollector.recordLatency` internally dispatches on these literal keys (e.g., to separate latency buckets), widening to `string` will silently accept arbitrary strings from callers without a type error. This is a type-safety regression for an exported function. The narrower union should be preserved or re-exported from `routes/state.ts` where `analyticsCollector` lives.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
+let shuttingDown = false;
+let httpServer: ReturnType<typeof createAdaptorServer> | null = null;
+let dreamingWorkerHandle: DreamingWorkerHandle | null = null;
+let shadowProcess: ChildProcess | null = null;
+let predictorClientRef: PredictorClient | null = null;
+let embeddingTrackerHandle: EmbeddingTrackerHandle | null = null;
+let skillReconcilerHandle: ReturnType<typeof startReconciler> | null = null;
+let structuralBackfillTimer: ReturnType<typeof setTimeout> | null = null;
+let telemetryRef: TelemetryCollector | undefined;
+let heartbeatTimer: ReturnType<typeof setInterval> | undefined;
```
PR-Reviewer-Ant, line 191: `commitPending` and `commitTimer` are declared as module-level `let` variables and are cleaned up in `cleanup()`, but all auto-commit scheduling logic (`scheduleAutoCommit`) moved to `routes/git-sync.ts`. These variables are never set to non-default values in daemon.ts itself — the cleanup block that clears them is a no-op. Either the git-sync module maintains its own timer state (in which case these are dead) or the cleanup of the actual commit timer is now missing from the shutdown sequence. Either way, the intent should be made explicit.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `77a689bd`

[Automated Review — pr-reviewer] Found one confirmed pre-call initialization bug and one missing permission guard that contradicts the PR's stated invariants. Two minor dead-code/type issues also flagged.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The startSessionCleanup() module-scope call is directly visible in the changed daemon.ts file content. The missing /api/auth/token permission guard is provable from the changed file: the PR description claims 'all permission guards remain in daemon.ts unchanged' but that guard is absent from the new daemon.ts permission guards section. The one uncertainty is whether the guard was re-added inside misc-routes.ts, which is not included in the diff.

packages/daemon/src/routes/pipeline-routes.ts
```diff
@@ -0,0 +1,533 @@
...
+		}
+
+		const contentType = c.req.header("content-type") ?? "";
+		let mode: "compact" | "incremental" = "incremental";
+		if (contentType.includes("application/json")) {
+			const raw: unknown = await c.req.json().catch(() => null);
+			if (raw === null) {
+				return c.json({ error: "Malformed JSON body" }, 400);
+			}
+			if (typeof raw === "object" && raw !== null) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in bf131ccd — removed redundant `raw !== null` check since the null case is already handled by the early return above.

packages/daemon/src/routes/telemetry-routes.ts
```diff
@@ -0,0 +1,467 @@
...
+
+		return c.json({
+			...result,
+			checkpoint_path: checkpointPath,
+			checkpoint_saved: checkpointSaved,
+		});
+	});
+
+	app.get("/api/telemetry/events", (c) => {
+		if (telemetryRef) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in bf131ccd — changed to `if (!telemetryRef)` early-return pattern as recommended, with main logic path below.

packages/daemon/src/routes/telemetry-routes.ts
```diff
@@ -0,0 +1,467 @@
...
+			const until = c.req.query("until");
+			const limit = Number.parseInt(c.req.query("limit") ?? "100", 10);
+			const events = telemetryRef.query({ event, since, until, limit });
+			return c.json({ events, enabled: true });
+		}
+		return c.json({ events: [], enabled: false });
+	});
+
+	app.get("/api/telemetry/stats", (c) => {
+		if (telemetryRef) {
```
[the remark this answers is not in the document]
aaf2tbz: Fixed in bf131ccd — same `if (!telemetryRef)` early-return pattern applied to /api/telemetry/stats.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `bf131ccd`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `bf131ccd`

[Automated Bot Review] This is a large structural refactor. Found one timer-state ownership split that may silently leak an auto-commit timer on shutdown, one exported function signature widening that weakens type safety, and one module-level side effect that could affect test isolation.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The auto-commit timer split is directly evidenced in the daemon.ts diff: `commitPending` and `commitTimer` are declared as local `let` variables in daemon.ts (lines ~180-182) and cleared in `cleanup()`, but `scheduleAutoCommit` is now imported from `routes/git-sync.ts`. Whether `stopGitSyncTimer()` cancels the auto-commit debounce (not just the periodic sync interval) cannot be confirmed from the diff alone—the git-sync.ts file is truncated. The `recordPredictorLatency` signature change is directly readable in the diff. The `startSessionCleanup()` placement is directly visible.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
-			name?: unknown;
-		}>
-	).some((column) => column.name === "session_id");
-	return hasMemoriesSessionIdColumnCache;
-}
+let shuttingDown = false;
+let httpServer: ReturnType<typeof createAdaptorServer> | null = null;
+let dreamingWorkerHandle: DreamingWorkerHandle | null = null;
+let shadowProcess: ChildProcess | null = null;
+let predictorClientRef: PredictorClient | null = null;
```
PR-Reviewer-Ant, line 186: **Auto-commit timer ownership split — potential shutdown leak.**

`commitPending` and `commitTimer` are declared here and cleared in `cleanup()`, but nothing in daemon.ts ever sets them anymore: `scheduleAutoCommit` was moved to `routes/git-sync.ts` and almost certainly maintains its own internal timer state. As a result, the cleanup block:
```ts
if (commitTimer) { clearTimeout(commitTimer); commitTimer = null; }
commitPending = false;
```
is a no-op — the real debounce timer lives in git-sync.ts and only gets cancelled if `stopGitSyncTimer()` explicitly cancels it too. If `stopGitSyncTimer` only clears the periodic sync interval and not the auto-commit debounce, a pending `git commit` can fire after `closeDbAccessor()` returns, risking a write to a closed or partially-torn-down database. Verify that `stopGitSyncTimer` cancels the auto-commit debounce, or move `commitPending`/`commitTimer` into git-sync.ts where they belong and export a `cancelAutoCommit()` for cleanup.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
-	model: string;
-	available: boolean;
-	modelCached?: boolean;
-	dimensions?: number;
-	base_url: string;
-	error?: string;
-	checkedAt: string;
-}
+app.use("/api/memory/remember", async (c, next) => {
+	return requirePermission("remember", authConfig)(c, next);
```
PR-Reviewer-Ant, line 387: **`startSessionCleanup()` called at module scope before `main()`.**

This starts a recurring interval timer the moment the module is imported, not when the daemon is intentionally started. While session cleanup currently uses only in-memory maps (no DB), calling it at module load time breaks test isolation: any test that imports this module will start a live background timer. The original code called this inside `main()`. This should be moved back inside `main()` or the setup functions that follow it.

packages/daemon/src/daemon.ts
```diff
@@ -271,10826 +162,613 @@ import {
...
-const OPENCLAW_STALE_MS = 10 * 60 * 1000; // 10 minutes
-
-// Prevents concurrent UMAP computations for the same dimension count
-const projectionInFlight = new Map<number, Promise<void>>();
-const projectionErrors = new Map<number, { message: string; expires: number }>();
-const PROJECTION_ERROR_TTL_MS = 30_000;
-let hasMemoriesSessionIdColumnCache: boolean | null = null;
-
-// Auth state — initialized lazily in main(), but middleware reads from here
+const OPENCLAW_STALE_MS = 10 * 60 * 1000;
```
PR-Reviewer-Ant: **`recordPredictorLatency` operation parameter type widened from discriminated union to `string`.**

Original signature: `operation: "predictor_score" | "predictor_train"`. New signature: `operation: string`. All existing callers pass literal strings that happen to be valid, so no runtime breakage — but the narrower union type was serving as compile-time documentation of which analytics buckets exist. Widening to `string` means typos in future callers (`"predictor_Score"`, `"predictor-train"`) will silently produce orphaned analytics entries rather than a type error. Restore the union type or document that the expansion is intentional.

packages/daemon/src/daemon.ts
```diff
@@ -4,141 +4,57 @@
...
-	realpathSync,
-	statSync,
-	unlinkSync,
-	writeFileSync,
-} from "node:fs";
+import { appendFileSync, copyFileSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync, readdirSync, statSync, unlinkSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { basename, dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
```
PR-Reviewer-Ant, line 14: `type serve` is imported from `@hono/node-server` but the `httpServer` variable type was changed to `ReturnType<typeof createAdaptorServer>`. If `type serve` is no longer referenced, this is an unused import that biome should flag. Verify with `bun run lint` on this file specifically.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `2a8000f9`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -271,10722 +173,955 @@ import {
 	stopUpdateTimer,
```
PR-Reviewer-Ant, line 171: **Auto-commit debounce leak on shutdown.** `cleanup()` calls `stopGitSyncTimer()` (imported from `routes/git-sync.ts`) and clears `syncTimer` (the harness-sync debounce), but the auto-commit debounce timer created by `scheduleAutoCommit()` lives inside `git-sync.ts`. If `stopGitSyncTimer` only cancels the periodic sync interval and not the pending auto-commit debounce, a commit can fire after shutdown completes — writing to the workspace after the DB is closed and the PID file is gone. This was flagged in prior bot rounds and remains unverifiable without seeing the full `stopGitSyncTimer` implementation. At minimum, the PR should document (or assert in code) that `stopGitSyncTimer` drains both timers.

PR-Reviewer-Ant, line 171: **Missing regression test for the `startSessionCleanup` bug fix.** The PR description explicitly states: *'Fixed missing `startSessionCleanup` import that was dropped during extraction.'* Per AGENTS.md: *'Every bug fix needs a test that would fail before the fix.'* The PR checklist marks regression tests as N/A and categorizes this as a pure refactor, but a dropped import that caused a broken startup path is a bug. A test that verifies `startSessionCleanup` is invoked during `main()` (or that the session cleanup loop actually runs) is required by repo convention.

PR-Reviewer-Ant, line 171: **Orphaned constant `OPENCLAW_STALE_MS`.** `const OPENCLAW_STALE_MS = 10 * 60 * 1000` is defined at module scope but `openClawHeartbeat`, `buildOpenClawHealth()`, and the `/api/diagnostics/openclaw/heartbeat` handler that consumed it were extracted to route modules. If the extracted module imports `OPENCLAW_STALE_MS` from `./state.ts` or re-defines it locally, the copy in `daemon.ts` is dead code. If neither module re-exports it, the extracted handler is silently using a hardcoded inline value instead. Either move the constant to `state.ts` or confirm the consuming module imports it from the correct source.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +94,70 @@ import {
 	stopOpenCodeServer,
 } from "./pipeline/provider";
 import { resolveRuntimeModel } from "./pipeline/provider-resolution";
```
PR-Reviewer-Ant, line 95: **Likely dead imports: `shouldEnforceScope`, `backfillSkippedSessions`, `type serve`, `type Context`, `type AnalyticsCollector`.** None of these appear to be referenced in the visible body of the new `daemon.ts`. If the extracted route modules import them directly from their source modules (as expected in a clean extraction), these are dead imports. The PR reports 8 pre-existing Biome warnings and claims lint passes — if these are among them, that's fine, but they should be confirmed dead and removed rather than accumulated.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2a8000f9`

Automated bot review (pr-reviewer) — Structural refactor extracting 13 route modules from daemon.ts. Prior bot rounds flagged startSessionCleanup order and auth token guard; both appear resolved in head SHA 2a8000f. Three remaining issues warrant attention.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - daemon.ts is fully visible and the auth guard + startSessionCleanup placement are confirmed correct in the current head. However, the 13 extracted route modules (memory-routes.ts, git-sync.ts, repair-routes.ts, misc-routes.ts, etc.) are not in the diff beyond a few samples, so I cannot verify that all route-handler logic was moved verbatim, that permissions in extracted modules are consistent, or that git-sync.ts's stopGitSyncTimer cancels the auto-commit debounce in addition to the periodic sync interval — which is the crux of the timer-leak concern.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `f407f9dc`) now and I'll follow up shortly with feedback.

BusyBee3333 (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f407f9dc`

```pr-review-json
{}
```

packages/daemon/src/daemon.ts
```diff
@@ -271,10722 +171,954 @@ import {
...
-	// This ensures information near boundaries appears in both chunks,
-	// improving recall when search queries match boundary content.
-	const chunks: string[] = [];
-	for (let i = 0; i < filtered.length; i++) {
-		if (i < filtered.length - 1) {
-			const overlap = filtered[i + 1].slice(0, Math.floor(targetChars * 0.25));
-			chunks.push(`${filtered[i]} ${overlap}`.trim());
-		} else {
-			chunks.push(filtered[i]);
+	for (const candidate of candidates) {
```
PR-Reviewer-Ant, line 617: **Blocking — silent data loss in non-local auth modes.** `ingestMemoryMarkdown` posts to `/api/memory/remember` with no `Authorization` header. When `authConfig.mode !== 'local'`, the route guard added at line ~335 (`requirePermission('remember', authConfig)`) will reject these requests with 401/403. The response status is checked and logged, but the caller treats non-OK as a skipped chunk and returns 0 — so memory files are silently not ingested whenever auth is enabled. Same problem in `syncClaudeMemoryFile` (~line 690). The fix is to either: (a) use an internal bypass path that skips auth middleware, (b) pass a pre-minted admin token, or (c) call the underlying ingest function directly rather than HTTP self-call.

packages/daemon/src/daemon.ts
```diff
@@ -271,10722 +171,954 @@ import {
 	stopUpdateTimer,
```
PR-Reviewer-Ant, line 171: **Warning — removed concurrency guard.** The old code had `let pipelineTransition = false;` (visible in the diff removal block). The diff is truncated before the removed route handlers, so it cannot be confirmed whether this flag was used as a mutex around `restartPipelineRuntime`. If it was, concurrent agent.yaml change events (e.g., rapid edits) can now race two simultaneous stop→start cycles, leaving workers in a partially-started state (e.g., one `startPipeline` call races with a `stopPipeline` from the other). Please confirm the flag was dead code, or re-introduce a single-flight guard around `restartPipelineRuntime`.

PR-Reviewer-Ant, line 171: **Warning — typecheck explicitly skipped.** The PR readiness checklist leaves `bun run typecheck` unchecked and notes pre-existing failures. This is the highest-risk surface for a refactor of this scale: module boundary changes, re-exported types from routes/state.ts, and new `ReturnType<typeof startReconciler>` inference replacing the explicit `ReconcilerHandle` type. Pre-existing failures should not mask new type errors introduced here. At minimum, run `bun run --filter @signet/daemon tsc --noEmit --skipLibCheck` scoped to the changed files and confirm no new errors are introduced.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f407f9dc`

[Automated Review — pr-reviewer bot] PR #457 extracts 13 route modules from a 12k-line daemon.ts. The structural intent is sound and the regression test for the session-cleanup side-effect is a good addition. Three issues warrant attention before merge: a silent auth failure path in self-HTTP ingestion calls, a removed concurrency guard whose original usage cannot be confirmed from the truncated diff, and an explicitly skipped typecheck that covers the highest-risk surface of this refactor.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The unauthenticated self-request pattern is directly visible in the diff (lines in ingestMemoryMarkdown and syncClaudeMemoryFile). The pipelineTransition removal is in the changed daemon.ts state block. The truncated diff prevents confirming how pipelineTransition was used in the removed route handlers, and routes/state.ts content is not included, so setter/consumer parity cannot be fully verified.


- packages/daemon/src/daemon.ts:168 - **Nitpick — missing blank line.** The last import statement ends and `let authConfig` begins on the very next line without separation. All other major sections use `// ===` banners or blank lines. Minor but affects readability of the file boundary.

PR #460, opened 2026-04-03

aaf2tbz:
## Summary

Split `packages/daemon/src/daemon.ts` from **12,693 → ~2,500 lines** (80% reduction) by extracting route handlers into 13 dedicated modules under `routes/`.

This is a clean redo of #457, incorporating all review feedback and an upstream fix. Nine commits against latest `main`.

## Changes

- `routes/state.ts` owns shared mutable runtime state, updated via explicit setters from `daemon.ts`
- Auth endpoints (`/api/auth/whoami`, `/api/auth/token`) remain in `daemon.ts` with their guards
- Session cleanup called from `main()` after `initDbAccessor()`, not at module scope
- Git auto-commit debounce timer properly drained via `cancelAutoCommit()` inside `stopGitSyncTimer()`
- `recordPredictorLatency` preserves discriminated union type
- Incorporated upstream fix: `MEMORY_BACKUP_FILENAME_RE` for filtering generated backup markdown
- Added regression tests for session cleanup startup side effect and auth state synchronization

|---|---|---|
| `memory-routes.ts` | 2,574 | `/api/memory/*`, `/memory/*`, embeddings, documents |
| `hooks-routes.ts` | 1,183 | `/api/hooks/*`, cross-agent, synthesis |
| `utils.ts` | 1,107 | Shared parsers, validators, query builders |
| `git-sync.ts` | 729 | Git pull/push/sync, auto-commit, credential resolution |
| `misc-routes.ts` | 806 | Logs, config, identity, agents, update, tasks |
| `repair-routes.ts` | 665 | `/api/repair/*`, `/api/troubleshoot/*` |
| `pipeline-routes.ts` | 534 | `/api/status`, pipeline, dream, diagnostics |
| `knowledge-routes.ts` | 498 | `/api/knowledge/*` |
| `telemetry-routes.ts` | 467 | Analytics, telemetry, timeline, checkpoints |
| `state.ts` | 487 | Shared runtime state, config utilities, network helpers |
| `connectors-routes.ts` | 442 | `/api/connectors/*`, harnesses |
| `session-routes.ts` | 296 | `/api/sessions/*` |
| `secrets-routes.ts` | 241 | `/api/secrets/*` |

### Commit 2 (`9134253a`) — first review on commit 1

| # | Finding | Resolution |
|---|---------|------------|
| 1 | **Dual auth state.** | **Fixed.** Single `reloadAuthState()` in state.ts. |
| 2 | **Rate-limiter live-binding.** | **Mitigated by #1.** |
| 3 | **Missing auth reload test.** | **Added.** |
| 4 | **Dead imports.** | **Fixed.** Removed all. |

### Commit 3 (`ef578ea1`) — second review on commit 2

| # | Finding | Resolution |
|---|---------|------------|
| 5 | **Test weakness.** | **Fixed.** Two-reload idempotency + consistency. |
| 6 | **Dual `shuttingDown`.** | **Fixed.** Import from state.ts. |
| 7 | **Dead `createProviderTracker`/`DiagnosticsReport`.** | **Fixed.** Removed. |
| 8 | **Scheduler handle leak.** | **Fixed.** Added stop to cleanup. |

### Commit 4 (`ee4fdf47`) — third review on commit 3

| # | Finding | Resolution |
|---|---------|------------|
| 9 | **Buffer reference equality in test.** | **Fixed.** `toString('hex')`. |
| 10 | **Unauthenticated self-HTTP calls.** | **Out of scope.** Pre-existing. |
| 11 | **Unbounded `syncedClaudeMemories`.** | **Out of scope.** Pre-existing. |

### Commit 5 (`2ac5148c`) — fourth review on commit 4

| # | Finding | Resolution |
|---|---------|------------|
| 12 | **`schedulerHandle.stop()` not awaited.** | **Fixed.** `try { await ... } catch {}`. |
| 13 | **Test only exercises local mode.** | **Fixed.** Token-mode test with temp dir. |

### Commit 6 (`c16cece0`) — fifth review on commit 5

| # | Finding | Resolution |
|---|---------|------------|
| 14 | **Unused `existsSync` import.** | **Fixed.** Removed. |
| 15 | **Token-mode test only asserts no-throw.** | **Commented.** Bun test-isolation limitation. |
| 16 | **Session cleanup test first-import only.** | **Commented.** ES module caching expected. |
| 17 | **Bun live-binding may affect runtime.** | **Commented.** Same module instance at runtime. |
| 18 | **`authRecallLlmLimiter` not in daemon.ts.** | **Verified.** In memory-routes.ts. |

### Commit 7 (`b470b2eb`) — sixth review on commit 6

| # | Finding | Resolution |
|---|---------|------------|
| 19 | **`skillReconcilerHandle.stop()` not awaited.** | **Fixed.** `try { await ... } catch {}`. |
| 20 | **`pipelineTransition` variable missing.** | **Verified.** In state.ts, used by pipeline-routes.ts. |
| 21 | **`afterEach` should be `beforeEach`.** | **Fixed.** Changed to `beforeEach`. |

### Commit 8 (`9252bf93`) — seventh review on commit 7

| # | Finding | Resolution |
|---|---------|------------|
| 22 | **`restartPipelineRuntimeRef` unwired.** | **Fixed.** Added `setRestartPipelineRuntime(...)`. |
| 23 | **Unverified typecheck coverage.** | **Verified.** 433 errors on both main and branch — zero new. |
| 24 | **Inconsistent scheduler import specifiers.** | **Fixed.** Consolidated to `"./scheduler/index.js"`. |

### Commit 9 (`e915bc7e`) — eighth review on commit 8

| # | Finding | Resolution |
|---|---------|------------|
| 25 | **Blocking: wrong function registered.** `startPipelineRuntime` skips stop, leaking workers/providers. | **Fixed.** Changed to `restartPipelineRuntime`. |
| 26 | **Idempotency test trivially true in local mode.** | **Fixed.** Explicit `toBeNull()` assertion. |

### No new commit needed — ninth review on commit 9

| # | Finding | Resolution |
|---|---------|------------|
| 27 | **`setupStaticServing()` module-scope side effect.** Log noise during test imports. | **Out of scope.** Pre-existing in `main` (line 9893). Deferring to `main()` would be a separate refactor. |
| 28 | **`SCRIPTS_DIR` import in connectors-routes.ts.** Verify it's consumed. | **Verified.** Used at `connectors-routes.ts:391` for harness config generation. |
| 29 | **Unauditable scoping in extracted modules.** `agent_id`/`visibility` threading can't be verified from diff. | **Pre-existing.** Extracted handlers are verbatim copies from `main` — scoping is identical to the monolith. Human spot-check recommended at merge. |

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Security checks applied to admin/mutation endpoints — `/api/auth/token` guard preserved in `daemon.ts`; other admin guards preserved
- [x] Docs updated for API/spec/status changes — no API/spec/schema change beyond restoring extracted auth wiring
- [x] Regression tests added for each bug fix — session cleanup startup; auth reload idempotency + consistency; token-mode reload
- [x] Lint/typecheck/tests pass locally — touched-file Biome passes; daemon-scoped tsc: 433 errors on both main and branch (zero new); focused daemon tests pass (84/84)

N/A — no migrations touched.

## Testing

- [x] `bun test packages/daemon/src/daemon-refactor.test.ts packages/daemon/src/auth/auth.test.ts packages/daemon/src/session-api.test.ts packages/daemon/src/update-route.test.ts`
- [x] `bunx biome check packages/daemon/src/daemon.ts packages/daemon/src/routes/state.ts packages/daemon/src/routes/hooks-routes.ts packages/daemon/src/routes/git-sync.ts packages/daemon/src/session-tracker.ts packages/daemon/src/daemon-refactor.test.ts`
- [x] `bunx tsc --noEmit --project packages/daemon/tsconfig.json` — 433 errors on main, 433 on branch (zero new)

## AI disclosure

## Notes

Shared mutable daemon runtime state is now centralized in `routes/state.ts` with a single `reloadAuthState()` entry point for auth mutations, so extracted route modules observe live state without reintroducing module-scope side effects.

Closes #457.

### DIFF
--- a/packages/daemon/src/daemon-refactor.test.ts
+++ b/packages/daemon/src/daemon-refactor.test.ts
@@ -1,0 +1,96 @@
+import { beforeEach, describe, expect, it } from "bun:test";
+import { mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { isSessionCleanupRunning, stopSessionCleanup } from "./session-tracker";
+
+describe("daemon route extraction refactor", () => {
+	beforeEach(() => {
+		stopSessionCleanup();
+	});
+
+	// NOTE: This test is only meaningful on the first run in a process —
+	// Bun/Node cache ES modules, so subsequent imports are no-ops. It
+	// guards against module-level side effects at import time (e.g.
+	// calling startSessionCleanup() during top-level evaluation).
+	it("does not start session cleanup when daemon is imported for route registration", async () => {
+		expect(isSessionCleanupRunning()).toBe(false);
+		await import("./daemon");
+		expect(isSessionCleanupRunning()).toBe(false);
+	});
+
+	it("reloadAuthState is idempotent and produces consistent auth state", async () => {
+		const state = await import("./routes/state.js");
+
+		state.reloadAuthState(state.AGENTS_DIR);
+		const mode1 = state.authConfig.mode;
+		const secret1 = state.authSecret;
+
+		state.reloadAuthState(state.AGENTS_DIR);
+		const mode2 = state.authConfig.mode;
+		const secret2 = state.authSecret;
+
+		expect(mode2).toBe(mode1);
+		expect(secret2?.toString("hex")).toBe(secret1?.toString("hex"));
+
+		// In local mode (the test environment default), authSecret is always
+		// null. Make this invariant explicit rather than relying on the
+		// trivially-true toString comparison above.
+		if (state.authConfig.mode === "local") {
+			expect(state.authSecret).toBeNull();
+		} else {
+			expect(state.authSecret).not.toBeNull();
+		}
+	});
+
+	// Exercises the non-local auth path (token mode) end-to-end:
+	// reloadAuthState reads agent.yaml, parses mode=token, and calls
+	// loadOrCreateSecret to populate authSecret. A regression where
+	// reloadAuthState silently swallowed an error would leave authSecret
+	// null in token mode, causing the auth middleware to 503 all requests.
+	//
+	// Limitation: Bun's ES module live bindings don't reliably propagate
+	// export let reassignments from function calls in the test runner,
+	// so we assert the function completes without error (a throw would
+	// indicate a parsing or secret-loading failure). At runtime in the
+	// daemon process, reloadAuthState is called from the file watcher
+	// and startPipelineRuntime where live bindings propagate correctly
+	// because state.ts and daemon.ts share the same module instance.
+	it("reloadAuthState completes without error in token mode", async () => {
+		const { reloadAuthState } = await import("./routes/state.js");
+
+		const tmpDir = join(tmpdir(), `signet-test-auth-${Date.now()}`);
+		mkdirSync(join(tmpDir, "memory"), { recursive: true });
+		mkdirSync(join(tmpDir, ".daemon"), { recursive: true });
+		writeFileSync(join(tmpDir, ".daemon", "auth-secret"), "test-secret-key-for-auth-mode");
+		writeFileSync(
+			join(tmpDir, "agent.yaml"),
+			[
+				"auth:",
+				"  mode: token",
+				"  rateLimits:",
+				"    forget:",
+				"      windowMs: 60000",
+				"      max: 30",
+				"    modify:",
+				"      windowMs: 60000",
+				"      max: 60",
+				"    batchForget:",
+				"      windowMs: 60000",
+				"      max: 5",
+				"    admin:",
+				"      windowMs: 60000",
+				"      max: 10",
+				"    recallLlm:",
+				"      windowMs: 60000",
+				"      max: 60",
+			].join("\n"),
+		);
+
+		try {
+			reloadAuthState(tmpDir);
+		} finally {
+			rmSync(tmpDir, { recursive: true, force: true });
+		}
+	});
+});
--- a/packages/daemon/src/daemon.ts
+++ b/packages/daemon/src/daemon.ts
@@ -4,142 +4,55 @@
  * Background service for memory, API, and dashboard hosting
  */
 
-import type { Database } from "bun:sqlite";
-import { type ChildProcess, execFileSync, spawn } from "node:child_process";
+import type { ChildProcess } from "node:child_process";
+import { execFileSync, spawn } from "node:child_process";
 import { createHash } from "node:crypto";
-import { copyFileSync } from "node:fs";
-import {
-	appendFileSync,
-	existsSync,
-	mkdirSync,
-	readFileSync,
-	readdirSync,
-	realpathSync,
-	statSync,
-	unlinkSync,
-	writeFileSync,
-} from "node:fs";
+import { appendFileSync, copyFileSync } from "node:fs";
+import { existsSync, mkdirSync, readFileSync, readdirSync, statSync, unlinkSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { basename, dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
-import { createAdaptorServer, type serve } from "@hono/node-server";
+import { createAdaptorServer } from "@hono/node-server";
 import { serveStatic } from "@hono/node-server/serve-static";
 import {
 	type AgentDefinition,
-	CONNECTOR_PROVIDERS,
-	type ConnectorConfig,
-	SIGNET_GIT_PROTECTED_PATHS,
-	type SyncCursor,
 	buildArchitectureDoc,
-	findSignetForgeBinary,
-	keywordSearch,
-	mergeSignetGitignoreEntries,
 	networkModeFromBindHost,
 	normalizeAgentRosterEntry,
 	parseSimpleYaml,
-	readNetworkMode,
-	readPipelinePauseState,
-	resolveNetworkBinding,
-	resolveSignetForgeManagedPath,
-	setPipelinePaused,
 	stripSignetBlock,
-	vectorSearch,
 } from "@signet/core";
 import { watch } from "chokidar";
-import { type Context, Hono } from "hono";
+import { Hono } from "hono";
 import { cors } from "hono/cors";
-import { logger as honoLogger } from "hono/logger";
-import { getAgentScope, resolveAgentId } from "./agent-id";
-import { type AnalyticsCollector, type ErrorStage, createAnalyticsCollector } from "./analytics";
+import { resolveAgentId } from "./agent-id";
 import {
-	type AuthConfig,
-	AuthRateLimiter,
 	type TokenRole,
 	type TokenScope,
 	checkScope,
 	createAuthMiddleware,
 	createToken,
-	loadOrCreateSecret,
-	parseAuthConfig,
 	requirePermission,
 	requireRateLimit,
-	requireScope,
 } from "./auth";
 import { bindWithRetry } from "./bind-with-retry";
 import { migrateConfig } from "./config-migration";
-import { createFilesystemConnector } from "./connectors/filesystem";
-import {
-	getConnector,
-	listConnectors,
-	registerConnector,
-	removeConnector,
-	updateConnectorStatus,
-	updateCursor,
-} from "./connectors/registry";
+import { listConnectors } from "./connectors/registry";
 import { normalizeAndHashContent } from "./content-normalization";
-import {
-	type AgentMessage,
-	type AgentMessageType,
-	clearAllPresence,
-	createAgentMessage,
-	getAgentPresenceForSession,
-	isMessageVisibleToAgent,
-	listAgentMessages,
-	listAgentPresence,
-	relayMessageViaAcp,
-	removeAgentPresence,
-	subscribeCrossAgentEvents,
-	touchAgentPresence,
-	upsertAgentPresence,
-} from "./cross-agent";
+import { clearAllPresence } from "./cross-agent";
 import { closeDbAccessor, getDbAccessor, initDbAccessor } from "./db-accessor";
-import { syncVecDeleteBySourceId, syncVecInsert, vectorToBlob } from "./db-helpers";
-import {
-	type DiagnosticsReport,
-	type PredictorHealthParams,
-	createProviderTracker,
-	getDiagnostics,
-	getPredictorHealth,
-} from "./diagnostics";
-import {
-	fetchEmbedding,
-	resolveEmbeddingApiKey,
-	resolveEmbeddingBaseUrl,
-	resolveOllamaUrl,
-	setNativeFallbackToOllama,
-} from "./embedding-fetch";
-import { buildEmbeddingHealth } from "./embedding-health";
+import { syncVecDeleteBySourceId, syncVecInsert } from "./db-helpers";
+import { fetchEmbedding, setNativeFallbackToOllama } from "./embedding-fetch";
 import { type EmbeddingTrackerHandle, startEmbeddingTracker } from "./embedding-tracker";
 import { getAllFeatureFlags, initFeatureFlags } from "./feature-flags";
 import { writeFileIfChangedAsync } from "./file-sync";
-import { walkImpact } from "./graph-impact";
 import { syncAgentWorkspaces } from "./identity-sync";
-import { linkMemoryToEntities } from "./inline-entity-linker";
-import {
-	getAttributesForAspectFiltered,
-	getEntityAspectsWithCounts,
-	getEntityDependenciesDetailed,
-	getEntityHealth,
-	getKnowledgeEntityDetail,
-	getKnowledgeGraphForConstellation,
-	getKnowledgeStats,
-	getPinnedEntities,
-	listKnowledgeEntities,
-	pinEntity,
-	resolveNamedEntity,
-	unpinEntity,
-} from "./knowledge-graph";
 import { closeLlmProvider, getLlmProvider, initLlmProvider } from "./llm";
-import { type LogCategory, type LogEntry, logger } from "./logger";
+import { type LogEntry, logger } from "./logger";
 import { type EmbeddingConfig, type ResolvedMemoryConfig, loadMemoryConfig } from "./memory-config";
-import { type RecallParams, hybridRecall } from "./memory-search";
-import { buildMemoryTimeline } from "./memory-timeline";
-import { ONEPASSWORD_SERVICE_ACCOUNT_SECRET, importOnePasswordSecrets, listOnePasswordVaults } from "./onepassword.js";
-import { recordPathFeedback } from "./path-feedback";
 import {
 	DEFAULT_RETENTION,
 	enqueueDocumentIngestJob,
-	enqueueExtractionJob,
 	ensureRetentionWorker,
 	getDreamingPasses,
 	getDreamingState,
@@ -153,16 +66,8 @@ import {
 	stopPipeline,
 } from "./pipeline";
 import { AlreadyRunningError, type DreamingWorkerHandle, startDreamingWorker } from "./pipeline/dreaming-worker";
-import { getFeedbackTelemetry } from "./pipeline/aspect-feedback";
-import { clusterEntities } from "./pipeline/community-detection";
-import { deadLetterExtractionJob, deadLetterPendingExtractionJobs } from "./pipeline/extraction-fallback";
-import { getGraphBoostIds } from "./pipeline/graph-search";
-import {
-	getTraversalStatus,
-	invalidateTraversalCache,
-	resolveFocalEntities,
-	traverseKnowledgeGraph,
-} from "./pipeline/graph-traversal";
+import { deadLetterPendingExtractionJobs } from "./pipeline/extraction-fallback";
+import { getTraversalStatus, invalidateTraversalCache } from "./pipeline/graph-traversal";
 import {
 	getAvailableModels,
 	getModelsByProvider,
@@ -184,80 +89,85 @@ import {
 	stopOpenCodeServer,
 } from "./pipeline/provider";
 import { resolveRuntimeModel } from "./pipeline/provider-resolution";
-import { type RerankCandidate, noopReranker, rerank } from "./pipeline/reranker";
-import { createEmbeddingReranker } from "./pipeline/reranker-embedding";
-import { type PredictorClient, createPredictorClient, resolvePredictorCheckpointPath } from "./predictor-client";
+import { startReconciler } from "./pipeline/skill-reconciler";
+import { type PredictorClient, createPredictorClient } from "./predictor-client";
 import { detectDrift } from "./predictor-comparison";
-import {
-	getComparisonsByEntity,
-	getComparisonsByProject,
-	listComparisons,
-	listTrainingRuns,
-} from "./predictor-comparisons";
 import { getPredictorState } from "./predictor-state";
+import { type RepairContext, structuralBackfill } from "./repair-actions";
 import {
-	type RepairContext,
-	type RepairResult,
-	backfillSkippedSessions,
-	checkFtsConsistency,
-	cleanOrphanedEmbeddings,
-	createRateLimiter,
-	deduplicateMemories,
-	findDeadMemories,
-	forgetDeadMemories,
-	getDedupStats,
-	getEmbeddingGapStats,
-	pruneChunkGroupEntities,
-	pruneSingletonExtractedEntities,
-	reclassifyEntities,
-	reembedMissingMemories,
-	releaseStaleLeases,
-	requeueDeadJobs,
-	resyncVectorIndex,
-	structuralBackfill,
-	triggerRetentionSweep,
-} from "./repair-actions";
-import {
-	resolveScopedAgent,
-	resolveScopedProject as resolveScopedProjectForClaims,
-	shouldEnforceScope,
-} from "./request-scope";
-import {
-	CRON_PRESETS,
-	computeNextRun,
-	isHarnessAvailable,
-	resolveSkillPrompt,
-	startSchedulerWorker,
-	validateCron,
-} from "./scheduler";
-import { emitTaskStream, getTaskStreamSnapshot, subscribeTaskStream } from "./scheduler/task-stream";
-import { deleteSecret, execWithSecrets, getSecret, hasSecret, listSecrets, putSecret } from "./secrets.js";
-import {
-	flushPendingCheckpoints,
-	getCheckpointsByProject,
-	getCheckpointsBySession,
-	initCheckpointFlush,
-	pruneCheckpoints,
-	redactCheckpointRow,
-} from "./session-checkpoints";
-import { parseFeedback, recordAgentFeedback } from "./session-memories";
+	AGENTS_DIR,
+	ALLOWED_ORIGINS,
+	BIND_HOST,
+	CURRENT_VERSION,
+	DAEMON_DIR,
+	HOST,
+	INTERNAL_SELF_HOST,
+	LOG_DIR,
+	MEMORY_DB,
+	NETWORK_MODE,
+	PID_FILE,
+	PORT,
+	analyticsCollector,
+	authAdminLimiter,
+	authBatchForgetLimiter,
+	authConfig,
+	authCrossAgentMessageLimiter,
+	authForgetLimiter,
+	authModifyLimiter,
+	authSecret,
+	bindAbort,
+	invalidateDiagnosticsCache,
+	isAllowedOrigin,
+	isManagedOpenCodeLocalEndpoint,
+	normalizeRuntimeBaseUrl,
+	providerRuntimeResolution,
+	providerTracker,
+	queueExtractionJob,
+	readEnvTrimmed,
+	redactUrlForLogs,
+	reloadAuthState,
+	repairLimiter,
+	setCheckpointPruneTimer,
+	setHeartbeatTimer,
+	setPredictorClientRef,
+	setRestartPipelineRuntime,
+	setShuttingDown,
+	setTelemetryRef,
+	shuttingDown,
+} from "./routes/state.js";
+import { isHarnessAvailable, startSchedulerWorker } from "./scheduler/index.js";
+import { getSecret, hasSecret } from "./secrets.js";
+import { flushPendingCheckpoints, initCheckpointFlush, pruneCheckpoints } from "./session-checkpoints";
+import { releaseAllSessions, startSessionCleanup, stopSessionCleanup } from "./session-tracker";
 import { createSingleFlightRunner } from "./single-flight-runner";
 import { closeSynthesisProvider, initSynthesisProvider } from "./synthesis-llm";
-import { readScopedTask, readTaskAgentId } from "./task-scope";
 import { type TelemetryCollector, type TelemetryEventType, createTelemetryCollector } from "./telemetry";
-import { expandTemporalNode } from "./temporal-expand";
-import { isNoiseSession } from "./session-noise";
-import { upsertThreadHead } from "./thread-heads";
-import { type TimelineSources, buildTimeline } from "./timeline";
-import {
-	type MutationContext,
-	txFinalizeAccessAndHistory,
-	txForgetMemory,
-	txIngestEnvelope,
-	txModifyMemory,
-	txRecoverMemory,
-} from "./transactions";
-import { cacheProjection, computeProjection, computeProjectionForQuery, getCachedProjection } from "./umap-projection";
+import { closeWidgetProvider, initWidgetProvider } from "./widget-llm";
+
+import { mountMcpRoute } from "./mcp/route.js";
+import { mountAppTrayRoutes } from "./routes/app-tray.js";
+import { mountChangelogRoutes } from "./routes/changelog.js";
+import { registerConnectorRoutes } from "./routes/connectors-routes.js";
+import { mountEventBusRoutes } from "./routes/event-bus.js";
+import { getGitStatus, gitSync, scheduleAutoCommit, startGitSyncTimer, stopGitSyncTimer } from "./routes/git-sync.js";
+import { registerHooksRoutes } from "./routes/hooks-routes.js";
+import { registerKnowledgeRoutes } from "./routes/knowledge-routes.js";
+import { mountMarketplaceReviewsRoutes } from "./routes/marketplace-reviews.js";
+import { mountMarketplaceRoutes } from "./routes/marketplace.js";
+import { mountMcpAnalyticsRoutes } from "./routes/mcp-analytics.js";
+import { registerMemoryRoutes } from "./routes/memory-routes.js";
+import { registerMiscRoutes } from "./routes/misc-routes.js";
+import { mountOsAgentRoutes } from "./routes/os-agent.js";
+import { mountOsChatRoutes } from "./routes/os-chat.js";
+import { registerPipelineRoutes } from "./routes/pipeline-routes.js";
+import { registerRepairRoutes } from "./routes/repair-routes.js";
+import { registerSecretRoutes } from "./routes/secrets-routes.js";
+import { registerSessionRoutes } from "./routes/session-routes.js";
+import { mountSkillAnalyticsRoutes } from "./routes/skill-analytics.js";
+import { mountSkillsRoutes, setFetchEmbedding } from "./routes/skills.js";
+import { registerTelemetryRoutes } from "./routes/telemetry-routes.js";
+import { getConfiguredProviderHints } from "./routes/utils.js";
+import { mountWidgetRoutes } from "./routes/widget.js";
 import {
 	MAX_UPDATE_INTERVAL_SECONDS,
 	MIN_UPDATE_INTERVAL_SECONDS,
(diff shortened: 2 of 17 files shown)


BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `d76f5c14`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `d76f5c14`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -271,10722 +171,957 @@ import {
 	stopUpdateTimer,
```
PR-Reviewer-Ant, line 183: **Dual auth state — fragile sync pattern.** `authConfig`, `authSecret`, and the five rate-limiter variables are declared here as local module-scope lets, AND mirrored into `routes/state.ts` via `setAuthConfig` / `setAuthSecret` / `setAuthRateLimiters`. The permission guards defined in this file close over the local variables; route modules use `routes/state.ts` copies. The file-watcher reload path (around line ~680 of the visible content) sets the locals first and then calls the setters — a pattern that works today but creates two sources of truth. If a future contributor adds a permission guard inside a route module and reads directly from state.ts while a concurrent request hits a guard in daemon.ts reading the local (or vice versa during a reload), the two copies can transiently disagree. Recommendation: either (a) remove the local declarations and have daemon.ts import everything from state.ts too, or (b) add a comment making the invariant explicit: 'local vars are always updated before calling setters, in the same synchronous block'.

packages/daemon/src/daemon.ts
```diff
@@ -271,10722 +171,957 @@ import {
...
+let authBatchForgetLimiter = new AuthRateLimiter(60_000, 5);
+let authAdminLimiter = new AuthRateLimiter(60_000, 10);
+let authRecallLlmLimiter = new AuthRateLimiter(60_000, 60);
 
-function normalizeBaseUrl(url: string | undefined): string | undefined {
-	if (!url) return undefined;
-	return url.endsWith("/") ? url.slice(0, -1) : url;
-}
+let shuttingDown = false;
+let httpServer: ReturnType<typeof createAdaptorServer> | null = null;
```
PR-Reviewer-Ant, line 187: **Rate-limiter live-binding assumption.** Route modules that import a rate-limiter from `routes/state.ts` (e.g. `authRecallLlmLimiter` likely used in `session-routes.ts`) will only see the new value after config reload if `state.ts` uses `export let` (which provides ES-module live bindings to importers) rather than exporting a wrapper object or constant. If state.ts exports these as `export const obj = { recallLlm: ... }` and the setter mutates the property, the binding is live. If it uses `export let recallLlm = ...` and the setter reassigns the export, the binding is also live. But if any route module captures the value at import time with `const limiter = recallLlmLimiter` (destructuring the live binding into a local const), the capture is frozen. This cannot be verified without seeing `routes/state.ts` and the consuming route files. A test that reloads auth config and verifies the rate-limiter enforces the new limit would close this gap.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,14 @@
+import { afterEach, describe, expect, it } from "bun:test";
+import { isSessionCleanupRunning, stopSessionCleanup } from "./session-tracker";
+
+describe("daemon route extraction refactor", () => {
+	afterEach(() => {
+		stopSessionCleanup();
+	});
+
+	it("does not start session cleanup when daemon is imported for route registration", async () => {
```
PR-Reviewer-Ant, line 16: The regression test covers only one behavioral invariant: no session cleanup on import. Per repo convention (AGENTS.md: 'every bug fix needs a test' / 'new features ship with tests that describe the contract'), a refactor of this scope warrants at least one test for the auth reload path — specifically that after a config change is processed by the file watcher, the permission guards and rate limiters in both daemon.ts and state.ts reflect the new values. Without it, the dual-state sync described above has no automated guard.

packages/daemon/src/daemon.ts
```diff
@@ -152,16 +70,8 @@ import {
...
-import { getGraphBoostIds } from "./pipeline/graph-search";
-import {
-	getTraversalStatus,
-	invalidateTraversalCache,
-	resolveFocalEntities,
-	traverseKnowledgeGraph,
-} from "./pipeline/graph-traversal";
+import { deadLetterPendingExtractionJobs } from "./pipeline/extraction-fallback";
+import { getTraversalStatus, invalidateTraversalCache } from "./pipeline/graph-traversal";
 import {
```
PR-Reviewer-Ant, line 166: **Possibly dead imports.** `blobToVector`, `chunkBySentence`, and `parseOptionalString` are imported from `./routes/utils.js`. In the visible portion of the new daemon.ts, `blobToVector` and `chunkBySentence` are not called (the local `chunkMarkdownHierarchically` is used instead for memory ingestion). `parseOptionalString` was used by the agents CRUD handlers which appear to have moved to `misc-routes.ts`. If these are genuinely unused in daemon.ts after the move, they are dead imports that will trigger Biome lint warnings. Verify with `bun run lint` and remove if unused.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d76f5c14`

[Automated Review — pr-reviewer bot] PR #460 refactors daemon.ts (12 693 → ~2 500 lines) by extracting 13 route modules. The structural split is clean and the stated goal is achieved. Two correctness concerns are flagged: a dual-source auth state pattern and an unverified live-binding assumption for hot-reloaded rate limiters. No security regressions are visible in the changed code.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The dual auth state concern is directly provable from daemon.ts lines 171–183 (local declarations) combined with the setAuthConfig/setAuthSecret/setAuthRateLimiters call sites in startPipelineRuntime and the file watcher. The live-binding concern depends on how routes/state.ts exports its rate-limiter variables, and that file is not included in the diff — only its public imports are visible. Confidence is medium rather than high because the route-module internals (how they consume state.ts exports) are not fully visible.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `9134253a`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9134253a`

[Automated Bot Review — pr-reviewer] This is a large structural refactor (12,693 → ~2,500 lines in daemon.ts) that extracts route handlers into 13 modules. The stated goals are achieved: auth state is consolidated into state.ts via a single reloadAuthState() entry point, session cleanup is no longer triggered at import time, and unused imports from the prior review are removed. No blocking correctness or security issues were found. Three findings below are worth discussing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The new daemon.ts and daemon-refactor.test.ts are fully visible in the diff. routes/state.ts is NOT included in the diff, so I cannot verify its internal implementation of reloadAuthState(), the shuttingDown export, or whether createProviderTracker and DiagnosticsReport moved there. The findings below are grounded in what is observable in daemon.ts and the test file.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,38 @@
...
+		expect(typeof state.authConfig.mode).toBe("string");
+
+		// authSecret must be consistent with mode.
+		if (state.authConfig.mode !== "local") {
+			expect(state.authSecret).not.toBeNull();
+		} else {
+			expect(state.authSecret).toBeNull();
+		}
+
+		// Verify reloadAuthState is the single update path — state.ts
```
PR-Reviewer-Ant: The final two assertions (`expect(state.authConfig.mode).toBe(configBefore)` and `expect(state.authSecret).toBe(secretBefore)`) test idempotency — that reloading the same on-disk config produces the same result — not that auth state synchronizes from state.ts to any other module. If `reloadAuthState` silently caught and swallowed an internal error and left values unchanged, both assertions would still pass. The test comment says it verifies "reloadAuthState is the single update path" but it cannot detect regressions where the reload ran but produced wrong values. Consider asserting a known state change: mutate a state.ts export directly (e.g., temporarily overwrite authConfig.mode), call reloadAuthState, and verify it is corrected. As-is the test provides weaker guarantees than the PR description claims.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
 	stopUpdateTimer,
```
PR-Reviewer-Ant, line 183: daemon.ts declares a local `let shuttingDown = false` and also imports `setShuttingDown` from routes/state.ts, setting both in `cleanup()`. The shutdown middleware on line 205 reads the local variable; if any extracted route module reads `shuttingDown` from state.ts, the two copies are structurally redundant and can diverge if one setter is missed in a future edit. This was not a problem in the old monolithic file. Consider driving the middleware off state.ts as well, or removing the local copy if state.ts owns the canonical value.

packages/daemon/src/daemon.ts
```diff
@@ -4,141 +4,55 @@
...
-	resolveOllamaUrl,
-	setNativeFallbackToOllama,
-} from "./embedding-fetch";
-import { buildEmbeddingHealth } from "./embedding-health";
+import { syncVecDeleteBySourceId, syncVecInsert } from "./db-helpers";
+import { type DiagnosticsReport, createProviderTracker } from "./diagnostics";
+import { fetchEmbedding, setNativeFallbackToOllama } from "./embedding-fetch";
 import { type EmbeddingTrackerHandle, startEmbeddingTracker } from "./embedding-tracker";
 import { getAllFeatureFlags, initFeatureFlags } from "./feature-flags";
 import { writeFileIfChangedAsync } from "./file-sync";
```
PR-Reviewer-Ant, line 48: `createProviderTracker` is imported from `./diagnostics` but `providerTracker` is also imported from `./routes/state.js` (line 121). If state.ts initializes `providerTracker` via `createProviderTracker()` at module scope (the expected pattern post-refactor), this import in daemon.ts is dead. Similarly `type DiagnosticsReport` is imported but the diagnostics cache that previously typed it has moved to state.ts. Both may generate lint warnings and add noise. Verify neither is actually referenced in the retained daemon.ts body and remove if unused.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
...
-				);
+				if (chunkTokens + paraTokens + 2 > maxTokens && chunkParas.length > 0) {
+					const text = currentHeader ? `${currentHeader}\n\n${chunkParas.join("\n\n")}` : chunkParas.join("\n\n");
+					results.push({
+						text,
+						tokenCount: chunkTokens,
+						header: currentHeader,
+						level: "paragraph",
+					});
+					chunkParas = [];
```
PR-Reviewer-Ant, line 918: `const schedulerHandle = startSchedulerWorker(getDbAccessor())` — the return value is never referenced again. If the scheduler has a shutdown path (e.g. `schedulerHandle.stop()`), it is not called in `cleanup()`. If the scheduler is designed to run until process exit this is fine, but it's inconsistent with how dreamingWorkerHandle, skillReconcilerHandle, and embeddingTrackerHandle are all explicitly stopped in cleanup. Worth confirming the scheduler worker self-terminates on process exit without needing explicit shutdown.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `ef578ea1`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,38 @@
...
+		const mode1 = state.authConfig.mode;
+		const secret1 = state.authSecret;
+
+		// Second reload — must produce identical values (idempotency).
+		state.reloadAuthState(state.AGENTS_DIR);
+		const mode2 = state.authConfig.mode;
+		const secret2 = state.authSecret;
+
+		expect(mode2).toBe(mode1);
+		expect(secret2).toBe(secret1);
```
PR-Reviewer-Ant: `expect(secret2).toBe(secret1)` uses `Object.is` (reference equality) to compare two `Buffer | null` values. If `reloadAuthState` allocates a new `Buffer` on each call — which is standard practice when reading a secret from disk — this assertion will always fail even when the content is identical, making the idempotency test permanently broken. Use `expect(secret2?.toString('hex')).toBe(secret1?.toString('hex'))` (or `.toEqual()`) to compare by value instead of by reference.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
...
 
-	try {
-		const raw = readFileSync(packageJsonPath, "utf8");
-		const parsed = JSON.parse(raw) as { version?: unknown };
-		return typeof parsed.version === "string" ? parsed.version : null;
-	} catch {
-		return null;
-	}
-}
+app.use("/api/secrets", async (c, next) => {
```
PR-Reviewer-Ant, line 521: `ingestMemoryMarkdown` (and `syncClaudeMemoryFile` a few hundred lines later) POST to `/api/memory/remember` with no `Authorization` header. That endpoint has `requirePermission('remember', authConfig)` applied as middleware. In any non-local auth mode (bearer / token), these daemon-internal calls will be rejected with 401/403, so file-watcher-triggered memory ingestion silently fails for the entire watcher lifecycle. The error surfaces only as a logged warning per chunk. A simple fix is to skip the self-HTTP path and call the persistence layer directly, or mint an internal service token at startup. If this pattern predates the PR, please confirm so this finding can be closed as out-of-scope.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
...
+	return requirePermission("remember", authConfig)(c, next);
+});
+app.use("/api/hook/remember", async (c, next) => {
+	return requirePermission("remember", authConfig)(c, next);
+});
+app.use("/api/memory/recall", async (c, next) => {
+	return requirePermission("recall", authConfig)(c, next);
+});
+app.use("/api/memory/search", async (c, next) => {
+	return requirePermission("recall", authConfig)(c, next);
```
PR-Reviewer-Ant, line 404: `const syncedClaudeMemories = new Set<string>()` accumulates SHA-256-derived chunk keys and is never pruned. Over the lifetime of a long-running daemon that watches many Claude projects, this set grows monotonically. Consider capping it (e.g., evict when size exceeds a threshold) or using the existing `ingestedMemoryFiles` content-hash map as the single deduplication mechanism, dropping `syncedClaudeMemories` entirely.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ef578ea1`

[Automated Bot Review] pr-reviewer: Refactor PR extracting daemon routes into 13 modules. Found one test reliability bug (Buffer reference equality), one latent correctness issue with unauthenticated self-HTTP calls in non-local auth mode, and one unbounded memory set. No new security regressions introduced in the auth surface. Code structure is sound.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff includes daemon.ts (full new content), the new test file, connectors-routes.ts, and git-sync.ts, which is enough to evaluate the auth wiring, test correctness, and watcher logic. However, routes/state.ts, memory-routes.ts, hooks-routes.ts, and the other 9 route modules are not shown, so I cannot verify live-binding reads inside those modules or confirm pipelineTransition removal had no side effects.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `ee4fdf47`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 183: `schedulerHandle.stop()` is not awaited, despite the type being `{ stop(): Promise<void> }`. The PR description calls fixing this handle leak a key improvement (finding #8), but the stop call is fire-and-forget. If the scheduler does async cleanup (e.g. flushing pending jobs), it may not complete before the process continues shutdown. Contrast with `embeddingTrackerHandle.stop()` and `predictorClientRef.stop()`, which are both `await`ed with try/catch in `stopPipelineRuntime`. `skillReconcilerHandle.stop()` is also un-awaited, so this is not new inconsistency introduced by this PR, but the scheduler was the explicit subject of the fix claim.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,38 @@
...
+		stopSessionCleanup();
+	});
+
+	it("does not start session cleanup when daemon is imported for route registration", async () => {
+		expect(isSessionCleanupRunning()).toBe(false);
+		await import("./daemon");
+		expect(isSessionCleanupRunning()).toBe(false);
+	});
+
+	it("reloadAuthState is idempotent and produces consistent auth state", async () => {
```
PR-Reviewer-Ant, line 22: The `reloadAuthState` idempotency test only exercises the mode of the test environment (almost certainly `'local'`). In local mode, `authSecret` is null both before and after reload, so `expect(secret2?.toString('hex')).toBe(secret1?.toString('hex'))` reduces to `expect(undefined).toBe(undefined)` — trivially true. The consistency assertion at line 31 similarly reduces to `expect(null).toBeNull()`. The non-local branch (`expect(state.authSecret).not.toBeNull()`) is never reached. This is understood and partly acknowledged in the PR description, but the test as written does not verify that `reloadAuthState` correctly populates `authSecret` when mode is not `'local'`, which is the case where a stale secret would be dangerous.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ee4fdf47`

Automated bot review (pr-reviewer). This is a large structural refactor with careful attention to the dual-state issues flagged in prior rounds. No blocking bugs found from the visible diff, but three concrete issues are worth discussion.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - daemon.ts and daemon-refactor.test.ts are fully visible. The 13 extracted route modules are named but their implementations are only partially shown (connectors-routes.ts and git-sync.ts appear in changed file contents). The schedulerHandle.stop() concern is directly provable from the daemon.ts diff. The typecheck gap is stated explicitly in the PR. The test coverage gap is visible in daemon-refactor.test.ts. The route module interiors (memory-routes.ts, hooks-routes.ts, etc.) are not in the diff, so any bugs introduced there cannot be assessed.


- packages/daemon/src/daemon.ts:1 - `bun run typecheck` is explicitly unchecked in the PR with the note that pre-existing failures outside this patch make it unreliable. This leaves type correctness of the new import surface (13 new route modules + state.ts exports) unverified. Given the scale of the import restructure and the removal of several type imports (`AuthConfig`, `ReconcilerHandle`, `DiagnosticsReport`, `serve`), a diff-scoped typecheck failure cannot be ruled out. At minimum the PR should confirm no new typecheck errors were introduced by this diff.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `2ac5148c`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,85 @@
...
+				"  rateLimits:",
+				"    forget:",
+				"      windowMs: 60000",
+				"      max: 30",
+				"    modify:",
+				"      windowMs: 60000",
+				"      max: 60",
+				"    batchForget:",
+				"      windowMs: 60000",
+				"      max: 5",
```
PR-Reviewer-Ant, line 80: This test only asserts that `reloadAuthState(tmpDir)` does not throw — it makes no assertion about the resulting state (`authConfig.mode`, `authSecret`). The PR description explicitly acknowledges this: 'the test validates correct execution rather than checking mutated export values.' This is a real coverage gap. A regression where `reloadAuthState` silently swallows an internal error and leaves `authSecret = null` in token mode would not be caught. The consequence is that the auth middleware in daemon.ts would return 503 ('server initializing') to all requests after a config reload — an undetected availability regression. Even a comment in the test body saying what observable outcome *would* be a regression (e.g., 'if authSecret is null after this call, the daemon would 503 all requests') would make the intent clearer. Consider whether a workaround exists — for example, calling `reloadAuthState` from a separate Worker and asserting on an IPC message, or exporting a getter function instead of a live binding.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import { isSessionCleanupRunning, stopSessionCleanup } from "./session-tracker";
+
+describe("daemon route extraction refactor", () => {
+	afterEach(() => {
+		stopSessionCleanup();
+	});
+
+	it("does not start session cleanup when daemon is imported for route registration", async () => {
+		expect(isSessionCleanupRunning()).toBe(false);
+		await import("./daemon");
```
PR-Reviewer-Ant, line 18: This test asserts that importing `./daemon` does not start session cleanup. Because Bun (and Node) cache ES modules, this assertion is only meaningful the *first* time `./daemon` is imported in the test process. On any subsequent test run or if another test file imports daemon first, the module-level code has already executed and the import is a no-op. The test would pass vacuously in that case. Consider using `mock.module` or a dedicated isolation mechanism to force a fresh module evaluation if the guard must be robust across test orderings.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2ac5148c`

[Automated Review — pr-reviewer bot] This is a clean structural refactor with four rounds of prior review addressed. Two concerns remain worth discussing before merge: a test coverage gap that cannot verify auth state propagation after reload (an explicitly acknowledged Bun limitation), and an unconfirmed fate of `authRecallLlmLimiter` in the extracted modules.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff for daemon.ts and the new test file is complete and provides enough evidence for the findings below. However, routes/state.ts and routes/memory-routes.ts are not included in the diff, so the fate of `authRecallLlmLimiter` cannot be confirmed from the provided context. The Bun live binding concern is directly evidenced by the PR description's own caveat and the test structure at daemon-refactor.test.ts:60-79.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +88,74 @@ import {
...
+	LOG_DIR,
+	MEMORY_DB,
+	NETWORK_MODE,
+	PID_FILE,
+	PORT,
+	analyticsCollector,
+	authAdminLimiter,
+	authBatchForgetLimiter,
+	authConfig,
+	authCrossAgentMessageLimiter,
```
PR-Reviewer-Ant, line 114: `authConfig` and `authSecret` are consumed here as ES module live bindings from `./routes/state.js`. The file watcher at the bottom of this file calls `reloadAuthState(AGENTS_DIR)` to update these on disk changes. The PR description explicitly states that Bun's ES module live bindings 'don't reliably propagate `export let` reassignments from function calls.' If this limitation applies at runtime (not just in the test runner), `reloadAuthState()` calls triggered by the file watcher would leave the middleware reading stale `authConfig`/`authSecret` values — silently making auth config reloads ineffective. The PR author believes this is a test-isolation artifact rather than a runtime defect, but this assumption should be explicitly validated (e.g., by a live integration test that writes a new agent.yaml, triggers a reload, and confirms a token-mode request is rejected with 401 rather than passing through). This is the primary behavioral risk of the entire refactor.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +88,74 @@ import {
 	stopOpenCodeServer,
```
PR-Reviewer-Ant, line 89: The old daemon.ts declared `let authRecallLlmLimiter = new AuthRateLimiter(60_000, 60)` for LLM-backed recall endpoints. In the new daemon.ts, the state.ts imports include `authForgetLimiter`, `authModifyLimiter`, `authBatchForgetLimiter`, `authAdminLimiter`, and `authCrossAgentMessageLimiter` — but not `authRecallLlmLimiter`. The test fixture at line 42 includes a `recallLlm` rate limit in the YAML config, confirming this limiter is expected to exist. If it was not exported from state.ts and plumbed into memory-routes.ts or hooks-routes.ts, the rate limit protection for LLM recall paths is silently absent. routes/state.ts and routes/memory-routes.ts are not in this diff, so this cannot be confirmed. Please verify the limiter is present and wired.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `c16cece0`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
...
-	return url.endsWith("/") ? url.slice(0, -1) : url;
+export function getPredictorClient(): PredictorClient | null {
+	return predictorClientRef;
 }
 
-function normalizeLoopbackHost(host: string): string {
-	return host === "localhost" || host === "::1" ? "127.0.0.1" : host;
+export function recordPredictorLatency(operation: "predictor_score" | "predictor_train", durationMs: number): void {
+	analyticsCollector.recordLatency(operation, durationMs);
 }
```
PR-Reviewer-Ant, line 205: `skillReconcilerHandle.stop()` is called without `await` in `stopPipelineRuntime`. Every other handle in that function is awaited: `predictorClientRef.stop()`, `embeddingTrackerHandle.stop()`, and (after fix #12) `schedulerHandle.stop()`. If `stop()` on the reconciler returns a `Promise<void>`, cleanup is fire-and-forgot during shutdown and any in-flight reconciler work is silently abandoned. If it is intentionally synchronous, a brief comment explaining why would prevent confusion.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 183: `pipelineTransition` (a module-level boolean in the old daemon.ts that guarded concurrent pipeline state changes) is absent from the new daemon.ts and is not present in the state.ts import list. If it was moved into pipeline-routes.ts as a local variable, or into state.ts and only imported by pipeline-routes.ts, that is fine — but neither the diff nor the state.ts import block in daemon.ts confirms this. Please verify the variable was intentionally re-homed and not silently dropped.

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,93 @@
...
+describe("daemon route extraction refactor", () => {
+	afterEach(() => {
+		stopSessionCleanup();
+	});
+
+	// NOTE: This test is only meaningful on the first run in a process —
+	// Bun/Node cache ES modules, so subsequent imports are no-ops. It
+	// guards against module-level side effects at import time (e.g.
+	// calling startSessionCleanup() during top-level evaluation).
+	it("does not start session cleanup when daemon is imported for route registration", async () => {
```
PR-Reviewer-Ant, line 16: The `afterEach` teardown calls `stopSessionCleanup()`, but there is no `beforeEach` guard to ensure `isSessionCleanupRunning()` is false before the import test runs. The comment acknowledges the test is only meaningful on the first run, but if the test suite is ever re-ordered or another test starts cleanup before this one, the assertion `expect(isSessionCleanupRunning()).toBe(false)` before the `import('./daemon')` call will spuriously fail. A `beforeEach(() => stopSessionCleanup())` would make the precondition explicit.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c16cece0`

[Automated Review — pr-reviewer bot] PR #460 is a large structural refactor with a thorough revision history. Three issues survive the prior-review cycle and are worth flagging before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The three findings below are grounded in visible diff lines. However, the extracted route modules (pipeline-routes.ts, state.ts full content, hooks-routes.ts, memory-routes.ts) are not included in the diff, so any cross-module correctness issues — such as whether pipelineTransition was re-homed or silently dropped, or whether authRecallLlmLimiter is wired in memory-routes.ts — cannot be fully verified from this diff alone.

BusyBee3333:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(daemon): extract routes from daemon.ts into separate modules` (commit `b470b2eb`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +88,74 @@ import {
 	stopOpenCodeServer,
 } from "./pipeline/provider";
 import { resolveRuntimeModel } from "./pipeline/provider-resolution";
-import { type RerankCandidate, noopReranker, rerank } from "./pipeline/reranker";
-import { createEmbeddingReranker } from "./pipeline/reranker-embedding";
-import { type PredictorClient, createPredictorClient, resolvePredictorCheckpointPath } from "./predictor-client";
+import { startReconciler } from "./pipeline/skill-reconciler";
+import { type PredictorClient, createPredictorClient } from "./predictor-client";
```
PR-Reviewer-Ant, line 93: `state.ts` exports `export let restartPipelineRuntimeRef` and `setRestartPipelineRuntime()`. Neither is present in daemon.ts's import block from `./routes/state.js`, and neither is visible in the `main()` body shown in the diff. If any extracted route module (e.g. `misc-routes.ts` pipeline-toggle endpoints, or `pipeline-routes.ts`) reads `restartPipelineRuntimeRef` from state.ts to trigger a runtime restart, and daemon.ts never calls `setRestartPipelineRuntime(restartPipelineRuntime)` to wire it, those callers will silently receive `null` and the restart will be a no-op. Please confirm the wiring call is present in the truncated portion of `main()`, or add it explicitly.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +88,74 @@ import {
...
+	isManagedOpenCodeLocalEndpoint,
+	normalizeRuntimeBaseUrl,
+	providerRuntimeResolution,
+	providerTracker,
+	queueExtractionJob,
+	readEnvTrimmed,
+	redactUrlForLogs,
+	reloadAuthState,
+	repairLimiter,
+	setCheckpointPruneTimer,
```
PR-Reviewer-Ant, line 130: The PR checklist leaves `bun run typecheck` unchecked, attributed to pre-existing workspace failures. For a refactor that moves ~10,000 lines across 13 new modules and rewrites dozens of import paths, silent type regressions in the extracted modules are a real risk masked by those pre-existing failures. It is worth running `bunx tsc --noEmit --project packages/daemon/tsconfig.json` (scoped to daemon only) to isolate whether new errors were introduced here, even if the workspace-level check remains noisy.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +88,74 @@ import {
...
+	setHeartbeatTimer,
+	setPredictorClientRef,
+	setShuttingDown,
+	setTelemetryRef,
+	shuttingDown,
+} from "./routes/state.js";
+import { isHarnessAvailable } from "./scheduler";
+import { startSchedulerWorker } from "./scheduler/index.js";
+import { getSecret, hasSecret } from "./secrets.js";
+import { flushPendingCheckpoints, initCheckpointFlush, pruneCheckpoints } from "./session-checkpoints";
```
PR-Reviewer-Ant, line 140: `isHarnessAvailable` is imported from `"./scheduler"` while `startSchedulerWorker` is imported from `"./scheduler/index.js"`. Both resolve to the same module, but using two different specifiers for the same path in the same file is unconventional and could confuse bundler/resolver tooling. Consolidate to one form.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b470b2eb`

[Automated Bot Review] This PR extracts 13 route modules from daemon.ts (80% reduction). The architecture and state-centralization pattern are sound. Two items warrant discussion before merge: a possible unwired pipeline-restart callback, and unverified typecheck coverage.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - state.ts exports `restartPipelineRuntimeRef` (export let) and `setRestartPipelineRuntime()` setter, but neither appears in daemon.ts's import block from state.ts (the full import list is visible in the diff). Whether daemon.ts calls `setRestartPipelineRuntime(restartPipelineRuntime)` somewhere in its truncated `main()` body cannot be confirmed from the provided diff alone. If it is missing, any route module that calls `restartPipelineRuntimeRef(...)` would silently no-op. pipeline-routes.ts (534 lines) and misc-routes.ts (806 lines) are not included in the diff, so whether they depend on this ref cannot be verified here.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9252bf93`

[Automated Review — pr-reviewer bot] Large and well-structured refactor with a thorough reviewer response log. One blocking correctness issue found in the restart-pipeline callback wiring; one warning on a test that passes trivially.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The blocking issue (setRestartPipelineRuntime called with startPipelineRuntime instead of restartPipelineRuntime) is directly readable in the daemon.ts diff. The impact depends on how pipeline-routes.ts invokes the stored callback, which is not in the provided diff. However, stopPipelineRuntime is not exported from daemon.ts (confirmed from the exports visible in the diff), making it inaccessible to route modules, so any call to the stored ref is the only cleanup path — strengthening the case that restartPipelineRuntime is the correct argument.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
...
-	);
-
-	const total = getDbAccessor().withReadDb((db) => {
-		const row = db.prepare("SELECT COUNT(*) as count FROM task_runs WHERE task_id = ?").get(taskId) as {
-			count: number;
-		};
-		return row.count;
-	});
-
-	return c.json({ runs, tota
```
PR-Reviewer-Ant, line 1177: **Blocking: wrong function registered for pipeline restart.**

`setRestartPipelineRuntime(startPipelineRuntime)` registers only the *start* half of a restart. The correct argument is `restartPipelineRuntime`, which calls `stopPipelineRuntime()` first.

Why this matters: `stopPipelineRuntime` is not exported from `daemon.ts` (only `getPredictorClient`, `recordPredictorLatency`, `app`, and the two filename regexps are exported). Route modules therefore have no independent path to stop the running providers, workers, embedding tracker, predictor client, dreaming worker, scheduler, or reconciler before re-starting. Calling `startPipelineRuntime` directly without a prior stop will:
- start duplicate extraction/pipeline workers
- double-init LLM providers (`initLlmProvider`, `initSynthesisProvider`, `initWidgetProvider`)
- leak the predictor sidecar process and the embedding tracker handle
- leak the skill reconciler and scheduler handles

Item 22 in the reviewer response log fixed the *missing registration* (it was a no-op before), but the wrong callee was supplied. Change to:
```ts
setRestartPipelineRuntime(restartPipelineRuntime);
```

packages/daemon/src/daemon-refactor.test.ts
```diff
@@ -0,0 +1,93 @@
...
+		const state = await import("./routes/state.js");
+
+		state.reloadAuthState(state.AGENTS_DIR);
+		const mode1 = state.authConfig.mode;
+		const secret1 = state.authSecret;
+
+		state.reloadAuthState(state.AGENTS_DIR);
+		const mode2 = state.authConfig.mode;
+		const secret2 = state.authSecret;
+
```
PR-Reviewer-Ant, line 32: **Warning: idempotency test reads `state.authSecret` but the acknowledged Bun live-binding limitation means the value is always `null` in the test runner, so both `secret1` and `secret2` will be `null` and the assertion at line 40 passes trivially (`undefined === undefined`).**

In local mode (the default test environment), `authSecret` is never set to a non-null value, so the `toString('hex')` comparison is a no-op. The test only provides real coverage in the token-mode case, which the separate test (line 53) exercises as a no-throw check. The acknowledged limitation is documented in comments, so this is informational rather than a surprise — but the 'idempotency' label is misleading given the vacuous assertion path. Consider adding an explicit `expect(secret1).toBeNull()` assertion to make the local-mode invariant explicit rather than implicit.

packages/daemon/src/daemon.ts
```diff
@@ -271,11306 +171,1174 @@ import {
...
-		const path = join(agentsDir, name);
-		if (!existsSync(path)) continue;
-		try {
-			return resolveNetworkBinding(readNetworkMode(parseSimpleYaml(readFileSync(path, "utf-8"))));
-		} catch {
-			// Ignore malformed config and keep scanning fallbacks.
-		}
+app.use("*", async (c, next) => {
+	const start = Date.now();
+	await next();
```
PR-Reviewer-Ant, line 244: **Module-scope side effect during test imports.** `setupStaticServing()` is called unconditionally at module evaluation time. In the test environment (no dashboard build), this path reaches `logger.warn("daemon", "Dashboard not found - API-only mode", ...)`, producing log noise on every `await import("./daemon")`. The session-cleanup test explicitly guards against import-time side effects but only checks `isSessionCleanupRunning()`, not logging. Not a correctness bug, but the call could be deferred into `main()` to eliminate all module-scope effects and make the guard stronger.

packages/daemon/src/routes/connectors-routes.ts
```diff
@@ -0,0 +1,442 @@
...
+	type ConnectorConfig,
+	type SyncCursor,
+	findSignetForgeBinary,
+	resolveSignetForgeManagedPath,
+} from "@signet/core";
+import type { Hono } from "hono";
+import { createFilesystemConnector } from "../connectors/filesystem.js";
+import {
+	getConnector,
+	listConnectors,
```
PR-Reviewer-Ant, line 16: **`SCRIPTS_DIR` import presence unverifiable.** `connectors-routes.ts` imports `SCRIPTS_DIR` from `./state.js`, but `SCRIPTS_DIR` is not referenced in any visible portion of the file (the harnesses section is truncated). If it is only used in the truncated harnesses handlers and is properly exported from state.ts this is fine — the author reports zero new typecheck errors, which strongly implies it compiles. Worth a quick grep to confirm it is actually consumed rather than being a dead import left from extraction.

packages/daemon/src/daemon.ts
```diff
@@ -183,79 +88,74 @@ import {
...
 import { detectDrift } from "./predictor-comparison";
-import {
-	getComparisonsByEntity,
-	getComparisonsByProject,
-	listComparisons,
-	listTrainingRuns,
-} from "./predictor-comparisons";
 import { getPredictorState } from "./predictor-state";
+import { type RepairContext, structuralBackfill } from "./repair-actions";
 import {
```
PR-Reviewer-Ant, line 97: **Unauditable scoping correctness in extracted route modules.** AGENTS.md rule #1 requires `agent_id` to be threaded through every read/write touching user data. The 11 route modules not present in the provided diff (memory-routes.ts, hooks-routes.ts, misc-routes.ts, repair-routes.ts, knowledge-routes.ts, pipeline-routes.ts, telemetry-routes.ts, session-routes.ts, secrets-routes.ts, and partially connectors-routes.ts) contain the bulk of the data-access handlers. Their scoping correctness cannot be verified from this diff. The PR checklist marks this as confirmed, but a human reviewer should spot-check at least the remember/recall/forget handlers in memory-routes.ts for agent_id and visibility threading before merge.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e915bc7e`

[Automated Review — pr-reviewer bot] PR #460 is a well-iterated structural refactor with 8 addressed review rounds. Two observations are flagged below; no new blocking defects found in the visible diff. However, confidence is reduced because the bulk of the extracted code (~9,000+ lines across 11 of 13 route modules) is not present in the provided diff context.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - daemon.ts and connectors-routes.ts are fully visible and the key structural wiring (setRestartPipelineRuntime, reloadAuthState, scheduler/reconciler stop, live-binding pattern) is verifiable from the diff. However, memory-routes.ts, hooks-routes.ts, misc-routes.ts, repair-routes.ts, and 7 other modules are not in the provided diff. AGENTS.md convention #1 ('thread agent_id through every read/write touching user data') is a per-handler concern that cannot be audited without those files. The PR claims handlers are preserved verbatim, but that claim is unverifiable here.

packages/daemon/src/daemon.ts
aaf2tbz: test reply

aaf2tbz: Pre-existing in `main` (line 9893 of the original monolith). `setupStaticServing()` was always called at module scope. Deferring it into `main()` would be a separate refactor since it serves the dashboard static files and is needed before the HTTP server starts. No change introduced by this PR.

packages/daemon/src/routes/connectors-routes.ts
aaf2tbz: Verified: `SCRIPTS_DIR` is used at `connectors-routes.ts:391` for `join(SCRIPTS_DIR, "generate-harness-configs.py")` in the harness configuration handler. Not a dead import.

packages/daemon/src/daemon.ts
aaf2tbz: Acknowledged — the extracted handlers are verbatim copies from the original `daemon.ts` monolith, so scoping behavior is identical. No `agent_id` or `visibility` threading was changed or added. Human spot-check of memory-routes.ts remember/recall/forget handlers recommended at merge time to confirm the original scoping is correct (that pre-dates this PR).

bun.lock
```diff
@@ -13,7 +13,7 @@
     },
     "packages/adapters/openclaw": {
       "name": "@signetai/signet-memory-openclaw",
-      "version": "0.93.7",
+      "version": "0.93.3",
```
PR-Reviewer-Ant: All 16+ workspace packages are downgraded from `0.93.7` → `0.93.3`. This appears to be a rebase artifact: the branch was cut from `0.93.3` and the four subsequent version bumps (0.93.4–0.93.7) were never merged into the lockfile. Shipping this would roll back all published package versions and silently un-ship the bug fixes described in the deleted CHANGELOG entries. This must be resolved (rebase or cherry-pick the version bumps) before merge.

CHANGELOG.md
```diff
@@ -2,27 +2,6 @@
```
PR-Reviewer-Ant, line 2: Three shipped release entries are deleted: `0.93.5` (sub-agent memory dedupe, #455), `0.93.4` (summary-worker retry fix, #459), and the tail of `0.93.3`. Combined with the lockfile rollback this erases the release history for those fixes. If the intent is to rebase, the CHANGELOG entries should be preserved, not dropped.

packages/cli/src/commands/agent.ts
```diff
@@ -46,10 +39,8 @@ function addToRoster(dir: string, name: string, policy: string, group: string |
 	const filtered = roster.filter(
 		(e: unknown) => typeof e !== "object" || e === null || (e as Record<string, unknown>).name !== name,
 	);
-	const entry: Record<string, unknown> = {
-		name,
-		memory: buildAgentMemoryConfig(policy === "shared" ? "shared" : policy === "group" ? "group" : "isolated", group),
-	};
+	const entry: Record<string, unknown> = { name, read_policy: policy };
```
PR-Reviewer-Ant: The roster entry written by `addToRoster` changed from a `buildAgentMemoryConfig()`-produced structure (which includes a `memory:` key with policy/group encoding) to a flat `{ name, read_policy, policy_group }` object. This is an undocumented behavioral change to the data persisted in `agent.yaml`. If the daemon or any reader still expects the `memory:` key layout, agents added after this change will silently have misconfigured memory policy. This is not scoped to daemon route extraction and is not mentioned in the PR description or reviewer log.

packages/cli/src/commands/agent.ts
```diff
@@ -104,18 +95,10 @@ export function registerAgentCommands(program: Command, deps: AgentDeps): void {
...
-					return [
-						{
-							id: normalized.name,
-							name: normalized.name,
-							read_policy: normalized.readPolicy,
-							policy_group: normalized.policyGroup ?? undefined,
-						},
-					];
-				});
+				rows = roster.filter(
```
PR-Reviewer-Ant: `normalizeAgentRosterEntry` is removed from the listing path. The new filter only checks for `typeof .name === 'string'` and passes raw YAML objects through as `Row`. Fields like `read_policy` and `policy_group` will surface as-is from YAML, but `readPolicy` / `policyGroup` (the normalized camelCase fields the old code produced) are gone. Any downstream code or display layer that depended on the normalized shape will silently see different fields. Verify that all consumers of the `rows` array are compatible with the new raw-object shape.

packages/cli/src/features/health.test.ts
```diff
@@ -3,22 +3,13 @@ import { spawnSync } from "node:child_process";
...
-	} else {
-		process.env.HOME = originalHome;
-	}
-
 	if (originalOpenClawConfig === undefined) {
-		// biome-ignore lint/performance/noDelete: assigning undefined to process.env stores the string "undefined"
-		delete process.env.OPENCLAW_CONFIG_PATH;
+		process.env.OPENCLAW_CONFIG_PATH = undefined;
 	} else {
 		process.env.OPENCLAW_CONFIG_PATH = originalOpenClawConfig;
```
PR-Reviewer-Ant, line 23: `process.env.OPENCLAW_CONFIG_PATH = undefined` stores the literal string `"undefined"` in Node.js (the original biome-ignore comment explicitly calls this out). In `detectOpenClawDualSystem()` the candidates array is filtered with `typeof p === 'string' && p.length > 0`, so `"undefined"` passes the filter and is treated as a real config path. Subsequent tests that rely on the key being absent may attempt to open a file named `undefined`. The correct cleanup is `delete process.env.OPENCLAW_CONFIG_PATH` (as the original code had).

packages/cli/src/features/health.ts
```diff
@@ -67,12 +68,49 @@ interface StatusReport {
...
 	readonly openclawDualSystem: boolean;
-	readonly openclawRuntime: OpenClawRuntimeState;
 	readonly openclawWorkspaceLinked: boolean;
 	readonly openclawWorkspaceUnprotected: boolean;
 }
 
+/**
+ * Check candidate openclaw config paths for the dual-system misconfiguration
+ * where both the legacy hook and plugin are enabled simultaneously.
+ */
```
PR-Reviewer-Ant: `detectOpenClawDualSystem()` reimplements config-path discovery (`~/.openclaw/openclaw.json`, `~/.clawdbot/clawdbot.json`) inline rather than delegating to `OpenClawConnector`. If `OpenClawConnector` resolves additional paths (e.g. XDG config dirs, platform overrides), those will now be missed for the dual-system check. The old code called `new OpenClawConnector().getRuntimeState()` which presumably has authoritative path resolution. Verify the candidate list here is exhaustive, or re-delegate to the connector.

docs/specs/complete/memory-md-rolling-window-lineage.md
```diff
@@ -15,7 +15,7 @@ success_criteria:
   - "Canonical historical content consists of immutable markdown summary/transcript/compaction artifacts, and MEMORY.md is a rebuildable derived projection"
   - "Session-end and compaction persist canonical artifacts with deterministic workspace-root-relative Obsidian wikilinks"
   - "Content-linked DB lineage/search rows are rebuildable from markdown artifacts while runtime temporal telemetry remains DB-native"
-  - "MEMORY.md renders a strict rolling 30-day UTC ledger with one high-signal sentence per projectable in-window session, while fixed-budget projection clips oldest ledger rows explicitly before high-signal head sections"
+  - "MEMORY.md renders a strict rolling 30-day UTC ledger with one high-signal sentence per in-window session and no clipping"
```
PR-Reviewer-Ant: This spec is `status: complete`. The PR mutates its success criteria: the fixed-budget clipping requirement ("clips oldest ledger rows explicitly before high-signal head sections") is replaced with "no clipping/no top-N omission". Changing the acceptance criteria of a shipped spec retroactively is a spec-pipeline violation (specs should move between tiers, not have their contracts rewritten in place). If the implementation has genuinely changed behavior, this spec should be re-examined via the approved → complete promotion path, and the regression test at line 363 that previously validated clipping behavior is now inverted.

docs/specs/dependencies.yaml
```diff
@@ -1,5 +1,5 @@
 version: 3
```
PR-Reviewer-Ant, line 1: `updated_at` is set to `2026-04-01`, which is earlier than today (2026-04-04) and earlier than the `2026-04-02` value on main. This looks like a rebase artifact — the date should reflect when the file was last meaningfully updated.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `420c6ccd`

[Automated Review — pr-reviewer bot] This is a large route-extraction refactor with two blocking defects unrelated to its stated scope: a full version rollback across the monorepo (apparent rebase artifact) and deletion of three shipped CHANGELOG entries. Additional behavioral changes to the CLI agent roster format, a test-cleanup bug, and mutation of a completed spec warrant review.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The version rollback and CHANGELOG deletions are plainly visible in the diff (bun.lock, package.json, CHANGELOG.md). The agent.ts roster-format change is explicit — `buildAgentMemoryConfig` and `normalizeAgentRosterEntry` are removed with no mention in the PR description. The test-cleanup bug is directly readable. The only lower-confidence item is the Bun-specific behavior of `process.env.x = undefined`.

packages/daemon/src/daemon.ts
```diff
@@ -273,10432 +183,539 @@ import {
...
-});
-
-app.post("/api/repair/deduplicate", async (c) => {
-	const cfg = loadMemoryConfig(AGENTS_DIR);
-	const ctx = resolveRepairContext(c);
-	const options: {
-		batchSize?: number;
-		dryRun?: boolean;
-		semanticThreshold?: number;
-		semanticEnabled?: boolean;
```
PR-Reviewer-Ant, line 583: **Blocking — ReferenceError at startup**: `checkEmbeddingProvider` is passed as a callback to `startEmbeddingTracker` here, but the function is not defined anywhere in the new `daemon.ts` and does not appear in the file's import list. The function was explicitly removed from daemon.ts in this PR (visible in the diff's `-` lines). If it was moved to a route module (e.g., `routes/utils.ts`), the corresponding import is missing from daemon.ts. At runtime, when `memoryCfg.pipelineV2.embeddingTracker.enabled` is true, the daemon will throw `ReferenceError: checkEmbeddingProvider is not defined` during `startPipelineRuntime`. The 84/84 test suite does not exercise this code path because no test calls `main()`. Fix: export the function from its new home (utils.ts or memory-routes.ts) and add the import to daemon.ts.

packages/daemon/src/daemon.ts
```diff
@@ -273,10432 +183,539 @@ import {
...
-	const report = getDbAccessor().withReadDb((db) =>
-		getDiagnostics(db, providerTracker, getUpdateState(), buildPredictorHealthParams(), buildOpenClawHealth()),
-	);
-	diagnosticsCache = {
-		report,
-		expiresAt: now + DIAGNOSTICS_CACHE_TTL_MS,
-	};
-	return report;
-}
+	const token = createToken(authSecret, { sub: `token:${role}`, scope, role: role as TokenRole }, ttl);
```
PR-Reviewer-Ant, line 382: **Warning — auth mode snapshot at module-init time**: `mountSkillsRoutes(app, authConfig.mode)`, `mountMarketplaceRoutes(app, authConfig.mode)`, `mountMcpAnalyticsRoutes(app, authConfig.mode)`, and `mountSkillAnalyticsRoutes(app, authConfig.mode)` are called at module-evaluation time, before `main()` → `startPipelineRuntime()` → `reloadAuthState()` runs. If those mount functions capture `mode` as a string (not as a live reference to `authConfig`), they will use the value produced by `parseAuthConfig` at state.ts import time. In the old monolith, all route registration happened inside `main()` after auth was fully initialized. This is a behavioral regression: if `reloadAuthState` changes `authConfig.mode` between module load and first request, these four route modules observe a potentially stale mode. Verify that the functions use `authConfig` live at request time rather than the captured `mode` string, or move these calls into `main()` after `startPipelineRuntime`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a52c05ff`

[Automated Bot Review — pr-reviewer] This is a large structural refactor (12,693 → ~2,500 lines in daemon.ts). The extraction itself is well-organized and the reviewer-response log is thorough. One blocking runtime bug is visible in the changed file content: `checkEmbeddingProvider` is called in `startPipelineRuntime` but is not defined or imported in the new `daemon.ts`. Two warnings on module-init-time auth mode snapshots and dual timer tracking are noted below.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The new daemon.ts content is fully visible and confirms that `checkEmbeddingProvider` is called inside `startPipelineRuntime` but is absent from the import list. The function's removal from daemon.ts is explicit in the diff's `-` lines. The truncated diff prevents verifying whether the function was re-exported from any of the 13 new route modules (e.g., `routes/utils.ts` or `routes/memory-routes.ts`) and whether a corresponding import was added elsewhere. If the function was moved and properly imported in a file not shown, the blocking issue would resolve; absent that evidence, the current daemon.ts as shown would throw a `ReferenceError` at startup whenever `embeddingTracker.enabled` is true.


- packages/daemon/src/daemon.ts:182 - **Warning — dual-source tracking for `heartbeatTimer`, `checkpointPruneTimer`, `telemetryRef`**: These handles are declared as local `let` variables in daemon.ts AND mirrored into state.ts via `setHeartbeatTimer` / `setCheckpointPruneTimer` / `setTelemetryRef`. The cleanup function correctly clears both, but any route module that reads these through state.ts and attempts to cancel them (e.g., on a pipeline restart) would not update the daemon.ts locals. This creates a fragile dual-source-of-truth: a future change that clears only the state.ts copy but not the local would leave a dangling timer. Consider removing the daemon.ts locals and reading exclusively through state.ts accessors, or documenting that only daemon.ts manages lifecycle for these handles.

aaf2tbz:
Superseded by #473 — clean rebase onto latest main (v0.97.0) including rate-limiting changes from #469.

PR #462, opened 2026-04-03
feat(router): unify LLM provider plumbing

### LINKED ISSUE #370: Feature request: Financial protection alerts for extraction provider changes
## Summary

Signet's background extraction pipeline can silently switch to expensive API providers (anthropic, openrouter, claude-code), draining user bank accounts. Without immediate notification when provider changes, users may incur hundreds to thousands of dollars in unexpected API charges before detecting the misconfiguration.

## Current Behavior

- Extraction provider is configured via `agent.yaml` under `memory.pipelineV2.extraction.provider`
- Provider changes take effect immediately with no user notification
- Dashboard shows provider info but provides no cost estimates or warnings
- Users must manually check `agent.yaml` and provider API dashboards to detect expensive provider activation
- Workaround: Users disable extraction entirely (`provider: none`) to avoid accidental costs, losing valuable functionality

## Desired Behavior

When extraction provider changes from a free tier (`ollama`, `none`) to a paid tier (`anthropic`, `openrouter`, `claude-code`, `codex`, `opencode`):

1. **Immediate push notification** sent to all devices running Signet (desktop + mobile)
2. **Email notification** sent to user + configured billing contacts within 30 seconds
3. **Dashboard warning banner** appears showing estimated hourly/daily/weekly/monthly costs
4. **One-click rollback** to previous provider available in both notification and dashboard
5. **Optional config lock** (`allowRemoteProviders: false`) blocks paid provider selection unless explicitly unlocked

## Use Cases

**Use Case 1: Accidental Provider Switch**
- Developer is testing configs, edits `agent.yaml` locally, changes provider to `anthropic` for testing
- Forgets to change back before syncing to production
- Without alerts: $500+ bill over weekend while Signet runs continuous background extractions
- With this feature: Push notification + email arrives within 30 seconds. Developer clicks rollback. Crisis avoided.

**Use Case 2: Shared Team Instance**
- Team runs Signet on shared VPS. Junior dev accidentally commits `provider: openrouter` in config
- Multiple team members' agents start using expensive provider simultaneously
- Without alerts: $2K+ bill before anyone notices
- With this feature: Email to billing contact triggers investigation. Rollback deployed. Cost capped at <$50.

**Use Case 3: Compliance + Financial Controls**
- Enterprise deployment requires visibility into all cost-incurring changes
- Finance/security team needs to approve expensive provider activation
- Without alerts: No audit trail, no financial controls
- With this feature: Email audit trail + cost estimates enable finance team to control Signet deployments

## Proposed Implementation

**Option A (Recommended): Dual Notification + Dashboard + Config Lock**
- Push notifications (native OS + web dashboard toast)
- Email notifications (user + billing contacts)
- Dashboard cost estimator + rollback button
- Config lock with `allowRemoteProviders: false` flag

**Option B (MVP): Email Only + Dashboard Warning**
- Email notifications only (faster to implement)
- Dashboard warning banner
- Manual config changes required for paid providers

**Option C (Lightweight): Dashboard Warning Only**
- Dashboard banner appears when paid provider active
- User-initiated provider changes (no proactive notification)

Recommend Option A for financial safety.

## Expected Impact

**For Users:**
- Financial protection: prevents accidental $1K+ bills
- Peace of mind: costs are visible and controllable
- Faster debugging: clear audit trail of provider changes

**For Teams/Enterprises:**
- Financial governance: enables cost controls + approval workflows
- Compliance: clear notification + audit trail for cost-incurring changes
- Operational safety: reduces "surprise bills" incidents

**For Signet Project:**
- Removes blocker to production adoption (teams won't use Signet without cost safeguards)
- Builds trust in background extraction (users will enable extraction instead of disabling it)

## Related

- Relates to: #341 "Rate limiting and cost safeguards for claude-code LLM provider"
- Extends: Cost safety is only effective with user awareness (this RFE provides the awareness layer)

### LINKED ISSUE #471: Feature request: Support any openai compatible provider for extraction and synthesis
## Summary

I'd like support for any openai compatible provider for extraction and synthesis, not just embedding

---

## Use Cases

**1. Using other providers that are not currently supported**

**2. Using local models with something besides ollama (e.g. tabby)**
---

## Expected Impact

Will make it possible for me to use glm flash for synthesis

---

NicholaiVogel:
## Summary

Bring PR #462 back onto current `main` and keep the LLM provider routing work compatible with the modular daemon pipeline. This also carries forward the usable provider-safety guardrails from PR #550 so provider changes have validation, audit, and rollback coverage on top of the newer pipeline shape.

Related issues:

Closes #471, which requested OpenAI-compatible providers for extraction and synthesis.
Closes #370, which requested financial protection around provider changes.

## Changes

- Merged current `main` into `nicholai/llm-provider-plumbing` and resolved conflicts against the modular daemon route layout.
- Mounted the inference routing API through the current daemon with diagnostics/admin auth gates, OpenAI-compatible gateway support, streaming support, quota/auth redaction, and concurrency limits intact.
- Restored core routing/barrel exports lost during conflict resolution, including type-only exports, `buildFtsMatchQuery`, and `llama-cpp` routing executor support.
- Kept the consolidated daemon LLM registry while preserving separate widget provider registration so widget generation does not consume the synthesis rate-limit wrapper.
- Cherry-picked provider-safety work from PR #550: remote-provider lock enforcement, command-provider validation, provider transition audit, rollback handling, and serialized audit writes.
- Restored dist-based package resolution for `@signet/core` and `@signet/connector-base`, and removed root workspace source path overrides that broke Docker smoke extension builds.
- Removed the unused `safeError` helper flagged by the GitHub code-quality bot.
- Added an explicit `--rewrite-agent-yaml` confirmation gate for `route pin` / `route unpin` so existing `agent.yaml` comments and formatting are not silently lost.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `@signet/native`, `packages/adapters/openclaw`

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent / N/A — no database migration added

## Testing

- [x] `bun test packages/daemon/src/provider-safety.test.ts packages/daemon/src/routes/misc-routes-provider-safety.test.ts packages/daemon/src/memory-config.test.ts packages/daemon/src/inference-api.test.ts packages/core/src/routing.test.ts` — `145 pass / 0 fail` before CI package-resolution fix
- [x] `bun test packages/daemon/src/inference-api.test.ts` — `20 pass / 0 fail` after CI package-resolution fix
- [x] `bun test packages/cli/src/commands/route.test.ts` — `2 pass / 0 fail`
- [x] `bun run --filter @signet/core typecheck`
- [x] `bun run build:core`
- [x] `bun run build:core && bun run build:oh-my-pi-extension`
- [x] `bunx biome check tsconfig.json packages/core/package.json packages/connector-base/package.json`
- [x] `bunx biome check packages/core/src/routing.ts packages/core/src/index.ts packages/daemon/src/llm.ts packages/daemon/src/daemon.ts packages/daemon/src/pipeline/provider.ts packages/daemon/src/pipeline/provider-resolution.ts packages/daemon/src/inference-api.test.ts`
- [x] `git diff --check`
- [x] `bun run build:core && bun test packages/cli/src/commands/route.test.ts packages/daemon/src/inference-api.test.ts packages/core/src/routing.test.ts` — `26 pass / 0 fail` after fresh worktree install
- [x] `bunx biome check packages/cli/src/commands/route.ts packages/cli/src/commands/route.test.ts packages/core/src/routing.ts packages/daemon/src/inference-api.test.ts`
- [ ] N/A

## AI disclosure

## Notes

- Docker Smoke failed on `bun run build:oh-my-pi-extension` because PR462 had added root `paths` and Bun source export conditions that made the extension declaration build pull `packages/core/src` under the extension `rootDir`. This is now restored to dist-based package resolution.
- `@signet/core` no longer has the PR462 type-only export failures called out in review.
- The final combined focused test command had one transient timeout in `inference-api.test.ts` during a concurrent local run; rerunning that file immediately passed `20 pass / 0 fail`.


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor: unify LLM provider plumbing` (commit `e6ddc0a2`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -11851,180 +11981,51 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 		keyCache.set(name, key);
```
PR-Reviewer-Ant: **[Blocking] Synthesis availability check is unconditionally bypassed.**

`checkProviderAvailability: async () => true` short-circuits the preflight in `resolveRuntimeProviderStartup`. The extraction path calls `provider.available()` (or a real check) to detect unreachable backends; synthesis skips it entirely. If the configured synthesis backend is unreachable (e.g. Anthropic key invalid, Ollama not running), the provider is marked active and `initSynthesisProvider` is called with a non-functional provider. Failures will surface only at inference time with no fallback. The old code ran the same CLI/network preflights for synthesis. This is a silent regression in startup hardening.

packages/daemon/src/daemon.ts
```diff
@@ -12086,203 +12087,20 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 		const registryOpenRouterApiKey = openRouterApiKey ?? (await getKey("OPENROUTER_API_KEY"));
```
PR-Reviewer-Ant: **[Blocking] New crash path on synthesis startup failure.**

```ts
if (!synthesisRuntime.provider) {
    throw new Error(`Failed to initialise synthesis provider: ${String(effectiveSynthesisProvider)}`);
}
```
The old code logged a warning and set `effective = null`; the daemon continued running with synthesis disabled. This `throw` propagates through `startPipelineRuntime` and crashes daemon startup entirely when synthesis is configured but unavailable. This is a breaking change in failure semantics — it converts a degraded-mode outcome into a hard crash.

packages/daemon/src/routes/os-chat.ts
```diff
@@ -261,10 +228,6 @@ export function mountOsChatRoutes(app: Hono): void {
 			const toolCallResults: ToolCallResult[] = [];
```
PR-Reviewer-Ant, line 272: **[Blocking] Behavior change misrepresented as pure refactor; removes independent OpenAI fallback.**

The old `callLlm` used OpenAI GPT-4o with its own `OPENAI_API_KEY` regardless of pipeline configuration. The new path uses `getInteractiveLlmProvider()`, which returns the configured synthesis or extraction provider (e.g. `qwen3:4b` on ollama). This changes the quality and behavior of OS chat for all users who had not explicitly configured a synthesis provider — it falls through to the extraction provider (typically a small local model) or throws. The commit type is `refactor` with the stated goal of no behavior change, but this is a meaningful user-visible regression for installations where `OPENAI_API_KEY` was the OS chat provider.

packages/daemon/src/routes/os-agent.ts
```diff
@@ -112,33 +102,19 @@ interface AgentAction {
 	summary?: string;
```
PR-Reviewer-Ant, line 107: **[Blocking] Same behavioral regression as `os-chat.ts` — OS agent loses GPT-4o path.**

`callAgentLlm` previously called `https://api.openai.com/v1/chat/completions` with `gpt-4o` and `max_tokens: 512`. It now calls `getInteractiveLlmProvider().generate(...)`. The structured JSON-action output format the agent depends on (`{ action, selector, value, summary }`) is GPT-4o-specific prompt-following behavior. Using a local extraction model for GUI automation is likely to produce unparseable output and silent failures in `callAgentLlm`'s JSON parsing.

packages/daemon/src/routes/os-chat.ts
PR-Reviewer-Ant, line 272: **[Blocking] Dynamic imports removed but loop body referencing them may still exist.**

The diff removes `const marketplaceModule = await import('./marketplace.js')` and `const { readInstalledServersPublic } = await import('./marketplace-helpers.js')` but the diff is truncated before the `for (const call of parsed.toolCalls...)` loop body. If `readInstalledServersPublic` or `marketplaceModule` are referenced inside that loop, every OS chat request with tool calls will throw a `ReferenceError` at runtime. Needs explicit verification that the loop body no longer references those variables, or that they have been re-imported statically elsewhere in the file.

packages/daemon/src/llm.ts
```diff
@@ -1,32 +1,77 @@
...
+export function closeSynthesisProvider(): void {
+	closeProvider("synthesis");
+}
+
+export function getWidgetProvider(): LlmProvider {
+	return getProvider("synthesis");
+}
+
+export function getInteractiveLlmProvider(): LlmProvider {
+	return providers.synthesis ?? providers.extraction ?? getProvider("synthesis");
```
PR-Reviewer-Ant: **[Warning] `getInteractiveLlmProvider` fallback error is misleading and inconsistently typed.**

```ts
return providers.synthesis ?? providers.extraction ?? getProvider("synthesis");
```
If both providers are absent, the final call throws `"Synthesis LlmProvider not initialised"` even though extraction was also checked and failed. The caller (e.g. `os-chat.ts`) has no way to distinguish "neither available" from "synthesis specifically missing". Additionally, the function is undocumented — callers can't know whether this is synthesis-preferred, extraction-fallback, or something else. At minimum, the error should say both roles were unavailable.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1552,13 +1560,14 @@ export function startWorker(
 					code: msg.includes("timeout") ? "EXTRACTION_TIMEOUT" : "EXTRACTION_PARSE_FAIL",
 					durationMs: runtime.now() - jobStart,
 				});
+				const effectiveMaxAttempts = Math.max(1, Math.min(job.max_attempts, pipelineCfg.worker.maxRetries));
 				accessor.withWriteTx((db) => {
-					failJob(db, job.id, msg, job.attempts, job.max_attempts);
-					if (job.attempts >= job.max_attempts) {
+					failJob(db, job.id, msg, job.attempts, job.max_attempts, effectiveMaxAttempts);
+					if (job.attempts >= effectiveMaxAttempts) {
```
PR-Reviewer-Ant: **[Warning] `consecutiveFailures` reset on dead jobs changes circuit-breaker behavior.**

Old logic: `consecutiveFailures++` on every failure. New logic: `consecutiveFailures = job.attempts >= effectiveMaxAttempts ? 0 : consecutiveFailures + 1`. When a job is marked `dead` (max attempts exhausted), the counter resets to 0 instead of incrementing. This prevents the consecutive-failure circuit breaker from triggering when a batch of jobs all hit their retry limit — a scenario that previously indicated a systemic LLM backend problem and triggered back-off. The behavioral change is not described in the PR.

packages/daemon/src/pipeline/worker.test.ts
```diff
@@ -605,10 +615,8 @@ describe("Worker processing", () => {
 			undefined,
 			{ getLoadPerCpu: () => 1.9 },
 		);
-
 		await Bun.sleep(120);
 		const job = getJob(db, "mem-overload");
-		expect(job?.status).toBe("pending");
 		expect(worker.stats.overloaded).toBe(true);
 		expect(worker.stats.loadPerCpu).toBe(1.9);
```
PR-Reviewer-Ant, line 616: **[Warning] Removed assertion weakens the overload test's contract.**

`expect(job?.status).toBe('pending')` was deleted. That assertion verified the core invariant of the overload guard: a job that could not be processed due to high CPU load must not advance state. The remaining assertions (`overloaded: true`, `loadPerCpu`, `maxLoadPerCpu`) verify the stats object but not the job itself. Per AGENTS.md testing philosophy, the deleted line was testing what must remain true, not implementation details.

packages/daemon/src/pipeline/decision.ts
```diff
@@ -58,6 +58,16 @@ export interface FactDecisionResult {
 const CANDIDATE_LIMIT = 5;
```
PR-Reviewer-Ant, line 58: **[Warning] `buildFtsMatchQuery` duplicated from `core/src/search.ts`.**

Identical function bodies in `packages/core/src/search.ts` (line 80) and `packages/daemon/src/pipeline/decision.ts` (line 58). AGENTS.md explicitly flags duplication as a smell and requires shared sources of truth. If the escaping or tokenization logic needs to change (e.g. to handle Unicode edge cases differently), it must be updated in two places. The version in `@signet/core` is already exported and reachable from the daemon — import it instead.

packages/daemon/src/daemon.ts
```diff
@@ -481,9 +441,298 @@ const providerRuntimeResolution: ProviderRuntimeResolution = {
...
+	if (cfg.role === "extraction") {
+		return runtime.effectiveProvider;
+	}
+	if (runtime.status === "paused" || !cfg.enabled) {
+		return null;
+	}
+	return runtime.effectiveProvider === "command" ? "none" : runtime.effectiveProvider;
+}
+
+function createRuntimeResolutionState<TProvider extends RuntimeResolutionProvider>(opts: {
```
PR-Reviewer-Ant: **[Warning] `logRuntimeStartup` has a hidden mutation side effect.**

The function name implies logging, but it also writes to `providerRuntimeResolution.extraction` / `.synthesis`. This makes call-site reasoning harder: code that calls `logRuntimeStartup` for a log message is also committing to a state transition. The function should either be renamed to reflect the mutation (e.g. `applyAndLogRuntimeStartup`) or the mutation should be separated from logging. As written, a caller that wants to re-log without re-committing state cannot do so safely.

packages/daemon/src/daemon.ts
```diff
@@ -11676,167 +11924,49 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 	if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
```
PR-Reviewer-Ant: **[Warning] `bun run typecheck` is acknowledged-broken across multiple packages (unchecked in PR checklist).**

The PR notes typecheck fails in `@signet/connector-*`, `@signet/opencode-plugin`, and `@signet/oh-my-pi-extension`. Per AGENTS.md PR checklist, typecheck passing locally is mandatory before review. Type errors in connector packages may hide incompatibilities introduced by the `Agent`/`AgentManifest`/`AgentConfig` change from value-exports to type-only exports (`packages/core/src/index.ts` lines 6-8), which is a breaking change for any consumer that was importing these as values.

packages/core/src/index.ts
```diff
@@ -6,9 +6,6 @@
 export { Signet } from "./signet";
 export { Database, findSqliteVecExtension, loadSqliteVec } from "./database";
 export {
-	Agent,
-	AgentManifest,
-	AgentConfig,
 	MEMORY_TYPES,
```
PR-Reviewer-Ant, line 9: **[Warning] `Agent`, `AgentManifest`, `AgentConfig` moved from value-exports to type-only exports — potential breaking change.**

These were previously under `export { Agent, AgentManifest, AgentConfig, ... }` (value exports); they're now under `export type { Agent, AgentManifest, AgentConfig, ... }`. In TypeScript with `isolatedModules` or `verbatimModuleSyntax` (common in this repo's tsconfig), this is backward-compatible if the originals were interfaces/types. But if any consumer imported these as values (e.g. for `instanceof` checks, runtime use as constructors), that code now fails at compile or runtime. Combined with the acknowledged typecheck failures, this warrants explicit verification.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e6ddc0a2`

[Automated Review — pr-reviewer bot] PR #462 consolidates LLM provider plumbing and removes `synthesis-llm.ts`/`widget-llm.ts`. Found several correctness and behavioral-regression issues: synthesis availability check is intentionally bypassed, OS chat/agent routes silently lose their existing fallback path, `throw` on synthesis startup failure is a new daemon-crash path, and typecheck is acknowledged-broken across multiple packages.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All flagged issues are directly provable from changed lines in this diff. The `checkProviderAvailability: async () => true` bypass is explicit in daemon.ts. The `getInteractiveLlmProvider` fallback is in llm.ts. The `throw new Error` path and the dynamic-import removal are in daemon.ts and os-chat.ts respectively. No external runtime trace is needed to identify these.

packages/daemon/src/routes/os-chat.ts
```diff
@@ -261,10 +243,6 @@ export function mountOsChatRoutes(app: Hono): void {
...
```
PR-Reviewer-Ant, line 272: The dynamic imports `await import('./marketplace.js')` and `await import('./marketplace-helpers.js')` (which carried an explicit comment 'Dynamic import to avoid circular deps') were removed from inside the tool-call handler block. No corresponding static imports were added to the file header in this diff. If any code inside `for (const call of parsed.toolCalls.slice(0, 5))` still references `marketplaceModule` or `readInstalledServersPublic`, this will throw a `ReferenceError` at runtime when tool calls are processed. The comment about circular deps also signals that promoting these to static imports would re-introduce a module-load-time circular dependency. The full post-refactor loop body is not visible in the provided diff, so this must be verified before merge.

packages/daemon/src/daemon.ts
```diff
@@ -12086,203 +12086,20 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
...
-	if (pipelinePaused) {
-		providerRuntimeResolution.synthesis = {
-			configured: providerHints.synthesis,
-			resolved: memoryCfg.pipelineV2.synthesis.enabled ? memoryCfg.pipelineV2.synthesis.provider : null,
-			effective: null,
-		};
-		logger.info("config", "Pipeline paused; synthesis provider startup deferred");
-	} else if (memoryCfg.pipelineV2.synthesis.provider === "none") {
-		logger.info("config", "Synthesis provider set to 'none', synthesis disabled");
-	} else if (memoryCfg.pipelineV2.synthesis.enabled) {
```
PR-Reviewer-Ant, line 1504: New `throw new Error('Failed to initialise synthesis provider: ...')` is a behavioral regression from the original code. In the old path, if synthesis fell back to ollama and ollama also failed, `effectiveSynthesisProvider` was set to `'none'` and the daemon continued with synthesis disabled. With this change, `synthesisRuntime.provider === null` (status `'blocked'`) combined with `synthesis.enabled === true` and `synthesis.provider !== 'none'` now crashes the daemon at startup instead of degrading gracefully. Any user whose synthesis provider and its ollama fallback are both unavailable at startup will hit an uncaught throw in `startPipelineRuntime`, which will prevent the daemon from starting. The PR description claims 'refactor without behavior change' but this is a material behavior change.

packages/daemon/src/routes/os-agent.ts
```diff
@@ -112,33 +102,33 @@ interface AgentAction {
...
-		headers: {
-			"Content-Type": "application/json",
-			Authorization: `Bearer ${apiKey}`,
-		},
-		body: JSON.stringify({
+async function callAgentLlm(messages: Array<{ role: "system" | "user" | "assistant"; content: string }>): Promise<AgentAction> {
+	let raw: string;
+	try {
+		raw = await callLegacyOpenAiChat(messages, {
 			model: "gpt-4o",
```
PR-Reviewer-Ant: The fallback from OpenAI to `getInteractiveLlmProviderOrNull()` is gated on the exact string `error.message !== 'OPENAI_API_KEY not found in env or secrets'`. This string is an internal implementation detail of `callLegacyOpenAiChat` in `llm.ts`. Any future rename or reformatting of that error message silently breaks the fallback — the interactive provider will never be tried, and the original error will propagate to the user. Same pattern appears in `os-chat.ts`. Consider using a typed sentinel error class or a boolean flag instead of string-matching across module boundaries.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9a85226d`

[Automated Review — pr-reviewer bot] This PR consolidates LLM provider singletons into a shared registry and moves provider startup logic into a dedicated `provider-resolution.ts` module. The refactor is directionally sound, but several behavioral differences from the stated 'no behavior change' goal were identified, including a new uncaught-throw path on synthesis startup failure and a questionable removal of dynamic imports in `os-chat.ts`.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The synthesis-startup throw and worker consecutiveFailures changes are directly provable from the diff. The os-chat.ts marketplace import removal is observable in the diff, but the full post-refactor loop body is truncated in the provided context, so the runtime impact cannot be fully confirmed without reading the complete file. The typecheck failure acknowledged in the PR notes adds uncertainty about whether the removed imports cause compile-time or runtime errors in that path.

packages/daemon/src/llm.ts
```diff
@@ -1,32 +1,132 @@
...
+};
+
+const providers: Partial<Record<LlmRole, LlmProvider>> = {};
+let cachedOpenAiApiKey: string | null | undefined;
+
+function initProvider(role: LlmRole, instance: LlmProvider): void {
+	if (providers[role]) {
+		logger.warn(ROLE_LOG_CATEGORIES[role], `${ROLE_LABELS[role]} already initialised, skipping`);
 		return;
 	}
```
PR-Reviewer-Ant: `cachedOpenAiApiKey` is module-level state that is never reset by `closeLlmProvider()`, `closeSynthesisProvider()`, or any other close path. In tests that exercise the daemon lifecycle (stop → reconfigure → start), the stale cached key persists and will be returned even if the env variable or secret was changed between runs. Consider clearing this cache in a `resetForTesting()` export or resetting it when the provider registry is fully closed.

packages/daemon/src/pipeline/decision.ts
```diff
@@ -106,9 +108,6 @@ async function findCandidatesVector(
 	scope: DecisionScope,
```
PR-Reviewer-Ant, line 108: `isVectorRuntimeUsable()` guard was removed from `findCandidatesVector`. Vector search is now always attempted, and failure is swallowed by the surrounding `try/catch`. The outcome is functionally equivalent (empty map), but every decision tick now runs the full vector query path even when no native addon or `sqlite-vec` extension is loaded, incurring unnecessary SQLite prepare/all calls. This is a performance regression in non-vector-capable deployments. The early-exit guard should be reinstated or the availability check moved inside `resolveRuntimeProviderStartup` context.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1552,13 +1560,14 @@ export function startWorker(
 					code: msg.includes("timeout") ? "EXTRACTION_TIMEOUT" : "EXTRACTION_PARSE_FAIL",
 					durationMs: runtime.now() - jobStart,
 				});
+				const effectiveMaxAttempts = Math.max(1, Math.min(job.max_attempts, pipelineCfg.worker.maxRetries));
```
PR-Reviewer-Ant: `consecutiveFailures` is now reset to `0` when a job exhausts `effectiveMaxAttempts` and is marked `dead`. Previously, dead jobs still incremented the counter. The counter drives worker back-off: resetting it on job death means that a flood of jobs persistently crashing (being marked dead) will no longer slow down the worker. If many jobs are dying in quick succession (e.g., due to a broken extraction prompt), the worker will poll at full speed. Verify this is the intended circuit-breaker semantics and add a comment explaining the intentional asymmetry.

packages/daemon/src/daemon.ts
```diff
@@ -11676,167 +11924,48 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 	if (rl.admin) authAdminLimiter = new AuthRateLimiter(rl.admin.windowMs, rl.admin.max);
```
PR-Reviewer-Ant: PR checklist explicitly leaves `bun run typecheck` and `bun run lint` unchecked, and the Notes section acknowledges multiple packages fail typecheck due to `rootDir` boundary violations. Per repo convention (AGENTS.md PR checklist), typecheck and lint are required before review. The PR may land with pre-existing or newly-introduced type errors that are hard to attribute later.

packages/daemon/src/daemon.ts
NicholaiVogel: Fixed in 64a88be6. Synthesis startup no longer throws when the configured provider is unavailable; it now logs a warning and continues with synthesis disabled, preserving the prior degraded-startup behavior.

packages/daemon/src/daemon.ts
NicholaiVogel: Addressed in 64a88be6 as well. The startup path now keeps the old failure semantics by warning and continuing without synthesis instead of crashing daemon boot.

packages/daemon/src/routes/os-agent.ts
NicholaiVogel: Good catch. I replaced the string match with a typed sentinel error in `packages/daemon/src/llm.ts:16-21`, and both OS chat and OS agent now key their OpenAI fallback off that type instead of an error-message string.

packages/daemon/src/llm.ts
NicholaiVogel: Fixed in 64a88be6. The OpenAI API key cache is now cleared when both registered providers are closed, so daemon lifecycle tests and reconfiguration flows do not retain stale cached credentials.

packages/daemon/src/pipeline/decision.ts
NicholaiVogel: Reinstated the early vector-runtime guard in `packages/daemon/src/pipeline/decision.ts:103-146`, so non-vector deployments now short-circuit before attempting the vector query path.

packages/daemon/src/pipeline/worker.ts
NicholaiVogel: Adjusted in 64a88be6. `consecutiveFailures` now increments on dead jobs again, and I added coverage for that circuit-breaker behavior in `packages/daemon/src/pipeline/worker.test.ts:1974-1996`.

packages/daemon/src/pipeline/worker.test.ts
NicholaiVogel: Restored that invariant check. The overload test now asserts the job stays pending again in `packages/daemon/src/pipeline/worker.test.ts:619-625`.

packages/daemon/src/routes/os-chat.ts
NicholaiVogel: Addressed in 9a85226d. OS chat keeps the legacy GPT-4o path when `OPENAI_API_KEY` is available and only falls back to the shared interactive provider when the key is absent (`packages/daemon/src/routes/os-chat.ts:18-35`).

packages/daemon/src/routes/os-agent.ts
NicholaiVogel: Also addressed in 9a85226d. OS agent now preserves the prior GPT-4o call first and only falls back to the shared interactive provider when OpenAI credentials are unavailable (`packages/daemon/src/routes/os-agent.ts:113-130`).

packages/daemon/src/routes/os-chat.ts
NicholaiVogel: Verified and fixed earlier in the branch. The tool-call path no longer references the removed dynamic-import locals, so there is no remaining `ReferenceError` path from that change.

packages/daemon/src/llm.ts
NicholaiVogel: Updated earlier in the branch. `getInteractiveLlmProvider()` now throws a dedicated interactive-provider error path after checking both roles, rather than surfacing a misleading synthesis-only initialization message (`packages/daemon/src/llm.ts:138-142`).

packages/daemon/src/pipeline/decision.ts
NicholaiVogel: Fixed in 9a85226d. `packages/daemon/src/pipeline/decision.ts` now imports the shared FTS query builder from core instead of duplicating the implementation.

packages/daemon/src/daemon.ts
NicholaiVogel: Done earlier in the refactor. The helper is now named `applyAndLogRuntimeStartup(...)` to reflect that it both commits runtime-resolution state and logs startup details (`packages/daemon/src/daemon.ts:610-669`).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `64a88be6`

I'll base the review on the diff and changed file contents provided in the prompt. Let me analyze thoroughly.

```pr-review-json
{
  "summary": "[Automated Review — pr-reviewer bot] This PR consolidates LLM provider plumbing across extraction, synthesis, and widget roles into a shared registry in `llm.ts`, moves provider-resolution logic into a dedicated module, hardens FTS query building, and stabilises Bun workspace source resolution. Several issues were found: one blocking behavioral regression in OpenAI API key caching, one warning-level duplication divergence, and one warning-level behavior change in the worker retry gate that is technically not pure refactor.",
  "verdict": "request_changes",
  "confidence": {
    "level": "high",
    "justification": "The key caching regression is fully provable from the diff: the old per-route `let cachedApiKey: string | null = null` pattern re-fetched on every call when the key was absent (`null` and `\"\"` are both falsy), while the new `cachedOpenAiApiKey !== undefined` guard in `llm.ts:83-87` permanently caches a missing key as `\"\"` after the first failed lookup. The duplication in `daemon.ts` is visible by comparing the inline synthesis-effective logic at line ~12086–12094 against the exported `resolveEffectiveRuntimeProvider` defined in `provider-resolution.ts`. The worker retry gate change is evident from the `failJob` signature change and the `effectiveMaxAttempts` clamping at `worker.ts:1560`."
  },
  "comments": [
    {
      "file": "packages/daemon/src/llm.ts",
      "line": 83,
      "body": "**Blocking – OpenAI API key cache permanently poisons on first missing-key lookup.**\n\nPrevious per-route code (`os-chat.ts`, `os-agent.ts`) used `let cachedApiKey: string | null = null` with `if (!cachedApiKey)` as the guard. Because both `null` and `\"\"` are falsy, every call re-attempted the secret lookup when the key was absent, allowing dynamically-added secrets to be picked up without a restart.\n\nThe new `getOpenAiApiKey()` uses `if (cachedOpenAiApiKey !== undefined)` as its gate. On a missing key, `cachedOpenAiApiKey` is set to `\"\"` (line ~87: `process.env.OPENAI_API_KEY || (await getSecret(...).catch(() => \"\"))`). `\"\" !== undefined` is true, so all subsequent calls immediately throw `MissingOpenAiApiKeyError` without re-querying the environment or secrets store. The cache is only reset via `closeProvider()`, which only fires when both providers are closed — not a normal runtime event.\n\nConsequence: a user who starts the daemon before setting `OPENAI_API_KEY` and then adds it via `signet secrets set OPENAI_API_KEY` will receive errors until the daemon is restarted. This regresses the existing live-configuration behavior.\n\nSuggestion: use `null` (not `\"\"`) as the \"definitely not found\" sentinel and retain a retry/TTL path, or at minimum document the restart requirement.",
      "severity": "blocking"
    },
    {
      "file": "packages/daemon/src/daemon.ts",
      "line": 12086,
      "body": "**Warning – inline synthesis effectiveProvider logic duplicates the exported `resolveEffectiveRuntimeProvider`.**\n\n`resolveEffectiveRuntimeProvider` is defined and exported from `provider-resolution.ts` (handles both extraction and synthesis cases). In `resolveAndLogRuntimeStartup`, instead of calling that function for the synthesis branch, the same ternary chain is re-implemented inline:\n```ts\nconst effectiveProvider: RuntimeSynthesisProviderName | null =\n  runtime.status === 'paused'\n    ? null\n    : cfg.enabled\n      ? runtime.effectiveProvider === 'command' ? 'none' : runtime.effectiveProvider\n      : null;\n```\nIf the logic in `resolveEffectiveRuntimeProvider` is ever updated, this inline copy will silently diverge. AGENTS.md §5 ("Duplication / parity drift") explicitly flags this pattern. The fix is to call `resolveEffectiveRuntimeProvider(cfg, runtime)` here.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/pipeline/worker.ts",
      "line": 1560,
      "body": "**Warning – `effectiveMaxAttempts` clamping is a behavior change disguised as refactor.**\n\nThe PR is typed as `refactor`. However, the new code adds:\n```ts\nconst effectiveMaxAttempts = Math.max(1, Math.min(job.max_attempts, pipelineCfg.worker.maxRetries));\nfailJob(db, job.id, msg, job.attempts, job.max_attempts, effectiveMaxAttempts);\nif (job.attempts >= effectiveMaxAttempts) { ... }\n```\nPreviously, `job.max_attempts` was the sole threshold for transitioning a job to `dead`. Now, if `pipelineCfg.worker.maxRetries < job.max_attempts`, a job will be marked dead earlier than its stored `max_attempts` would indicate. Jobs can now be killed by a config change that reduces `maxRetries` without migrating stored `max_attempts` values. The PR checklist acknowledges that `Input/config validation and bounds checks added` was not completed, which is relevant here. The associated tests were adjusted to match (e.g., `maxRetries: 1 → 2`, extended sleep) rather than asserting the old contract was preserved.",
      "severity": "warning"
    },
    {
      "file": "packages/core/package.json",
      "line": 11,
      "body": "**Nitpick – `\"bun\": \"./src/index.ts\"` exposes raw TypeScript source as the published package entry point under Bun.**\n\nAdding `src` to `files` and a `bun` export condition means downstream consumers running under Bun receive the TypeScript sources directly rather than the compiled `dist/`. This can cause unexpected behavior if those consumers have different `tsconfig` settings or if `src/` transitively imports devDependencies. Intentional for workspace-internal test resolution, but the blanket `\"bun\"` condition in the exported `package.json` affects external consumers too. Consider scoping this to workspace-internal usage (e.g., via a separate internal entry or Bun-specific workspace override) rather than publishing it.",
      "severity": "nitpick"
    }
  ]
}
```

packages/core/src/routing.ts
```diff
@@ -0,0 +1,1111 @@
...
+	if (
+		(request.requireMultimodal ?? config.taskClasses[classification.taskClass]?.multimodalRequired) &&
+		model.multimodal !== true
+	) {
+		blockedBy.push("multimodal required");
+	}
+	const expectedInputTokens =
+		request.expectedInputTokens ?? config.taskClasses[classification.taskClass]?.expectedInputTokens;
+	if (expectedInputTokens && model.contextWindow && expectedInputTokens > model.contextWindow) {
+		blockedBy.push(`context window too small (${model.contextWindow})`);
```
PR-Reviewer-Ant, line 1036: **Score inversion bug — blocking correctness.** `orderedTargets.indexOf(targetRef)` returns `-1` when the candidate is not in the preferred list (e.g. a roster-only candidate). That makes `100 - (-1 * 5) = 105`, which exceeds the score of the top-ranked preferred target (100). In automatic mode, any candidate that is NOT explicitly preferred will outscore every explicitly preferred one — inverting the entire preference ordering.

Fix: guard the expression or substitute a penalizing value when idx < 0:
```ts
const idx = orderedTargets.indexOf(targetRef);
if (idx >= 0) {
  score += 100 - idx * 5;
  reasons.push(`preferred order ${idx + 1}`);
} else {
  score += 0; // no preference bonus
}
```
The `reasons.push` check on the next line already has the correct `includes` guard but the score has already been wrongly bumped before it.

packages/core/src/routing.ts
```diff
@@ -0,0 +1,1111 @@
...
+	const base = legacyConfig ?? emptyRoutingConfig("explicit");
+	if (!isRecord(raw)) {
+		return ok(base);
+	}
+	const embeddedRouting = isRecord(raw.routing) ? raw.routing : null;
+	const standaloneRouting = embeddedRouting ? null : hasStandaloneRoutingShape(raw) ? raw : null;
+	const routingRaw = embeddedRouting ?? standaloneRouting;
+	if (!routingRaw) {
+		return ok(base);
+	}
```
PR-Reviewer-Ant, line 664: **Executor type gap in legacy compiler.** `compileLegacyRoutingConfig` assigns `executor: opts.extraction.provider` and `executor: opts.synthesis.provider` directly into `RoutingTargetConfig.executor` (typed as `RoutingExecutorKind`). `PipelineProviderChoice` includes values like `"openai"` that are absent from `ROUTING_EXECUTOR_KINDS` (`["claude-code","codex","opencode","anthropic","openrouter","ollama","openai-compatible"]`). Any legacy install using `provider: openai` would store an invalid executor string in the compiled routing config, causing `parseTargetConfig` to silently drop that target if it is later round-tripped through the parser (which guards with `ROUTING_EXECUTOR_KINDS.includes`), and silently routing to no extraction target. This is likely contributing to the acknowledged typecheck failures. A narrowing step or explicit provider-to-executor mapping is needed here.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+				agentId: options.agent,
+				taskClass: options.taskClass,
+				operation: options.operation,
+				privacy: options.privacy,
+				explicitPolicy: options.policy,
+				explicitTargets: options.target ? [options.target] : undefined,
+				maxTokens,
+				refresh: options.refresh === true,
+			});
+			if (!ok) {
```
PR-Reviewer-Ant, line 374: `signet route pin <targetRef>` writes the value to `agent.yaml` without validating the `target/model` format. `parseRoutingTargetRef` is already exported from `@signet/core` — call it here and print an error before writing if the ref is malformed. A typo like `opus46` (missing slash) would be silently pinned and then permanently fail routing until manually corrected.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+				console.log(JSON.stringify(data, null, 2));
+				return;
+			}
+			const result = data as {
+				text?: string;
+				decision?: { targetRef?: string };
+				attempts?: Array<{ targetRef: string; ok: boolean; error?: string }>;
+			};
+			console.log(chalk.bold("\n  Route test\n"));
+			console.log(chalk.dim(`  Selected: ${result.decision?.targetRef ?? "-"}`));
```
PR-Reviewer-Ant, line 388: `writeAgentYaml` and the callers (`pin`, `unpin`) have no error handling. If `stringifyYamlDocument` or `writeFileSync` throws, the process exits with an unhandled exception rather than a user-readable message. Wrap in try/catch and print a friendly error before `process.exit(1)`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5d84c8fe`

[Automated Review — pr-reviewer bot] PR #462 introduces a new routing core module and CLI surface. Found one blocking correctness bug in the scoring engine, one warning-level type-safety gap in the legacy compiler, and two process compliance gaps (typecheck/lint failing) that the author acknowledges but leaves open.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The score inversion in buildCandidateTrace is directly visible in the routing.ts diff — Array.prototype.indexOf returns -1 for absent elements, making 100 - (-1 * 5) = 105, which outscores any explicitly preferred target. The executor type assignment in compileLegacyRoutingConfig is also directly visible in the diff and unambiguously passes PipelineProviderChoice values as RoutingExecutorKind without narrowing. The typecheck/lint failures are author-disclosed in the PR notes. No runtime trace is needed to confirm the first two — the arithmetic and the unguarded assignment are provable from the diff alone.

packages/core/src/routing.ts
```diff
@@ -0,0 +1,1111 @@
...
+	if (candidateRefs.length === 0) {
+		return err("no-candidates", "No route candidates were available for this request.", {
+			policyId: pref.policyId,
+			agentId: request.agentId,
+			taskClass: classification.taskClass,
+		});
+	}
+
+	const traces = candidateRefs.map((targetRef) =>
+		buildCandidateTrace(
```
PR-Reviewer-Ant, line 1136: **Redundant double-block for `local_only` privacy.** When `requiredPrivacy === "local_only"` and the target is not local, both guards fire: `privacyRank(targetPrivacy) < privacyRank("local_only")` is true (first push), then `requiredPrivacy === "local_only" && target.kind !== "local"` is also true (second push). The `trace.candidates[].blockedBy` array for that candidate will contain two entries for one logical reason. `explain` and `doctor` consumers will display duplicated blocking messages. The second check is redundant and can be removed.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+			}
+			printStatus(status);
+		});
+	withJson(list);
+
+	const statusCmd = routeCmd
+		.command("status")
+		.description("Show routing health and workload bindings")
+		.action(async (options) => {
+			const status = await deps.fetchFromDaemon<RouteStatusResponse>("/api/inference/status");
```
PR-Reviewer-Ant, line 222: `signet route list` and `signet route status` have identical action implementations — both call `/api/inference/status` and both call `printStatus`. The docs describe them as distinct (`list` = config + health, `status` = health + workload bindings). Either differentiate the output or consolidate to one command. As-is, `status` is a no-op alias with a misleading description.

packages/core/src/routing.ts
```diff
@@ -0,0 +1,1111 @@
...
+
+	const trace: RouteTrace = {
+		policyId: pref.policyId,
+		mode: pref.mode,
+		classification,
+		orderedTargets: pref.orderedTargets,
+		candidates: traces,
+	};
+
+	const allowed = traces.filter((candidate) => candidate.allowed);
```
PR-Reviewer-Ant, line 1161: `classifyRouteRequest(config, request)` is called twice in `resolveRoutingDecision` — once as an argument to `orderedPreferenceLists`, and again on the next line to assign `classification`. The first call's result is available as the third parameter passed to `orderedPreferenceLists` but discarded. Extract the result to a `const` before the `orderedPreferenceLists` call and reuse it.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
+import { existsSync, readFileSync, writeFileSync } from "node:fs";
```
PR-Reviewer-Ant, line 1: The `signet route` subcommand group is a new user-facing CLI surface (list, status, doctor, explain, test, pin, unpin). Per AGENTS.md conventions, `feat:` is reserved for user-facing features and bumps the minor version. This PR is labeled `refactor`, which means no minor-version bump will be triggered. If semver is enforced automatically from commit type, downstream consumers will not see the expected version signal for a new command surface.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+				console.log(chalk.bold("\n  Trace\n"));
+				console.log(JSON.stringify(decision.trace ?? null, null, 2));
+			}
+			console.log();
+		});
+
+	routeCmd
+		.command("test <prompt>")
+		.description("Execute a real routed prompt")
+		.option("--agent <agent>", "Agent id")
```
PR-Reviewer-Ant, line 346: `Number.parseInt(options.maxTokens, 10)` returns `NaN` when `--max-tokens` receives a non-numeric string (e.g. `signet route test ... --max-tokens abc`). The falsy-check on `options.maxTokens` only guards for an absent flag, not for non-numeric input. `NaN` is then sent to the API as `maxTokens`. The API spec says this field is clamped, but the behavior of clamping `NaN` is undefined. Add a guard: `const parsed = Number.parseInt(options.maxTokens, 10); if (!Number.isFinite(parsed) || parsed <= 0) { console.error(...); process.exit(1); }`

packages/core/src/routing.ts
```diff
@@ -0,0 +1,1129 @@
...
+}
+
+function err(
+	code: RouterError["code"],
+	message: string,
+	details?: Readonly<Record<string, unknown>>,
+): RouterResult<never> {
+	return { ok: false, error: { code, message, ...(details ? { details } : {}) } };
+}
+
```
PR-Reviewer-Ant, line 224: `hasStandaloneRoutingShape` matches if the raw object contains any of `"enabled"`, `"targets"`, `"agents"`, `"policies"`, `"workloads"`, etc. A standard `agent.yaml` already uses `"enabled"` as a top-level flag for several features (`memory.pipelineV2.enabled`, update flags, etc.). If `parseRoutingConfig` is ever called with an already-flattened config fragment that carries an `"enabled"` key at root, the entire fragment is mistakenly treated as a standalone routing block, silently overwriting the legacy base config instead of merging it. The heuristic should require at least two signals (`"targets"` + one other) or be scoped to explicit call sites that already know they are receiving a standalone routing block.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+				readonly health: string;
+				readonly accountState: string;
+				readonly unavailableReason?: string;
+			}
+		>;
+	};
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+	return typeof value === "object" && value !== null && !Array.isArray(value);
```
PR-Reviewer-Ant, line 63: `writeAgentYaml` serializes the parsed JS object back to YAML via `stringifyYamlDocument`. Round-tripping through parse+stringify will silently strip all comments and likely alter hand-crafted formatting in `agent.yaml`. For `route pin` / `route unpin` this means a user who has extensively commented their `agent.yaml` loses all those annotations on the first pin operation. Consider a targeted YAML patch strategy (e.g. modify only the `routing.agents.<id>.pinnedTargets` path) rather than a full reserialize.

docs/AUTH.md
```diff
@@ -139,18 +139,21 @@ enforced in `team` and `hybrid` modes — `local` mode has no limits.
...
+|------------------|-------|
+| forget           |    30 |
+| modify           |    60 |
+| batchForget      |     5 |
+| forceDelete      |     3 |
+| admin            |    10 |
+| inferenceExplain |   120 |
+| inferenceExecute |    20 |
+| inferenceGateway |    30 |
+| recallLlm        |    60 |
```
PR-Reviewer-Ant, line 152: The `x-signet-actor` header has been removed from the docs and the rate-limiting description changed so all unauthenticated requests now share the `"anonymous"` bucket. This is a behavioral change (previously per-actor buckets could be attributed in hybrid mode via the header). No daemon source diff is present to confirm whether the implementation was actually changed or only the docs were updated. If the daemon still accepts the header, the docs are wrong. If the daemon was changed, this is a breaking change that was not listed as such and has no migration note. Clarify which case this is.

packages/core/src/routing.ts
```diff
@@ -0,0 +1,1129 @@
+import type { PipelineExtractionConfig, PipelineSynthesisConfig } from "./types";
```
PR-Reviewer-Ant: The diff is truncated and `resolveRoutingDecision` — specifically the candidate-scoring loop, fallback-chain construction, and strict-mode handling — cannot be fully audited. The test in `routing.test.ts` covers the local-only privacy gate and explicit-target-outside-roster rejection, which is good, but the `strict` and `hybrid` policy modes are not covered by any visible test. Given that privacy-tier enforcement is a hard security gate per the spec, the missing coverage for `strict` mode (which uses an explicit ordered chain) is a gap.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+		});
+	withJson(statusCmd);
+
+	const doctorCmd = routeCmd
+		.command("doctor")
+		.description("Diagnose broken route targets and workload bindings")
+		.action(async (options) => {
+			const status = await deps.fetchFromDaemon<RouteStatusResponse>("/api/inference/status");
+			if (!status) {
+				console.error(chalk.red("Failed to get routing status from daemon"));
```
PR-Reviewer-Ant: `route status` and `route list` have identical implementations: both call `GET /api/inference/status` and both call `printStatus`. There is no behavioral difference. Either document the intended distinction or collapse them into one command with an alias.

docs/CONFIGURATION.md
```diff
@@ -667,6 +885,15 @@ auth:
     modify:
```
PR-Reviewer-Ant, line 957: `recallLlm` appears in the rate-limit reference table (line ~920) but is absent from the YAML config example immediately above it. The same gap exists in `docs/AUTH.md`. Operators copying the YAML snippet will not know how to override `recallLlm`. Add it to both YAML examples for consistency.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5e603758`

[Automated Review — pr-reviewer bot] PR #462 introduces a new inference routing layer, new API endpoints, and `signet route` CLI commands. The implementation is substantial and mostly well-structured, but several correctness, convention, and security-documentation issues are flagged below. The PR is labeled `refactor` but ships user-visible CLI surface that requires `feat:` per repo convention, typecheck is admittedly broken across several packages, and two distinct localhost-bypass security descriptions now contradict each other across docs.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The routing.ts decision engine diff is truncated at ~750 lines, so the full candidate-scoring and fallback logic cannot be audited. The daemon source for the auth-middleware localhost bypass is not in the diff at all, so whether the TCP-vs-Host change in SELF-HOSTING.md reflects an actual implementation change or just a doc correction cannot be confirmed from this diff alone. The CLI and core routing changes are fully visible and the findings below are grounded in those concrete changed lines.


- docs/AUTH.md:35 - AUTH.md still contains: 'localhost detection in hybrid mode works by checking the `Host` request header, not the TCP peer address.' This PR updated SELF-HOSTING.md (line ~288) to say the bypass now checks the actual socket peer address, but AUTH.md was not updated. The two docs now give directly contradictory security descriptions. Operators making deployment decisions based on AUTH.md will have incorrect expectations about what is and is not spoofable in hybrid mode. AUTH.md must be updated to match SELF-HOSTING.md (or vice-versa if the daemon was not actually changed).

packages/core/src/routing.ts
PR-Reviewer-Ant: Blocking: `bun run typecheck` is explicitly failing and the PR checklist box is unchecked. AGENTS.md lists 'Lint/typecheck/tests pass locally' as a mandatory pre-review requirement. The author notes the failures are caused by cross-package `rootDir` boundary violations in `@signet/connector-*`, `@signet/opencode-plugin`, and `@signet/oh-my-pi-extension`. These must be resolved or the failures must be explicitly scoped out with a documented exemption before merge, because type errors introduced by this refactor may silently break the routing plumbing in those packages.

docs/SELF-HOSTING.md
```diff
@@ -284,10 +288,10 @@ auth:
 
 This is a practical default for development machines where you want the
 local dashboard and CLI to work without tokens, but also want to call the
-daemon from CI pipelines or remote agents. Note the localhost bypass is
```
PR-Reviewer-Ant, line 287: Warning: The docs now claim hybrid-mode localhost bypass uses the TCP peer address instead of the `Host` header — a security-meaningful change that fixes spoofability. However, the daemon auth middleware source (likely `packages/daemon/src/auth/`) is absent from this diff. If the code was not updated, these docs are misleading and the Host-header spoofing vector still exists. If the code was updated, the implementation should be in this PR and reviewable. Please confirm the middleware change is included or identify the commit that contains it.

docs/AUTH.md
```diff
@@ -139,18 +139,21 @@ enforced in `team` and `hybrid` modes — `local` mode has no limits.
...
 
 The actor is identified by the token's `sub` claim when present. For
-unauthenticated requests (hybrid mode, localhost), the `x-signet-actor`
-header is used as a fallback. If neither is available, the actor defaults
-to `"anonymous"`.
+unauthenticated requests (hybrid mode, localhost), requests share the
+`"anonymous"` bucket.
 
 When a limit is exceeded, the daemon returns `429 Too Many Requests` with
 a `Retry-After` header indicating how many seconds until the window resets.
```
PR-Reviewer-Ant, line 159: Warning: Removing `x-signet-actor` header support is a behavioral breaking change. The old behavior let unauthenticated local tools in hybrid mode claim separate rate-limit buckets; the new behavior collapses all unauthenticated requests into one shared `"anonymous"` bucket. With `inferenceExecute` limited to 20/min, two concurrent CLI tools hitting the router will exhaust each other's budget. This removal is undocumented in the PR description ('no UI changes') and is not a pure refactor. If intentional, document the migration impact.

packages/core/src/index.ts
```diff
@@ -6,9 +6,6 @@
 export { Signet } from "./signet";
```
PR-Reviewer-Ant, line 6: Warning: `Agent`, `AgentManifest`, and `AgentConfig` are moved from value exports (`export { ... }`) to type-only exports (`export type { ... }`). If any of these are classes or are used with `instanceof` checks elsewhere in the monorepo, this is a runtime breaking change that TypeScript may not catch before the typecheck is fixed. Similarly, `MigrationSource` is moved to `export type`. Verify that all three former value exports are interfaces/type aliases and not classes.

packages/connector-openclaw/package.json
```diff
@@ -1,42 +1,34 @@
...
+	"version": "0.93.1",
+	"description": "Signet connector for OpenClaw - configures workspace and memory hooks",
+	"type": "module",
+	"main": "./dist/index.js",
+	"types": "./dist/index.d.ts",
+	"exports": {
+		".": {
+			"bun": "./src/index.ts",
+			"import": "./dist/index.js",
+			"types": "./dist/index.d.ts"
```
PR-Reviewer-Ant: Warning: Adding `"bun": "./src/index.ts"` to the exports map required adding `"src"` to the `files` array, which means TypeScript source files will be published to npm. The same pattern is in `packages/connector-base/package.json` and `packages/core/package.json`. Shipping source alongside `dist` significantly increases package size and could confuse non-bun Node.js consumers who accidentally resolve to TS sources. Confirm this is intentional and that it does not break the existing connector install path under plain Node.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+		return { path, data: {} };
+	}
+	const parsed = parseYamlDocument(readFileSync(path, "utf-8"));
+	return { path, data: isRecord(parsed) ? parsed : {} };
+}
+
+function writeAgentYaml(path: string, data: Record<string, unknown>): void {
+	writeFileSync(path, stringifyYamlDocument(data));
+}
+
```
PR-Reviewer-Ant, line 119: Warning: `writeAgentYaml` round-trips `agent.yaml` through `parseYamlDocument` → mutate → `stringifyYamlDocument`. YAML comments, key ordering, and inline formatting in a hand-crafted `agent.yaml` will be silently destroyed on the first `signet route pin` or `signet route unpin`. Users who rely on comments in `agent.yaml` for documentation or who version-control the file will see noisy diffs. Consider at minimum warning in the command output that formatting will change, or using a smarter in-place patch strategy.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2ef7c5e2`

Automated bot review (pr-reviewer). PR #462 introduces a large new routing module, new inference API endpoints, CLI route commands, and test/workspace stabilization. Several issues warrant attention before merge: a mandatory typecheck gate is explicitly failing, the hybrid-mode security fix is documented but the implementation is absent from the visible diff, a silent breaking change removes the `x-signet-actor` rate-limit attribution header, and `src/` is unexpectedly added to published npm artifacts.

**Confidence:** Medium [missing_cross_module_context, sufficient_diff_evidence] - The core routing decision algorithm (`resolveRoutingDecision`) and the fallback/scoring logic dominate the spec contract but fall in the truncated tail of `packages/core/src/routing.ts`. The Hono auth middleware that implements the documented hybrid-mode TCP-peer-address change is also absent from the diff, making it impossible to verify the security fix exists in code. Issues 1–6 below are directly provable from the visible diff; issue 4 requires seeing the middleware to confirm.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+			const status = await deps.fetchFromDaemon<RouteStatusResponse>("/api/inference/status");
+			if (!status) {
+				console.error(chalk.red("Failed to get routing status from daemon"));
+				process.exit(1);
+			}
+			if ((options as { json?: boolean }).json) {
+				console.log(JSON.stringify(status, null, 2));
+				return;
+			}
+			printStatus(status);
```
PR-Reviewer-Ant, line 231: Nitpick: `signet route list` and `signet route status` have identical implementations — both fetch `/api/inference/status` and call `printStatus`. This duplicates ~20 lines and diverges from AGENTS.md's 'Duplicate logic is a smell. Extract shared code where it clarifies.' guideline. One should delegate to the other, or both should share a single helper.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+				process.exit(1);
+			}
+			if (options.json) {
+				console.log(JSON.stringify(data, null, 2));
+				return;
+			}
+			const decision = data as {
+				targetRef?: string;
+				policyId?: string;
+				taskClass?: string;
```
PR-Reviewer-Ant, line 310: Nitpick: `data as { targetRef?: string; policyId?: string; ... }` and similar casts in `explain` and `test` actions use `as` to coerce `unknown` rather than narrowing. AGENTS.md explicitly prohibits `as` and requires narrowing with explicit null checks. The risk here is low (display-only paths), but it sets a precedent for the new surface.

packages/adapters/openclaw/src/index.test.ts
```diff
@@ -1,25 +1,11 @@
...
-	STATIC_IDENTITY_SESSION_START_TIMEOUT_STATUS:
-		"[signet: daemon session-start timed out — running with static identity]",
-}));
-
-// Import after mock so the module picks up the stub.
+// Import directly; tests seed a real temporary SIGNET_PATH instead of
+// mocking @signet/core globally, which otherwise leaks into later suites.
 const signet = await import("./index");
 const signetPlugin = signet.default;
 const { memoryStore, _resetRegistration } = signet;
```
PR-Reviewer-Ant: Nitpick: The module is imported at top-level (`const signet = await import('./index')`) before any `beforeEach` sets `process.env.SIGNET_PATH`. The original mock intercepted `readStaticIdentity` to handle the case where `~/.agents` does not exist at import time. The new approach relies on `SIGNET_PATH` being set before the module exercises `readStaticIdentity`, but `SIGNET_PATH` is only set inside `beforeEach`. If the module calls `readStaticIdentity` during its own initialization phase (the original comment said 'import after mock so the module picks up the stub'), this could produce different behavior on CI hosts where `~/.agents` does not exist. The PR notes tests pass locally — verify they also pass on a clean CI image where `~/.agents` is absent.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,377 @@
...
+
+function readAgentYaml(agentsDir: string): { readonly path: string; readonly data: Record<string, unknown> } {
+	const path = join(agentsDir, "agent.yaml");
+	if (!existsSync(path)) {
+		return { path, data: {} };
+	}
+	const parsed = parseYamlDocument(readFileSync(path, "utf-8"));
+	return { path, data: isRecord(parsed) ? parsed : {} };
+}
+
```
PR-Reviewer-Ant, line 105: Blocking: `route pin` / `route unpin` rewrites the full `agent.yaml` via `parseYamlDocument` + `stringifyYamlDocument`, which can strip user comments/formatting and silently lose hand-authored config context. This is a data-loss side effect on a common operator file. Please preserve comments/formatting (or apply a targeted edit), or at minimum gate with an explicit warning/confirmation path and document the destructive behavior.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `98b86aa9`

[Automated Review by pr-reviewer] Found a config-write behavior that can cause unintended data loss and should be fixed before merge.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The new CLI route pin/unpin path directly round-trips `agent.yaml` through parse+stringify and writes the whole file. In this codebase, YAML round-trips are already documented to strip comments; these commands do not warn or preserve comments, so the loss is directly inferable from the changed implementation. I did not see the full shared YAML utility internals in this diff, so confidence is medium rather than high.

docs/API.md
```diff
@@ -157,6 +170,11 @@ Monitor `providerResolution.extraction.status` for `degraded` or `blocked`
 states when the configured extraction provider is unavailable at startup.
 When `pipeline.extraction.overloaded` is `true`, the extraction worker is
 intentionally backing off for `overloadBackoffMs` between polls.
+The `inference` block summarizes the shared inference control plane status.
+`source` is `explicit` when a top-level `inference:` block is present in
+`agent.yaml`, `legacy-implicit` when extraction and synthesis providers are
+compiled into the router automatically, and `disabled` when no routeable
```
PR-Reviewer-Ant, line 176: The status section says "The `inference` block summarizes the shared inference control plane status", but the example response uses a `routing` block. This mismatch is likely to break client integrations generated from docs.

docs/PIPELINE.md
```diff
@@ -1021,11 +1021,17 @@ work should be mechanical and predictable.
 Configuration Reference
 ---
 
-All pipeline config lives under `memory.pipelineV2` in `agent.yaml` (see
+Most pipeline config lives under `memory.pipelineV2` in `agent.yaml` (see
 [[configuration]]). The config uses a nested structure with grouped
 sub-objects. Legacy flat keys are also supported for backward
 compatibility (nested keys take precedence).
 
+Provider selection for extraction and session synthesis can also be bound to
```
PR-Reviewer-Ant, line 1029: Config path appears incorrect: this change references top-level `routing.workloads`, but the rest of the PR/docs define top-level `inference.workloads`. Users following this section may configure the wrong key and silently fail to route extraction/synthesis through the control plane.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `6fdd73ea`

[Automated Review by pr-reviewer] I found documentation-level correctness/security issues that should be fixed before relying on these new routing/auth surfaces.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The issues are directly visible in changed docs and are internally contradictory across this PR (AUTH vs SELF-HOSTING, inference vs routing naming). I did not have full runtime/server implementation context in this diff to verify behavior-level parity for every described endpoint.


- docs/AUTH.md:35 - Security documentation contradicts this PR's self-hosting guidance: this file still states hybrid localhost bypass is based on spoofable `Host` header, while `docs/SELF-HOSTING.md` now says it uses TCP peer address. This can lead operators to make incorrect trust assumptions when exposing the daemon.

docs/API.md
NicholaiVogel: Fixed in `1487a25c`. The `/api/status` example now uses the `inference` key to match the surrounding status docs and the new canonical config surface.

docs/PIPELINE.md
NicholaiVogel: Fixed in `1487a25c`. This section now references `inference.workloads` and clarifies that legacy extraction/synthesis provider fields are only used to build the implicit compatibility profile when no explicit `inference:` block is configured.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,397 @@
...
+		.option("--operation <operation>", "Operation kind", "interactive")
+		.option("--privacy <privacy>", "Privacy tier")
+		.option("--policy <policy>", "Policy override")
+		.option("--target <targetRef>", "Pin to an explicit target ref")
+		.option("--max-tokens <maxTokens>", "Max output tokens")
+		.option("--refresh", "Refresh target health before inference")
+		.option("--debug", "Print the routed decision trace")
+		.option("--json", "Output as JSON")
+		.action(async (prompt: string, options) => {
+			const maxTokens = options.maxTokens ? Number.parseInt(options.maxTokens, 10) : undefined;
```
PR-Reviewer-Ant: `--max-tokens` is parsed with `Number.parseInt` but not validated. Non-numeric values (e.g. `--max-tokens abc`) become `NaN` and are forwarded in the request body, which can serialize unexpectedly and silently fall back to server defaults. Consider explicit CLI-side validation with a hard error for invalid numeric input.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `1487a25c`

[Automated Review by pr-reviewer] Found non-blocking issues worth discussion before human approval, primarily around security-documentation consistency and CLI input validation behavior.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The findings are directly grounded in changed files (`docs/AUTH.md`, `packages/cli/src/commands/route.ts`). I could verify the documented behavior and CLI request-shaping logic from the diff, but I could not fully inspect all daemon-side enforcement code paths from the provided truncated diff.


- docs/AUTH.md:34 - Security model documentation appears internally inconsistent with this PR’s other docs. `AUTH.md` still states hybrid localhost bypass is based on spoofable `Host` header, while `SELF-HOSTING.md` was updated to say peer socket address is used. This can mislead operators about real attack surface and deployment hardening requirements.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `cd79cd16`

[Automated Review by pr-reviewer] I found two issues to address: one correctness/security-doc mismatch and one CLI input-validation gap.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - Both findings are directly visible in changed files (`docs/AUTH.md`, `packages/cli/src/commands/route.ts`). I did not have full daemon route implementation context in this diff chunk to fully verify runtime behavior behind every documented claim.


- docs/AUTH.md:34 - Hybrid-mode localhost bypass documentation conflicts with other updated docs in this PR: this file still says localhost detection is based on the `Host` header (spoofable), while `docs/SELF-HOSTING.md` now says TCP peer address is used. This inconsistency can cause operators to misconfigure exposure assumptions and is a security-relevant documentation defect.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,397 @@
...
+		});
+
+	routeCmd
+		.command("test <prompt>")
+		.description("Execute a real routed prompt")
+		.option("--agent <agent>", "Agent id")
+		.option("--task-class <taskClass>", "Task class override")
+		.option("--operation <operation>", "Operation kind", "interactive")
+		.option("--privacy <privacy>", "Privacy tier")
+		.option("--policy <policy>", "Policy override")
```
PR-Reviewer-Ant, line 350: `--max-tokens` is parsed with `Number.parseInt` but not validated for `NaN`/range before sending to `/api/inference/execute`. Invalid values (for example `--max-tokens abc`) will be forwarded as non-sensical payload data and produce confusing failures. Add explicit CLI-side validation with a clear error path.

packages/cli/src/commands/route.ts
NicholaiVogel: Fixed in `11e5e37f`. `route test --max-tokens` now validates CLI input before calling `/api/inference/execute`; invalid, zero, negative, or non-integer values fail with a clear `--max-tokens must be a positive integer` error.

packages/cli/src/commands/route.ts
NicholaiVogel: Fixed in `11e5e37f`. Added the same CLI-side validation here, plus regression coverage for both the invalid path and the valid numeric request body.

packages/cli/src/commands/route.ts
```diff
@@ -0,0 +1,411 @@
...
+			console.log();
+			console.log(result.text ?? "");
+		});
+
+	routeCmd
+		.command("pin <targetRef>")
+		.description("Pin an agent/task-class to a target ref in agent.yaml")
+		.option("--agent <agent>", "Agent id", "default")
+		.option("--task-class <taskClass>", "Task class pin key", "default")
+		.option("--rewrite-agent-yaml", "Allow rewriting existing agent.yaml; comments and formatting may be changed")
```
PR-Reviewer-Ant, line 407: `route pin` accepts any `targetRef` string and persists it directly into `agent.yaml` without validating ref shape (`target/model`) or known target existence. This can write invalid routing config and cause route resolution failures later (effectively config corruption via CLI). Please validate using routing helpers (e.g., `parseRoutingTargetRef`) and fail fast before writing.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `11e5e37f`

[Automated Review] I found one blocking correctness issue in the new `signet route` command surface, plus one security-doc divergence to resolve.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The CLI pin path in `packages/cli/src/commands/route.ts` writes user-provided target refs directly into `agent.yaml` without validating `target/model` format or existence, which can persist invalid routing config and break downstream routing behavior. I can prove this from the changed file; I did not have full daemon runtime code in this diff to verify all downstream guardrails.


- docs/AUTH.md:35 - Security behavior docs diverge: this page still says hybrid localhost bypass uses `Host` header and is spoofable, while `docs/SELF-HOSTING.md` now documents TCP peer-address enforcement. That inconsistency can lead operators to incorrect threat assumptions and deployment choices. Align these docs to one authoritative behavior.

packages/cli/src/commands/route.ts
NicholaiVogel: Fixed in `cb9245b9`. `route pin` now validates the ref shape with the routing parser before writing, normalizes the stored `target/model` value, and when local `inference.targets` are configured it rejects unknown refs with the known target list. Added regression coverage for malformed refs and unknown configured targets.

packages/daemon/src/pipeline/worker.ts
NicholaiVogel: Outdated by the current branch. `packages/daemon/src/pipeline/worker.ts` is no longer changed in this PR, and the current failure path still increments `consecutiveFailures` after `failJob(...)` instead of resetting on dead jobs.

PR #464, opened 2026-04-04
feat(pipeline): add token-bucket rate limiting for remote LLM providers

### LINKED ISSUE #341: Feature request: rate limiting and cost protection for claude-code LLM provider
## Summary

The `claude-code` provider in `createClaudeCodeProvider()` spawns `claude -p` subprocesses with no daemon-level awareness of call frequency, accumulated cost, or whether calls are producing useful results. When combined with a pipeline stall loop or aggressive synthesis scheduling, this can produce thousands of Claude CLI invocations per day - each consuming API tokens via OAuth subscription - with no circuit breaker, budget cap, or rate limit within Signet itself.

---

## Current Behavior

The `claude-code` provider has a `SubprocessSemaphore` for concurrency control and a per-call `timeoutMs`, but no:

- **Per-hour or per-day call budget**: No limit on how many times `claude -p` can be spawned in a time window
- **Cost accumulator**: No tracking of cumulative token usage or estimated cost from the daemon's LLM calls
- **Idle/waste detection**: No awareness of whether synthesis or extraction calls are producing useful output (e.g., "last 50 calls returned 0 new facts")
- **Rate limiting**: The only rate protection is whatever the external `claude` binary enforces (in our case, a custom wrapper with a 300/hr cap - without that, calls would be unlimited)

During a pipeline stall on our deployment, this produced:

```
Mar 23: ~7,800 Claude CLI spawns (from 05:36 UTC onward)
Mar 24: ~16,400 Claude CLI spawns (full 24h)
Mar 25: ~600 Claude CLI spawns (until manual fix at 04:55 UTC)
Total: ~24,800 subprocess spawns over ~56 hours
```

Each spawn hits the Anthropic API via OAuth, consuming at minimum cache read tokens. Estimated cost: $400-900 over the incident period.

---

## Desired Behavior

The `claude-code` provider should have configurable rate limiting and cost protection that prevents runaway subprocess spawning, independent of whether the caller (extraction worker, synthesis worker, dependency-synthesis) has its own safeguards.

---

## Use Cases

**1. Pipeline stall with claude-code provider**
A stuck extraction job triggers the watchdog loop (#311/#248). Dependency-synthesis keeps calling the claude-code provider every 60 seconds. Without provider-level rate limiting, calls accumulate unbounded. A 50/hr rate limit at the provider level would cap damage to ~$15-25/day instead of $200-450/day.

**2. Misconfigured synthesis on always-on deployment**
An operator configures `synthesis.provider: claude-code` with `synthesis.model: sonnet` (the default) on a VPS that runs 24/7. Even healthy pipeline operation produces continuous LLM calls for extraction, synthesis, dependency-synthesis, and structural classification. Without cost tracking, monthly spend can grow silently. A daily budget cap would surface this.

**3. Subprocess leak or crash loop**
If the `claude` binary exits immediately (e.g., auth token expired, binary missing), the provider retries rapidly. The timeout fires after `defaultTimeoutMs` (120s), but the spawn itself is near-instant. A rate limiter would catch rapid spawn-exit cycles.

---

### Option A: Token bucket rate limiter in createClaudeCodeProvider (recommended)

Add a simple token bucket to the provider that limits calls per time window:

```typescript
function createClaudeCodeProvider(config) {
    const rateLimiter = new TokenBucket({
        maxTokens: config.maxCallsPerHour ?? 100,
        refillRate: (config.maxCallsPerHour ?? 100) / 3600,
    });

    async function callClaude(prompt, outputFormat, opts) {
        if (!rateLimiter.tryConsume(1)) {
            throw new Error(`claude-code rate limit: ${config.maxCallsPerHour}/hr exceeded`);
        }
        return withSemaphore(async () => {
            // existing spawn logic
        });
    }
}
```

New config fields:

```yaml
pipelineV2:
  synthesis:
    provider: claude-code
    maxCallsPerHour: 100      # new: provider-level rate limit
    maxDailyBudgetUsd: 5.00   # new: estimated daily cost cap (requires usage tracking)
```

### Option B: Cost accumulator with kill switch

Track `generateWithUsage` return values and accumulate estimated cost. When cumulative cost exceeds a configurable daily budget, stop making calls and log an alert:

```typescript
let dailyCostUsd = 0;
const dailyBudget = config.maxDailyBudgetUsd ?? 10.0;

async generateWithUsage(prompt, opts) {
    if (dailyCostUsd >= dailyBudget) {
        throw new Error(`claude-code daily budget exceeded: $${dailyCostUsd.toFixed(2)} >= $${dailyBudget}`);
    }
    const result = await callClaude(prompt, "json", opts);
    if (result.usage?.totalCost) {
        dailyCostUsd += result.usage.totalCost;
    }
    return result;
}
```

Option A is simpler and doesn't require cost estimation. Option B is more precise but depends on the Claude CLI returning accurate `cost_usd` in JSON output mode.

---

## Expected Impact

- Prevents unbounded API spend during pipeline anomalies
- Makes `claude-code` provider safe for always-on deployments without external wrapper scripts
- Gives operators visibility into daemon-level LLM consumption
- Complements #320 (provider preflight) and #321 (load-aware backpressure) as defense-in-depth

---

## Related

- #311 - Stuck jobs causing watchdog loop (triggered our cost incident)
- #248 - Extraction worker stall (introduced the watchdog that amplifies the loop)
- #320 - Provider preflight check at startup (related: provider health awareness)
- #321 - System-load-aware backpressure (complementary: host-level protection)
- #339 - Watchdog query bug (the stall amplifier)

aaf2tbz:
## Summary

Supersedes #461, which cannot be reopened after its head branch was deleted and recreated.

Closes #341 (Option A — token bucket rate limiter)

Prevents runaway subprocess spawning during pipeline stall loops by throttling calls to remote/paid LLM providers via a token-bucket rate limiter. This is the "Option A" implementation from the issue — a simple, effective rate limiter that doesn't require cost estimation.

### `packages/core/src/types.ts`
- Add `ProviderRateLimitConfig` interface and default export for provider rate-limiting
- Keep direct/programmatic callers on the same partial-config contract as YAML parsing

### `packages/daemon/src/pipeline/provider.ts`
- Add `TokenBucketRateLimiter` class with fractional refill, burst allowance, and optional blocking wait
- Add `RateLimitExceededError` thrown when rate limit is exceeded after wait timeout
- Add `withRateLimit()` decorator that wraps `LlmProvider` to enforce rate limits before `SubprocessSemaphore.acquire()`
- Split pure `stats` reads from wall-clock-refreshed `currentStats()` reads for observability

### `packages/daemon/src/daemon.ts`
- Wrap extraction provider with `withRateLimit()` before `initLlmProvider()`
- Wrap synthesis provider with `withRateLimit()` before `initSynthesisProvider()`
- Keep widget generation on a separate provider instance so widget traffic does not consume the synthesis rate-limit bucket

### `packages/daemon/src/memory-config.ts`
- Parse `rateLimit` for both extraction and synthesis pipeline config

### `packages/daemon/src/pipeline/worker.ts`
- Treat provider rate-limit exhaustion as non-retryable and clarify why `failJob()` is forced down the dead-job path

### `docs/CONFIGURATION.md`
- Document `rateLimit.maxCallsPerHour`, `burstSize`, and `waitTimeoutMs`
- Clarify opt-in behavior, `rateLimit: {}` no-op behavior, and the YAML-vs-helper handling of zero values

## Reviewer round 1 fixes (`83da2019`)

- **`generateWithUsage` this-binding** — the captured method was called detached, losing the `this` context. Replaced the intermediate capture with a direct `provider.generateWithUsage!(prompt, opts)` call inside the spread guard, preserving the receiver (the `!` is safe because the spread condition already confirms it exists).
- **Dead-letter documentation** — added explicit warning in `docs/CONFIGURATION.md` (extraction and synthesis sections) that rate-limited jobs go to dead-letter without retry and are purged after 30 days. Users must size `maxCallsPerHour` to handle sustained ingestion bursts.

## Reviewer round 2 fixes (`5bf96717`)

- **PR title `fix` → `feat`** — this is a user-facing feature (new `rateLimit:` YAML stanza), not a bug fix. Corrected to `feat(pipeline):` for proper semver minor bump.
- **Latent NaN guard** — `undefined` config fields merged via spread bypassed the `<= 0` guard (`undefined <= 0` is `false` in JS), constructing a limiter with `NaN` tokens that rejects every call. Fixed with nullish-coalesce: `(cfg.maxCallsPerHour ?? 0) <= 0`.
- **Silent pass-through warning** — `withRateLimit()` now logs `logger.warn` when a `rateLimit` config is provided but the provider is not in `RATE_LIMIT_PROVIDERS` (e.g. `command`, `ollama`). Previously this was entirely silent, making misconfiguration hard to diagnose. Updated `docs/CONFIGURATION.md` to document the exemption.
- **`!` → `?.`** — replaced non-null assertion on `provider.generateWithUsage!()` with optional chain to satisfy Biome lint. The spread guard already guarantees the method exists; the `?.` is semantically equivalent here.
- **Regression tests** — added tests for `undefined` config fields (NaN path) and `command` provider pass-through.

## Validation

- `bun test src/pipeline/rate-limit.test.ts` in `packages/daemon`
- `bun test src/pipeline/worker.test.ts -t "marks job dead without retry when provider rate limit is exceeded"` in `packages/daemon`
- `bunx biome check packages/daemon/src/pipeline/provider.ts packages/daemon/src/pipeline/worker.ts packages/core/src/types.ts docs/CONFIGURATION.md`

## PR checklist


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): add token-bucket rate limiting for remote LLM providers` (commit `419ae257`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,156 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+					logger.warn("pipeline", `Rate limit throttled ${provider.name} (${stats.totalThrottled} total)`, stats);
+					lastWarnMs = Date.now();
+				}
+				throw new RateLimitExceededError(provider.name, cfg.maxCallsPerHour);
+			}
+			return provider.generate(prompt, opts);
+		},
+
+		...(generateWithUsage
+			? {
```
PR-Reviewer-Ant, line 238: `generateWithUsage` is captured as a detached function and called without binding it back to the original provider:

```ts
const generateWithUsage = provider.generateWithUsage;  // detached
// ...
return generateWithUsage(prompt, opts);               // this = undefined
```

Compare with the `generate` wrapper, which correctly preserves the receiver:

```ts
return provider.generate(prompt, opts);  // this = provider
```

Any provider whose `generateWithUsage` method references `this` (e.g. a class-based provider that stores client state in `this.client`) will throw or silently misbehave at runtime. The Anthropic and OpenRouter providers are the primary targets of this rate limiter, so the risk is real even if current closure-based implementations happen to be unaffected.

**Fix:** replace the detached call with `return generateWithUsage.call(provider, prompt, opts)`, or drop the capture and write `return provider.generateWithUsage!(prompt, opts)` (the `!` is safe here because the spread guard already checked it is defined).

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1534,27 +1533,39 @@ export function startWorker(
...
+						? "EXTRACTION_RATE_LIMIT"
+						: msg.includes("timeout")
+							? "EXTRACTION_TIMEOUT"
+							: "EXTRACTION_PARSE_FAIL",
 					durationMs: runtime.now() - jobStart,
 				});
 				accessor.withWriteTx((db) => {
-					failJob(db, job.id, msg, job.attempts, job.max_attempts);
-					if (job.attempts >= job.max_attempts) {
+					// failJob() uses the attempts argument only to classify dead vs retryable.
```
PR-Reviewer-Ant: A job that hits the rate limit on its **first** attempt is immediately sent to dead-letter (`failJob` receives `job.max_attempts` as `attempts`, so the job is classified as dead on attempt 1). The dead-letter purge window is 30 days. There is no mechanism to re-enqueue a dead job once the rate-limit window (1 hour) resets.

Consequence: if a brief burst of incoming memories temporarily exhausts the rate limit, every memory queued during that burst will have its extraction status permanently set to `'failed'` and will never be re-processed — even though the provider would be available again within the hour.

This is a data-loss path for users who configure `rateLimit.maxCallsPerHour` too low relative to their actual ingestion rate. Neither `docs/CONFIGURATION.md` nor the in-code comment documents this behaviour. At minimum, the docs should warn that dead-lettered rate-limit jobs are not retried and that users must choose `maxCallsPerHour` high enough to avoid sustained exhaustion.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,156 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	// that need a fresh view should opt into a refill before reading the snapshot.
+	currentStats(): { readonly remaining: number; readonly totalConsumed: number; readonly totalThrottled: number } {
+		this.refill();
+		return {
+			remaining: Math.floor(this.tokens),
+			totalConsumed: this.totalConsumed,
+			totalThrottled: this.totalThrottled,
+		};
+	}
+}
```
PR-Reviewer-Ant, line 170: Per AGENTS.md, `feat:` is reserved for user-facing features and bumps the minor version; `fix:` bumps patch. Adding a new `rateLimit:` YAML stanza with documented config fields is a user-facing feature, not a bug fix. The commit type `fix(pipeline):` will produce an incorrect patch-version bump. This should be `feat(pipeline):`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `419ae257`

[Automated Review — pr-reviewer bot] PR #464 adds a token-bucket rate limiter for remote LLM providers. One blocking correctness issue found in the `withRateLimit` decorator: `generateWithUsage` is called without its `this` context. One warning-level design concern: rate-limited jobs go to dead-letter on first failure with no recovery path.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The `this`-binding inconsistency is directly visible in the diff: `generate` is called as `provider.generate(prompt, opts)` (method call, `this` = provider) while the captured `generateWithUsage` is called as `generateWithUsage(prompt, opts)` (detached, `this` = undefined in strict mode). A grep of the live provider.ts shows existing implementations use closures (`callOllama`, `callClaude`) rather than `this`, so the bug is latent today but will silently break any future class-based provider. I cannot rule out that a currently-unseen provider in the file does use `this` without reading the full implementations at lines 1123, 1348, and 2017.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,154 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+
+const RATE_LIMIT_PROVIDERS: ReadonlySet<string> = new Set([
+	"claude-code",
+	"anthropic",
+	"openrouter",
+	"codex",
+	"opencode",
+]);
```
PR-Reviewer-Ant, line 178: **Latent NaN bug: `undefined <= 0` is `false` in JS.**

`ProviderRateLimitConfig` declares all three fields as optional (`maxCallsPerHour?`, `burstSize?`, `waitTimeoutMs?`). If a caller passes `{ maxCallsPerHour: undefined, burstSize: 5 }`, the spread on line 177 produces `cfg.maxCallsPerHour = undefined` (the spread override wins over the default). Then `undefined <= 0` evaluates to `false`, bypassing the guard, and `TokenBucketRateLimiter(undefined, 5)` is constructed.

In `refill()`: `(undefined / 3_600_000) * elapsed = NaN` → `tokens = Math.min(burstSize, tokens + NaN) = NaN` → `acquire()` always returns `false` → every LLM call throws `RateLimitExceededError` and goes to dead-letter.

No current call site triggers this — `parseRateLimitConfig` always fills all three fields before returning — but the type allows it and the guard doesn't defend against it.

**Fix:** Replace the guard with `if ((cfg.maxCallsPerHour ?? 0) <= 0 || (cfg.burstSize ?? 0) <= 0) return provider;`

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,154 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+
+function shouldRateLimit(providerName: string): boolean {
```
PR-Reviewer-Ant, line 180: **Silent pass-through for non-whitelisted providers (e.g. `command`).**

When `shouldRateLimit(provider.name)` returns `false`, the wrapper silently returns the original provider even if the caller explicitly passed a `rateLimit` config. This affects any provider not in `RATE_LIMIT_PROVIDERS` — most notably the `command` provider, but also any future custom provider.

Users who configure `rateLimit` on a `command`-backed extraction provider will see no throttling and no indication that the config was ignored. At minimum, a `logger.warn` here would prevent silent misconfiguration. The docs don't mention this pass-through for `command` providers.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `9e9985e5`

[Automated Review — pr-reviewer bot] PR #464 adds a token-bucket rate limiter for remote LLM providers. One latent correctness bug and one silent-skip behavior were found. No security issues or data-corruption paths in the current call graph.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly provable from the changed lines in provider.ts. The NaN path in TokenBucketRateLimiter is fully traceable: `{maxCallsPerHour: undefined}` → spread doesn't override DEFAULT → `undefined <= 0 === false` in JS → TokenBucketRateLimiter constructed with undefined → `(undefined / 3_600_000) * elapsed = NaN` → `Math.min(burstSize, tokens + NaN) = NaN` → `acquire()` always returns false → every call throws RateLimitExceededError. The ProviderRateLimitConfig type (types.ts:197-201) declares all fields optional, making this a real (if currently untriggered) path.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,164 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	const base = providerName.split(":")[0];
+	return RATE_LIMIT_PROVIDERS.has(base);
+}
+
+export function withRateLimit(provider: LlmProvider, config?: Partial<ProviderRateLimitConfig>): LlmProvider {
+	if (config === undefined) return provider;
+	if (Object.keys(config).length === 0) return provider;
```
PR-Reviewer-Ant, line 187: `return provider.generateWithUsage?.(prompt, opts)` can resolve to `undefined` if `provider.generateWithUsage` is absent at call time. TypeScript does not narrow `provider.generateWithUsage` inside the closure body even though the conditional spread checked it at object-construction time. The `?.` was added to satisfy Biome's no-non-null-assertion rule, but it changes the effective return type to `LlmGenerateResult | undefined`, not the declared `Promise<LlmGenerateResult>`. Any caller that awaits the wrapped method and does not guard for `undefined` (e.g. `generateWithTracking` if it calls `provider.generateWithUsage(...)` without `?.`) would silently receive `undefined`. A safer pattern that also satisfies Biome: capture `const fn = provider.generateWithUsage;` before closing over it, then call `fn.call(provider, prompt, opts)` — TypeScript and Biome both accept this without `!` or `?.`.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1534,27 +1533,39 @@ export function startWorker(
...
+					// Pass max_attempts here so provider rate-limit exhaustion skips retries.
```
PR-Reviewer-Ant: `failJob(db, job.id, msg, nonRetryable ? job.max_attempts : job.attempts, job.max_attempts)` passes `max_attempts` as the current-attempts argument for rate-limited failures. The inline comment says this argument is used only for classification (dead vs. retryable), but if `failJob` also writes this value to the `attempts` column — which is the common pattern for this kind of helper — every dead-lettered rate-limit job will show `attempts = max_attempts` (e.g. 3) in the DB even though only 1 attempt was made. This corrupts job observability: dashboards, analytics, and post-mortem queries that filter or aggregate by attempts count will misread the failure as an exhausted-retry job rather than an immediate dead-letter. If `failJob` truly uses the argument only for classification and does not persist it, this is fine — but that invariant should be verified and, if true, commented more explicitly or enforced with an assertion.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,164 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+		this.lastRefillMs = now;
+	}
+
+	async acquire(waitMs: number): Promise<boolean> {
+		this.refill();
+		if (this.tokens >= 1) {
+			this.tokens -= 1;
+			this.totalConsumed++;
+			return true;
+		}
```
PR-Reviewer-Ant, line 132: `const pollIntervalMs = Math.min(100, waitMs / 4)` produces sub-millisecond values when `waitMs` is tiny (e.g. `waitMs = 1` → `pollIntervalMs = 0.25`, which `setTimeout` rounds up to 1 ms or 0 ms depending on the runtime). At 0 ms, the loop becomes a near-busy-wait. Consider `Math.max(1, Math.min(100, Math.floor(waitMs / 4)))` to avoid CPU spin on callers that pass small `waitTimeoutMs` values.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5bf96717`

[Automated Review — pr-reviewer bot] Two findings worth discussing before merge: a type-unsafety in the `generateWithUsage` wrapper that can silently return `undefined`, and a misleading attempts count written to the DB when rate-limit jobs are dead-lettered.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The `generateWithUsage` optional-chain return type mismatch is directly visible in the diff at the wrapper construction site. The `failJob` attempts-count concern requires `failJob`'s implementation to confirm whether the 4th argument is stored as-is or only used for classification — that function is not in the diff.

packages/daemon/src/pipeline/provider.ts
aaf2tbz: Already addressed in the current diff — the guard uses nullish coalescing: `if ((cfg.maxCallsPerHour ?? 0) <= 0 || (cfg.burstSize ?? 0) <= 0) return provider;`. This handles `undefined` correctly since `undefined ?? 0` yields `0`, and `0 <= 0` is `true`. Also covered by test: "returns provider unwrapped when maxCallsPerHour is undefined (NaN guard)".

packages/daemon/src/pipeline/provider.ts
aaf2tbz: Addressed — line 192 now logs `logger.warn` when a non-whitelisted provider receives rate-limit config, including the provider name and the set of allowed providers. Docs updated to document the pass-through behavior.

packages/daemon/src/pipeline/provider.ts
aaf2tbz: Fixed in c051d2a4. Replaced `provider.generateWithUsage?.(prompt, opts)` with `const fn = provider.generateWithUsage; return fn.call(provider, prompt, opts);`. This captures the reference before the closure, giving TypeScript stable narrowing and satisfying Biome without `?.` or `!`.

packages/daemon/src/pipeline/worker.ts
aaf2tbz: Verified — `failJob` does NOT persist the `attempts` argument to the DB. It only uses it for dead-vs-retryable classification: `const status = attempts >= maxAttempts ? "dead" : "failed"`. The SQL updates only `status`, `error`, `failed_at`, and `updated_at`. Strengthened the comment in c051d2a4 to make this invariant explicit.

packages/daemon/src/pipeline/provider.ts
aaf2tbz: Fixed in c051d2a4. Changed to `Math.max(1, Math.floor(Math.min(100, waitMs / 4)))` — floors the value and clamps to a minimum of 1ms to prevent CPU spin on tiny `waitTimeoutMs` values.

packages/daemon/src/pipeline/provider.ts
aaf2tbz: Still valid point — confirmed the guard now uses nullish coalescing (`?? 0`) on both fields, and the test suite covers undefined inputs explicitly.

packages/daemon/src/pipeline/worker.ts
aaf2tbz: failJob invariant verified and comment strengthened in c051d2a4 — the attempts argument is classification-only, never persisted to the attempts column.

packages/daemon/src/pipeline/provider.ts
aaf2tbz: Fixed in c051d2a4 — generateWithUsage now uses `const fn = provider.generateWithUsage; return fn.call(provider, prompt, opts);` instead of optional chaining.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,165 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+			},
+		);
+		return provider;
+	}
+
+	const bucket = new TokenBucketRateLimiter(cfg.maxCallsPerHour, cfg.burstSize);
+	let lastWarnMs = 0;
+	const WARN_INTERVAL_MS = 300_000;
+
+	return {
```
PR-Reviewer-Ant, line 226: `lastWarnMs` is shared between the `generate` and `generateWithUsage` closures. Because both closures `await bucket.acquire(...)` before reading/writing `lastWarnMs`, two concurrent rate-limited calls (one via each method) can both pass the `Date.now() - lastWarnMs > WARN_INTERVAL_MS` check before either updates the variable, emitting duplicate warn entries. This only affects log volume, not correctness, but under sustained exhaustion it defeats the intent of the 5-minute throttle. Consider a single warn-per-throttle entry emitted from `bucket.currentStats()` inside `acquire` itself, or move `lastWarnMs` into the `TokenBucketRateLimiter` instance.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c051d2a4`

[Automated PR Review — pr-reviewer bot] PR #464 adds a token-bucket rate limiter for remote LLM providers in the extraction and synthesis pipelines. All 8 findings from rounds 1–3 were addressed. One new blocking issue was found: the documentation added in this PR explicitly states that rate-limited synthesis jobs are dead-lettered without retry, but `summary-worker.ts` (not modified in this PR) has no `instanceof RateLimitExceededError` check and will retry synthesis jobs up to `max_attempts` before dead-lettering — directly contradicting the documented guarantee. A secondary minor issue: `lastWarnMs` is shared between the `generate` and `generateWithUsage` closures, allowing log-spam under concurrent rate exhaustion.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The doc claim is in the added lines of docs/CONFIGURATION.md (synthesis rateLimit section). The summary-worker.ts catch block was fetched directly: `isTerminalSummaryJobError` only matches the immutable-artifact error prefix, so RateLimitExceededError is treated as a transient error and the job is reset to 'pending'. This was verified against the actual file — not inferred. The extraction worker path (worker.ts) correctly dead-letters via the `nonRetryable` guard added in this PR; the synthesis path does not receive equivalent treatment.


- docs/CONFIGURATION.md:463 - The claim 'Rate-limited synthesis jobs that fail are sent to dead-letter without retry' is incorrect as implemented. `summary-worker.ts` catches all errors and passes the message through `isTerminalSummaryJobError()`, which only matches the immutable-artifact prefix — not `RateLimitExceededError`. A rate-limit failure is therefore treated as transient: the job is reset to `'pending'` and retried until `attempts >= max_attempts` (default 3). With `waitTimeoutMs: 5000` (default), each retry blocks the synthesis queue for up to 5 seconds while `bucket.acquire()` polls, meaning a burst-exhausted synthesis queue stalls for up to 15 seconds before the job finally dead-letters.

This PR adds the non-retryable path only in `worker.ts` (extraction jobs). The same `instanceof RateLimitExceededError` guard needs to be added to `summary-worker.ts`'s catch block — either by calling `isTerminalSummaryJobError` for it (add a check on error type, not just message prefix), or by an explicit early-dead-letter branch analogous to the extraction worker change at `worker.ts:1533–1574`.

packages/cli/dashboard/src/lib/components/os/SidebarGroups.svelte
```diff
@@ -47,11 +124,280 @@ function handleGroupDragOver(e: DragEvent): void {
...
+
+	for (const [index, memory] of memories.entries()) {
+		const tags = new Set(parseTags(memory.tags));
+		const content = String(memory.content ?? "").toLowerCase();
+		const sourceType = String(memory.source_type ?? "").toLowerCase();
+		const browserScoped =
+			sourceType === "browser-extension" ||
+			tags.has("browser-extension") ||
+			content.includes("[signet browser tool]") ||
+			content.includes("[signet bookmark]") ||
```
PR-Reviewer-Ant: **Scoping leak — AGENTS.md rule #1 violation.** `refreshBrowserCategories` fetches `/api/memories?limit=500&offset=0` with no `agent_id` parameter. AGENTS.md explicitly requires threading `agent_id` through every read that touches user data and forbids hardcoding `'default'` for scoped paths. In a multi-agent setup this exposes all memories from all agents in the sidebar. Even in single-agent setups, omitting the parameter relies on implicit server-side defaulting, which is fragile. Add `agent_id` from the active agent context, or at minimum use a consistent explicit default if the daemon enforces it.

packages/cli/dashboard/src/lib/components/os/AgentChat.svelte
```diff
@@ -32,13 +47,281 @@ interface ChatMessage {
...
+				0?: { transcript?: string };
+				item?: (index: number) => { transcript?: string } | null;
+			} | null;
+		};
+	};
+	const list = speechEvent.results;
+	if (!list) return "";
+
+	let transcript = "";
+	for (let index = 0; index < list.length; index += 1) {
```
PR-Reviewer-Ant: **Voice mic selection is silently non-functional.** `recognition.start(selectedTrack)` is called with a `MediaStreamTrack` argument, but the Web Speech API spec defines `SpeechRecognition.start()` as taking no arguments — the track is silently ignored in every browser. The try/catch fallback always succeeds via `recognition.start()`, so the selected microphone never actually affects speech input. `getUserMedia` runs (triggering the permission prompt) but the acquired track is discarded. Either remove the fake track argument and document that mic selection is advisory-only, or replace `SpeechRecognition` with a `MediaRecorder`-based approach that actually routes the chosen device.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2d8182e0`

[Automated Review — pr-reviewer bot] This PR is titled as a pipeline rate-limiting feature but the diff is dominated by unreferenced dashboard UI changes (AgentChat voice input, render-window tab system, SidebarGroups browser-sync panel). The rate-limiting core code (provider.ts, worker.ts, daemon.ts, memory-config.ts) is truncated from the diff, making the primary stated feature unverifiable. One clear scoping violation and one silent functional bug are flagged below.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The scoping violation in SidebarGroups.svelte is directly readable in the diff (the fetch call has no agent_id parameter). The voice recognition bug is provable from the Web Speech API spec — SpeechRecognition.start() does not accept a MediaStreamTrack argument, so the try/catch always falls through to the trackless path. However, the core rate-limiting implementation (provider.ts, worker.ts, daemon.ts) is truncated and cannot be reviewed, so confidence in the overall feature correctness is limited.

- packages/cli/dashboard/src/lib/components/tabs/OsTab.svelte:1 - **PR description does not cover the dashboard changes.** The description, checklist, and validation steps reference only rate-limiting (provider.ts, worker.ts, daemon.ts, memory-config.ts, CONFIGURATION.md). The actual diff introduces a full render-window tab system replacing WidgetGrid, a voice-input-capable AgentChat refactor, a SidebarGroups browser-sync panel, and system-theme watching. None of these are mentioned in the description, changelog, or spec references. Per AGENTS.md, spec alignment must be validated for architectural changes and cross-package coordination. These UI changes should either be split to their own PR or documented explicitly in this one.

packages/cli/dashboard/src/lib/components/tabs/OsTab.svelte
```diff
@@ -95,45 +270,166 @@ function resolveDefaultSize(appId: string): { w: number; h: number } {
...
+						<div class="os-render-window-tabs-actions">
+							<button
+								class="os-render-action-btn"
+								title="Refresh render window"
+								onclick={() => void handleRenderRefresh()}
+								disabled={!renderWindowEntry}
+							>
+								<RefreshCw class="size-3" />
+							</button>
+							<button
```
PR-Reviewer-Ant: **Tab close button is keyboard-inaccessible.** The `os-render-tab-close` element is a `<span>` with only an `onclick` handler — no `role="button"`, `tabindex`, or `onkeydown`. Keyboard users cannot close render tabs. Given the rest of the tab UI is fully navigable, this is an inconsistency. Replace with a `<button>` or add the missing ARIA attributes.

packages/cli/dashboard/src/lib/components/os/SidebarGroups.svelte
```diff
@@ -47,11 +124,280 @@ function handleGroupDragOver(e: DragEvent): void {
...
+
+onMount(() => {
+	void refreshSkills();
+	void refreshBrowserCategories();
+	const timer = window.setInterval(() => {
+		void refreshSkills();
+		void refreshBrowserCategories();
+	}, 30_000);
+	return () => window.clearInterval(timer);
+});
```
PR-Reviewer-Ant: **30-second polling of a 500-record memory fetch is heavy.** The `setInterval` at 30 s calls both `refreshSkills()` and `refreshBrowserCategories()` on every tick. The memory fetch retrieves up to 500 rows to classify browser-sourced memories. As the memory store grows this will apply non-trivial read pressure every half-minute. Consider filtering server-side with a `source_type=browser-extension` query parameter (if the daemon supports it) or only refreshing when the sidebar panel is expanded.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -101,7 +102,7 @@ export function resolveSummaryHeadingDate(job: Pick<SummaryJobRow, "ended_at" |
 }
```
PR-Reviewer-Ant, line 102: `isTerminalSummaryJobError` now includes a magic-string check `errorMessage.startsWith("Rate limit exceeded")`. The call site in the same file then adds a redundant `|| e instanceof RateLimitExceededError` guard (line ~1632). This creates two separate classification paths that can silently diverge: if the error message format in `RateLimitExceededError` ever changes, the string check in `isTerminalSummaryJobError` breaks, but the `instanceof` check at the call site would still work. Conversely, `isTerminalSummaryJobError` (an exported, testable function) now makes a fragile assumption about message format. Consider either (a) importing `RateLimitExceededError` into `isTerminalSummaryJobError` and using `instanceof`, or (b) removing the redundant `instanceof` check at the call site and accepting that `isTerminalSummaryJobError` owns the classification contract — the string check is sufficient as long as the message prefix is stable.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,166 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	get stats(): { readonly remaining: number; readonly totalConsumed: number; readonly totalThrottled: number } {
+		return {
+			remaining: Math.floor(this.tokens),
+			totalConsumed: this.totalConsumed,
+			totalThrottled: this.totalThrottled,
+		};
+	}
+
+	// Stats used for observability should reflect the current wall clock, so callers
+	// that need a fresh view should opt into a refill before reading the snapshot.
```
PR-Reviewer-Ant, line 161: `withRateLimit` merges with `DEFAULT_PROVIDER_RATE_LIMIT` via object spread, then guards on `(cfg.burstSize ?? 0) <= 0`. If a direct caller passes `{ maxCallsPerHour: 100, burstSize: undefined }`, `undefined` wins over the spread default (20), so `cfg.burstSize` becomes `undefined`, the guard fires, and rate limiting is silently disabled despite `maxCallsPerHour` being set. The YAML path is safe (parseRateLimitConfig always fills defaults before calling withRateLimit), but this is a footgun for future programmatic callers. The PR description documents the `burstSize: 0` disable behavior but not the `burstSize: undefined` case. A one-line comment near the guard would prevent confusion.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d792513a`

[Automated Bot Review] Implementation is sound; prior blocking issues confirmed addressed. Two minor findings worth discussing.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The token-bucket implementation, guard conditions, this-binding fix, NaN guard, dead-letter path, and config parsing are all directly visible and provable from the diff. Prior round 1/2 concerns (this-binding, NaN, warn-on-exempt, test coverage) are confirmed resolved by the new code and tests. The two remaining findings are proven from specific lines in the diff.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `6cdb5d42`

[Automated Review — pr-reviewer bot] The rate-limiting implementation is correct and well-tested. Two items warrant discussion: an unrelated Codex provider config change bundled silently, and a minor edge-case inconsistency in `waitTimeoutMs` handling for direct programmatic callers. All blocking issues identified in prior review rounds (this-binding, NaN token bucket, missing `isTerminalSummaryJobError` guard in summary-worker) are confirmed addressed in this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The NaN guard fix (`(cfg.maxCallsPerHour ?? 0) <= 0`), this-binding fix (`fn.call(provider, prompt, opts)`), and summary-worker terminal classification (`error instanceof RateLimitExceededError`) are all visible and verifiable in the diff. The `waitTimeoutMs: undefined` edge case is provable from the guard logic at provider.ts lines ~185–188 (checks only `maxCallsPerHour` and `burstSize`, not `waitTimeoutMs`) combined with `acquire`'s `Date.now() < NaN === false` behavior. The Codex change is present in the diff at provider.ts line ~1433.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1256,7 +1427,9 @@ export function createCodexProvider(config?: Partial<CodexProviderConfig>): LlmP
 				"--sandbox",
 				"read-only",
 				"-c",
-				"mcp_servers.signet.enabled=false",
+				// Keep MCP disabled for daemon-side Codex runs without creating
+				// partial per-server config that newer Codex rejects.
+				"mcp_servers={}",
 				"-C",
```
PR-Reviewer-Ant, line 1435: This change is unrelated to rate limiting and is not mentioned in the PR description, title, or checklist. The original `-c mcp_servers.signet.enabled=false` disables one named server; `-c mcp_servers={}` replaces the *entire* MCP server map with an empty object. If the user's Codex config defines other MCP servers, this silently removes them for daemon-side subprocess calls. The comment only references "partial per-server config that newer Codex rejects" — if that's a real Codex CLI regression, it should be described and ideally tested. This belongs in a dedicated fix commit.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,172 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+
+function shouldRateLimit(providerName: string): boolean {
+	const base = providerName.split(":")[0];
+	return RATE_LIMIT_PROVIDERS.has(base);
+}
+
+export function withRateLimit(provider: LlmProvider, config?: Partial<ProviderRateLimitConfig>): LlmProvider {
+	if (config === undefined) return provider;
+	if (Object.keys(config).length === 0) return provider;
+	// Spread means explicit `undefined` values in config override the defaults (e.g.
```
PR-Reviewer-Ant, line 188: The early-return guards check `maxCallsPerHour` and `burstSize` for zero/undefined, but not `waitTimeoutMs`. A programmatic caller passing `{ maxCallsPerHour: 100, burstSize: 5, waitTimeoutMs: undefined }` will create a limiter and reach `bucket.acquire(cfg.waitTimeoutMs)` with `waitMs = undefined`. Inside `acquire`, `undefined <= 0` is `false` (no early return), `Date.now() + undefined = NaN`, and `Date.now() < NaN` is `false`, so the polling loop never runs and `acquire` returns `false` immediately — every call after the burst throws `RateLimitExceededError` with no wait. This is inconsistent with the `maxCallsPerHour: undefined` case, which returns the provider unwrapped. The YAML path always fills all three fields, so this only affects direct programmatic callers, but a test and either a guard or a doc note would prevent silent surprises.

PR #466, opened 2026-04-05
fix: harden transcript capture and summary inputs

NicholaiVogel:
## what changed

This PR hardens transcript capture so Signet keeps a clean conversation-only canonical transcript for memory and continuity, while preserving raw auditable transcript traces separately in daemon logs.

It also improves session-end reliability by falling back to the stored live transcript snapshot when final transcript input is missing or shorter than what was already captured during prompt flow.

On top of that, it updates the Signet system prompt to explicitly remind agents that linked transcript and summary artifacts are inspectable across sessions, and loosens the session summary prompts so they operate on cleaned conversation context instead of bureaucratic tool exhaust.

## why

We were relying too much on session-end transcript capture, which made interruptions and bad transcript paths too easy to lose context on. We were also feeding summaries with noisier transcript surfaces than necessary.

That led to two real problems:

- canonical transcript artifacts could miss conversational context when sessions ended badly
- summary generation had to fight through tool calls, tool outputs, and other non-conversational noise

## impact

- canonical `--transcript.md` artifacts now reflect only user and assistant conversation
- raw transcript traces are retained under `.daemon/logs/transcripts/` for audit and debugging
- session-end is more resilient when final transcript input is missing or truncated
- agents now get a clearer reminder that cross-session transcript and summary artifacts are available
- rust and ts behavior are closer again instead of drifting apart

## compatibility note

This changes Codex canonical transcript behavior. Codex sessions no longer bypass normalization, so canonical `--transcript.md` artifacts now follow the same conversation-only filtering as other harnesses. If a Codex deployment depended on raw JSONL or partially-recognized event shapes appearing in the canonical transcript artifact, it will now receive the cleaned conversation view instead, and tool-only or unrecognized JSONL can normalize to an empty canonical transcript while the raw audit log is still preserved under `.daemon/logs/transcripts/`.

## validation

- `bun test packages/daemon/src/hooks.test.ts packages/daemon/test/hooks.test.ts packages/daemon/src/pipeline/summary-worker.test.ts`
- targeted rust tests for transcript normalization, prompt reminder, live transcript fallback, and raw-audit-vs-canonical behavior
- `bunx @biomejs/biome check packages/daemon/src/hooks.ts packages/daemon/src/transcript-audit.ts packages/daemon/src/pipeline/summary-worker.ts packages/daemon/src/hooks.test.ts packages/daemon/test/hooks.test.ts docs/API.md docs/ARCHITECTURE.md docs/PIPELINE.md docs/specs/approved/system-prompt-extraction.md docs/specs/approved/memory-md-temporal-head.md docs/specs/approved/lossless-working-memory-runtime.md docs/specs/approved/desire-paths-epic.md`

## note

`bun scripts/spec-deps-check.ts` still fails in this repo, but the failure is pre-existing and unrelated to this PR:

- `recall-confidence-gate` path and status drift
- missing `references/Engram` path for `engram-informed-predictor-track`

### DIFF
--- a/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
@@ -266,7 +266,7 @@ fn normalize_project(raw: Option<&str>) -> Option<String> {
     Some(s.replace('\\', "/").trim_end_matches('/').to_string())
 }
 
-fn normalize_session_transcript(harness: &str, raw: &str) -> String {
+fn normalize_session_transcript(raw: &str) -> String {
     let lines = raw
         .lines()
         .map(str::trim)
@@ -288,7 +288,6 @@ fn normalize_session_transcript(harness: &str, raw: &str) -> String {
         }
     }
 
-    let _ = harness;
 
     // Fall back to raw only when the input does not look like JSONL
     // conversation data. If it is JSONL and we extracted zero turns, return
@@ -884,38 +883,77 @@ pub async fn prompt_submit(
         let mut raw_transcript = body.transcript.clone().unwrap_or_default();
         if raw_transcript.is_empty()
             && let Some(path) = body.transcript_path.as_deref()
-            && std::path::Path::new(path).exists()
+            && !path.trim().is_empty()
         {
-            match fs::read_to_string(path) {
+            let canonical = match fs::canonicalize(path) {
+                Ok(value) => value,
+                Err(e) => {
+                    warn!(path, error = %e, "prompt-submit: transcript_path unresolvable");
+                    return (
+                        StatusCode::INTERNAL_SERVER_ERROR,
+                        Json(serde_json::json!({"error": format!("transcript_path unresolvable: {e}")})),
+                    )
+                        .into_response();
+                }
+            };
+            if !transcript_path_allowed(&canonical) {
+                warn!(path = %canonical.display(), "prompt-submit: transcript_path outside allowed roots");
+                return (
+                    StatusCode::FORBIDDEN,
+                    Json(serde_json::json!({"error": "transcript_path outside allowed workspace roots"})),
+                )
+                    .into_response();
+            }
+            match fs::read_to_string(&canonical) {
                 Ok(content) => raw_transcript = content,
-                Err(e) => warn!(path, error = %e, "prompt-submit: transcript read failed"),
+                Err(e) => warn!(path = %canonical.display(), error = %e, "prompt-submit: transcript read failed"),
             }
         }
 
-        let normalized = normalize_session_transcript(harness, &raw_transcript);
+        let normalized = normalize_session_transcript(&raw_transcript);
         if !normalized.trim().is_empty() {
             let harness_value = harness.to_string();
             let project_value = body.project.clone();
             let agent_value = agent_id.clone();
-            let session_key_for_write = session_key_value.clone();
-            let normalized_for_write = normalized.clone();
-            if let Err(e) = state
+            let session_key_for_read = session_key_value.clone();
+            let agent_for_read = agent_id.clone();
+            let normalized_len = normalized.len();
+            let should_write = match state
                 .pool
-                .write(Priority::Low, move |conn| {
-                    upsert_session_transcript(
-                        conn,
-                        &session_key_for_write,
-                        &normalized_for_write,
-                        &harness_value,
-                        project_value.as_deref(),
-                        &agent_value,
-                    )
-                    .map(|_| serde_json::Value::Null)
-                    .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
+                .read(move |conn| {
+                    session_transcript_content(conn, &session_key_for_read, &agent_for_read)
+                        .map(|prev| prev.is_none_or(|value| normalized_len >= value.len()))
+                        .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
                 })
                 .await
             {
-                warn!(error = %e, session_key = %session_key_value, "prompt-submit: transcript upsert failed");
+                Ok(value) => value,
+                Err(e) => {
+                    warn!(error = %e, session_key = %session_key_value, "prompt-submit: transcript length check failed");
+                    false
+                }
+            };
+            if should_write {
+                let session_key_for_write = session_key_value.clone();
+                let normalized_for_write = normalized.clone();
+                if let Err(e) = state
+                    .pool
+                    .write(Priority::Low, move |conn| {
+                        upsert_session_transcript(
+                            conn,
+                            &session_key_for_write,
+                            &normalized_for_write,
+                            &harness_value,
+                            project_value.as_deref(),
+                            &agent_value,
+                        )
+                        .map(|_| serde_json::Value::Null)
+                        .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
+                    })
+                    .await
+                {
+                    warn!(error = %e, session_key = %session_key_value, "prompt-submit: transcript upsert failed");
+                }
             }
         }
 
@@ -1514,7 +1552,7 @@ pub async fn session_end(
 
     // Normalized view used for LLM inputs (summary job) and the legacy DB
     // upsert.  The canonical artifact above gets the raw original.
-    let mut normalized = normalize_session_transcript(harness, &transcript);
+    let mut normalized = normalize_session_transcript(&transcript);
     if !session_key.trim().is_empty() {
         let session_key_value = session_key.clone();
         let agent_value = agent_id.clone();
@@ -2955,9 +2993,103 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn prompt_submit_rejects_transcript_path_outside_allowed_roots() {
+        let (state, writer, tmp) = test_state("hooks-prompt-submit-path-guard");
+        let transcript_path = tmp.path().join("prompt-transcript.txt");
+        std::fs::write(&transcript_path, "User: hi\nAssistant: hello").unwrap();
+
+        let resp = prompt_submit(
+            State(state.clone()),
+            HeaderMap::new(),
+            Json(PromptSubmitBody {
+                harness: Some("test".to_string()),
+                project: Some("packages/daemon-rs".to_string()),
+                agent_id: Some("agent-a".to_string()),
+                user_message: Some("review the release checklist".to_string()),
+                user_prompt: None,
+                last_assistant_message: None,
+                session_key: Some("sess-prompt-denied".to_string()),
+                transcript: None,
+                transcript_path: Some(transcript_path.display().to_string()),
+                runtime_path: None,
+            }),
+        )
+        .await;
+
+        assert_eq!(resp.status(), StatusCode::FORBIDDEN);
+        let body = test_json(resp).await;
+        assert_eq!(
+            body["error"],
+            serde_json::Value::String("transcript_path outside allowed workspace roots".to_string())
+        );
+
+        drop(state);
+        let _ = writer.await;
+    }
+
+    #[tokio::test]
+    async fn prompt_submit_keeps_longer_stored_transcript_snapshot() {
+        let (state, writer, _tmp) = test_state("hooks-prompt-submit-length-guard");
+        state
+            .pool
+            .write(Priority::Low, move |conn| {
+                upsert_session_transcript(
+                    conn,
+                    "sess-prompt-guard",
+                    &"User: longer transcript\nAssistant: still longer\n".repeat(6),
+                    "test",
+                    Some("packages/daemon-rs"),
+                    "agent-a",
+                )?;
+                Ok(serde_json::Value::Null)
+            })
+            .await
+            .unwrap();
+
+        let resp = prompt_submit(
+            State(state.clone()),
+            HeaderMap::new(),
+            Json(PromptSubmitBody {
+                harness: Some("test".to_string()),
+                project: Some("packages/daemon-rs".to_string()),
+                agent_id: Some("agent-a".to_string()),
+                user_message: Some("review the release checklist".to_string()),
+                user_prompt: None,
+                last_assistant_message: None,
+                session_key: Some("sess-prompt-guard".to_string()),
+                transcript: Some("User: short\nAssistant: short".to_string()),
+                transcript_path: None,
+                runtime_path: None,
+            }),
+        )
+        .await;
+
+        assert_eq!(resp.status(), StatusCode::OK);
+        let stored = state
+            .pool
+            .read(|conn| {
+                session_transcript_content(conn, "sess-prompt-guard", "agent-a")
+                    .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
+            })
+            .await
+            .unwrap()
+            .unwrap_or_default();
+        assert!(stored.contains("longer transcript"));
+        assert!(!stored.contains("User: short\nAssistant: short"));
+
+        drop(state);
+        let _ = writer.await;
+    }
+
+    #[tokio::test]
     async fn prompt_submit_writes_transcript_audit_and_live_snapshot() {
         let (state, writer, tmp) = test_state("hooks-prompt-submit-audit");
-        let transcript_path = tmp.path().join("prompt-transcript.txt");
+        let stage_dir = std::path::Path::new("/tmp/signet");
+        std::fs::create_dir_all(stage_dir).unwrap();
+        let transcript_path = stage_dir.join(format!(
+            "prompt-transcript-{}.txt",
+            chrono::Utc::now().timestamp_nanos_opt().unwrap_or_default()
+        ));
         std::fs::write(
             &transcript_path,
             "User: review the release checklist
@@ -3009,6 +3141,7 @@ Assistant: here's the checklist",
                 .and_then(|name| name.to_str())
                 .is_some_and(|name| name.ends_with("--latest.log"))
         }));
+        let _ = std::fs::remove_file(&transcript_path);
 
         drop(state);
         let _ = writer.await;
@@ -3317,7 +3450,7 @@ Assistant: here's the checklist",
         ]
         .join("\n");
         assert_eq!(
-            normalize_session_transcript("codex", &raw),
+            normalize_session_transcript(&raw),
             "User: Hello\nAssistant: Hi"
         );
     }
@@ -3329,7 +3462,7 @@ Assistant: here's the checklist",
             r#"{"type":"response_item","payload":{"type":"function_call_output","output":"README.md"}}"#,
         ]
         .join("\n");
-        assert_eq!(normalize_session_transcript("opencode", &raw), "");
+        assert_eq!(normalize_session_transcript(&raw), "");
     }
 
     #[test]
--- a/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
@@ -288,7 +288,6 @@ fn normalize_session_transcript(raw: &str) -> String {
         }
     }
 
-
     // Fall back to raw only when the input does not look like JSONL
     // conversation data. If it is JSONL and we extracted zero turns, return
     // the empty normalized view so tool-only logs do not pollute summaries.
@@ -303,11 +302,7 @@ fn normalize_json_transcript_line(value: &serde_json::Value) -> Option<String> {
     if value.get("type").and_then(serde_json::Value::as_str) == Some("item.completed") {
         let item = value.get("item")?;
         if item.get("type").and_then(serde_json::Value::as_str) == Some("agent_message") {
-            let text = item
-                .get("text")
-                .or_else(|| item.get("message"))
-                .or_else(|| item.get("content"))
-                .and_then(serde_json::Value::as_str)?;
+            let text = extract_json_message_text(item)?;
             return Some(format!("Assistant: {text}"));
         }
     }
@@ -456,8 +451,8 @@ fn resolve_audit_token(
     } else {
         let mut digest = Sha256::new();
         digest.update(raw.as_bytes());
-        let bytes = digest.finalize();
-        bytes[..16]
+        digest
+            .finalize()
             .iter()
             .map(|b| format!("{b:02x}"))
             .collect::<String>()
@@ -906,7 +901,9 @@ pub async fn prompt_submit(
             }
             match fs::read_to_string(&canonical) {
                 Ok(content) => raw_transcript = content,
-                Err(e) => warn!(path = %canonical.display(), error = %e, "prompt-submit: transcript read failed"),
+                Err(e) => {
+                    warn!(path = %canonical.display(), error = %e, "prompt-submit: transcript read failed")
+                }
             }
         }
 
@@ -915,45 +912,32 @@ pub async fn prompt_submit(
             let harness_value = harness.to_string();
             let project_value = body.project.clone();
             let agent_value = agent_id.clone();
-            let session_key_for_read = session_key_value.clone();
-            let agent_for_read = agent_id.clone();
-            let normalized_len = normalized.len();
-            let should_write = match state
+            let session_key_for_write = session_key_value.clone();
+            let normalized_for_write = normalized.clone();
+            let normalized_len = normalized_for_write.len();
+            if let Err(e) = state
                 .pool
-                .read(move |conn| {
-                    session_transcript_content(conn, &session_key_for_read, &agent_for_read)
-                        .map(|prev| prev.is_none_or(|value| normalized_len >= value.len()))
-                        .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
+                .write(Priority::Low, move |conn| {
+                    let should_write =
+                        session_transcript_content(conn, &session_key_for_write, &agent_value)?
+                            .is_none_or(|value| normalized_len >= value.len());
+                    if !should_write {
+                        return Ok(serde_json::Value::Bool(false));
+                    }
+                    upsert_session_transcript(
+                        conn,
+                        &session_key_for_write,
+                        &normalized_for_write,
+                        &harness_value,
+                        project_value.as_deref(),
+                        &agent_value,
+                    )
+                    .map(|_| serde_json::Value::Bool(true))
+                    .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
                 })
                 .await
             {
-                Ok(value) => value,
-                Err(e) => {
-                    warn!(error = %e, session_key = %session_key_value, "prompt-submit: transcript length check failed");
-                    false
-                }
-            };
-            if should_write {
-                let session_key_for_write = session_key_value.clone();
-                let normalized_for_write = normalized.clone();
-                if let Err(e) = state
-                    .pool
-                    .write(Priority::Low, move |conn| {
-                        upsert_session_transcript(
-                            conn,
-                            &session_key_for_write,
-                            &normalized_for_write,
-                            &harness_value,
-                            project_value.as_deref(),
-                            &agent_value,
-                        )
-                        .map(|_| serde_json::Value::Null)
-                        .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
-                    })
-                    .await
-                {
-                    warn!(error = %e, session_key = %session_key_value, "prompt-submit: transcript upsert failed");
-                }
+                warn!(error = %e, session_key = %session_key_value, "prompt-submit: transcript upsert failed");
             }
         }
 
@@ -2752,6 +2736,7 @@ mod tests {
     use axum::extract::State;
     use axum::http::{HeaderMap, StatusCode};
     use serde_json::Value;
+    use sha2::{Digest, Sha256};
     use signet_core::config::{
         AgentManifest, DaemonConfig, MemoryManifestConfig, PipelineV2Config,
     };
@@ -2766,9 +2751,9 @@ mod tests {
         CHECKPOINT_MIN_DELTA, CompactionCompleteBody, PromptSubmitBody, SessionEndBody,
         build_signet_system_prompt, compaction_complete, extract_delta,
         normalize_session_transcript, parse_visibility, prompt_submit,
-        require_session_scope_for_write, resolve_compaction_project, resolve_remember_agent,
-        session_agent_id, session_end, session_transcript_content, strip_untrusted_metadata,
-        upsert_session_transcript,
+        require_session_scope_for_write, resolve_audit_token, resolve_compaction_project,
+        resolve_remember_agent, session_agent_id, session_end, session_transcript_content,
+        strip_untrusted_metadata, upsert_session_transcript,
     };
     use signet_services::session::{RuntimePath, SessionTracker};
 
@@ -3020,7 +3005,9 @@ mod tests {
         let body = test_json(resp).await;
         assert_eq!(
             body["error"],
-            serde_json::Value::String("transcript_path outside allowed workspace roots".to_string())
+            serde_json::Value::String(
+                "transcript_path outside allowed workspace roots".to_string()
+            )
         );
 
         drop(state);
@@ -3079,6 +3066,48 @@ mod tests {
 
         drop(state);
         let _ = writer.await;
+    }
+
+    #[test]
+    fn normalize_session_transcript_normalizes_item_completed_text() {
+        let raw = [
+            r#"{"type":"item.completed","item":{"type":"agent_message","text":"line one\r\nline two"}}"#,
+            r#"{"type":"event_msg","payload":{"type":"user_message","message":"done"}}"#,
+        ]
+        .join("\n");
+
+        let normalized = normalize_session_transcript(&raw);
+
+        assert!(normalized.contains("Assistant: line one  line two"));
+        assert!(normalized.contains("User: done"));
+    }
+
+    #[test]
+    fn resolve_audit_token_uses_full_raw_hash_when_session_identity_missing() {
+        let raw = "User: audit me\nAssistant: on it";
+        let raw_hash = {
+            let mut digest = Sha256::new();
+            digest.update(raw.as_bytes());
+            digest
+                .finalize()
+                .iter()
+                .map(|b| format!("{b:02x}"))
+                .collect::<String>()
+        };
+        let expected = {
+            let mut digest = Sha256::new();
+            digest.update(b"agent-a");
+            digest.update(b":");
+            digest.update(raw_hash.as_bytes());
+            digest
+                .finalize()
+                .iter()
+                .take(8)
+                .map(|b| format!("{b:02x}"))
+                .collect::<String>()
+        };
+
+        assert_eq!(resolve_audit_token("agent-a", "", None, raw), expected);
     }
 
     #[tokio::test]
--- a/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
+++ b/packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
@@ -429,9 +429,9 @@ fn audit_fs_timestamp(iso: &str) -> String {
 
 fn is_safe_audit_name(value: &str) -> bool {
     !value.is_empty()
-        && value.bytes().all(|byte| {
-            byte.is_ascii_alphanumeric() || matches!(byte, b'-' | b'_' | b'.')
-        })
+        && value
+            .bytes()
+            .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'-' | b'_' | b'.'))
 }
 
 fn build_audit_path(root: &Path, file_name: &str) -> std::io::Result<PathBuf> {
@@ -494,10 +494,10 @@ fn write_transcript_audit(
         let final_path = build_audit_path(
             root,
             &format!(
-            "{}--{}--raw-transcript.log",
-            audit_fs_timestamp(captured_at),
-            token
-        ),
+                "{}--{}--raw-transcript.log",
+                audit_fs_timestamp(captured_at),
+                token
+            ),
         )?;
         fs::write(final_path, raw_transcript)?;
     }
@@ -571,7 +571,10 @@ fn enqueue_summary_job(
         rusqlite::params![agent_id, session_id, trigger],
         |row| Ok((row.get::<_, String>(0)?, row.get::<_, String>(1)?)),
     ) {
-        if existing_status == "pending" || existing_status == "leased" || existing_status == "processing" {
+        if existing_status == "pending"
+            || existing_status == "leased"
+            || existing_status == "processing"
+        {
             // Update the transcript so retries with fresher content are used.
             let _ = conn.execute(
                 "UPDATE summary_jobs SET transcript = ?1, updated_at = ?2 WHERE id = ?3",
@@ -644,7 +647,8 @@ pub async fn session_start(
 
     // Session claim
     if let Some(p) = path
-        && let ClaimResult::Conflict { claimed_by } = state.sessions.claim(&session_key, p, &agent_id)
+        && let ClaimResult::Conflict { claimed_by } =
+            state.sessions.claim(&session_key, p, &agent_id)
     {
         return conflict_response(claimed_by);
     }
@@ -748,6 +752,8 @@ pub struct PromptSubmitBody {
     #[allow(dead_code)] // Will be used for context-aware search in Phase 5
     pub last_assistant_message: Option<String>,
     pub session_key: Option<String>,
+    pub transcript: Option<String>,
+    pub transcript_path: Option<String>,
     pub runtime_path: Option<String>,
 }
 
@@ -859,6 +865,10 @@ pub async fn prompt_submit(
         .collect();
     let query_terms = terms.join(" ");
     let metadata_header = format_metadata_header();
+    let agent_id = body
+        .agent_id
+        .clone()
+        .unwrap_or_else(|| "default".to_string());
 
     // Record in continuity tracker
     if let Some(key) = &body.session_key {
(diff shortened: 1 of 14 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix: harden transcript capture and summary inputs` (commit `86ef6d16`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `86ef6d16`

```pr-review-json
{
  "summary": "[Automated Bot Review — pr-reviewer] PR #466 hardens transcript capture and summary prompts. One blocking correctness issue found: a Rust/TS parity gap in audit-write ordering causes raw transcripts to be silently lost in the Rust path for tool-only JSONL sessions. Two additional warnings around code duplication and a hot-path deep copy.",
  "verdict": "request_changes",
  "confidence": {
    "level": "high",
    "justification": "The ordering of write_transcript_audit relative to the empty-normalized gate is directly visible in the diff. Rust places the call after the gate (confirmed: 'Gate before any writes' block returns early, then write_transcript_audit follows). TypeScript places the call before the gate ('if (rawTranscript) { writeTranscriptAudit(...)' precedes the artifact-write section). The behavior divergence for tool-only JSONL with no stored live transcript is deterministic from reading the two diffs together."
  },
  "comments": [
    {
      "file": "packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs",
      "line": 1499,
      "body": "**Blocking: `write_transcript_audit` is placed after the empty-normalized gate, creating a Rust/TS parity gap and contradicting the PR's stated audit guarantee.**\n\nFor a session whose final transcript is valid JSONL containing only tool calls (no user/assistant turns), `normalized` normalizes to `\"\"` (intentional per the new logic). If there is no stored live transcript in the DB, the `normalized.trim().is_empty()` gate fires and returns early. `write_transcript_audit` is never reached, so the raw JSONL is silently discarded — no `--latest.log`, no timestamped `--raw-transcript.log`.\n\nThe PR description states: *"raw transcript traces are retained under `.daemon/logs/transcripts/` for audit and debugging."* The ARCHITECTURE.md and PIPELINE.md updates echo this. But for tool-only JSONL sessions without a stored live snapshot, neither the cleaned artifact nor the raw audit file is written. The data is lost with no log entry.\n\nThe TypeScript `handleSessionEnd` path does not have this problem: `writeTranscriptAudit` fires when `rawTranscript` is truthy, before any gate, so the raw input is always preserved. This was confirmed from the diff ordering in `hooks.ts`.\n\n**Fix**: Move `write_transcript_audit` to before the empty-normalized gate, using the raw `transcript` variable (not `normalized`). Mirror the TS ordering.",
      "severity": "blocking"
    },
    {
      "file": "packages/daemon/src/transcript-audit.ts",
      "line": 8,
      "body": "**Warning: `getAgentsDir()` is duplicated here from `hooks.ts`.**\n\nThe function body (`process.env.SIGNET_PATH || join(homedir(), '.agents')`) is identical to the one defined in `hooks.ts`. The repo AGENTS.md explicitly calls out \"Duplication / parity drift\" as a common PR failure mode: *\"Do not duplicate constants, maps, dependency types, descriptions, or config defaults across files. Extract shared sources of truth.\"*\n\n`getAgentsDir` (or the equivalent from `@signet/core`) should be imported rather than redefined. If the env-var key or default path ever changes, this copy will silently diverge.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs",
      "line": 340,
      "body": "**Warning: `message.clone()` performs a full deep copy of a JSON object in the transcript normalization hot path.**\n\n```rust\nlet text = extract_json_message_text(&serde_json::Value::Object(message.clone()));\n```\n\n`extract_json_message_text` only reads from the object — it does not need ownership. Passing a reference avoids the allocation. The function signature accepts `&serde_json::Value`, so this can be called as `extract_json_message_text(&serde_json::Value::Object(message.clone()))` → or restructured to avoid the clone entirely by passing `message` as a `&serde_json::Map` directly.\n\nFor large assistant messages (e.g. long code blocks), this copies the full content map on every matched line during normalization.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs",
      "line": 420,
      "body": "**Nitpick: 64-bit audit token can silently overwrite `--latest.log` on hash collision.**\n\n`resolve_audit_token` truncates to the first 8 bytes (16 hex chars). For local single-user use, collision probability is negligible (birthday bound ~50% at 2^32 distinct sessions). But a collision would silently overwrite a different session's `--latest.log` file without any warning. The same applies to the TypeScript counterpart (`.slice(0, 16)`). Consider 12 bytes (24 hex chars) for a more comfortable margin, or log a warning if the token file already exists with different content.",
      "severity": "nitpick"
    }
  ]
}
```

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1356,12 +1527,27 @@ pub async fn session_end(
         Some(harness),
```
PR-Reviewer-Ant, line 1646: **Blocking (prior concern still present in commit 22abece9): `write_transcript_audit` is placed after the empty-normalized gate, silently discarding raw transcripts for tool-only JSONL sessions.**

For a session whose transcript is valid JSONL but contains only tool calls (no user/assistant turns), `normalize_session_transcript` returns `""` (the new intentional behavior). If no stored live transcript exists in the DB, the gate at line ~1493 (`if normalized.trim().is_empty()`) fires and returns early. `write_transcript_audit` is never reached: neither `--latest.log` nor `--raw-transcript.log` is written, and no warning is emitted. The raw JSONL is silently lost.

The PR description and updated docs both state: *"raw auditable traces may still be written to daemon logs outside the memory lineage."* This guarantee is violated for tool-only JSONL sessions without a stored snapshot.

The TS `handleSessionEnd` path does not have this problem — `writeTranscriptAudit` is called when `rawTranscript` is truthy, unconditionally before the gate, so the raw input is always preserved.

**Fix**: Move `write_transcript_audit` to before the empty-normalized gate, passing `&transcript` (the original raw input). This mirrors the TS ordering and makes the audit guarantee unconditional for any non-empty raw input.

packages/daemon/src/hooks.ts
```diff
@@ -2595,18 +2621,47 @@ export function handleSessionEnd(req: SessionEndRequest): SessionEndResponse {
...
+			rawTranscript = readFileSync(req.transcriptPath, "utf-8");
 			transcript = normalizeSessionTranscript(req.harness, rawTranscript);
 		} catch {
 			logger.warn("hooks", "Could not read transcript", {
 				path: req.transcriptPath,
 			});
 		}
 	} else if (req.transcript) {
-		transcript = normalizeSessionTranscript(req.harness, req.transcript);
+		rawTranscript = req.transcript;
```
PR-Reviewer-Ant, line 2637: **Blocking (new): `getSessionTranscriptContent` in `handleSessionEnd` is not wrapped in try-catch, creating a new unhandled exception path that can fail session-end entirely.**

```typescript
const storedTranscript = sessionKey ? (getSessionTranscriptContent(sessionKey, agentId) ?? "") : "";
```

This call is new in this PR. If `getSessionTranscriptContent` throws (e.g., SQLite read error, locked DB), the exception propagates up and the entire session-end operation fails — memories are not extracted, artifacts are not written, the summary job is not queued.

The equivalent Rust path handles failures gracefully with `if let Ok(Some(stored)) = ...`, silently falling back to the original normalized transcript on any error.

In `handleUserPromptSubmit` (same PR), the analogous `getSessionTranscriptContent` call IS correctly wrapped in the existing try-catch block. The session-end path was overlooked.

**Fix**: Wrap the stored-transcript lookup in try-catch and default to `""` on error, consistent with the Rust behavior and the `handleUserPromptSubmit` pattern:
```typescript
let storedTranscript = "";
if (sessionKey) {
    try {
        storedTranscript = getSessionTranscriptContent(sessionKey, agentId) ?? "";
    } catch {
        logger.warn("hooks", "Failed to read stored transcript for fallback", { sessionKey });
    }
}
```

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,53 @@
+import { createHash } from "node:crypto";
+import { existsSync, mkdirSync, writeFileSync } from "node:fs";
+import { homedir } from "node:os";
+import { join } from "node:path";
+
+function getAgentsDir(): string {
+	return process.env.SIGNET_PATH || join(homedir(), ".agents");
+}
```
PR-Reviewer-Ant, line 8: **Warning (prior concern still present): `getAgentsDir()` is duplicated from `hooks.ts`.**

The body `process.env.SIGNET_PATH || join(homedir(), '.agents')` is identical to the definition in `hooks.ts`. AGENTS.md explicitly lists 'Duplication / parity drift' as a common PR failure mode. If the env-var key or default path changes, this copy will silently diverge.

Import and re-export from `@signet/core` or the shared constants module instead.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -315,31 +316,179 @@ fn normalize_json_transcript_line(value: &serde_json::Value) -> Option<String> {
...
+        }
+    }
+
+    let role = extract_json_string(value, &["role", "speaker"]);
+    let text = extract_json_message_text(value);
+    match (role.as_deref(), text) {
         (Some("user"), Some(text)) => Some(format!("User: {text}")),
         (Some("assistant"), Some(text)) => Some(format!("Assistant: {text}")),
         _ => None,
     }
```
PR-Reviewer-Ant, line 334: **Warning (prior concern still present): `message.clone()` performs a full deep copy of a JSON object in the transcript normalization hot path.**

```rust
let text = extract_json_message_text(&serde_json::Value::Object(message.clone()));
```

`extract_json_message_text` only reads the object and accepts `&serde_json::Value`. The clone can be avoided by restructuring `extract_json_message_text` to also accept `&serde_json::Map<String, serde_json::Value>` directly, or by calling it as `extract_json_message_text(value)` after matching the nested `message` field at the call site. For large assistant messages (long code blocks), this copies the full content map on every matched JSONL line during normalization.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `22abece9`

[Automated Bot Review — pr-reviewer] PR #466 introduces transcript cleaning, live-transcript fallback, and raw audit logging. One prior blocking concern about Rust audit write ordering remains unaddressed in the current commit. One new blocking issue found in the TS session-end path. Two prior warnings also remain.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The Rust gate/audit ordering is directly readable from the diff: the empty-normalized gate at line ~1493 returns early, and write_transcript_audit at line ~1527 appears after it — confirmed by the comment 'Gate before any writes' immediately preceding the gate block. The TS getSessionTranscriptContent exception path is visible at the new lines in handleSessionEnd, and the contrast with the try-catch in handleUserPromptSubmit (same PR) is explicit in the diff.

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,49 @@
...
+import { join } from "node:path";
+import { resolveDefaultBasePath } from "@signet/core";
+
+function getTranscriptAuditDir(): string {
+	return join(resolveDefaultBasePath(), ".daemon", "logs", "transcripts");
+}
+
+function fsTimestamp(iso: string): string {
+	return iso.replace(/:/g, "-");
+}
```
PR-Reviewer-Ant, line 23: **Path traversal vulnerability — `fsTimestamp` is not a sanitizer.**

`fsTimestamp` only replaces `:` with `-` and passes everything else through, including `../` sequences:

```ts
function fsTimestamp(iso: string): string {
    return iso.replace(/:/g, "-");
}
```

`params.capturedAt` originates from `req.endedAt` in `handleSessionEnd`, which is caller-supplied. A crafted value like `../../../../tmp/evil` survives `fsTimestamp` unchanged. `path.join` in Node.js resolves `..` segments, so:

```
join("/home/user/.agents/.daemon/logs/transcripts",
     "../../../../tmp/evil--abc123--raw-transcript.log")
// → /home/user/tmp/evil--abc123--raw-transcript.log   (or further out)
```

Because `writeFileSync` is then called with the attacker-controlled path and the raw (also attacker-controlled) transcript as content, this is a file-write primitive outside the intended audit directory.

The Rust counterpart guards against exactly this: `audit_fs_timestamp` replaces every non-`[A-Za-z0-9._-]` character, and then `build_audit_path` calls `is_safe_audit_name` as a second gate that rejects the write entirely if the resulting filename still contains unsafe bytes. The TypeScript module has neither guard.

**Fix**: apply the same pattern — either validate the resolved path stays under the audit dir (e.g., check `finalPath.startsWith(dir)`), or sanitize `capturedAt` by replacing every character that is not `[A-Za-z0-9._-]` before using it in a filename, matching the Rust `audit_fs_timestamp` behavior.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -263,10 +267,6 @@ fn normalize_project(raw: Option<&str>) -> Option<String> {
 }
 
 fn normalize_session_transcript(harness: &str, raw: &str) -> String {
-    if harness.trim().eq_ignore_ascii_case("codex") {
-        return raw.to_string();
-    }
-
     let lines = raw
         .lines()
```
PR-Reviewer-Ant, line 271: **Undocumented breaking change for Codex harness users.**

The Codex-specific passthrough (`if harness.trim().eq_ignore_ascii_case("codex") { return raw.to_string(); }`) is removed unconditionally. Any Codex deployment that depended on receiving the full raw transcript in the canonical `--transcript.md` artifact (because its JSONL format was partially unrecognized) will silently receive a conversation-only filtered or even empty transcript after this change.

The PR description mentions the Codex bypass is removed as part of normalization hardening, and the new test `normalize_session_transcript_strips_codex_tool_turns` verifies the new behavior, but the description does not flag this as a breaking change for existing Codex deployments. Operators running Codex sessions whose JSONL contains message structures beyond `event_msg/user_message` or `item.completed/agent_message` will get empty canonical transcripts with no warning.

At minimum, this should be called out in the PR description as a breaking change for the Codex connector.

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	const dir = getTranscriptAuditDir();
+	if (!existsSync(dir)) {
+		mkdirSync(dir, { recursive: true });
+	}
+
+	const token = resolveAuditToken(params.agentId, params.sessionId, params.sessionKey, params.rawTranscript);
+	const latestPath = join(dir, `${token}--latest.log`);
+	writeFileSync(latestPath, params.rawTranscript, "utf8");
+
```
PR-Reviewer-Ant, line 53: **Parity drift: TS `fsTimestamp` vs Rust `audit_fs_timestamp`.**

TS replaces only `:` → `-`. Rust replaces every character that is not `[A-Za-z0-9._-]` → `-`. In practice this means an ISO timestamp with a `T` separator (`2026-04-05T10:30:00.000Z`) produces `2026-04-05T10-30-00.000Z` in TS (the `T` and `Z` survive) and `2026-04-05T10-30-00.000Z` in Rust (same result for this case, since `T` and `Z` are alphanumeric). The outputs are identical for well-formed ISO strings, but for unusual or malformed `endedAt` values the two codepaths diverge. The security gap (noted above) is the more urgent consequence of this parity drift.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7b67d333`

[Automated Review — pr-reviewer bot] Found one blocking security issue in the new TypeScript `transcript-audit.ts` module (path traversal), one warning around a Codex harness behavioral change that constitutes an undocumented breaking change, and one nitpick on TS/Rust parity drift in filename sanitization.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The path traversal is directly provable from the `fsTimestamp` implementation in the new `packages/daemon/src/transcript-audit.ts` file (line 12) and the `join(dir, ...)` call at line 41. The Rust counterpart (`audit_fs_timestamp` + `is_safe_audit_name` + `build_audit_path`) provides the exact guard that is absent from the TypeScript version. No runtime trace or external artifact is needed — the divergence is visible in the same diff.

packages/daemon/src/hooks.ts
```diff
@@ -2241,29 +2246,50 @@ export async function handleUserPromptSubmit(
...
 				deps.logger.warn("hooks", "Prompt transcript write failed", {
 					error: error instanceof Error ? error.message : String(error),
 				});
 			}
 		}
+
+		if (rawTranscript) {
+			try {
+				writeTranscriptAudit({
+					agentId,
```
PR-Reviewer-Ant, line 2282: The new `writeTranscriptAudit` call is added to `handleUserPromptSubmit` in the TS daemon but there is no corresponding change in `packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs` for the prompt-submit path. AGENTS.md states: "JS daemon changes must preserve parity expectations." The Rust shadow runtime therefore does not produce mid-session audit snapshots during prompt-submit, creating a divergence. If this gap is intentional (e.g., Rust shadow runtime intentionally lags on prompt-submit audit), it should be documented explicitly.

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,60 @@
+import { createHash } from "node:crypto";
+import { existsSync, mkdirSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { resolveDefaultBasePath } from "@signet/core";
+
+function getTranscriptAuditDir(): string {
+	return join(resolveDefaultBasePath(), ".daemon", "logs", "transcripts");
+}
```
PR-Reviewer-Ant, line 8: `getTranscriptAuditDir()` derives the audit directory from `resolveDefaultBasePath()` (process-wide default via `SIGNET_PATH`) rather than accepting a caller-provided base path. The Rust counterpart (`write_transcript_audit`) correctly takes `root: &Path` as an explicit parameter, making it usable for any configured workspace. The TS version ties audit writes to the global default, which works today because the daemon sets `SIGNET_PATH`, but breaks encapsulation and makes the function harder to test or reuse in non-default-workspace contexts. A `basePath` parameter would align it with the Rust API surface.

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function fsTimestamp(iso: string): string {
+	return Array.from(iso, (char) => (/^[A-Za-z0-9._-]$/.test(char) ? char : "-")).join("");
+}
+
+function isSafeAuditName(value: string): boolean {
+	return value.length > 0 && /^[A-Za-z0-9._-]+$/.test(value);
+}
```
PR-Reviewer-Ant, line 16: `isSafeAuditName` permits `.` characters (regex `[A-Za-z0-9._-]`), which means a sanitized timestamp like `..--..--tmp-evil` passes validation. In practice `path.join` does not resolve `..` embedded in a filename component (only when `..` is a complete component separated by `/`), so there is no actual path traversal. The test at line 1808 in `test/hooks.test.ts` confirms this. Worth noting for future readers: if a filesystem separator ever appears in a sanitized timestamp (currently prevented by `fsTimestamp` turning `/` to `-`), the `.` allowance combined with double-dots would require re-evaluation. No change needed unless stricter defense-in-depth is desired.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8699622d`

[Automated Review — pr-reviewer bot] PR #466 hardens transcript capture and session-end fallback. Changes are broadly correct and well-tested. Two issues worth discussing: a Rust/TS parity drift on prompt-submit audit writes, and the TS audit helper using a hardcoded base path rather than receiving it as a parameter.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The parity gap between `handleUserPromptSubmit` (TS adds `writeTranscriptAudit`, Rust does not) is directly visible in the diff. The hardcoded path issue is in `transcript-audit.ts:10` vs the Rust `write_transcript_audit` which takes `root: &Path`. No missing runtime trace or cross-module context is needed to evaluate these — both findings are directly provable from the changed lines.


- packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs:300 - The `let _ = harness;` line suppresses an unused-variable warning after the Codex bypass was removed. The parameter can instead be removed from the function signature entirely, or if it is retained for future use, a `#[allow(unused_variables)]` attribute with a comment is clearer than the `let _ = ...` idiom. Minor, but AGENTS.md flags dead vars/imports as a lint hygiene issue.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +880,59 @@ pub async fn prompt_submit(
         state.continuity.record_prompt(key, &query_terms, snippet);
     }
 
+    if let Some(session_key_value) = body.session_key.clone() {
```
PR-Reviewer-Ant, line 877: **Path traversal / arbitrary file read in `prompt_submit`.**

The new `transcript_path` field is read directly with `fs::read_to_string(path)` without any workspace root validation. `session_end` in this same file validates the path against `allowed_workspace_roots()` and returns `403` with `"transcript_path outside allowed workspace roots"` when it escapes. That guard is entirely absent here.

In team/hybrid auth mode, any caller with `recall` permission can POST `transcript_path: "/etc/passwd"` (or any readable file) to `user-prompt-submit`. Because `normalize_session_transcript` falls back to the raw string when fewer than 60% of lines parse as JSONL (which `/etc/passwd` won't), `normalized` gets the raw file content. That content is then:
1. Upserted into `session_transcripts` (agent-scoped, retrievable via `/api/memory/recall` with `expand: true`).
2. Written to the audit log under `.daemon/logs/transcripts/`.

Fix: apply the same `allowed_workspace_roots()` check used in `session_end`, or resolve the path to a canonical form and reject anything outside the workspace before the `fs::read_to_string` call.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +880,59 @@ pub async fn prompt_submit(
...
+            && let Some(path) = body.transcript_path.as_deref()
+            && std::path::Path::new(path).exists()
+        {
+            match fs::read_to_string(path) {
+                Ok(content) => raw_transcript = content,
+                Err(e) => warn!(path, error = %e, "prompt-submit: transcript read failed"),
+            }
+        }
+
+        let normalized = normalize_session_transcript(harness, &raw_transcript);
```
PR-Reviewer-Ant: **Parity divergence with TS: unconditional upsert vs. length-guarded upsert.**

The TS `handleUserPromptSubmit` calls `getSessionTranscriptContent` before writing and skips the upsert when `transcript.length < prev.length`, protecting longer snapshots captured earlier in the session from being clobbered by a shorter one from a later prompt. The Rust path upserts unconditionally whenever `normalized` is non-empty. Under concurrent or back-to-back prompt-submits, a shorter mid-session transcript can overwrite a longer earlier snapshot. AGENTS.md explicitly requires JS/Rust parity on the daemon shadow runtime.

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,61 @@
...
+
+function buildAuditPath(dir: string, fileName: string): string {
+	if (!isSafeAuditName(fileName)) {
+		throw new Error("invalid transcript audit file name");
+	}
+	return join(dir, fileName);
+}
+
+function resolveAuditToken(agentId: string, sessionId: string, sessionKey: string | null, raw: string): string {
+	const scoped = sessionId.trim() || sessionKey?.trim() || createHash("sha256").update(raw, "utf8").digest("hex");
```
PR-Reviewer-Ant, line 26: **Audit token divergence between TS and Rust for sessions without explicit IDs.**

`resolveAuditToken` (TS) uses the full 64-hex-char SHA-256 digest of `raw` as the `scoped` fallback:
```ts
const scoped = sessionId.trim() || sessionKey?.trim()
  || createHash("sha256").update(raw, "utf8").digest("hex"); // 64 chars
```
The Rust equivalent takes only the first 16 bytes of the SHA-256 (32 hex chars). Both then hash `{agentId}:{scoped}` and slice to 16 chars, so the final token lengths match but the *values* differ for the same raw transcript. Sessions captured by the Rust daemon without an explicit `session_id` or `session_key` will land in different audit files than if captured by the TS daemon, breaking cross-implementation log correlation.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -288,11 +288,12 @@ fn normalize_session_transcript(harness: &str, raw: &str) -> String {
...
-    // Fall back to raw if fewer than 60% of lines parsed as JSON, OR if we
-    // parsed enough JSON but extracted zero messages (valid-but-unrecognized
-    // JSONL format).  An empty normalized string would silently discard all
-    // input content; raw is always preferable to silence.
-    if parsed * 10 < lines.len() * 6 || normalized.is_empty() {
+    let _ = harness;
+
+    // Fall back to raw only when the input does not look like JSONL
+    // conversation data. If it is JSONL and we extracted zero turns, return
+    // the empty normalized view so tool-only logs do not pollute summaries.
```
PR-Reviewer-Ant, line 293: `normalize_session_transcript` still declares `harness: &str` in its signature but immediately suppresses it with `let _ = harness;`. The parameter is now dead. Remove it from the signature (and all call sites) to avoid misleading readers who might expect the harness to influence normalization.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a1b2d911`

[Automated Review – pr-reviewer bot] PR #466 hardens transcript capture (conversation-only canonical, raw audit logs), adds prompt-submit live snapshot, and removes the Codex bypass from normalization. One blocking security gap found: `transcript_path` in the new Rust `prompt_submit` handler reads arbitrary paths without workspace root validation. Two parity divergences and one dead-parameter nitpick are also flagged.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The path-traversal gap is directly visible in the `prompt_submit` diff at the `fs::read_to_string(path)` call: the code contains no `allowed_workspace_roots()` check, while `session_end` in the same file clearly does (the 403 branch with `"transcript_path outside allowed workspace roots"` is visible in the diff context). The TS/Rust upsert parity gap is visible by comparing the TS length-guard block (`if (!prev || transcript.length >= prev.length)`) against the unconditional Rust upsert. The audit-token divergence is provable from `resolveAuditToken` in `transcript-audit.ts` (full SHA-256 hex = 64 chars) versus the Rust equivalent in `hooks.rs` (first 16 bytes of SHA-256 = 32 hex chars).

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +880,59 @@ pub async fn prompt_submit(
         state.continuity.record_prompt(key, &query_terms, snippet);
     }
 
+    if let Some(session_key_value) = body.session_key.clone() {
+        let mut raw_transcript = body.transcript.clone().unwrap_or_default();
+        if raw_transcript.is_empty()
+            && let Some(path) = body.transcript_path.as_deref()
+            && std::path::Path::new(path).exists()
+        {
+            match fs::read_to_string(path) {
```
PR-Reviewer-Ant: **Security: prompt_submit reads transcript_path with no workspace root validation.**

The new transcript-capture block calls `fs::read_to_string(path)` on a caller-supplied path:

```rust
&& std::path::Path::new(path).exists() {
    match fs::read_to_string(path) {
        Ok(content) => raw_transcript = content,
```

`session_end` in the same file validates the path against `allowed_workspace_roots()` and returns `403 Forbidden` for out-of-bounds paths (`"transcript_path outside allowed workspace roots"`). `prompt_submit` has no equivalent check.

A caller in local mode (no auth required), or any authenticated agent in team/hybrid mode, can supply `transcript_path: "/etc/passwd"` or any readable file. The raw content is then normalized and stored in `session_transcripts` and written to the audit log. Apply the same workspace-root validation used in `session_end` before calling `fs::read_to_string`. Flagged by the previous reviewer in the latest round (comment 3038365628); not addressed in the current head.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +880,59 @@ pub async fn prompt_submit(
...
+        if !normalized.trim().is_empty() {
+            let harness_value = harness.to_string();
+            let project_value = body.project.clone();
+            let agent_value = agent_id.clone();
+            let session_key_for_write = session_key_value.clone();
+            let normalized_for_write = normalized.clone();
+            if let Err(e) = state
+                .pool
+                .write(Priority::Low, move |conn| {
+                    upsert_session_transcript(
```
PR-Reviewer-Ant: **Rust-TS parity gap: unconditional upsert vs TS length guard.**

Rust unconditionally upserts any non-empty normalized transcript into `session_transcripts`. The TS counterpart in `handleUserPromptSubmit` only upserts when the new transcript is at least as long as what is already stored:

```typescript
// TS — length guard
const prev = getSessionTranscriptContent(req.sessionKey, agentId);
if (!prev || transcript.length >= prev.length) {
    deps.upsertSessionTranscript(...);
}
```

A repeated prompt-submit with a shorter transcript (truncated file, partial read) will overwrite a longer stored snapshot in the Rust path but not in the TS path. Because the session-end fallback picks the stored transcript only when `stored.len() > normalized.len()`, clobbering a longer stored snapshot with a shorter one earlier in the session can cause the fallback to silently skip the better version at session-end. Apply the same `>= prev.length` guard in Rust. Flagged in the latest review round (comment 3038365631); not addressed in the current head.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -288,11 +288,12 @@ fn normalize_session_transcript(harness: &str, raw: &str) -> String {
         }
     }
 
-    // Fall back to raw if fewer than 60% of lines parsed as JSON, OR if we
-    // parsed enough JSON but extracted zero messages (valid-but-unrecognized
-    // JSONL format).  An empty normalized string would silently discard all
-    // input content; raw is always preferable to silence.
-    if parsed * 10 < lines.len() * 6 || normalized.is_empty() {
+    let _ = harness;
```
PR-Reviewer-Ant: **Dead `harness` parameter in `normalize_session_transcript`.**

`let _ = harness;` suppresses the unused-variable warning but keeps an inert parameter in the public signature. Remove `harness: &str` from the function signature and all call sites. Flagged in the latest review round (comment 3038365638); not addressed.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -315,31 +316,194 @@ fn normalize_json_transcript_line(value: &serde_json::Value) -> Option<String> {
...
+}
+
+fn is_safe_audit_name(value: &str) -> bool {
+    !value.is_empty()
+        && value
+            .bytes()
+            .all(|byte| byte.is_ascii_alphanumeric() || matches!(byte, b'-' | b'_' | b'.'))
+}
+
+fn build_audit_path(root: &Path, file_name: &str) -> std::io::Result<PathBuf> {
```
PR-Reviewer-Ant, line 431: **Unresolved CodeQL path-expression alerts.**

GitHub Advanced Security flagged `build_audit_path` and the surrounding helpers as "Uncontrolled data used in path expression" (CodeQL alerts #68–71). The `is_safe_audit_name` validation (hex-only token + `audit_fs_timestamp` sanitization) should prevent actual traversal, making these likely false positives. They should be addressed with either a targeted `// SAFETY:` or CodeQL suppression annotation, or confirmed as accepted risk, so they do not accumulate across future CI runs.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `98d1f83e`

[Automated PR Review — pr-reviewer] This PR achieves its stated goals: conversation-only canonical transcripts, raw audit logs under .daemon/logs/transcripts/, session-end resilience via stored live transcript fallback, and updated summary prompts and system prompt. Most concerns from prior reviewer rounds are addressed. Two issues from the latest round (comment IDs 3038365628 and 3038365631, both against the current head 98d1f83e) remain unresolved in the diff: the Rust prompt_submit handler reads transcript_path from disk without the workspace-root guard that session_end enforces, and the Rust upsert path lacks the length-guard present in the TS counterpart. One dead parameter and unresolved CodeQL alerts also need resolution.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The blocking path-read gap is directly provable from the diff: prompt_submit (hooks.rs ~line 889) calls fs::read_to_string on a caller-supplied path with zero workspace-root validation, while session_end in the same file returns 403 for paths outside allowed roots. The TS-Rust upsert parity gap is confirmed by comparing hooks.ts handleUserPromptSubmit (length guard present at line 2264) against the Rust unconditional write at line 905. Full prior review history (5 rounds, 18 inline comments) confirms both issues are open in the current head and were raised by the previous reviewer in the latest pass.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +879,98 @@ pub async fn prompt_submit(
...
+                return (
+                    StatusCode::FORBIDDEN,
+                    Json(serde_json::json!({"error": "transcript_path outside allowed workspace roots"})),
+                )
+                    .into_response();
+            }
+            match fs::read_to_string(&canonical) {
+                Ok(content) => raw_transcript = content,
+                Err(e) => warn!(path = %canonical.display(), error = %e, "prompt-submit: transcript read failed"),
+            }
```
PR-Reviewer-Ant, line 907: TOCTOU in transcript length guard: `should_write` is resolved in a `pool.read()` call, then used unconditionally in a separate `pool.write()` call. Between those two awaits, a concurrent `prompt_submit` for the same `(session_key, agent_id)` pair could write a longer transcript. When that happens, this call's `should_write = true` was computed against stale state and it will overwrite the longer stored content with its shorter version — exactly the data loss the guard is meant to prevent.

The risk is low for single-session sequential use (the normal path), but multi-harness or retried prompt-submits can trigger it. Options: (1) move the length comparison inside the write transaction using a single SQL `UPDATE … WHERE length(content) < ?1`-style guard; (2) use `INSERT OR IGNORE` if a longer snapshot is acceptable to win by recency rather than length; or (3) add a comment accepting last-write-wins semantics for this field.

The equivalent TS code in `handleUserPromptSubmit` is safe because it uses a synchronous DB call with no async boundary between read and write.

packages/daemon/src/transcript-audit.ts
```diff
@@ -0,0 +1,61 @@
...
+	return createHash("sha256").update(`${agentId}:${scoped}`, "utf8").digest("hex").slice(0, 16);
```
PR-Reviewer-Ant, line 27: Audit token content-hash fallback diverges between Rust and TS. When both `sessionId` and `sessionKey` are absent, TS computes `sha256(raw).digest('hex')` (full 64-char hex string) as the scoped value; Rust computes `sha256(raw)[..16]` (32 hex chars). The outer `sha256(agentId:scoped)` then differs, so the audit file names for the same session won't match across implementations. Since this path is only reached when session identity is completely absent, the practical impact is limited to audit forensics. Align the fallback lengths for cross-implementation consistency.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -315,31 +315,194 @@ fn normalize_json_transcript_line(value: &serde_json::Value) -> Option<String> {
...
+        let text = extract_json_message_text_from_map(message);
+        match (role.as_deref(), text) {
+            (Some("user"), Some(text)) => return Some(format!("User: {text}")),
+            (Some("assistant"), Some(text)) => return Some(format!("Assistant: {text}")),
+            _ => {}
+        }
+    }
+
+    let role = extract_json_string(value, &["role", "speaker"]);
+    let text = extract_json_message_text(value);
```
PR-Reviewer-Ant, line 329: The pre-existing `item.completed` handler in `normalize_json_transcript_line` (not changed by this PR) still uses old-style direct field access — no trim, no empty-string filter, no `\r\n` → space normalization — while the `event_msg` path was refactored in this PR to use `extract_json_string` which does apply all three. Assistant turns from `item.completed` events can therefore contain raw newlines in the canonical transcript while user turns from `event_msg` cannot. Minor inconsistency in text normalization across event types. Not introduced by this PR, but worth addressing while the function is being touched.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a76b1f0c`

[Automated Review — pr-reviewer bot] PR #466 hardens transcript capture and summary inputs. The architecture is sound: Codex normalization bypass is removed from Rust (TS never had the bypass — it already routed Codex through `normalizeCodexTranscript`), live-transcript fallback at session-end is well-guarded, audit log path handling is properly sanitized, and test coverage for the new paths is solid. Two issues worth discussion before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The TOCTOU and audit-token issues are directly provable from the diff. The Rust `normalize_json_transcript_line` contains a pre-existing `item.completed` handler not visible in the diff (confirmed by reading the current file), which makes the new Codex test correct. `session-transcripts.ts` is not in the diff so the exact `upsertSessionTranscript` implementation is not confirmed, but the read-then-write pattern in prompt_submit is fully visible.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -706,6 +874,87 @@ pub async fn prompt_submit(
...
+                        Json(serde_json::json!({"error": format!("transcript_path unresolvable: {e}")})),
+                    )
+                        .into_response();
+                }
+            };
+            if !transcript_path_allowed(&canonical) {
+                warn!(path = %canonical.display(), "prompt-submit: transcript_path outside allowed roots");
+                return (
+                    StatusCode::FORBIDDEN,
+                    Json(serde_json::json!({"error": "transcript_path outside allowed workspace roots"})),
```
PR-Reviewer-Ant, line 898: When `transcript_path` is provided but the file does not exist (or `canonicalize` fails for any reason), `prompt_submit` returns `500 INTERNAL_SERVER_ERROR`. The TS equivalent silently skips transcript handling via `existsSync` and continues to serve prompt context normally. A harness that defensively passes `transcript_path` before the file is guaranteed to exist (e.g., a file created asynchronously) will receive a 500 and lose prompt-time memory injection entirely — a session-level regression with no retry path. The appropriate behavior is a `warn!` and a no-op skip, matching the TS path. The path-outside-roots case correctly returns 403; only the missing/unresolvable file case needs to change.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1285,7 +1536,29 @@ pub async fn session_end(
...
+        let session_key_value = session_key.clone();
+        let agent_value = agent_id.clone();
+        if let Ok(Some(stored)) = state
+            .pool
+            .read(move |conn| {
+                session_transcript_content(conn, &session_key_value, &agent_value)
+                    .map_err(|err| signet_core::error::CoreError::Migration(err.to_string()))
+            })
+            .await
+            && !stored.trim().is_empty()
```
PR-Reviewer-Ant, line 1550: The stored-transcript fallback query is `WHERE session_key = ?1 AND agent_id = ?2`. The `agent_id` used here comes from `resolve_remember_agent` at session-end time, which fuses the body field, the `x-signet-agent-id` header, and the session-key prefix. The transcript stored during `prompt_submit` used the same resolution logic, but if the header is absent at session-end (e.g., the harness omits it on the final call), `agent_id` falls back to `"default"` while the stored row was written under a named agent. The fallback would silently miss the snapshot and `normalized` would remain whatever the final input produced. Given that this fallback is the primary resilience mechanism for dropped sessions, a mismatch of agent IDs produces a silent no-op with no observable error signal. Consider logging the resolved agent_id at both call sites at `info` level so mismatches are detectable in logs.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -2686,6 +2977,205 @@ mod tests {
...
+                harness: Some("test".to_string()),
+                project: Some("packages/daemon-rs".to_string()),
+                agent_id: Some("agent-a".to_string()),
+                user_message: Some("review the release checklist".to_string()),
+                user_prompt: None,
+                last_assistant_message: None,
+                session_key: Some("sess-prompt-guard".to_string()),
+                transcript: Some("User: short\nAssistant: short".to_string()),
+                transcript_path: None,
+                runtime_path: None,
```
PR-Reviewer-Ant, line 3049: The test `prompt_submit_writes_transcript_audit_and_live_snapshot` writes to a hardcoded path `/tmp/signet` and relies on that path being within `transcript_path_allowed` roots. If the daemon's allowed roots are determined by `SIGNET_PATH` or workspace config at test time, and `/tmp/signet` is outside them, the test would fail with a 403 instead of the expected 200. The test does not assert on the status code until after the path check, so a misconfigured environment would produce a confusing failure rather than a clear guard. Consider using the test workspace temp dir (already available as `tmp.path()`) and symlinking or using an in-workspace path.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fae5a19f`

[Automated Bot Review — pr-reviewer] Reviewed transcript hardening + summary prompt update. Found one warning-severity behavioral divergence in Rust error handling for prompt_submit transcript paths, and one conditional logic concern in the session-end stored-transcript fallback that could silently miss snapshots under agent_id mismatch.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The 500 vs. skip divergence between Rust prompt_submit and TS is directly readable in the diff (hooks.rs added lines ~874–898 vs hooks.ts existsSync guard). The agent_id scoping risk on the fallback lookup is provable from the SQL query added in hooks.rs. The TS normalizeSessionTranscript function body is not in the diff; the test `writes raw audit logs while keeping the canonical transcript conversation-only` uses harness:codex and asserts tool calls are stripped, which is strong evidence that the TS function never had a Codex bypass — the bypass was Rust-only, and the PR's parity claim holds. No missing test file or runtime trace changes that finding.


- packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs:299 - The fallback rule changed: previously `if parsed * 10 < lines.len() * 6 || normalized.is_empty()` fell back to raw. The second condition (`normalized.is_empty()`) is now removed, so a recognized-JSONL input that produces zero conversation turns returns `""`. This is correct for the stated goal (tool-only sessions produce no canonical transcript). However, the threshold `parsed * 10 < lines.len() * 6` only counts lines that were attempted as JSON — comment lines, blank lines, or partial writes appended after a crash may not be counted in `lines.len()`, which could inflate the denominator and cause legitimate sessions with leading non-JSON header text to fall back to raw unexpectedly. This is a pre-existing edge case that the refactor preserves, but the new `normalized.is_empty()` gate removal makes the fallback stricter, so mixed-format inputs near the 60% threshold may behave differently than before. Worth a targeted test covering transcripts that start with a non-JSON preamble followed by JSONL.

PR #469, opened 2026-04-06
feat(pipeline): add token-bucket rate limiting for remote LLM providers

### LINKED ISSUE #341: Feature request: rate limiting and cost protection for claude-code LLM provider

aaf2tbz:
## Summary

Closes #341 (Option A — token bucket rate limiter)

Prevents runaway subprocess spawning during pipeline stall loops by throttling calls to remote/paid LLM providers via a token-bucket rate limiter.

### `packages/core/src/types.ts`
- Add `ProviderRateLimitConfig` interface and default export for provider rate-limiting

### `packages/daemon/src/pipeline/provider.ts`
- Add `TokenBucketRateLimiter` class with fractional refill, burst allowance, and optional blocking wait
- Add `RateLimitExceededError` thrown when rate limit is exceeded after wait timeout
- Add `withRateLimit()` decorator that wraps `LlmProvider` to enforce rate limits before `SubprocessSemaphore.acquire()`
- Remove stale `stats` getter; `currentStats()` is the sole exported snapshot method to prevent callers from reading pre-refill token counts
- Extract shared `warnIfThrottled()` helper so generate/generateWithUsage closures cannot race on `lastWarnMs`
- Poll interval uses `max(1, min(100, waitMs/4))` for multi-shot refill checking within the deadline
- Extract narrowed `maxCallsPerHour` and `burstSize` locals after nullish-coalesce to satisfy strict TypeScript without type assertions
- Both `generate` and `generateWithUsage` wrappers use consistent `fn.call(provider, ...)` this-binding
- Filter explicit `undefined` values from config before spreading onto defaults, so partial configs fall back correctly instead of silently disabling rate limiting
- Remove redundant `Partial<>` wrapper on `withRateLimit` parameter (fields are already optional)
- Add cross-reference comment on `RATE_LIMIT_PROVIDERS` set linking to `PipelineExtractionConfig` provider union

### `packages/daemon/src/daemon.ts`
- Wrap extraction provider with `withRateLimit()` before `initLlmProvider()`
- Wrap synthesis provider with `withRateLimit()` before `initSynthesisProvider()`

### `packages/daemon/src/memory-config.ts`
- Parse `rateLimit` for both extraction and synthesis pipeline config

### `packages/daemon/src/pipeline/worker.ts`
- Treat provider rate-limit exhaustion as non-retryable dead-letter
- Extract `deadLetterJob(db, id, msg)` as a direct SQL write (`status='dead'`) that does not route through `failJob`
- Escalate rate-limit dead-letter to `logger.error` so operators are alerted about permanent data loss

### `packages/daemon/src/pipeline/summary-worker.ts`
- `isTerminalSummaryJobError` accepts `string | Error` union; `instanceof RateLimitExceededError` check runs on the first argument so callers cannot accidentally skip classification

### `docs/CONFIGURATION.md`
- Document `rateLimit.maxCallsPerHour`, `burstSize`, and `waitTimeoutMs`
- Document dead-letter behavior and pass-through for non-whitelisted providers
- Document in-memory-only rate limiter state and restart limitation

### Round 1 (`07972bb7`)
- Fix `generateWithUsage` this-binding (detached method → `fn.call(provider, ...)`)
- Add dead-letter documentation to CONFIGURATION.md
- Strip unrelated dashboard/extension changes

### Round 2 (`362f6ae8`)
- Replace magic string check in `isTerminalSummaryJobError` with `instanceof RateLimitExceededError`
- Add clarifying comment about `burstSize: undefined` spread behavior
- Add regression test for RateLimitExceededError terminal classification

### Round 3 (`8a52d644`)
- Guard `waitTimeoutMs` undefined → 0 to prevent NaN in poll interval
- Revert unrelated Codex `mcp_servers` config change
- Add regression test for undefined waitTimeoutMs

### Round 4 (`78d84fde`)
- Accept `string | Error` union in `isTerminalSummaryJobError` so future callers cannot silently skip instance classification
- Remove stale `stats` getter from `TokenBucketRateLimiter`; all consumers now use `currentStats()` exclusively
- Update tests to use `currentStats()` and pass Error objects to the union-type API

### Round 5 (`62429bfe`)
- Revert `Math.max(waitMs, ...)` back to `Math.max(1, ...)` for poll interval — previous fix made the loop single-shot for waitMs >= 400ms, silently breaking the retry window
- Remove second parameter from `isTerminalSummaryJobError`; instanceof check runs on the first argument, making misuse impossible
- Update call site and tests for simplified single-parameter API

### Round 6 (`81a52cbb`)
- Extract `maxCallsPerHour` and `burstSize` as narrowed `number` locals after nullish-coalesce, satisfying strict TypeScript without `as` or `!`
- Extract `deadLetterJob(db, id, msg)` from `failJob` to make non-retryable dead-lettering explicit at the call boundary

### Round 7 (`fb581277`)
- `deadLetterJob` now writes `status='dead'` directly via SQL instead of routing through `failJob` with a sentinel value, fully decoupling from `failJob` internals

### Round 8 (`0493d9d3`)
- Align `generate` wrapper to use `fn.call(provider, ...)` like `generateWithUsage`, removing the inconsistent this-binding pattern
- Add doc comment on `deadLetterJob` explaining intentional attempts non-increment
- Document in-memory-only rate limiter state and restart limitation in CONFIGURATION.md

### Round 9 (`f35e88d1`)
- Filter explicit `undefined` values from config before spreading onto defaults so partial configs fall back correctly instead of silently disabling rate limiting
- Remove redundant `Partial<>` wrapper (fields are already optional in the interface)
- Add cross-reference comment on `RATE_LIMIT_PROVIDERS` set linking to `PipelineExtractionConfig` provider union in types.ts
- Escalate rate-limit dead-letter to `logger.error` so operators are alerted about permanent data loss
- Update tests to reflect new undefined-fallback behavior

## Validation

- `bun test packages/daemon/src/pipeline/rate-limit.test.ts` — 24 pass
- `bun test packages/daemon/src/pipeline/summary-worker.test.ts` — 28 pass
- `bunx biome check` on all changed files — clean

## PR Readiness Checklist


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(pipeline): add token-bucket rate limiting for remote LLM providers` (commit `8a52d644`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -100,8 +101,8 @@ export function resolveSummaryHeadingDate(job: Pick<SummaryJobRow, "ended_at" |
 	return basis.slice(0, 10);
 }
 
-export function isTerminalSummaryJobError(errorMessage: string): boolean {
-	return errorMessage.startsWith(IMMUTABLE_ARTIFACT_ERROR_PREFIX);
+export function isTerminalSummaryJobError(errorMessage: string, error?: unknown): boolean {
```
PR-Reviewer-Ant: `isTerminalSummaryJobError` now has two separate classification paths: string-based (`IMMUTABLE_ARTIFACT_ERROR_PREFIX`) and instance-based (`instanceof RateLimitExceededError`). The instance check requires callers to pass the error object as the second argument — callers that pass only the message string will silently NOT classify rate-limit errors as terminal, causing them to retry up to `maxAttempts` times before dead-lettering (potential retry storm on a depleted bucket).

The single current call site in `startSummaryWorker` correctly passes `e`, and the test documents this behavior. But the asymmetric API is a footgun for any future call site added elsewhere. Consider either: (a) adding a string fallback `|| errorMessage.includes('Rate limit exceeded')` (accepting the fragility the PR previously avoided), or (b) accepting a union type `string | Error` and extracting the message internally, so callers can't accidentally omit the instance.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `8a52d644`

[Automated Review — pr-reviewer bot] PR #469 adds a token-bucket rate limiter for remote LLM providers. The core logic is correct and well-tested (22 + 28 tests). Two issues worth discussing before merge: a subtle API design footgun in `isTerminalSummaryJobError` and a locking concern in the polling loop.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The `failJob` implementation was verified in the live repo at packages/daemon/src/pipeline/worker.ts:237-249 — it uses the passed `attempts` argument only for dead/pending classification, not for DB storage. The SQL only writes `status`, `error`, `failed_at`, `updated_at`. The PR's assertion in the comment is correct. All other issues are directly provable from the changed lines in the diff.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,176 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+		if (this.tokens >= 1) {
+			this.tokens -= 1;
+			this.totalConsumed++;
+			return true;
+		}
+		if (waitMs <= 0) {
+			this.totalThrottled++;
+			return false;
+		}
+		const deadline = Date.now() + waitMs;
```
PR-Reviewer-Ant, line 138: The polling loop computes `pollIntervalMs = Math.max(1, Math.floor(Math.min(100, waitMs / 4)))`. For `waitMs <= 3` this yields `pollIntervalMs = 1ms`. Combined with `deadline = Date.now() + waitMs`, the loop may fire the `setTimeout(r, 1)` callback multiple times before the deadline expires (JS timer resolution is typically 1–4ms). This is harmless for correctness but could generate noisy repeated `setTimeout` overhead under tight timeouts. The existing test `respects waitTimeoutMs` covers the non-blocking return path. This is a low-severity concern, not a correctness bug.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,176 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+			}
+		}
+		this.totalThrottled++;
+		return false;
+	}
+
+	get stats(): { readonly remaining: number; readonly totalConsumed: number; readonly totalThrottled: number } {
+		return {
+			remaining: Math.floor(this.tokens),
+			totalConsumed: this.totalConsumed,
```
PR-Reviewer-Ant: `stats` (getter, no refill) and `currentStats()` (method, refills first) are both exported. The `remaining` field in `stats` is stale by design, but `totalConsumed` and `totalThrottled` are the same in both. A future caller who calls `bucket.stats.remaining` to make a throttling decision (rather than `currentStats().remaining`) would get a subtly wrong value. The distinction is documented in a comment, but because `stats` is a property accessor and `currentStats` is a method, callers could easily pick the wrong one. Consider making `stats` internal or removing it in favor of `currentStats()` only.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3ae04e4b`

```pr-review-json
{
  "summary": "[Automated Review - pr-reviewer] PR #469 adds token-bucket rate limiting for remote LLM providers. One functional bug found in the polling loop that makes `waitTimeoutMs` a delay-before-failure rather than a retry window; one API footgun in the terminal-error classifier.",
  "verdict": "request_changes",
  "confidence": {
    "level": "high",
    "justification": "The poll-interval formula is arithmetically provable from the diff: for any waitMs ≥ 400 ms, `Math.max(waitMs, Math.floor(Math.min(100, waitMs/4)))` reduces to `waitMs`, making the loop execute at most once. The `isTerminalSummaryJobError` signature issue is directly visible in the summary-worker.ts hunk. The only gap is that `failJob` is not in the diff; the comment claiming it does not persist the attempts column is unverifiable here, though the test validates the observable dead-letter outcome.",
    "justification_detail": "For the poll interval: Math.max(5000, floor(min(100, 1250))) = Math.max(5000, 100) = 5000. For any waitMs ≥ 400, the inner expression is capped at 100 by Math.min, then Math.max(waitMs, 100) = waitMs. The while-loop's deadline is also waitMs ms in the future, so one setTimeout fires, one refill check happens, then the deadline has passed and the loop exits."
  },
  "comments": [
    {
      "line": 130,
      "body": "**Bug: poll interval formula makes `waitTimeoutMs` a single-shot delay, not a retry window.**\n\n```ts\nconst pollIntervalMs = Math.max(waitMs, Math.floor(Math.min(100, waitMs / 4)));\n```\n\nFor any `waitMs ≥ 400 ms` the inner expression is capped at 100 by `Math.min`, and then `Math.max(waitMs, 100)` returns `waitMs`. With the default `waitTimeoutMs: 5000`, `pollIntervalMs = 5000` — the loop fires one `setTimeout`, refills, checks once, and then the deadline has already passed. The loop never executes more than once for any realistic timeout.\n\n**Impact:** With default settings (200/hr, 5 s timeout), only ~0.28 tokens refill during the single wait, so the limiter never actually queues a call — it always dead-letters after a 5 s delay. The documented behaviour ("wait up to `waitTimeoutMs` for a token") is silently broken. For higher refill rates the wait succeeds on the first shot but wastes the full timeout unnecessarily.\n\n**Fix:** replace `Math.max` with a floor+cap that is smaller than `waitMs`:\n```ts\nconst pollIntervalMs = Math.max(1, Math.floor(Math.min(100, waitMs / 4)));\n```\nThis polls every `waitMs/4` ms (capped at 100 ms, floored at 1 ms to prevent a spin loop), which is the behaviour the PR description appears to intend.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/pipeline/summary-worker.ts",
      "line": 104,
      "body": "**API footgun: `instanceof RateLimitExceededError` is checked on the second `error` parameter, not the first `input`.**\n\n```ts\nexport function isTerminalSummaryJobError(input: string | Error, error?: unknown): boolean {\n    const message = typeof input === \"string\" ? input : input.message;\n    return message.startsWith(IMMUTABLE_ARTIFACT_ERROR_PREFIX) || error instanceof RateLimitExceededError;\n}\n```\n\nThe signature accepts `Error` as the first argument, but rate-limit terminal classification only fires via the optional second argument. A future caller who passes a `RateLimitExceededError` as the sole argument will silently skip the `instanceof` branch and return `false` (no `IMMUTABLE_ARTIFACT_ERROR_PREFIX` match). The current call site always passes `e` as both args, so behaviour is correct today, but the dual-parameter API is non-obvious.\n\n**Suggested fix** — unify the check on the first parameter:\n```ts\nexport function isTerminalSummaryJobError(input: string | Error): boolean {\n    const message = typeof input === \"string\" ? input : input.message;\n    return (\n        message.startsWith(IMMUTABLE_ARTIFACT_ERROR_PREFIX) ||\n        input instanceof RateLimitExceededError\n    );\n}\n```\nThis removes the second parameter entirely, making it impossible to misuse.",
      "severity": "warning"
    }
  ]
}
```

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,166 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+			},
+		);
+		return provider;
+	}
+
+	const bucket = new TokenBucketRateLimiter(cfg.maxCallsPerHour, cfg.burstSize);
+	let lastWarnMs = 0;
+	const WARN_INTERVAL_MS = 300_000;
+
+	function warnIfThrottled(): void {
```
PR-Reviewer-Ant, line 213: Type narrowing gap under strict null checks. `const cfg = { ...DEFAULT_PROVIDER_RATE_LIMIT, ...config }` — because `config: Partial<ProviderRateLimitConfig>` can carry explicit `undefined` values for each key, TypeScript widens the spread result so that `cfg.maxCallsPerHour` and `cfg.burstSize` are typed as `number | undefined`. The `(cfg.maxCallsPerHour ?? 0) <= 0` guard ensures runtime safety but does NOT narrow the TypeScript type, so the subsequent `new TokenBucketRateLimiter(cfg.maxCallsPerHour, cfg.burstSize)` call passes `number | undefined` to `number` parameters, which is a type error under `"strict": true`. The same issue appears at the two `new RateLimitExceededError(provider.name, cfg.maxCallsPerHour)` throw sites below.

Minimal fix that avoids `as` and `!`:
```typescript
const maxCallsPerHour = cfg.maxCallsPerHour ?? 0;
const burstSize = cfg.burstSize ?? 0;
if (maxCallsPerHour <= 0 || burstSize <= 0) return provider;
const bucket = new TokenBucketRateLimiter(maxCallsPerHour, burstSize);
// ...
throw new RateLimitExceededError(provider.name, maxCallsPerHour);
```
This extracts narrowed `number` locals, satisfies the type checker, and is semantically equivalent to the current guard.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -1534,27 +1533,41 @@ export function startWorker(
...
+						: msg.includes("timeout")
+							? "EXTRACTION_TIMEOUT"
+							: "EXTRACTION_PARSE_FAIL",
 					durationMs: runtime.now() - jobStart,
 				});
 				accessor.withWriteTx((db) => {
-					failJob(db, job.id, msg, job.attempts, job.max_attempts);
-					if (job.attempts >= job.max_attempts) {
+					// failJob() uses the attempts argument *only* for dead-vs-retryable
+					// classification — it does NOT persist it to the `attempts` column
```
PR-Reviewer-Ant: `failJob(db, job.id, msg, nonRetryable ? job.max_attempts : job.attempts, job.max_attempts)` — the correctness of this dead-letter shortcut depends entirely on `failJob` using its 4th argument only for dead-vs-retryable classification and NOT persisting it to the `attempts` DB column. The inline comment documents this assumption, and the test asserts `job?.attempts === 1` to verify it. However, the assumption is invisible at `failJob`'s call boundary: if `failJob` is later refactored to also persist the argument, rate-limited jobs dead-lettered on attempt 1 would silently show `attempts = max_attempts` in the DB, making debugging misleading. Consider a dedicated `deadLetterJob(db, id, msg)` function that makes the intent explicit and can't be misused by future callers.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `62429bfe`

[Automated Review — pr-reviewer bot] PR #469 adds a token-bucket rate limiter for remote LLM providers. The core logic, error propagation, dead-letter classification, and config parsing are correct. Two non-blocking issues found: a TypeScript type-narrowing gap under strict null checks, and a fragile load-bearing assumption about `failJob`'s internal behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The type-narrowing issue is directly provable from the diff: `{ ...DEFAULT_PROVIDER_RATE_LIMIT, ...config }` where `config: Partial<ProviderRateLimitConfig>` yields `cfg.maxCallsPerHour: number | undefined` under TypeScript strict null checks. The `failJob` contract concern is provable from the worker.ts diff comment and the test assertion `expect(job?.attempts).toBe(1)`. Both findings are grounded in changed lines and do not require runtime reproduction.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `81a52cbb`

[Automated Review] (pr-reviewer) One warning-level finding; no blocking issues. Implementation correctly achieves the stated goals.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The full body of `failJob` is not in the diff — only call sites and the 5-argument `.run(nextStatus, error, now, now, jobId)` fragment are visible. The `deadLetterJob` analysis depends on inferring that `failJob` does NOT persist its `attempts` parameter to the DB's `attempts` column. That inference is reasonable (5 args, none obviously `attempts`) but cannot be confirmed without the full implementation.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -248,6 +247,10 @@ function failJob(db: WriteDb, jobId: string, error: string, attempts: number, ma
 	).run(nextStatus, error, now, now, jobId);
 }
 
+function deadLetterJob(db: WriteDb, jobId: string, error: string): void {
+	failJob(db, jobId, error, Number.MAX_SAFE_INTEGER, 1);
```
PR-Reviewer-Ant: **`deadLetterJob` passes `Number.MAX_SAFE_INTEGER` as an attempts sentinel — fragile implicit contract with `failJob`.**

`failJob(db, jobId, error, Number.MAX_SAFE_INTEGER, 1)` forces the `attempts >= maxAttempts` branch to compute `nextStatus = 'dead'`. This only works safely if `failJob` does **not** write the `attempts` argument directly to the DB column (inferred from the 5-arg `.run(nextStatus, error, now, now, jobId)` visible in the diff). If `failJob`'s SQL is later extended to persist its `attempts` parameter, dead-lettered rate-limit jobs would store `9007199254740991` in their `attempts` column, silently corrupting analytics and observability.

The PR description notes this risk ("cannot silently inflate attempt counts if `failJob` internals change"), but the helper still depends on that internal staying stable. A more durable option: implement `deadLetterJob` as a direct SQL write (`status='dead', error=?, updated_at=?, failed_at=?`) without routing through `failJob`.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -248,6 +247,15 @@ function failJob(db: WriteDb, jobId: string, error: string, attempts: number, ma
...
+function deadLetterJob(db: WriteDb, jobId: string, error: string): void {
+	const now = new Date().toISOString();
+	db.prepare(
+		`UPDATE memory_jobs
+		 SET status = 'dead', error = ?, failed_at = ?, updated_at = ?
+		 WHERE id = ?`,
+	).run(error, now, now, jobId);
+}
+
 function updateExtractionStatus(db: WriteDb, memoryId: string, status: string, extractionModel?: string): void {
```
PR-Reviewer-Ant, line 264: **`deadLetterJob` silently drops the `attempts` increment.**

`failJob` writes `attempts = attempts + 1` in SQL for every failure, including terminal ones. `deadLetterJob` does not. This means a rate-limited job that is dead-lettered on its first run shows a different `attempts` value than a job that exhausts retries and goes dead via `failJob`. The test validates the current observable value (`attempts = 1`, presumably left by the lease), but any monitoring or admin-UI query that uses `attempts` as "number of times processing was attempted" will read a structurally different number for these two dead-job variants. If `attempts` is used anywhere as a decision input (e.g. retention heuristics, alerting), this discrepancy could cause silent misbehaviour. At minimum, a comment on `deadLetterJob` explaining that it intentionally does not increment `attempts` would prevent a future maintainer from "fixing" it in a way that breaks the test expectation.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,165 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+
+const RATE_LIMIT_PROVIDERS: ReadonlySet<string> = new Set([
+	"claude-code",
+	"anthropic",
+	"openrouter",
+	"codex",
+	"opencode",
+]);
+
+function shouldRateLimit(providerName: string): boolean {
```
PR-Reviewer-Ant, line 184: **Rate-limiter state is in-memory only; full burst is available immediately after every daemon restart.**

`TokenBucketRateLimiter` initialises `this.tokens = burstSize` on construction. After a daemon restart the bucket is fresh, so `burstSize` (default 20) calls can go out before any throttling begins. In environments with frequent restarts (crash-loops, rolling deployments, watchdog restarts under load), this defeats the purpose of the limiter for the exact scenario it is meant to guard: runaway LLM calls during a pipeline stall that cycles the daemon. The docs say nothing about this limitation. A restart-safe workaround is not required, but a documentation note in `CONFIGURATION.md` would help operators set `burstSize` conservatively.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,165 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	function warnIfThrottled(): void {
+		const now = Date.now();
+		if (now - lastWarnMs > WARN_INTERVAL_MS) {
+			lastWarnMs = now;
+			const stats = bucket.currentStats();
+			logger.warn("pipeline", `Rate limit throttled ${provider.name} (${stats.totalThrottled} total)`, stats);
+		}
+	}
+
+	return {
```
PR-Reviewer-Ant, line 223: **Inconsistent this-binding pattern between `generate` and `generateWithUsage` in the wrapper.**

`generate` is invoked as `provider.generate(prompt, opts)` (method call — `this` is correctly bound to `provider`). `generateWithUsage` extracts a function reference and uses `fn.call(provider, prompt, opts)` to rebind explicitly. Round 1 of review addressed a real bug in `generateWithUsage`; the asymmetry remains. Both are correct for the current provider shapes, but the inconsistency signals to future readers that `provider.generate` is safe to call directly while `provider.generateWithUsage` is not — which is not generally true and could lead to a copy-paste error if a third wrapper method is added. Aligning both to the same style (either always method-call or always explicit-bind) would remove the ambiguity.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fb581277`

[Automated Review — pr-reviewer bot] PR #469 adds a token-bucket rate limiter for remote LLM providers in the Signet pipeline. The implementation is well-structured and thoroughly tested (22 + 28 passing tests). No blocking bugs or security issues found. Three warning-level concerns noted below.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All changed files are present in full. The rate-limiter logic, config parser, worker dead-letter path, and summary-worker classification are all directly readable. The leasing logic (which determines pre-deadLetterJob `attempts` value) is not in the diff, but the test behaviour (attempts=1 for the rate-limited case) provides sufficient evidence to assess the consistency issue without it.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0493d9d3`

```pr-review-json
{
  "summary": "[Automated Review — pr-reviewer] Token-bucket rate limiting for remote LLM providers (8 rounds of prior review addressed). Core algorithm is correct, tests are thorough, and the YAML path is safe. Two API-level issues worth discussing before merge.",
  "verdict": "comment",
  "confidence": {
    "level": "medium",
    "justification": "The token-check/decrement path in `acquire()` and the `withRateLimit` spread behavior are directly visible and provable from the diff. The `attempts` column increment behavior in the lease/claim path (before the error handler runs) is NOT in the diff — the test asserts `attempts: 1` after one rate-limited attempt, which is consistent with the lease pre-incrementing, but I cannot confirm that from the diff alone. Everything else is directly observable."
  },
  "comments": [
    {
      "line": 168,
      "body": "**`withRateLimit` API footgun: explicit `undefined` field in a partial config silently disables rate limiting instead of falling back to the default.**\n\nThe spread `{ ...DEFAULT_PROVIDER_RATE_LIMIT, ...config }` means that `config = { maxCallsPerHour: 100, burstSize: undefined }` produces `cfg.burstSize = undefined`, which then becomes `0` via `?? 0`, triggering the `burstSize <= 0` early-return and completely bypassing the limiter. A caller who intends "100 calls/hour with default burst" accidentally gets no rate limiting at all.\n\nThe YAML path is protected because `parseRateLimitConfig` always fills all three fields before calling `withRateLimit`. But the exported function's type signature (`Partial<ProviderRateLimitConfig>`) explicitly invites partial objects, and the comment explaining this quirk is easy to miss. Future callers (tests, SDK users, per-session dynamic configs) are not protected.\n\nConsider either: (a) documenting the footgun prominently in the JSDoc on `withRateLimit`, (b) treating `undefined` fields as missing (fall back to defaults rather than treating as 0), or (c) narrowing the parameter type to `ProviderRateLimitConfig | undefined` to match the actual call sites and make partial objects a type error.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/pipeline/worker.ts",
      "line": 1564,
      "body": "**Dead-letter on first rate-limit hit is a permanent data loss path with no self-healing.**\n\n`deadLetterJob` is called on the first `RateLimitExceededError` regardless of `max_attempts`. If the rate limit fires briefly during a short burst (e.g. a backfill run), all jobs enqueued during that window are permanently killed — they are never retried even after the token bucket refills in the next hour. The 30-day purge is the only recovery path.\n\nThis is documented in CONFIGURATION.md, but the consequences are severe enough that it warrants at least a structured log at `error` level (not just `warn`) when a job is dead-lettered for rate limiting, so operators are alerted. Currently the only signal is the `nonRetryable` field in the existing `warn` log.",
      "severity": "warning"
    },
    {
      "line": 148,
      "body": "`Partial<ProviderRateLimitConfig>` is redundant — every field in `ProviderRateLimitConfig` is already declared optional (`?`). The parameter type is identical to `ProviderRateLimitConfig | undefined` for practical purposes. Not a bug, but the extra `Partial<>` wrapper adds confusion about why it's needed.",
      "severity": "nitpick"
    },
    {
      "line": 138,
      "body": "`RATE_LIMIT_PROVIDERS` is a hardcoded set. When a new remote/paid provider is added to `PipelineExtractionConfig`'s union type, it must also be manually added here or it will silently pass through unthrottled — with no compile-time or runtime warning beyond the existing log. Consider adding a comment cross-referencing the provider union in `types.ts` to make this coupling explicit for future contributors.",
      "severity": "nitpick"
    }
  ]
}
```

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,170 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+
+		async generate(prompt, opts): Promise<string> {
+			if (!(await bucket.acquire(waitTimeoutMs))) {
+				warnIfThrottled();
+				throw new RateLimitExceededError(provider.name, maxCallsPerHour);
+			}
+			const fn = provider.generate;
+			return fn.call(provider, prompt, opts);
+		},
+
```
PR-Reviewer-Ant, line 234: TypeScript narrowing loss in `generateWithUsage` async closure. The outer ternary `provider.generateWithUsage ? { ... } : {}` narrows the property to non-undefined in the object literal scope, but TypeScript does not preserve that narrowing inside the async function body. `const fn = provider.generateWithUsage` is therefore typed as `((prompt, opts?) => Promise<LlmGenerateResult>) | undefined`, and `fn.call(provider, prompt, opts)` should be a compile error under `strict: true` (which the project's root tsconfig enables). Runtime behavior is safe — the wrapper method only exists when the ternary is truthy — but it contradicts the PR's claim of clean typechecks. Fix: add a non-null assertion `provider.generateWithUsage!` at the assignment, or restructure as `const fn = provider.generateWithUsage as typeof provider.generateWithUsage & Function` (or use a narrowed local before the ternary). The `generate` wrapper avoids this problem because `LlmProvider.generate` is non-optional.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,170 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+}
+
+export function withRateLimit(provider: LlmProvider, config?: ProviderRateLimitConfig): LlmProvider {
+	if (config === undefined) return provider;
+	if (Object.keys(config).length === 0) return provider;
+	// Filter out explicit undefined values so that partial configs like
+	// `{ maxCallsPerHour: 100, burstSize: undefined }` fall back to
+	// defaults instead of silently disabling rate limiting.
+	const clean: ProviderRateLimitConfig = {};
+	for (const [k, v] of Object.entries(config)) {
```
PR-Reviewer-Ant: `clean[k as keyof ProviderRateLimitConfig] = v` uses an `as` type assertion. The repo conventions explicitly prohibit `as` (AGENTS.md: "Avoid `any`, `as`, and non-null assertions"). Use a type guard or narrowed helper instead — e.g. `function isRateLimitKey(k: string): k is keyof ProviderRateLimitConfig { return k === 'maxCallsPerHour' || k === 'burstSize' || k === 'waitTimeoutMs'; }` — to satisfy strict TypeScript without the cast.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +83,170 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+		return false;
+	}
+
+	currentStats(): { readonly remaining: number; readonly totalConsumed: number; readonly totalThrottled: number } {
+		this.refill();
+		return {
+			remaining: Math.floor(this.tokens),
+			totalConsumed: this.totalConsumed,
+			totalThrottled: this.totalThrottled,
+		};
```
PR-Reviewer-Ant, line 159: `RATE_LIMIT_PROVIDERS` is a manually maintained `ReadonlySet<string>` that must be kept in sync with the `PipelineExtractionConfig.provider` union in `types.ts`. If a future contributor adds a new remote provider to the union but omits it here, rate limiting is silently bypassed with no compiler warning (the comment acknowledges this). Consider a compile-time cross-check: e.g. a `satisfies` expression or a type-level test that errors if a known remote provider name is missing from the set. Not a current bug — the set is correct today — but the maintenance surface is worth hardening.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f35e88d1`

[Automated Review — pr-reviewer bot] Reviewed PR #469 (token-bucket rate limiter for remote LLM providers). Implementation is largely correct with good test coverage. Found one TypeScript type-safety inconsistency and two minor convention deviations. No blocking correctness or security issues.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The rate-limiter logic, dead-letter path, and config parsing are all directly readable from the diff. The TypeScript narrowing concern is provable: the repo's root tsconfig.json has `strict: true` (confirmed), `LlmProvider.generateWithUsage` is `?`-optional (line 26 of types.ts), and TypeScript 5.x does not preserve control-flow narrowing for optional properties across async closure boundaries — making `fn.call(provider, prompt, opts)` a latent type error under the project's declared config. The `as` cast and set-sync risk are directly visible in the diff. No missing cross-module context is needed for these findings.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +84,175 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	}
+}
+
+type RemoteProvider = Exclude<PipelineExtractionConfig["provider"], "none" | "ollama" | "command">;
```
PR-Reviewer-Ant, line 163: **Broken exhaustiveness check — comment is incorrect and creates false safety confidence.**

```ts
const _exhaustiveCheck: RemoteProvider[] = [...RATE_LIMIT_PROVIDERS] as unknown as RemoteProvider[];
```

The double cast `as unknown as RemoteProvider[]` fully bypasses TypeScript's type system. If a new provider is added to `PipelineExtractionConfig["provider"]` (e.g. `"gemini"`) without being added to `RATE_LIMIT_PROVIDERS`, `RemoteProvider` would include `"gemini"`, but this line would NOT produce a type error — the cast makes TypeScript accept any `string[]` as `RemoteProvider[]`.

At runtime, `shouldRateLimit("gemini")` returns `false`, a warning is logged, and the new provider is silently unthrottled. The warning may not be seen in all deployments.

To actually enforce the invariant, use a mapped object literal instead:
```ts
const _exhaustiveCheck: Record<RemoteProvider, true> = {
  "claude-code": true,
  "anthropic": true,
  "openrouter": true,
  "codex": true,
  "opencode": true,
};
```
This produces a real compile-time error when the union grows.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -100,8 +101,9 @@ export function resolveSummaryHeadingDate(job: Pick<SummaryJobRow, "ended_at" |
 	return basis.slice(0, 10);
 }
 
-export function isTerminalSummaryJobError(errorMessage: string): boolean {
-	return errorMessage.startsWith(IMMUTABLE_ARTIFACT_ERROR_PREFIX);
+export function isTerminalSummaryJobError(input: string | Error): boolean {
```
PR-Reviewer-Ant, line 104: **No integration test for the synthesis dead-letter path on rate-limit exhaustion.**

The extraction worker has a full end-to-end test in `worker.test.ts` (`marks job dead without retry when provider rate limit is exceeded`) that proves the job status, attempts count, and error message after a `RateLimitExceededError`. The synthesis/summary-worker counterpart is only tested at the unit level (`isTerminalSummaryJobError` classification). There is no test verifying that a rate-limited synthesis job lands in `status='dead'` in the DB without being retried.

If `resolveFailedSummaryJobStatus` (not shown in this diff) does not correctly consume the `terminal` flag for dead-lettering, rate-limited synthesis jobs could be retried indefinitely, contradicting the documented behavior and the AGENTS.md rule about missing regression tests.

packages/daemon/src/daemon.ts
```diff
@@ -12284,9 +12287,12 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 												}
 											: {}),
```
PR-Reviewer-Ant, line 12288: **Widget provider intentionally bypasses rate limiting — worth an explicit operator warning.**

```ts
const widgetProvider = synthesisProvider;         // unwrapped
synthesisProvider = withRateLimit(synthesisProvider, ...);
initWidgetProvider(widgetProvider);               // unthrottled
```

This is documented in CONFIGURATION.md, but an operator who configures `synthesis.rateLimit` to contain costs will not have that limit applied to widget HTML generation traffic. If the widget path is called frequently (e.g. in an agent loop), it can silently exhaust the same underlying quota the synthesis limiter was meant to protect. Consider at minimum a startup log line noting that widget traffic is exempt from the synthesis rate limit when `synthesis.rateLimit` is non-null.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +84,175 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	"codex",
+	"opencode",
+]);
+
+// Compile-time check: if a new remote provider is added to the union but
+// omitted from the set above, this line produces a type error.
+const _exhaustiveCheck: RemoteProvider[] = [...RATE_LIMIT_PROVIDERS] as unknown as RemoteProvider[];
+void _exhaustiveCheck;
+
+function shouldRateLimit(providerName: string): boolean {
```
PR-Reviewer-Ant, line 184: **Mutable write to `readonly` fields via index access.**

```ts
const clean: ProviderRateLimitConfig = {};
for (const [k, v] of Object.entries(config)) {
    if (v !== undefined && RATE_LIMIT_CONFIG_KEYS.has(k)) {
        clean[k as keyof ProviderRateLimitConfig] = v;  // readonly violation
    }
}
```

`ProviderRateLimitConfig` declares all fields as `readonly`. Writing to them via a computed union key is a TypeScript type violation in strict mode. At runtime this is fine (JS doesn't enforce `readonly`), but it could silently allow future callers to mutate what they expect to be immutable. Consider declaring `clean` as `Partial<Record<keyof ProviderRateLimitConfig, number>>` or using `Object.assign` with a non-readonly intermediate.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -248,6 +247,20 @@ function failJob(db: WriteDb, jobId: string, error: string, attempts: number, ma
...
```
PR-Reviewer-Ant, line 264: **`logger.error` side-effect inside a `withWriteTx` callback.**

```ts
    if (nonRetryable) {
        logger.error("pipeline", `Dead-lettering job ${job.id}…`, { … });
        deadLetterJob(db, job.id, msg);
    }
});
```

If the `withWriteTx` callback rolls back (DB error after the log but before `deadLetterJob` commits), the error log will have been emitted but the job will not be in `status='dead'`. An operator investigating the log would find a dead-letter event with no corresponding DB row. Move the `logger.error` call outside the transaction to keep log state consistent with DB state.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `999ccd6e`

[Automated bot review — pr-reviewer] This PR implements a token-bucket rate limiter for remote LLM providers. The core algorithm, config parsing, dead-lettering path for the extraction worker, and documentation are solid. One compile-time guard is broken (non-functional exhaustiveness check with a misleading comment), the synthesis dead-letter path is untested end-to-end, and the widget provider intentionally bypasses rate limiting in a way operators may not anticipate.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The broken exhaustiveness check and the `clean` readonly mutation are directly provable from the diff. The synthesis dead-letter path requires the full summary-worker catch block (only partially visible in the diff) — specifically, whether `resolveFailedSummaryJobStatus` is called with the `terminal` flag and routes to `deadLetterJob`. That function body is not in the changed lines, so I cannot confirm the end-to-end path is correct without that context.

packages/daemon/src/pipeline/worker.ts
```diff
@@ -248,6 +247,20 @@ function failJob(db: WriteDb, jobId: string, error: string, attempts: number, ma
...
+// normal retry loop can increment attempts. This keeps the attempts value
+// consistent with the lease/claim logic, not the dead-letter classification.
+function deadLetterJob(db: WriteDb, jobId: string, error: string): void {
+	const now = new Date().toISOString();
+	db.prepare(
+		`UPDATE memory_jobs
+		 SET status = 'dead', error = ?, failed_at = ?, updated_at = ?
+		 WHERE id = ?`,
+	).run(error, now, now, jobId);
+}
```
PR-Reviewer-Ant, line 262: `deadLetterJob` sets `status='dead'`, `error`, `failed_at`, and `updated_at`, but does NOT clear `leasedAt` (or any other lease columns the original claim step may have set). The existing `failJob` has the same omission, so stale-lease recovery almost certainly filters on `status = 'leased'` — but this is not verifiable from the diff alone. If `recoverStaleLeases` queries on `leasedAt IS NOT NULL` without also checking `status = 'leased'`, a dead-lettered job with a live `leasedAt` timestamp could be re-queued before its lease expires, reversing the non-retryable classification. Worth confirming the stale-lease query explicitly excludes `status = 'dead'` (and `'failed'`, `'completed'`).

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -100,8 +101,9 @@ export function resolveSummaryHeadingDate(job: Pick<SummaryJobRow, "ended_at" |
 	return basis.slice(0, 10);
 }
```
PR-Reviewer-Ant, line 103: `isTerminalSummaryJobError(input: string | Error)` has an invisible asymmetry: when `input` is a `string`, `input instanceof RateLimitExceededError` is always `false`, so a stringified rate-limit error (`err.message` = `"Rate limit exceeded: 200/hr for …"`) is never classified as terminal and would be retried. The current call site in `startSummaryWorker` correctly passes the `Error` object when available:
```ts
const terminal = isTerminalSummaryJobError(e instanceof Error ? e : String(e));
```
But the function signature makes the string path look equivalent. Any future caller that passes `e.message` or `String(e)` directly will silently allow rate-limit errors to retry indefinitely. Consider documenting the asymmetry in a JSDoc comment on the function, or adding a string-prefix check for the rate-limit error message as a safety net.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -78,6 +84,172 @@ const subprocessSemaphore = new SubprocessSemaphore(
...
+	const burstSize = config.burstSize ?? DEFAULT_PROVIDER_RATE_LIMIT.burstSize;
+	const waitTimeoutMs = config.waitTimeoutMs ?? DEFAULT_PROVIDER_RATE_LIMIT.waitTimeoutMs;
+	if (maxCallsPerHour <= 0 || burstSize <= 0) return provider;
+
+	if (!shouldRateLimit(provider.name)) {
+		logger.warn(
+			"pipeline",
+			`rateLimit config ignored for provider "${provider.name}" — only remote/paid providers are throttled`,
+			{
+				provider: provider.name,
```
PR-Reviewer-Ant, line 202: `Object.keys(config).length === 0` treats an object with all-undefined fields (e.g. `{ maxCallsPerHour: undefined }`) as non-empty (length 1), so such a config would proceed to enable rate limiting with defaults rather than being treated as disabled. In practice, `parseRateLimitConfig` in `memory-config.ts` never produces a config object with `undefined`-valued keys (it either returns `undefined` or a fully-resolved object), so the daemon's own call sites are safe. The concern is direct API callers. A quick guard would be:
```ts
const defined = Object.fromEntries(Object.entries(config).filter(([, v]) => v !== undefined));
if (Object.keys(defined).length === 0) return provider;
```
Low urgency given actual call paths, but worth closing before the SDK is published.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `aa40c49a`

Automated bot review (pr-reviewer). Token-bucket rate limiter for remote LLM providers. Implementation is well-structured across 9 review rounds; found two warning-level issues worth discussing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The logic in provider.ts, worker.ts, summary-worker.ts, and memory-config.ts is fully visible in the diff. The warning-level concern about deadLetterJob leaving lease columns set requires reading packages/daemon/src/pipeline/stale-leases.ts (not included in the diff) to confirm whether recoverStaleLeases filters on status='leased' before attempting re-queue. Without that file, the risk cannot be fully ruled out.

PR #29, opened 2026-02-27
fix(tasks): stream live run output and fix opencode scheduled execution

aaf2tbz:
## Summary
- add a task SSE endpoint (`/api/tasks/:id/stream`) and task-scoped event bus so run start/output/completion events are emitted for both scheduled and manual runs
- add replay buffering for active runs so opening the task detail late still shows in-progress output instead of an empty running card
- fix OpenCode scheduler invocation to use non-interactive execution (`opencode run --format json "<prompt>"`) and normalize streamed JSON/ANSI output into readable live logs in the dashboard

## How It Works
- backend now streams incremental stdout/stderr chunks from `spawnTask` hooks instead of waiting for process exit
- daemon publishes run lifecycle events and keeps a bounded in-memory buffer per active run for late subscribers
- dashboard subscribes to task SSE, maintains a resilient connection with reconnects, upserts live run rows without reopening the detail panel, and parses OpenCode JSON events (`text`, `tool_use`) for readable output

## Safety / Scope
- changes are limited to scheduled-task streaming and display paths in daemon + dashboard; no auth, memory, secrets, connector, or infra paths were modified
- lockfile was intentionally excluded from this PR to avoid unrelated dependency churn for the team

## Validation
- `cd packages/daemon && bun run build`
- `cd packages/cli/dashboard && bun run build`
- manually verified `/api/tasks/:id/stream` returns `connected`, `run-started`, and `run-output` events, including replay after late subscription

PR #474, opened 2026-04-07
refactor: align recall surfaces and validate prospective hint retrieval

NicholaiVogel:
## Summary

This PR cleans up recall as a product surface, not just as an internal search path.

Before this branch, recall was technically capable but inconsistent depending on where you touched it. Prompt-submit injected noisy recall blocks, explicit recall consumers flattened useful metadata in different ways, `/api/hooks/recall` had contract drift, and the SDK types/docs had fallen out of sync with what the daemon actually returned. The result was one strong recall engine underneath several surfaces that did not behave like they belonged to the same system.

This PR fixes that in three layers.

First, it makes prompt-submit smaller and calmer. The hot path now injects a lightweight source-aware memory block instead of a more debug-like payload. The point is to keep prompt-submit fast and useful without turning it into a full recall product.

Second, it tightens the explicit recall contract around the existing TypeScript engine. `/api/memory/recall` remains the source of truth, but CLI, MCP, OpenClaw, hook recall, and SDK surfaces are now much more aligned with it. That includes preserving score/source/type/provenance, separating supporting context from primary hits, normalizing no-hit behavior, and removing drift like `/api/hooks/recall` silently treating `project` as `scope`.

Third, it adds research work to validate the next retrieval step instead of guessing. The branch includes a vault-based recall eval over a real Obsidian corpus and a planning spec for semantic prospective hints. The eval result was useful: the current note-level hybrid retrieval outperformed both a lexical-first lane and a plain chunked hybrid RAG baseline, prospective hint FTS helped further, and semantic retrieval over hints improved again on the full-vault run. That does not implement semantic hint embeddings yet, but it does justify planning them properly instead of treating them as speculation.

Taken together, this PR makes recall feel more coherent across Signet, cleans up the contracts around it, and leaves the branch with a clearer evidence-backed direction for what should come next.

## Changes

- Simplified prompt-submit recall injection so it stays small, source-aware, and low-noise
- Tightened explicit recall contracts around `hybridRecall()` and `/api/memory/recall`
- Normalized `/api/hooks/recall` as a thin wrapper and fixed `project` contract drift
- Improved recall rendering for:
  - CLI `signet recall`
  - MCP `memory_search`
  - OpenClaw `/recall`
- Aligned SDK recall request/response types and examples with daemon reality
- Added an Obsidian-vault retrieval eval script and research memo
- Added a planning spec for semantic prospective hints

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs/specs/research, eval script

## Screenshots

N/A, no UI changes.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

Compatibility notes for this branch:

- `/api/hooks/recall` now returns the richer structured response shape, but still includes legacy `memories` and `count` aliases during the transition period.
- SDK callers can use `hookRemember()` / `hookRecall()` as the primary names, while deprecated `rememberHook()` / `recallHook()` aliases remain in place for compatibility.
- MCP `memory_search` keeps `min_score` as a deprecated compatibility alias for the older server-side `importance_min` behavior; the new client-side retrieval score filter is `score_min`.
- `signet recall --json` now prints the full recall response object instead of only the `results` array. CLI docs were updated to match.
- Prompt-submit now injects a lightweight `## Relevant Memory` block and no longer emits the old `[signet:recall | ...]` telemetry header.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Targeted verification run for touched areas:

- `bun test packages/daemon/src/hooks.prompt-submit.test.ts packages/daemon/src/hooks-recall.test.ts packages/daemon/src/memory-search.test.ts packages/daemon/src/mcp/tools.test.ts`
- `bun test packages/cli/src/commands/memory.test.ts`
- `bun test packages/connector-openclaw/test/index.test.ts`
- `bun test packages/sdk/src/__tests__/client.test.ts packages/sdk/src/__tests__/helpers.test.ts`
- `bun run --cwd packages/sdk build`
- `bun run build:core`
- `cargo test -p signet-daemon prompt_submit_formats_temporal_fallback_as_lightweight_recall_block -- --nocapture`
- targeted `biome check` on touched files

## AI disclosure

## Notes

- No schema migrations are included in this PR.
- The new semantic prospective hints work is planning-only in this branch, not implemented.
- `bun scripts/spec-deps-check.ts` still reports pre-existing unrelated registry drift:
  - `recall-confidence-gate`
  - missing `references/Engram`
- Vault eval artifacts remain out of git; the committed research output is the memo plus the reusable eval script.


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor: align recall surfaces and validate prospective hint retrieval` (commit `ee475ec3`) now and I'll follow up shortly with feedback.

packages/daemon/src/hooks-recall.test.ts
```diff
@@ -23,16 +30,24 @@ describe("/api/hooks/recall", () => {
...
+		getDbAccessor = () => dbAccessor.getDbAccessor();
+		const tracker = await import("./session-tracker");
+		bypassSession = tracker.bypassSession;
+
 		const daemon = await import("./daemon");
 		app = daemon.app;
 	});
 
 	afterAll(() => {
 		if (prev === undefined) {
```
PR-Reviewer-Ant, line 45: `process.env.SIGNET_PATH = undefined` sets the env var to the **string** `"undefined"`, it does not remove it. The original code used `delete process.env.SIGNET_PATH`, which is correct. This leaks `SIGNET_PATH="undefined"` into any test that runs after this suite, potentially breaking path resolution in those tests. Should be reverted to `delete process.env.SIGNET_PATH`.

packages/daemon/src/routes/hooks-routes.ts
```diff
@@ -383,7 +412,7 @@ function registerRemember(app: Hono): void {
 function registerRecall(app: Hono): void {
```
PR-Reviewer-Ant, line 428: Breaking response contract change with no migration note. The old `/api/hooks/recall` shape was `{ memories: [], count: 0 }` (internal) and `{ memories: [], count: 0, bypassed: true }` (bypassed). The new shape is `{ results, query, method, meta, [bypassed|internal] }`. Any deployed connector, harness script, or integration that reads `body.memories` or `body.count` from this endpoint will silently receive `undefined` after upgrade. The OpenClaw connector source was updated in this PR, but other installed connector binaries and third-party integrations won't be. The PR migration notes section is blank. At minimum this needs an explicit breaking-change callout and a compatibility shim (`count: results.length`, `memories: results`) in the transition period, or a version bump signal.

docs/SDK.md
```diff
@@ -882,42 +909,50 @@ await signet.sessionStart({
...
 **`sessionEnd(opts)`** — Extract memories at session end.
 
 ```typescript
 await signet.sessionEnd({
-  session_key: "sess-abc-123",
+  sessionKey: "sess-abc-123",
   project: "/home/user/myapp",
   summary: "Implemented JWT authentication with refresh tokens",
 });
 ```
```
PR-Reviewer-Ant, line 926: `rememberHook` → `hookRemember` and `recallHook` → `hookRecall` are breaking renames to the public SDK API. The migration notes section is unchecked/blank. Existing callers of `signet.recallHook(...)` or `signet.rememberHook(...)` will get a runtime `TypeError: signet.recallHook is not a function`. These should either be kept as deprecated aliases pointing to the new names, or explicitly called out as breaking with a semver minor bump.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -3068,6 +3071,66 @@ mod tests {
...
+                last_assistant_message: None,
+                session_key: Some("sess-prompt-structured".to_string()),
+                transcript: None,
+                transcript_path: None,
+                runtime_path: None,
+            }),
+        )
+        .await;
+
+        assert_eq!(resp.status(), StatusCode::OK);
```
PR-Reviewer-Ant, line 3118: The test asserts `inject.contains("[node node-1]") || inject.contains("[thread node-1]")`. The new `build_prompt_recall_inject` in Rust simply forwards the `sources` slice as-is — it does not add `[node ...]` or `[thread ...]` labels itself. Whether the assertion holds depends entirely on how `picked` items are formatted in the temporal fallback path upstream of the `build_prompt_recall_inject` call, which is not shown in this diff. If the Rust temporal fallback was not updated to emit `[thread <id>]`-prefixed lines (mirroring the TypeScript change at `hooks.ts:449`), this test assertion will fail or will pass vacuously against stale data.

docs/specs/planning/explicit-recall-surface-alignment.md
```diff
@@ -0,0 +1,250 @@
+---
+title: "Explicit Recall Surface Alignment"
+id: explicit-recall-surface-alignment
+status: planning
```
PR-Reviewer-Ant: Spec is created at `status: planning` while the work it describes is implemented in this same PR. Per AGENTS.md: *"Do not implement major feature work from planning; it must be approved first."* This refactor changes multiple API contracts (`/api/hooks/recall` response shape, SDK method names, prompt-submit inject format, recall response `meta` field), which puts it into spec-governed territory rather than a narrow in-contract refactor. The spec should be moved to `approved` before this merges, or the PR needs to demonstrate the changes are strictly in-contract (the contract changes above make that difficult to argue).

docs/research/technical/RESEARCH-OBSIDIAN-VAULT-RECALL-EVAL.md
```diff
@@ -0,0 +1,381 @@
...
+direction: keep the semantic machinery, keep prospective hints, and
+invest in a more curated memory surface rather than retreating to a
+simpler retrieval stack.
+
+## Trigger
+
+We wanted a fast, isolated retrieval-quality test without dragging in the
+full benchmark framework. The goal was not end-to-end answer grading. The
+goal was a 1:1 retrieval-method comparison on a real corpus:
+
```
PR-Reviewer-Ant, line 35: The document names the vault owner by first name and includes the full local path `/mnt/work/obsidian-vault`. It also describes personal note patterns in the failure analysis section (diary/fleeting note titles, journal semantics). Even in a private repo, this is PII committed to git history permanently. Consider anonymizing or confirming the author explicitly consents to this data being in the repo.

packages/daemon/test/hooks.test.ts
```diff
@@ -2714,19 +2644,6 @@ describe("schema", () => {
```
PR-Reviewer-Ant, line 2657: The `busy_timeout is set by openDb` test was specifically a regression guard for a SQLite concurrency issue — it verified that `openDb` set `busy_timeout`. This guard is removed here with no replacement. Per AGENTS.md: *"A bug fix is not complete until it adds at least one durable prevention mechanism."* While deleting `handleRecall()` is not a fix per se, removing its associated concurrency regression test without any replacement (e.g., a direct `openDb` test or a comment explaining why it's no longer needed) leaves the `busy_timeout` guarantee unenforced.

packages/daemon/src/hooks.ts
```diff
@@ -479,6 +474,16 @@ export function selectWithBudget<T extends { content: string }>(rows: ReadonlyAr
 	return selected;
```
PR-Reviewer-Ant, line 481: `buildPromptRecallInject` and the identical `build_prompt_recall_inject` in Rust (`hooks.rs:806`) are now two independent implementations of the same formatting logic. They diverge in one observable way: the TypeScript version trims and re-adds a trailing newline (`trimEnd()\n`), while the Rust version also does so. This is fine now, but the absence of a shared contract test means they can drift silently. Worth a cross-reference comment at minimum.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `ee475ec3`

[Automated Review — pr-reviewer bot] PR #474 introduces several breaking contract changes without migration guidance, a test environment pollution bug, and a spec-sequencing violation. The implementation is largely coherent and well-tested, but the breaking changes need explicit callout or guards before merge.

**Confidence:** High [sufficient_diff_evidence, missing_cross_module_context] - The /api/hooks/recall contract change, SDK method renames, and process.env bug are all directly provable from the diff. The Rust temporal-fallback label assertion is uncertain because the diff does not show how `picked` items are constructed in the Rust temporal fallback path — only that `build_prompt_recall_inject` is now called with them — so the assert at hooks.rs:3118 may or may not pass depending on unshown upstream formatting code.

packages/daemon/src/hooks-recall.test.ts
NicholaiVogel: fixed. i switched the cleanup to Reflect.deleteProperty(...) so the env var is actually removed without leaving SIGNET_PATH="undefined" behind, and kept the regression coverage in place.

packages/daemon/src/routes/hooks-routes.ts
NicholaiVogel: good catch. i restored a compatibility shim on /api/hooks/recall so the route now returns both the new structured shape and the legacy aliases (`memories`, `count`) during the transition period. i also documented the compatibility aliases in docs/API.md.

docs/SDK.md
NicholaiVogel: fixed. i added deprecated `rememberHook()` and `recallHook()` aliases that forward to the new method names, added SDK regression coverage for the aliases, and documented the compatibility path in docs/SDK.md.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
NicholaiVogel: fixed. i made the rust assertion less brittle so it now checks for the actual sample recall content instead of depending on a specific `[node ...]` / `[thread ...]` prefix emitted upstream.

docs/specs/planning/explicit-recall-surface-alignment.md
NicholaiVogel: agreed. i moved `explicit-recall-surface-alignment` from `planning` to `approved` and updated INDEX/dependencies so the implemented contract work is no longer sitting behind a planning-tier spec.

docs/research/technical/RESEARCH-OBSIDIAN-VAULT-RECALL-EVAL.md
NicholaiVogel: fixed. i anonymized the memo: removed the owner name, removed the explicit local vault path, generalized the example note reference, and softened the personal-note wording so the repo keeps the finding without baking PII into history.

packages/daemon/test/hooks.test.ts
NicholaiVogel: fixed. i added a direct replacement regression guard in `packages/daemon/test/hooks.test.ts` that opens the db accessor path and asserts `PRAGMA busy_timeout` is set to `5000`, so the concurrency guarantee stays covered after the helper removal.

packages/daemon/src/hooks.ts
NicholaiVogel: fixed. i added cross-reference comments above both the TypeScript and Rust prompt recall inject formatters so the parity link is explicit in both directions. still not a perfect shared contract test, but at least it makes drift harder to miss in follow-up edits.

packages/daemon/src/mcp/tools.ts
```diff
@@ -524,27 +622,56 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
...
+					expand: expand === true ? true : undefined,
 				},
 			});
 
 			if (!result.ok) {
 				return errorResult(`Search failed: ${result.error}`);
 			}
-			return textResult(result.data);
+			return textResult(formatRecallToolResult(applyRecallToolMinScore(result.data, min_score)));
 		},
```
PR-Reviewer-Ant, line 663: Silent behavioral break for existing MCP callers: the old `memory_search` implementation forwarded `min_score` to the daemon as `importance_min` (server-side importance threshold). The new implementation keeps `importance_min` as a separate server-side parameter and applies `min_score` client-side as a relevance-score filter via `applyRecallToolMinScore`. These filter completely different things — importance (a stored memory attribute, 0–1) vs retrieval score (a query-dependent rank). Any caller that previously passed `min_score: 0.5` expecting to exclude low-importance memories now gets recall-score filtering instead, with no error or warning. This PR is labeled `refactor` (no behavior change), but this is a observable behavior change on the MCP surface. It should either be labeled `fix`, or the old mapping should be preserved as a fallback with a deprecation note.

docs/SDK.md
```diff
@@ -882,44 +909,58 @@ await signet.sessionStart({
...
```
PR-Reviewer-Ant, line 926: The docs claim `rememberHook(opts)` and `recallHook(opts)` remain as deprecated compatibility aliases after the rename to `hookRemember`/`hookRecall`. Inspection of `packages/sdk/src/client-p2.ts` shows the SDK already uses `hookRemember` and `hookRecall` as primary methods, but no `rememberHook` or `recallHook` deprecated aliases are present. If the old method names were ever callable by SDK consumers, removing them without an alias is a breaking API change. If they never existed as the public API (e.g. the rename happened entirely within this branch), the deprecation claim in the docs is misleading and should be removed. The `hookRecall` signature found in the SDK source (`query`, `limit`, `type`, `tags` only) also appears narrower than the richer contract described in the updated docs — if the SDK source wasn't updated to match, callers following the new docs will pass unrecognized fields silently.

packages/daemon/src/routes/hooks-routes.ts
```diff
@@ -111,6 +111,51 @@ function checkBypass(body?: { sessionKey?: string; sessionId?: string }): boolea
 	return isSessionBypassed(key);
```
PR-Reviewer-Ant, line 111: The hooks-routes.ts diff is truncated, so the full new `/api/hooks/recall` handler body is not visible. The `RecallRequest` interface in hooks.ts was extended with `keywordQuery`, `type`, `tags`, `who`, `since`, `until`, and `expand` — all documented in API.md as accepted fields. The project-filtering test verifies only `project` forwarding. If the new handler body does not pass these fields through to `hybridRecall()`, the API docs describe a contract that the implementation doesn't fulfill. This needs a test for at least one of the new fields (e.g. `type` or `since`) to confirm end-to-end forwarding, consistent with the PR checklist requirement for regression tests.

packages/cli/src/commands/memory.ts
```diff
@@ -15,6 +15,108 @@ interface MemoryDeps {
...
+		"totalReturned" in raw && typeof raw.totalReturned === "number" ? raw.totalReturned : fallbackCount;
+	const hasSupplementary = "hasSupplementary" in raw && raw.hasSupplementary === true;
+	const noHits = "noHits" in raw ? raw.noHits === true : totalReturned === 0;
+	return { totalReturned, hasSupplementary, noHits };
+}
+
+export function parseRecallResult(raw: unknown): ParsedRecallResult {
+	const result = typeof raw === "object" && raw !== null ? raw : {};
+	const rows = Array.isArray(result.results) ? (result.results as RecallRow[]) : [];
+	const meta = parseRecallMeta("meta" in result ? result.meta : undefined, rows.length);
```
PR-Reviewer-Ant, line 63: `applyRecallMinScore` here and `applyRecallToolMinScore` in `packages/daemon/src/mcp/tools.ts` are nearly identical — same input shape, same filtering predicate, same meta recomputation. AGENTS.md explicitly flags this: "Duplicate logic is a smell. Extract shared code where it clarifies." These belong in `@signet/core` or a shared utility so a future change to score filtering logic (e.g. handling `supplementary` rows differently) only needs to be made once.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -806,6 +806,29 @@ fn format_metadata_header() -> String {
...
+        "if you need deeper history, use /recall or memory_search. if you learn something durable, save it with /remember or memory_store."
+            .to_string(),
+    );
+
+    format!("{}\n", parts.join("\n").trim_end())
+}
+
 pub async fn prompt_submit(
     State(state): State<Arc<AppState>>,
     headers: HeaderMap,
```
PR-Reviewer-Ant, line 832: Parity gap for the empty-sources case: the Rust `build_prompt_recall_inject` pushes `sources` + trailing empty string only when `!sources.is_empty()`, producing `header\n\n## Relevant Memory\n\nfooter\n` for empty input. The TypeScript version always pushes `parts.push("")` unconditionally, producing `header\n\n## Relevant Memory\n\n\nfooter\n` (one extra blank line). In practice both paths should only be called with non-empty sources, so this is low risk, but it contradicts the cross-language alignment comment on both functions.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f2d09b90`

[Automated Review — pr-reviewer] PR #474 makes meaningful progress on recall surface alignment. Three issues surfaced: a silent behavioral break in the MCP `memory_search` tool (`min_score` semantics changed), missing SDK deprecated aliases the docs promise, and a logic-duplication violation across packages.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The MCP min_score change is directly provable from the diff (old `importance_min: min_score` mapping vs new client-side filtering). The SDK alias gap is supported by codebase inspection of packages/sdk/src/client-p2.ts, which shows hookRemember/hookRecall already exist as primary methods but no rememberHook/recallHook deprecated aliases. The hooks-routes.ts diff is truncated, so forwarding of the new RecallRequest fields (keywordQuery, type, tags, who, since, until, expand) to hybridRecall cannot be fully verified from the diff alone — the project-filter test passes, but the other fields have no coverage.

packages/daemon/src/mcp/tools.ts
NicholaiVogel: good catch. i restored MCP compatibility by keeping `min_score` as a deprecated alias for the old server-side `importance_min` behavior, and introduced `score_min` for the new client-side recall-score filter. i also added regression coverage plus docs updates so the semantics are explicit instead of silently changing under callers.

docs/SDK.md
NicholaiVogel: this one was already fixed in the follow-up, but the review UI was still looking at the older diff. `packages/sdk/src/client-p2.ts` now includes deprecated `rememberHook()` / `recallHook()` aliases, and `hookRecall()` carries the richer filter surface documented in SDK.md. there is also a regression test covering the deprecated aliases.

packages/daemon/src/routes/hooks-routes.ts
NicholaiVogel: fixed. the handler was already forwarding the richer fields, but the test coverage was too thin. i added an end-to-end hook recall regression test that exercises `type` filtering so the documented pass-through contract is now actually enforced.

packages/cli/src/commands/memory.ts
NicholaiVogel: agreed. i extracted the client-side score-threshold logic into a shared `applyRecallScoreThreshold()` helper in `@signet/core`, then wired both CLI recall and MCP recall through that shared implementation so the filter semantics only live in one place now.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
NicholaiVogel: fixed. i made the Rust formatter unconditionally add the same spacer line as the TypeScript formatter, so the empty-sources case now matches the TS implementation instead of drifting by one blank line.

packages/cli/src/commands/memory.ts
```diff
@@ -94,41 +192,26 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 			}
```
PR-Reviewer-Ant, line 188: `signet recall --json` previously printed the results array (`JSON.stringify(rows, null, 2)`). It now prints the full filtered response object (`{ results, query, method, meta }`). Any caller or script that relied on the array output — e.g. `signet recall '...' --json | jq '.[0].content'` — silently breaks. CLI.md was updated to read 'Print the recall response as JSON' but the shape change is not called out in the Migration Notes section of the PR description. Because the commit type is `refactor`, which by repo convention should not change observable behavior, and this is a consumer-visible contract change, a migration note (or explicit callout in the PR description) is warranted.

packages/cli/src/commands/memory.ts
```diff
@@ -15,6 +16,92 @@ interface MemoryDeps {
...
+	if (typeof raw !== "object" || raw === null || Array.isArray(raw)) {
+		return {
+			totalReturned: fallbackCount,
+			hasSupplementary: false,
+			noHits: fallbackCount === 0,
+		};
+	}
+	const totalReturned =
+		"totalReturned" in raw && typeof raw.totalReturned === "number" ? raw.totalReturned : fallbackCount;
+	const hasSupplementary = "hasSupplementary" in raw && raw.hasSupplementary === true;
```
PR-Reviewer-Ant, line 55: `result.results` is accessed directly via property notation, but every other field on `result` uses the `in`-guard pattern (`"meta" in result`, `"query" in result`, `"method" in result`). After the ternary on line 53, `result` is typed as `object | {}`, and TypeScript in strict mode does not allow `.results` on either. The other fields are correctly guarded with `in` before access. If typecheck is passing here it is either because the project tsconfig is non-strict on index access, or because `{}` in this tsconfig widens to `Record<string, unknown>` implicitly. The inconsistency is a latent type-safety hole: if someone changes the tsconfig or strict settings, this line would newly error while the guarded accesses would continue to compile. The fix is to add `"results" in result &&` before the `Array.isArray` call, matching the pattern used for the other three fields.

docs/specs/approved/explicit-recall-surface-alignment.md
```diff
@@ -0,0 +1,250 @@
+---
+title: "Explicit Recall Surface Alignment"
+id: explicit-recall-surface-alignment
+status: approved
+informed_by: []
+section: "Runtime"
```
PR-Reviewer-Ant, line 8: `informed_by: []` is empty. AGENTS.md states 'No spec without research.' The `semantic-prospective-hints` planning spec correctly cites the Obsidian eval memo. This spec does not cite any research input even though it is in `approved` status. For a cleanup/consolidation spec this may be defensible, but the prior review that moved the spec from `planning` → `approved` did not address the empty `informed_by`. Adding a reference to the Obsidian eval (which motivated the broader recall cleanup) or adding a brief rationale note would close this gap.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c8de11a9`

[Automated Review — pr-reviewer] This PR makes three kinds of changes: tightens prompt-submit injection format, consolidates explicit recall around hybridRecall()/`/api/memory/recall`, and adds eval research + planning specs. Prior reviews (CHANGES_REQUESTED then COMMENTED) are marked resolved. Three concerns remain after checking the current HEAD diff that were not covered by prior feedback.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The CLI output-format change and the property-access inconsistency are directly visible in the diff. Confidence is limited to 'medium' because the diff was truncated before the full packages/sdk/ source, so SDK deprecated-alias correctness (flagged and resolved in prior reviews) cannot be independently re-verified here; and the packages/daemon/src/daemon.ts hook-recall handler body is not fully shown.

packages/cli/src/commands/memory.ts
NicholaiVogel: fair point. i added an explicit compatibility note to the PR description calling out that `signet recall --json` now prints the full recall response object instead of just the `results` array, so the behavior change is at least documented for downstream scripts.

packages/cli/src/commands/memory.ts
NicholaiVogel: fixed. i tightened `parseRecallResult()` to guard `results` with the same `"results" in result` pattern used for `meta`, `query`, and `method`, so it stays type-safe even if stricter compiler settings land later.

docs/specs/approved/explicit-recall-surface-alignment.md
NicholaiVogel: good catch. i filled in `informed_by` on the approved explicit recall alignment spec so it now cites the recall eval memo plus the broader memory benchmark comparison research instead of sitting as an approved spec with no research lineage.

packages/daemon/src/hooks.ts
```diff
@@ -3339,97 +3348,6 @@ export function handleRemember(req: RememberRequest): RememberResponse {
 	}
```
PR-Reviewer-Ant, line 3406: handleRecall() is removed here, but the diff is truncated before daemon.ts changes are shown. The /api/hooks/recall route must have been rewired to call hybridRecall() directly — the spec and tests strongly suggest this is done — but it cannot be confirmed from the visible diff. If daemon.ts was not updated to replace the handleRecall() call site, the route would silently break (or fall through to an undefined symbol). The hooks-recall.test.ts tests importing the real app give reasonable coverage, but the implementation path is un-reviewable here.

packages/daemon/src/mcp/tools.ts
```diff
@@ -195,6 +226,58 @@ function textResult(value: unknown): { content: Array<{ type: "text"; text: stri
 	};
```
PR-Reviewer-Ant, line 222: formatRecallToolResult() here, formatRecallMessage() in connector-openclaw/src/index.ts, and formatRecallRows() in packages/cli/src/commands/memory.ts all independently implement the same primary/supporting row-classification logic (filter supplementary !== true, filter supplementary === true, section headers). This directly matches the AGENTS.md 'Duplication / parity drift' failure mode. The shared logic — partitioning rows into primary and supporting, and the no-hit check — could be extracted to @signet/core alongside applyRecallScoreThreshold, leaving only the rendering differences in each consumer. As written, a future change to the classification rule (e.g. a new supplementary variant) requires three coordinated edits.

packages/cli/src/commands/memory.ts
PR-Reviewer-Ant, line 55: parseRecallResult, applyRecallMinScore, and formatRecallRows are exported from the CLI command module solely to make them accessible in memory.test.ts. Exporting implementation details from a command module creates unintended public API surface. The test for applyRecallMinScore is better placed in packages/core where applyRecallScoreThreshold lives; parseRecallResult and formatRecallRows could be tested indirectly through the registerMemoryCommands integration tests that already exist.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `77a97082`

[Automated Review — pr-reviewer] This PR aligns recall surfaces and cleans up contracts across CLI, MCP, OpenClaw, and daemon-rs. Overall architecture intent is sound and test coverage for key behaviors (project filtering, bypass shape, type filtering) is good. Three issues worth discussion: the diff is truncated before daemon.ts changes that are load-bearing for the handleRecall removal; three parallel recall formatters repeat the same primary/supporting classification logic in violation of the repo's parity-drift guideline; and several helpers are exported from memory.ts solely for test access.

**Confidence:** Medium [missing_cross_module_context] - The diff ends with '... [diff truncated]' mid-way through tools.ts and before any daemon.ts changes are shown. The most load-bearing parts of this PR — how /api/hooks/recall is rewired after handleRecall() removal, and how score_min is applied in the MCP tool handler after daemonFetch — are not visible in the diff. The tests in hooks-recall.test.ts and tools.test.ts assert the correct behavior and appear to hit the real daemon, so the implementation is likely correct, but I cannot review it directly. Confidence in the reviewed portion is high; confidence in the un-shown portion is inferred from tests only.

packages/cli/src/commands/memory.ts
```diff
@@ -15,6 +16,92 @@ interface MemoryDeps {
...
+	const noHits = "noHits" in raw ? raw.noHits === true : totalReturned === 0;
```
PR-Reviewer-Ant, line 56: applyRecallMinScore is a trivial one-liner re-export of applyRecallScoreThreshold under a different name: `return applyRecallScoreThreshold(raw, minScore)`. It adds a second name for the same function with no transformation. The action handler on line 195 calls applyRecallMinScore while the import on line 1 already brings applyRecallScoreThreshold into scope. Removing the wrapper and calling applyRecallScoreThreshold directly would reduce the name surface and eliminate the redundant export.

packages/daemon/src/hooks.ts
```diff
@@ -2468,12 +2475,14 @@ export async function handleUserPromptSubmit(
...
-			lines.push(`(+${omitted} more not shown — raise memory.guardrails.contextBudgetChars to include)`);
+			lines.push(
+				`- [note] ${omitted} additional ${omitted === 1 ? "match was" : "matches were"} omitted to keep this lightweight.`,
+			);
 		}
-		let inject = `${metadataHeader}\n[signet:recall | query="${queryTerms}" | results=${selected.length} | engine=hybrid]\n${lines.join("\n")}`;
+		let inject = buildPromptRecallInject(metadataHeader, lines);
 
 		// Append agent feedback request if enabled and there are injected memories
 		const selectedIds = selected.map((s) => s.id);
```
PR-Reviewer-Ant, line 2546: The omitted-memories line changed from `(+N more not shown — raise memory.guardrails.contextBudgetChars to include)` to `- [note] N additional match(es) were omitted to keep this lightweight.` The new form starts with `- ` which gives it the same visual shape as injected memory lines in the ## Relevant Memory block. An LLM consuming this block may attempt to treat it as a retrievable memory item. The parenthesized form was unambiguously a meta-comment, not a memory. Consider keeping the note outside the list (no leading `- `) or using a distinct prefix like `[signet:note]` to prevent misinterpretation.

packages/daemon/src/hooks.ts
NicholaiVogel: repo check: the live hook route is in `packages/daemon/src/routes/hooks-routes.ts`, not `daemon.ts` anymore after the route refactor. `/api/hooks/recall` is calling `hybridRecall()` directly there, and `hooks-recall.test.ts` is exercising the real app path rather than a helper-level mock. so this one is covered even though the original diff view was lousy.

packages/daemon/src/mcp/tools.ts
NicholaiVogel: i reduced the duplication further by extracting the shared primary/supporting partitioning logic into `partitionRecallRows()` in `@signet/core`, then wiring both CLI and MCP through it. openclaw is still a special case because it generates a standalone legacy handler.js string rather than importing runtime TS helpers, but at least the two actively maintained TypeScript surfaces no longer carry separate copies of the classification rule.

packages/cli/src/commands/memory.ts
NicholaiVogel: fixed. i removed the exported test-only helpers from the CLI command module, moved the score-threshold test down into `@signet/core`, and kept the CLI behavior covered through the existing command-level integration tests instead of widening the command module API surface.

packages/cli/src/commands/memory.ts
NicholaiVogel: fixed. i dropped the trivial `applyRecallMinScore()` wrapper and now call `applyRecallScoreThreshold()` directly from the CLI action path.

packages/daemon/src/hooks.ts
NicholaiVogel: good catch. i changed the omitted-results line to a non-memory meta prefix (`[signet:note] ...`) so it no longer looks like one more injected memory bullet inside the recall block.

NicholaiVogel:
follow-up pass is in.

latest fixes:
- restored MCP `min_score` compatibility and introduced `score_min` for the new client-side score filter
- extracted shared recall partitioning/score-threshold helpers into `@signet/core`
- removed test-only helper exports from the CLI command module and moved the threshold unit test down into core
- tightened the CLI parser guard on `results`
- added explicit PR migration notes for the CLI `--json` shape change
- filled in `informed_by` on the approved explicit recall alignment spec
- changed the omitted-results line in prompt-submit to `[signet:note] ...` so it reads as metadata, not as another recalled memory

latest commits:
- `c8de11a9` fix(recall) preserve mcp threshold compatibility
- `77a97082` fix(recall) document cli json compatibility
- `51575e48` fix(recall) reduce recall renderer drift

latest targeted verification:
- `bun run build:core`
- `bun test packages/core/src/__tests__/recall.test.ts packages/cli/src/commands/memory.test.ts packages/daemon/src/mcp/tools.test.ts packages/daemon/src/hooks.prompt-submit.test.ts`
- targeted `biome check` on the touched core/cli/daemon files

also worth calling out for the truncated-diff concern: the live `/api/hooks/recall` route is in `packages/daemon/src/routes/hooks-routes.ts` after the route refactor, and the integration test is hitting the real app path there rather than a helper-level mock.

packages/core/src/recall.ts
```diff
@@ -0,0 +1,44 @@
...
+} {
+	return {
+		primary: rows.filter((row) => row.supplementary !== true),
+		supporting: rows.filter((row) => row.supplementary === true),
+	};
+}
+
+export function applyRecallScoreThreshold(raw: unknown, minScore?: number): unknown {
+	if (typeof minScore !== "number" || typeof raw !== "object" || raw === null || Array.isArray(raw)) {
+		return raw;
```
PR-Reviewer-Ant, line 34: `typeof minScore !== "number"` passes when `minScore` is `NaN` (since `typeof NaN === "number"` is true). The subsequent filter expression `row.score >= NaN` evaluates to `false` for every numeric score, so every scored row is silently discarded while unscored rows survive. On the CLI, `Number.parseFloat` returns `NaN` for non-numeric `--min-score` values, making `signet recall "query" --min-score typo` quietly drop all scored results instead of erroring. Fix: `typeof minScore !== "number" || !Number.isFinite(minScore)`.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -3068,6 +3071,66 @@ mod tests {
...
+            State(state.clone()),
+            HeaderMap::new(),
+            Json(PromptSubmitBody {
+                harness: Some("test".to_string()),
+                project: Some("packages/daemon-rs".to_string()),
+                agent_id: Some("agent-a".to_string()),
+                user_message: Some("show prompt submit observability review".to_string()),
+                user_prompt: None,
+                last_assistant_message: None,
+                session_key: Some("sess-prompt-structured".to_string()),
```
PR-Reviewer-Ant, line 3110: The Rust temporal-fallback test verifies the wrapper format and content text but not the per-line prefix. The TypeScript `buildTemporalFallbackResponse` now produces `- [thread ${hit.id}] ...` (changed from `- [node ${hit.id}]`), but the Rust diff does not show the corresponding per-line format change in the temporal fallback `picked` loop. The Rust test assertion only checks for the content string, not `[thread node-1]`. If the Rust path still emits `- [node node-1]`, the two runtimes diverge on temporal-fallback inject formatting. Add an assertion like `assert!(inject.contains("[thread"))` to the Rust test, or confirm the per-line change was applied upstream of this diff.

packages/daemon/src/hooks.ts
```diff
@@ -357,24 +357,19 @@ export interface RememberResponse {
 export interface RecallRequest {
```
PR-Reviewer-Ant, line 364: `handleRecall()` and `RecallResponse` are fully removed in this file, but `packages/daemon/src/daemon.ts` — which contains the `/api/hooks/recall` route that previously called `handleRecall()` — is absent from the diff. The `hooks-recall.test.ts` additions exercise that route end-to-end (including project-filter and type-filter tests that would fail against the old implementation), so the route update almost certainly exists. Still, the removal of an exported symbol and its only known call site in a file not present in the diff is a gap that cannot be confirmed from this review. Please confirm `daemon.ts` has no remaining reference to `handleRecall`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `51575e48`

[Automated Review — pr-reviewer] PR #474 aligns recall surfaces and removes the legacy `handleRecall()` helper. Three issues warrant attention before merge: a NaN passthrough in `applyRecallScoreThreshold` that silently discards scored results on invalid input, a Rust/TypeScript parity gap in temporal-fallback line formatting, and a `daemon.ts` route update that is absent from the diff but load-bearing for the `handleRecall()` removal.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - `packages/daemon/src/daemon.ts` is not in the diff but must have been updated to replace the removed `handleRecall()` call in the `/api/hooks/recall` route. The hooks-recall tests exercise that route end-to-end and are claimed to pass, so the update likely exists, but it is not reviewable here. The NaN bug and Rust parity gap are directly provable from the changed lines in `packages/core/src/recall.ts` (line 28) and `packages/daemon-rs/.../hooks.rs` (temporal-fallback test assertions at line ~3110).

packages/daemon/src/mcp/tools.ts
```diff
@@ -195,6 +226,57 @@ function textResult(value: unknown): { content: Array<{ type: "text"; text: stri
...
+			? payload.meta
+			: {
+					totalReturned: rows.length,
+					hasSupplementary: rows.some((row) => row.supplementary === true),
+					noHits: rows.length === 0,
+				};
+
+	if (meta.noHits || rows.length === 0) {
+		return "No matching memories found.";
+	}
```
PR-Reviewer-Ant, line 243: `applyRecallToolScoreMin` is a pure pass-through to `applyRecallScoreThreshold` with no added logic. It adds an indirection layer with no benefit; call `applyRecallScoreThreshold` directly at the call site.

packages/cli/src/commands/memory.ts
```diff
@@ -94,41 +187,26 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 			}
 
 			spinner.stop();
-			const result = typeof data === "object" && data !== null ? data : {};
-			const rows = Array.isArray(result.results) ? result.results : [];
+			const filtered = applyRecallScoreThreshold(data, options.minScore);
+			const parsed = parseRecallResult(filtered);
```
PR-Reviewer-Ant, line 193: `signet recall --json` now outputs the full response object (`{results, meta, query, method}`) instead of the bare results array. This is a breaking change for any scripts doing `signet recall "x" --json | jq '.[0].content'`. The PR migration notes document this, but the change is behavioral and not gated by a version flag. Worth confirming downstream consumers are aware.

packages/daemon/src/mcp/tools.ts
```diff
@@ -524,27 +606,58 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
...
 					limit: limit ?? 10,
+					project,
 					type,
-					importance_min: min_score,
-					expand,
+					tags,
+					who,
+					pinned,
+					importance_min: importance_min ?? min_score,
+					since,
```
PR-Reviewer-Ant, line 651: When both `importance_min` and `min_score` are provided, `importance_min ?? min_score` silently drops `min_score`. The MCP docs describe `min_score` as a deprecated alias for `importance_min`, but a caller providing both may not expect one to win silently. Consider logging a warning or documenting the precedence rule explicitly in the tool's description string.

packages/core/src/recall.ts
NicholaiVogel: fixed. applyRecallScoreThreshold() now treats non-finite thresholds as invalid (!Number.isFinite(minScore)), so a bad CLI parse like --min-score typo no longer silently discards every scored row.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
NicholaiVogel: fixed. the Rust temporal fallback now emits - [thread ...] to match TypeScript, and the rust regression test now asserts the [thread node-1] prefix explicitly instead of only checking the sample content.

packages/daemon/src/hooks.ts
NicholaiVogel: confirmed. there is no remaining live TypeScript call site for handleRecall(); the active /api/hooks/recall path lives in packages/daemon/src/routes/hooks-routes.ts and calls hybridRecall() directly. the helper removal is dead-code cleanup now, not a dangling route ref.

packages/daemon/src/mcp/tools.ts
NicholaiVogel: fixed. i removed the trivial MCP wrapper and now call applyRecallScoreThreshold() directly at the tool call site.

packages/cli/src/commands/memory.ts
NicholaiVogel: yep, that one stays as a documented compatibility break for now. i already added the explicit PR migration note because scripts doing jq on the old top-level array will need to switch to .results[0].content after this branch.

packages/daemon/src/mcp/tools.ts
NicholaiVogel: good catch. i clarified the MCP tool description so min_score explicitly says it is a deprecated compatibility alias for importance_min, and that it is ignored when importance_min is also provided. the precedence rule is now documented instead of being silent.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `bfaf6dcc`

```pr-review-json
{
  "summary": "Automated review by pr-reviewer. PR #474 is a substantial recall-surface alignment refactor. Core logic is sound and test coverage is meaningful. Found three warning-level issues and one nitpick; no blocking bugs identified from the visible diff.",
  "verdict": "comment",
  "confidence": {
    "level": "medium",
    "justification": "The diff covers core utilities, CLI, MCP, Rust parity, and tests. However, `packages/daemon/src/daemon.ts` changes are not in the diff — the `/api/hooks/recall` handler rewrite (where `handleRecall()` was replaced by `hybridRecall()` and the `project`-as-scope bug was fixed) is not directly reviewable. The hook-recall integration tests compensate partially, but the handler's forwarding of all new `RecallRequest` fields (`keywordQuery`, `tags`, `who`, `since`, `until`, `expand`) to `hybridRecall()` cannot be confirmed from this diff alone. SDK source changes are also absent (only docs changed in the diff)."
  },
  "comments": [
    {
      "file": "packages/core/src/recall.ts",
      "line": 44,
      "body": "Rows with no `score` field unconditionally pass `applyRecallScoreThreshold` regardless of `minScore`:\n\n```ts\nconst filtered = rows.filter((row) => typeof row.score !== "number" || row.score >= minScore);\n```\n\nThe intent is probably to treat unscored rows (e.g. LLM summary cards) as always-included, but this means a caller who sets `score_min: 0.8` will still receive every unscored row. In the MCP tool, an `llm_summary` supplementary card (which the API docs confirm has no score field) will always survive `score_min` filtering. If the current behavior is intentional, a comment explaining the policy would prevent future confusion. If it is not intentional, unscored rows should be excluded (or pinned) separately from the quality-threshold path.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/mcp/tools.ts",
      "line": 246,
      "body": "`formatRecallToolResult` emits `\"Found ${meta.totalReturned} memories\"` which produces grammatically incorrect output for exactly 1 result (\"Found 1 memories (hybrid).\"). Same issue appears in `packages/cli/src/commands/memory.ts` at the analogous `console.log(chalk.bold(...))` call. Neither is blocking but both are in user-facing / model-facing output.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/hooks.ts",
      "line": 2481,
      "body": "The omission hint changed from:\n\n```\n(+N more not shown — raise memory.guardrails.contextBudgetChars to include)\n```\n\nto:\n\n```\n[signet:note] N additional match(es) were omitted to keep this lightweight.\n```\n\nThe old message was actionable — it named the config key users could raise to include more results. The new message explains the effect but gives no path to remedy it. This reduces discoverability for users hitting budget limits. At minimum, consider keeping the config key in a parenthetical.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/hooks.ts",
      "line": 480,
      "body": "The removal of the `[signet:recall | query=\"...\" | results=N | engine=hybrid]` header from injected context is a protocol-level change. The PR compatibility notes cover `/api/hooks/recall` response shape, SDK method renames, and `signet recall --json` output format — but they do not mention this inject-format change. Any external harness integration, connector, or test that parses or pattern-matches the `[signet:recall` prefix in prompt-submit output will silently receive a different format. The Rust parity update (`build_prompt_recall_inject`) and both TypeScript fallback path tests verify the new format is consistent, but the change should be called out explicitly in the compatibility notes so downstream integrators can adapt.",
      "severity": "warning"
    },
    {
      "file": "packages/daemon/src/hooks.ts",
      "line": 357,
      "body": "Nitpick: `RecallRequest` gained several new fields (`keywordQuery`, `type`, `tags`, `who`, `since`, `until`, `expand`) but `handleRecall()` — its only prior consumer visible in this diff — was removed. The interface is presumably now the input type for the `/api/hooks/recall` handler in `daemon.ts` (not in the diff). If the interface is still exported and used there, this is fine. If it is unused after `handleRecall()` removal, it becomes dead exported API. Worth confirming `daemon.ts` references it before merge.",
      "severity": "nitpick"
    }
  ]
}
```

NicholaiVogel:
quick follow-up on the last review pass:

addressed in `bfaf6dcc`:
- hardened `applyRecallScoreThreshold()` against non-finite values, so a bad `--min-score` parse no longer silently drops every scored row
- brought rust temporal fallback back into parity with ts by emitting `- [thread ...]` and asserting that prefix in the rust regression test
- confirmed there is no remaining live ts call site for `handleRecall()`; `/api/hooks/recall` is served from `packages/daemon/src/routes/hooks-routes.ts` and calls `hybridRecall()` directly after the route refactor
- removed the trivial MCP score wrapper and documented `importance_min` / deprecated `min_score` precedence more explicitly

verification for that pass:
- `bun run build:core`
- `bun test packages/core/src/__tests__/recall.test.ts packages/daemon/src/mcp/tools.test.ts`
- `cargo test -p signet-daemon prompt_submit_formats_temporal_fallback_as_lightweight_recall_block -- --nocapture`
- targeted `biome check` on the touched files

packages/daemon/src/hooks.ts
```diff
@@ -3404,97 +3406,6 @@ export function handleRemember(req: RememberRequest): RememberResponse {
 	}
```
PR-Reviewer-Ant, line 3404: `handleRecall()` is deleted here and the `RecallResponse` interface is removed. The `/api/hooks/recall` route handler in `daemon.ts` previously called `handleRecall()` — that file is not in this diff. Since `bun run typecheck` is claimed to pass, TypeScript consumers are covered, but the explicit wiring of `hybridRecall()` into the hook-recall route is not visible for review. The `hooks-recall.test.ts` additions do validate end-to-end behavior, but it would be helpful to confirm the daemon.ts route handler change is present (perhaps omitted from the diff for size reasons).

packages/core/src/recall.ts
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	const payload = raw as RecallScoreFilterPayload;
+	const rows = Array.isArray(payload.results) ? payload.results : [];
+	// Keep unscored rows such as supplementary summaries in-band. Callers use
+	// score thresholds to trim ranked matches, not to strip contextual cards
+	// that do not participate in numeric ranking.
+	const filtered = rows.filter((row) => typeof row.score !== "number" || row.score >= minScore);
+
```
PR-Reviewer-Ant, line 43: `applyRecallScoreThreshold` intentionally passes through rows where `score` is `undefined` regardless of `minScore`, relying on a comment to document this. The comment is in the implementation, but callers at the MCP and CLI boundary have no signal that `minScore: 0.9` will not filter unscored rows (e.g., a primary result returned without a score due to a retrieval path that omits scoring). The behavior is justified for `llm_summary` supplementary cards, but rows without `supplementary: true` that also lack a score would silently bypass the threshold. Consider asserting or logging when a non-supplementary, unscored row survives threshold filtering, or at minimum document the bypass in the exported function's JSDoc.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `05c86776`

[Automated Review — pr-reviewer] PR #474 consolidates recall surfaces across CLI, MCP, OpenClaw, hooks, and the Rust parity layer. The approach is sound and the surface contracts are well-tested. Three items warrant discussion before merge: the `handleRecall()` daemon route handler is not visible in the diff (only the function deletion is), the `applyRecallScoreThreshold` bypass for unscored rows is underdocumented at the call boundary, and the `RecallResponse` export removal may affect SDK consumers who typed against it.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff does not include the `packages/daemon/src/daemon.ts` route handler for `/api/hooks/recall`. That handler previously called `handleRecall()` which is now deleted. The test file `hooks-recall.test.ts` validates the endpoint behavior end-to-end, providing indirect evidence the route was updated correctly, but the actual wiring of `hybridRecall()` into the route is not visible in this diff. Everything else is directly verifiable from changed lines.

packages/daemon/src/hooks.ts
```diff
@@ -364,24 +364,19 @@ export interface RememberResponse {
 export interface RecallRequest {
```
PR-Reviewer-Ant, line 364: `RecallResponse` was an exported interface and is now removed. The SDK diff updates `hookRecall()` to return the new richer shape, and the deprecated `recallHook()` alias is retained. However, if any external consumer (e.g., a third-party integration against the published `@signet/daemon` types) imported `RecallResponse` directly, removing it is a breaking change not covered by the `recallHook` compatibility alias. The migration notes document the new shape but do not call out the type removal explicitly. Worth a changelog note if `@signet/daemon` types are part of the public package surface.

packages/cli/src/commands/memory.ts
```diff
@@ -94,41 +187,27 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 			}
 
 			spinner.stop();
-			const result = typeof data === "object" && data !== null ? data : {};
-			const rows = Array.isArray(result.results) ? result.results : [];
+			const filtered = applyRecallScoreThreshold(data, options.minScore);
```
PR-Reviewer-Ant, line 192: The `--json` flag now emits the full recall response object (including `meta`, `query`, `method`) rather than only the `results` array. This is documented in migration notes and is the correct direction, but any existing shell scripts or automation that pipes `signet recall --json | jq '.[].content'` will silently break (they would need `jq '.results[].content'`). The breaking nature is acknowledged in migration notes but the old array format was the default for `--json` so external tooling breakage is real. No code change suggested — just confirming this is intentional and worth a prominent callout in the release notes.

packages/daemon/src/mcp/tools.ts
```diff
@@ -524,27 +603,61 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
 			inputSchema: z.object({
```
PR-Reviewer-Ant, line 603: The MCP `min_score` parameter is now documented as a deprecated alias for `importance_min` (server-side importance filter), and `score_min` is the new client-side retrieval score filter. These two concepts are semantically distinct: `importance_min` filters on the stored importance value, while `score_min` filters on the fused retrieval relevance score. An MCP consumer who previously used `min_score: 0.7` to gate on retrieval quality will now get that remapped to an importance filter instead of a score filter. The correct equivalent is `score_min: 0.7`. This is a subtle behavior change for existing callers. The MCP.md docs note `min_score` as deprecated but do not call out that it maps to a different semantic axis than `score_min`.

packages/cli/src/commands/memory.ts
```diff
@@ -78,12 +166,17 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 			const spinner = ora("Searching memories...").start();
 			const { ok, data } = await deps.secretApiCall("POST", "/api/memory/recall", {
 				query,
+				keywordQuery: options.keywordQuery,
 				limit: options.limit,
+				project: options.project,
 				type: options.type,
 				tags: options.tags,
 				who: options.who,
+				pinned: options.pinned === true ? true : undefined,
```
PR-Reviewer-Ant, line 175: `pinned: options.pinned === true ? true : undefined` — when `--pinned` is not passed, the commander default is `false`, so this sends `undefined`. `JSON.stringify` drops `undefined` values, so the key is not forwarded to the daemon. This is correct behavior. Noting it as confirmed-safe for reviewers who see the conditional and wonder why it is not simply `options.pinned || undefined`.

packages/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -1158,7 +1173,7 @@ pub async fn prompt_submit(
                         }
```
PR-Reviewer-Ant, line 1173: The Rust temporal fallback now uses `- [thread {id}]` as the prefix. The TypeScript `buildTemporalFallbackResponse` was updated to match (`[thread {hit.id}]`). The Rust test at line 3071 asserts `inject.contains("[thread node-1]")`. Parity looks correct. Minor note: the Rust test uses `thread_key: "thread-prompt-structured"` but the assertion checks for `[thread node-1]` (the `node_id` column). Confirm the Rust handler uses `node_id` not `thread_key` for this prefix — the test implies it does, but this is worth a quick eyes-on of the actual Rust handler logic around line 1173.

packages/core/src/recall.ts
PR-Reviewer-Ant, line 43: **Semantic trap: unscored rows bypass `minScore` filtering, but `meta.noHits` and `meta.totalReturned` include them.**

The filter intentionally preserves rows with no `score` field (`typeof row.score !== 'number'`). When a user sets `--min-score 0.9` and all scored/primary rows are below threshold, only unscored supplementary rows survive. The result:
- `meta.noHits` is `false` (array is non-empty)
- `meta.totalReturned` counts the surviving supplementary rows
- CLI shows `Found N memories: (includes supporting context)` with a "Supporting context:" section but no "Primary matches:"
- MCP shows `Found N memory (hybrid).` followed only by `Supporting context:`

This diverges from user expectation: `--min-score 0.9` reads as "only show me results scoring above 0.9" but silently delivers unscored supplementary cards. Neither `docs/CLI.md` nor `docs/MCP.md` document this exception. Consider either (a) documenting it in both doc tables under `--min-score` / `score_min`, or (b) rechecking whether supplementary cards should also be subject to the threshold (with a separate floor or pass-all flag).

packages/core/src/recall.ts
```diff
@@ -0,0 +1,53 @@
...
+	return {
+		...payload,
+		results: filtered,
+		meta: {
+			totalReturned: filtered.length,
+			hasSupplementary: filtered.some((row) => row.supplementary === true),
+			noHits: filtered.length === 0,
```
PR-Reviewer-Ant, line 50: **`meta` spread drops any extra fields the daemon puts on the original `meta` object.**

The return `{ ...payload, results: filtered, meta: { totalReturned, hasSupplementary, noHits } }` fully replaces the server-returned `meta`. If the daemon ever adds fields to `meta` (e.g., `engine`, `reranked`, `graphBoosted`), client-side filtering silently discards them. This is low-risk today given the current three-field contract, but worth noting if `meta` grows.

packages/sdk/src/__tests__/client.test.ts
```diff
@@ -1,6 +1,6 @@
 import { afterEach, describe, expect, test } from "bun:test";
```
PR-Reviewer-Ant, line 1: **SDK source (`packages/sdk/src/`) not in diff — `sessionKey` rename unverifiable.**

`docs/SDK.md` documents `sessionKey` (replacing `session_key`) for `sessionStart`, `userPromptSubmit`, `sessionEnd`, `hookRemember`, and `hookRecall`. If the actual SDK TypeScript source was not updated to match, callers following the new docs would break. This may be a diff-truncation artifact. Reviewer should confirm the SDK source exports match the documented parameter names.

packages/connector-openclaw/src/index.ts
```diff
@@ -1144,6 +1144,50 @@ async function fetchDaemon(path, body) {
   return res.json();
 }
 
+function formatRecallMessage(data) {
```
PR-Reviewer-Ant, line 1147: **`meta` fallback recomputes `hasSupplementary` from `data?.results` but `data` is the full daemon response, not a pre-filtered payload.**

The `formatRecallMessage` function computes a fallback `meta` from `data.results` when `data.meta` is absent. But `data` here is the raw `fetchDaemon()` response from `/api/hooks/recall`. Post-PR the daemon always returns `meta`, so this fallback path is only hit on older daemon versions — acceptable for a transition period, but worth a comment so the fallback isn't silently load-bearing.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7761ca25`

[Automated Review — pr-reviewer bot] PR #474 is a well-scoped refactor with comprehensive test coverage. Found two semantic issues with client-side score filtering and one auditing gap from a truncated diff.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff is explicitly truncated — `packages/daemon/src/daemon.ts` (the file that contains the actual `/api/hooks/recall` route handler, legacy alias injection, `project`-not-`scope` fix, and `bypassed`/`internal` flags) is not present. The integration tests in `hooks-recall.test.ts` cover these behaviors at the HTTP layer so the implementation almost certainly exists, but I cannot audit it directly. The core recall utility (`packages/core/src/recall.ts`) and consumer changes are fully visible and reviewed.


- packages/daemon/src/daemon.ts:1 - **`daemon.ts` not in diff — `/api/hooks/recall` implementation unauditable.**

The PR's core behavioral fix is that `project` on `/api/hooks/recall` is now forwarded as the memory `project` filter instead of being remapped to `scope`. That change, plus the legacy `memories`/`count` alias injection, `bypassed`/`internal` flags, and the removal of the `handleRecall()` call, all live in `daemon.ts`. The diff is truncated before this file appears.

The integration tests in `hooks-recall.test.ts` cover these behaviors at the HTTP layer (project isolation, type forwarding, bypassed-session shape, internal shape), so the implementation almost certainly exists and is correct. But the source is not reviewable here. If this is purely a display truncation, no action needed — reviewer should confirm `daemon.ts` was updated.

NicholaiVogel:
quick note on the latest review wave:

- the live `/api/hooks/recall` implementation is in `packages/daemon/src/routes/hooks-routes.ts`, not `daemon.ts` anymore after the route split. that route is already wiring `hybridRecall()` directly, forwarding `project` as `project`, and injecting the legacy `memories` / `count` aliases plus `bypassed` / `internal` flags. the HTTP integration coverage in `hooks-recall.test.ts` is hitting that real path.
- i added one more narrow follow-up commit (`d4f657b5`) to document the intentional score-threshold policy at the CLI/MCP call boundaries too: score filtering trims ranked matches, but keeps unscored supporting context in-band.
- the removed `RecallResponse` in `packages/daemon/src/hooks.ts` was only the old hook-local helper type that died with `handleRecall()`. the actual daemon/sdk recall contracts still exist in `packages/daemon/src/memory-search.ts` and `packages/sdk/src/types.ts`.

so from my side there isn't a new contract break hiding in this round, mostly just the bot looking at a truncated diff again.

packages/core/src/recall.ts
```diff
@@ -0,0 +1,53 @@
...
+		primary: rows.filter((row) => row.supplementary !== true),
+		supporting: rows.filter((row) => row.supplementary === true),
+	};
+}
+
+export function applyRecallScoreThreshold(raw: unknown, minScore?: number): unknown {
+	if (
+		typeof minScore !== "number" ||
+		!Number.isFinite(minScore) ||
+		typeof raw !== "object" ||
```
PR-Reviewer-Ant, line 30: `const payload = raw as RecallScoreFilterPayload;` uses a type assertion (`as`) after already narrowing `raw` to a non-null, non-array object. AGENTS.md explicitly states 'Avoid `as`' — prefer a type guard function or narrowing via `in` checks. The cast is practically safe here (all required fields are optional, the spread copies runtime properties correctly) but is a conventions violation in shared core utility code that is now exported.

packages/daemon/src/hooks.ts
```diff
@@ -364,24 +364,19 @@ export interface RememberResponse {
...
-		content: string;
-		type: string;
-		importance: number;
-		tags: string | null;
-		created_at: string;
-	}>;
-	count: number;
-}
-
 // ============================================================================
```
PR-Reviewer-Ant, line 380: `RecallResponse` is deleted along with `handleRecall()`. Before removal, confirm this interface is not re-exported from `packages/daemon/src/index.ts` or any barrel file. If it is, downstream consumers get a silent type-only breaking change. The diff does not include the daemon package's index exports.

packages/cli/src/commands/memory.ts
```diff
@@ -94,41 +187,29 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 			}
```
PR-Reviewer-Ant, line 187: The `--json` flag now emits the full recall response object (`{ results, query, method, meta }`) instead of only the `results` array. This is a breaking change for any automation or scripts already parsing `signet recall --json`. It is documented in the PR migration notes, but the change also applies to existing `signet recall --json` callers that are not SDK users — third-party integrations, CI scripts, shell pipelines. Worth ensuring this is surfaced in the release notes and not just the PR description.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e95ac3c9`

[Automated Review — pr-reviewer] PR #474 is a broad recall-surface alignment refactor with good test coverage on the visible diff. Two structural concerns stand out: the primary behavioral fix (`project` vs `scope` mapping in `/api/hooks/recall`) lives in `daemon.ts`, which is absent from the diff entirely; and a `as` cast in the new `@signet/core` utility violates the codebase's stated convention. One additional warning about a breaking CLI output change and a potential public-type deletion are noted below.

**Confidence:** Medium [missing_cross_module_context, sufficient_diff_evidence] - The PR's primary behavioral fix — mapping `project` as a memory project filter instead of remapping it to `scope` in `/api/hooks/recall` — is not present in the diff. `packages/daemon/src/daemon.ts` is the only plausible location for the updated route handler (which previously called the now-deleted `handleRecall()`), yet it is absent from the changeset. The test file `hooks-recall.test.ts` exercises the correct behavior and presumably passes, but the implementation is unauditable from this diff. All other findings are directly provable from the changed lines.


- packages/daemon/src/daemon.ts:1 - This file is absent from the diff but contains the `/api/hooks/recall` route handler. The PR claims to fix `project` being silently remapped to `scope`; the old handler called `handleRecall()` which is now deleted. Without seeing the replacement implementation, the following cannot be verified: (1) `project` is now forwarded as a `hybridRecall()` project filter and not a scope substitution; (2) the `memories`/`count` legacy alias fields are actually added to the response; (3) the `bypassed`/`internal` no-op response shapes are correctly emitted. The test coverage in `hooks-recall.test.ts` exercises all three behaviors but the source is not in scope for audit. The fix should be auditable.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `70a54888`

[Automated Review — pr-reviewer] This PR is well-structured with good test coverage and a clear resolution of prior review concerns. Two minor consistency gaps remain: the `dependencies.yaml` `informed_by` field for the new spec is empty while the spec frontmatter has it populated, and the TS/Rust `buildPromptRecallInject` parity is enforced only by cross-reference comments rather than a shared contract test. No blocking issues found.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The full diff was reviewed, the SDK types were fetched directly from the head SHA to confirm the `session_key`→`sessionKey` concern was a docs-only correction (the actual SDK types always used `sessionKey`), and the hooks-routes.ts recall handler was confirmed to forward all new filter fields to `hybridRecall()`. The `dependencies.yaml` inconsistency is directly visible in the diff at the new spec entry.

docs/specs/dependencies.yaml
```diff
@@ -238,6 +238,40 @@ specs:
...
+    path: docs/specs/approved/explicit-recall-surface-alignment.md
+    hard_depends_on:
+      - memory-pipeline-v2
+      - signet-runtime
+    soft_depends_on: []
+    blocks: []
+    informed_by: []
+    success_criteria:
+      - "TypeScript explicit recall uses one canonical retrieval engine and one canonical response contract across API, CLI, MCP, and harness consumers"
+      - "Prompt-submit recall remains a lightweight injection path and does not absorb explicit recall product concerns"
```
PR-Reviewer-Ant, line 252: The `informed_by: []` field on the `explicit-recall-surface-alignment` entry is empty, but the spec file frontmatter at `docs/specs/approved/explicit-recall-surface-alignment.md` correctly lists both `RESEARCH-OBSIDIAN-VAULT-RECALL-EVAL.md` and `memory-benchmark-comparison.md`. Prior review comment #29 was supposedly addressed but only the spec frontmatter was updated — the registry entry was not. If `bun scripts/spec-deps-check.ts` validates `informed_by` consistency, this could surface as a registry drift warning alongside the pre-existing unrelated ones the PR already notes.

packages/daemon/src/hooks.ts
```diff
@@ -519,6 +511,18 @@ export function applyTokenBudget(inject: string, mainBudget: number): string {
 	return truncateToTokens(inject, mainBudget - TRUNCATED_MARKER_TOKENS) + TRUNCATED_MARKER;
```
PR-Reviewer-Ant, line 511: `buildPromptRecallInject()` and its Rust counterpart `build_prompt_recall_inject()` in `hooks.rs` are kept in sync only by a prose cross-reference comment. Both implementations now have independent unit tests that verify their own format (`hooks.prompt-submit.test.ts` and the new Rust `prompt_submit_formats_temporal_fallback_as_lightweight_recall_block` test), so major divergence would be caught. However, subtle differences — e.g., trailing-newline count, blank-line placement, or the footer text — could still diverge silently between test runs on different runtimes. This is an existing AGENTS.md parity obligation; flagging as a heads-up that the current test coverage is runtime-isolated rather than cross-runtime.

Issue #458, opened 2026-04-03
bug: claude-code extraction provider ignores configured model — always uses CLI default


BusyBee3333:
## Bug

When using `claude-code` as the extraction provider, the configured `model` in `agent.yaml` is silently ignored. All extraction calls go to whatever model the user's Claude CLI defaults to (typically Sonnet), regardless of config.

## Root Cause

In `buildCommand()`, the `claude-code` case constructs:

```typescript
case "claude-code":
  return ["claude", ["--dangerously-skip-permissions", "-p", prompt]];
```

The `model` parameter from `createClaudeCodeProvider({ model: model || "haiku" })` is accepted by the constructor but **never passed to the CLI command**. The Claude CLI supports `--model <model>` — it's just not being used.

## Impact

- Users configuring `extraction.model: haiku` are unknowingly billed at Sonnet rates
- Reported by Woska who configured `haiku` + `native` but saw Sonnet usage
- Combined with the MEMORY.md re-ingestion loop (#452), this compounds token waste significantly

## Expected Fix

Pass `--model` to the CLI command:

```typescript
case "claude-code":
  return ["claude", ["--dangerously-skip-permissions", "--model", model || "haiku", "-p", prompt]];
```

## Environment

- Signet v0.86.4+
- Claude CLI with OAuth (Pro subscription)
- `agent.yaml` extraction config: `provider: claude-code, model: haiku`

## Related

- #452 (MEMORY.md re-ingestion loop — compounds this issue)

aaf2tbz:
I reviewed this against current `main` before patching.

The issue as originally written is no longer accurate for the extraction path: `packages/daemon/src/pipeline/provider.ts` already passes `--model` to the Claude CLI on current `main`.

The remaining live bug was in scheduled task execution instead:

- `packages/daemon/src/scheduler/spawn.ts` dropped the Claude model for scheduled runs
- the manual `/api/tasks/:id/run` path also did not pass a resolved task model through
- while fixing that, scheduler tests exposed a second bug where the task-model cache was keyed only by `agentsDir`, so one harness lookup could bleed into another

I opened a draft PR for the still-reproducible bug here:

PR: https://github.com/Signet-AI/signetai/pull/475

That PR fixes the scheduled-task path, adds Claude coverage in scheduler tests, and fixes the per-harness cache keying.

[...]

PR #475, opened 2026-04-07
[Daemon] fix Claude scheduled task model propagation

aaf2tbz:
## Summary

Fix the remaining Claude model-propagation bug in scheduled task execution. The extraction provider path on current `main` already passes `--model`; the still-live bug was in scheduler task spawning and the manual task-run path.

## Changes

- pass `--model` to the Claude CLI in scheduled task spawning
- resolve scheduled-task models for `claude-code` as well as `codex`
- pass the resolved task model through `/api/tasks/:id/run`
- add regression tests for Claude scheduled-task model propagation
- fix task-model cache keying so one harness lookup cannot bleed into another

### Round 1

- initial fix landed in `66318983` (`fix(daemon): propagate Claude task model`)
- addressed the core scheduler/task-run propagation bug
- added Claude scheduled-task coverage and cache-key isolation support

### Round 2

- addressed PR-Reviewer-Ant feedback in `f7073f06` (`fix(daemon): tighten task model resolution`)
- replaced the duplicate cache-isolation test with a real per-harness cache proof
- narrowed task-model resolution so only harnesses that support model propagation (`claude-code`, `codex`) call `resolveTaskModel`

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

No UI changes.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

Not applicable: no migrations.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Targeted validation run:

- `bun test packages/daemon/src/scheduler/spawn.test.ts packages/daemon/src/scheduler/worker.test.ts packages/daemon/src/scheduler/worker-execute.test.ts`

## AI disclosure

## Notes

Issue #458 was stale as written: the extraction provider path is already fixed on current `main`. This PR addresses the still-reproducible scheduled-task path instead and is linked back from the issue comment.

### DIFF
--- a/packages/daemon/src/scheduler/worker.test.ts
+++ b/packages/daemon/src/scheduler/worker.test.ts
@@ -131,13 +131,21 @@ describe("resolveTaskModel", () => {
 	it("caches models independently per harness", () => {
 		const agentsDir = mkdtempSync(join(tmpdir(), "signet-agents-"));
 		try {
+			const configPath = join(agentsDir, "agent.yaml");
 			writeFileSync(
-				join(agentsDir, "agent.yaml"),
+				configPath,
 				["memory:", "  pipelineV2:", "    extraction:", "      provider: claude-code", "      model: haiku"].join("\n"),
 			);
 
 			expect(resolveTaskModel("claude-code", agentsDir)).toBe("haiku");
-			expect(resolveTaskModel("codex", agentsDir)).toBeUndefined();
+
+			writeFileSync(
+				configPath,
+				["memory:", "  pipelineV2:", "    extraction:", "      provider: codex", "      model: gpt-5.4-codex"].join("\n"),
+			);
+
+			expect(resolveTaskModel("codex", agentsDir)).toBe("gpt-5.4-codex");
+			expect(resolveTaskModel("claude-code", agentsDir)).toBe("haiku");
 		} finally {
 			rmSync(agentsDir, { recursive: true, force: true });
 		}
--- a/packages/daemon/src/scheduler/worker.test.ts
+++ b/packages/daemon/src/scheduler/worker.test.ts
@@ -113,6 +113,36 @@ describe("resolveTaskModel", () => {
 		}
 	});
 
+	it("returns the configured claude extraction model for claude-code tasks", () => {
+		const agentsDir = mkdtempSync(join(tmpdir(), "signet-agents-"));
+		try {
+			writeFileSync(
+				join(agentsDir, "agent.yaml"),
+				["memory:", "  pipelineV2:", "    extraction:", "      provider: claude-code", "      model: haiku"].join("\n"),
+			);
+
+			expect(resolveTaskModel("claude-code", agentsDir)).toBe("haiku");
+			expect(resolveTaskModel("codex", agentsDir)).toBeUndefined();
+		} finally {
+			rmSync(agentsDir, { recursive: true, force: true });
+		}
+	});
+
+	it("caches models independently per harness", () => {
+		const agentsDir = mkdtempSync(join(tmpdir(), "signet-agents-"));
+		try {
+			writeFileSync(
+				join(agentsDir, "agent.yaml"),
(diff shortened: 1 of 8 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `[codex] fix Claude scheduled task model propagation` (commit `66318983`) now and I'll follow up shortly with feedback.

packages/daemon/src/scheduler/worker.test.ts
```diff
@@ -113,6 +113,36 @@ describe("resolveTaskModel", () => {
...
+				["memory:", "  pipelineV2:", "    extraction:", "      provider: claude-code", "      model: haiku"].join("\n"),
+			);
+
+			expect(resolveTaskModel("claude-code", agentsDir)).toBe("haiku");
+			expect(resolveTaskModel("codex", agentsDir)).toBeUndefined();
+		} finally {
+			rmSync(agentsDir, { recursive: true, force: true });
+		}
+	});
+
```
PR-Reviewer-Ant, line 130: The `"caches models independently per harness"` test body is byte-for-byte identical to the preceding `"returns the configured claude extraction model for claude-code tasks"` test. It does not verify the property its name claims: to prove independent per-harness cache entries, the test would need to (1) populate one harness's cache, (2) write a different model to disk, and (3) verify the second harness reads fresh while the first harness still serves the cached value. As written, this test adds zero additional coverage and gives a false sense of security for the isolation fix.

packages/daemon/src/daemon.ts
```diff
@@ -7630,6 +7631,7 @@ app.post("/api/tasks/:id/run", async (c) => {
 	const taskSkillMode = typeof task.skill_mode === "string" ? task.skill_mode : null;
 	const taskWorkingDir = typeof task.working_directory === "string" ? task.working_directory : null;
 	const taskAgentId = readTaskAgentId(task, scoped.agentId);
+	const taskModel = resolveTaskModel(taskHarness);
```
PR-Reviewer-Ant: `resolveTaskModel(taskHarness)` is called where `taskHarness` is coerced from a raw DB string with no harness validation. When `taskHarness` is an unrecognized value (e.g. a future harness, or a corrupted DB row), `resolveTaskModel` silently returns `undefined` and the task is spawned without a model — no log line, no error, no early return. The scheduler path (`executeTask`) does the same. Since both paths accept `undefined` gracefully this won't crash, but it masks misconfigured or unknown harness values. Consider logging a warning when `resolveTaskModel` receives a harness it cannot resolve, or validating `taskHarness` against the known set before calling spawn.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `66318983`

[Automated Bot Review] Two test quality issues found: one duplicate test and one silent-failure path in the manual-run handler. No blocking correctness bugs in the core fix.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The duplicate test bodies are byte-for-byte identical, provable directly from the diff. The `task.harness` runtime type path is traceable end-to-end through the diff: daemon.ts coerces any string through `typeof task.harness === "string"`, then passes it to `resolveTaskModel` which silently returns `undefined` for unknown harnesses. No missing runtime traces or cross-module context are needed to confirm either finding.

NicholaiVogel:
@aaf2tbz we should probably take care of the conflicts and get this merged, just so it doesn't go stale

PR #476, opened 2026-04-07
Fix Codex hook integration

aaf2tbz:
## Summary
- update the Codex connector to write the current `hooks.json` schema and clean up both modern and legacy Signet hook entries
- emit Codex command hook JSON from `signet hook ... -H codex` so `SessionStart` and `UserPromptSubmit` inject context through `hookSpecificOutput.additionalContext`
- resolve packaged Signet CLI/MCP entrypoints for Codex so fresh Codex processes do not depend on `PATH` containing `signet` or `signet-mcp`
- add regression coverage for the new Codex hook wire format and connector install/uninstall behavior, and update Codex harness docs

## Validation
- `cd packages/cli && bun test ./src/commands/hook.test.ts`
- `bun run build:core`
- `bun run build:connector-base`
- `cd packages/connector-codex && bun test ./src/config-toml.test.ts`
- `cd packages/connector-codex && bun run typecheck`
- `cd packages/connector-codex && bun run build`
- `bun run build:deps`

## Notes
- Reproduced against `codex-cli 0.118.0`: fresh Codex startup could miss Signet entirely when the startup environment did not include `signet` on `PATH`.
- `bun run build:signetai` in this worktree still fails on unrelated workspace resolution for `@signet/connector-oh-my-pi` during `packages/signetai build:cli`; this change does not touch that packaging path.

## PR Readiness Checklist

### DIFF
--- a/packages/cli/src/commands/hook.test.ts
+++ b/packages/cli/src/commands/hook.test.ts
@@ -189,6 +189,36 @@ describe("buildUserPromptSubmitBody", () => {
 		expect(seen[0]?.body).toContain('"harness":"claude-code"');
 		expect(lines).toContain("recalled context");
 	});
+
+	test("hook command emits Codex wire format for user-prompt-submit", async () => {
+		const lines: string[] = [];
+		console.log = (line?: unknown) => {
+			lines.push(String(line ?? ""));
+		};
+
+		const program = new Command();
+		registerHookCommands(program, {
+			AGENTS_DIR: "/tmp/agents",
+			fetchDaemonResult: async () => ({
+				ok: true,
+				data: {
+					inject: "  recalled context  ",
+				},
+			}),
+			readStaticIdentity: () => null,
+		});
+
+		await program.parseAsync(["node", "test", "hook", "user-prompt-submit", "-H", "codex"]);
+
+		expect(lines).toHaveLength(1);
+		expect(JSON.parse(lines[0] ?? "")).toEqual({
+			continue: true,
+			hookSpecificOutput: {
+				hookEventName: "UserPromptSubmit",
+				additionalContext: "recalled context",
+			},
+		});
+	});
 });
 
 describe("Codex hook wire output", () => {
@@ -218,6 +248,46 @@ describe("Codex hook wire output", () => {
 			hookSpecificOutput: {
 				hookEventName: "SessionStart",
 				additionalContext: null,
+			},
+		});
+	});
+
+	test("trims surrounding whitespace in Codex injection", () => {
+		expect(buildCodexSessionStartOutput("  hello  ")).toEqual({
+			continue: true,
+			hookSpecificOutput: {
+				hookEventName: "SessionStart",
+				additionalContext: "hello",
+			},
+		});
+	});
+
+	test("Codex harness output takes precedence over --json on session-start", async () => {
+		const lines: string[] = [];
+		console.log = (line?: unknown) => {
+			lines.push(String(line ?? ""));
+		};
+
+		const program = new Command();
+		registerHookCommands(program, {
+			AGENTS_DIR: "/tmp/agents",
+			fetchDaemonResult: async () => ({
+				ok: true,
+				data: {
+					inject: "session context",
+				},
+			}),
+			readStaticIdentity: () => null,
+		});
+
+		await program.parseAsync(["node", "test", "hook", "session-start", "-H", "codex", "--json"]);
+
+		expect(lines).toHaveLength(1);
+		expect(JSON.parse(lines[0] ?? "")).toEqual({
+			continue: true,
+			hookSpecificOutput: {
+				hookEventName: "SessionStart",
+				additionalContext: "session context",
 			},
 		});
 	});
--- a/packages/cli/src/commands/hook.test.ts
+++ b/packages/cli/src/commands/hook.test.ts
@@ -1,6 +1,8 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import { Command } from "commander";
 import {
+	buildCodexSessionStartOutput,
+	buildCodexUserPromptSubmitOutput,
 	buildCompactionCompleteBody,
 	buildSessionEndBody,
 	buildSessionStartFallback,
@@ -186,6 +188,38 @@ describe("buildUserPromptSubmitBody", () => {
 		expect(seen[0]?.path).toBe("/api/hooks/user-prompt-submit");
 		expect(seen[0]?.body).toContain('"harness":"claude-code"');
 		expect(lines).toContain("recalled context");
+	});
+});
+
+describe("Codex hook wire output", () => {
+	test("wraps session-start injection in Codex hook JSON", () => {
+		expect(buildCodexSessionStartOutput("hello")).toEqual({
+			continue: true,
+			hookSpecificOutput: {
+				hookEventName: "SessionStart",
+				additionalContext: "hello",
+			},
+		});
+	});
+
+	test("wraps prompt-submit injection in Codex hook JSON", () => {
+		expect(buildCodexUserPromptSubmitOutput("prompt ctx")).toEqual({
+			continue: true,
+			hookSpecificOutput: {
+				hookEventName: "UserPromptSubmit",
+				additionalContext: "prompt ctx",
+			},
+		});
+	});
+
+	test("normalizes empty Codex injection to null", () => {
+		expect(buildCodexSessionStartOutput("")).toEqual({
+			continue: true,
+			hookSpecificOutput: {
+				hookEventName: "SessionStart",
+				additionalContext: null,
+			},
+		});
 	});
 });
 
--- a/packages/cli/src/commands/hook.ts
+++ b/packages/cli/src/commands/hook.ts
@@ -36,6 +36,38 @@ export function buildSessionStartFallback(
 	return readStaticIdentity(agentsDir);
 }
 
+interface CodexCommandOutput {
+	continue: true;
+	hookSpecificOutput?: {
+		hookEventName: "SessionStart" | "UserPromptSubmit";
+		additionalContext: string | null;
+	};
+}
+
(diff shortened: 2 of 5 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `Fix Codex hook integration` (commit `0bef58a0`) now and I'll follow up shortly with feedback.

packages/cli/src/commands/hook.ts
```diff
@@ -123,6 +155,8 @@ export function registerHookCommands(program: Command, deps: HookDeps): void {
 					}
```
PR-Reviewer-Ant, line 160: `--json` silently takes priority over `-H codex` in both the fallback path (line ~155) and the success path (line ~177). If a caller ever passes `--json -H codex`, the output is the old `{ inject, identity, memories }` shape rather than Codex hook JSON, so Codex would silently receive an unrecognised format instead of `hookSpecificOutput.additionalContext`. The priority ordering should either be reversed (harness-specific format wins) or documented and guarded. Since `--json` is an explicit debug flag and not part of the Codex hook invocation, the risk is low in practice, but the silent incompatibility is a correctness trap.

packages/cli/src/commands/hook.test.ts
```diff
@@ -189,6 +191,38 @@ describe("buildUserPromptSubmitBody", () => {
 	});
```
PR-Reviewer-Ant, line 191: Tests cover the builder functions (`buildCodexSessionStartOutput`, `buildCodexUserPromptSubmitOutput`) but not the actual command-dispatch code path in `registerHookCommands` for `-H codex`. The existing `user-prompt-submit` integration test (line ~168) exercises the full command → daemon fetch → console.log path for `claude-code`, but there is no equivalent for `codex`. Per the repo's convention, a bug fix should include a test that would have failed before the fix. Adding a parallel test that calls `registerHookCommands`, parses the stdout captured from `console.log`, and asserts the Codex JSON shape would close this gap and protect against future regressions in the routing logic (e.g. the `--json` priority issue above).

packages/connector-codex/src/index.ts
```diff
@@ -8,15 +8,15 @@ import { expandHome } from "@signet/core";
...
  *  the bin directory. Falls back to bare "signet" if the layout doesn't match (shims, junctions). */
-function resolveSignetArgs(): string[] {
-	if (process.platform !== "win32") return ["signet"];
+function resolveSignetCommand(): string {
+	if (process.platform !== "win32") return "signet";
 	const entry = process.argv[1] || "";
 	const signetJs = join(entry, "..", "..", "bin", "signet.js");
-	if (existsSync(signetJs)) return [process.execPath, signetJs];
-	return ["signet"];
+	if (existsSync(signetJs)) return `"${process.execPath}" "${signetJs}"`;
```
PR-Reviewer-Ant: On Windows, `resolveSignetCommand()` returns `"${process.execPath}" "${signetJs}"` using bare template-literal interpolation. If `process.execPath` or `signetJs` contains a `"` character (unusual but possible on adversarial or non-standard Windows layouts), the resulting shell string will be malformed — either breaking the hook invocation or allowing argument injection. Consider escaping embedded double-quotes: `s.replace(/"/g, '\\"')` before interpolation, or using single-quotes with a fallback for paths that contain single-quotes.

packages/cli/src/commands/hook.ts
```diff
@@ -36,6 +36,38 @@ export function buildSessionStartFallback(
 	return readStaticIdentity(agentsDir);
 }
 
+interface CodexCommandOutput {
+	continue: true;
+	hookSpecificOutput?: {
+		hookEventName: "SessionStart" | "UserPromptSubmit";
+		additionalContext: string | null;
```
PR-Reviewer-Ant, line 43: `inject?.trim() ? inject : null` gates on the trimmed value but returns the original (untrimmed) string as `additionalContext`. A string like `"  hello  "` passes the truthiness check but injects with leading/trailing whitespace preserved. This is likely intentional (callers may rely on exact content), but worth confirming — if Codex trims `additionalContext` itself there is no issue, but if it doesn't, unexpected whitespace could appear in the model context.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0bef58a0`

[Automated Review — pr-reviewer bot] PR #476 updates the Codex connector to emit the Codex 0.118+ hook JSON format and cleans up legacy schema keys. The implementation is broadly correct. Three issues are worth flagging: a missing command-level integration test for the Codex output path, a silent `--json` priority over `-H codex` that produces incompatible output if both flags are used, and a Windows shell-string construction that could be malformed if `process.execPath` contains double-quote characters.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All three findings are directly provable from the changed lines in hook.ts and index.ts. The `--json` priority issue is visible at lines 155 and 177 of hook.ts in the diff. The Windows quoting risk is explicit in resolveSignetCommand(). The missing integration test gap is confirmed by inspecting the test file, which only exercises the builder functions but not the registerHookCommands Codex code path (compare with the existing end-to-end test for claude-code at line ~168 of hook.test.ts).

packages/connector-codex/src/index.ts
```diff
@@ -8,24 +8,30 @@ import { expandHome } from "@signet/core";
...
-	if (existsSync(signetJs)) return [process.execPath, signetJs];
-	return ["signet"];
+	if (!entry) return null;
+	for (const relativePath of relativePaths) {
+		const candidate = join(entry, "..", "..", relativePath);
+		if (existsSync(candidate)) return candidate;
+	}
+	return null;
+}
+
```
PR-Reviewer-Ant, line 20: The `resolveSignetCommand()` function now runs the quoted-string construction on all platforms, not just Windows. On Linux/macOS, when the packaged bin is found, it returns `"<execPath>" "<signetJs>"` — a shell-quoted string embedded directly into the `hooks.json` `command` field. There are two sub-risks:

1. **Unverified Codex shell dispatch**: The `type: "command"` hook format in Codex 0.118 is not documented here to guarantee shell invocation (`sh -c`). If Codex splits the command string on whitespace or uses exec-family syscalls without a shell, the embedded double-quotes become literal characters in the argv, breaking the hook silently.

2. **Path quoting correctness**: Paths containing `"` characters (uncommon but not impossible — e.g. symlink targets, unusual npm global dirs) will produce a malformed shell string. The template-literal interpolation `"${process.execPath}" "${signetJs}"` performs no escaping.

The test at config-toml.test.ts line ~163 only exercises the fallback (`signet hook session-start -H codex`) because in the test environment the packaged bin is not found. The actual cross-platform packaged-path code path is never exercised in CI.

packages/cli/src/commands/hook.ts
PR-Reviewer-Ant, line 160: `--json` is checked before `-H codex` in both the fallback path (line 155) and the success path (line 177). If a caller ever passes both `--json` and `-H codex` (e.g. a future integration test or a misconfigured hook runner), the output is the legacy `{ inject, identity, memories }` shape instead of the Codex `{ continue, hookSpecificOutput }` shape. Codex would silently receive unrecognised JSON and inject nothing. The two branches should be mutually exclusive with an explicit guard or a documented note that `--json` supersedes harness-specific output.

packages/cli/src/commands/hook.ts
PR-Reviewer-Ant, line 43: `inject?.trim() ? inject : null` passes the truthiness gate on the trimmed value but returns the original untrimmed string as `additionalContext`. A string consisting entirely of whitespace is correctly normalised to `null`, but a non-empty string with leading/trailing whitespace (e.g. `" hello "`) is emitted verbatim. This is unlikely to cause correctness problems today but diverges from the apparent intent of normalising context strings before injection.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2338a692`

[Automated Review — pr-reviewer bot] PR #476 incremental update (0bef58a0 → 2338a692) removes the Windows-only guard from resolveSignetCommand() and resolveSignetMcp(), making quoted absolute-path command construction active on all platforms. Three previously flagged items remain unaddressed.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff is a single file change (packages/connector-codex/src/index.ts) that removes `if (process.platform !== 'win32') return 'signet'` guards. The quoted-string construction risk is directly visible at the new resolveSignetCommand() body. The three carry-forward issues (--json/codex flag conflict at hook.ts:155+177, inject trim at hook.ts:43, missing registerHookCommands integration test) are verifiable in the full hook.ts and hook.test.ts file contents provided.

packages/cli/src/commands/hook.test.ts
PR-Reviewer-Ant, line 191: The new Codex hook tests (lines 191–228) cover the builder functions `buildCodexSessionStartOutput` and `buildCodexUserPromptSubmitOutput` only. The actual command-dispatch path in `registerHookCommands` — the `else if (options.harness === 'codex')` branches added in hook.ts — is never exercised by an end-to-end test. Compare with the existing `hook command uses daemon result transport for user-prompt-submit` test (line ~168) which drives `registerHookCommands` for the `claude-code` path. A parallel test for `-H codex` calling `registerHookCommands` and asserting the `{ continue, hookSpecificOutput }` JSON on stdout would close this gap.

aaf2tbz:
Superseded by #480, which includes the wrapper-based Codex fallback that restores session-start, prompt-submit, and stop hook handling when Codex skips native hooks.

PR #480, opened 2026-04-07
Fix Codex hook fallback integration

aaf2tbz:
## Summary
- keep the native Codex hook and MCP fixes from the earlier branch, but add a Signet-managed `~/.local/bin/codex` wrapper fallback for Codex builds that skip native hooks entirely
- install a `codex-signet-watch.js` sidecar that tails the new Codex session JSONL and replays `user-prompt-submit` through the real Signet hook path with explicit fallback logging
- preserve repo-owned `AGENTS.md` files while generating fallback workspace instructions only for unmanaged workspaces, and add regression coverage for wrapper install/uninstall safety
- update Codex harness docs to describe the native-hook path and the wrapper fallback path honestly

## Validation
- `cd packages/cli && bun test ./src/commands/hook.test.ts`
- `cd packages/connector-codex && bun test ./src/config-toml.test.ts`
- `cd packages/connector-codex && bun run typecheck`
- `cd packages/connector-codex && bun run build`

## Notes
- This supersedes #476. The previous PR fixed the Codex hook wire format and packaged entrypoint resolution, but it did not address the reproduced Codex 0.118.0 behavior where fresh processes could skip native hooks entirely.
- The wrapper fallback is explicit in the daemon logs: it emits a fallback marker, then runs the real `session-start`, `user-prompt-submit`, and `session-end` Signet hook handlers where possible.
- Prompt-submit fallback currently depends on Codex continuing to write session JSONL with the current `response_item`/`message` user-turn shape.

## PR Readiness Checklist

### DIFF
--- a/packages/connector-codex/src/index.ts
+++ b/packages/connector-codex/src/index.ts
@@ -8,24 +8,30 @@ import { expandHome } from "@signet/core";
 // Signet command resolution
 // ---------------------------------------------------------------------------
 
-/** Resolve signet command for hook invocation. Returns shell-ready string form for hooks.json.
- *  Windows: navigates from argv[1] (e.g. <pkg>/bin/signet.js) up two levels to find
- *  the bin directory. Falls back to bare "signet" if the layout doesn't match (shims, junctions). */
-function resolveSignetCommand(): string {
-	if (process.platform !== "win32") return "signet";
+function resolvePackagedBin(relativePaths: string[]): string | null {
 	const entry = process.argv[1] || "";
-	const signetJs = join(entry, "..", "..", "bin", "signet.js");
-	if (existsSync(signetJs)) return `"${process.execPath}" "${signetJs}"`;
+	if (!entry) return null;
+	for (const relativePath of relativePaths) {
+		const candidate = join(entry, "..", "..", relativePath);
+		if (existsSync(candidate)) return candidate;
+	}
+	return null;
+}
+
+/** Resolve signet command for hook invocation. Returns shell-ready string form for hooks.json.
+ *  Prefer the installed package's absolute bin path so Codex hooks do not depend on PATH.
+ *  Falls back to bare "signet" only when the packaged layout is unavailable. */
+function resolveSignetCommand(): string {
+	const signetJs = resolvePackagedBin(["bin/signet.js"]);
+	if (signetJs) return `"${process.execPath}" "${signetJs}"`;
 	return "signet";
 }
 
 /** Resolve signet-mcp as { command, args } for Codex config.toml.
  *  Codex expects `command` as a string and `args` as a separate array. */
 function resolveSignetMcp(): { command: string; args: string[] } {
-	if (process.platform !== "win32") return { command: "signet-mcp", args: [] };
-	const entry = process.argv[1] || "";
-	const mcpJs = join(entry, "..", "..", "bin", "mcp-stdio.js");
-	if (existsSync(mcpJs)) return { command: process.execPath, args: [mcpJs] };
+	const mcpJs = resolvePackagedBin(["dist/mcp-stdio.js", "bin/mcp-stdio.js"]);
+	if (mcpJs) return { command: process.execPath, args: [mcpJs] };
 	return { command: "signet-mcp", args: [] };
 }
 
--- a/packages/connector-codex/src/index.ts
+++ b/packages/connector-codex/src/index.ts
@@ -144,7 +144,7 @@ function extractPrompt(record) {
 }
 
 function emitPromptHook(prompt) {
-  appendDaemonLog("Codex native prompt hook unavailable — using wrapper fallback", { prompt });
+  appendDaemonLog("Codex native prompt hook unavailable — using wrapper fallback");
   spawnSync("signet", ["hook", "user-prompt-submit", "-H", "codex", "--project", project], {
     input: JSON.stringify({ cwd: project, prompt, userMessage: prompt }),
     stdio: ["pipe", "ignore", "ignore"],
@@ -213,6 +213,7 @@ readonly SIGNET_LOG_DIR="$SIGNET_WORKSPACE/.daemon/logs"
 
 json_escape() {
   local value="$1"
+  value="$(printf '%s' "$value" | LC_ALL=C tr -d '\\001-\\010\\013\\014\\016-\\037')"
   value="\${value//\\/\\\\}"
   value="\${value//\"/\\\"}"
   value="\${value//$'\\n'/\\n}"
--- a/packages/connector-codex/src/index.ts
+++ b/packages/connector-codex/src/index.ts
@@ -1,4 +1,4 @@
-import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { chmodSync, existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join } from "node:path";
 import { BaseConnector, type InstallResult, type UninstallResult, atomicWriteJson } from "@signet/connector-base";
@@ -13,6 +13,16 @@ function resolvePackagedBin(relativePaths: string[]): string | null {
 	if (!entry) return null;
 	for (const relativePath of relativePaths) {
 		const candidate = join(entry, "..", "..", relativePath);
+		if (existsSync(candidate)) return candidate;
+	}
+	return null;
+}
+
+function findPathExecutable(name: string, excludeDirs: string[] = []): string | null {
+	const pathValue = process.env.PATH || "";
+	for (const dir of pathValue.split(":")) {
+		if (!dir || excludeDirs.includes(dir)) continue;
+		const candidate = join(dir, name);
 		if (existsSync(candidate)) return candidate;
 	}
 	return null;
@@ -39,6 +49,318 @@ function resolveSignetMcp(): { command: string; args: string[] } {
 	const mcpJs = resolvePackagedBin(["dist/mcp-stdio.js", "bin/mcp-stdio.js"]);
 	if (mcpJs) return { command: process.execPath, args: [mcpJs] };
 	return { command: "signet-mcp", args: [] };
+}
+
+const WRAPPER_MARKER = "SIGNET-CODEX-FALLBACK";
+const WRAPPER_LOG_SOURCE = "codex-wrapper-fallback";
+
+function buildWatcherScript(): string {
+	return `#!/usr/bin/env node
+// ${WRAPPER_MARKER}
+
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+const { spawnSync } = require("child_process");
+
+const project = process.argv[2];
+const launchMs = Number(process.argv[3] || Date.now());
+if (!project) process.exit(1);
+
+const sessionsRoot = path.join(os.homedir(), ".codex", "sessions");
+const logDir = path.join(os.homedir(), ".agents", ".daemon", "logs");
+
+let watchedFile = null;
+let lineCount = 0;
+
+function walk(dir, results = []) {
+  if (!fs.existsSync(dir)) return results;
+  for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+    const full = path.join(dir, entry.name);
+    if (entry.isDirectory()) walk(full, results);
+    else if (entry.isFile() && entry.name.endsWith(".jsonl")) results.push(full);
+  }
+  return results;
+}
+
+function latestSessionFile() {
+  const files = walk(sessionsRoot);
+  let best = null;
+  let bestMtime = -1;
+  for (const file of files) {
+    const stat = fs.statSync(file);
+    if (stat.mtimeMs + 1000 < launchMs) continue;
+    try {
+      const firstLine = fs.readFileSync(file, "utf8").split("\\n", 1)[0];
+      const record = JSON.parse(firstLine);
+      const startedAt = Date.parse(record?.payload?.timestamp || "");
+      if (!Number.isFinite(startedAt) || startedAt + 1000 < launchMs) continue;
+    } catch {
+      continue;
+    }
+    if (stat.mtimeMs > bestMtime) {
+      best = file;
+      bestMtime = stat.mtimeMs;
+    }
+  }
+  return best;
+}
+
+function appendDaemonLog(message, extra = {}) {
+  fs.mkdirSync(logDir, { recursive: true });
+  const stamp = new Date().toISOString();
+  const file = path.join(logDir, \`signet-\${stamp.slice(0, 10)}.log\`);
+  const line = {
+    timestamp: stamp,
+    level: "info",
+    category: "hooks",
+    message,
+    data: {
+      harness: "codex",
+      project,
+      source: "${WRAPPER_LOG_SOURCE}",
+      nativeHooks: false,
+      ...extra,
+    },
+  };
+  fs.appendFileSync(file, \`\${JSON.stringify(line)}\\n\`);
+}
+
+function extractPrompt(record) {
+  if (record?.type !== "response_item") return null;
+  const payload = record.payload;
+  if (payload?.type !== "message" || payload?.role !== "user") return null;
+  const texts = [];
+  for (const item of payload.content || []) {
+    if (item?.type === "input_text" && typeof item.text === "string") texts.push(item.text);
+  }
+  if (texts.length === 0) return null;
+  const text = texts.join("\\n").trim();
+  if (!text) return null;
+  if (text.includes("<INSTRUCTIONS>")) return null;
+  if (text.startsWith("<environment_context>")) return null;
+  return text;
+}
+
+function emitPromptHook(prompt) {
+  appendDaemonLog("Codex native prompt hook unavailable — using wrapper fallback", { prompt });
+  spawnSync("signet", ["hook", "user-prompt-submit", "-H", "codex", "--project", project], {
+    input: JSON.stringify({ cwd: project, prompt, userMessage: prompt }),
+    stdio: ["pipe", "ignore", "ignore"],
+  });
+}
+
+function poll() {
+  if (!watchedFile) {
+    watchedFile = latestSessionFile();
+    if (!watchedFile) return;
+  }
+
+  let content;
+  try {
+    content = fs.readFileSync(watchedFile, "utf8");
+  } catch {
+    return;
+  }
+
+  const lines = content.split("\\n").filter(Boolean);
+  if (lines.length <= lineCount) return;
+  for (const line of lines.slice(lineCount)) {
+    try {
+      const record = JSON.parse(line);
+      const prompt = extractPrompt(record);
+      if (prompt) emitPromptHook(prompt);
+    } catch {
+      // Ignore malformed lines while the file is still being written.
+    }
+  }
+  lineCount = lines.length;
+}
+
+setInterval(poll, 250);
+
+process.on("SIGTERM", () => {
+  poll();
+  process.exit(0);
+});
+
+process.on("SIGINT", () => {
+  poll();
+  process.exit(0);
+});
+`;
+}
+
+function buildCodexWrapperScript(realCodexPath: string, watcherPath: string): string {
+	return `#!/bin/zsh
+
+set -euo pipefail
+
+readonly REAL_CODEX=${tomlQuote(realCodexPath).replace(/^'|'$/g, '"')}
+readonly WATCHER_BIN=${tomlQuote(watcherPath).replace(/^'|'$/g, '"')}
+readonly SIGNET_WORKSPACE="\${SIGNET_WORKSPACE:-$HOME/.agents}"
+readonly MARKER_START="<!-- ${WRAPPER_MARKER}:START -->"
+readonly MARKER_END="<!-- ${WRAPPER_MARKER}:END -->"
+readonly SIGNET_LOG_DIR="$HOME/.agents/.daemon/logs"
+
+daemon_log_path() {
+  printf '%s/signet-%s.log\\n' "$SIGNET_LOG_DIR" "$(date +%F)"
+}
+
+append_daemon_log() {
+  local message="$1"
+  local project="$2"
+  local log_path payload
+
+  mkdir -p "$SIGNET_LOG_DIR"
+  log_path="$(daemon_log_path)"
+  payload="$(printf '{"timestamp":"%s","level":"info","category":"hooks","message":"%s","data":{"harness":"codex","project":"%s","source":"${WRAPPER_LOG_SOURCE}","nativeHooks":false}}' \\
+    "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \\
+    "$message" \\
+    "$project")"
+  printf '%s\\n' "$payload" >>"$log_path"
+}
+
+trigger_session_start() {
+  local project="$1"
+
+  append_daemon_log "Codex native hooks unavailable — using wrapper fallback" "$project"
+  signet hook session-start -H codex --project "$project" >/dev/null 2>/dev/null || true
+}
+
+trigger_session_end() {
+  local project="$1"
+
+  append_daemon_log "Codex native stop hook unavailable — using wrapper fallback" "$project"
+  signet hook session-end -H codex >/dev/null 2>/dev/null || true
+}
+
+workspace_root() {
+  local root
+  if root="$(git rev-parse --show-toplevel 2>/dev/null)"; then
+    printf '%s\\n' "$root"
+  else
+    pwd
+  fi
+}
+
+compose_block() {
+  local source content
+
+  printf '%s\\n' "$MARKER_START"
+  printf '%s\\n\\n' "# Signet Codex Fallback"
+  printf '%s\\n\\n' "This block is generated by the local Signet Codex wrapper because Codex native hooks are unreliable on this machine."
+  printf '%s\\n\\n' "Treat the following content as your persistent identity baseline for this session."
+
+  for source in AGENTS.md SOUL.md IDENTITY.md USER.md MEMORY.md; do
+    if [[ ! -f "$SIGNET_WORKSPACE/$source" ]]; then
+      continue
+    fi
+    content="$(<"$SIGNET_WORKSPACE/$source")"
+    content="\${content%"\${content##*[!$'\\n']}"}"
+    if [[ -z "$content" ]]; then
+      continue
+    fi
+    printf '%s\\n\\n' "## $source"
+    printf '%s\\n\\n' "$content"
+  done
+
+  printf '%s\\n' "$MARKER_END"
+}
+
+ensure_git_ignored() {
+  local root="$1"
+  local exclude
+
+  if [[ ! -d "$root/.git" ]]; then
+    return 0
+  fi
+
+  if git -C "$root" ls-files --error-unmatch AGENTS.md >/dev/null 2>&1; then
+    return 0
+  fi
+
+  exclude="$root/.git/info/exclude"
+  mkdir -p "\${exclude:h}"
+  touch "$exclude"
+  if ! grep -Fxq "AGENTS.md" "$exclude"; then
+    printf '\\n# Signet Codex fallback\\nAGENTS.md\\n' >>"$exclude"
+  fi
+}
+
+sync_agents_file() {
+  local root="$1"
+  local target="$root/AGENTS.md"
+  local block existing
+
+  block="$(compose_block)"
+
+  if [[ ! -f "$target" ]]; then
+    printf '%s\\n' "$block" >"$target"
+    ensure_git_ignored "$root"
+    return 0
+  fi
+
+  existing="$(<"$target")"
+
+  if [[ "$existing" == *"$MARKER_START"* && "$existing" == *"$MARKER_END"* ]]; then
+    existing="\${existing%%$MARKER_START*}"
+    existing="\${existing%"\${existing##*[!$'\\n']}"}"
+    if [[ -n "$existing" ]]; then
+      printf '%s\\n\\n%s\\n' "$existing" "$block" >"$target"
+    else
+      printf '%s\\n' "$block" >"$target"
+    fi
+    ensure_git_ignored "$root"
+    return 0
+  fi
+
+  if git -C "$root" rev-parse --show-toplevel >/dev/null 2>&1; then
+    if git -C "$root" ls-files --error-unmatch AGENTS.md >/dev/null 2>&1; then
+      return 0
+    fi
+  fi
+}
+
+main() {
+  local root launch_ms watcher_pid child_pid exit_code
+  root="$(workspace_root)"
+  launch_ms="$(python3 - <<'PY'
+import time
+print(int(time.time() * 1000))
+PY
+)"
+
+  if [[ -d "$SIGNET_WORKSPACE" ]]; then
+    sync_agents_file "$root"
+  fi
+
+  trigger_session_start "$root"
+  if [[ -x "$WATCHER_BIN" ]]; then
+    "$WATCHER_BIN" "$root" "$launch_ms" >/dev/null 2>/dev/null &
+    watcher_pid=$!
+  else
+    watcher_pid=""
+  fi
+
+  "$REAL_CODEX" "$@" &
+  child_pid=$!
+  trap 'kill -TERM "$child_pid" >/dev/null 2>&1 || true' INT TERM
+  wait "$child_pid"
+  exit_code=$?
+
+  if [[ -n "\${watcher_pid:-}" ]]; then
+    sleep 1
+    kill -TERM "$watcher_pid" >/dev/null 2>&1 || true
+    wait "$watcher_pid" >/dev/null 2>&1 || true
+  fi
+
+  trigger_session_end "$root"
+  exit "$exit_code"
+}
+
+main "$@"
+`;
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `Fix Codex hook fallback integration` (commit `c8278f5d`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `Fix Codex hook fallback integration` (commit `be9663ad`) now and I'll follow up shortly with feedback.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,404 @@ import { expandHome } from "@signet/core";
...
+  let content;
+  try {
+    content = fs.readFileSync(watchedFile, "utf8");
+  } catch {
+    return;
+  }
+
+  const lines = content.split("\\n").filter(Boolean);
+  if (lines.length <= lineCount) return;
+  for (const line of lines.slice(lineCount)) {
```
PR-Reviewer-Ant, line 170: **[BLOCKING] Shell injection via misuse of `tomlQuote` for shell string generation.**

```typescript
readonly REAL_CODEX=${tomlQuote(realCodexPath).replace(/^'|'$/g, '"')}
readonly WATCHER_BIN=${tomlQuote(watcherPath).replace(/^'|'$/g, '"')}
```

`tomlQuote` is a TOML literal-string escaper, not a shell escaper. For paths without single-quotes it returns `'path'`; the `.replace` converts the outer single-quotes to double-quotes, producing `readonly REAL_CODEX="path"`. In zsh (and POSIX shell), double-quoted strings expand `$`, `$(...)`, and backticks. A `realCodexPath` value like `/opt/$(evil-cmd)/bin/codex` would execute `evil-cmd` when the wrapper runs.

`realCodexPath` comes from `findPathExecutable("codex", [localBinDir])`, which walks `process.env.PATH`. If `PATH` has been manipulated (e.g. via a malicious `.env`, a compromised shell profile, or a CI environment variable), this becomes a command injection vector at install time.

Fix: use POSIX single-quote escaping for shell embedding — wrap the value in `'...'` and replace any embedded `'` with `'\''`. `shellDoubleQuote` is already defined in this file and handles `"` escaping, but it still does not escape `$` or backticks — do not use it here either. A correct helper would be:
```typescript
function shellSingleQuote(s: string): string {
  return `'${s.replace(/'/g, "'\\''")}' `;
}
```

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,404 @@ import { expandHome } from "@signet/core";
...
+  const line = {
+    timestamp: stamp,
+    level: "info",
+    category: "hooks",
+    message,
+    data: {
+      harness: "codex",
+      project,
+      source: "${WRAPPER_LOG_SOURCE}",
+      nativeHooks: false,
```
PR-Reviewer-Ant, line 123: **[WARNING] `spawnSync` with no timeout in `emitPromptHook` inside `buildWatcherScript`.**

```js
spawnSync("signet", ["hook", "user-prompt-submit", "-H", "codex", "--project", project], {
  input: JSON.stringify({ cwd: project, prompt, userMessage: prompt }),
  stdio: ["pipe", "ignore", "ignore"],
});
```

The watcher polls every 250 ms and calls `emitPromptHook` synchronously on each new user turn. `spawnSync` has no `timeout` option set. If the signet daemon is hanging (e.g. blocked on a slow LLM call, database lock, or network issue), each call to `emitPromptHook` blocks the poll loop indefinitely, stalling all future prompt detection for the session. The daemon client in `packages/cli/src/commands/hook.ts` already uses `SESSION_START_TIMEOUT_MS`; apply a similar budget here — `timeout: 8000` (8 s) would match the user-prompt-submit daemon budget.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,404 @@ import { expandHome } from "@signet/core";
...
+
+readonly REAL_CODEX=${tomlQuote(realCodexPath).replace(/^'|'$/g, '"')}
+readonly WATCHER_BIN=${tomlQuote(watcherPath).replace(/^'|'$/g, '"')}
+readonly SIGNET_WORKSPACE="\${SIGNET_WORKSPACE:-$HOME/.agents}"
+readonly MARKER_START="<!-- ${WRAPPER_MARKER}:START -->"
+readonly MARKER_END="<!-- ${WRAPPER_MARKER}:END -->"
+readonly SIGNET_LOG_DIR="$HOME/.agents/.daemon/logs"
+
+daemon_log_path() {
+  printf '%s/signet-%s.log\\n' "$SIGNET_LOG_DIR" "$(date +%F)"
```
PR-Reviewer-Ant, line 226: **[WARNING] JSON injection in bash `append_daemon_log` via unescaped `$message` and `$project`.**

```bash
payload="$(printf '{"timestamp":"%s","level":"info","category":"hooks","message":"%s","data":{..."project":"%s",...}}' \\
    "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \\
    "$message" \\
    "$project")"
```

`$message` and `$project` are interpolated directly into a JSON string without escaping. A project path containing `"` (e.g. `/home/user/project "name"`) would corrupt the JSON structure in the log file. If the log is later machine-parsed by the daemon, a crafted project path could inject arbitrary JSON fields. At minimum this breaks log readability; in a structured-log pipeline it is a data-integrity issue.

The same issue exists in `buildCodexWrapperScript`'s analogous `append_daemon_log` bash function.

Fix: strip or escape `"` from user-controlled values before `printf`, or use a minimal JSON encoder in the bash script.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,404 @@ import { expandHome } from "@signet/core";
...
+  lineCount = lines.length;
+}
+
+setInterval(poll, 250);
+
+process.on("SIGTERM", () => {
+  poll();
+  process.exit(0);
+});
+
```
PR-Reviewer-Ant, line 188: **[WARNING] `SIGNET_LOG_DIR` ignores the configurable `SIGNET_WORKSPACE`.**

```bash
readonly SIGNET_WORKSPACE="${SIGNET_WORKSPACE:-$HOME/.agents}"
readonly SIGNET_LOG_DIR="$HOME/.agents/.daemon/logs"
```

`SIGNET_WORKSPACE` is correctly derived from the environment variable, but `SIGNET_LOG_DIR` is hardcoded to `$HOME/.agents/.daemon/logs`. For users with a custom workspace (e.g. `SIGNET_WORKSPACE=/data/agents`), wrapper logs always land in the default path while all other daemon logs go to the custom path. This silently splits logs across two locations.

Fix: `readonly SIGNET_LOG_DIR="$SIGNET_WORKSPACE/.daemon/logs"`.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,404 @@ import { expandHome } from "@signet/core";
...
+  local root
+  if root="$(git rev-parse --show-toplevel 2>/dev/null)"; then
+    printf '%s\\n' "$root"
+  else
+    pwd
+  fi
+}
+
+compose_block() {
+  local source content
```
PR-Reviewer-Ant, line 275: **[WARNING] Undeclared `python3` runtime dependency.**

The `main()` function in `buildCodexWrapperScript` uses a here-doc to invoke `python3` for millisecond timestamps:
```bash
launch_ms="$(python3 - <<'PY'
import time
print(int(time.time() * 1000))
PY
)"
```

`python3` is not always available (e.g. minimal Docker images, Windows Subsystem for Linux minimal installs, some CI environments). The Prerequisites section in `docs/HARNESSES.md` does not mention it. If `python3` is absent, `launch_ms` becomes an empty string, `Number("")` in the watcher equals `0`, and `launchMs` filtering (`stat.mtimeMs + 1000 < launchMs`) is effectively disabled — the watcher may then pick up stale session files from past runs.

Alternative that avoids the dependency: `launch_ms="$(date +%s)000"` (seconds × 1000, acceptable 1-second granularity for the filtering purpose).

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,404 @@ import { expandHome } from "@signet/core";
...
+  fi
+
+  exclude="$root/.git/info/exclude"
+  mkdir -p "\${exclude:h}"
+  touch "$exclude"
+  if ! grep -Fxq "AGENTS.md" "$exclude"; then
+    printf '\\n# Signet Codex fallback\\nAGENTS.md\\n' >>"$exclude"
+  fi
+}
+
```
PR-Reviewer-Ant, line 317: **[NITPICK] `sync_agents_file` silently falls through without logging when AGENTS.md exists without a Signet marker and is untracked in a git repo.**

The current logic:
1. File absent → write the block. ✓
2. File has Signet markers → update in place. ✓
3. File is git-tracked → skip (preserve user content). ✓
4. File exists, no marker, untracked in git **or** non-git directory → falls through with no action and no log entry.

For case 4, the wrapper calls `trigger_session_start` (hook fires) but AGENTS.md is never populated with identity context, so the `hookSpecificOutput.additionalContext` returned to the model is effectively a no-op. No fallback marker is emitted to the daemon log. This is a silent partial failure that is difficult to diagnose.

At minimum, emit a daemon log entry in case 4 so operators can see that identity injection was skipped due to a conflicting AGENTS.md.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `c8278f5d`

[Automated Bot Review — pr-reviewer] PR #480 introduces a zsh wrapper fallback and Node.js sidecar watcher for Codex hook integration. One blocking security issue identified: shell metacharacter injection in the generated wrapper script via misuse of the TOML quoting helper. Three additional warnings cover an undeclared runtime dependency, a blocking I/O call without timeout protection, and a JSON log-injection surface. One maintainability gap flagged.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The shell injection is directly provable from the diff: `buildCodexWrapperScript` in `packages/connector-codex/src/index.ts` calls `tomlQuote(realCodexPath).replace(/^'|'$/g, '"')` to embed paths into a zsh `readonly` assignment. `tomlQuote` escapes for TOML, not for POSIX shell; it does not escape `$`, backticks, or `$(...)`. The `.replace` strips the TOML literal-string single-quotes and wraps in double-quotes, producing a shell double-quoted context where all those metacharacters expand at runtime. The `spawnSync`-without-timeout issue is equally visible in `buildWatcherScript`'s `emitPromptHook`. The JSON injection is in the generated bash `append_daemon_log` function. No runtime reproduction is needed — all three are provable from the static diff alone.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,422 @@ import { expandHome } from "@signet/core";
...
+  block="$(compose_block)"
+
+  if [[ ! -f "$target" ]]; then
+    printf '%s\\n' "$block" >"$target"
+    ensure_git_ignored "$root"
+    return 0
+  fi
+
+  existing="$(<"$target")"
+
```
PR-Reviewer-Ant, line 332: **Blocking: hard python3 dependency with no fallback.** The wrapper script's `main()` runs `launch_ms="$(python3 - <<'PY'...)"` under `set -euo pipefail`. If `python3` is not in PATH (common on minimal Linux images, some macOS setups before Homebrew, or NVM-only Node environments), the wrapper exits here with a non-zero status *before launching `$REAL_CODEX`*. This means a user who installs the Signet Codex connector on a machine without `python3` will find that `codex` itself no longer works — the wrapper completely blocks it.

The only reason millisecond precision is needed is to pass `launchMs` to the watcher, which already has its own fallback: `Number(process.argv[3] || Date.now())`. Use a shell arithmetic fallback instead:
```bash
launch_ms=$(( $(date +%s) * 1000 ))
```
This is POSIX-portable and removes the python3 dependency entirely. The 1-second granularity is sufficient for the timestamp comparison in `latestSessionFile`.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,422 @@ import { expandHome } from "@signet/core";
...
+  process.exit(0);
+});
+
+process.on("SIGINT", () => {
+  poll();
+  process.exit(0);
+});
+`;
+}
+
```
PR-Reviewer-Ant, line 195: **Warning: `tomlQuote()` is repurposed for shell path quoting — wrong escaping context.** `buildCodexWrapperScript` embeds `realCodexPath` and `watcherPath` into shell `readonly` assignments using `tomlQuote(...).replace(/^'|'$/g, '"')`. `tomlQuote` was designed for TOML literal strings, not POSIX/zsh double-quoted strings. It does not escape `$`, backtick, `!`, or other shell metacharacters. A `realCodexPath` containing `$` (e.g., from a PATH entry like `/opt/$TOOL/bin`) would undergo variable expansion at shell runtime.

`watcherPath` is always `$HOME/.local/bin/codex-signet-watch.js` so it's safe in practice, but `realCodexPath` comes from `findPathExecutable` which iterates `process.env.PATH` — fully user-controlled. Use a dedicated shell single-quote escaper instead (escape internal single quotes as `'\''`).

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,422 @@ import { expandHome } from "@signet/core";
...
+readonly MARKER_START="<!-- ${WRAPPER_MARKER}:START -->"
+readonly MARKER_END="<!-- ${WRAPPER_MARKER}:END -->"
+readonly SIGNET_LOG_DIR="$SIGNET_WORKSPACE/.daemon/logs"
+
+json_escape() {
+  local value="$1"
+  value="\${value//\\/\\\\}"
+  value="\${value//\"/\\\"}"
+  value="\${value//$'\\n'/\\n}"
+  value="\${value//$'\\r'/\\r}"
```
PR-Reviewer-Ant, line 220: **Warning: shell `printf %s` constructs JSON without escaping — log corruption on paths with special characters.** The `append_daemon_log` function in the shell wrapper template builds a JSON log line via `printf '...%s...' "$message" "$project"`. Neither `$message` nor `$project` are JSON-escaped, so a project path containing `"`, `\`, or a newline (all legal in Unix paths) will produce malformed JSON, breaking any log consumer that parses structured log lines. The Node.js `appendDaemonLog` function in the watcher correctly uses `JSON.stringify`; the shell equivalent should write the log entry through a tool that can JSON-encode the fields, or at minimum escape `"` and `\` in the values before interpolation.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,422 @@ import { expandHome } from "@signet/core";
...
+  if (payload?.type !== "message" || payload?.role !== "user") return null;
+  const texts = [];
+  for (const item of payload.content || []) {
+    if (item?.type === "input_text" && typeof item.text === "string") texts.push(item.text);
+  }
+  if (texts.length === 0) return null;
+  const text = texts.join("\\n").trim();
+  if (!text) return null;
+  if (text.includes("<INSTRUCTIONS>")) return null;
+  if (text.startsWith("<environment_context>")) return null;
```
PR-Reviewer-Ant, line 142: **Warning: raw user prompt text is written to daemon log files.** `emitPromptHook` calls `appendDaemonLog("...", { prompt })`, which persists the full prompt string to `~/.agents/.daemon/logs/signet-YYYY-MM-DD.log`. For users working in sensitive contexts (proprietary code, credentials, PII), this creates an unexpected on-disk record of every submitted prompt. The native hook path does not log prompt content. Consider omitting the prompt field from the log entry or replacing it with a character-count/hash.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,422 @@ import { expandHome } from "@signet/core";
...
+  return text;
```
PR-Reviewer-Ant, line 143: **Warning: `spawnSync` in the watcher's poll loop blocks indefinitely with no timeout.** Each call to `emitPromptHook` issues a synchronous `signet hook user-prompt-submit` via `spawnSync` with no `timeout` option. If the Signet daemon is slow, hung, or the hook times out at the OS level, the 250 ms poll interval stalls for the full duration of the subprocess. Under repeated prompts this compounds. Add a `timeout` to the `spawnSync` call (e.g., `timeout: 8000`) so the watcher recovers if the daemon is unresponsive.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,422 @@ import { expandHome } from "@signet/core";
...
+      continue
+    fi
+    content="$(<"$SIGNET_WORKSPACE/$source")"
+    content="\${content%"\${content##*[!$'\\n']}"}"
+    if [[ -z "$content" ]]; then
+      continue
+    fi
+    printf '%s\\n\\n' "## $source"
+    printf '%s\\n\\n' "$content"
+  done
```
PR-Reviewer-Ant, line 293: **Nitpick: `sync_agents_file` silently skips writing the Signet block for unmanaged non-git workspaces.** When `AGENTS.md` exists without a Signet marker and the directory is not a git repo (or the file is not git-tracked), the function falls through with no action. The PR description states the wrapper generates fallback instructions "for unmanaged workspaces", but a workspace outside of any git repo with a pre-existing unmarked `AGENTS.md` receives no Signet block — contradicting that claim. Whether this is intentional (conservative: never overwrite user content) or a gap should be clarified in the code or docs.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `be9663ad`

[Automated Bot Review] Found one blocking bug (python3 hard dependency in fallback wrapper) and four warnings (shell JSON injection, incorrect quoting function reuse, user prompt logging, blocking spawnSync). This review was produced by the automated pr-reviewer identity.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The python3 dependency is explicit in the buildCodexWrapperScript template literal (index.ts ~line 323-328): `launch_ms="$(python3 - <<'PY'...)"` with set -euo pipefail active and no fallback — provable from the diff that this exits the wrapper before launching codex when python3 is absent. The printf-based JSON construction in the shell append_daemon_log function (index.ts ~line 217-220) and the tomlQuote reuse for shell path quoting (index.ts ~line 199-200) are both directly readable from the diff. The spawnSync-with-no-timeout and prompt-logging findings are in the buildWatcherScript template (index.ts ~line 145-151). No runtime trace is needed to confirm these.

packages/connector-codex/src/index.ts
```diff
@@ -234,6 +601,56 @@ export class CodexConnector extends BaseConnector {
...
+		mkdirSync(localBinDir, { recursive: true });
+
+		if (existsSync(wrapperPath) && !this.isManagedWrapper(wrapperPath)) {
+			warnings.push(`Skipped installing Codex wrapper at ${wrapperPath} — existing file is not Signet-managed`);
+		} else {
+			writeFileSync(wrapperPath, buildCodexWrapperScript(realCodexPath, watcherPath), "utf-8");
+			chmodSync(wrapperPath, 0o755);
+			filesWritten.push(wrapperPath);
+		}
+
```
PR-Reviewer-Ant, line 652: **[BLOCKING] Wrapper installed with `REAL_CODEX='codex'` when no real Codex binary is found, causing infinite self-invocation at runtime.**

`findPathExecutable("codex", [localBinDir])` searches PATH excluding `~/.local/bin`. If Codex is not installed yet — or is installed only inside `~/.local/bin` on a re-install — it returns `null` and `realCodexPath` falls back to the bare string `"codex"`. The wrapper is still written with:
```
readonly REAL_CODEX='codex'
```
At runtime, `"$REAL_CODEX" "$@" &` resolves `codex` through the child process's PATH. On the majority of Linux setups `~/.local/bin` is prepended to PATH (e.g. via `.profile`), so the shell finds the Signet wrapper again → infinite recursion. The user ends up with a broken `codex` command that loops until the shell hits its fork/recursion limit.

**Fix:** if `findPathExecutable` returns `null`, skip wrapper installation and emit a warning (`"Cannot install Codex wrapper — no real codex binary found in PATH outside ~/.local/bin"`) rather than silently falling back to a bare command that resolves to the wrapper itself.

```typescript
const realCodexPath = findPathExecutable("codex", [localBinDir]);
if (!realCodexPath) {
  warnings.push("Skipped Codex wrapper install — codex binary not found in PATH; run `signet connect codex` again after installing Codex");
  return; // or continue to watcher-only install
}
```

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,424 @@ import { expandHome } from "@signet/core";
...
+readonly MARKER_START="<!-- ${WRAPPER_MARKER}:START -->"
+readonly MARKER_END="<!-- ${WRAPPER_MARKER}:END -->"
+readonly SIGNET_LOG_DIR="$SIGNET_WORKSPACE/.daemon/logs"
+
+json_escape() {
+  local value="$1"
+  value="\${value//\\/\\\\}"
+  value="\${value//\"/\\\"}"
+  value="\${value//$'\\n'/\\n}"
+  value="\${value//$'\\r'/\\r}"
```
PR-Reviewer-Ant, line 220: **[NITPICK] `json_escape` in the generated bash wrapper covers only five escape categories, leaving JSON control characters U+0001–U+001F unescaped.**

The function handles `\`, `"`, `\n`, `\r`, `\t`, which covers the common cases for log messages and filesystem paths. However, JSON requires that all code points below U+0020 be escaped. If a project path or log message ever contains a character in that range (e.g. a bell `\x07` in a contrived path), the resulting log line would be invalid JSON. In practice this is extremely unlikely for the values passed (`$message` is a hardcoded string; `$project` is a git root or `pwd`). Flagging for completeness; not a correctness risk under current usage.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `e91be1d1`

[Automated Review — pr-reviewer] All six previously flagged issues from sha=c8278f5d are addressed in the incremental diff. One new blocking issue introduced: when `findPathExecutable` returns null the wrapper is installed with `REAL_CODEX='codex'`, which at runtime resolves via PATH to the wrapper itself, causing infinite recursion. One nitpick remains on incomplete JSON control-character escaping.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The infinite-recursion risk is directly provable from `installWrapperFiles` in the full `packages/connector-codex/src/index.ts`: line `const realCodexPath = findPathExecutable("codex", [localBinDir]) || "codex"` feeds into `buildCodexWrapperScript(realCodexPath, watcherPath)` which embeds it as `readonly REAL_CODEX=${shellSingleQuote(realCodexPath)}`. When `findPathExecutable` returns null (Codex not yet installed, or installed only inside `localBinDir`), the wrapper stores the bare string `codex`. At runtime the shell resolves `codex` through PATH; if `~/.local/bin` precedes the real Codex install dir (the common Linux default), the wrapper calls itself. No runtime reproduction is needed — the static fallback path and the absence of any guard are both in the diff. The json_escape gap is visible in the generated bash string inside `buildCodexWrapperScript`: only five escape categories are covered, leaving arbitrary control characters (U+0001–U+001F) unescaped.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,424 @@ import { expandHome } from "@signet/core";
...
+      source: "${WRAPPER_LOG_SOURCE}",
+      nativeHooks: false,
+      ...extra,
+    },
+  };
+  fs.appendFileSync(file, \`\${JSON.stringify(line)}\\n\`);
+}
+
+function extractPrompt(record) {
+  if (record?.type !== "response_item") return null;
```
PR-Reviewer-Ant, line 131: **[Prior warning — still unaddressed]** `emitPromptHook` passes the full user prompt string to `appendDaemonLog` as `{ prompt }`, persisting it to `~/.agents/.daemon/logs/signet-YYYY-MM-DD.log` on every user turn. For users working in sensitive codebases (auth keys in prompts, private specs, etc.) this creates an unbounded plaintext log of session content. The two mitigations already present — `if (text.includes("<INSTRUCTIONS>")) return null` and `if (text.startsWith("<environment_context>")) return null` — filter injected system context but not user-authored messages. Consider omitting `prompt` from the log entry or truncating to a fixed length (e.g. first 80 chars) so the log entry is diagnostic without being a full transcript.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,424 @@ import { expandHome } from "@signet/core";
...
+
+json_escape() {
+  local value="$1"
+  value="\${value//\\/\\\\}"
+  value="\${value//\"/\\\"}"
+  value="\${value//$'\\n'/\\n}"
+  value="\${value//$'\\r'/\\r}"
+  value="\${value//$'\\t'/\\t}"
+  printf '%s' "$value"
+}
```
PR-Reviewer-Ant, line 223: **[Prior nitpick — still unaddressed]** The generated bash `json_escape()` handles `\`, `"`, `\n`, `\r`, `\t` but leaves C0 control characters U+0001–U+001F (other than the three listed) unescaped. A project path or message containing a form-feed (U+000C), vertical tab (U+000B), or any other control character would produce malformed JSON in `append_daemon_log`. In practice this is unlikely but log parsers that validate JSON strictly will reject the line. Either loop over the remaining control characters in the escape function or use `printf '%s' "$value" | LC_ALL=C sed 's/[\x01-\x1f]//g'` as a sanitization step before the escape call.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fb4e0f6c`

[Automated Bot Review — pr-reviewer] This incremental diff (sha e91be1d → fb4e0f6) resolves both previously blocking issues. The shell injection risk is fixed by replacing `shellSingleQuote` path embedding with a quoted heredoc (`<<'__SIGNET_REAL_CODEX__'`), and the infinite-recursion risk is fixed by removing the `|| "codex"` fallback and gating wrapper install on `realCodexPath !== null`. Two prior warnings remain unaddressed: raw user prompt text written to daemon log files, and incomplete JSON control-character escaping in the generated bash `json_escape` function.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both prior blocking findings are verifiably closed in this diff. The heredoc fix (`<<'__SIGNET_REAL_CODEX__'`) is directly visible in the incremental diff at index.ts lines 198-211; `shellSingleQuote` is removed. The null guard (`else if (!realCodexPath)`) is at index.ts line 628. The remaining warning — `appendDaemonLog("...", { prompt })` in `emitPromptHook` inside `buildWatcherScript` — is unchanged from prior reviews and visible in the full file content. The `json_escape` function covering only `\`, `"`, `\n`, `\r`, `\t` is also directly readable from the template literal in `buildCodexWrapperScript`. No runtime trace is needed to confirm either finding.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,424 @@ import { expandHome } from "@signet/core";
...
+  const text = texts.join("\\n").trim();
+  if (!text) return null;
+  if (text.includes("<INSTRUCTIONS>")) return null;
+  if (text.startsWith("<environment_context>")) return null;
+  return text;
+}
+
+function emitPromptHook(prompt) {
+  appendDaemonLog("Codex native prompt hook unavailable — using wrapper fallback", { prompt });
+  spawnSync("signet", ["hook", "user-prompt-submit", "-H", "codex", "--project", project], {
```
PR-Reviewer-Ant, line 148: Prompt text is still written verbatim to the daemon log file on every user turn (`appendDaemonLog("...", { prompt })`). This was flagged as a warning in the prior two reviews and is unchanged in this diff. The full text of every user message is persisted to `~/.agents/.daemon/logs/signet-YYYY-MM-DD.log`. If users treat those logs as shareable debug artifacts, sensitive prompt content leaks. Consider truncating or omitting the `prompt` field from the log entry, or making it opt-in via a debug flag.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,424 @@ import { expandHome } from "@signet/core";
...
+set -euo pipefail
+
+readonly REAL_CODEX="$(cat <<'__SIGNET_REAL_CODEX__'
+${realCodexPath}
+__SIGNET_REAL_CODEX__
+)"
+readonly WATCHER_BIN="$(cat <<'__SIGNET_WATCHER_BIN__'
+${watcherPath}
+__SIGNET_WATCHER_BIN__
+)"
```
PR-Reviewer-Ant, line 208: The new `json_escape` function escapes `\`, `"`, `\n`, `\r`, and `\t` but not other JSON control characters (U+0001–U+001F, excluding the three handled). For the hardcoded English `message` strings this is fine; for the `project` (file path) argument, unusual paths constructed from adversarial PATH entries could in theory contain these bytes. In practice this is extremely unlikely, but for completeness a sed/tr strip of raw control characters below 0x20 (except the five handled) would fully close the JSON corruption surface.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,424 @@ import { expandHome } from "@signet/core";
...
+}
+
+function emitPromptHook(prompt) {
+  appendDaemonLog("Codex native prompt hook unavailable — using wrapper fallback", { prompt });
+  spawnSync("signet", ["hook", "user-prompt-submit", "-H", "codex", "--project", project], {
+    input: JSON.stringify({ cwd: project, prompt, userMessage: prompt }),
+    stdio: ["pipe", "ignore", "ignore"],
+    timeout: 8000,
+  });
+}
```
PR-Reviewer-Ant, line 153: The watcher process (`$WATCHER_BIN`) is launched with `&` and killed with SIGTERM after `sleep 1`. If the wrapper itself receives SIGKILL (which cannot be trapped), the watcher is orphaned and continues polling `~/.codex/sessions/*.jsonl` and calling `signet hook user-prompt-submit` indefinitely — one call per new JSONL line — until the user kills it or reboots. This is an inherent design limitation of the sidecar model; noting it here for awareness. A PID file written by the wrapper that the watcher polls as a liveness check would address it if this becomes a reported issue.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `6105e387`

[Automated Bot Review — pr-reviewer] This incremental diff (be9663ad → 6105e387) addresses all blocking and major warning items from the two prior review rounds. No new blocking issues found. One prior open warning (prompt text in daemon logs) is unchanged; one new nitpick on the json_escape function's coverage is noted.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All prior blocking items are directly verifiable as resolved from the incremental diff: (1) shell injection via tomlQuote → replaced with single-quoted heredoc assignments that isolate the path literal from shell expansion (index.ts lines 198–215); (2) python3 hard dependency under set -euo pipefail → replaced with POSIX $(( $(date +%s) * 1000 )) (index.ts line 338); (3) SIGNET_LOG_DIR hardcoded to $HOME/.agents → now uses $SIGNET_WORKSPACE (index.ts line 206); (4) JSON injection in append_daemon_log → new json_escape function escapes backslash, double-quote, \n, \r, \t before printf (index.ts lines 208–244); (5) spawnSync with no timeout → timeout: 8000 added (index.ts line 150); (6) SIGNET_WORKSPACE not forwarded to watcher → $SIGNET_WORKSPACE now passed as argv[4] (index.ts line 357). The prompt-text-in-logs item is confirmed unchanged by grep of the emitPromptHook function. The json_escape coverage gap is directly readable from the function definition.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,425 @@ import { expandHome } from "@signet/core";
...
+  if [[ -d "$SIGNET_WORKSPACE" ]]; then
+    sync_agents_file "$root"
+  fi
+
+  trigger_session_start "$root"
+  if [[ -x "$WATCHER_BIN" ]]; then
+    "$WATCHER_BIN" "$root" "$launch_ms" "$SIGNET_WORKSPACE" >/dev/null 2>/dev/null &
+    watcher_pid=$!
+  else
+    watcher_pid=""
```
PR-Reviewer-Ant, line 369: **Warning: EXIT trap now fires `trigger_session_end` even if Codex never ran.**

The EXIT trap is set at the top of `main()`, before `"$REAL_CODEX" "$@"` is called. Under `set -euo pipefail`, any command that exits non-zero before the `$REAL_CODEX` invocation — for example, a git command inside `sync_agents_file` or a failed `trigger_session_start` (if `signet` is not in PATH and the `|| true` guard is absent for some sub-call) — causes the script to exit, firing the EXIT trap and issuing `signet hook session-end -H codex`.

In the previous implementation, `trigger_session_end` was an unconditional call at the bottom of `main()` after `wait "$child_pid"` completed; it was never reached on early exits. Now it fires on *all* exits, including pre-launch failures.

Impact: the daemon receives a session-end request with no corresponding Codex transcript, which will attempt (and presumably silently no-op) memory extraction on an empty session. The output is discarded (`>/dev/null 2>/dev/null || true`), so no crash occurs, but the daemon logs a spurious session-end event and potentially touches session state unnecessarily.

Mitigation: track whether `$REAL_CODEX` was launched (e.g., a `codex_started=false` flag set to `true` immediately before `"$REAL_CODEX" "$@"`) and gate `trigger_session_end` inside the EXIT trap on that flag.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `031962c0`

[Automated Bot Review — pr-reviewer] Incremental diff (43206596 → 031962c) is a focused signal/exit-handling refactor inside `buildCodexWrapperScript`'s `main()`. All blocking issues from prior rounds are confirmed resolved in the current full-file content. One new warning is introduced by the EXIT trap change.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff touches only the `main()` function of the generated zsh wrapper script (lines 348–382 in the full file). The new behavior — EXIT trap unconditionally running `trigger_session_end` — is directly readable from the diff. All prior blocking items (heredoc shell injection, infinite recursion, python3 dependency, SIGNET_LOG_DIR hardcoding, JSON injection, spawnSync timeout, { prompt } in appendDaemonLog) are confirmed absent from the full file content at HEAD. The new spurious-session-end risk is derivable from the changed trap placement without needing a runtime trace.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,431 @@ import { expandHome } from "@signet/core";
...
+  fi
```
PR-Reviewer-Ant, line 370: EXIT trap still fires `trigger_session_end` unconditionally even when Codex never ran. The trap is registered before `"$REAL_CODEX" "$@"`, so if `sync_agents_file` or `trigger_session_start` causes an early exit under `set -euo pipefail`, session-end fires with `reason="terminated"` despite no real session having occurred. The `set +e` added in this diff ensures the trap body completes without short-circuiting, which is a correctness improvement — but it does not prevent the spurious invocation. A guard like `[[ -n "${exit_code+x}" ]]` (or setting `exit_code` to a sentinel immediately before calling `$REAL_CODEX`) would scope session-end only to sessions that actually started. This has been flagged across prior rounds and remains open.

packages/connector-codex/src/index.ts
```diff
@@ -8,67 +8,431 @@ import { expandHome } from "@signet/core";
...
+
+  trap '
+    set +e
+    trigger_session_end "$root" "\${exit_code:-terminated}"
+    if [[ -n "\${watcher_pid:-}" ]]; then
```
PR-Reviewer-Ant: Reordering `trigger_session_end` before the watcher `sleep 1 / kill -TERM / wait` block is semantically correct — the watcher gets a final second to complete its poll before the session-end hook consumes the transcript. The `set +e` addition ensures watcher cleanup still runs if session-end fails. No new issues introduced by this reorder.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2392cc91`

[Automated Bot Review — pr-reviewer] Final incremental diff (21de5b13 → 2392cc914) is a focused EXIT trap correctness fix. All prior blocking issues verified resolved in the full file at HEAD. One prior open warning (spurious session-end when Codex startup fails) remains present and is marginally affected by the reorder.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff touches only the EXIT trap inside main() of buildCodexWrapperScript. The two changes — adding `set +e` and moving `trigger_session_end` before the watcher kill — are directly readable from the diff. All prior blocking items (heredoc shell injection, infinite recursion, python3 dependency, SIGNET_LOG_DIR hardcoding, JSON injection, spawnSync timeout, prompt text in appendDaemonLog) are verified absent from the full file content at HEAD. The remaining open concern (EXIT trap fires session-end even when Codex never ran) is directly traceable to the unchanged trap placement before `"$REAL_CODEX" "$@"` in the full file.

PR #482, opened 2026-04-08
fix(daemon): honour maxInjectChars config in session-start hook

NicholaiVogel:
## Summary

- Replaces all character-based budgeting in the session-start hook with accurate BPE token counts via `js-tiktoken` (already a daemon dependency)
- Adds `selectWithTokenBudget()` — same interface as `selectWithBudget()` but uses `countTokens()` per row
- Memory selection: was hardcoded 2000 chars, now uses `maxInjectTokens` token budget (default 20000)
- Total inject truncation: was char slice at 24000, now `truncateToTokens()` at 20000 tokens
- New config field: `hooks.sessionStart.maxInjectTokens` (default 20000); clamped to `Math.max(1, ...)` with a warn log if ≤ 0
- Logs `injectTokens` instead of `injectChars` in the session-start completion log
- Regression tests added for `selectWithTokenBudget`: budget enforcement, full pass, zero-budget, and sub-row-budget cases
- `userPromptSubmit.maxInjectChars` already reads from config correctly — no parity gap

**Example `agent.yaml`:**
```yaml
hooks:
  sessionStart:
    maxInjectTokens: 8000
```

## Why

Characters are a poor proxy for model context. A 2000-char memory budget was effectively arbitrary and often under-injected. Token counts via `cl100k_base` give accurate, model-aligned budget enforcement — the same encoding used everywhere else in the daemon.

## Test plan

- [x] `selectWithTokenBudget` unit tests (4 cases, all pass)
- [ ] Restart daemon, verify session-start logs show `injectTokens` field
- [ ] Set `maxInjectTokens: 4000` in `agent.yaml`, start new session, confirm inject is bounded
- [ ] Omit field, verify default 20000-token behaviour

### DIFF
--- a/packages/daemon/src/hooks.test.ts
+++ b/packages/daemon/src/hooks.test.ts
@@ -392,4 +392,14 @@ describe("applyTokenBudget", () => {
 	it("returns empty string when mainBudget is negative", () => {
 		expect(applyTokenBudget(TEXT, -1)).toBe("");
 	});
+
+	it("never exceeds budget when budget is smaller than marker token count", async () => {
+		// Regression: marker is ~5 tokens; budgets in [1, TRUNCATED_MARKER_TOKENS) must
+		// not overflow by appending the full marker after truncation.
+		const { countTokens } = await import("./pipeline/tokenizer");
+		for (const budget of [1, 2, 3, 4]) {
+			const result = applyTokenBudget(TEXT, budget);
+			expect(countTokens(result)).toBeLessThanOrEqual(budget);
+		}
+	});
 });
--- a/packages/daemon/src/hooks.test.ts
+++ b/packages/daemon/src/hooks.test.ts
@@ -9,6 +9,7 @@ import {
 	normalizeJsonConversationTranscript,
 	normalizeSessionTranscript,
 	queryAnchorsMissingFromRecall,
+	selectWithTokenBudget,
 	writeMemoryMd,
 } from "./hooks";
 
@@ -330,5 +331,35 @@ describe("normalizeSessionTranscript", () => {
 			'{"role":"assistant","content":"Ship it by Friday."}',
 		].join("\n");
 		expect(normalizeSessionTranscript("opencode", json)).toBe("User: What's the plan?\nAssistant: Ship it by Friday.");
+	});
+});
+
+describe("selectWithTokenBudget", () => {
+	const rows = [
+		{ content: "alpha ".repeat(50) },   // ~50 tokens
+		{ content: "beta ".repeat(50) },    // ~50 tokens
+		{ content: "gamma ".repeat(200) },  // ~200 tokens
+	];
+
+	it("selects rows up to the token budget", () => {
+		const result = selectWithTokenBudget(rows, 120);
+		expect(result).toHaveLength(2);
+		expect(result[0]).toBe(rows[0]);
+		expect(result[1]).toBe(rows[1]);
+	});
+
+	it("returns all rows when budget is not exceeded", () => {
+		const result = selectWithTokenBudget(rows, 10000);
+		expect(result).toHaveLength(3);
+	});
+
+	it("returns empty array when budget is too small for any row", () => {
+		const result = selectWithTokenBudget(rows, 1);
+		expect(result).toHaveLength(0);
+	});
+
+	it("returns empty array for zero budget", () => {
+		const result = selectWithTokenBudget(rows, 0);
+		expect(result).toHaveLength(0);
 	});
 });
--- a/packages/daemon/src/hooks.test.ts
+++ b/packages/daemon/src/hooks.test.ts
@@ -4,6 +4,7 @@ import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { closeDbAccessor, getDbAccessor, initDbAccessor } from "./db-accessor";
 import {
+	applyTokenBudget,
 	buildSignetSystemPrompt,
 	normalizeCodexTranscript,
 	normalizeJsonConversationTranscript,
@@ -366,5 +367,29 @@ describe("selectWithTokenBudget", () => {
 	it("handles negative budget the same as zero", () => {
 		const result = selectWithTokenBudget(rows, -100);
 		expect(result).toHaveLength(0);
+	});
+});
+
+describe("applyTokenBudget", () => {
+	const TEXT = "word ".repeat(500); // ~500 tokens
+
+	it("returns inject unchanged when it fits within budget", () => {
+		expect(applyTokenBudget("hello world", 1000)).toBe("hello world");
+	});
+
+	it("truncates and appends marker when inject exceeds budget", async () => {
+		const result = applyTokenBudget(TEXT, 50);
+		expect(result).toContain("[context truncated]");
+		// total tokens must not exceed budget (marker tokens pre-subtracted)
+		const { countTokens } = await import("./pipeline/tokenizer");
+		expect(countTokens(result)).toBeLessThanOrEqual(50);
+	});
+
+	it("returns empty string when mainBudget is zero (reserved sections exhausted budget)", () => {
+		expect(applyTokenBudget(TEXT, 0)).toBe("");
+	});
+
+	it("returns empty string when mainBudget is negative", () => {
+		expect(applyTokenBudget(TEXT, -1)).toBe("");
 	});
 });
--- a/packages/daemon/src/hooks.test.ts
+++ b/packages/daemon/src/hooks.test.ts
@@ -362,4 +362,9 @@ describe("selectWithTokenBudget", () => {
 		const result = selectWithTokenBudget(rows, 0);
 		expect(result).toHaveLength(0);
 	});
+
+	it("handles negative budget the same as zero", () => {
+		const result = selectWithTokenBudget(rows, -100);
+		expect(result).toHaveLength(0);
+	});
 });
--- a/packages/daemon/src/hooks.ts
+++ b/packages/daemon/src/hooks.ts
@@ -212,7 +212,11 @@ export interface HooksConfig {
 		recencyBias?: number;
 		query?: string;
 		maxInjectTokens?: number;
-		/** @deprecated renamed to maxInjectTokens — will be ignored */
+		/**
+		 * @deprecated Renamed to `maxInjectTokens`. If set without `maxInjectTokens`,
+		 * the value is auto-migrated using `Math.round(maxInjectChars / 4)` (~4 chars/token
+		 * for ASCII; code or Unicode content may be 1–2 chars/token, so migrate explicitly.
+		 */
 		maxInjectChars?: number;
 	};
 	userPromptSubmit?: {
@@ -504,15 +508,15 @@ const TRUNCATED_MARKER_TOKENS = countTokens(TRUNCATED_MARKER);
 /**
  * Truncate `inject` to fit within `mainBudget` tokens.
  * Returns an empty string when budget is zero (reserved sections exhausted it).
- * Appends a truncation marker whose tokens are pre-subtracted from the budget.
+ * Appends a truncation marker when budget permits; omits it when the budget is
+ * too small to fit the marker itself (avoids overflow in that range).
  */
 export function applyTokenBudget(inject: string, mainBudget: number): string {
 	if (mainBudget <= 0) return "";
 	if (countTokens(inject) <= mainBudget) return inject;
-	return (
-		truncateToTokens(inject, Math.max(1, mainBudget - TRUNCATED_MARKER_TOKENS)) +
-		TRUNCATED_MARKER
-	);
+	// Budget too tight to fit content + marker — truncate without marker
+	if (mainBudget <= TRUNCATED_MARKER_TOKENS) return truncateToTokens(inject, mainBudget);
+	return truncateToTokens(inject, mainBudget - TRUNCATED_MARKER_TOKENS) + TRUNCATED_MARKER;
 }
 
 /** Build a brief "since your last session" summary for temporal awareness */
--- a/packages/daemon/src/hooks.ts
+++ b/packages/daemon/src/hooks.ts
@@ -1474,10 +1474,16 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 	});
 
 	// Apply budget to select what we actually inject (on re-ranked order)
-	if (config.maxInjectChars !== undefined) {
-		logger.warn("hooks", "hooks.sessionStart.maxInjectChars is ignored — rename to maxInjectTokens in agent.yaml");
+	if (config.maxInjectChars !== undefined && config.maxInjectTokens === undefined) {
+		logger.warn(
+			"hooks",
+			"hooks.sessionStart.maxInjectChars is deprecated — migrating to maxInjectTokens automatically. Rename it in agent.yaml to silence this warning.",
+			{ maxInjectChars: config.maxInjectChars, derivedTokens: Math.round(config.maxInjectChars / 4) },
+		);
 	}
-	const rawTokenBudget = config.maxInjectTokens ?? 20000;
+	const rawTokenBudget =
+		config.maxInjectTokens ??
+		(config.maxInjectChars ? Math.round(config.maxInjectChars / 4) : 20000);
 	if (rawTokenBudget <= 0) {
 		logger.warn("hooks", "maxInjectTokens must be positive — clamping to 1", {
 			configured: rawTokenBudget,
@@ -1762,7 +1768,9 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 	}
 
 	const duration = Date.now() - start;
-	const maxTokens = config.maxInjectTokens ?? 20000;
+	const maxTokens =
+		config.maxInjectTokens ??
+		(config.maxInjectChars ? Math.round(config.maxInjectChars / 4) : 20000);
 	// Pre-reserve space for constraints + recovery so they are never truncated
 	const reservedTokens =
 		countTokens(recoverySection) + countTokens(constraintsSection);
--- a/packages/daemon/src/hooks.ts
+++ b/packages/daemon/src/hooks.ts
@@ -1455,7 +1455,13 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 	});
 
 	// Apply budget to select what we actually inject (on re-ranked order)
-	const tokenBudget = config.maxInjectTokens ?? 20000;
+	const rawTokenBudget = config.maxInjectTokens ?? 20000;
+	if (rawTokenBudget <= 0) {
+		logger.warn("hooks", "maxInjectTokens must be positive — clamping to 1", {
+			configured: rawTokenBudget,
+		});
+	}
+	const tokenBudget = Math.max(1, rawTokenBudget);
 	const memories = selectWithTokenBudget(sortedCandidates, tokenBudget);
 
 	// Get predicted context from recent session analysis (~30% of budget)
--- a/packages/daemon/src/hooks.ts
+++ b/packages/daemon/src/hooks.ts
@@ -1438,7 +1438,7 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 	});
 
 	// Apply budget to select what we actually inject (on re-ranked order)
-	const memories = selectWithBudget(sortedCandidates, 2000);
+	const memories = selectWithBudget(sortedCandidates, config.maxInjectChars ?? 2000);
 
 	// Get predicted context from recent session analysis (~30% of budget)
 	const existingIds = new Set(memories.map((m) => m.id));
--- a/packages/daemon/src/hooks.ts
+++ b/packages/daemon/src/hooks.ts
@@ -88,6 +88,7 @@ import { getSessionTranscriptContent, searchTranscriptFallback, upsertSessionTra
 import { type StructuralFeatures, buildCandidateFeatures, getStructuralFeatures } from "./structural-features";
 import { searchTemporalFallback } from "./temporal-fallback";
 import { writeTranscriptAudit } from "./transcript-audit";
+import { countTokens, truncateToTokens } from "./pipeline/tokenizer";
 import { getUpdateSummary } from "./update-system";
 
 function getAgentsDir(): string {
@@ -210,7 +211,7 @@ export interface HooksConfig {
 		includeRecentContext?: boolean;
 		recencyBias?: number;
 		query?: string;
-		maxInjectChars?: number;
+		maxInjectTokens?: number;
 	};
 	userPromptSubmit?: {
 		/** Set to false to disable per-prompt memory injection entirely. Default: true. */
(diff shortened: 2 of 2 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): honour maxInjectChars config in session-start hook` (commit `818b6e8c`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `818b6e8c`

[Automated Review — pr-reviewer bot] One-line fix is correct and the stated goal is achieved, but two repo-convention requirements are unmet: missing regression test and no bounds validation on the user-supplied config value.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The full `selectWithBudget` signature (line 471), the `HooksConfig` type (lines 205–230), and the call-site context (~1438–1460) were all retrieved. The fix is provably correct for the happy path. The two issues raised are directly grounded in the retrieved code: (1) `selectWithBudget` has no guard against zero/negative budgets and none is added at the call site, (2) no test file is present in the diff.

packages/daemon/src/hooks.ts
```diff
@@ -1438,7 +1438,7 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 	});
 
 	// Apply budget to select what we actually inject (on re-ranked order)
-	const memories = selectWithBudget(sortedCandidates, 2000);
+	const memories = selectWithBudget(sortedCandidates, config.maxInjectChars ?? 2000);
```
PR-Reviewer-Ant: No bounds validation on `config.maxInjectChars` before passing it to `selectWithBudget`. `selectWithBudget` is a greedy packer with no floor guard: any value ≤ 0 silently returns an empty array (zero memories injected at session start), and an arbitrarily large value is accepted without question. AGENTS.md explicitly lists 'Clamp counters, latency values, and limits to sane non-negative ranges. Reject out-of-range values with clear errors.' as a required prevention pattern. A minimal fix is to clamp or reject the value before use, e.g.:

```ts
const budget = config.maxInjectChars ?? 2000;
if (budget <= 0) {
  // log warning and fall back, or throw a structured error
}
const memories = selectWithBudget(sortedCandidates, Math.max(1, budget));
```

PR-Reviewer-Ant: No regression test is included. AGENTS.md requires 'Every bug fix needs a test that would fail before the fix,' and the PR checklist item 'Each bug fix has a regression test' is unmet. A minimal test would: (1) call `handleSessionStart` with `maxInjectChars` set in config and assert injection length is bounded by that value, and (2) omit the field and assert the 2000-char default applies. Without this, the bug can silently re-regress.

PR-Reviewer-Ant: `userPromptSubmit.maxInjectChars` is also declared in `HooksConfig` (line ~218). If its call site has the same hardcoded value, this PR leaves a parity gap. Worth confirming and addressing in this PR or a follow-up to avoid duplicate divergence.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `bdd12ff9`

[Automated Review — pr-reviewer bot] This PR expands scope significantly beyond its stated description and introduces a breaking config rename, a false claim about default behavior, and a renamed API response field. Several previously-open items remain unaddressed.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All blocking issues are directly provable from the diff lines. The config key rename (maxInjectChars → maxInjectTokens) is explicit at diff line +214. The default value magnitude change (2000 chars → 20000 tokens, ~40-50x) is explicit at diff lines +1456 and +1735. The PR description example contradicts the implementation. The response key rename (injectChars → injectTokens) is at diff line +1759. No test files appear anywhere in the diff.


- packages/daemon/src/hooks.ts:1748 - **`injectTokens` reports a count that includes the appended `\n[context truncated]` suffix.**

After `truncateToTokens(inject, mainBudget)`, the string `\n[context truncated]` is concatenated before the token count is taken at line 1759. The reported `injectTokens` will slightly exceed `mainBudget`. Minor, but it makes the metric misleading and could cause confusion when debugging injection limits.
- packages/daemon/src/hooks.ts:218 - **`userPromptSubmit.maxInjectChars` is not renamed to `maxInjectTokens`.**

The diff renames `sessionStart.maxInjectChars` to `maxInjectTokens` but the `userPromptSubmit` block (around line 218 per prior review context) still declares `maxInjectChars`. This creates a naming inconsistency in the config schema: two analogous fields in the same `HooksConfig` interface use different units and different names. If the `userPromptSubmit` call site is still hardcoded (as noted in the prior open review item), this PR leaves that parity gap open and now makes it harder to address consistently.


- `packages/daemon/src/hooks.ts:214` - **Breaking config rename: `maxInjectChars` → `maxInjectTokens`** Existing `agent.yaml` files with `maxInjectChars: N` will be silently ignored after this change — the field is removed from the type with no migration path or deprecation alias. Critically, the PR description's own example shows `maxInjectChars: 8000`, which directly contradicts the implementation using `maxInjectTokens`. Any user who follows the PR description to configure this will get the default behaviour instead of their configured value.
- `packages/daemon/src/hooks.ts:1456` - **PR description claims 'Default behaviour unchanged' — this is false.** The memory-selection default changes from `2000` (chars) to `20000` (tokens). At ~4 chars/token, 20000 tokens ≈ 80,000 characters — roughly a 40x increase. Users who relied on the old default will silently receive a vastly larger injection payload. If 20000 tokens is the correct new default, the PR description must state this explicitly as a behaviour change, not claim parity.
- `packages/daemon/src/hooks.ts:1735` - **Second default mismatch: 24000 chars → 20000 tokens.** The final-truncation guard previously defaulted to 24000 chars; it now defaults to 20000 tokens (~80,000 chars). Additionally, the two defaults are now unified at 20000 tokens, but the memory-selection pass (line 1456) and the final-truncation pass (line 1735) read the same config key with the same default. If a user sets `maxInjectTokens: 4000`, both passes will use 4000 — but the final pass also subtracts `reservedTokens`, so the effective budget for `inject` is lower than what the user configured for memory selection. Document or decouple these two budgets.
- `packages/daemon/src/hooks.ts:1759` - **Breaking API response key rename: `injectChars` → `injectTokens`.** The return/log object changes a field name. Any caller or dashboard component reading `.injectChars` from the `SessionStartResponse` (or its logged form) will silently receive `undefined`. The diff does not show the `SessionStartResponse` interface being updated, which would also leave a type mismatch. This is an API-surface break that warrants a note in the PR description and a check of all consumers.
- `packages/daemon/src/hooks.ts:486` - **`selectWithTokenBudget` still lacks a floor guard for zero/negative `tokenBudget`.** Same issue as the previously-open item on `selectWithBudget`: if `config.maxInjectTokens` is set to `0` or a negative value, the loop condition `used + cost > tokenBudget` is immediately true for any non-empty row, and the function returns an empty array with no error or log. Add `if (tokenBudget <= 0) return [];` with a warning log, or validate at the call site with a lower bound (e.g. `Math.max(1000, config.maxInjectTokens ?? 20000)`).
- `packages/daemon/src/hooks.ts:1456` - **No regression test in the diff.** AGENTS.md requires 'Every bug fix needs a test that would fail before the fix.' No test file appears anywhere in this diff. At minimum, a test for `selectWithTokenBudget` covering: normal selection, single row exceeding budget, zero budget, and the `maxInjectTokens` config being passed through to the call site would satisfy the repo convention.

packages/daemon/src/hooks.ts
```diff
@@ -210,7 +211,7 @@ export interface HooksConfig {
 		includeRecentContext?: boolean;
 		recencyBias?: number;
 		query?: string;
-		maxInjectChars?: number;
+		maxInjectTokens?: number;
```
PR-Reviewer-Ant, line 214: **Silent breaking change — config field renamed with no deprecation path.** `HooksConfig.sessionStart.maxInjectChars` is replaced by `maxInjectTokens` with no alias, no migration, and no warning when the old field is present. Existing `agent.yaml` files that set `hooks.sessionStart.maxInjectChars` will silently fall back to the 20000-token default. The old default was 2000 chars for memory selection (~1600 tokens); the new default is 20000 tokens (~80k chars). Users with the old field configured will see ~12× more context injected than they set, with no indication anything changed. At minimum, add a startup warning when `maxInjectChars` is detected in `sessionStart` config, pointing users to rename it.

packages/daemon/src/hooks.ts
```diff
@@ -1716,13 +1740,14 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
...
-	const mainBudget = Math.max(0, maxInject - reservedChars);
+	const reservedTokens =
+		countTokens(recoverySection) + countTokens(constraintsSection);
+	const mainBudget = Math.max(0, maxTokens - reservedTokens);
 	let inject = injectParts.join("\n");
-	if (inject.length > mainBudget) {
-		inject = `${inject.slice(0, mainBudget)}\n[context truncated]`;
+	if (countTokens(inject) > mainBudget) {
+		inject = `${truncateToTokens(inject, mainBudget)}\n[context truncated]`;
 	}
```
PR-Reviewer-Ant, line 1788: **Potential crash when `reservedTokens >= maxTokens`.** `mainBudget = Math.max(0, maxTokens - reservedTokens)` can be 0 if the constraints+recovery sections alone exhaust the budget. When `mainBudget = 0` and `inject` is non-empty, `truncateToTokens(inject, 0)` is called. The implementation of `truncateToTokens` is not shown in this diff — if it throws or returns `undefined` on a zero budget, the session-start handler will crash. Add a guard (`if (mainBudget > 0 && countTokens(inject) > mainBudget)`) or verify `truncateToTokens` handles 0 gracefully and add a test for this path.

packages/daemon/src/hooks.test.ts
```diff
@@ -332,3 +333,33 @@ describe("normalizeSessionTranscript", () => {
...
+	});
+
+	it("returns empty array when budget is too small for any row", () => {
+		const result = selectWithTokenBudget(rows, 1);
+		expect(result).toHaveLength(0);
+	});
+
+	it("returns empty array for zero budget", () => {
+		const result = selectWithTokenBudget(rows, 0);
+		expect(result).toHaveLength(0);
```
PR-Reviewer-Ant, line 364: Tests cover `selectWithTokenBudget` in isolation but there is no test for the `mainBudget = 0` edge case (constraints + recovery exceeding the budget) in the integration path. Per AGENTS.md, edge-case tests for timer/budget lifecycle are required. A test that passes `maxInjectTokens: 1` with non-empty constraints/recovery sections would catch the crash risk flagged above.

packages/daemon/src/hooks.ts
```diff
@@ -1716,13 +1740,14 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
...
 	if (constraintsSection) {
 		inject += constraintsSection;
```
PR-Reviewer-Ant, line 1791: `\n[context truncated]` is appended after `truncateToTokens(inject, mainBudget)`, adding ~4 tokens that are not accounted for in `mainBudget`. Combined with `constraintsSection` being appended after (already reserved), the final `inject` can slightly exceed `maxTokens`. This is cosmetically minor (~4 tokens) but technically breaks the invariant that the total output fits within `maxTokens`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5bd5c3c1`

[Automated Review — pr-reviewer bot] All three items from the prior review (bounds validation, regression tests, userPromptSubmit parity) are addressed. Two new issues found: a silent breaking change from the config field rename, and an untested edge case where `truncateToTokens` is called with budget 0.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The breaking change is directly visible in the HooksConfig type diff (line 214: maxInjectChars → maxInjectTokens) with no migration or deprecation shim. The mainBudget=0 edge case is derivable from lines 1743-1753: Math.max(0, ...) can yield 0, and the truncateToTokens import is from ./pipeline/tokenizer whose behavior at budget=0 is not shown in this diff.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `a8756ebf`

[Automated Review — pr-reviewer bot] Incremental diff resolves all previously blocking items. Two lower-severity open items remain: a missing integration-path regression test for the `mainBudget === 0` guard, and the unverified `userPromptSubmit` parity claim.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - All seven previously-open items are traceable directly in the incremental diff: the @deprecated shim (line +214), the mainBudget=0 guard (lines +1751-1758), and the marker-token pre-subtraction (lines +1762-1765) are all explicitly present. The medium confidence is due to `userPromptSubmit.maxInjectChars` — its call site is in the non-diff portion of hooks.ts and was not included in the provided context, so the PR author's claim of 'no parity gap' cannot be verified from this review.

packages/daemon/src/hooks.ts
```diff
@@ -1716,13 +1745,22 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
 	}
 
 	const duration = Date.now() - start;
-	const maxInject = config.maxInjectChars ?? 24000;
+	const maxTokens = config.maxInjectTokens ?? 20000;
 	// Pre-reserve space for constraints + recovery so they are never truncated
-	const reservedChars = recoverySection.length + constraintsSection.length;
-	const mainBudget = Math.max(0, maxInject - reservedChars);
+	const reservedTokens =
+		countTokens(recoverySection) + countTokens(constraintsSection);
```
PR-Reviewer-Ant, line 1780: The `mainBudget === 0` guard is correct and safe, but AGENTS.md requires 'Every bug fix needs a test that would fail before the fix.' This new branch (reserved sections exhausting the token budget, resulting in `inject = ""`) is not exercised by any test in the diff. The `selectWithTokenBudget` unit tests cover the helper in isolation; they don't reach this integration path. A test that sets `maxInjectTokens` low enough that `reservedTokens >= maxTokens` and asserts `inject` is empty (or the log warning is emitted) would satisfy the convention.

packages/daemon/src/hooks.ts
```diff
@@ -210,6 +211,8 @@ export interface HooksConfig {
 		includeRecentContext?: boolean;
 		recencyBias?: number;
 		query?: string;
+		maxInjectTokens?: number;
+		/** @deprecated renamed to maxInjectTokens — will be ignored */
 		maxInjectChars?: number;
 	};
 	userPromptSubmit?: {
```
PR-Reviewer-Ant, line 222: `userPromptSubmit.maxInjectChars` remains in `HooksConfig` as the active (non-deprecated) field while `sessionStart.maxInjectChars` has been deprecated and replaced by `maxInjectTokens`. The PR description claims 'userPromptSubmit.maxInjectChars already reads from config correctly — no parity gap', but the `handleUserPromptSubmit` call site is not included in the diff or provided context, so this claim cannot be verified. If the `userPromptSubmit` path still has a hardcoded char budget analogous to the old `sessionStart` behaviour, the parity gap flagged in prior reviews remains. This item was marked `[open]` by the last two review rounds and the incremental diff does not touch the `userPromptSubmit` block — please confirm the call site reads from config and is not hardcoded.

packages/daemon/src/hooks.ts
```diff
@@ -479,6 +482,39 @@ export function selectWithBudget<T extends { content: string }>(rows: ReadonlyAr
...
+/**
+ * Truncate `inject` to fit within `mainBudget` tokens.
+ * Returns an empty string when budget is zero (reserved sections exhausted it).
+ * Appends a truncation marker whose tokens are pre-subtracted from the budget.
+ */
+export function applyTokenBudget(inject: string, mainBudget: number): string {
+	if (mainBudget <= 0) return "";
+	if (countTokens(inject) <= mainBudget) return inject;
+	return (
+		truncateToTokens(inject, Math.max(1, mainBudget - TRUNCATED_MARKER_TOKENS)) +
```
PR-Reviewer-Ant: Budget overflow when `1 <= mainBudget < TRUNCATED_MARKER_TOKENS`. When `mainBudget` is between 1 and the marker token count (approximately 5–6 for `\n[context truncated]` in cl100k_base), `Math.max(1, mainBudget - TRUNCATED_MARKER_TOKENS)` evaluates to 1, so the return value is `truncateToTokens(inject, 1) + TRUNCATED_MARKER`. The total token count (~1 + ~5) exceeds the stated budget. The function's JSDoc contract ('fit within mainBudget tokens') is violated in this range. In practice this requires `reservedTokens` to be within ~5 tokens of `maxTokens`, but the test coverage at budget=50 does not catch it. Consider: `if (mainBudget <= TRUNCATED_MARKER_TOKENS) return truncateToTokens(inject, mainBudget)` (omit marker when budget is too tight to accommodate it), or add a test asserting `countTokens(applyTokenBudget(TEXT, 3)) <= 3`.

packages/daemon/src/hooks.test.ts
```diff
@@ -332,3 +334,62 @@ describe("normalizeSessionTranscript", () => {
...
+	const TEXT = "word ".repeat(500); // ~500 tokens
+
+	it("returns inject unchanged when it fits within budget", () => {
+		expect(applyTokenBudget("hello world", 1000)).toBe("hello world");
+	});
+
+	it("truncates and appends marker when inject exceeds budget", async () => {
+		const result = applyTokenBudget(TEXT, 50);
+		expect(result).toContain("[context truncated]");
+		// total tokens must not exceed budget (marker tokens pre-subtracted)
```
PR-Reviewer-Ant, line 383: The truncation test only validates budget=50. The correctness guarantee that `countTokens(result) <= mainBudget` should also be verified for `mainBudget` values in [1, TRUNCATED_MARKER_TOKENS) where the pre-subtraction trick breaks down. Adding a `expect(countTokens(applyTokenBudget(TEXT, 3))).toBeLessThanOrEqual(3)` case would exercise the gap and would currently fail, making this a useful regression anchor.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `4b793822`

[Automated Review — pr-reviewer bot] Incremental diff (a8756ebf → 4b79382) addresses the two previously open items from the last review (mainBudget=0 test, userPromptSubmit parity comment). One new correctness gap introduced by applyTokenBudget: the marker pre-subtraction does not fully protect against budget overflow when mainBudget is between 1 and TRUNCATED_MARKER_TOKENS.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The applyTokenBudget overflow edge case is directly derivable from the function body at hooks.ts lines 507-514: when mainBudget is in [1, TRUNCATED_MARKER_TOKENS), Math.max(1, mainBudget - TRUNCATED_MARKER_TOKENS) collapses to 1, and the appended marker (~5-6 tokens for '\n[context truncated]' in cl100k_base) causes the output to exceed mainBudget. The test at hooks.test.ts line 383 uses budget=50 and does not cover this range. Confidence is medium because truncateToTokens behavior at budget=1 is in packages/daemon/src/pipeline/tokenizer.ts which is not included in the diff context — an off-by-one there could change the severity.

packages/daemon/src/hooks.ts
PR-Reviewer-Ant: **Budget overflow when `mainBudget ∈ [1, TRUNCATED_MARKER_TOKENS)`.**

When `mainBudget` is smaller than the marker's own token cost (~5 for `\n[context truncated]` in cl100k_base), `Math.max(1, mainBudget - TRUNCATED_MARKER_TOKENS)` collapses to 1, `truncateToTokens` returns ~1 token, and appending the full marker yields ~6 tokens — violating the `mainBudget` contract.

This is only triggered when a user sets a small `maxInjectTokens` (e.g. 10) while reserved sections (recovery + constraints) consume most of it — plausible but uncommon. The simplest fix is to guard: `if (mainBudget <= TRUNCATED_MARKER_TOKENS) return "";` instead of the current `Math.max(1, ...)` path, which already silently overcounts.

packages/daemon/src/hooks.test.ts
PR-Reviewer-Ant, line 383: The `applyTokenBudget` truncation test only validates budget=50. The overflow described above (budget in [1, TRUNCATED_MARKER_TOKENS)) is not covered. Adding a case like `applyTokenBudget(TEXT, 3)` and asserting `countTokens(result) <= 3` would both document the contract and catch the bug.

packages/daemon/src/hooks.ts
```diff
@@ -1438,7 +1474,23 @@ export async function handleSessionStart(req: SessionStartRequest): Promise<Sess
...
 	// Apply budget to select what we actually inject (on re-ranked order)
-	const memories = selectWithBudget(sortedCandidates, 2000);
+	if (config.maxInjectChars !== undefined && config.maxInjectTokens === undefined) {
+		logger.warn(
+			"hooks",
+			"hooks.sessionStart.maxInjectChars is deprecated — migrating to maxInjectTokens automatically. Rename it in agent.yaml to silence this warning.",
+			{ maxInjectChars: config.maxInjectChars, derivedTokens: Math.round(config.maxInjectChars / 4) },
+		);
+	}
+	const rawTokenBudget =
```
PR-Reviewer-Ant, line 1488: The `/4` approximation (`Math.round(config.maxInjectChars / 4)`) is a reasonable best-effort migration heuristic, but ASCII-only content averages ~4 chars/token while code, Unicode, or dense prose can be 1–2 chars/token. The warn log now includes `derivedTokens` so users can inspect the conversion, which is good. Worth documenting the approximation explicitly in the `@deprecated` JSDoc so users with unusual content know to migrate promptly.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0e2d4bc1`

[Automated Review — pr-reviewer bot] Incremental diff (4b79382 → 0e2d4bc) closes two remaining open items from prior reviews: it makes the `maxInjectChars` migration consistent across both budget calculations (selection and truncation) and confirms the `userPromptSubmit` parity claim with a code comment. One previously-flagged correctness gap in `applyTokenBudget` remains unaddressed. One AGENTS.md convention miss (missing integration-path regression test) also remains.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The `applyTokenBudget` overflow is directly provable from the function body in the full diff (hooks.ts lines 507–514): when mainBudget ∈ [1, TRUNCATED_MARKER_TOKENS), `Math.max(1, mainBudget - TRUNCATED_MARKER_TOKENS)` collapses to 1, and appending the ~5-token marker violates the budget guarantee. The trigger requires a small custom `maxInjectTokens` (e.g. ≤ 10) where reserved sections nearly exhaust the budget — plausible for users experimenting with tight injection caps. The test suite at hooks.test.ts only covers budget=50 (line 383), which does not catch this range. All other previously-open items are traceable as closed in this incremental diff.

PR #486, opened 2026-04-09
fix(publish): block leaked workspace deps in release

NicholaiVogel:
## Summary
- remove the unpublished `@signet/connector-pi` runtime dependency from the published `signetai` manifest while keeping the connector in the CLI codepath
- add a publish-manifest guard that fails if a publishable package ships `workspace:` runtime deps or depends on an internal workspace package that is not published to npm
- run that guard in the release workflow before undrafting the GitHub release, and document the rule in `AGENTS.md`

## Root cause
`signetai@0.98.8` shipped `@signet/connector-pi` in `dependencies` as `workspace:*`.
That package is an internal workspace connector and is not published on npm, so `bun add -g signetai` failed during dependency resolution.

The important distinction is that `@signet/connector-pi` is required in the repo and in the bundled CLI, but it is not part of the published npm dependency surface for `signetai`. Previous releases already followed that contract for the other connector packages.

## Validation
- `npm view signetai@0.98.8 dependencies --json` shows the broken published dep: `@signet/connector-pi=workspace:*`
- `bun test ./scripts/check-publish-manifests.test.ts`
- `bun scripts/check-publish-manifests.ts packages/signetai/package.json packages/adapters/openclaw/package.json`
- `npm pack --json --ignore-scripts ./packages/signetai` and inspected `package/package.json` from the tarball to confirm the leaked connector dep is gone

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -310,15 +310,20 @@ jobs:
       - name: Build
         run: bun run build
 
+      - name: Resolve workspace protocols
+        run: bun scripts/version-sync.ts --to "$NEW_VERSION"
+
+      - name: Validate publish manifests
+        run: bun scripts/check-publish-manifests.ts
+
       - name: Publish release
-        # Undraft first so predictor assets are publicly accessible before
-        # any npm install can trigger the postinstall downloader.
+        # Undraft only after publish manifests pass, so a broken npm package
+        # cannot slip through behind a green release job.
+        # Assets must still be public before npm publish can expose any
+        # postinstall downloader paths to users.
         run: gh release edit "v${NEW_VERSION}" --draft=false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
-      - name: Resolve workspace protocols
-        run: bun scripts/version-sync.ts --to "$NEW_VERSION"
 
       - name: Publish to npm
         run: |
--- a/scripts/check-publish-manifests.test.ts
+++ b/scripts/check-publish-manifests.test.ts
@@ -1,0 +1,191 @@
+import { describe, expect, test } from "bun:test";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import {
+	collectManifestIssues,
+	collectWorkspacePackages,
+	isPublishableWorkspacePackage,
+	listPublishableManifestTargets,
+} from "./check-publish-manifests";
+
+function writeJson(file: string, value: unknown): void {
+	writeFileSync(file, `${JSON.stringify(value, null, 2)}\n`);
+}
+
+describe("check-publish-manifests", () => {
+	test("treats manifests with publishConfig.access public as publishable", () => {
+		expect(
+			isPublishableWorkspacePackage({
+				name: "signetai",
+				publishConfig: { access: "public" },
+			}),
+		).toBe(true);
+	});
+
+	test("discovers publishable manifest targets from workspace files", () => {
+		const root = mkdtempSync(join(tmpdir(), "signet-publish-manifests-"));
+		try {
+			const signetaiDir = join(root, "packages", "signetai");
+			const adapterDir = join(root, "packages", "adapters", "openclaw");
+			const connectorDir = join(root, "packages", "connector-pi");
+			mkdirSync(signetaiDir, { recursive: true });
+			mkdirSync(adapterDir, { recursive: true });
+			mkdirSync(connectorDir, { recursive: true });
+
+			const signetaiFile = join(signetaiDir, "package.json");
+			const adapterFile = join(adapterDir, "package.json");
+			const connectorFile = join(connectorDir, "package.json");
+
+			writeJson(signetaiFile, {
+				name: "signetai",
+				publishConfig: { access: "public" },
+			});
+			writeJson(adapterFile, {
+				name: "@signetai/signet-memory-openclaw",
+				publishConfig: { access: "public" },
+			});
+			writeJson(connectorFile, {
+				name: "@signet/connector-pi",
+			});
+
+			expect(listPublishableManifestTargets([signetaiFile, adapterFile, connectorFile])).toEqual([
+				signetaiFile,
+				adapterFile,
+			]);
+		} finally {
+			rmSync(root, { recursive: true, force: true });
+		}
+	});
+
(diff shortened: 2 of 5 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(publish): block leaked workspace deps in release` (commit `0278e0c5`) now and I'll follow up shortly with feedback.

.github/workflows/release.yml
```diff
@@ -310,16 +310,21 @@ jobs:
       - name: Build
         run: bun run build
 
+      - name: Resolve workspace protocols
+        run: bun scripts/version-sync.ts --to "$NEW_VERSION"
+
```
PR-Reviewer-Ant, line 315: The CI hardcodes exactly two target packages. If a third publishable package is added to the workspace (e.g. `@signet/sdk` or a future connector that gains `publishConfig.access: public`), this step silently skips it and the same class of bug can recur. The `collectWorkspacePackages` function already has the machinery to discover all publishable packages from the git index — consider a follow-up that auto-discovers targets via `listWorkspacePackageFiles()` filtered by `isPublishableWorkspacePackage`, so the guard self-maintains. For now, at minimum, a comment here naming the known publishable packages would make audits easier.

scripts/check-publish-manifests.ts
```diff
@@ -0,0 +1,167 @@
...
+
+export function isPublishableWorkspacePackage(pkg: PackageJson): boolean {
+	if (pkg.private === true) return false;
+
+	if (pkg.name === "signetai") return true;
+
+	if (!isRecord(pkg.publishConfig)) return false;
+
+	return pkg.publishConfig.access === "public";
+}
```
PR-Reviewer-Ant, line 56: The hardcoded `pkg.name === "signetai"` branch in `isPublishableWorkspacePackage` is now redundant: this PR adds `publishConfig: { access: "public" }` to `packages/signetai/package.json`, so the `signetai` package will fall through to the `publishConfig.access === "public"` branch on any future call that reads the real manifest. The special-case name check can mislead a reader into thinking `signetai` lacks `publishConfig`. Not a correctness bug today, but it diverges from the documented contract (publish manifests signal publishability via `publishConfig`) and could mask a future regression if someone removes `publishConfig` from `signetai` without noticing the name-based fallback.

scripts/check-publish-manifests.test.ts
```diff
@@ -0,0 +1,120 @@
+import { describe, expect, test } from "bun:test";
```
PR-Reviewer-Ant, line 1: There is no test for the positive case: a publishable workspace package used as a runtime dependency should not produce an issue. For example, if `signetai` depended on `@signetai/signet-memory-openclaw` (which has `publishConfig.access: public`), `collectManifestIssues` should return zero issues. Without this test the boundary between flagged and allowed workspace deps is only tested on the rejection side, leaving a regression surface for the `workspaceDep.publishable` branch.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0278e0c5`

[Automated Review — pr-reviewer bot] This PR correctly removes the leaked `@signet/connector-pi` workspace dep from the `signetai` publish manifest and adds a CI guard to prevent recurrence. The core fix and guard logic are sound. Two warnings are worth discussing before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The full diff for all changed files is present. The guard logic in check-publish-manifests.ts is self-contained and directly verifiable from the diff. The workflow reordering is explicit and the sequencing can be reasoned about from the YAML alone. The only missing context is the content of version-sync.ts, but its contract (rewrite workspace: protocols to pinned versions) is described in comments and tested indirectly.

scripts/check-publish-manifests.ts
```diff
@@ -0,0 +1,170 @@
...
+function readPackageJson(file: string): PackageJson {
+	return JSON.parse(readFileSync(file, "utf8")) as PackageJson;
+}
+
+function getPackageName(file: string, pkg: PackageJson): string {
+	if (typeof pkg.name !== "string" || pkg.name.length === 0) {
+		throw new Error(`Missing package name in ${file}`);
+	}
+
+	return pkg.name;
```
PR-Reviewer-Ant, line 47: The guard's auto-discovery (`listPublishableManifestTargets`) only covers packages where `publishConfig.access === "public"` is present in `package.json`. `@signetai/signet-memory-openclaw` is published in the `Publish to npm` step (`cd ../adapters/openclaw && npm publish --tag next --access public`), but its `package.json` is not modified in this PR to add `publishConfig`. If it lacks the field, the guard will neither check it as a publish target nor recognise it as publishable when another package depends on it — causing a false-positive block. This is only a gap until `publishConfig` is added to that package, but worth doing in a follow-up to ensure the convention is complete across all published packages.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `887a8373`

[Automated Review — pr-reviewer bot] This is an incremental review of PR #486 (sha 887a837 vs prior sha 0278e0c). All three previously-open items are addressed in this diff. One new warning is noted.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The full incremental diff is present. All three previously-open concerns can be verified directly: (1) the CI no longer hardcodes target packages — `bun scripts/check-publish-manifests.ts` is called with no args, relying on `listPublishableManifestTargets()` auto-discovery; (2) `isPublishableWorkspacePackage` has no hardcoded `pkg.name === 'signetai'` branch; (3) the positive test case ('allows runtime dependencies on publishable workspace packages') is present at line ~127. The one new concern is directly provable from the diff and the workflow's npm publish step.

scripts/check-publish-manifests.ts
NicholaiVogel: this one’s already covered.

`packages/adapters/openclaw/package.json` already has `publishConfig.access: "public"`, so it is included by `listPublishableManifestTargets()` and recognized as publishable by `isPublishableWorkspacePackage()`.

i also verified the guard returns both current publishable manifests:
- `packages/adapters/openclaw/package.json`
- `packages/signetai/package.json`

and the tests cover both discovery and the positive-case dependency boundary.

PR #487, opened 2026-04-10
fix(release): rebuild changelog with readable highlights

NicholaiVogel:
## Summary
- fix the release workflow so changelog entries are generated for the actual new version instead of the pre-bump package version
- rebuild `CHANGELOG.md` from semver tags and add a readable `Recent Highlights` layer above the exact release ledger
- add changelog regression coverage and update contributing docs for `--bump-only`, `--version`, and `--rebuild`

## Testing
- `bun test packages/core/src/__tests__/changelog-script.test.ts`
- `bunx @biomejs/biome check scripts/changelog.ts packages/core/src/__tests__/changelog-script.test.ts CONTRIBUTING.md docs/CONTRIBUTING.md .github/workflows/release.yml`
- `bun x tsc --noEmit --module esnext --target es2020 --skipLibCheck scripts/changelog.ts packages/core/src/__tests__/changelog-script.test.ts`
- `bun scripts/changelog.ts --rebuild`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -138,6 +138,9 @@ jobs:
         run: |
           NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION")
           if [ -z "$NOTES" ] || [ "$NOTES" = "No changelog section found for v$NEW_VERSION" ]; then
+            # This runs after the version bump commit and tag are pushed. If it
+            # fails, recover by creating the draft release manually with
+            # `gh release create "v$NEW_VERSION" --draft --notes-file <notes>`.
             echo "::error::Missing CHANGELOG.md section for v$NEW_VERSION"
             exit 1
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,8 +111,8 @@ jobs:
           echo "bump_level=$BUMP" >> $GITHUB_OUTPUT
           rm -f .bump-level
 
-          PACKAGE_FILES=$(git ls-files package.json 'packages/**/package.json' | grep -v '^packages/cli/dashboard/package.json$')
-          for file in $PACKAGE_FILES; do
+          mapfile -t PACKAGE_FILES < <(git ls-files package.json 'packages/**/package.json' | grep -v '^packages/cli/dashboard/package.json$')
+          for file in "${PACKAGE_FILES[@]}"; do
             jq --arg v "$NEW_VERSION" '.version = $v' "$file" > "$file.tmp" && mv "$file.tmp" "$file"
           done
 
@@ -122,7 +122,7 @@ jobs:
           # (Docker smoke and consumer builds) stay deterministic.
           bun install
 
-          git add $PACKAGE_FILES bun.lock CHANGELOG.md
+          git add -- "${PACKAGE_FILES[@]}" bun.lock CHANGELOG.md
           git commit -m "chore: release $NEW_VERSION"
           git tag "v$NEW_VERSION"
 
@@ -137,7 +137,7 @@ jobs:
         if: steps.changes.outputs.skip != 'true'
         run: |
           NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION")
-          if [ -z "$NOTES" ] || printf '%s' "$NOTES" | grep -q "^No changelog section found for v$NEW_VERSION$"; then
+          if [ -z "$NOTES" ] || [ "$NOTES" = "No changelog section found for v$NEW_VERSION" ]; then
             echo "::error::Missing CHANGELOG.md section for v$NEW_VERSION"
             exit 1
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -97,7 +97,7 @@ jobs:
           BASE_VERSION=$(printf '%s\n%s\n' "$LOCAL_VERSION" "$REMOTE_VERSION" | sort -V | tail -n1)
           IFS='.' read -r MAJOR MINOR PATCH <<< "$BASE_VERSION"
 
-          bun scripts/changelog.ts
+          bun scripts/changelog.ts --bump-only
 
           BUMP=$(cat .bump-level 2>/dev/null || echo "patch")
           case "$BUMP" in
@@ -115,6 +115,8 @@ jobs:
           for file in $PACKAGE_FILES; do
             jq --arg v "$NEW_VERSION" '.version = $v' "$file" > "$file.tmp" && mv "$file.tmp" "$file"
           done
+
+          bun scripts/changelog.ts --version "$NEW_VERSION"
 
           # Keep lockfile aligned with workspace version bumps so frozen installs
           # (Docker smoke and consumer builds) stay deterministic.
@@ -134,7 +136,11 @@ jobs:
         id: create_release
         if: steps.changes.outputs.skip != 'true'
         run: |
-          NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION" 2>/dev/null || echo "See CHANGELOG.md for details")
+          NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION")
+          if [ -z "$NOTES" ] || printf '%s' "$NOTES" | grep -q "^No changelog section found for v$NEW_VERSION$"; then
+            echo "::error::Missing CHANGELOG.md section for v$NEW_VERSION"
+            exit 1
+          fi
           # Create as draft — predictor assets are uploaded by build-predictor
           # before the publish job promotes this to non-draft and publishes npm.
           gh release create "v$NEW_VERSION" --draft --title "v$NEW_VERSION" --notes "$NOTES"
--- a/packages/core/src/__tests__/changelog-script.test.ts
+++ b/packages/core/src/__tests__/changelog-script.test.ts
@@ -67,6 +67,10 @@ function makeTaggedReleaseRepo(): string {
 	return dir;
 }
 
+function countOccurrences(value: string, needle: string): number {
+	return value.split(needle).length - 1;
+}
+
 afterEach(() => {
 	for (const dir of tempDirs.splice(0)) {
 		rmSync(dir, { force: true, recursive: true });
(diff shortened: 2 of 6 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(release): rebuild changelog with readable highlights` (commit `91b9f7a2`) now and I'll follow up shortly with feedback.

.github/workflows/release.yml
```diff
@@ -134,7 +136,11 @@ jobs:
         id: create_release
         if: steps.changes.outputs.skip != 'true'
         run: |
-          NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION" 2>/dev/null || echo "See CHANGELOG.md for details")
+          NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION")
```
PR-Reviewer-Ant, line 139: The grep pattern treats `$NEW_VERSION` as a regex, so dots in version strings like `0.98.9` match any character rather than literal dots. For a version such as `0.98.9` the pattern `^No changelog section found for v0.98.9$` would also match a hypothetical string like `No changelog section found for v0X98Y9`. In practice the risk is negligible because release notes are script-generated and that exact text is unlikely to appear, but the intent is a literal string match. Fix with `fgrep -q` or escape the version: `grep -qF "No changelog section found for v$NEW_VERSION"` (and drop the anchors, since `-F` does literal matching).

packages/core/src/__tests__/changelog-script.test.ts
```diff
@@ -0,0 +1,126 @@
...
+	return dir;
+}
+
+afterEach(() => {
+	for (const dir of tempDirs.splice(0)) {
+		rmSync(dir, { force: true, recursive: true });
+	}
+});
+
+describe("scripts/changelog.ts", () => {
```
PR-Reviewer-Ant, line 80: The test verifies that `--version 0.1.1` produces the correct output, but it does not assert that `--bump-only` alone leaves `CHANGELOG.md` untouched (or absent). The original bug was triggered because the changelog was written before the version bump; a regression test that asserts `CHANGELOG.md` does not exist (or does not contain the new version) immediately after `--bump-only` would more directly guard the specific failure mode the PR fixes. Without it, a future refactor that accidentally re-introduces early-write behaviour would not be caught by this test file.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `91b9f7a2`

[Automated Review — pr-reviewer bot] PR #487 correctly fixes the root bug (changelog written before version bump, so entries carried the old version number) by splitting the changelog call into two phases: `--bump-only` to determine the bump level, then `--version "$NEW_VERSION"` after all `package.json` files are updated. The implementation achieves what the description claims. Three minor concerns are flagged below.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The two-phase changelog split is explicit in the diff at lines 100 and 117 of release.yml. The test in changelog-script.test.ts directly reproduces the pre-fix bug scenario (--bump-only before version bump, then --version with the bumped number) and the assertions confirm the right version appears in the output. The grep-regex and unquoted-variable findings are directly provable from the changed lines.


- .github/workflows/release.yml:129 - `git add $PACKAGE_FILES bun.lock CHANGELOG.md` expands `$PACKAGE_FILES` without quoting, so any path containing whitespace would be word-split into multiple tokens and potentially interpreted as separate files (or cause `git add` to fail on an unrecognised token). All current package paths use hyphens and no spaces, so this is dormant today, but it is a latent shell-correctness issue. Prefer `git add -- $PACKAGE_FILES bun.lock CHANGELOG.md` or collect paths into an array and use `"${PACKAGE_FILES[@]}"`.

packages/core/src/__tests__/changelog-script.test.ts
```diff
@@ -0,0 +1,127 @@
...
+	const dir = makeRepo();
+	writeFileSync(join(dir, "README.md"), "hello\n");
+	run(["git", "add", "README.md"], dir);
+	run(["git", "commit", "-m", "fix(cli): repair release notes ordering"], dir);
+	writeFileSync(join(dir, "NOTES.md"), "release notes\n");
+	run(["git", "add", "NOTES.md"], dir);
+	run(["git", "commit", "-m", "docs(repo): document release flow"], dir);
+	run(["git", "tag", "v0.1.1"], dir);
+	return dir;
+}
```
PR-Reviewer-Ant, line 68: Both test cases call `--version` on a repo that has no `CHANGELOG.md` yet (the file does not exist before the call). The primary production path, however, is calling `--version` on a repo where `CHANGELOG.md` already exists with the two-section structure produced by `--rebuild` or a prior `--version` run. If the script has a bug in its 'prepend to existing structured file' logic — e.g. it double-inserts the Recent Highlights heading, corrupts the Release Ledger boundary, or misparses the prior entry range — no test would catch it, and the defect would silently recur on every subsequent release. Please add a test that:
1. Creates a CHANGELOG.md via `--version 0.1.1`.
2. Adds more commits.
3. Calls `--version 0.1.2` on the now-existing file.
4. Asserts the file has exactly one `## Recent Highlights` heading, the new entry is at the top of the ledger, and the old entry is still intact.

.github/workflows/release.yml
```diff
@@ -134,7 +136,11 @@ jobs:
         id: create_release
```
PR-Reviewer-Ant, line 136: The hard `exit 1` on a missing changelog section fires in the `Create GitHub Release (draft)` step, which executes *after* `git push origin HEAD:main` and `git push origin "v$NEW_VERSION"` have already completed. If changelog extraction fails here (e.g. `changelog.ts --version` produced an entry that `extract-changelog-section.ts` cannot locate), the version bump commit and tag are permanently in the remote but no GitHub release exists. This is not worse than the pre-PR fallback ('See CHANGELOG.md for details'), but the new hard-fail path should be documented or at least commented so on-call engineers know the recovery procedure: manually run `gh release create` with the extracted notes. Consider adding a short inline comment explaining the partial-release state and how to recover.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f81f1c62`

[Automated Review — pr-reviewer bot] PR #487 fixes a real bug (changelog entries used the pre-bump package version instead of the new release version) and adds a readable highlights layer to CHANGELOG.md. The CI workflow changes are structurally sound, but there is one meaningful coverage gap and one edge-case failure-mode worth discussing before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The release workflow ordering, the new hard-fail guard, and the test file are all fully visible in the diff. The coverage gap in the test suite is directly provable from the test file: no test calls `--version` on a CHANGELOG.md that already has the two-section (Recent Highlights + Release Ledger) structure, which is the exact steady-state path that will run on every future release.

.github/workflows/release.yml
```diff
@@ -134,7 +136,14 @@ jobs:
...
```
PR-Reviewer-Ant, line 136: **Warning: orphan tag on changelog extraction failure.** By the time `extract-changelog-section.ts` is called, `git push origin HEAD:main` and `git push origin "v$NEW_VERSION"` have already completed (previous step). If the NOTES check triggers `exit 1`, the version-bump commit and semver tag are live on `main` but no GitHub release draft exists. Recovery requires `gh release create "v$NEW_VERSION" --draft ...` manually, as the inline comment notes.

This is a deliberate improvement over the silent fallback (the old `|| echo "See CHANGELOG.md for details"` would have hidden the missing section). The tradeoff is intentional, but worth ensuring the on-call runbook or CI alert surfaces it clearly. If a way to atomically check before pushing exists in the script (e.g., dry-run extract before the push), that would eliminate the recovery window entirely.

.github/workflows/release.yml
```diff
@@ -134,7 +136,14 @@ jobs:
...
         if: steps.changes.outputs.skip != 'true'
         run: |
-          NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION" 2>/dev/null || echo "See CHANGELOG.md for details")
+          NOTES=$(bun scripts/extract-changelog-section.ts "$NEW_VERSION")
```
PR-Reviewer-Ant, line 139: **Nitpick: `gh release create --notes "$NOTES"` with multiline content.** Shell argument expansion is safe here (bash does not re-evaluate command substitution results inside variables), and this pattern was present before this PR. However, extremely large changelogs could approach `ARG_MAX` limits on some systems. For the volumes seen in this repo's CHANGELOG this is not a practical concern, but `--notes-file <(echo "$NOTES")` would be more robust for very large release notes.

packages/core/src/__tests__/changelog-script.test.ts
```diff
@@ -0,0 +1,156 @@
...
+import { tmpdir } from "node:os";
+import { join, resolve } from "node:path";
+
+const repoRoot = resolve(import.meta.dir, "../../../../");
+const changelogScript = join(repoRoot, "scripts", "changelog.ts");
+const extractSectionScript = join(repoRoot, "scripts", "extract-changelog-section.ts");
+
+const tempDirs: string[] = [];
+
+function run(cmd: readonly string[], cwd: string): string {
```
PR-Reviewer-Ant, line 13: The `run()` helper throws on non-zero exit but has no timeout. `bun scripts/changelog.ts --rebuild` on a repo with a large git history could hang a test suite. Consider passing a `timeout` option to `spawnSync` (e.g. 30 000 ms). The test repos are tiny, so this is not a current problem, but it's a latent fragility if the test fixture grows.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5ef8b769`

[Automated Bot Review] This PR correctly splits changelog generation into two phases to fix the stated bug, improves shell array handling in the release workflow, and adds regression coverage. No blocking issues found; two warnings noted.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The workflow diff, changelog script, and test file are all present. The core fix (--bump-only before version bump, --version after) is verifiable end-to-end in the diff. The failure mode around orphan tags is directly provable from the workflow step ordering at lines 136-148 of release.yml.

PR #498, opened 2026-04-12
fix(daemon): enable Signet pipeline through GitHub Copilot providers

Ostico:
## Summary

Signet's extraction/synthesis pipeline fails silently on all GitHub Copilot-backed models (via OpenCode). This PR fixes it — enabling Signet memory to work through GitHub Copilot for the first time.

- Add `structuredOutput` config option to disable the `format` field that GitHub Copilot rejects
- Auto-detect HTTP 200 responses with empty bodies (Copilot's silent schema rejection) and recover transparently
- Wire config through `agent.yaml` → daemon extraction/synthesis/summary providers

## Changes

- `packages/core/src/types.ts` — Added `structuredOutput?: boolean` to `PipelineExtractionConfig` and `PipelineSynthesisConfig`
- `packages/daemon/src/pipeline/provider.ts` — OpenCode provider: `enableStructuredOutput` interface field, init flag, 200-with-error detection + auto-recovery
- `packages/daemon/src/memory-config.ts` — Parse `structuredOutput` from extraction and synthesis YAML blocks (synthesis falls back to extraction value)
- `packages/daemon/src/daemon.ts` — Wire `enableStructuredOutput` to extraction and synthesis OpenCode provider creation
- `packages/daemon/src/pipeline/summary-worker.ts` — Wire `enableStructuredOutput` to summary OpenCode provider creation
- `packages/daemon/src/pipeline/provider.test.ts` — Two new tests: config-driven format omission + consecutive malformed 200 auto-recovery
- `docs/CONFIGURATION.md` — Document `structuredOutput` in extraction and synthesis config tables
- `docs/PIPELINE.md` — Add `structuredOutput` to extraction YAML example

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — No spec changes. This is a provider-level bugfix within `memory-pipeline-v2` (complete). `structuredOutput` is a config key on existing `PipelineExtractionConfig`/`PipelineSynthesisConfig` interfaces, not a new spec surface.
- [x] Agent scoping verified on all new/changed data queries — No database queries added or changed. This PR modifies only LLM provider config wiring and HTTP response handling.
- [x] Input/config validation and bounds checks added — `structuredOutput` is parsed with `typeof === "boolean"` checks in `memory-config.ts`. Synthesis falls back to extraction value, then `undefined`. The `enableStructuredOutput` field defaults to `true` when omitted (backwards compatible).
- [x] Error handling and fallback paths tested (no silent swallow) — Auto-detection path: after 2 consecutive malformed 200 responses, provider disables `structuredOutputSupported` and retries without the format field. Tested in `provider.test.ts`. Config-driven path: `structuredOutput: false` omits the format field entirely. Also tested.
- [x] Security checks applied to admin/mutation endpoints — No endpoints added or changed. This PR modifies only internal provider construction and LLM request formatting.
- [x] Docs updated for API/spec/status changes — `docs/CONFIGURATION.md` updated with `structuredOutput` rows in both extraction and synthesis config tables. `docs/PIPELINE.md` updated with YAML example.
- [x] Regression tests added for each bug fix — Two new tests in `provider.test.ts`: (1) config-driven format omission, (2) consecutive malformed 200 auto-recovery with format field removal.
- [x] Lint/typecheck/tests pass locally — `bun run typecheck` ✅ all packages. `bun run lint` / `bun run format` only pre-existing monorepo-wide errors (1727 lint, 148 format — none in changed files). `bun test` 1751 pass, 37 fail — all failures pre-existing (verified by stashing changes).

N/A — no migrations.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Note: 37 test failures are all pre-existing (verified by running tests with changes stashed). No new failures introduced. Pre-existing failures include: 5 `createCodexProvider` tests (no Codex binary), 1 CORS test, 1 skills route timeout, 1 `resolveSummaryProvider` codex test, and ~28 worker integration tests.

## AI disclosure

### Problem detail

GitHub Copilot's API rejects Signet's structured output `format` field (`{"type": "json_schema", "schema": {"type": "object", "additionalProperties": true}}`) but wraps the rejection in HTTP 200 with `"parts": []` instead of returning a proper 422. The daemon only detects 422 rejections (lines 2087-2118 of provider.ts), so `structuredOutputSupported` stays `true` forever — every extraction and synthesis call fails silently.

### Solution

1. **Config option** — `structuredOutput: false` in `agent.yaml` disables the `format` field entirely. Safe for all models because extraction prompts already include "Return ONLY the JSON object" instructions and `stripFences()` handles cleanup.

2. **Auto-detection** — After two consecutive malformed 200 responses (original POST + retry after session reset), the provider disables `structuredOutputSupported` and retries without the format field — recovering transparently without user intervention.

3. **Proper scoping** — `structuredOutput` is a first-class field on both `PipelineExtractionConfig` and `PipelineSynthesisConfig`. The synthesis config falls back to the extraction value when omitted, so a single `extraction.structuredOutput: false` covers both pipelines while allowing per-pipeline override.


PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): enable Signet pipeline through GitHub Copilot providers` (commit `f49e1ca3`) now and I'll follow up shortly with feedback.

packages/connector-oh-my-pi/src/extension-bundle.ts
```diff
@@ -5,43 +5,43 @@
  * Re-run the build to regenerate from ../oh-my-pi-extension/dist/signet-oh-my-pi.mjs
```
PR-Reviewer-Ant, line 5: This file is explicitly marked as a generated artifact ('Re-run the build to regenerate from ../oh-my-pi-extension/dist/signet-oh-my-pi.mjs'). The PR description mentions no intentional changes to the oh-my-pi extension, yet the bundle has been regenerated with wholesale variable renaming throughout. This may bundle in unintended changes from an intermediate build state of the oh-my-pi extension. The regeneration should either be explained and confirmed deliberate, or reverted if it was an accidental side-effect of the local build environment.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f49e1ca3`

[Automated Bot Review] PR #498 claims fixes to the Signet pipeline for GitHub Copilot-backed models, but the provided diff contains only a regenerated extension bundle. The substantive changes (provider.ts, daemon.ts, memory-config.ts, core/types.ts, summary-worker.ts) are entirely absent from the visible diff, making it impossible to validate the implementation against the stated goals.

**Confidence:** Low [missing_cross_module_context, ambiguous_requirements] - The visible diff covers only packages/connector-oh-my-pi/src/extension-bundle.ts — a generated artifact. The PR description names five additional files that carry the entire logic of this fix (packages/daemon/src/pipeline/provider.ts, packages/daemon/src/daemon.ts, packages/daemon/src/pipeline/summary-worker.ts, packages/core/src/types.ts, packages/daemon/src/memory-config.ts). None of these appear in the supplied diff. Without those files, none of the three stated mechanisms (structuredOutput config flag, consecutive-200 auto-detection, provider creation wiring) can be inspected, making correctness, safety, and race-condition analysis impossible.


- packages/daemon/src/pipeline/provider.ts:1 - The PR description states that after two consecutive malformed HTTP 200 responses the provider permanently disables structuredOutputSupported for the session. This file is not present in the supplied diff. Before merge, reviewers must verify: (1) the counter and flag are session-scoped, not shared across concurrent sessions (shared mutable state would be a data-corruption / correctness bug); (2) the flag is never reset mid-session in a way that could re-trigger the rejection loop; (3) there is no TOCTOU window where two concurrent extractions both observe count==1 and both attempt the retry, leading to two retries and double-writes.
- packages/daemon/src/memory-config.ts:1 - The new structuredOutput boolean is read from agent.yaml per the PR description, but this file is absent from the diff. Reviewers should confirm: the field is validated and defaults to true (not undefined/null) so existing configs are unaffected; the flag is not silently coerced in a way that disables structured output for models that support it correctly.
- packages/daemon/src/daemon.ts:1 - The PR states enableStructuredOutput is threaded through extraction, synthesis, and summary provider creation. This file is absent from the diff. Reviewers must confirm all three provider construction call-sites receive the flag, and that a missing or false value in one branch does not silently fall back to the old behaviour in a mismatched way (e.g. synthesis enabled but extraction disabled).
- docs/API.md:1 - A new user-facing config key (pipeline.extraction.structuredOutput) is introduced in agent.yaml. Per AGENTS.md PR checklist, docs/API.md must be updated in the same PR. The diff does not include any documentation update. This is a doc/spec drift violation.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3575a2bc`

[Automated PR Review — bot identity: pr-reviewer] This PR adds GitHub Copilot structured-output compatibility to the Signet daemon pipeline. The visible diff is almost entirely a large regenerated bundle (`extension-bundle.ts`). The substantive logic changes (provider.ts, daemon.ts, summary-worker.ts, memory-config.ts, types.ts) were truncated from the diff context, so several findings below are inferred from the PR description plus the bundle anomaly.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The bundle-scope-leak finding is directly provable from the visible diff: the extension-bundle.ts header explicitly states it is generated from `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs`, and the diff contains new Lexer content (lines 151–165 of the diff add previously absent `parsePlainScalar`/`pushCount` branches) that go beyond variable renaming — confirming an unreviewed dependency or source change. The race-condition and doc-drift concerns cannot be fully verified because provider.ts, daemon.ts, summary-worker.ts, and docs/API.md were not present in the supplied diff context.


- packages/daemon/src/pipeline/provider.ts:1 - Based on the PR description: 'After two consecutive malformed 200 responses… the provider disables `structuredOutputSupported` and retries without the format field.' If `structuredOutputSupported` and the consecutive-failure counter are mutable instance fields on a provider object shared across concurrent extraction/synthesis calls (which is the common pattern in this codebase), there is a race condition:

1. Two in-flight requests both observe a malformed 200 simultaneously.
2. Both increment the counter; the flag is set twice, which is benign.
3. BUT if a third concurrent request already read `structuredOutputSupported = true` and built its request body with `format` before either counter write landed, it will still send the rejected field — meaning the recovery window can exceed two requests under concurrent load.

Additionally: the PR description says 'original POST + retry after session reset' implies the counter relies on a session-reset happening between attempt 1 and attempt 2. If the session reset is asynchronous, the counter may not be correctly sequenced. Please confirm the counter is per-session-context (not per-provider-singleton) and that reads/writes are serialized or atomic.
- packages/daemon/src/pipeline/provider.ts:1 - The auto-detection recovery silently swallows the first two failed requests. The PR description says 'recovering transparently without user intervention,' but the two calls that triggered the detection produce no extracted memories and no summaries. Under low-session-frequency usage (e.g. one session per day), users could lose memory extraction for two consecutive sessions before the flag self-corrects. The word 'transparent' is misleading; a warning log when the auto-disable triggers would satisfy the AGENTS.md requirement that errors are not swallowed silently ('Do not swallow errors. Log with context.').
- docs/API.md:1 - AGENTS.md rule 'Doc/spec drift': 'Update behavior, API, schema, and status docs in the same PR. Keep `docs/API.md` accurate when affected.' The PR introduces a new user-facing `agent.yaml` config key (`pipeline.extraction.structuredOutput`). The PR description and diff context contain no evidence that `docs/API.md` was updated to document this option, its default value, or which providers require it. This is a correctness gap for users of non-Copilot providers who may accidentally set this flag.
- packages/daemon/src/pipeline/provider.ts:1 - AGENTS.md: 'Every bug fix needs a test that would fail before the fix.' The PR claims 23/23 existing tests pass and adds two new tests. However, the regression test for the core described bug — Copilot returning HTTP 200 with `"parts": []` causing `structuredOutputSupported` to remain `true` forever — must be a test that would have **failed** against the old code. Please confirm this test exists and that it covers the steady-state case (i.e. `structuredOutputSupported` remains `false` after detection across subsequent calls, not just for the two recovery calls).


- `packages/connector-oh-my-pi/src/extension-bundle.ts:5` - This file is declared at the top as auto-generated from `../oh-my-pi-extension/dist/signet-oh-my-pi.mjs`. The diff is not merely variable renaming: lines roughly 151–165 of the diff add new `parsePlainScalar` / `pushCount` / `pushToIndex` branches to the bundled YAML Lexer class that did not exist in the prior bundle. This indicates either (a) a YAML library version bump in the oh-my-pi-extension package or (b) an upstream source change in that package, neither of which is mentioned in the PR description. This is a scope leak. The PR's stated goal is GitHub Copilot provider support in the daemon pipeline, which has no dependency on the oh-my-pi browser extension bundle. Committing a silently-updated third-party bundle without a corresponding diff of the source (oh-my-pi-extension) or an explanation of why the dependency changed makes this change unauditable. Per AGENTS.md, the build order runs `build:oh-my-pi-extension → build:connector-oh-my-pi`; if this was regenerated as a side-effect of running `bun run build`, the triggering source change needs to be isolated and reviewed separately or the bundle reverted to its prior state.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `d5cf9a46`

```pr-review-json
{
  "summary": "[Automated Review — pr-reviewer bot] PR #498 adds a `structuredOutput` config option and auto-detection for GitHub Copilot's silent 200-with-empty-body rejection. Three issues found: one blocking cross-config coupling bug, one warning about an undocumented side-effect (system prompt stripped), and one warning about inconsistent polling in the fallback path.",
  "verdict": "request_changes",
  "confidence": {
    "level": "high",
    "justification": "All three findings are directly provable from the diff. The synthesis/summary config bug is explicit at daemon.ts lines 2082 and summary-worker.ts line 1509 (both read `extraction.structuredOutput`). The system-prompt omission is confirmed by the new test assertion at provider.test.ts line 749. The polling gap is visible in the new provider.ts block where `parsePostResponse` is called once without a poll loop."
  },
  "comments": [
    {
      "file": "packages/daemon/src/daemon.ts",
      "line": 2082,
      "body": "**[Blocking] Synthesis provider reads `extraction.structuredOutput` — wrong config namespace.**\n\nThis line reads `memoryCfg.pipelineV2.extraction.structuredOutput` to configure the *synthesis* OpenCode provider. The same pattern appears in `summary-worker.ts` line 1509.\n\nThis creates a semantic bug: if a user runs extraction on Ollama/none and synthesis on OpenCode (Copilot), there is no way to disable structured output for synthesis because `extraction.structuredOutput` is documented and named as an extraction-scoped key. The user must set a config field that is semantically wrong for their topology.\n\nThe PR description says "Wire config through `agent.yaml` → daemon extraction/synthesis/summary providers", implying all three read the same field by design. But the field is nested under `pipeline.extraction.*` in the YAML schema, so users with split provider configurations have no working escape hatch.\n\n**Minimal fix**: add `structuredOutput?: boolean` to `PipelineSynthesisConfig`, read it in `loadPipelineConfig`, and pass `memoryCfg.pipelineV2.synthesis.structuredOutput` here (and in `summary-worker.ts`). Fall back to `extraction.structuredOutput` for the common single-provider case if desired, but expose synthesis as its own key.",
      "severity": "blocking"
    },
    {
      "file": "packages/daemon/src/pipeline/summary-worker.ts",
      "line": 1509,
      "body": "**[Blocking] Same cross-config coupling as daemon.ts line 2082.** Summary provider reads `cfg.pipelineV2.extraction.structuredOutput` instead of a synthesis/summary-scoped flag. Same fix applies.",
      "severity": "blocking"
    },
    {
      "file": "packages/daemon/src/pipeline/provider.test.ts",
      "line": 749,
      "body": "**[Warning] Test asserts that `system` is also `undefined` when `enableStructuredOutput: false` — undocumented and potentially harmful side-effect.**\n\n```ts\nexpect(capturedBody.system).toBeUndefined();\n```\n\nThe `system` field in the OpenCode request body carries the extraction system prompt (JSON schema instructions, output constraints, etc.). Stripping it along with `format` means Copilot models receive zero system-level guidance about the expected JSON output shape. The PR description justifies this with "extraction prompts already include 'Return ONLY the JSON object'", but that claim is about the user-prompt text, not the system field.\n\nIf `buildMessageBody` only populates `system` when structured output is on, that coupling should be explicit and documented. If the system prompt genuinely contains only format-enforcement guidance (and is fully redundant with the user prompt), that should be verified. As written, Copilot users will silently get lower-quality extraction with no indication why.",
      "severity": "warning"
    },
    {
      "line": 2190,
      "body": "**[Warning] Third POST (after SO disabled) does not poll — inconsistent with the retry path.**\n\nThe existing retry path (steps 1–2) both go through the polling loop (`parsePostResponse` → poll GET in a loop until assistant message or timeout). The new fallback POST at this line calls `parsePostResponse(fallbackRes, fallbackSid)` exactly once with no poll loop. If the Copilot backend returns another empty 200 for the third request (possible under load), the code silently falls through to the Ollama fallback with no log line specific to this failure.\n\nAlso, at this point the original `controller.signal` (AbortSignal) may already be near expiry after two full poll timeouts. The third request inherits the same signal, so a tight timeout budget can cause it to abort immediately — making the recovery path unreliable under the default timeout.\n\nConsider either: (a) reusing the same polling loop for the fallback POST, or (b) creating a fresh AbortController with a minimum guaranteed window (e.g. `Math.min(remaining, 10_000)`).",
      "severity": "warning"
    }
  ]
}
```

Ostico:
## Response to Automated Review (commit `d5cf9a46`)

### 1. "Blocking: Synthesis/summary reads `extraction.structuredOutput`"

This follows the existing inheritance pattern for synthesis fields. Per `docs/CONFIGURATION.md`:

> If the `synthesis` block is omitted entirely, Signet falls back to the resolved extraction provider, model, endpoint, and timeout.

`provider`, `model`, `endpoint`, and `timeout` all inherit from extraction when the synthesis block is omitted. `structuredOutput` follows the same convention — single source of truth for the common case where extraction and synthesis share a provider.

That said, the reviewer is correct that unlike `timeout` (which has a resolved synthesis-scoped value via `loadPipelineConfig`), `structuredOutput` reads `extraction.structuredOutput` directly at daemon.ts:2082 and summary-worker.ts:1509 rather than going through synthesis-scoped resolution. This is a deliberate simplification: the GitHub Copilot structured-output issue affects **all** OpenCode provider instances identically (extraction, synthesis, and summary all hit the same upstream API). A future PR can add `synthesis.structuredOutput` with fallback to `extraction.structuredOutput` for split-provider topologies.

Not blocking for this PR — the real-world use case this PR fixes (all three pipelines using the same OpenCode/Copilot provider) works correctly.

### 2. "Blocking: Same cross-config coupling in summary-worker.ts"

Same reasoning as above.

### 3. "Warning: Test asserts `system` is undefined — undocumented side-effect"

The test assertion reflects `buildMessageBody` behavior (provider.ts:1974). When `structured` is true and `structuredOutputSupported` is true, `buildMessageBody` sets both `format` and `system`. When either is false, neither is set. This is intentional: the `system` field contains only JSON schema enforcement instructions (`"Return ONLY valid JSON matching the requested schema"`), not extraction domain guidance. The extraction domain prompt lives in the `parts[].text` user content.

Verified empirically: the daemon successfully extracts memories from GitHub Copilot models without the `system` field. The user-prompt text already contains complete output format instructions.

### 4. "Warning: Third POST does not poll"

~~Correction from my initial analysis~~: the third POST **does** go through the full poll cycle. `parsePostResponse` (provider.ts:2070-2078) calls `parseMessagePayload` first, then falls back to `pollForAssistantMessage` if the initial parse returns null — identical behavior to the first two POSTs. The reviewer's concern about inconsistent polling is therefore not applicable.

The AbortSignal timeout budget concern is valid in edge cases. The third POST shares the original controller, so if the first two attempts consumed most of the timeout budget, the fallback window could be tight. In practice the default 90s timeout provides sufficient headroom for three sequential request+poll cycles, and worst-case the Ollama fallback (provider.ts:2194) catches any remaining failures.

---

**Action taken**: Added `structuredOutput` documentation to `docs/CONFIGURATION.md` (extraction config table) and `docs/PIPELINE.md` (YAML reference example). Commit incoming.

packages/connector-oh-my-pi/src/extension-bundle.ts
Ostico: Resolved in commit `d5cf9a46` — all three bundle files reverted to match `origin/main` exactly. The regeneration was an unintended side-effect of `bun run build`.

packages/daemon/src/daemon.ts
```diff
@@ -2078,6 +2079,7 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 								ollamaFallbackBaseUrl: synthesisOllamaFallbackBaseUrl,
 								ollamaFallbackMaxContextTokens: ollamaFallbackMaxContextTokens,
 								defaultTimeoutMs: memoryCfg.pipelineV2.synthesis.timeout,
+								enableStructuredOutput: memoryCfg.pipelineV2.extraction.structuredOutput,
```
PR-Reviewer-Ant: **Copy-paste bug: synthesis reads extraction's `structuredOutput`.**

The synthesis OpenCode provider is wired with `enableStructuredOutput: memoryCfg.pipelineV2.extraction.structuredOutput`, but `PipelineSynthesisConfig` has no `structuredOutput` field of its own. This silently couples synthesis behavior to the extraction config block.

For any user whose `extraction.provider` is **not** `opencode` but whose `synthesis.provider` **is** `opencode` (e.g., Ollama for extraction, OpenCode+Copilot for synthesis), the `structuredOutput: false` flag under `extraction:` has no semantic relationship to extraction — but they must still set it there to fix synthesis. The documentation in `CONFIGURATION.md` documents this flag only under the Extraction section, so those users will not find the knob and will be stuck with silent synthesis failures.

Fix options:
1. Add `structuredOutput?: boolean` to `PipelineSynthesisConfig` and load it in `memory-config.ts`, then wire `memoryCfg.pipelineV2.synthesis.structuredOutput` here.
2. At minimum, document in `CONFIGURATION.md` and `PIPELINE.md` that this flag controls both extraction and synthesis when either uses the OpenCode provider.

packages/daemon/src/pipeline/summary-worker.ts
```diff
@@ -1506,6 +1506,7 @@ export async function resolveSummaryProvider(cfg: ReturnType<typeof loadMemoryCo
 				ollamaFallbackBaseUrl: "http://127.0.0.1:11434",
 				ollamaFallbackMaxContextTokens: ollamaMaxContextTokens,
 				defaultTimeoutMs: timeout,
+				enableStructuredOutput: cfg.pipelineV2.extraction.structuredOutput,
```
PR-Reviewer-Ant: Same root issue as `daemon.ts`: the summary provider (used for OpenCode synthesis in the summary worker) is created with `enableStructuredOutput: cfg.pipelineV2.extraction.structuredOutput`. A user whose synthesis uses OpenCode but extraction does not would need to set the flag under the wrong config key. See the daemon.ts comment for the full analysis.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -971,6 +997,52 @@ describe("createOpenCodeProvider", () => {
...
+		expect(result).toBe("recovered without format");
+
+		// The third POST body should NOT have the format field
+		expect(postBodies.length).toBeGreaterThanOrEqual(3);
+		expect(postBodies[0]?.format).toBeDefined(); // first: had format
+		expect(postBodies[1]?.format).toBeDefined(); // second (retry): still had format
+		expect(postBodies[2]?.format).toBeUndefined(); // third: format disabled
+	}, 15000);
+
 	it("generate() does not disable structured output on an unrelated 400", async () => {
```
PR-Reviewer-Ant, line 1046: **Test fidelity concern: auto-detection recovery test may not model real OpenCode behavior.**

The new test mocks the third POST to return `openCodeResponse("recovered without format")` directly in the POST response body, while mocking GET polls to return `[]` (empty). Other tests in this file model OpenCode responses arriving via GET polling (the GET mock returns the data, not the POST). If the real OpenCode API always delivers content via GET polling rather than in the POST body, `parsePostResponse` on the third POST response would return null, the recovery block would silently fall through, and the auto-detection feature would not actually recover — it would just disable `structuredOutputSupported` and still fall back to Ollama/error.

Please verify whether `parsePostResponse` can successfully extract a complete result from a POST response body in the real (non-mocked) OpenCode flow. If not, the fallback block needs the same GET-polling loop used in the normal `generate()` path.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3704ac23`

[Automated Review — pr-reviewer bot] Two issues identified in this PR: one blocking config-wiring bug affecting mixed-provider setups, and one warning about test fidelity for the auto-detection recovery path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The synthesis config wiring bug is directly visible in the diff: line +7 of the daemon.ts hunk reads `memoryCfg.pipelineV2.extraction.structuredOutput` inside the synthesis OpenCode provider creation block, and `PipelineSynthesisConfig` (visible in the full types.ts content) has no `structuredOutput` field. The test-fidelity concern is grounded in the mocking pattern in the new test (POST returns data directly) vs the existing provider tests where OpenCode results come via GET poll.

packages/daemon/src/daemon.ts
Ostico: Fixed in 1896d046. `structuredOutput` is now a first-class field on `PipelineSynthesisConfig` (types.ts) with its own parsing in `memory-config.ts`. Fallback chain: `synthesis.structuredOutput` → `extraction.structuredOutput` → `undefined`. Both `daemon.ts` and `summary-worker.ts` now read from `synthesis.structuredOutput`.

packages/daemon/src/pipeline/summary-worker.ts
Ostico: Same fix applied here — now reads `cfg.pipelineV2.synthesis.structuredOutput` (commit 1896d046).

packages/daemon/src/pipeline/provider.test.ts
Ostico: Valid concern. The test covers `parsePostResponse` which handles both paths: inline JSON parse and poll fallback (line 2077 of provider.ts). The mock returning data directly exercises the inline parse path, which is the common case for most providers. A poll-variant test would strengthen coverage but isn't blocking — the detection logic being tested (200-with-error → `structuredOutputSupported = false`) fires before the poll path.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1962,7 +1964,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		return sessionId;
```
PR-Reviewer-Ant, line 1964: **Fragile default: `DEFAULT_OPENCODE_CONFIG.enableStructuredOutput: true` is silently overridden to `undefined` on every call-site.**

The daemon wires the flag as:
```ts
createOpenCodeProvider({
  ...
  enableStructuredOutput: memoryCfg.pipelineV2.extraction.structuredOutput, // undefined when not set in yaml
})
```

When `structuredOutput` is absent from `agent.yaml`, the call passes `enableStructuredOutput: undefined` explicitly. Object spread `{ ...DEFAULT_OPENCODE_CONFIG, enableStructuredOutput: undefined }` sets `cfg.enableStructuredOutput = undefined`, shadowing the `true` default. The current initialization `cfg.enableStructuredOutput !== false` returns `true` for `undefined`, so behavior is correct — but the default value in `DEFAULT_OPENCODE_CONFIG` is effectively dead code and misleads future readers.

If someone later refactors the check to `cfg.enableStructuredOutput ?? true` (natural), it would break differently. Or if the condition changes to `=== true`, undefined would become `false` — silently disabling structured output for all users who don't set the config key.

Fix: use `cfg.enableStructuredOutput ?? true` in the initialization to make the default explicit and not reliant on the `!== false` idiom:
```ts
let structuredOutputSupported = cfg.enableStructuredOutput ?? true;
```
This also makes `DEFAULT_OPENCODE_CONFIG.enableStructuredOutput: true` meaningful.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2162,6 +2164,30 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+				const fallbackSid = await createSession();
+				sessionId = fallbackSid;
+				const fallbackRes = await fetch(`${cfg.baseUrl}/session/${fallbackSid}/message`, {
+					method: "POST",
+					headers: { "Content-Type": "application/json" },
+					body: buildMessageBody(prompt, false),
+					signal: controller.signal,
+				});
+				if (fallbackRes.ok) {
+					const fallbackParsed = await parsePostResponse(fallbackRes, fallbackSid);
```
PR-Reviewer-Ant, line 2186: **Silent path when the auto-detection-triggered fallback POST returns a non-OK status.**

When the third POST (after `structuredOutputSupported` is disabled) returns a non-OK response, the `if (fallbackRes.ok)` block is skipped and execution falls through to the generic `logger.warn("OpenCode response remained malformed after retry; using fallback")`. That warn message was written for the two-malformed-200 case, not for the case where the third attempt returned a proper error code.

From a debugging standpoint, a user investigating why Copilot extraction is failing would see:
1. `logger.info`: "Consecutive malformed 200 responses; disabling structured output" — suggests the feature kicked in
2. `logger.warn`: "OpenCode response remained malformed after retry" — misleading, implies a third malformed 200, not a 4xx/5xx failure

Minimal fix: add a log entry before the fall-through when `fallbackRes.ok` is false:
```ts
if (fallbackRes.ok) {
  const fallbackParsed = await parsePostResponse(fallbackRes, fallbackSid);
  if (fallbackParsed) return fallbackParsed;
} else {
  logger.warn("pipeline", "Structured-output-disabled fallback POST failed", {
    sessionId: fallbackSid,
    status: fallbackRes.status,
  });
}
```
This satisfies the AGENTS.md rule: "Do not swallow errors. Log with context."

packages/daemon/src/memory-config.ts
```diff
@@ -805,6 +809,12 @@ export function loadPipelineConfig(yaml: Record<string, unknown>): PipelineV2Con
 			timeout: resolvedSynthesisTimeout,
 			maxTokens: clampPositive(synthesisRaw?.maxTokens ?? synthesisRaw?.max_tokens, 1000, 32000, d.synthesis.maxTokens),
 			idleGapMinutes: clampPositive(synthesisRaw?.idleGapMinutes, 1, 1440, d.synthesis.idleGapMinutes),
+			structuredOutput: (() => {
```
PR-Reviewer-Ant, line 812: **No config-parsing test for the synthesis `structuredOutput` fallback-to-extraction inheritance.**

The new synthesis resolution logic:
```ts
structuredOutput: (() => {
  const candidate = synthesisRaw?.structuredOutput;
  if (typeof candidate === 'boolean') return candidate;
  const extractionCandidate = extractionRaw?.structuredOutput;
  return typeof extractionCandidate === 'boolean' ? extractionCandidate : undefined;
})(),
```
…has no test in the test suite. If the fallback order were accidentally reversed (extraction checked first, synthesis second), or if the boolean guard were removed, no existing test would catch it. AGENTS.md requires: "Every bug fix needs a test that would fail before the fix." The config-parsing change is part of the fix path for synthesis-over-OpenCode topologies. A test covering: (1) synthesis own value wins, (2) synthesis absent → extraction value inherited, (3) both absent → undefined, would close this gap.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `1896d046`

[Automated Review — pr-reviewer bot] PR #498 resolves GitHub Copilot silent 200-rejection handling for the Signet pipeline. All blocking issues from prior rounds (synthesis reading extraction's structuredOutput, extension bundle scope leak) are fixed in the final commit (1896d046). Two new findings remain that were not raised or addressed in prior rounds.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The DEFAULT_OPENCODE_CONFIG shadowing concern is directly provable from the diff: the spread `{ ...DEFAULT_OPENCODE_CONFIG, enableStructuredOutput: undefined }` nullifies the documented default of `true` when undefined is passed from config. The missing-log finding is visible from the new fallback block structure. Confidence is medium rather than high because the full `generate()` function body in provider.ts is not visible in the diff, making it impossible to independently verify Ostico's claim that `parsePostResponse` covers the polling path on the third POST.

NicholaiVogel:
@Ostico Looks good from my end, merging this. Thanks for the contribution!

P.S. signet-first looks cool.

NicholaiVogel (approved):
Approved. The main config-shape issue has been addressed: structuredOutput is now scoped on both extraction and synthesis, synthesis falls back to extraction when omitted, and the docs/checks are in a reasonable state. Remaining reviewer notes look like polish or follow-up territory rather than merge blockers.

PR #500, opened 2026-04-13
fix(hooks): prompt agents to check memory before acting

NicholaiVogel:
## Summary

- add a Memory Check Loop to session-start system prompt injection
- add per-turn Memory Check guidance to user-prompt-submit injection, including no strong match paths
- mirror the prompt shaping in daemon-rs shadow hook handling
- document the new hook behavior and add regression coverage

## Why

Agents were being told that memory tools exist, but not given an explicit pre-execution habit for using them. This makes it too easy for an agent to skip recall before touching context-sensitive work. The new prompt shape gives agents a compact trigger, procedure, pitfalls, and verification loop.

## Validation

- `bun run build:core`
- `bun test packages/daemon/src/hooks.test.ts packages/daemon/src/hooks.prompt-submit.test.ts`
- `cargo test -p signet-daemon hooks`
- `bunx biome check packages/daemon/src/hooks.ts packages/daemon/src/hooks.test.ts packages/daemon/src/hooks.prompt-submit.test.ts docs/HOOKS.md`
- `rustfmt --edition 2024 --check crates/signet-daemon/src/routes/hooks.rs`
- `git diff --check`

## Notes

`bun run --filter '@signet/daemon' build:types` still fails on pre-existing broader daemon type issues, including rootDir/core source inclusion, logger category mismatches, and unrelated route/test typing errors. Those failures were not introduced by this prompt-shaping patch.

## PR readiness checklist

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) (routine bug-fix/docs/test follow-up; no spec-governed contract changed)
- [x] Agent scoping verified on all new/changed data queries (no new data queries added)
- [x] Input/config validation and bounds checks added (not applicable; no new external input/config surface)
- [x] Security checks applied to admin/mutation endpoints (not applicable; no admin/mutation endpoint changes)

### DIFF
--- a/packages/daemon/src/hooks.prompt-submit.test.ts
+++ b/packages/daemon/src/hooks.prompt-submit.test.ts
@@ -133,6 +133,7 @@ describe("handleUserPromptSubmit observability", () => {
 		expect(result.inject).toContain("## Memory Check");
 		expect(result.inject).toContain("No strong automatic memory match was injected");
 		expect(result.inject).toContain("run 1-3 targeted Signet recalls before executing commands");
+		expect(result.inject).toContain("save it with /remember or memory_store");
 		const submitCalls = infoMock.mock.calls.filter((call) => call[1] === "User prompt submit");
 		expect(submitCalls).toHaveLength(1);
 		const payload = submitCalls[0]?.[2];
@@ -311,10 +312,32 @@ describe("handleUserPromptSubmit observability", () => {
 		expect(result.inject).toContain("Current Date & Time");
 		expect(result.inject).toContain("## Memory Check");
 		expect(result.inject).toContain("No strong automatic memory match was injected");
+		expect(result.inject).toContain("save it with /remember or memory_store");
 		expect(result.inject).not.toContain("[signet:recall");
 		const submitCalls = infoMock.mock.calls.filter((call) => call[1] === "User prompt submit");
 		expect(submitCalls).toHaveLength(1);
 		const payload = submitCalls[0]?.[2];
 		expect(payload?.engine).toBe("low-confidence");
+	});
+
+	it("returns Memory Check guidance when hybrid recall fails", async () => {
+		hybridRecallMock.mockRejectedValueOnce(new Error("synthetic recall failure"));
+
+		const result = await handleUserPromptSubmit(
+			{
+				harness: "vscode-custom-agent",
+				userMessage: "show memory failure behavior",
+				sessionKey: "session-recall-failure",
+			},
+			makeDeps(),
+		);
+
+		expect(result.memoryCount).toBe(0);
+		expect(result.inject).toContain("Current Date & Time");
+		expect(result.inject).toContain("## Memory Check");
+		expect(result.inject).toContain("No strong automatic memory match was injected");
+		expect(result.inject).toContain("run 1-3 targeted Signet recalls before executing commands");
+		expect(result.inject).toContain("save it with /remember or memory_store");
+		expect(errorMock).toHaveBeenCalledTimes(1);
 	});
 });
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(hooks): prompt agents to check memory before acting` (commit `66973c58`) now and I'll follow up shortly with feedback.

packages/daemon/src/hooks.ts
```diff
@@ -2692,7 +2714,7 @@ export async function handleUserPromptSubmit(
 		);
```
PR-Reviewer-Ant, line 2718: The catch block now calls `buildNoStrongMemoryMatchInject(metadataHeader)` instead of returning `inject: ""`. If `metadataHeader` is declared and assigned **before** the try block this is safe, but if the error is thrown during header construction (e.g., `getMetadataHeader()` throws) then `metadataHeader` would be `undefined` at catch time, producing a malformed inject or a secondary throw. The full function body is not in the diff context, so this cannot be confirmed from the diff alone. Please verify (or add a comment) that `metadataHeader` is always assigned before the try block begins.

packages/daemon/src/hooks.prompt-submit.test.ts
```diff
@@ -130,6 +130,9 @@ describe("handleUserPromptSubmit observability", () => {
 		);
```
PR-Reviewer-Ant, line 130: The PR description says it adds regression coverage, and it does cover the happy-path no-match and match cases. However, the changed catch branch (`} catch (e) { return { inject: buildNoStrongMemoryMatchInject(metadataHeader), ... } }`) is not exercised by any test. A test that forces the inner try body to throw (e.g., by having `hybridRecall` reject with an error after `metadataHeader` is built) would confirm both that the catch fires and that `metadataHeader` is in scope with its expected value.

packages/daemon/src/hooks.ts
```diff
@@ -528,15 +534,31 @@ export function applyTokenBudget(inject: string, mainBudget: number): string {
 function buildPromptRecallInject(metadataHeader: string, lines: ReadonlyArray<string>): string {
```
PR-Reviewer-Ant, line 534: The `buildPromptRecallInject` footer previously read: `"if you need deeper history, use /recall or memory_search. if you learn something durable, save it with /remember or memory_store."` — one sentence covering both recall and save hints. The new version removes the recall hint from this footer (it moved into the `## Memory Check` block) and keeps only the save hint. For the no-match path (`buildNoStrongMemoryMatchInject`) the save hint is absent entirely. This is likely intentional, but agents working on a no-context turn will no longer be reminded to save durable facts. Worth confirming this is the desired UX.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `66973c58`

[Automated Bot Review - pr-reviewer] PR #500 adds a Memory Check Loop to session-start system prompt injection and per-turn Memory Check guidance to user-prompt-submit, mirrored in the Rust shadow daemon. The JS/Rust implementations are in close parity. Two issues worth discussing before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The JS/Rust parity and test coverage claims are directly verifiable from the diff. The catch-block concern (see comment below) requires seeing the full `handleUserPromptSubmit` body to confirm `metadataHeader` is always assigned before the catch scope is entered — that function body is not in the provided context.

PR #503, opened 2026-04-13
fix(daemon,openclaw,cli): five correctness bugs found via sqmd structural review

aaf2tbz:
## Summary

Five correctness bugs found via sqmd structural search, with regression tests for all fixes.

### Bug 1: Map-mutation-during-iteration in openclaw `cleanupTimedMap`
`packages/adapters/openclaw/src/index.ts:876` — Fixed with collect-then-delete pattern. Note: JS `for...of` + `delete` of current key is defined behavior per the spec, so this is a defensive hardening rather than a correctness fix. **Regression test added** (2 tests).

### Bug 2: `markDeprecatedVersions` not applied to anthropic/openrouter models
`packages/daemon/src/pipeline/model-registry.ts:489` — Now wraps both with the deprecation marker (idempotent). **Regression tests added** (5 tests including OpenRouter-style `provider/model` IDs and cross-provider non-deprecation).

### Bug 3: Null `policyGroup` causes group-scope SQL to return nothing
`packages/daemon/src/memory-search.ts` — Falls back to isolated scope when policyGroup is null. **Regression test added**.

### Bug 4: LIKE metacharacter injection in Python and TypeScript
Python scripts + `buildFilterClause` + `buildProjectionWhere` — Added `escapeLike`/`escape_like` with ESCAPE clause. The `who` filter in `buildFilterClause` uses exact match (`m.who = ?`), not LIKE, so it is not vulnerable. **Regression tests added** (TS unit tests + pytest-discoverable `test_like_escaping.py` in `tests/` subdirectory of both template copies).

### Bug 5: `hasMore` pagination false-negative in UMAP projection
`packages/daemon/src/umap-projection.ts` — Over-fetch-by-1 sentinel for `hasMore`. Unconditional `typeof(e.vector) = 'blob'` COUNT provides stable `total` for dashboard display across all pages. Sentinel handles `hasMore` in the no-discard case; COUNT-based `hasMore` only activates when discards occur. **Regression tests added** (3 tests including discard-path with offset>0).

## sqmd review rounds

- **R1**: Found 5 bugs via sqmd structural search. Fixed all.
- **R2**: LIKE injection in TS daemon, missing regression tests for bugs 3/4/5, hasMore edge case. Fixed.
- **R3**: Missing regression tests for bugs 1/2, duplicate escapeLike, test quality, total/hasMore inconsistency. Fixed.
- **R4**: Residual hasMore false-negative when discards + more pages, discard-path test, JSDoc direction. Fixed.
- **R5**: Import ordering for Biome compliance, bounded vector-only re-count instead of full-row scan. Fixed.
- **R6**: Import ordering (confirmed correct), `who` filter confirmed safe (exact match), `total` inflation in no-discard COUNT path. Fixed by unifying COUNT to always use `typeof(e.vector) = 'blob'`.
- **R7**: `who` LIKE injection (N/A), `toEmbeddingRow` other discard reasons (N/A — schema NOT NULL TEXT), unconditional COUNT perf (minor), `cleanupTimedMap` framing (noted in description), discard-path offset>0 test (already present). No code changes needed.
- **R8**: Python LIKE-escaping regression tests missing for both template copies. Added inline `_test_like_escaping()`.
- **R9**: Unconditional COUNT perf (acknowledged), unused `cursor` variable in both memory.py copies. Removed.
- **R10**: Python tests not CI-discoverable (moved to `test_like_escaping.py` with pytest), OpenRouter-style ID deprecation not tested (added 2 tests), unconditional COUNT perf (gated behind discard check). Fixed.
- **R11**: `total` undercount in no-discard path breaks dashboard display (restored unconditional COUNT — dashboard needs stable total), pytest files in template dirs not CI-discoverable (moved to `tests/` subdirectory, added `conftest.py`), per-page COUNT on discard path (noted, acceptable). Fixed.
- **R12**: sqmd-review — no_issues. Gated unconditional COUNT behind discard check (no-discard path uses cheap offset arithmetic). Removed `conftest.py` (sys.path injection was a no-op). Renamed test files to unique module names (`test_cli_like_escaping.py`, `test_signetai_like_escaping.py`) for pytest discoverability without `__init__.py` or `conftest.py`. All 6 pytest tests pass, 3 bun tests pass, typecheck clean.
- **R13**: Bot flagged R12's offset arithmetic for `total` as regression — non-terminal pages report `total=offset+limit` instead of actual row count, breaking dashboard display. Restored unconditional `typeof(e.vector) = 'blob'` COUNT for stable `total` across all pages. Sentinel still handles `hasMore` in no-discard case; COUNT-based `hasMore` only used when discards occur. 3 bun tests pass.

## Checklist


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon,openclaw,cli): five correctness bugs found via sqmd structural review` (commit `3fcdb135`) now and I'll follow up shortly with feedback.

packages/daemon/src/memory-search.ts
```diff
@@ -101,13 +101,13 @@ export function buildAgentScopeClause(
 			args: [agentId],
 		};
 	}
-	if (readPolicy === "group") {
+	if (readPolicy === "group" && policyGroup) {
```
PR-Reviewer-Ant, line 105: Bug 4 fixes LIKE metacharacter injection in the Python scripts, but the same class of vulnerability exists in this TypeScript file and was not addressed. In `buildFilterClause`, the tags branch constructs `args.push(\`%${t}%\`)` and passes it directly into `m.tags LIKE ?` without escaping `%` or `_` in the user-supplied tag value. A tag search containing `%` returns all rows with any tags. The SQL is parameterized so there is no injection into the query structure, but the LIKE wildcard semantics are broken. The fix is the same pattern as Bug 4: escape `\`, `%`, `_` before interpolation and add an `ESCAPE` clause. This is in the same file as the Bug 3 change.

packages/daemon/src/umap-projection.ts
```diff
@@ -418,7 +418,9 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
 	const rawRows = db.prepare(sql).all(...rowParams) as Record<string, unknown>[];
 	const rows = rawRows.map(toEmbeddingRow).filter((row): row is EmbeddingRow => row !== null);
 	const limit = requestedLimit ?? Math.max(0, total - offset);
-	const hasMore = offset + rows.length < total;
+	const hasMore = requestedLimit !== null
```
PR-Reviewer-Ant: The hasMore fix is an improvement but introduces a subtle false-negative. The new condition `offset + rows.length >= requestedLimit` is intended to detect a 'saturated page', but `rows.length` is post-`toEmbeddingRow` null-filter while `requestedLimit` rows were fetched from the DB. If `toEmbeddingRow` discards any rows (malformed vector blobs that slip through the `vector IS NOT NULL` guard), `rows.length < requestedLimit`, so `offset + rows.length >= requestedLimit` is false and `hasMore` is returned as false even though more rows exist. The root cause — `total` from `COUNT(*)` before the null filter — is not addressed. In normal operation this is rare, but a cleaner fix is to re-count after filtering or over-fetch by 1 to detect saturation directly.

packages/daemon/src/pipeline/model-registry.ts
```diff
@@ -486,14 +486,14 @@ async function _refreshRegistryInner(
 	}
 
 	if (anthropicModels.length > 0) {
-		state.models.set("anthropic", anthropicModels);
+		state.models.set("anthropic", markDeprecatedVersions(anthropicModels));
```
PR-Reviewer-Ant, line 489: The Bug 2 description is inaccurate. Both `discoverAnthropicModels` and `discoverOpenRouterModels` already apply `markDeprecatedVersions` on every return path (including fallbacks) in the full file as shown. Wrapping the results again in `_refreshRegistryInner` is idempotent and harmless, but the stated rationale — 'older versions never filtered out for anthropic/openrouter' — was not true before this change. The actual pre-fix state stored already-marked models. Consider verifying whether the sqmd analysis was looking at a stale snapshot of these two functions. The fix itself is safe to land.

packages/daemon/src/memory-search.ts
PR-Reviewer-Ant, line 105: AGENTS.md explicitly requires: 'Every bug fix needs a test that would fail before the fix' and 'A bug fix is not complete until it adds at least one durable prevention mechanism.' This PR fixes five bugs but adds zero new regression tests for any of them. The PR description notes Bug 3 is 'validated by existing memory-search.test.ts (7/7 pass)' — passing existing tests is not a regression test for the new behavior (the null policyGroup path was presumably not tested before if the bug existed). Bugs 1, 2, 4, and 5 also have no new tests. At minimum, tests should be added for: the null-policyGroup group-policy fallback (Bug 3), the LIKE metacharacter escaping in Python (Bug 4), and the hasMore pagination boundary (Bug 5).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `3fcdb135`

[Automated Review — pr-reviewer] Five targeted bug fixes, all structurally correct, but three substantive issues require attention: (1) the same LIKE metacharacter injection that Bug 4 fixes in Python exists unaddressed in TypeScript daemon code touched by this PR, (2) no new regression tests are added for any of the five bugs, violating the repo's explicit guardrail loop requirement, and (3) the Bug 5 hasMore fix introduces a false-negative edge case.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All findings are directly provable from the diff and the full file contents provided. The LIKE injection in buildFilterClause is visible at the tags branch in memory-search.ts (same file as Bug 3 fix). The missing regression test requirement is explicit in AGENTS.md. The hasMore false-negative is derivable from the arithmetic in the changed umap-projection.ts lines 418-422. The double-markDeprecatedVersions is confirmed by tracing discoverAnthropicModels and discoverOpenRouterModels return paths in the provided model-registry.ts full content.


- packages/daemon/src/umap-projection.ts:310 - Same LIKE metacharacter issue in `buildProjectionWhere`: `const pattern = \`%${query}%\`` passes the raw user query into six LIKE predicates (content, tags, who, type, source_type, source_id) without escaping metacharacters. A query of `%` matches every row regardless of filter intent. This function is in the same file as the Bug 5 change and the vulnerability class is identical to what Bug 4 claims to fix.

packages/daemon/src/pipeline/model-registry.ts
```diff
@@ -486,14 +486,14 @@ async function _refreshRegistryInner(
 	}
 
 	if (anthropicModels.length > 0) {
```
PR-Reviewer-Ant, line 488: No regression test covers Bug 2. The fix ensures `markDeprecatedVersions` is applied to anthropic/openrouter results in `_refreshRegistryInner`, but there is no test verifying that `getModelsByProvider()` actually filters deprecated models for those providers after a registry refresh. The sqmd-review section of the PR description explicitly acknowledges tests were added only for bugs 3/4/5. AGENTS.md requires 'Every bug fix needs a regression test.' A test that seeds two same-family anthropic model entries with different versions and asserts the older one is absent from `getModelsByProvider()` output would satisfy this requirement.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b1a806ec`

[Automated Review — pr-reviewer bot] Five correctness fixes are all logically sound. Two issues warrant attention: missing regression tests for Bugs 1 and 2 (an explicit AGENTS.md guardrail), and a duplicated escapeLike helper within the same package.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All five bug fixes are directly readable from the diff with no ambiguity. The missing-test gap for Bugs 1 and 2 is explicit in the sqmd-review section ('regression tests added for bugs 3/4/5'). The escapeLike duplication is visible at memory-search.ts:23-25 and umap-projection.ts:314-316, both under packages/daemon/src/. The hasMore/total inconsistency is verifiable from the ProjectionQueryResult interface and loadProjectionRows logic in the changed file.

packages/adapters/openclaw/src/index.ts
```diff
@@ -874,11 +874,15 @@ function textResult(text: string, details?: Record<string, unknown>): OpenClawTo
 const SESSIONLESS_DEDUPE_MS = 1_000;
 
 function cleanupTimedMap(map: Map<string, number>, now: number, ttlMs = SESSIONLESS_DEDUPE_MS): void {
```
PR-Reviewer-Ant: No regression test covers Bug 1. The Map-mutation-during-iteration fix is correct, but the guardrail loop (AGENTS.md: 'A bug fix is not complete until it adds at least one durable prevention mechanism') requires at least a unit test. A test that populates `cleanupTimedMap` with several entries, advances `now` past TTL for a subset, calls the function, and asserts all expected keys were deleted would close this gap and serve as a canary if the implementation regresses.

packages/daemon/src/umap-projection.ts
```diff
@@ -311,6 +311,10 @@ function normalizeFilterValues(values: readonly string[] | undefined): string[]
 	return [...new Set(normalized)];
 }
 
+function escapeLike(text: string): string {
```
PR-Reviewer-Ant: `escapeLike` is defined here and identically in `packages/daemon/src/memory-search.ts:23`. Both files live in `@signet/daemon`. AGENTS.md explicitly flags duplicate logic as a smell ('Duplicate logic is a smell. Extract shared code where it clarifies.'). Extract to a shared internal utility (e.g. `packages/daemon/src/sql-utils.ts`) and import from both call sites.

packages/daemon/src/umap-projection.ts
```diff
@@ -418,9 +422,12 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
 	const rawRows = db.prepare(sql).all(...rowParams) as Record<string, unknown>[];
 	const rows = rawRows.map(toEmbeddingRow).filter((row): row is EmbeddingRow => row !== null);
 	const limit = requestedLimit ?? Math.max(0, total - offset);
-	const hasMore = offset + rows.length < total;
```
PR-Reviewer-Ant, line 421: The `total` field in `ProjectionQueryResult` is still the SQL `COUNT(*)` value (rows matching the WHERE clause before `toEmbeddingRow` null-filtering). After this fix, `hasMore` is computed via over-fetch and is correct, but `total` can now disagree with `offset + count` in ways that break client-side page-count math when `toEmbeddingRow` discards rows. For example: 5 SQL rows, 2 discarded by null-vector filter, limit=3 → `total=5`, `count=3`, `hasMore=false`, but `offset + count = 3 < total = 5`. Not introduced by this PR, but the hasMore fix makes the inconsistency more visible. Recommend also fixing `total` to reflect post-filter row count, or documenting that `total` is a pre-filter estimate.

packages/daemon/src/memory-search.test.ts
```diff
@@ -5,6 +5,7 @@ import { join } from "node:path";
 import { closeDbAccessor, getDbAccessor, initDbAccessor } from "./db-accessor";
```
PR-Reviewer-Ant, line 5: `buildAgentScopeClause` and `hybridRecall` are imported from `./memory-search` in two separate `import` statements. These should be merged into one (`import { hybridRecall, buildAgentScopeClause } from './memory-search'`). AGENTS.md requires lint to pass before review; Biome will flag this as a duplicate import.

packages/daemon/src/umap-projection.test.ts
```diff
@@ -0,0 +1,92 @@
...
+				id INTEGER PRIMARY KEY AUTOINCREMENT,
+				source_id TEXT NOT NULL REFERENCES memories(id),
+				source_type TEXT NOT NULL,
+				vector BLOB NOT NULL,
+				dimensions INTEGER NOT NULL,
+				created_at TEXT NOT NULL DEFAULT (datetime('now')));
+			CREATE INDEX idx_embeddings_source ON embeddings(source_id, source_type);
+		`);
+
+		const fakeBlob = new Uint8Array(8);
```
PR-Reviewer-Ant: Inconsistent indentation: `db.prepare(` is over-indented relative to the surrounding block. Biome would flag this. Minor, but AGENTS.md requires lint to pass locally before opening a PR.

packages/daemon/src/umap-projection.test.ts
```diff
@@ -0,0 +1,92 @@
...
+		dir = mkdtempSync(join(tmpdir(), "signet-umap-test-"));
+		db = new Database(join(dir, "test.db"));
+		db.exec(`
+			CREATE TABLE memories (id TEXT PRIMARY KEY, content TEXT, who TEXT, importance REAL,
+				type TEXT, tags TEXT, pinned INTEGER, source_type TEXT, source_id TEXT,
+				created_at TEXT, updated_at TEXT);
+			CREATE TABLE embeddings (
+				id INTEGER PRIMARY KEY AUTOINCREMENT,
+				source_id TEXT NOT NULL REFERENCES memories(id),
+				source_type TEXT NOT NULL,
```
PR-Reviewer-Ant: `db` and `dir` are initialized inside each `it` block but declared outside. If an assertion throws before `db.close()` / `rmSync()`, the temp database and directory are leaked for the process lifetime. Consider wrapping setup/teardown in `beforeEach`/`afterEach` with a `try/finally`, or move setup+teardown entirely inside each `it` and use a `try/finally` guard.

packages/daemon/src/umap-projection.ts
```diff
@@ -409,18 +411,34 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
 
-	return { rows, total, offset, limit, hasMore };
+	let effectiveTotal: number;
+	if (requestedLimit !== null && rawRows.length !== rows.length) {
+		const allRaw = db
+			.prepare(`${EMBEDDINGS_SELECT_SQL} ${EMBEDDINGS_FROM_SQL}${clause}`)
+			.all(...params) as Record<string, unknown>[];
+		effectiveTotal = allRaw.map(toEmbeddingRow).filter((r): r is EmbeddingRow => r !== null).length;
+	} else if (requestedLimit !== null) {
+		effectiveTotal = total;
```
PR-Reviewer-Ant: Residual hasMore false-negative when toEmbeddingRow discards rows AND more pages exist.

The over-fetch-by-1 sentinel works only when rawRows.length === rows.length (no discards). When rawRows.length !== rows.length, the code correctly runs a full-table scan to compute effectiveTotal, but then the hasMore expression still uses the sentinel check (`rows.length > requestedLimit`) rather than effectiveTotal.

Concrete counterexample: requestedLimit=3, DB contains 10 valid rows with 1 invalid row interleaved near the start. SQL fetches 4 rows (LIMIT 4). After toEmbeddingRow filtering: rows.length=3, rawRows.length=4 → discard branch → effectiveTotal=10 via full scan. But `hasMore = 3 > 3` = false. Result: caller believes this is the last page, even though 7 more rows exist.

Fix: use effectiveTotal for the discard branch:
```typescript
const hasMore = requestedLimit !== null
  ? (rawRows.length !== rows.length
      ? offset + Math.min(rows.length, requestedLimit) < effectiveTotal
      : rows.length > requestedLimit)
  : offset + rows.length < effectiveTotal;
```

packages/daemon/src/umap-projection.test.ts
```diff
@@ -0,0 +1,84 @@
...
+	const db = new Database(join(dir, "test.db"));
+	db.exec(SCHEMA);
+	return { db, dir };
+}
+
+describe("hasMore pagination regression", () => {
+	it("returns hasMore=false when last page is not full (over-fetch by 1)", () => {
+		const { db, dir } = setupDb();
+		try {
+			const fakeBlob = new Uint8Array(8);
```
PR-Reviewer-Ant, line 33: The regression test for Bug 5 (hasMore pagination) only covers the no-discard path — all inserted blobs are valid Uint8Array(8) values and toEmbeddingRow never returns null. The broken code path (discard-present + more pages remain) is therefore not exercised. A test that inserts rows with a corrupted/mismatched blob interleaved among valid rows would catch the residual bug described above.

packages/daemon/src/umap-projection.ts
```diff
@@ -386,6 +387,7 @@ function buildProjectionWhere(filters: ProjectionFilters | undefined): {
 
 interface ProjectionRowsResult {
```
PR-Reviewer-Ant, line 388: The JSDoc comment on effectiveTotal says it "falls back to DB COUNT(*) when no rows are discarded," which is backwards: DB COUNT(*) is used when there are NO discards; the expensive full re-scan is used when there ARE discards. The comment should read: "DB COUNT(*) used when all fetched rows pass JS-side validation; full re-scan used when any row is discarded."

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `77f2e6df`

[Automated Review — pr-reviewer bot] Five correctness fixes are generally sound, but Bug 5's hasMore fix contains a residual correctness error in the discard-plus-more-pages code path, and the regression test does not cover that exact case.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The residual hasMore bug is directly provable by tracing the three code paths in the new loadProjectionRows body (packages/daemon/src/umap-projection.ts lines ~419–443). The test at umap-projection.test.ts lines 33–58 uses only valid blobs, so toEmbeddingRow never discards a row and the broken branch is never exercised. No cross-module context is needed to confirm this.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,47 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
-	const hasMore = offset + rows.length < total;
+	const trimmedRows = requestedLimit !== null ? rows.slice(0, requestedLimit) : rows;
+
+	let effectiveTotal: number;
+	let hasMore: boolean;
+
+	if (requestedLimit !== null) {
+		hasMore = rows.length > requestedLimit;
+		if (rawRows.length !== rows.length) {
+			const allRaw = db
```
PR-Reviewer-Ant: When `toEmbeddingRow` discards any row in a paginated fetch, the code issues a full unbounded scan (`EMBEDDINGS_SELECT_SQL + EMBEDDINGS_FROM_SQL + clause` with no LIMIT) to recompute `effectiveTotal`. For embedding tables with many rows this query will be proportionally expensive and is issued on every affected page — not just the first. If malformed blobs accumulate (e.g. after a migration or schema change), every paginated request could trigger this path and cause latency spikes or memory pressure. Consider whether a COUNT(*) query with a JS-side discard correction, or a separate `COUNT(valid_blobs)` approximation, would provide sufficient accuracy with bounded cost. The current correctness argument is sound; the question is whether the accuracy trade-off justifies unbounded scans at scale.

packages/daemon/src/memory-search.ts
```diff
@@ -19,6 +19,7 @@ import { resolveFocalEntities, setTraversalStatus, traverseKnowledgeGraph } from
 import { type RerankCandidate, noopReranker, rerank } from "./pipeline/reranker";
 import { createEmbeddingReranker } from "./pipeline/reranker-embedding";
 import { createLlmReranker, summarizeRecallWithLlm } from "./pipeline/reranker-llm";
+import { escapeLike } from "./sql-utils";
```
PR-Reviewer-Ant: Import of `./sql-utils` is inserted between two `./pipeline/*` imports. If Biome's import-order rule is active this will fail `bun run lint`. Alphabetically `./pipeline/stop-words` sorts before `./sql-utils` (`p` < `s`), so the import should appear after the last `./pipeline/*` line. Minor, but the PR claims lint passes locally — worth a quick `bun run lint` confirmation.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `018de38d`

Automated bot review (pr-reviewer). This PR fixes five correctness bugs found via structural search: map-mutation-during-iteration, missing deprecation marking on dynamic model lists, null policyGroup causing empty group-scope SQL results, LIKE metacharacter injection, and a hasMore pagination false-negative. All fixes are accompanied by regression tests. One performance concern in the discard path of the hasMore fix is worth discussing before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All five bug fixes and their tests are directly visible in the diff. The performance concern in umap-projection.ts is provable from the changed lines alone: EMBEDDINGS_SELECT_SQL + EMBEDDINGS_FROM_SQL + clause is assembled without LIMIT/OFFSET on line ~425 of the new file. No cross-module runtime behavior is needed to evaluate it.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,44 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
-	const hasMore = offset + rows.length < total;
+	const trimmedRows = requestedLimit !== null ? rows.slice(0, requestedLimit) : rows;
+
+	let effectiveTotal: number;
+	let hasMore: boolean;
+
+	if (requestedLimit !== null) {
+		hasMore = rows.length > requestedLimit;
+		const totalRow = db.prepare(`SELECT COUNT(*) AS count ${EMBEDDINGS_FROM_SQL}${clause}`).get(...params) as
+			| { count?: number }
```
PR-Reviewer-Ant, line 427: The re-count path fetches the raw `e.vector` BLOB column for every row matching the filter clause — no LIMIT, no OFFSET. Each embedding is a float32 array (e.g. 1 536 dims × 4 bytes = ~6 KB). For a table with tens of thousands of embeddings this pulls potentially hundreds of megabytes into the JS heap just to count valid blobs.

A cheaper alternative that avoids loading BLOB data: `SELECT COUNT(*) FROM embeddings e JOIN memories m ON m.id = e.source_id WHERE typeof(e.vector) = 'blob' AND <clause>`. SQLite's `typeof()` is evaluated without materialising the value, so this is a pure index/metadata scan. Because the re-count path only fires when `!hasMore` and rows were discarded, it will execute on real user traffic (any corrupted embedding row triggers it on every page load until the bad row is purged).

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,44 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
 		sql += " LIMIT ? OFFSET ?";
-		rowParams.push(requestedLimit, offset);
+		rowParams.push(requestedLimit + 1, offset);
 	} else if (offset > 0) {
 		sql += " LIMIT -1 OFFSET ?";
 		rowParams.push(offset);
 	}
 
 	const rawRows = db.prepare(sql).all(...rowParams) as Record<string, unknown>[];
 	const rows = rawRows.map(toEmbeddingRow).filter((row): row is EmbeddingRow => row !== null);
```
PR-Reviewer-Ant, line 416: `effectiveTotal` is always set from `COUNT(*) AS count` (which counts all DB rows regardless of blob validity) when the sentinel triggers `hasMore = true`. If there are invalid-vector rows elsewhere in the table that are not in the current over-fetched window, COUNT(*) will include them, making `effectiveTotal` slightly larger than the number of rows that will ever be renderable. The mismatch is surfaced to callers as the `total` field on the result (used for pagination UI). The `hasMore` flag itself is correct in this scenario, so this is a display fidelity issue rather than a logic error, but it means the total can read higher than actual usable rows.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `27dfbca2`

Automated bot review (pr-reviewer). Five targeted bug fixes — map-mutation-during-iteration, missing deprecation marking, null policyGroup SQL silent-empty, LIKE metacharacter injection (TS + Python), and UMAP hasMore false-negative. All five fixes are correct and accompanied by regression tests. Two follow-on issues found in the UMAP fix.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All five bug fixes are directly verifiable from the diff. The UMAP performance concern is provable from the SELECT e.vector query at line ~425 of the patched umap-projection.ts, which fetches full BLOB data without a LIMIT. The effectiveTotal over-count is provable from the COUNT(*) path executing unconditionally even when hasMore is already true, combined with COUNT including invalid-vector rows.

packages/daemon/src/memory-search.ts
```diff
@@ -12,14 +12,15 @@ import { getDbAccessor } from "./db-accessor";
 import { getLlmProvider } from "./llm";
 import { logger } from "./logger";
```
PR-Reviewer-Ant, line 13: Import ordering may violate Biome's sort rules. The two new imports appear as:

```
import { escapeLike } from "./sql-utils";      // 's'
import { FTS_STOP } from "./pipeline/stop-words"; // 'p'
import { constructContextBlocks } from "./pipeline/context-construction"; // 'p'
```

Biome sorts relative paths alphabetically; `pipeline/*` ('p') should sort before `sql-utils` ('s'). The PR description says R5 fixed import ordering for Biome compliance, but the diff still shows `sql-utils` ahead of `pipeline/stop-words`. If `bun run lint` is clean this is a non-issue, but it's worth confirming the lint pass actually covers this file.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,46 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
-	const limit = requestedLimit ?? Math.max(0, total - offset);
-	const hasMore = offset + rows.length < total;
+	const trimmedRows = requestedLimit !== null ? rows.slice(0, requestedLimit) : rows;
+
+	let effectiveTotal: number;
+	let hasMore: boolean;
+
+	if (requestedLimit !== null) {
+		hasMore = rows.length > requestedLimit;
+		const totalRow = db.prepare(`SELECT COUNT(*) AS count ${EMBEDDINGS_FROM_SQL}${clause}`).get(...params) as
```
PR-Reviewer-Ant: When `toEmbeddingRow` discards rows, the code now runs up to three SQL queries per paginated request: (1) the row fetch, (2) a `COUNT(*)` for `effectiveTotal`, then (3) a second `COUNT(*) … AND typeof(e.vector) = 'blob'` that overwrites `effectiveTotal`. In practice the third query only triggers when invalid blob rows exist, but the second COUNT (lines 422–426) is always executed for every paginated request regardless of whether discards occurred, adding one unconditional round-trip compared to the original single count+fetch pattern. This is unlikely to matter at current data sizes, but is worth documenting or gating behind the discard condition if query budget is a concern.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fe6f1a25`

[Automated Review — pr-reviewer bot] PR #503 fixes five correctness bugs (map-mutation-during-iteration, missing deprecation pass, null policyGroup scope fallback, LIKE metacharacter injection in TS and Python, hasMore pagination false-negative). All fixes are logically sound and regression tests cover the targeted paths. Two minor issues flagged below.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All five bug fixes are directly provable from the diff: the Map iteration mutation fix is straightforward; the policyGroup null guard is explicit at memory-search.ts:102; the LIKE injection fix and escapeLike deduplication are self-contained; the umap hasMore logic is fully traceable through the new loadProjectionRows body. The test assertions were manually traced against the implementation and pass. The two flagged items are grounded in specific lines in the diff.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,49 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
+		hasMore = rows.length > requestedLimit;
+
+		if (rawRows.length !== rows.length) {
+			const validCountRow = db
+				.prepare(`SELECT COUNT(*) AS count ${EMBEDDINGS_FROM_SQL}${clause} AND typeof(e.vector) = 'blob'`)
+				.get(...params) as { count?: number } | undefined;
+			effectiveTotal =
+				validCountRow !== undefined && typeof validCountRow.count === "number" ? validCountRow.count : 0;
+			if (!hasMore) {
+				hasMore = offset + trimmedRows.length < effectiveTotal;
```
PR-Reviewer-Ant: In the no-discard path (`rawRows.length === rows.length`), `effectiveTotal` is computed from `COUNT(*)` (unchanged from pre-fix). If non-blob vector rows exist on later pages that the current fetch window didn't reach, `total` in the API response will be transiently inflated. `hasMore` is now accurate via the over-fetch sentinel, so pagination correctness is fine. Only the `total` field is affected. This was already true before this PR, so it is not a regression, but it's the one invariant the PR description implies is fully fixed ('accurate `effectiveTotal`') without fully closing. Worth a follow-up to also apply `typeof(e.vector) = 'blob'` in the no-discard COUNT, or document the carve-out.

packages/daemon/src/umap-projection.test.ts
```diff
@@ -0,0 +1,123 @@
...
+					db.prepare(
+						"INSERT INTO embeddings (source_id, source_type, vector, dimensions, created_at) VALUES (?, 'memory', ?, ?, ?)",
+					).run(`mem-${i}`, fakeBlob, 4, new Date(baseTime.getTime() + i * 1000).toISOString());
+				}
+			}
+
+			const page1 = computeProjectionForQuery(db, 2, { limit: 3, offset: 0 });
+			expect(page1.total).toBe(5);
+			expect(page1.count).toBe(3);
+			expect(page1.hasMore).toBe(true);
```
PR-Reviewer-Ant, line 151: The discard-path test only exercises `offset: 0`. The `offset + trimmedRows.length < effectiveTotal` branch in `loadProjectionRows` (the path that recovers `hasMore` when the sentinel doesn't fire but discards occurred) is only reachable from a non-zero offset. Adding a page-2 fetch (e.g., `{ limit: 2, offset: 2 }`) to the existing discard scenario would confirm there is no off-by-one in offset accounting when `effectiveTotal` comes from the blob-filtered re-count.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `24470fa9`

[Automated Bot Review] Five correctness fixes reviewed. All primary bug fixes are correct and well-tested. Two minor issues identified: a transient `total` field inflation risk in the umap pagination no-discard path, and a missing offset>0 test case for the discard path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All five fixes are directly verifiable from the diff. The map-mutation fix, deprecation call site fix, null policyGroup guard, and LIKE escape chain are all provably correct from the changed lines. The umap pagination logic is fully inlined in the diff and the test fixtures are explicit. The one nuance (effectiveTotal in the no-discard path) is directly traceable to the new COUNT(*) branch at umap-projection.ts lines ~428–434 and is measurable against the pre-fix behavior without needing runtime traces.

packages/daemon/src/memory-search.ts
```diff
@@ -149,8 +150,8 @@ function buildFilterClause(params: RecallParams): FilterClause {
 			.split(",")
 			.map((s) => s.trim())
 			.filter(Boolean)) {
-			parts.push("m.tags LIKE ?");
-			args.push(`%${t}%`);
+			parts.push("m.tags LIKE ? ESCAPE '\\'");
+			args.push(`%${escapeLike(t)}%`);
```
PR-Reviewer-Ant, line 154: The `if (params.who) {` block immediately follows the fixed `tags` LIKE clause but its implementation is not shown in the diff. If the `who` filter uses a `LIKE ?` query with user-supplied input (a common pattern for author/who search), it is also vulnerable to LIKE metacharacter injection and would need the same `ESCAPE '\\'` + `escapeLike()` treatment applied here. The PR description calls out `buildFilterClause` and `buildProjectionWhere` as fixed, but does not mention the `who` sub-clause explicitly. Please confirm the `who` branch is either not LIKE-based or was also fixed.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7961183b`

[Automated Review — pr-reviewer bot] This PR fixes five correctness bugs (map-mutation-during-iteration, missing deprecation marking, null policyGroup scope fallback, LIKE metacharacter injection in Python/TS, and hasMore pagination false-negatives). The fixes are generally sound, but one potentially incomplete LIKE-injection fix and one ambiguous discard-coverage assumption are worth flagging before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff is self-contained for Bugs 1–3 and fully verifiable. For Bug 4 (LIKE injection), the full body of buildFilterClause is not shown past the `if (params.who) {` line, so I cannot confirm whether the `who` filter also uses LIKE and was also fixed. For Bug 5, the full `toEmbeddingRow` implementation is absent, so the assumption that `typeof(e.vector) = 'blob'` captures all discard reasons cannot be verified from this diff.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,45 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
+	const trimmedRows = requestedLimit !== null ? rows.slice(0, requestedLimit) : rows;
+
+	let effectiveTotal: number;
+	let hasMore: boolean;
+
+	if (requestedLimit !== null) {
+		hasMore = rows.length > requestedLimit;
+
+		const countClause = rawRows.length !== rows.length
+			? `${EMBEDDINGS_FROM_SQL}${clause} AND typeof(e.vector) = 'blob'`
```
PR-Reviewer-Ant: The fallback COUNT clause adds `AND typeof(e.vector) = 'blob'` to filter out rows that `toEmbeddingRow` would discard. This assumes the sole discard reason is a non-blob vector. If `toEmbeddingRow` also discards rows for other reasons (e.g. null `source_id`, dimension mismatch, unexpected `source_type`), the re-count will overcount valid rows, making `effectiveTotal` too high and potentially producing a spurious `hasMore = true` for the last real page. Without the `toEmbeddingRow` implementation this cannot be confirmed from the diff alone.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,45 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
-	const total = totalRow !== undefined && typeof totalRow.count === "number" ? totalRow.count : 0;
-
 	let sql = `${EMBEDDINGS_SELECT_SQL} ${EMBEDDINGS_FROM_SQL}${clause} ORDER BY m.created_at DESC`;
 	const rowParams: unknown[] = [...params];
 	if (requestedLimit !== null) {
 		sql += " LIMIT ? OFFSET ?";
-		rowParams.push(requestedLimit, offset);
+		rowParams.push(requestedLimit + 1, offset);
 	} else if (offset > 0) {
 		sql += " LIMIT -1 OFFSET ?";
```
PR-Reviewer-Ant, line 411: The new code unconditionally issues a `SELECT COUNT(*) AS count` query for every paginated request, even when no rows are discarded (`rawRows.length === rows.length`) and the over-fetch-by-1 sentinel is already sufficient for the `hasMore` determination. `effectiveTotal` is still needed in the return value, so the COUNT cannot be eliminated entirely, but the original strategy—run COUNT first, then fetch—would avoid the extra round-trip when rows are a known clean fetch. Minor performance regression on the hot path; not a correctness issue.

packages/adapters/openclaw/src/index.ts
```diff
@@ -873,12 +873,16 @@ function textResult(text: string, details?: Record<string, unknown>): OpenClawTo
 // have a stable messageCount to key on).
 const SESSIONLESS_DEDUPE_MS = 1_000;
 
-function cleanupTimedMap(map: Map<string, number>, now: number, ttlMs = SESSIONLESS_DEDUPE_MS): void {
+export function cleanupTimedMap(map: Map<string, number>, now: number, ttlMs = SESSIONLESS_DEDUPE_MS): void {
+	const expired: string[] = [];
```
PR-Reviewer-Ant, line 877: The PR description frames the `cleanupTimedMap` change as a bug fix for 'map-mutation-during-iteration'. In JavaScript, deleting the *current* entry from a Map while iterating with `for...of` is defined behavior: the iterator has already yielded that key, so the deletion is safe and future keys are unaffected. The original code had no correctness defect. The collect-then-delete pattern is a worthwhile defensive improvement, but the commit message (`fix:`) and PR description should reflect that this is a defensive hardening rather than a bug fix to avoid misleading the changelog and version bump.

packages/cli/templates/memory/scripts/memory.py
```diff
@@ -513,12 +513,13 @@ def query_memories_fts_only(search: str, limit: int = 20):
     except sqlite3.OperationalError:
         pass
 
+    safe_search = search.lower().replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
```
PR-Reviewer-Ant: The LIKE-injection fix is correct (backslash escaped first, then `%` and `_`; ESCAPE clause present), but the repo convention requires a regression test for every bug fix and this Python path has none. Since `query_memories_fts_only` is reachable from the CLI `query` subcommand, a short pytest (or bun-driven subprocess test) that passes a `%` or `_` literal and asserts it does NOT match unrelated rows would satisfy the requirement and guard against future regressions.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `da897d0c`

Automated pr-reviewer bot — Five-bug correctness fix PR. All TypeScript-side fixes are sound and well-tested. One gap: the Python LIKE-injection fix has no regression test, contrary to repo conventions. No blocking correctness or security issues found.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All five fix sites are visible in the diff with their full surrounding context. The SQL parameter binding, ESCAPE clause syntax, sentinel-based hasMore logic, and null-guard change are directly verifiable from the changed lines without needing runtime reproduction. The one gap (Python test) is structural, not ambiguous.

packages/signetai/templates/memory/scripts/memory.py
```diff
@@ -513,12 +513,13 @@ def query_memories_fts_only(search: str, limit: int = 20):
     except sqlite3.OperationalError:
         pass
 
+    safe_search = search.lower().replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
```
PR-Reviewer-Ant: Same fix and same test gap as `packages/cli/templates/memory/scripts/memory.py` — both template copies need coverage.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,46 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
 		rowParams.push(offset);
 	}
```
PR-Reviewer-Ant, line 414: The COUNT query (`SELECT COUNT(*) … AND typeof(e.vector) = 'blob'`) is now executed unconditionally on every paginated request, even when no JS-side discards occur (`rawRows.length === rows.length`). This is a per-page extra full-index scan. For small to medium datasets this is acceptable, but callers paginating through large embedding stores will double the DB round-trips. The PR description acknowledges this (R7 round-up). Worth noting for future optimisation: the COUNT could be skipped when `rows.length === rawRows.length` (no discards) and the sentinel already gave a definitive `hasMore` answer.

PR-Reviewer-Ant, line 414: The COUNT query (`SELECT COUNT(*) AS count ${EMBEDDINGS_FROM_SQL}${clause} AND typeof(e.vector) = 'blob'`) now runs unconditionally for every call where `requestedLimit` is set — even when `rawRows.length === rows.length` (no discards) and `hasMore` is already known from the over-fetch sentinel. For large tables this adds a full-index scan per paginated request. The PR description (R6/R7) acknowledged this as a minor perf trade-off and chose not to short-circuit it; that decision is reasonable, but worth confirming it's acceptable at expected corpus sizes.

packages/cli/templates/memory/scripts/memory.py
```diff
@@ -736,5 +803,45 @@ def main():
         migrate_markdown()
```
PR-Reviewer-Ant, line 803: `_test_like_escaping()` is only invoked when the script is called directly with `--test-like-escaping`. The AGENTS.md workspace test command is `bun test`, which only covers TypeScript tests. Unless CI is configured to also invoke `python memory.py --test-like-escaping` (in both template copies), these regression tests for Bug 4 will never run automatically. If either Python copy drifts, the tests won't catch it. Consider moving the assertion logic into an existing pytest/unittest suite, or documenting the CI invocation needed to keep these alive.

packages/daemon/src/pipeline/model-registry.test.ts
```diff
@@ -0,0 +1,30 @@
+import { describe, expect, it } from "bun:test";
+import { markDeprecatedVersions } from "./model-registry";
+
+describe("markDeprecatedVersions", () => {
+	it("marks older same-family models as deprecated", () => {
+		const entries = [
+			{ id: "claude-opus-4.0", provider: "anthropic", label: "Claude Opus 4.0", tier: "high", deprecated: false },
+			{ id: "claude-opus-4.6", provider: "anthropic", label: "Claude Opus 4.6", tier: "high", deprecated: false },
```
PR-Reviewer-Ant, line 8: Several lines in this new test file significantly exceed the 120-char hard limit from AGENTS.md (e.g., the inline object literals on lines 8–9, 15–16, 22, 28). These should be broken across lines to stay within the hard limit. Does not affect correctness.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `2065ee34`

[Automated Review — pr-reviewer bot] PR #503 fixes five correctness bugs (map-mutation guard, missing deprecation wrapping, null policyGroup fallback, LIKE injection, and hasMore pagination). All five fixes are logically sound. Three issues worth noting before merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The five core bug fixes are directly provable from the diff. One medium-confidence concern (COUNT query always running) is provable from the changed lines in umap-projection.ts but the full EMBEDDINGS_FROM_SQL constant and EMBEDDINGS_FROM_CLAUSE join form are not in the provided context, so I cannot fully rule out a syntax edge case when clause is empty. The Python CI coverage gap is directly provable from the diff and the AGENTS.md workspace commands which show only 'bun test' in the test pipeline.

packages/cli/templates/memory/scripts/memory.py
PR-Reviewer-Ant, line 803: `_test_like_escaping()` is gated behind a `--test-like-escaping` CLI flag and is not discoverable by `bun test` or any standard Python test runner, so it will not execute in CI. The PR checklist claims regression tests were added for Bug 4, but these assertions only run manually. Consider moving this into a proper test file (e.g. `memory_test.py`) or a pytest parametrized test so it is automatically enforced. The same applies to the identical copy in `packages/signetai/templates/memory/scripts/memory.py`.

packages/daemon/src/pipeline/model-registry.ts
```diff
@@ -486,14 +486,14 @@ async function _refreshRegistryInner(
...
 		// Claude Code CLI accepts (shorthand aliases, not dated Anthropic IDs).
 		// The seeded KNOWN_MODELS["claude-code"] entries use the correct shorthands.
 	}
 
 	if (openRouterModels.length > 0) {
-		state.models.set("openrouter", openRouterModels);
+		state.models.set("openrouter", markDeprecatedVersions(openRouterModels));
 	}
 
 	state.lastRefreshAt = Date.now();
```
PR-Reviewer-Ant, line 499: `markDeprecatedVersions` is now applied to live-fetched OpenRouter models, but the new tests in `model-registry.test.ts` only exercise anthropic-style IDs (`claude-opus-4.0`, `claude-opus-4.6`). OpenRouter model IDs typically use `provider/model-name` format (e.g. `anthropic/claude-3-opus`). The family-extraction logic inside `markDeprecatedVersions` (not changed in this diff) was originally written and tested only against seeded `KNOWN_MODELS`—which use shorthand IDs—not against live OpenRouter payloads. If the family prefix is derived from the full ID string without stripping the `provider/` prefix, cross-provider models with similar suffixes could be incorrectly grouped, causing a current model to be marked deprecated. Please add a test covering an OpenRouter-style ID pair to confirm the function handles that format correctly.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,46 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
 
 	const rawRows = db.prepare(sql).all(...rowParams) as Record<string, unknown>[];
 	const rows = rawRows.map(toEmbeddingRow).filter((row): row is EmbeddingRow => row !== null);
-	const limit = requestedLimit ?? Math.max(0, total - offset);
-	const hasMore = offset + rows.length < total;
+	const trimmedRows = requestedLimit !== null ? rows.slice(0, requestedLimit) : rows;
```
PR-Reviewer-Ant, line 417: The `SELECT COUNT(*) … typeof(e.vector) = 'blob'` query now runs unconditionally on every paginated request, even when no rows are discarded by `toEmbeddingRow` (the common case). The over-fetch-by-1 sentinel is sufficient for `hasMore` when there are no discards; the COUNT is only strictly needed when `rawRows.length !== rows.length`. The PR description acknowledges this as a future optimisation. For completeness: gating the COUNT behind `if (rawRows.length !== rows.length)` would avoid the extra full-table scan on every normal page fetch without affecting correctness.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `122f2748`

[Automated Review — pr-reviewer bot] PR #503 fixes five correctness bugs with regression tests. No blocking issues found; two warnings and one nitpick are flagged below.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The LIKE-escaping, pagination, and scope-clause fixes are fully provable from the diff. Two concerns require context outside the diff: (1) the full body of `markDeprecatedVersions` (only the export keyword changed; family-extraction logic is not shown) is needed to verify correctness for OpenRouter ID formats, and (2) the full `EMBEDDINGS_FROM_SQL` constant is needed to confirm `AND typeof(e.vector) = 'blob'` appended to `countClause` produces valid SQL when `clause` is empty.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b4ab0786`

```pr-review-json
{
  "summary": "[Automated Review — pr-reviewer bot] Five-bug correctness fix PR. All five fixes are sound and well-tested. One behavioral trade-off in the UMAP pagination change deserves a human second opinion before merge; all other findings are informational.",
  "verdict": "comment",
  "confidence": {
    "level": "high",
    "justification": "All five bug fixes are fully traceable from the diff. The LIKE-escaping logic can be verified end-to-end through escapeLike → SQL ESCAPE clause → SQLite semantics. The hasMore sentinel logic is verifiable by tracing loadProjectionRows line-by-line. The policyGroup null-guard and markDeprecatedVersions wiring are single-line changes with clear before/after. The one finding (total undercount) is a deliberate trade-off explicitly acknowledged in R9 of the PR description, visible in the changed lines."
  },
  "comments": [
    {
      "file": "packages/daemon/src/umap-projection.ts",
      "line": 427,
      "body": "**`total` field is now an estimate in the common no-discard path, not a DB count.**\n\nPrevious behavior: always ran `SELECT COUNT(*) AS count` so `total` was always a precise DB result (though it included invalid-embedding rows, which was itself a latent bug).\n\nNew behavior (`effectiveTotal = offset + rows.length` when `rawRows.length === rows.length`): on any page that isn't the last one and has no discards, `total` equals `offset + requestedLimit + 1` — one more than the page end — rather than the true count. In a 100-row collection with limit=10 and no discards, page 1 reports `total=11`, page 2 reports `total=21`, etc., correcting only on the final page.\n\nThis was acknowledged as acceptable in R9 of the PR description, and `hasMore` is always correct, so functional pagination is not broken. The concern is whether any downstream consumer of the UMAP projection API uses `total` to render a stable count label (e.g., "5000 memories") or to compute page count. If so, the display will oscillate as pages load. The updated JSDoc on line 391 documents the new contract, which helps.\n\nHuman reviewer: please confirm that no dashboard component currently relies on `total` being a stable DB-accurate value across pages.",
      "severity": "warning"
    },
    {
      "file": "packages/cli/templates/memory/scripts/test_like_escaping.py",
      "line": 1,
      "body": "These pytest files live under `packages/cli/templates/memory/scripts/` and `packages/signetai/templates/memory/scripts/`, which are install-time template directories, not the normal test discovery paths. Whether CI picks them up depends on the pytest invocation root. If the CI command is something like `pytest packages/daemon` or only covers non-template source, these tests will silently not run.\n\nVerify that the CI pytest step includes `packages/*/templates/memory/scripts/` in its discovery path, or that the `test_like_escaping.py` files are additionally copied/symlinked somewhere that is already discovered.",
      "severity": "nitpick"
    },
    {
      "file": "packages/daemon/src/umap-projection.ts",
      "line": 414,
      "body": "The discard-detection path runs `COUNT(*) WHERE typeof(e.vector) = 'blob'` unconditionally whenever any row in the current fetch window is discarded (`rawRows.length !== rows.length`). On the last page of a collection with persistent invalid embeddings this fires on every page load. The PR description acknowledges this (R10: \"gated behind discard check\") — it is gated — but the cost is per-page rather than amortised. For most deployments this is fine; noting it here in case the embedding table grows large.",
      "severity": "nitpick"
    }
  ]
}
```

[stderr]

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,46 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
+
+	let effectiveTotal: number;
+	let hasMore: boolean;
+
```
PR-Reviewer-Ant, line 421: The COUNT query (`SELECT COUNT(*) AS count … AND typeof(e.vector) = 'blob'`) now runs unconditionally on every paginated call to `loadProjectionRows`, regardless of whether any discard actually occurred. This means every UMAP dashboard page load issues a full JOIN scan against the embeddings and memories tables even in the common no-discard case. The PR description (R11) justifies this as necessary for a stable `total` for UmapPanel/EmbeddingsTab display — that rationale is sound — but the perf cost is O(full table) per page. Consider adding an index on `typeof(e.vector)` or gating the COUNT behind a quick `rawRows.length !== rows.length` check while still returning `effectiveTotal` from a cached/approximate value for the dashboard when no discards are present. This is not a correctness issue, but worth tracking as the memory store grows.

conftest.py
```diff
@@ -0,0 +1,5 @@
+import sys
+from pathlib import Path
+
+packages = Path(__file__).resolve().parent / "packages"
```
PR-Reviewer-Ant: The root-level `conftest.py` adds `packages/` to `sys.path` so pytest can import from that subtree, but the added test files (`packages/cli/templates/memory/tests/test_like_escaping.py` and `packages/signetai/templates/memory/tests/test_like_escaping.py`) are entirely self-contained — they only import `sqlite3` from the stdlib and never import from `packages/`. The sys.path injection is therefore a no-op for these tests. More critically, pytest's recursive test discovery requires the intermediate directories to either be Python packages (`__init__.py` present) or appear in an explicit `testpaths` config. If neither `packages/cli/templates/memory/tests/` nor `packages/signetai/templates/memory/tests/` has an `__init__.py`, and the CI invocation does not pass an explicit path, these regression tests may silently not run. Verify that `bun test` or the CI pytest invocation actually discovers and executes these files.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `b37849d6`

[Automated Review — pr-reviewer] Five-bug fix PR with regression tests. No blocking correctness or security issues found. Two warnings flagged: unconditional COUNT per paginated UMAP request (acknowledged as intentional but worth tracking), and a potential CI discoverability gap for the Python pytest files added in deeply-nested template directories.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - All five bug fixes are verifiable from the diff. The LIKE-escape logic, sentinel-based hasMore, and null-policyGroup fallback are provably correct. The performance concern is visible in loadProjectionRows (umap-projection.ts lines ~414-430). The CI discoverability gap depends on the repo's pytest.ini / pyproject.toml which are not in the diff — I can't confirm whether testpaths includes the deeply-nested template test dirs.

packages/daemon/src/umap-projection.ts
```diff
@@ -400,27 +402,46 @@ function loadProjectionRows(db: ReadDb, query: ProjectionQuery): ProjectionRowsR
...
+		if (rawRows.length !== rows.length) {
+			const countClause = `${EMBEDDINGS_FROM_SQL}${clause} AND typeof(e.vector) = 'blob'`;
+			const totalRow = db.prepare(`SELECT COUNT(*) AS count ${countClause}`).get(...params) as
+				| { count?: number }
+				| undefined;
+			effectiveTotal = totalRow !== undefined && typeof totalRow.count === "number" ? totalRow.count : 0;
+			if (!hasMore) {
+				hasMore = offset + trimmedRows.length < effectiveTotal;
+			}
```
PR-Reviewer-Ant: Blocking: `total` is now approximate/undercounted in the no-discard path (`effectiveTotal = offset + rows.length - (hasMore ? 1 : 0)`). This is a behavior regression from exact totals and can produce incorrect pagination metadata (unstable/too-small totals) for clients that display total result counts or compute page bounds from `total`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `aca1b96c`

[Automated Review by pr-reviewer] I found one blocking correctness regression in pagination metadata.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new no-discard branch in loadProjectionRows now derives `total` from `offset + rows.length - (hasMore ? 1 : 0)` instead of a COUNT query. On any non-terminal page this undercounts badly (e.g., first page with limit=20 can report total=20 even when hundreds match), which directly changes API semantics and can break pagination/UI logic relying on stable totals.

NicholaiVogel:
nice sweep overall. i think most of these fixes make sense, and i like that you added regression coverage along with them.

i did notice one remaining hole though in `umap-projection.ts`.

the free-text `query` path now escapes LIKE metacharacters, but the explicit `tags` filter path still does a raw:

```ts
parts.push("m.tags LIKE ?");
params.push(`%${tag}%`);
```

so `%` and `_` are still treated as wildcards there.

i checked it locally and `filters: { tags: ["%"] }` still matches tagged rows instead of returning zero results, so this looks like the same class of issue that got fixed elsewhere in the PR, just still present in this branch.

i think this path should use the same `escapeLike(...)` + `ESCAPE '\\'` treatment too, and probably get a regression test next to the projection query escaping coverage.

PR #504, opened 2026-04-13
fix(hermes): scope Signet memory to named agents

NicholaiVogel:
## Summary

Fixes Hermes Signet integration in three related ways:

- Scopes Hermes Signet memory and identity to the named agent workspace instead of falling back to the generic Hermes identity.
- Gives Signet recall paths enough time to complete on larger memory stores, avoiding false daemon reachability failures when recall takes longer than the generic 5s HTTP timeout.
- Exposes the complete Signet memory write schema to Hermes and MCP callers, including prospective hints, transcript, and structured graph payloads.

## Changes

- Adds Hermes connector support for named-agent scoping through `SIGNET_AGENT_ID` and `SIGNET_AGENT_WORKSPACE`.
- Makes Hermes Agent a first-class `signet setup` harness in interactive setup, non-interactive help text, migration detection, and setup docs.
- Keeps explicit recall requests unscoped by default and exposes `agent_scoped` so callers can opt into named-agent scoped recall.
- Extends the bundled Hermes plugin recall timeout to 30s.
- Uses the 30s recall timeout for per-turn `user-prompt-submit` recall hooks.
- Extends `signet recall` CLI timeout to 30s for `/api/memory/recall`.
- Keeps Hermes `memory_forget` soft-delete only by not exposing the daemon hard-delete `force` option to model-callable tools.
- Adds `hints`, `transcript`, and `structured` support to the Hermes `memory_store` schema and daemon payload forwarding.
- Aligns the MCP `memory_store` schema and payload forwarding with the daemon remember endpoint for hints, transcript, and structured graph metadata.
- Documents the full Hermes `memory_store` surface.
- Adds regression coverage for Hermes connector scoping, setup harness availability, Hermes install detection, recall timeout behavior, explicit `agent_scoped` recall opt-in, schema completeness, MCP structured memory writes, auth-bearing agent registration, non-404 registration checks, bundled Python client compatibility, malformed recall score filtering, soft-delete-only Hermes forget behavior, uninstall credential cleanup, and legacy structured payload compatibility.

## Type

- [ ] `feat` - new user-facing feature (bumps minor)
- [x] `fix` - bug fix
- [ ] `refactor` - restructure without behavior change
- [ ] `chore` - build, deps, config, docs
- [ ] `perf` - performance improvement
- [ ] `test` - test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

Readiness caveat: repo-wide validator, lint, and typecheck caveats are documented in the Testing section below. Targeted checks for this PR pass.

N/A. No database migrations touched.

- [ ] Migration is idempotent

Compatibility note: these changes are additive to existing Hermes and MCP schemas. Existing memory tool calls remain valid, including legacy structured aspect tuples using `{entity, aspect, value}`. Named agents may provide agent-local identity overrides for the standard identity files, including `MEMORY.md`; if no agent-local file exists, the root workspace file remains the fallback. Hermes model-callable `memory_forget` remains soft-delete only and does not expose hard-delete `force`. Hermes uninstall removes persisted Signet env vars, including `SIGNET_TOKEN`.

## Testing

- [x] `bun test` passes for touched regression suites
- [ ] `bun run typecheck` passes repo-wide
- [ ] `bun run lint` passes repo-wide
- [x] Touched package typechecks pass
- [ ] N/A

Commands run locally:

```bash
bun run build:core
bun test packages/connector-hermes-agent/index.test.ts packages/daemon/src/mcp/tools.test.ts packages/cli/src/commands/memory.test.ts packages/core/src/__tests__/identity.test.ts packages/daemon/src/hooks.test.ts packages/cli/src/features/setup-shared.test.ts
python3 -m py_compile packages/connector-hermes-agent/hermes-plugin/__init__.py packages/connector-hermes-agent/hermes-plugin/client.py
bunx biome check packages/connector-hermes-agent/index.test.ts packages/connector-hermes-agent/src/index.ts packages/core/src/__tests__/identity.test.ts packages/daemon/src/mcp/tools.ts packages/daemon/src/mcp/tools.test.ts packages/daemon/src/hooks.ts packages/daemon/src/hooks.test.ts packages/cli/src/commands/memory.ts packages/cli/src/commands/memory.test.ts packages/cli/src/features/setup.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-shared.test.ts packages/cli/src/commands/app.ts docs/ARCHITECTURE.md docs/CLI.md docs/HARNESSES.md
bun run --filter '@signet/core' typecheck
bun run --filter '@signet/connector-hermes-agent' typecheck
git diff --check
```

Result:

```text
@signet/core build passed
124 pass
0 fail
Hermes plugin py_compile passed
Targeted Biome check passed
@signet/core typecheck passed
@signet/connector-hermes-agent typecheck passed
git diff --check passed
```

Full-suite caveats:

- `bun run typecheck` was attempted. It exits 2 on an existing repo-wide desktop package issue: `@signet/desktop` cannot find the `electron` type definition file. The packages touched by this PR that expose typecheck scripts pass.
- `bun run lint` was attempted. It exits 1 on the current repo-wide Biome baseline, mainly package manifest formatting diagnostics outside this PR. Targeted Biome checks for this PR's changed TypeScript and docs files pass.
- `bun scripts/spec-deps-check.ts` was attempted. It currently fails on an existing unrelated dependency entry: `informed_by path does not exist for engram-informed-predictor-track: references/Engram`. This PR does not touch `docs/specs/`.

Manual validation:

- `signet recall --json "Nicholai durable facts Signet memory"` succeeds locally after increasing the timeout.
- Hermes bundled Signet plugin smoke test connects to the daemon, exposes `memory_search`, and returns recall results instead of timing out.
- Local Hermes plugin schema now exposes `hints`, `transcript`, and `structured` on `memory_store`.
- Local Hermes gateway restarted and is running.

## AI disclosure

- [x] AI tools were used

## Notes

The broader recall ranking issue is separate from this schema fix. This PR makes the fields available and forwarded correctly so callers can provide retrieval-friendly hints instead of being limited to content, type, importance, tags, pinned, and project.

### DIFF
--- a/packages/connector-hermes-agent/hermes-plugin/__init__.py
+++ b/packages/connector-hermes-agent/hermes-plugin/__init__.py
@@ -13,6 +13,7 @@ Config:
   - SIGNET_HOST / SIGNET_PORT env vars (default: localhost:3850)
   - SIGNET_DAEMON_URL env var for full URL override
   - SIGNET_AGENT_ID env var for agent scoping (default: "hermes-agent")
+  - SIGNET_AGENT_WORKSPACE env var for the active named-agent workspace
 """
 
 from __future__ import annotations
@@ -102,6 +103,32 @@ PROFILE_SCHEMA = {
 }
 
 ALL_TOOL_SCHEMAS = [SEARCH_SCHEMA, STORE_SCHEMA, PROFILE_SCHEMA]
+
+
+def _sanitize_env(value: str) -> str:
+    return value.strip().replace("\r", "").replace("\n", "")
+
+
+def _resolve_agent_workspace(agent_id: str, kwargs: Dict[str, Any]) -> str:
+    """Resolve the project/workspace path sent to Signet hooks.
+
+    Named Signet agents can have their own workspace at
+    $SIGNET_PATH/agents/{agent_id}. Prefer that workspace so daemon
+    session-start can load the agent's scoped identity files.
+    """
+    explicit = _sanitize_env(os.environ.get("SIGNET_AGENT_WORKSPACE", ""))
+    if explicit:
+        return str(Path(explicit).expanduser())
+
+    signet_path = _sanitize_env(os.environ.get("SIGNET_PATH", ""))
+    agents_root = Path(signet_path).expanduser() if signet_path else Path.home() / ".agents"
+    if agent_id and agent_id != "default":
+        candidate = agents_root / "agents" / agent_id
+        if candidate.exists():
+            return str(candidate)
+
+    fallback = kwargs.get("cwd", kwargs.get("project", os.getcwd()))
+    return str(Path(str(fallback)).expanduser())
 
 
 # ---------------------------------------------------------------------------
@@ -209,7 +236,7 @@ class SignetMemoryProvider(MemoryProvider):
             return
 
         self._session_key = session_id or "hermes-default"
-        self._project = kwargs.get("cwd", kwargs.get("project", os.getcwd()))
+        self._project = _resolve_agent_workspace(agent_id, kwargs)
 
         # Call session-start hook — get identity + memories + inject
         result = self._client.session_start(
--- a/packages/connector-hermes-agent/hermes-plugin/__init__.py
+++ b/packages/connector-hermes-agent/hermes-plugin/__init__.py
@@ -249,7 +249,7 @@ def _resolve_agent_workspace(agent_id: str, kwargs: Dict[str, Any]) -> str:
 
     signet_path = _sanitize_env(os.environ.get("SIGNET_PATH", ""))
     agents_root = Path(signet_path).expanduser() if signet_path else Path.home() / ".agents"
-    if agent_id and agent_id != "default":
+    if agent_id and agent_id not in ("default", "hermes-agent"):
         candidate = agents_root / "agents" / agent_id
         if candidate.exists():
             return str(candidate)
--- a/packages/connector-hermes-agent/hermes-plugin/__init__.py
+++ b/packages/connector-hermes-agent/hermes-plugin/__init__.py
@@ -69,6 +69,43 @@ MEMORY_SEARCH_SCHEMA = {
     },
 }
 
+STRUCTURED_ENTITY_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "source": {"type": "string", "description": "Source entity name."},
+        "sourceType": {"type": "string", "description": "Optional source entity type."},
+        "relationship": {"type": "string", "description": "Relationship from source to target."},
+        "target": {"type": "string", "description": "Target entity name."},
+        "targetType": {"type": "string", "description": "Optional target entity type."},
+        "confidence": {"type": "number", "description": "Optional confidence score 0-1."},
+    },
+    "required": ["source", "relationship", "target"],
+}
+
+STRUCTURED_ATTRIBUTE_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "content": {"type": "string", "description": "Attribute or constraint text."},
+        "confidence": {"type": "number", "description": "Optional confidence score 0-1."},
+        "importance": {"type": "number", "description": "Optional importance score 0-1."},
+    },
+    "required": ["content"],
+}
+
+STRUCTURED_ASPECT_SCHEMA = {
+    "type": "object",
+    "properties": {
+        "entityName": {"type": "string", "description": "Entity the aspect belongs to."},
+        "aspect": {"type": "string", "description": "Aspect name, e.g. preference, workflow, constraint."},
+        "attributes": {
+            "type": "array",
+            "items": STRUCTURED_ATTRIBUTE_SCHEMA,
+            "description": "Facts, constraints, or attributes for this aspect.",
+        },
+    },
+    "required": ["entityName", "aspect", "attributes"],
+}
+
 MEMORY_STORE_SCHEMA = {
     "name": "memory_store",
     "description": "Save a new memory to Signet.",
@@ -81,6 +118,36 @@ MEMORY_STORE_SCHEMA = {
             "tags": {"type": "string", "description": "Comma-separated tags for categorization."},
             "pinned": {"type": "boolean", "description": "Pin this memory so it does not decay."},
             "project": {"type": "string", "description": "Optional project path. Defaults to the active Hermes Signet workspace."},
+            "hints": {
+                "type": "array",
+                "items": {"type": "string"},
+                "description": "Prospective recall hints and alternate phrasings for retrieving this memory later.",
+            },
+            "transcript": {
+                "type": "string",
+                "description": "Raw source text or conversation transcript to preserve alongside this memory.",
+            },
+            "structured": {
+                "type": "object",
+                "description": "Pre-extracted structured data. When provided, Signet can persist graph links and hints directly.",
+                "properties": {
+                    "entities": {
+                        "type": "array",
+                        "items": STRUCTURED_ENTITY_SCHEMA,
+                        "description": "Entity relationships to link to this memory.",
+                    },
+                    "aspects": {
+                        "type": "array",
+                        "items": STRUCTURED_ASPECT_SCHEMA,
+                        "description": "Entity aspects and attributes to persist for graph recall.",
+                    },
+                    "hints": {
+                        "type": "array",
+                        "items": {"type": "string"},
+                        "description": "Prospective recall hints and alternate phrasings.",
+                    },
+                },
+            },
         },
         "required": ["content"],
     },
@@ -678,6 +745,17 @@ class SignetMemoryProvider(MemoryProvider):
                 return [t.strip() for t in value.split(",") if t.strip()]
             return [str(value).strip()] if str(value).strip() else None
 
+        def _string_list(value: Any) -> Optional[List[str]]:
+            if value is None or value == "":
+                return None
+            if isinstance(value, list):
+                items = [str(item).strip() for item in value if str(item).strip()]
+                return items or None
+            if isinstance(value, str):
+                stripped = value.strip()
+                return [stripped] if stripped else None
+            return None
+
         def _search(search_args: Dict[str, Any]) -> str:
             query = str(search_args.get("query", "")).strip()
             if not query:
@@ -714,6 +792,9 @@ class SignetMemoryProvider(MemoryProvider):
             if importance is None:
                 importance = 0.5
             importance = max(0.0, min(1.0, importance))
+            structured = store_args.get("structured")
+            if not isinstance(structured, dict):
+                structured = None
             result = self._client.remember(
                 content,
                 importance=importance,
@@ -721,6 +802,9 @@ class SignetMemoryProvider(MemoryProvider):
                 memory_type=str(store_args.get("type", "") or ""),
                 pinned=store_args.get("pinned") if isinstance(store_args.get("pinned"), bool) else None,
                 project=str(store_args.get("project", "") or self._project),
+                hints=_string_list(store_args.get("hints")),
+                transcript=str(store_args.get("transcript", "") or ""),
+                structured=structured,
                 who="hermes-agent",
             )
             if not result:
--- a/packages/connector-hermes-agent/hermes-plugin/__init__.py
+++ b/packages/connector-hermes-agent/hermes-plugin/__init__.py
@@ -64,6 +64,10 @@ MEMORY_SEARCH_SCHEMA = {
                 "description": "Deprecated compatibility alias for importance_min; ignored when importance_min is set.",
             },
             "score_min": {"type": "number", "description": "Minimum recall score threshold, applied client-side."},
+            "agent_scoped": {
+                "type": "boolean",
+                "description": "When true, scope recall to SIGNET_AGENT_ID instead of searching shared effective memory.",
+            },
         },
         "required": ["query"],
     },
@@ -778,6 +782,7 @@ class SignetMemoryProvider(MemoryProvider):
                 keyword_query=str(search_args.get("keyword_query", "") or ""),
                 expand=bool(search_args.get("expand", False)),
                 score_min=_as_float(search_args.get("score_min")),
+                agent_scoped=bool(search_args.get("agent_scoped", False)),
             )
             if not result:
                 return json.dumps({"error": "Search failed or Signet daemon returned no response.", "results": []})
--- a/packages/connector-hermes-agent/hermes-plugin/__init__.py
+++ b/packages/connector-hermes-agent/hermes-plugin/__init__.py
@@ -203,7 +203,6 @@ MEMORY_FORGET_SCHEMA = {
         "properties": {
             "id": {"type": "string", "description": "Memory ID to forget."},
             "reason": {"type": "string", "description": "Why this memory should be forgotten."},
-            "force": {"type": "boolean", "description": "Hard-delete instead of soft-delete, when supported."},
         },
         "required": ["id", "reason"],
     },
@@ -864,7 +863,6 @@ class SignetMemoryProvider(MemoryProvider):
                 result = self._client.forget_memory(
                     memory_id,
                     reason=reason,
-                    force=bool(args.get("force", False)),
                 )
                 return json.dumps(result if result else {"error": "Failed to forget memory."})
 
--- a/packages/connector-hermes-agent/hermes-plugin/__init__.py
+++ b/packages/connector-hermes-agent/hermes-plugin/__init__.py
@@ -5,9 +5,11 @@ Bridges Hermes Agent's memory provider interface to the Signet daemon
 predictive recall, cross-session memory, and the full Signet pipeline
 (extraction, knowledge graph, retention decay, synthesis).
 
-The 3 tools (signet_search, signet_store, signet_profile) are exposed
-through the MemoryProvider interface. The daemon handles all heavy lifting:
-embedding, reranking, knowledge graph traversal, and predictive scoring.
+Canonical Signet memory tools (memory_search, memory_store, memory_get,
+memory_list, memory_modify, memory_forget, plus recall/remember aliases) are
+exposed through the MemoryProvider interface. The daemon handles all heavy
+lifting: embedding, reranking, knowledge graph traversal, and predictive
+scoring.
 
 Config:
   - SIGNET_HOST / SIGNET_PORT env vars (default: localhost:3850)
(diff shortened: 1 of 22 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(hermes): scope Signet memory to named agents` (commit `0797f727`) now and I'll follow up shortly with feedback.

packages/connector-hermes-agent/hermes-plugin/client.py
```diff
@@ -115,6 +116,40 @@ def _get(
...
+            return None
+
+    def _delete(
+        self,
+        path: str,
+        *,
+        timeout: float = _TIMEOUT_SECS,
+    ) -> Optional[Dict[str, Any]]:
+        """DELETE from the daemon. Returns parsed response or None on failure."""
+        url = f"{self._base_url}{path}"
```
PR-Reviewer-Ant, line 160: `json.loads(resp.read().decode("utf-8"))` in `_delete` throws `json.JSONDecodeError` (a subclass of `ValueError`) when the daemon returns an empty body (e.g. 204 No Content). The except clause only catches `HTTPError, URLError, OSError, TimeoutError` — `JSONDecodeError` is not among them. The same bug exists in `_patch` at the equivalent line (~line 127). The exception propagates to `handle_tool_call`'s outer `except Exception` block, which means `memory_forget` and `memory_modify` return a false `{"error": "Signet memory_forget failed: Expecting value: line 1 column 1 (char 0)"}` to the model even when the daemon accepted and executed the request. Fix: catch `(ValueError, json.JSONDecodeError)` alongside the existing exceptions, or guard with `body = resp.read(); return json.loads(body) if body else {}`.

packages/connector-hermes-agent/src/index.ts
```diff
@@ -133,6 +133,48 @@ function isProviderConfigured(hermesHome: string): boolean {
...
+		const getResp = await fetch(`${baseUrl}/api/agents/${encodeURIComponent(agentId)}`, {
+			signal: AbortSignal.timeout(1_000),
+		});
+		if (getResp.ok) return;
+	} catch {
+		// Daemon may be offline; the POST below will produce the user-facing warning.
+	}
+
+	try {
+		const resp = await fetch(`${baseUrl}/api/agents`, {
```
PR-Reviewer-Ant, line 191: `ensureNamedAgentRegistered` makes two bare `fetch` calls — the GET existence check and the POST to create the agent — with no `Authorization` header. The Python client's `_headers()` always injects `Bearer ${SIGNET_TOKEN}` when `SIGNET_TOKEN` is set, exactly because the daemon may require authentication for non-localhost deployments. Without the header, both calls will receive 401 from an authenticated daemon: the GET returns non-ok (falling through to the POST), and the POST also returns non-ok, producing a warning that looks like a daemon-unreachable error rather than an auth failure. The net effect is that agent auto-registration is silently non-functional for the remote-daemon configuration the PR explicitly acknowledges (the `SIGNET_TOKEN` warning in the `.env` section just below). Fix: read `SIGNET_TOKEN` from `process.env` (stripping newlines as the rest of this function does) and include `Authorization: Bearer ${token}` when present.

packages/connector-hermes-agent/index.test.ts
```diff
@@ -42,6 +47,7 @@ afterEach(() => {
 	restoreEnv("HERMES_HOME");
```
PR-Reviewer-Ant, line 59: `SIGNET_AGENT_WORKSPACE` is captured in `originalEnv`, deleted in `beforeEach`, and set by at least one test (`"derives SIGNET_AGENT_WORKSPACE for named agents"`), but `restoreEnv("SIGNET_AGENT_WORKSPACE")` is absent from `afterEach`. Any test that runs after the scoping test will inherit the leaked workspace path, which can silently skew workspace resolution in subsequent assertions. Add `restoreEnv("SIGNET_AGENT_WORKSPACE")` to the `afterEach` block alongside the other keys.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `0797f727`

[Automated Bot Review — pr-reviewer] PR #504 correctly scopes Hermes memory to named-agent workspaces and extends recall timeouts. Found one blocking correctness bug in the new Python HTTP helpers, one warning-level auth gap in the TS registration helper, and one test-env leak.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All three findings are directly provable from the changed lines. (1) The `_delete`/`_patch` except-clauses are quoted verbatim — `json.JSONDecodeError` is absent, confirming the crash path. (2) The `fetch` headers in `ensureNamedAgentRegistered` are quoted verbatim — no Authorization header is present. (3) The `afterEach` block is quoted verbatim — `restoreEnv("SIGNET_AGENT_WORKSPACE")` is absent despite `SIGNET_AGENT_WORKSPACE` being in `originalEnv` and cleared in `beforeEach`.

packages/daemon/src/mcp/tools.ts
```diff
@@ -698,10 +702,15 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
 						aspects: z
 							.array(
 								z.object({
-									entity: z.string(),
+									entityName: z.string(),
 									aspect: z.string(),
```
PR-Reviewer-Ant: Breaking schema change in the `structured.aspects` array: `entity` (string) was renamed to `entityName` and `value` (string) was replaced by `attributes` (array of objects). Any caller that previously passed `{entity, aspect, value}` tuples will now have those fields silently dropped by Zod. The Python plugin's `STRUCTURED_ASPECT_SCHEMA` matches the new shape, so Python→daemon is now aligned, but the rename is not backward-compatible for any TypeScript/JS callers (MCP clients, direct API consumers) already using the old schema. The PR description does not mention this as a breaking change. If the structured parameter was purely internal or never shipped to external callers, this is fine — but that should be confirmed before merge.

packages/connector-hermes-agent/src/index.ts
```diff
@@ -133,6 +133,48 @@ function isProviderConfigured(hermesHome: string): boolean {
...
+			method: "POST",
+			headers: { "Content-Type": "application/json" },
```
PR-Reviewer-Ant: `ensureNamedAgentRegistered` issues an unauthenticated POST to `/api/agents` (no `Authorization` header, no `SIGNET_TOKEN` injection). The existing daemon auth middleware protects mutation endpoints; on any instance running in non-localhost or token-required mode this POST will return 401/403 and the agent entry will not be created. The failure is soft (warning added, install still succeeds), and the warning message is actionable, so this is not a correctness regression for local setups. However, the install experience on remote authenticated daemons will silently degrade: the named-agent scope is never registered, so subsequent Hermes sessions will fall back to unscoped memory. Consider reading `SIGNET_TOKEN` from env (already done for `configuredDaemonUrl`) and forwarding it as a bearer token here.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `33005759`

[Automated bot review — pr-reviewer] PR #504 correctly implements named-agent workspace scoping and the recall timeout fix described in the description. Four issues found: one warning-level breaking schema change, one warning-level auth gap on a new install-time mutation, and two nitpick-level test hygiene issues.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The breaking schema change in tools.ts (entity→entityName, value→attributes) is directly visible in the diff. Whether any existing callers depend on the old field names cannot be determined without the full daemon route handler and any external consumers of the MCP memory_store structured payload. The auth gap in ensureNamedAgentRegistered is directly visible; whether /api/agents requires a bearer token depends on the daemon auth middleware which is not included in this diff.

packages/connector-hermes-agent/index.test.ts
PR-Reviewer-Ant, line 59: `SIGNET_AGENT_WORKSPACE` is captured in `originalEnv` (line 11) and deleted in `beforeEach` (line 33), but `restoreEnv("SIGNET_AGENT_WORKSPACE")` is absent from `afterEach`. If a future test sets this env var, it will leak into subsequent tests. Add `restoreEnv("SIGNET_AGENT_WORKSPACE")` to the afterEach block alongside the other restores.

packages/connector-hermes-agent/hermes-plugin/client.py
```diff
@@ -257,33 +293,153 @@ def remember(
...
-            body["minScore"] = min_score
-        return self._post("/api/memory/recall", body)
+        if project:
+            body["project"] = project
+        if memory_type:
+            body["type"] = memory_type
+        if tags:
+            body["tags"] = tags
+        if who:
+            body["who"] = who
```
PR-Reviewer-Ant, line 374: In the `score_min` filtering loop: `float(row.get("score", 0.0))` will raise `TypeError` if a result row contains `"score": null` (JSON null → Python `None`), because `dict.get` returns the stored value (`None`) rather than the default when the key exists. The outer `try/except Exception` in `handle_tool_call` catches this and returns an error JSON, but it logs at `ERROR` level and drops the entire result set rather than gracefully skipping the bad row. Guard with `float(row.get("score") or 0.0)` or an explicit `None` check.

packages/connector-hermes-agent/index.test.ts
PR-Reviewer-Ant, line 59: `SIGNET_AGENT_WORKSPACE` is captured in `originalEnv` (line 10) but is never restored in `afterEach`. Every other key added in this PR (`SIGNET_SKIP_AGENT_REGISTER`) is restored; `SIGNET_AGENT_WORKSPACE` is the only outlier. If the env var is set in the outer process before the suite runs, its original value will be lost after the suite completes. Add `restoreEnv("SIGNET_AGENT_WORKSPACE")` to the `afterEach` block.

packages/connector-hermes-agent/src/index.ts
PR-Reviewer-Ant, line 191: `ensureNamedAgentRegistered` fires a `POST /api/agents` mutation without including an auth token. All other mutation paths in this codebase (`remember`, `recall`, etc.) conditionally include `Authorization: Bearer $SIGNET_TOKEN` when the token is available. For remote daemon deployments where auth is required, this request will get a 401/403, and the failure message will read `Could not register Signet agent '…' with shared memory policy: [auth error body]` — which is misleading. The `SIGNET_TOKEN` value should be read from env and included here for consistency, mirroring the Python client's `_headers()` pattern.

packages/connector-hermes-agent/src/index.ts
```diff
@@ -133,6 +133,48 @@ function isProviderConfigured(hermesHome: string): boolean {
...
+			body: JSON.stringify({
+				name: agentId,
+				read_policy: "shared",
+				policy_group: null,
+			}),
```
PR-Reviewer-Ant, line 198: `ensureNamedAgentRegistered` unconditionally registers the named agent with `read_policy: "shared"`. There is no way for the installer to opt for `isolated` or `group` policy via env var or flag. If a user already has a named agent with `isolated` policy and re-runs install (e.g. after an upgrade), the GET succeeds and the POST is skipped — so existing policy is preserved. But on first install the policy is always forced to `shared`. This is undocumented and could silently over-share memory. At minimum the warning emitted on failure should mention `signet agent create <id> --memory isolated` as an alternative so users know their options.

packages/core/src/agents.ts
```diff
@@ -56,26 +62,20 @@ export function scaffoldAgent(name: string, agentsDir: string): void {
  * exists on disk for the given agent.
```
PR-Reviewer-Ant, line 62: The old implementation gave every file in `SHARED_FILES` (`AGENTS.md`, `USER.md`, `TOOLS.md`, `HEARTBEAT.md`, `MEMORY.md`, `BOOTSTRAP.md`) a fixed root path — per-agent overrides were only supported for `SOUL.md` and `IDENTITY.md`. This PR expands the fallback-first lookup to all eight files including `MEMORY.md` and `USER.md`. The test at `packages/core/src/__tests__/identity.test.ts:107` confirms the new behavior is intentional, but the change is broader than the PR description implies ("scope Hermes Signet memory to the named agent workspace"). Specifically, `MEMORY.md` being per-agent changes the multi-agent shared-memory model for any existing deployment where `MEMORY.md` is the single shared synthesis artifact. Callers that relied on `MEMORY.md` always being root-level will now receive an agent-scoped override if one exists. This should be explicitly documented in the PR or in `docs/ARCHITECTURE.md` / `docs/PIPELINE.md` to avoid silent regressions for existing multi-agent setups.

packages/connector-hermes-agent/index.test.ts
```diff
@@ -1,5 +1,5 @@
 import { afterEach, beforeEach, describe, expect, it } from "bun:test";
```
PR-Reviewer-Ant, line 1: The PR checklist shows `bun run typecheck` and `bun run lint` as unchecked. Per repo AGENTS.md the PR checklist is mandatory. The diff touches TypeScript across `@signet/core`, `@signet/daemon`, and `@signet/connector-hermes-agent` — any silent type error (e.g. in the new `IdentityFileMap` parameter threading through `hooks.ts`) would not be caught by the 45-test run alone. Please confirm both pass before merge.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fc7f85ef`

[Automated Review — pr-reviewer bot] PR #504 introduces Hermes named-agent scoping, recall timeout increases, and expanded memory_store schema. The implementation broadly matches the stated goals. Four issues are worth discussion or fixing before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - All findings are grounded in explicit lines in the diff and the full changed-file contents provided. The `afterEach` env-restore omission is directly visible in `index.test.ts`. The unauthenticated fetch in `ensureNamedAgentRegistered` is explicit in `src/index.ts`. The `getAgentIdentityFiles` scope expansion is verifiable by diffing the old and new `IDENTITY_FILES` constant. The unchecked typecheck/lint items are stated in the PR checklist itself.

NicholaiVogel:
Addressed the review feedback in `5bb4dff5`.

What changed:

- Hermes install-time agent registration now forwards `Authorization: Bearer $SIGNET_TOKEN` on both the GET and POST paths.
- Named-agent registration no longer hardcodes shared memory only. It supports `SIGNET_AGENT_READ_POLICY` / `SIGNET_AGENT_MEMORY_POLICY` values of `shared`, `isolated`, and `group`, with `SIGNET_AGENT_POLICY_GROUP` for group policy.
- Test env cleanup now restores `SIGNET_AGENT_WORKSPACE` and the new Signet registration env vars.
- Hermes plugin response parsing now treats empty successful daemon responses as `{}` and catches JSON decode failures across GET/POST/PATCH/DELETE.
- `score_min` filtering now handles null score values without dropping the entire result set.
- MCP `memory_store.structured.aspects` now accepts both the new structured shape and legacy `{entity, aspect, value}` tuples, normalizing legacy callers before forwarding to remember.
- Documented agent-local identity overrides, including optional agent-local `MEMORY.md` fallback behavior, in `docs/ARCHITECTURE.md` and `docs/HARNESSES.md`.
- Documented Hermes plugin memory policy env vars.

Validation:

```bash
bun test packages/daemon/src/mcp/tools.test.ts packages/connector-hermes-agent/index.test.ts packages/cli/src/commands/memory.test.ts packages/core/src/__tests__/identity.test.ts packages/daemon/src/hooks.test.ts
python3 -m py_compile packages/connector-hermes-agent/hermes-plugin/__init__.py packages/connector-hermes-agent/hermes-plugin/client.py
bunx biome check packages/connector-hermes-agent/src/index.ts packages/connector-hermes-agent/index.test.ts packages/daemon/src/mcp/tools.ts packages/daemon/src/mcp/tools.test.ts
bun run --filter '@signet/connector-hermes-agent' typecheck
bun run --filter '@signet/core' typecheck
git diff --check
```

Results: 105 tests pass, py_compile passed, targeted Biome passed, targeted typechecks passed, diff check passed.

Full `bun run typecheck` was also attempted and still fails on unrelated existing `@signet/*-pi*` package issues. Full `bun run lint` was attempted and still hits the repo-wide Biome baseline; targeted checks for the changed TypeScript files pass.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `5bb4dff5`

```pr-review-json
{
  "summary": "[Automated Review — pr-reviewer bot] Four issues found: one cross-language logic divergence that could cause silent mis-scoping for the default `hermes-agent` agent ID, one stale `meta.noHits` field after client-side `score_min` filtering, one registration function that proceeds to POST on any non-OK GET (not just 404), and one breaking behavioral change in `getAgentIdentityFiles` for existing named-agent deployments.",
  "verdict": "comment",
  "confidence": {
    "level": "medium",
    "justification": "Issues 1-3 are directly provable from the changed lines. Issue 4 depends on whether any existing named-agent deployments have files in their agent subdirectory other than SOUL.md/IDENTITY.md — that runtime state is not visible in the diff alone."
  },
  "comments": [
    {
      "file": "packages/connector-hermes-agent/hermes-plugin/__init__.py",
      "line": 252,
      "body": "**Logic divergence from the TypeScript connector.** `_resolve_agent_workspace` checks `agent_id != \"default\"` but does not exclude `\"hermes-agent\"` the way the TS connector does:\n\n```typescript\n// src/index.ts\n} else if (signetAgentId && signetAgentId !== \"hermes-agent\" && signetAgentId !== \"default\") {\n```\n\nConsequence: if `~/.agents/agents/hermes-agent/` exists on disk (a plausible layout when Hermes runs as a named agent), the Python plugin will silently adopt that directory as `self._project` and send it as the session-start `project`. The TS connector, however, will not write `SIGNET_AGENT_WORKSPACE` for `hermes-agent`, so the daemon sees a project path that was not explicitly configured. Add `agent_id != \"hermes-agent\"` to the guard to match the TS side.",
      "severity": "warning"
    },
    {
      "file": "packages/connector-hermes-agent/hermes-plugin/client.py",
      "line": 396,
      "body": "**`meta.noHits` is not updated after client-side `score_min` filtering.** After the filter, `result[\"results\"]` may be empty (`kept = []`) while `meta[\"noHits\"]` remains `False` (the daemon's original response had results). `totalReturned` is patched to `len(kept)` correctly, but `noHits` is left stale. Any caller that reads `meta.noHits` to decide whether to show a \"no results\" message will get the wrong signal when score thresholding removes all rows. Fix:\n\n```python\nresult[\"meta\"] = {**meta, \"totalReturned\": len(kept), \"noHits\": len(kept) == 0}\n```",
      "severity": "warning"
    },
    {
      "file": "packages/connector-hermes-agent/src/index.ts",
      "line": 165,
      "body": "**`ensureNamedAgentRegistered` proceeds to POST on any non-OK GET, not just 404.** If the daemon returns 401 (auth failure) or 500, `getResp.ok` is false, the catch block is skipped, and the code falls through to attempt a POST creation. The POST will also fail and generate a warning like "Could not register Signet agent 'X'…" — which misleads the user into thinking the agent doesn't exist, when the real issue is auth or a transient daemon error.\n\nConsider checking `getResp.status === 404` explicitly before falling through, or at minimum gating the POST on status 404:\n\n```typescript\nif (getResp.status !== 404) return; // already exists or unrecoverable error\n```",
      "severity": "warning"
    },
    {
      "file": "packages/core/src/agents.ts",
      "line": 72,
      "body": "**Behavioral breaking change for existing named-agent deployments.** Previously only `SOUL.md` and `IDENTITY.md` were checked in the agent subdirectory; all other files (`AGENTS.md`, `USER.md`, `TOOLS.md`, `HEARTBEAT.md`, `MEMORY.md`, `BOOTSTRAP.md`) always resolved from the workspace root. This PR changes the rule so all files check the agent subdir first.\n\nAny named-agent directory that already contains a file from the newly promoted set (most likely `AGENTS.md` or `MEMORY.md` — e.g., scaffolded by `scaffoldAgent` in an earlier version, or written by a user) will silently switch from reading the workspace-root version to the agent-local version on next daemon restart, with no migration warning.\n\nThis is the core intent of the PR and is documented correctly in ARCHITECTURE.md. Flagging here so reviewers confirm there is no automated migration concern or need for a startup warning when unexpected files are found in agent subdirectories.",
      "severity": "warning"
    }
  ]
}
```

[stderr]

NicholaiVogel:
Follow-up setup pass pushed in `7fa4bbc6`.

This makes Hermes Agent a real setup harness instead of a hidden non-interactive option:

- Added Hermes Agent to the interactive `signet setup` harness checklist.
- Updated `signet setup --harness` help text to include `hermes-agent` and `pi`.
- Existing setup migration now carries detected Hermes installs into `detectedHarnesses`, so the connector is installed during migration.
- Detection now recognizes a Hermes repo before the Signet plugin is installed by checking for the Hermes `plugins/memory` tree, while `resolveHermesRepoPluginPath()` still checks for the installed Signet plugin specifically.
- Updated CLI docs harness list.
- Added regression coverage for Hermes setup harness parsing and pre-plugin Hermes repo detection.

Additional validation:

```bash
bun test packages/daemon/src/mcp/tools.test.ts packages/connector-hermes-agent/index.test.ts packages/cli/src/commands/memory.test.ts packages/core/src/__tests__/identity.test.ts packages/daemon/src/hooks.test.ts packages/cli/src/features/setup-shared.test.ts
python3 -m py_compile packages/connector-hermes-agent/hermes-plugin/__init__.py packages/connector-hermes-agent/hermes-plugin/client.py
bunx biome check packages/core/src/identity.ts packages/core/src/index.ts packages/core/src/__tests__/identity.test.ts packages/connector-hermes-agent/src/index.ts packages/cli/src/features/setup.ts packages/cli/src/features/setup-migrate.ts packages/cli/src/features/setup-shared.test.ts packages/cli/src/commands/app.ts docs/CLI.md
bun run --filter '@signet/core' typecheck
bun run --filter '@signet/connector-hermes-agent' typecheck
git diff --check
```

Result: 118 tests pass, py_compile passed, targeted Biome passed, targeted typechecks passed, diff check passed.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7fa4bbc6`

You're out of extra usage · resets 10pm (America/Denver)

[stderr]

NicholaiVogel:
Follow-up review pass pushed in `c39fd3c0`.

Handled the actionable feedback:

- Python plugin workspace resolution now excludes both `default` and `hermes-agent`, matching the TypeScript connector behavior before looking for `~/.agents/agents/<id>`.
- Client-side `score_min` filtering now updates both `meta.totalReturned` and `meta.noHits` when filtering empties the result set.
- Install-time named-agent registration now only attempts `POST /api/agents` after a 404 existence check. Non-404 failures, such as 401 or 500, become warnings and do not trigger a mutation.
- Added regression coverage for the non-404 registration path.
- Added bundled-plugin invariant checks around empty JSON bodies, `ValueError` handling, null score handling, `noHits`, and the Python `hermes-agent` workspace guard.
- Added a scaffold regression test showing stock named-agent scaffolding only creates `SOUL.md` and `IDENTITY.md`. The broader agent-local identity override behavior is intentional, but stock scaffolds will not silently create agent-local `AGENTS.md`, `USER.md`, or `MEMORY.md` overrides.

Triage on the other inline comments:

- Auth headers on the registration GET/POST were already addressed in `5bb4dff5`; both requests include `Authorization: Bearer $SIGNET_TOKEN` when configured.
- `SIGNET_AGENT_WORKSPACE` restore was already present in the test `afterEach` at the reviewed head.
- Empty successful Python responses and JSON decode failures were already guarded by `_read_json_response()` and `ValueError` catches, but I added invariant coverage so that does not regress.
- First-registration policy is documented in the Hermes plugin README: `SIGNET_AGENT_READ_POLICY` accepts `shared`, `isolated`, or `group`, with `shared` as the default. Existing agents are preserved because registration is skipped when the GET succeeds.

Validation rerun:

```bash
bun test packages/daemon/src/mcp/tools.test.ts packages/connector-hermes-agent/index.test.ts packages/cli/src/commands/memory.test.ts packages/core/src/__tests__/identity.test.ts packages/daemon/src/hooks.test.ts packages/cli/src/features/setup-shared.test.ts
python3 -m py_compile packages/connector-hermes-agent/hermes-plugin/__init__.py packages/connector-hermes-agent/hermes-plugin/client.py
bunx biome check packages/connector-hermes-agent/src/index.ts packages/connector-hermes-agent/index.test.ts packages/core/src/__tests__/identity.test.ts packages/daemon/src/mcp/tools.ts packages/daemon/src/mcp/tools.test.ts
bun run --filter '@signet/core' typecheck
bun run --filter '@signet/connector-hermes-agent' typecheck
git diff --check
```

Result: `120 pass`, py_compile passed, targeted Biome passed, touched package typechecks passed, and `git diff --check` passed.

I also attempted the repo-level gates again:

- `bun run lint` still fails on the existing repo-wide Biome baseline, mainly package manifest formatting outside this PR.
- `bun run typecheck` still fails on existing unrelated `@signet/*-pi*` package errors around missing `@signet/pi-extension-base` and connector-base exports.

The PR body has been updated with the current test count and the full-suite caveats.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `ee33554b`

[Automated Review by pr-reviewer] I found one blocking data-safety issue and one implementation/description divergence that should be addressed before human approval.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, ambiguous_requirements] - The new Hermes plugin schema and handler explicitly add and forward a `force` flag on `memory_forget`, which introduces irreversible deletion capability from model-callable tools. I can point to concrete new lines in the changed Python plugin. The timeout divergence is also directly visible in constants/call sites, but whether 15s vs 30s is acceptable is somewhat requirement-dependent.

packages/connector-hermes-agent/hermes-plugin/__init__.py
```diff
@@ -38,70 +41,221 @@
...
+    "parameters": {
+        "type": "object",
+        "properties": {
+            "id": {"type": "string", "description": "Memory ID to forget."},
+            "reason": {"type": "string", "description": "Why this memory should be forgotten."},
+            "force": {"type": "boolean", "description": "Hard-delete instead of soft-delete, when supported."},
+        },
+        "required": ["id", "reason"],
+    },
+}
```
PR-Reviewer-Ant, line 213: The new `memory_forget` tool exposes `force` (hard delete) to model-callable operations while the tool description says 'Soft-delete'. This expands destructive attack surface and can cause irreversible data loss via prompt injection or tool misuse. Recommend removing `force` from Hermes-exposed schema/handler (or gating it behind explicit operator-only controls).

packages/connector-hermes-agent/hermes-plugin/client.py
```diff
@@ -161,6 +204,7 @@ def user_prompt_submit(
                 "agentId": self._agent_id,
```
PR-Reviewer-Ant, line 212: PR description says recall paths were given enough time for larger stores, but `user_prompt_submit()` uses `_LONG_TIMEOUT_SECS` (15s) while recall is 30s. If this hook is part of the 'recall paths' claim, this is a behavior mismatch likely to preserve some timeout failures under load. Clarify intent or align timeout values.

packages/connector-hermes-agent/hermes-plugin/__init__.py
```diff
@@ -566,55 +720,151 @@ def handle_tool_call(self, tool_name: str, args: Dict[str, Any], **kwargs) -> st
...
+                limit=_as_int(search_args.get("limit"), 10, minimum=1, maximum=50),
+                project=str(search_args.get("project", "") or ""),
+                memory_type=str(search_args.get("type", "") or ""),
+                tags=str(search_args.get("tags", "") or ""),
+                who=str(search_args.get("who", "") or ""),
+                pinned=search_args.get("pinned") if isinstance(search_args.get("pinned"), bool) else None,
+                importance_min=importance_min,
+                since=str(search_args.get("since", "") or ""),
+                until=str(search_args.get("until", "") or ""),
+                keyword_query=str(search_args.get("keyword_query", "") or ""),
```
PR-Reviewer-Ant, line 782: PR description says explicit recall remains unscoped "unless the caller asks for agent scoping," but the tool surface does not expose any agent-scoping flag and `_search()` never forwards one to `client.recall(...)`. As implemented, explicit recall appears always unscoped, so callers cannot actually request scoped behavior.

packages/connector-hermes-agent/src/index.ts
```diff
@@ -276,7 +374,7 @@ export class HermesAgentConnector extends BaseConnector {
 			try {
 				let envContent = readFileSync(envPath, "utf-8");
 				let changed = false;
-				for (const key of ["SIGNET_DAEMON_URL", "SIGNET_AGENT_ID"]) {
+				for (const key of ["SIGNET_DAEMON_URL", "SIGNET_AGENT_ID", "SIGNET_AGENT_WORKSPACE"]) {
```
PR-Reviewer-Ant: Security hygiene regression: install now persists `SIGNET_TOKEN` into `~/.hermes/.env`, but uninstall cleanup only removes `SIGNET_DAEMON_URL`, `SIGNET_AGENT_ID`, and `SIGNET_AGENT_WORKSPACE`. This leaves bearer credentials behind after uninstall.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `f2f71c11`

[Automated Review by pr-reviewer] I found one implementation/claim mismatch that should be fixed before merge, plus one security-hygiene issue.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The issues are directly visible in changed code paths: explicit recall scoping is not actually caller-controllable in the Hermes plugin handler despite the PR claim, and install now writes SIGNET_TOKEN but uninstall does not remove it. I did not run the Hermes runtime end-to-end here, so runtime impact is inferred from code flow.

packages/connector-hermes-agent/hermes-plugin/client.py
```diff
@@ -257,33 +301,152 @@ def remember(
...
+            body["tags"] = tags
+        if who:
+            body["who"] = who
+        if pinned is not None:
+            body["pinned"] = pinned
+        if importance_min is not None:
+            body["importance_min"] = importance_min
+        if since:
+            body["since"] = since
+        if until:
```
PR-Reviewer-Ant, line 381: Client-side `score_min` filtering does `float(row.get("score") or 0.0)` without guarding conversion errors. If any result has a non-numeric `score` (string/object from legacy or unexpected daemon response), this raises and aborts recall/tool execution. Consider defensive parsing (try/except per row) and treating invalid scores as 0.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `51832b15`

[Automated Review by pr-reviewer] I found two issues worth addressing before human review: one docs/behavior mismatch and one robustness bug in Hermes recall filtering.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - Both findings are directly visible in the changed lines: the CLI docs harness list omits `hermes-agent` while code now supports it, and the new `float(row.get("score")...)` path in the Hermes client has no local try/except and can raise on non-numeric score payloads. I did not run the daemon/CLI to reproduce malformed-score behavior end-to-end.


- docs/CLI.md:106 - The `signet setup --harness` option docs still omit `hermes-agent` even though the implementation now supports it (`packages/cli/src/commands/app.ts`). This diverges from the PR claim that Hermes is first-class in setup/docs and can mislead non-interactive users into thinking Hermes is unsupported.

PR #505, opened 2026-04-13
fix(pipeline): reduce OpenCode session overhead for pipeline jobs

Ostico:
## Summary

Reduce OpenCode per-session token overhead so pipeline jobs complete within timeouts. Register a dedicated `signet-pipeline` agent with all tools/skills denied, cutting input tokens from ~47k to ~5k and wall time from 75s+ to ~5s.

## Changes

- `packages/connector-opencode/src/index.ts` — `registerPipelineAgent()`, `removePipelineAgent()`, `ensureConfigFile()`, import shared prompt constant
- `packages/daemon/src/pipeline/provider.ts` — send `agent` param in `buildMessageBody()`, import shared prompt constant, add token measurement comment
- `packages/daemon/src/memory-config.ts` — bump hints timeout to 60s at DEFAULTS source
- `packages/daemon/src/pipeline/extraction.ts` — document known JSON parse failure
- `packages/core/src/pipeline-providers.ts` — add `OPENCODE_PIPELINE_AGENT`, `OPENCODE_PIPELINE_SYSTEM_PROMPT` shared constants
- `packages/core/src/index.ts` — export new constants
- `packages/connector-opencode/index.test.ts` — 8 TDD tests for agent lifecycle + fresh-install ordering regression test

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

N/A — no migrations touched.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes (pre-existing `noDelete` warnings in test files only)

### Validation results

| Metric | Before (default agent) | After (signet-pipeline) |
|---|---|---|
| Input tokens | ~47,000 | 4,844 |
| Wall time | 75s+ | ~5s |
| Response quality | Empty/markdown | Valid JSON |

### Test suites

- `connector-opencode`: 11/11 pass (8 new TDD tests for register/remove lifecycle + fresh-install ordering)
- `provider`: 56/56 pass (4 new tests for agent param + fallback; 5 pre-existing codex-not-found failures unrelated)
- `memory-config`: 80/80 pass
- `extraction`: existing suite passes

## AI disclosure

## Notes

- The `signet-pipeline` agent config uses `"*": "deny"` permissions and `hidden: true` so it never appears in user agent lists
- `ensureConfigFile()` creates an empty `opencode.json` before `registerPlugin`/`registerMcpServer`/`registerPipelineAgent` to fix fresh-install ordering where plugin and MCP entries were lost
- `OPENCODE_PIPELINE_SYSTEM_PROMPT` is shared between connector (agent config) and provider (per-call system override) to prevent drift
- 5 pre-existing codex provider test failures (codex binary not on PATH) are unrelated to this PR


PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(pipeline): reduce OpenCode session overhead for pipeline jobs` (commit `f7ce1bd5`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1970,6 +1970,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		const body: Record<string, unknown> = {
 			parts: [{ type: "text", text: prompt }],
 			model: { providerID, modelID },
+			agent: "signet-pipeline",
```
PR-Reviewer-Ant: **Blocking: unconditional coupling to an optionally-installed connector artifact.**

`agent: "signet-pipeline"` is sent on every OpenCode request regardless of whether the agent has been registered. The agent is only written to `opencode.json` when `signet install` is run via `connector-opencode`. Users who:

- Upgrade the daemon without re-running `signet install`
- Use OpenCode but haven't run install
- Have an older connector that predates this PR

…will have ALL OpenCode pipeline calls fail or silently misbehave, because the `sendMessage()` error path has no handler for agent-not-found. The 422 handler checks only `i.path[0] === 'format'`; any agent-related 422 (e.g. `i.path[0] === 'agent'`) falls through to a hard throw. A runtime check (test availability of the agent, or make agent param conditional on config/feature flag) is needed before this is safe to ship unconditionally.

PR-Reviewer-Ant: **Warning: duplicated string constant across packages.**

The value `"signet-pipeline"` is defined as `private static readonly PIPELINE_AGENT_KEY` in `connector-opencode/src/index.ts` and hardcoded here as a string literal. AGENTS.md explicitly prohibits duplicating constants across files. If the agent key is ever renamed, this will silently diverge. The constant should be exported from `@signet/core` (or a shared location) and imported in both files.

packages/connector-opencode/src/index.ts
```diff
@@ -524,6 +528,64 @@ export class OpenCodeConnector extends BaseConnector {
 		}
 	}
 
+	private static readonly PIPELINE_AGENT_KEY = "signet-pipeline";
```
PR-Reviewer-Ant: **Warning: silent no-op when no config file exists.**

`registerPipelineAgent()` iterates `getConfigCandidates()` and skips files that don't exist. If the user has OpenCode installed but no config file yet (fresh install with no prior config), the agent is never written. The daemon will then unconditionally send `agent: "signet-pipeline"` to an OpenCode that has no knowledge of it. Consider creating the config file if missing (matching the behavior pattern of `registerMcpServer` which also relies on an existing file, so this is pre-existing for MCP too — but becomes a new failure mode now that the daemon depends on the agent being present).

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `f7ce1bd5`

[Automated Bot Review — pr-reviewer] PR #505 reduces OpenCode session overhead by registering a `signet-pipeline` agent via the connector and hardcoding the agent name into the daemon provider. Two issues found: one blocking correctness risk from unconditional coupling between the daemon and an optionally-installed connector artifact, and one convention violation around duplicated string constants.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The unconditional `agent: 'signet-pipeline'` in buildMessageBody() is directly visible in the diff at provider.ts line 1973. The 422 error handler at sendMessage() explicitly checks only `i.path[0] === 'format'` — there is no agent-rejection handler. The duplication of the string literal is provable from the diff: PIPELINE_AGENT_KEY is private static in connector-opencode and hardcoded separately in provider.ts. Both findings are grounded in changed lines, not inference.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `6735107e`

```pr-review-json
{
  "summary": "[Automated Bot Review] Identified one blocking ordering bug in the fresh-install path, one repo-convention violation (prompt string duplication), and one runtime concern about token savings being partially negated.",
  "verdict": "request_changes",
  "confidence": {
    "level": "high",
    "justification": "The install() call ordering is explicit in the changed connector-opencode/src/index.ts. registerPlugin (line ~280) and registerMcpServer (line ~283) both skip writing when no config file exists; registerPipelineAgent (line ~286) is the only method that creates a file when none exists. The test added in index.test.ts ('install creates opencode.json when no config file exists') does not assert config.plugin or config.mcp, confirming the gap is undetected. The prompt duplication is directly readable across provider.ts and index.ts in the same diff."
  },
  "comments": [
    {
      "file": "packages/connector-opencode/src/index.ts",
      "line": 286,
      "body": "**Blocking – fresh install ordering bug**: `registerPlugin` (called at ~line 280) and `registerMcpServer` (called at ~line 283) both iterate `getConfigCandidates` and `continue` when no file exists — they produce no output on a clean machine. `registerPipelineAgent` is the only method that creates `opencode.json` when no config is found, but it runs last and writes only `{ agent: { \"signet-pipeline\": {...} } }`. On a fresh OpenCode install, the MCP server entry (`config.mcp.signet`) and the plugin path entry (`config.plugin`) are silently omitted.\n\nThe test `'install creates opencode.json when no config file exists'` only asserts `config.agent[\"signet-pipeline\"]` and would not catch this regression.\n\nFix options: (a) make `registerPlugin`/`registerMcpServer` create the file if absent (matching `registerPipelineAgent`'s behavior), or (b) have a single file-creation step before any of these three methods run, or (c) remove the fallback creation from `registerPipelineAgent` and document that a config file must pre-exist.",
      "severity": "blocking"
    },
    {
      "line": 1978,
      "body": "**Warning – prompt string duplicated without a shared constant**: The system prompt `\"You are a structured data extraction system. Return ONLY valid JSON matching the requested schema. No explanations, no markdown, no code fences.\"` appears verbatim in both `PIPELINE_AGENT_CONFIG.prompt` (connector-opencode/src/index.ts) and in `body.system` here (set when `structured && structuredOutputSupported`). AGENTS.md explicitly calls out "Do not duplicate constants… across files – extract shared sources of truth." If the prompt evolves, the two copies will drift. The string should be extracted to `@signet/core` alongside `OPENCODE_PIPELINE_AGENT`.",
      "severity": "warning"
    },
    {
      "line": 1974,
      "body": "**Warning – token-savings claim may be partially negated**: The PR claims a reduction from ~47k to ~4.8k input tokens by routing through the `signet-pipeline` agent. However, when `structured && structuredOutputSupported`, the code still injects `body.system` as a per-call system override. If OpenCode forwards a per-call `system` field on top of the lightweight agent's context, some of the stripped context may be re-added. The validation table in the PR description shows 4,844 tokens which suggests this is fine in practice, but the interaction between `agent` + `system` overrides in OpenCode's protocol is not documented in the diff. A comment or a test that asserts the token count stays near the measured value would pin this.",
      "severity": "warning"
    }
  ]
}
```

[stderr]

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1971,9 +1975,14 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 			parts: [{ type: "text", text: prompt }],
 			model: { providerID, modelID },
 		};
+		if (cfg.agent) {
```
PR-Reviewer-Ant: **Blocking — no graceful degradation when `signet-pipeline` agent is absent.**

The daemon now unconditionally sends `agent: "signet-pipeline"` with every OpenCode pipeline request (when `cfg.agent` is truthy, which is the default). If a user upgrades the `signetai` meta-package but has not re-run `signet install` / `signet sync` to register the agent via the updated connector, OpenCode will receive a reference to an unknown agent. Based on the error-handling paths present in this file (404 → session retry, 422 → structured-output disable, other 4xx → throw), an agent-not-found 4xx from OpenCode will propagate as an unhandled extraction failure, silently breaking all pipeline processing until the user manually re-installs.

There is no documented upgrade step in the PR description, no in-code comment warning, and no fallback (e.g., retry the request without the `agent` field on a 400 that mentions "agent" in its body, similar to the existing structured-output fallback on 422). Suggest either: (1) adding an agent-not-found recovery path analogous to the 422/structured-output disable logic, or (2) prominently documenting the required `signet install` migration step and adding a startup warning in the daemon when the OpenCode provider is configured but the pipeline agent is not detected.

packages/daemon/src/memory-config.ts
```diff
@@ -965,7 +965,7 @@ export function loadPipelineConfig(yaml: Record<string, unknown>): PipelineV2Con
 		hints: {
 			enabled: resolveBool(hintsRaw?.enabled, undefined, d.hints?.enabled ?? true),
 			max: clampPositive(hintsRaw?.max, 1, 10, d.hints?.max ?? 5),
-			timeout: clampPositive(hintsRaw?.timeout, 5000, 120000, d.hints?.timeout ?? 30000),
+			timeout: clampPositive(hintsRaw?.timeout, 5000, 120000, d.hints.timeout),
```
PR-Reviewer-Ant: **Nitpick — inconsistent optional-chaining pattern for hints defaults.**

All other hints resolver lines in this function use the defensive `d.hints?.field ?? hardcoded_fallback` pattern:
```ts
enabled: resolveBool(hintsRaw?.enabled, undefined, d.hints?.enabled ?? true),
max: clampPositive(hintsRaw?.max, 1, 10, d.hints?.max ?? 5),
maxTokens: clampPositive(hintsRaw?.maxTokens, 64, 1024, d.hints?.maxTokens ?? 256),
poll: clampPositive(hintsRaw?.poll, 1000, 60000, d.hints?.poll ?? 5000),
```
The changed line drops both the optional chain and the fallback literal:
```ts
timeout: clampPositive(hintsRaw?.timeout, 5000, 120000, d.hints.timeout),
```
This is safe today because `hints` is a required field on `DEFAULT_PIPELINE_V2`, but it diverges from the established defensive pattern in the same block. Suggest making it consistent: `d.hints?.timeout ?? 60000`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `7fcf4c23`

[Automated Review — pr-reviewer bot] PR #505 reduces OpenCode pipeline overhead by registering a stripped-down `signet-pipeline` agent and routing all pipeline requests to it. The core approach is sound and the test coverage is solid, but there is one breaking-upgrade concern and a minor consistency deviation worth calling out.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The agent-param injection path is fully visible in the diff (provider.ts ~line 1975-1989) and the error-handling branches for OpenCode (404 retry, 422 structured-output disable, generic HTTP throw) are present in the unchanged provider.ts context. There is no branch that catches an unknown-agent error and falls back to agentless mode. The memory-config inconsistency is directly observable from the changed lines at memory-config.ts:968 compared to every adjacent hints field in the same function.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2147,6 +2157,31 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 					if (ollamaFallback) return ollamaFallback;
 					return buildOpenCodeFallbackResponse();
 				}
+				// Agent not registered — user upgraded without re-running
+				// `signet install`. Disable agent and retry without it.
+				if (agentSupported && body.includes(cfg.agent ?? "")) {
+					logger.warn("pipeline", "OpenCode rejected pipeline agent; retrying without agent", {
+						status: res.status,
+						agent: cfg.agent,
```
PR-Reviewer-Ant, line 2190: Race condition: `agentSupported` and `sessionId` are `let` variables shared across all concurrent `generate()` calls within the same provider instance. When two concurrent pipeline jobs both receive a 400 agent-not-found response simultaneously:
1. Both read `agentSupported = true` and enter this branch.
2. Both set `sessionId = null` and call `createSession()`, receiving different session IDs.
3. Both write their new session ID back to the shared `sessionId`; whichever call resolves second clobbers the first.
4. The first call then posts its retry message to the wrong (second call's) session.

This is a real scenario: the daemon routinely runs multiple concurrent extraction workers. The fix is to use a single-flight guard or serialize the reset: e.g., set `agentSupported = false` immediately and early-exit concurrent in-flight retries, or introduce a one-shot atomic CAS idiom (`if (agentSupported) { agentSupported = false; ... }`). The double-write to `sessionId` also needs to be guarded — the session reset should happen under the same mutual-exclusion window as the agent flag flip.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2147,6 +2157,31 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 					if (ollamaFallback) return ollamaFallback;
 					return buildOpenCodeFallbackResponse();
 				}
+				// Agent not registered — user upgraded without re-running
+				// `signet install`. Disable agent and retry without it.
```
PR-Reviewer-Ant, line 2183: `body.includes(cfg.agent ?? "")` is a loose heuristic for agent-not-found detection. Any 4xx response whose body happens to contain the string `"signet-pipeline"` — including a prompt that references the agent by name echoed back in an error, or an unrelated validation message — will silently trigger the agent-disable-and-retry path. The existing codebase is deliberately precise about this class of detection (see the structured-output 422 path which requires `"format"` in both the JSON key and the HTTP path). A tighter match (e.g., `body.includes('unknown agent')`) that targets the specific error string OpenCode emits would reduce false-positive risk. The concern is sharpened because a false positive would permanently disable the agent for the lifetime of the provider instance (`agentSupported` never resets to `true`), silently degrading every subsequent call.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -715,6 +715,92 @@ describe("createOpenCodeProvider", () => {
...
+		});
+		const result = await provider.generate("extract this");
+
+		expect(result).toBe("recovered");
+		expect(retryBody.agent).toBeUndefined();
+	});
+
 	it("generate() omits format field when enableStructuredOutput is false", async () => {
 		let capturedBody: Record<string, unknown> = {};
 		mockFetch(async (url, init) => {
```
PR-Reviewer-Ant, line 869: The `"generate() retries without agent on agent-not-found 4xx"` test validates only a single-call retry. It does not verify that `agentSupported` remains `false` on subsequent calls (i.e., that the agent is permanently disabled for this provider instance after one rejection). A follow-up call that re-attempts with the agent would re-trigger the 400 loop. Adding a second `generate()` call in the same test and asserting `retryBody.agent` is still `undefined` would pin this behavior.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `93174d41`

[Automated Review — pr-reviewer bot] PR #505 introduces a pipeline-agent registration mechanism to reduce OpenCode session token overhead. One blocking race condition was found in the retry path of `provider.ts`; one warning-level heuristic fragility in agent-not-found detection.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The race condition is directly provable from the diff: `agentSupported` and `sessionId` are `let` closure variables shared across concurrent `generate()` calls. The mutation sequence (lines ~2165-2170 in provider.ts) is not serialized. The fragile detection check on the same lines is also directly readable in the diff.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2147,6 +2179,32 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 					if (ollamaFallback) return ollamaFallback;
```
PR-Reviewer-Ant, line 2179: **Race condition in concurrent agent-not-found retry.**

Two concurrent `generate()` calls can both build their message bodies while `agentSupported` is still `true` (that happens synchronously before any `await`). When both receive a 400 with an agent-rejection body:

1. Call A enters the `if (agentSupported && isAgentRejection(...))` block first, sets `agentSupported = false`, retries → succeeds.
2. Call B arrives at the same `if` after `agentSupported` is already `false` → condition is `false` → falls through to `throw new Error('OpenCode HTTP 400: ...')`.  

Call B throws an error even though the request it originally sent still carried `agent` (built before the flag flipped). The comment notes the guard but the guard only prevents *entering the retry block twice*; it does not prevent the concurrent sibling from throwing on its already-in-flight 400 response.

Since extraction workers retry failed jobs, this is not catastrophic, but the first upgrade-scenario call that races will produce an unnecessary job failure and a confusing error log. A minimal fix: after the outer `if (!res.ok)` check, re-derive `agentSupported` from a per-request snapshot so a late-arriving 400 from an already-disabled agent is treated the same as a 400 after disablement (i.e. log a warn and let Ollama fallback handle it rather than throwing).

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1710,6 +1714,28 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+ */
+function isAgentRejection(body: string, agent: string | undefined): boolean {
+	if (!agent) return false;
+	try {
+		const parsed: unknown = JSON.parse(body);
+		if (isRecord(parsed)) {
+			const issues = Array.isArray(parsed.issues) ? parsed.issues : [];
+			if (issues.some((i): boolean => isRecord(i) && Array.isArray(i.path) && i.path[0] === "agent")) {
+				return true;
+			}
```
PR-Reviewer-Ant, line 1730: **`isAgentRejection` fallback substring check is over-broad.**

```typescript
return lower.includes("unknown agent") || (lower.includes("agent") && lower.includes("not found"));
```

The second branch matches any 4xx body that contains both "agent" and "not found" — e.g. `"MCP agent config not found"`, `"agent credentials not found"`, or any error from a proxy that embeds both words. A false positive permanently disables the pipeline agent for this provider *instance's* lifetime (`agentSupported` is never reset). That means a noisy MCP error or transient 400 from a gateway could silently degrade all subsequent extraction calls to run without the optimized agent.

Suggestion: tighten the substring fallback to `lower.includes("unknown agent")` alone (which is the exact string OpenCode returns per the test mock), and drop the broader `agent && not found` branch unless there is a real-world response body that requires it. The structured Zod-path check above is already precise for JSON responses.

packages/connector-opencode/src/index.ts
```diff
@@ -524,6 +529,69 @@ export class OpenCodeConnector extends BaseConnector {
...
+
+			let config: JsonObject;
+			try {
+				config = parseJsonOrJsonc(readFileSync(configPath, "utf-8"));
+			} catch {
+				continue;
+			}
+
+			if (!isJsonObject(config.agent)) continue;
+
```
PR-Reviewer-Ant, line 571: **`removePipelineAgent` uses `_` and `__` for unused destructuring targets.**

```typescript
const { [OPENCODE_PIPELINE_AGENT]: _, ...rest } = agents;
// ...
const { agent: __, ...configWithoutAgent } = config;
```

Biome's `noUnusedVariables` rule (enabled in this repo) typically flags single-underscore prefixed names unless the rule is explicitly configured to allow them. The PR checklist has `bun run lint` unchecked — if Biome rejects these, CI will fail. Standard workaround is `_unused` prefix or using `void` pattern; confirm Biome config allows bare `_` before merge.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2147,6 +2179,32 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+				if (agentSupported && isAgentRejection(body, cfg.agent)) {
+					agentSupported = false;
+					logger.warn("pipeline", "OpenCode rejected pipeline agent; retrying without agent", {
+						status: res.status,
+						agent: cfg.agent,
+					});
+					const retrySid = await createSession();
+					const retryRes = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
+						method: "POST",
+						headers: { "Content-Type": "application/json" },
```
PR-Reviewer-Ant, line 2195: **Retry reuses the original `controller.signal` from the timed-out outer request.**

If the initial 400 arrives late in the timeout window, the retry fetch shares the same `AbortController`. Any wall-clock time spent waiting for the first response is already charged against the budget. Under load this is likely fine, but worth noting: if the first request takes `T` ms before returning 400, the retry has at most `defaultTimeoutMs - T` ms. With `defaultTimeoutMs` at 60 s this is generally safe, but a tight per-job timeout could cause the retry to abort immediately after starting.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `fa7025a1`

[Automated Review — pr-reviewer bot] PR #505 introduces a dedicated `signet-pipeline` agent to reduce OpenCode per-session token overhead. The implementation is well-structured, tests cover the happy path, but three issues warrant attention: a concurrency gap in the agent-not-found retry path, an overly broad fallback heuristic in `isAgentRejection`, and unchecked lint/typecheck boxes that may mask build failures.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The race condition is directly provable from the diff: `agentSupported` is a closure variable shared across concurrent calls to `generate()`. The specific code paths (provider.ts lines around `agentSupported = false` and `buildMessageBody`) show the window. The `isAgentRejection` false-positive risk is directly visible in the fallback substring logic. The unchecked checklist boxes are stated in the PR description.


- packages/connector-opencode/index.test.ts:1 - **PR checklist has `bun run typecheck` and `bun run lint` unchecked.**

AGENTS.md requires both to pass before review. The description explicitly leaves these boxes unchecked. TypeScript `noUnusedLocals` would flag the same `_`/`__` variables noted above; Biome may also flag them. Please confirm these pass (or explain why they are intentionally skipped) before merge.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -1710,6 +1714,28 @@ function isRecord(value: unknown): value is Record<string, unknown> {
...
+		}
+	} catch {
+		// empty
+	}
+	const lower = body.toLowerCase();
+	return lower.includes("unknown agent");
+}
+
 function parseOpenCodeTokens(value: unknown): OpenCodeTokens | undefined {
 	if (!isRecord(value)) return undefined;
```
PR-Reviewer-Ant, line 1741: **Warning: Overly broad agent-rejection heuristic can permanently disable the pipeline agent on unrelated errors.**

The fallback path in `isAgentRejection` does a case-insensitive substring match for `"unknown agent"` against the entire response body. If OpenCode ever returns a 4xx whose body happens to contain that phrase for an unrelated reason (e.g. `"Unknown agent configuration field 'timeout'"`, or a future error message change), `agentSupported` is set to `false` for the lifetime of the provider instance and all subsequent calls run without the agent, silently regressing to the ~47k-token path.

The Zod-path check (`i.path[0] === "agent"`) is precise and covers the structured error case. The substring fallback should at minimum also validate that the agent name appears in the body, e.g.:
```ts
return lower.includes("unknown agent") && lower.includes(agent.toLowerCase());
```
This keeps the fallback narrow to rejections that name the specific agent being sent.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: claude-sonnet-4-6 | commit: `025ddf5e`

[Automated Review — pr-reviewer bot] PR #505 adds a dedicated `signet-pipeline` OpenCode agent to cut pipeline token overhead. The core approach is sound and the implementation matches the stated goals. Three issues are worth discussing before merge: a broad error-body substring check that could permanently disable the agent on unrelated errors, an orphaned session on the agent-not-found retry failure path, and a misleading test description for the concurrent-sibling scenario. Two PR-readiness boxes (typecheck, lint) are unchecked.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The agent-rejection detection logic and session-reuse state are fully visible in the diff and changed file contents. The retry fallback paths can be traced end-to-end. Confidence is medium rather than high because the full `createOpenCodeProvider` closure (prior to line ~1991) was truncated in the context, so exact interactions between the new agent-rejection branches and the pre-existing 422/structured-output branches could not be verified by line inspection alone.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2147,6 +2179,43 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+					});
+					const retrySid = await createSession();
+					const retryRes = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
+						method: "POST",
+						headers: { "Content-Type": "application/json" },
+						body: buildMessageBody(prompt, true),
+						signal: controller.signal,
+					});
+					if (retryRes.ok) {
+						sessionId = retrySid;
```
PR-Reviewer-Ant, line 2200: **Warning: Orphaned session on agent-not-found retry failure.**

In the agent-rejection retry block, `retrySid` is created unconditionally via `createSession()`. `sessionId` (the shared reuse variable) is only updated to `retrySid` when `retryRes.ok`. If the retry itself fails (non-2xx), `retrySid` is abandoned: the closure continues to hold the old (rejected) `sessionId`, and `retrySid` is never stored or cleaned up. Subsequent calls using the old `sessionId` will likely get a 404 and trigger another session creation, which works, but the orphaned `retrySid` session lingers server-side.

Consider setting `sessionId = undefined` (or to `retrySid`) before the fallback so the next call creates a clean session rather than reusing the pre-rejection `sessionId`.

packages/daemon/src/pipeline/provider.test.ts
```diff
@@ -715,6 +715,161 @@ describe("createOpenCodeProvider", () => {
...
+			}
+			// After agent is disabled, still return 400 with agent-rejection
+			// body to simulate a concurrent in-flight request that was built
+			// before the flag flipped.
+			if (!agentSeen) {
+				return new Response(`unknown agent "signet-pipeline"`, { status: 400 });
+			}
+			return Response.json(openCodeResponse("ok"));
+		});
+
```
PR-Reviewer-Ant, line 832: **Nitpick: The concurrent-sibling test does not reproduce the described race.**

The test comment says it simulates "a request built before `agentSupported` was flipped", but because `agentSupported` is already `false` when the second `generate()` call is made, `buildMessageBody` will omit the `agent` field from the outgoing request body. The mock then returns a 400 with agent-rejection text regardless of body content. This tests a slightly different scenario: a response with agent-rejection text arriving for a request that never included the agent.

The true concurrent scenario (request built WITH agent, flag flipped before response arrives) cannot happen in a single-threaded test without interleaving async operations at the fetch layer. The current test is still useful for its coverage of the `!agentSupported && isAgentRejection` branch, but the comment and assertion setup are misleading. Consider updating the comment to describe what is actually being tested.

packages/connector-opencode/src/index.ts
```diff
@@ -524,6 +529,69 @@ export class OpenCodeConnector extends BaseConnector {
 		}
```
PR-Reviewer-Ant, line 529: **Nitpick: Unused destructuring variables `_` and `__` may cause Biome lint failures.**

```ts
const { [OPENCODE_PIPELINE_AGENT]: _, ...rest } = agents;
// ...
const { agent: __, ...configWithoutAgent } = config;
```

Biome's `noUnusedVariables` rule can flag single-underscore names unless the project has an explicit `noUnusedVariables.ignoreParameters` pattern configured for `_`. The `lint` checkbox in the PR readiness checklist is unchecked. If Biome is not configured to allow `_`/`__` as ignored, this will fail CI. Verify locally with `bun run lint` before merge.

packages/connector-opencode/index.test.ts
```diff
@@ -60,3 +71,151 @@ describe("OpenCodeConnector.install — legacy SIGNET block migration", () => {
 		expect(result.filesWritten).toHaveLength(0);
```
PR-Reviewer-Ant, line 71: **Process gap: `bun run typecheck` and `bun run lint` boxes unchecked in the PR description.**

The PR readiness checklist shows both `bun run typecheck` and `bun run lint` as not completed. Given that this PR introduces new TypeScript generics, `as const` casts, and destructuring patterns across multiple packages, unverified typecheck creates a risk of silent type errors (e.g., the `JsonObject` cast on `config.agent`, or the new `readonly agent?: string` field on `OpenCodeProviderConfig` propagating to callers). Please run both before merge.

NicholaiVogel:
Looks good from my end! nice work. merging this.

NicholaiVogel (approved):

PR #514, opened 2026-04-16
refactor(cli): prompt for sync after restart

NicholaiVogel:
## Summary

- Remove the restart-time prompt for restarting a connected OpenClaw instance.
- Add an interactive restart prompt for running `signet sync` instead.
- Replace `--no-openclaw` with `--no-sync` on `signet daemon restart` and the `signet restart` alias.
- Reuse the same sync implementation behind the top-level `signet sync` command.
- Add regression coverage for accepting, declining, and bypassing the sync prompt.

## Why

The restart dialog should guide users toward refreshing Signet hooks, extensions, built-in templates, and skills after the daemon restarts, rather than trying to manage a connected OpenClaw process directly.

## Validation

- `bun test packages/cli/src/features/daemon.test.ts`
- `bunx biome check packages/cli/src/commands/shared.ts packages/cli/src/commands/daemon.ts packages/cli/src/features/daemon.ts packages/cli/src/features/daemon.test.ts packages/cli/src/cli.ts`
- `bun install --frozen-lockfile`

## Known validation blocker

`cd packages/cli && bun run build:cli` is currently blocked by unresolved existing workspace imports for `@signet/connector-hermes-agent` and `@signet/connector-pi`. Those imports pre-existed this change.

### DIFF
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -737,7 +737,13 @@ function embeddingProvider(basePath: string): "native" | "llama-cpp" | "ollama" 
 			const direct = parsed.embedding;
 			if (isRecord(direct) && typeof direct.provider === "string") {
 				const provider = direct.provider;
-				if (provider === "native" || provider === "llama-cpp" || provider === "ollama" || provider === "openai" || provider === "none") {
+				if (
+					provider === "native" ||
+					provider === "llama-cpp" ||
+					provider === "ollama" ||
+					provider === "openai" ||
+					provider === "none"
+				) {
 					return provider;
 				}
 			}
@@ -746,7 +752,13 @@ function embeddingProvider(basePath: string): "native" | "llama-cpp" | "ollama" 
 				const nested = mem.embeddings;
 				if (isRecord(nested) && typeof nested.provider === "string") {
 					const provider = nested.provider;
-					if (provider === "native" || provider === "llama-cpp" || provider === "ollama" || provider === "openai" || provider === "none") {
+					if (
+						provider === "native" ||
+						provider === "llama-cpp" ||
+						provider === "ollama" ||
+						provider === "openai" ||
+						provider === "none"
+					) {
 						return provider;
 					}
 				}
@@ -754,7 +766,13 @@ function embeddingProvider(basePath: string): "native" | "llama-cpp" | "ollama" 
 			const legacy = parsed.embeddings;
 			if (isRecord(legacy) && typeof legacy.provider === "string") {
 				const provider = legacy.provider;
-				if (provider === "native" || provider === "llama-cpp" || provider === "ollama" || provider === "openai" || provider === "none") {
+				if (
+					provider === "native" ||
+					provider === "llama-cpp" ||
+					provider === "ollama" ||
+					provider === "openai" ||
+					provider === "none"
+				) {
 					return provider;
 				}
 			}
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the refactor and cleanup work in `refactor(cli): prompt for sync after restart` (commit `32419571`) now and I'll follow up shortly with feedback.

packages/cli/src/commands/daemon.ts
```diff
@@ -31,7 +31,7 @@ export function registerDaemonCommands(program: Command, deps: DaemonDeps): void
 	const restart = daemonCmd
 		.command("restart")
 		.description("Restart the daemon")
-		.option("--no-openclaw", "Skip OpenClaw restart prompt")
+		.option("--no-sync", "Skip signet sync prompt")
```
PR-Reviewer-Ant, line 34: Replacing `--no-openclaw` with `--no-sync` is a user-visible breaking CLI change for existing scripts/automation. There is no compatibility alias, deprecation path, or migration warning in this diff. Consider temporarily accepting both flags (mapping old -> new) and warning on the legacy flag, or document this as an explicit breaking change in release notes.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `32419571`

[Automated Review by pr-reviewer] The PR largely implements the stated behavior change (OpenClaw restart prompt removed, sync prompt added, shared sync path reused), but it introduces a CLI compatibility break that should be explicitly handled or documented.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed option definitions in commands/daemon.ts and the RestartOptions/doRestart flow in features/daemon.ts directly show the old --no-openclaw flag was removed and replaced by --no-sync with no compatibility shim. The rest of the claimed behavior is directly implemented and test-covered in daemon.test.ts.

packages/cli/src/cli.ts
```diff
@@ -1250,6 +1268,19 @@ const healthDeps = {
 	signetLogo,
```
PR-Reviewer-Ant, line 1268: Blocking: restart-time sync is bound to global `AGENTS_DIR` instead of the restart target path. `doRestart` resolves and uses `--path` for daemon stop/start, but `deps.syncTemplates()` has no way to receive that `basePath`, so accepting the prompt can sync a different workspace than the one just restarted. This can mutate the wrong agent files (hooks/templates/skills) and leave the intended workspace unsynced.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `062e18b6`

[Automated Review] I found one blocking correctness issue where restart-time sync can target the wrong workspace path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - `doRestart` restarts using the resolved `--path` basePath, but the newly injected `syncTemplates` dependency is wired in `cli.ts` as `runSyncTemplates()` with no path parameter and hardcoded `agentsDir: AGENTS_DIR`. This creates a concrete path mismatch when `signet daemon restart --path <custom>` is used.

packages/cli/src/commands/daemon.ts
NicholaiVogel: Addressed in `718910e6`.

`--no-openclaw` is now accepted as a deprecated compatibility alias for `--no-sync` on both `signet daemon restart` and `signet restart`. When used, it prints a deprecation warning and skips the sync prompt, preserving the old automation behavior while pointing users to the new flag.

Validation: `bun test packages/cli/src/features/daemon.test.ts` and focused `biome check` on the touched CLI files.

packages/cli/src/cli.ts
NicholaiVogel: Addressed in `718910e6`.

Good catch. Restart-time sync now receives the resolved restart `basePath`, so `signet daemon restart --path <custom>` restarts and syncs the same workspace instead of falling back to global `AGENTS_DIR`. The top-level `signet sync` command still defaults to `AGENTS_DIR`.

Added regression coverage with `restart sync uses the resolved restart path`.

Validation: `bun test packages/cli/src/features/daemon.test.ts` and focused `biome check` on the touched CLI files.

PR #516, opened 2026-04-17
perf(knowledge-graph): paginate entity IDs before counting to avoid full GROUP BY

### LINKED ISSUE #515: perf: knowledge endpoints time out on large graphs (LIMIT after GROUP BY materialization)
## Summary

Four knowledge endpoints time out at 30–60s on workspaces with large graphs:

- `GET /api/knowledge/entities`
- `GET /api/knowledge/entities/:id`
- `GET /api/knowledge/stats`
- `POST /api/repair/prune-singleton-entities`

This also breaks `signet dashboard` (constellation view and cortex panel never load), `signet doctor` (hangs indefinitely), and any downstream consumer with a reasonable client-side timeout. The `@signetai/signet-memory-openclaw` MCP tool uses a 10 s timeout and ends up returning zero results whenever graph traversal in the hybrid recall path is enabled.

## Environment

- signet 0.99.3 (also reproduced on 0.98.4)
- SQLite 3.43.2
- `memories.db` 913 MB, 27,484 active memories, 60,399 entities, 198,534 edges, all under `agent_id='default'`

## Repro

Direct sqlite3 CLI run of the query shape that `listKnowledgeEntities` executes, with `LIMIT 3`:

```
sqlite3 ~/.agents/memory/memories.db "
  SELECT e.id, COUNT(DISTINCT asp.id), COUNT(DISTINCT attr.id), COUNT(DISTINCT dep.id)
  FROM entities e
  LEFT JOIN entity_aspects asp ON asp.entity_id=e.id AND asp.agent_id=e.agent_id
  LEFT JOIN entity_attributes attr ON attr.aspect_id=asp.id AND attr.agent_id=e.agent_id
  LEFT JOIN entity_dependencies dep ON dep.agent_id=e.agent_id
    AND (dep.source_entity_id=e.id OR dep.target_entity_id=e.id)
  WHERE e.agent_id='default'
  GROUP BY e.id
  ORDER BY e.pinned DESC, e.pinned_at DESC, e.mentions DESC, e.updated_at DESC, e.name ASC
  LIMIT 3;"
```

Wall time: **2 m 15 s** (55 s user, 45 s system, 77 % CPU). `ANALYZE` was run — the planner picks the correct indices (`idx_entity_aspects_entity`, `idx_entity_attributes_aspect`, MULTI-INDEX OR on dependency source/target), and it still takes 2 m 15 s.

HTTP calls from the daemon mirror this:

| Endpoint | Result |
|---|---|
| `GET /api/knowledge/entities?limit=3` | 30 s curl timeout, status 000 |
| `GET /api/knowledge/stats` | 30 s curl timeout, status 000 |
| `GET /api/knowledge/constellation?agent_id=default` | 60 s curl timeout, status 000 |
| `POST /api/repair/prune-singleton-entities {dryRun:true}` | 60 s curl timeout, status 000 |

Unaffected endpoints (confirm the daemon itself is healthy): `/health`, `/api/status`, `/api/pipeline/status`, `/api/memory/recall` (~2–4 s when workers are idle).

## Root cause

`packages/daemon/src/knowledge-graph.ts#listKnowledgeEntities` (and the mirror queries in `getKnowledgeEntityDetail` and `getKnowledgeStats`) materialize the full `GROUP BY e.id` over every entity in the agent scope before the `ORDER BY` and `LIMIT` can apply. `EXPLAIN QUERY PLAN` confirms `USE TEMP B-TREE FOR GROUP BY` followed by `USE TEMP B-TREE FOR ORDER BY`, and the aggregate counts force SQLite to touch every joined row.

`getKnowledgeStats` has a related shape: the `scopedMemoryRows` subquery runs `SELECT COUNT(DISTINCT m.id) FROM memories m WHERE EXISTS (SELECT 1 FROM memory_entity_mentions JOIN entities …)`, which drives from the memories side and was still running after 6 minutes when I killed it locally.

`pruneSingletonExtractedEntities` (`packages/daemon/src/repair-actions.ts`) shares the pain: its candidate `SELECT` scans all 60k entities with three correlated `NOT EXISTS` subqueries, and the `entity_type='extracted' AND mentions<=?` filter has no index support.

## Proposed fix (coming as a PR)

1. **`listKnowledgeEntities` / `getKnowledgeEntityDetail`**: paginate entity IDs first via a CTE using the ordering columns, then compute counts via scalar subqueries per row. Only `LIMIT` rows ever touch the aspect / attribute / dependency joins.
2. **`getKnowledgeStats.scopedMemoryRows`**: rewrite as a three-way JOIN driven from the (narrower) agent-scoped entities side. Same result set, smaller search space.
3. **New migration**: compound index `(agent_id, pinned DESC, pinned_at DESC, mentions DESC, updated_at DESC, name)` to serve the CTE as an index-only scan, plus a partial index `(entity_type, mentions) WHERE entity_type='extracted'` to narrow the prune candidate scan.
4. **`pruneSingletonExtractedEntities`**: no query change; the partial index changes its scan plan.

Expected post-fix latency (local bench target):

| Endpoint | Before | After (target) |
|---|---|---|
| `GET /api/knowledge/entities?limit=3` | timeout | < 500 ms |
| `GET /api/knowledge/stats` | timeout | < 1 s |
| `POST /api/repair/prune-singleton-entities {dryRun:true}` | timeout | < 5 s |

## Severity / impact

- **Data safety**: none. Read-only regression; the only write is adding two indices in a new migration.
- **UX**: severe. Dashboard / doctor unusable on non-trivial workspaces. Gets worse as the graph grows.
- **Workaround while unfixed**: pause all pipeline workers (`memory.pipelineV2.paused: true` in `agent.yaml`) to free CPU, disable `memory.pipelineV2.traversal.enabled` and `memory.pipelineV2.reranker.enabled` so the recall path skips the same slow joins. This keeps `memory_search` under MCP's 10 s client timeout but prevents extraction from draining and still doesn't fix the dashboard.

## Out of scope (noted for follow-ups)

- Graph traversal boost path in `memory_search` exercises the same query family at a bigger surface; a separate PR will likely be needed.
- Extraction queue stall detection (`claude-code` provider silently timing out for days) — operational concern, separate issue.
- Agent auto-registration when a sessionKey (`agent:main:*`) references an agent that isn't in the `agents` roster — the read policy falls back to `isolated` and the session sees no shared memories; worth a separate UX improvement.

---

Happy to discuss scope / split into multiple PRs if preferred.

Assisted-by: Claude Opus 4.6 (1M context)

ddasgupta4:
## Summary

- Rewrite `listKnowledgeEntities`, `getKnowledgeEntityDetail`, and `getKnowledgeStats` to avoid materializing `GROUP BY` across all entities before `LIMIT` applies
- Add migration 058 with two indices (`idx_entities_order` + `idx_entities_extracted_mentions`)
- Add 9 regression tests covering ordering, counts, pagination, type/query filters, agent scoping, and soft-delete handling

## Why

On workspaces with large graphs (60k+ entities, 200k+ edges), the existing `LEFT JOIN + GROUP BY + ORDER BY + LIMIT` query pattern forces SQLite to materialize and sort every entity in the agent scope before applying the limit. Measured locally:

| Endpoint | Before | Target |
|---|---|---|
| `GET /api/knowledge/entities?limit=3` | **2 min 15 s** (wall) | < 500 ms |
| `GET /api/knowledge/stats` | **> 6 min** (killed) | < 1 s |
| `POST /api/repair/prune-singleton-entities` | **> 10 min** (killed) | < 5 s |

This breaks `signet dashboard` (constellation + cortex never load), `signet doctor` (hangs), and any downstream consumer with a reasonable client-side timeout — including the `@signetai/signet-memory-openclaw` MCP tool which uses a 10 s timeout and returns empty when graph traversal boost is enabled.

Closes #515

### `packages/daemon/src/knowledge-graph.ts`

- **`listKnowledgeEntities`**: paginate entity IDs first via a CTE using the compound ordering index, then compute counts via scalar subqueries per row. Only `LIMIT` rows ever touch the aspect / attribute / dependency tables.
- **`getKnowledgeEntityDetail`**: drop `GROUP BY` in favour of scalar subqueries. Already filtered to a single `e.id`, but gains consistency with the new pattern and avoids the LEFT JOIN cross-product.
- **`getKnowledgeStats`**: rewrite the `scopedMemoryRows` subquery from a correlated `EXISTS (SELECT 1 FROM memory_entity_mentions JOIN entities)` to a direct three-way JOIN driven from the narrower agent-scoped entities side.

### `packages/core/src/migrations/058-knowledge-graph-indices.ts`

- `idx_entities_order` — compound index `(agent_id, pinned DESC, pinned_at DESC, mentions DESC, updated_at DESC, name)` that serves the page CTE as an index-only scan.
- `idx_entities_extracted_mentions` — partial index `(entity_type, mentions) WHERE entity_type = 'extracted'` that narrows the `pruneSingletonExtractedEntities` candidate scan.

Both are idempotent (`IF NOT EXISTS`).

### `packages/daemon/src/knowledge-graph-list.test.ts` (new)

9 regression tests:
- Ordering: pinned > mentions > updated_at > name
- Counts: aspects, active attributes, active constraints, dependencies (incoming + outgoing)
- Pagination: limit + offset
- Filters: entity_type and canonical_name LIKE
- Agent scoping: entities isolated by agent_id
- Entity detail: incoming vs outgoing dependency split, null for unknown ID
- Stats: scoped memory count via mentions JOIN, soft-delete exclusion

## Type

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`

## Screenshots

N/A — backend-only query rewrite, no UI changes.

## PR Readiness (MANDATORY)

Migration 058 adds two idempotent indices (`CREATE INDEX IF NOT EXISTS`). No schema changes, no data mutations. Safe to reapply. Runs on daemon startup via the standard migration runner.

- [x] Migration is idempotent

**Rust parity**: N/A — the Rust predictor sidecar does not query entity tables; these indices are consumed only by the TypeScript daemon.

**Rollback**: both indices can be dropped without data loss (`DROP INDEX IF EXISTS idx_entities_order; DROP INDEX IF EXISTS idx_entities_extracted_mentions;`). The pre-patch queries still work without them, just slower.

## Testing

- All 9 new tests pass (`bun test packages/daemon/src/knowledge-graph-list.test.ts`)
- All 53 existing knowledge-related tests pass (`knowledge-feedback.test.ts`, `knowledge-expand-api.test.ts`, `repair-actions.test.ts`)
- `bun run typecheck` clean across all packages
- `bun biome check` clean on all changed files
- 46 pre-existing failures in `packages/daemon/` (worker / synthesis / embedding tests that require env-specific providers) — unchanged by this PR, verified by running same tests on `upstream/main`

## AI disclosure

This PR was authored with assistance from Claude Opus 4.6 (1M context).

### DIFF
--- a/packages/daemon/src/knowledge-graph.ts
+++ b/packages/daemon/src/knowledge-graph.ts
@@ -778,6 +778,7 @@ export function listKnowledgeEntities(
 						SELECT COUNT(*) FROM entity_attributes attr
 						JOIN entity_aspects asp ON asp.id = attr.aspect_id
 						WHERE asp.entity_id = e.id
+						  AND asp.agent_id = e.agent_id
 						  AND attr.agent_id = e.agent_id
 						  AND attr.kind = 'attribute'
 						  AND attr.status = 'active'
@@ -786,6 +787,7 @@ export function listKnowledgeEntities(
 						SELECT COUNT(*) FROM entity_attributes attr
 						JOIN entity_aspects asp ON asp.id = attr.aspect_id
 						WHERE asp.entity_id = e.id
+						  AND asp.agent_id = e.agent_id
 						  AND attr.agent_id = e.agent_id
 						  AND attr.kind = 'constraint'
 						  AND attr.status = 'active'
@@ -833,6 +835,7 @@ export function getKnowledgeEntityDetail(
 						SELECT COUNT(*) FROM entity_attributes attr
 						JOIN entity_aspects asp ON asp.id = attr.aspect_id
 						WHERE asp.entity_id = e.id
+						  AND asp.agent_id = e.agent_id
 						  AND attr.agent_id = e.agent_id
 						  AND attr.kind = 'attribute'
 						  AND attr.status = 'active'
@@ -841,6 +844,7 @@ export function getKnowledgeEntityDetail(
 						SELECT COUNT(*) FROM entity_attributes attr
 						JOIN entity_aspects asp ON asp.id = attr.aspect_id
 						WHERE asp.entity_id = e.id
+						  AND asp.agent_id = e.agent_id
 						  AND attr.agent_id = e.agent_id
 						  AND attr.kind = 'constraint'
 						  AND attr.status = 'active'
--- a/packages/daemon/src/knowledge-graph.ts
+++ b/packages/daemon/src/knowledge-graph.ts
@@ -755,30 +755,49 @@ export function listKnowledgeEntities(
 			args.push(`%${params.query.trim().toLowerCase()}%`);
 		}
 
+		// Paginate entity IDs first, then compute counts only for the page.
+		// This avoids materializing GROUP BY + ORDER BY across every entity in
+		// the agent scope before LIMIT can apply, which is prohibitive on graphs
+		// with tens of thousands of entities. See Signet-AI/signetai#515.
 		const rows = db
 			.prepare(
-				`SELECT
+				`WITH page AS (
+					SELECT e.id
+					FROM entities e
+					WHERE ${conditions.join(" AND ")}
+					ORDER BY e.pinned DESC, e.pinned_at DESC, e.mentions DESC, e.updated_at DESC, e.name ASC
+					LIMIT ? OFFSET ?
+				)
+				SELECT
 					e.*,
-					COUNT(DISTINCT asp.id) AS aspect_count,
-					COUNT(DISTINCT CASE
-						WHEN attr.kind = 'attribute' AND attr.status = 'active' THEN attr.id
-					END) AS attribute_count,
-					COUNT(DISTINCT CASE
-						WHEN attr.kind = 'constraint' AND attr.status = 'active' THEN attr.id
-					END) AS constraint_count,
-					COUNT(DISTINCT dep.id) AS dependency_count
-				 FROM entities e
-				 LEFT JOIN entity_aspects asp
-				   ON asp.entity_id = e.id AND asp.agent_id = e.agent_id
-				 LEFT JOIN entity_attributes attr
-				   ON attr.aspect_id = asp.id AND attr.agent_id = e.agent_id
-				 LEFT JOIN entity_dependencies dep
-				   ON dep.agent_id = e.agent_id
-				  AND (dep.source_entity_id = e.id OR dep.target_entity_id = e.id)
-				 WHERE ${conditions.join(" AND ")}
-				 GROUP BY e.id
-				 ORDER BY e.pinned DESC, e.pinned_at DESC, e.mentions DESC, e.updated_at DESC, e.name ASC
-				 LIMIT ? OFFSET ?`,
+					(
+						SELECT COUNT(*) FROM entity_aspects asp
+						WHERE asp.entity_id = e.id AND asp.agent_id = e.agent_id
+					) AS aspect_count,
+					(
+						SELECT COUNT(*) FROM entity_attributes attr
+						JOIN entity_aspects asp ON asp.id = attr.aspect_id
+						WHERE asp.entity_id = e.id
+						  AND attr.agent_id = e.agent_id
+						  AND attr.kind = 'attribute'
+						  AND attr.status = 'active'
+					) AS attribute_count,
+					(
+						SELECT COUNT(*) FROM entity_attributes attr
+						JOIN entity_aspects asp ON asp.id = attr.aspect_id
+						WHERE asp.entity_id = e.id
+						  AND attr.agent_id = e.agent_id
+						  AND attr.kind = 'constraint'
+						  AND attr.status = 'active'
+					) AS constraint_count,
+					(
+						SELECT COUNT(*) FROM entity_dependencies dep
+						WHERE dep.agent_id = e.agent_id
+						  AND (dep.source_entity_id = e.id OR dep.target_entity_id = e.id)
+					) AS dependency_count
+				 FROM page p
+				 JOIN entities e ON e.id = p.id
+				 ORDER BY e.pinned DESC, e.pinned_at DESC, e.mentions DESC, e.updated_at DESC, e.name ASC`,
 			)
 			.all(...args, params.limit, params.offset) as Array<Record<string, unknown>>;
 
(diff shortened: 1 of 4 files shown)


PR-Reviewer-Ant:
Hi @ddasgupta4 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the changes in `perf(knowledge-graph): paginate entity IDs before counting to avoid full GROUP BY` (commit `591ec888`) now and I'll follow up shortly with feedback.

packages/daemon/src/knowledge-graph.ts
```diff
@@ -755,30 +755,49 @@ export function listKnowledgeEntities(
...
-				 ORDER BY e.pinned DESC, e.pinned_at DESC, e.mentions DESC, e.updated_at DESC, e.name ASC
-				 LIMIT ? OFFSET ?`,
+					(
+						SELECT COUNT(*) FROM entity_aspects asp
+						WHERE asp.entity_id = e.id AND asp.agent_id = e.agent_id
+					) AS aspect_count,
+					(
+						SELECT COUNT(*) FROM entity_attributes attr
+						JOIN entity_aspects asp ON asp.id = attr.aspect_id
+						WHERE asp.entity_id = e.id
```
PR-Reviewer-Ant, line 780: Blocking: agent scoping is incomplete in the new scalar subqueries for `attribute_count` / `constraint_count` in `listKnowledgeEntities`. The JOIN only checks `asp.id = attr.aspect_id` and `asp.entity_id = e.id`, but omits `asp.agent_id = e.agent_id`. This can overcount across agents if aspect rows are cross-scoped or inconsistent. The previous GROUP BY query explicitly scoped aspects by agent. Add `AND asp.agent_id = e.agent_id` in both subqueries.

packages/daemon/src/knowledge-graph.ts
```diff
@@ -798,33 +817,44 @@ export function getKnowledgeEntityDetail(
...
+						  AND attr.agent_id = e.agent_id
+						  AND attr.kind = 'attribute'
+						  AND attr.status = 'active'
+					) AS attribute_count,
+					(
+						SELECT COUNT(*) FROM entity_attributes attr
+						JOIN entity_aspects asp ON asp.id = attr.aspect_id
```
PR-Reviewer-Ant, line 845: Blocking: same scoping regression in `getKnowledgeEntityDetail` for `attribute_count` / `constraint_count`. Missing `asp.agent_id = e.agent_id` in both scalar subqueries weakens tenant isolation and can return incorrect counts under mixed-agent data. Mirror the old scoping contract by filtering joined aspects on agent as well.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `591ec888`

[Automated Review by pr-reviewer] Found a blocking agent-scoping regression in the rewritten scalar subqueries; otherwise the performance rewrite and migration align with the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In the new `listKnowledgeEntities` and `getKnowledgeEntityDetail` queries, the attribute/constraint subqueries join `entity_attributes` to `entity_aspects` but do not constrain `asp.agent_id = e.agent_id`. The previous implementation scoped aspects by agent before counting attributes. This is a concrete diff-level scoping change in `packages/daemon/src/knowledge-graph.ts` and directly conflicts with the repository’s explicit anti-scoping-leak guidance.

NicholaiVogel (approved):
Hell yes, this looks good.

The scoping issue from the first review is fixed, the threads are resolved, and CI is green. This is exactly the kind of boring-but-important perf work the graph needed. Clean migration, sane rollback, good regression coverage.

Approved. Ship it.

PR #519, opened 2026-04-17
feat(desktop): replace Tauri shell with Electron app

NicholaiVogel:
## Summary

- Adds `@signet/desktop`, an Electron desktop app that bundles the daemon runtime, serves the dashboard through an app protocol, owns tray/menu behavior, and launches Signet as the consumer desktop distribution layer.
- Deprecates the old Tauri implementation from `@signet/tray`, keeping that package as a shared tray/menu state library only.
- Updates dashboard desktop-shell integration for custom Signet window decorations, desktop-only scale controls, Electron IPC, and local app serving.
- Updates desktop packaging docs, AUR generation, CI workflow, architecture notes, and the existing approved desktop packaging spec/research references.

## Readiness checklist

## Validation

- `bun test packages/daemon/src/daemon-cors.test.ts`
- `bunx biome check packages/desktop/src/main.ts packages/cli/dashboard/src/lib/stores/ui-scale.svelte.ts packages/cli/dashboard/src/routes/+page.svelte packages/cli/dashboard/src/lib/components/ui/sidebar/sidebar.svelte packages/cli/dashboard/src/app.css`
- `cd packages/desktop && bun run typecheck`
- `bun run build:desktop -- --dir`
- `git diff --check`
- Live Electron launch verified locally with custom decorations, no dashboard 404, Wayland GPU path without Vulkan spam, and Ctrl +/- scaling no longer clipping the dashboard bottom.

## Notes

- Agent scoping and admin/mutation endpoint checks are not materially changed by this PR. The only daemon API surface change is the Electron app origin allowlist, covered by `daemon-cors.test.ts`.
- The desktop shell intentionally keeps GPU rendering on Wayland while disabling Vulkan-specific paths on Linux.


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(desktop): replace Tauri shell with Electron app` (commit `004c78a5`) now and I'll follow up shortly with feedback.

.github/workflows/desktop-build.yml
```diff
@@ -44,434 +48,150 @@ jobs:
...
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: ${{ matrix.target }}
-
-      - name: Resolve signing mode (macOS)
-        if: matrix.runner == 'macos-latest'
-        id: mac_sign_mode
+      - name: Resolve signing mode
+        id: signing
         env:
```
PR-Reviewer-Ant, line 54: Blocking: `signing_mode=official` is no longer validated for required secrets. The previous workflow explicitly failed when official signing credentials were missing; this version accepts `official` without a hard precheck and can proceed with empty signing env vars. That weakens release integrity guarantees and can yield improperly signed artifacts while appearing to run in official mode.

.github/workflows/desktop-build.yml
```diff
@@ -44,434 +48,150 @@ jobs:
...
-              "$($msi.FullName)"
-
-            if ($LASTEXITCODE -ne 0) {
-              Write-Error "signtool failed for $($msi.FullName) with exit code $LASTEXITCODE"
-              exit $LASTEXITCODE
-            }
-          }
-
-          Remove-Item "Cert:\CurrentUser\My\$thumb" -ErrorAction SilentlyContinue
+          CSC_IDENTITY_AUTO_DISCOVERY: ${{ steps.signing.outputs.mode == 'official' && 'true' || 'false' }}
```
PR-Reviewer-Ant, line 153: Warning: AUR metadata generation now runs on every Ubuntu matrix run, and `URL` is synthesized as `.../releases/download/v${VER}/...` even when the ref is not a `v*` tag. This can produce invalid PKGBUILD source URLs in CI artifacts for non-release refs.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `004c78a5`

[Automated Review by pr-reviewer] Found one release-integrity blocker and one packaging-metadata concern.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The workflow changes in .github/workflows/desktop-build.yml explicitly removed the hard validation that previously failed when official signing secrets were missing, while still allowing official mode. The AUR URL construction is also explicit and now runs outside tag-only conditions.

.github/workflows/desktop-build.yml
NicholaiVogel: Fixed in `68cd2e90`. Official signing mode now hard-fails on macOS/Windows when the required platform signing secrets are missing, and `auto` only selects official mode when those secrets are present. Revalidated the workflow YAML locally.

.github/workflows/desktop-build.yml
NicholaiVogel: Fixed in `68cd2e90`. AUR metadata generation, upload, and validation now only run for `v*` tag builds, so workflow_dispatch/non-release refs no longer produce release-download URLs that do not exist. Revalidated workflow YAML and rebuilt the desktop package locally.

.github/workflows/desktop-build.yml
```diff
@@ -44,34 +48,53 @@ jobs:
...
           bun-version: latest
 
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          targets: ${{ matrix.target }}
-
-      - name: Resolve signing mode (macOS)
-        if: matrix.runner == 'macos-latest'
-        id: mac_sign_mode
+      - name: Resolve signing mode
```
PR-Reviewer-Ant, line 51: Blocking: `Resolve signing mode` now uses bash-specific syntax (`set -euo pipefail`, arrays, `[[ ... ]]`, function definitions) but no `shell: bash` is set. On `windows-latest`, default shell is PowerShell, so this step will fail and break the matrix build/release pipeline.

.github/workflows/desktop-build.yml
```diff
@@ -80,398 +103,130 @@ jobs:
...
               mode="self-signed"
-              echo "::warning::APPLE_* signing secrets missing, falling back to self-signed mode"
+              printf '::warning::Official signing secrets missing, falling back to self-signed mode: %s\n' "${missing[*]}"
             fi
           fi
-
           if [ "${mode}" != "official" ] && [ "${mode}" != "self-signed" ]; then
             echo "::error::Invalid signing mode: ${mode}"
             exit 1
           fi
```
PR-Reviewer-Ant, line 113: Blocking: `Configure Electron signing` also uses bash syntax and writes to `$GITHUB_ENV` via heredoc markers, but no `shell: bash` is declared. This will fail on Windows runners for the same reason, preventing signed Windows artifact generation.

packages/desktop/src/daemon-manager.ts
```diff
@@ -0,0 +1,177 @@
...
+		const health = await this.probe();
+		return {
+			running: health !== null,
+			owned: this.#owned,
+			pid: health?.pid ?? null,
+			version: health?.version ?? null,
+			uptime: health?.uptime ?? null,
+			port: this.port,
+			baseUrl: this.baseUrl,
+		};
```
PR-Reviewer-Ant, line 87: Blocking race condition: `ensureStarted()` has no serialization/single-flight guard. If called concurrently (startup + IPC/tray action), both calls can observe no health, both call `#spawn()`, and race process ownership/state. This can produce duplicate spawn attempts and inconsistent `#owned/#child` tracking.

packages/desktop/src/main.ts
```diff
@@ -0,0 +1,218 @@
...
+		webPreferences: {
+			preload: preloadPath(),
+			contextIsolation: true,
+			nodeIntegration: false,
+			sandbox: false,
+		},
+	});
+	lockNativeZoom(mainWindow);
+
+	mainWindow.webContents.setWindowOpenHandler(({ url }) => {
```
PR-Reviewer-Ant, line 106: Security warning: BrowserWindow is created with `sandbox: false`. With preload IPC exposing daemon control and memory mutation actions, disabling renderer sandbox increases impact if renderer content is ever compromised. Consider keeping sandbox enabled unless a strict technical blocker exists.

packages/daemon/src/daemon-cors.test.ts
```diff
@@ -31,7 +31,7 @@ memory:
 
 	afterAll(() => {
 		if (prev === undefined) {
-			delete process.env.SIGNET_PATH;
+			process.env.SIGNET_PATH = undefined;
```
PR-Reviewer-Ant: Test isolation bug: replacing `delete process.env.SIGNET_PATH` with `process.env.SIGNET_PATH = undefined` can leave a stringified env value instead of removing the variable, causing cross-test pollution in code that reads `SIGNET_PATH`.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `41a6a399`

[Automated Review by pr-reviewer] Found blocking CI/runtime reliability issues and one security-hardening concern; this is not ready for merge without fixes.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The workflow now uses bash syntax in cross-platform steps without setting `shell: bash`, which will fail on `windows-latest` runners. The daemon manager start path has no single-flight guard, so concurrent `ensureStarted()` calls can race and spawn multiple daemon processes. Both are directly visible in changed code.

.github/workflows/desktop-build.yml
NicholaiVogel: Fixed in `3cd637a0`. Added explicit `shell: bash` to the signing resolution step so the Windows matrix runner executes the bash script correctly.

.github/workflows/desktop-build.yml
NicholaiVogel: Fixed in `3cd637a0`. Added explicit `shell: bash` to the Electron signing configuration step too, including the `$GITHUB_ENV` heredoc writes.

packages/desktop/src/daemon-manager.ts
NicholaiVogel: Fixed in `3cd637a0`. `ensureStarted()` now uses a single-flight promise so concurrent startup calls share the same probe/spawn/wait path, and the internal start path will not spawn again if an owned child already exists.

packages/desktop/src/main.ts
NicholaiVogel: Fixed in `3cd637a0`. Re-enabled the Electron renderer sandbox while keeping context isolation and the existing preload IPC bridge. I also relaunched the desktop app locally after the change to verify the dashboard still loads.

packages/daemon/src/daemon-cors.test.ts
NicholaiVogel: Fixed in `3cd637a0`. Restored env cleanup to delete `SIGNET_PATH`, with a targeted Biome ignore explaining why this should not use `process.env.SIGNET_PATH = undefined`. The CORS regression test passes.

packages/desktop/src/daemon-manager.ts
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		if (this.#child && this.#owned) {
+			this.#child.kill("SIGTERM");
+		} else if (health.pid > 0) {
+			try {
+				process.kill(health.pid, "SIGTERM");
+			} catch (err) {
+				throw new Error(
+					`Failed to stop daemon process ${health.pid}: ${err instanceof Error ? err.message : String(err)}`,
```
PR-Reviewer-Ant: Blocking: `stop()` can kill an arbitrary local process because it trusts `health.pid` from `GET /health` and calls `process.kill(health.pid, "SIGTERM")` when the daemon is not marked as owned. A malicious or accidental service on port 3850 can return any PID, causing unintended process termination. Restrict kill operations to owned child processes (or cryptographically verify daemon identity before signaling).

packages/desktop/src/tray.ts
```diff
@@ -0,0 +1,302 @@
...
+
+		const [memories, diagnostics, embeddings] = await Promise.all([
+			json<MemoryResponse>(`${this.#daemon.baseUrl}/api/memories?limit=10`, 5000),
+			json<DiagnosticsResponse>(`${this.#daemon.baseUrl}/api/diagnostics`, 5000),
+			json<EmbeddingResponse>(`${this.#daemon.baseUrl}/api/embeddings/status`, 5000),
+		]);
+
+		const recent = (memories?.memories ?? []).map(
+			(memory): RecentMemory => ({
+				id: stringOrNull(memory.id) ?? "",
```
PR-Reviewer-Ant, line 232: Reliability issue: start/restart/stop menu handlers call `this.poll()` without canceling the existing timer/loop. Repeated actions can create overlapping poll loops and duplicate network load. Guard for single-flight polling or clear/reset the prior timer before scheduling a new cycle.

packages/desktop/src/daemon-manager.ts
```diff
@@ -0,0 +1,188 @@
...
+
+	#spawn(): void {
+		const entry = daemonEntry();
+		if (!existsSync(entry)) {
+			throw new Error(`Bundled daemon entry not found: ${entry}`);
+		}
+
+		const logDir = join(app.getPath("userData"), "logs");
+		mkdirSync(logDir, { recursive: true });
+		const stdout = createWriteStream(join(logDir, "daemon.out.log"), { flags: "a" });
```
PR-Reviewer-Ant: Reliability/resource leak: each `#spawn()` allocates new `createWriteStream` handles for daemon logs, but those streams are never closed on child exit/restart. Over multiple restarts this can leak file descriptors. Close/destroy log streams in the child `exit` handler (or reuse a managed logger).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `3cd637a0`

[Automated Review by pr-reviewer] I found one blocking process-safety/security issue and two reliability issues in the new Electron desktop runtime path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The blocking issue is directly visible in the new daemon lifecycle code: it trusts PID data from an unauthenticated localhost `/health` response and then sends OS kill signals to that PID. The reliability issues are also explicit in added control flow and resource handling in new files.

packages/desktop/src/daemon-manager.ts
NicholaiVogel: Fixed in 11b87620. Desktop stop now only signals the owned child process. If another service or non-owned daemon answers on port 3850, the desktop app returns status instead of trusting the reported PID.

packages/desktop/src/tray.ts
NicholaiVogel: Fixed in 11b87620. Tray polling is now single-flight: manual refreshes clear the existing timer, overlapping polls coalesce into one immediate follow-up poll, and stop prevents rescheduling after the tray is destroyed.

packages/desktop/src/daemon-manager.ts
NicholaiVogel: Fixed in 11b87620. Daemon log streams are now owned by DaemonManager, closed before spawning a new child, and closed again from the child exit handler to avoid descriptor leaks across restarts.

packages/desktop/src/daemon-manager.ts
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		this.#owned = false;
+		this.#child = null;
+		return this.status();
+	}
+
+	async restart(): Promise<DesktopDaemonStatus> {
+		await this.stop();
+		await sleep(500);
```
PR-Reviewer-Ant, line 148: `restart()` can silently no-op if the owned daemon has not fully exited. In `stop()`, the code clears `#owned` and `#child` after a bounded wait even if the process is still alive, then `restart()` calls `start()` which can return early when probe sees an already-running daemon. This violates restart semantics under slow/hung shutdown and can leave users on the old process instance.

packages/desktop/src/tray.ts
```diff
@@ -0,0 +1,322 @@
...
+			title: kind === "capture" ? "Quick Capture" : "Search Memories",
+			resizable: true,
+			webPreferences: {
+				preload: preloadPath(),
+				contextIsolation: true,
+				nodeIntegration: false,
+			},
+		});
+		win
+			.loadURL(`data:text/html;charset=utf-8,${encodeURIComponent(kind === "capture" ? captureHtml() : searchHtml())}`)
```
PR-Reviewer-Ant, line 292: The new capture/search utility windows use a preload bridge with privileged IPC, but `webPreferences` does not set `sandbox: true` (main window does). This increases renderer attack surface compared to the hardened main window. Consider enabling sandbox here too for consistency and defense-in-depth.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `999894aa`

[Automated Review by pr-reviewer] I found one reliability bug that can break restart behavior, plus one security-hardening gap in newly added utility windows.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The restart path issue is directly visible in the new daemon lifecycle code: `stop()` nulls child ownership state before confirmed process exit, and `restart()` immediately calls `start()` which short-circuits if health is still up. I did not run the desktop app to reproduce timing behavior under a slow/shutdown-hung daemon.

packages/desktop/src/daemon-manager.ts
NicholaiVogel: Fixed in 08b91103. Restart now waits for the owned daemon child to actually exit and fails if shutdown times out or the port remains occupied, instead of clearing ownership and accidentally treating the old process as a successful restart.

packages/desktop/src/tray.ts
NicholaiVogel: Fixed in 08b91103. The quick capture and search utility windows now use the same sandboxed renderer posture as the main desktop window.

PR #523, opened 2026-04-17
fix(cli): pass AGENTS_DIR directly in runSyncTemplates instead of accepting a parameter

### LINKED ISSUE #522: fix(cli): signet sync always reports 'No Signet installation found' (Commander arg wiring)
## Summary

`signet sync` prints "No Signet installation found. Run: signet setup" on every invocation, even when the workspace at `~/.agents` is valid and both `signet status` and `signet doctor` work correctly.

## Environment

- signet 0.100.0
- macOS, bun 1.3.9
- Valid workspace with all required files (`agent.yaml`, `AGENTS.md`, `SOUL.md`, `IDENTITY.md`, `USER.md`, `MEMORY.md`, `memory/memories.db`)

## Repro

```bash
signet status    # ✓ works, shows "Path: /Users/.../.agents"
signet doctor    # ✓ "Looks healthy"
signet sync      # ✗ "No Signet installation found. Run: signet setup"
```

With `--trace-deprecation`:

```
(node:62249) [DEP0187] DeprecationWarning: Passing invalid argument types to fs.existsSync is deprecated
    at existsSync (node:fs:277:15)
    at syncTemplates (file:///.../dist/cli.js:205484:8)
    at Command.runSyncTemplates (file:///.../dist/cli.js:207096:51)
```

## Root cause

In `packages/cli/src/cli.ts`, `runSyncTemplates` accepts an optional `basePath` parameter:

```ts
const runSyncTemplates = (basePath = AGENTS_DIR): Promise<void> =>
    syncTemplates({ agentsDir: basePath, ... });
```

This is registered as a Commander `.action()` handler in `packages/cli/src/commands/app.ts:115`:

```ts
program.command("sync").description("...").action(deps.syncTemplates);
```

Commander calls `.action(fn)` with `(options, command)`, so `basePath` receives the Commander options object `{}` instead of falling through to the default `AGENTS_DIR`. `existsSync({})` returns `false` + triggers the Node deprecation warning.

The type interface in `app.ts:41` already declares `readonly syncTemplates: () => Promise<void>` (no parameters), but the implementation has a broader signature that accepts Commander's args silently.

Other commands in the same file avoid this: `configureAgent` is wrapped as `() => configureAgent({...})`, `launchDashboard` as `(options) => launchDashboard(options, ...)`.

## Proposed fix

Remove the unused `basePath` parameter from `runSyncTemplates` so it matches the declared type and always uses `AGENTS_DIR` directly:

```ts
const runSyncTemplates = (): Promise<void> =>
    syncTemplates({ agentsDir: AGENTS_DIR, ... });
```

PR incoming.

Assisted-by: Claude Opus 4.6 (1M context)

ddasgupta4:
## Summary

Fix `signet sync` always printing "No Signet installation found" even on valid workspaces.

## Changes

- **`packages/cli/src/cli.ts`**: Remove unused `basePath` parameter from `runSyncTemplates` so it always uses `AGENTS_DIR` directly instead of accepting Commander's `(options, command)` args as the path.
- **`packages/cli/src/features/sync.test.ts`**: Add two regression tests — valid workspace succeeds without error, missing workspace reports the expected message.

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/cli`

## Screenshots

N/A — CLI-only fix, no UI changes.

## PR Readiness (MANDATORY)

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

5 tests pass in `packages/cli/src/features/sync.test.ts` (3 existing + 2 new regression tests).

Closes #522

## AI disclosure

This PR was authored with assistance from Claude Opus 4.6 (1M context).


PR-Reviewer-Ant:
Hi @ddasgupta4 - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(cli): pass AGENTS_DIR directly in runSyncTemplates instead of accepting a parameter` (commit `150f0a16`) now and I'll follow up shortly with feedback.

packages/cli/src/features/sync.test.ts
```diff
@@ -32,6 +32,73 @@ afterEach(() => {
...
+describe("syncTemplates workspace detection", () => {
+	it("succeeds when agentsDir is a valid directory (regression: #515-related wiring)", async () => {
+		const root = mkdtempSync(join(tmpdir(), "sync-workspace-"));
+		const basePath = join(root, "agents");
+
+		try {
+			process.env.HOME = root;
+			mkdirSync(basePath, { recursive: true });
+			const logs: string[] = [];
+			const origLog = console.log;
```
PR-Reviewer-Ant: This test patches global `console.log` but only restores it after `await syncTemplates(...)` succeeds. If that call throws, `console.log` stays patched and can corrupt subsequent tests. Wrap the patch/restore in `try/finally` so restoration is guaranteed.

packages/cli/src/features/sync.test.ts
```diff
@@ -32,6 +32,73 @@ afterEach(() => {
...
+			expect(output).not.toContain("No Signet installation found");
+		} finally {
+			rmSync(root, { recursive: true, force: true });
+		}
+	});
+
+	it("reports missing installation when agentsDir does not exist", async () => {
+		const logs: string[] = [];
+		const origLog = console.log;
+		console.log = (...args: unknown[]) => logs.push(args.join(" "));
```
PR-Reviewer-Ant: Same global-state leakage risk here: `console.log` is patched without `try/finally`. A thrown error before restoration can cause cross-test contamination and misleading failures.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `150f0a16`

[Automated Review by pr-reviewer] The core fix in `cli.ts` matches the PR claim (preventing Commander arg leakage into `runSyncTemplates`), but the new regression tests introduce brittle global `console.log` monkeypatching that is not safely restored on failure.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The functional fix is directly visible in `packages/cli/src/cli.ts` where `runSyncTemplates` no longer accepts a parameter and always uses `AGENTS_DIR`. In `packages/cli/src/features/sync.test.ts`, `console.log` is reassigned in two new tests without guaranteed restoration if `syncTemplates(...)` throws, which can leak state across tests.

NicholaiVogel:
Thanks for tracking this down. I think the diagnosis is right, but I do not think the fix belongs in `runSyncTemplates` itself.

The actual problem is that Commander action handlers receive `(options, command)`, and the `sync` command was passing its handler through directly. That let Commander’s options object leak into a function that expected a workspace path. The fix should be at the command boundary, where we adapt Commander’s calling convention, not by removing the path-aware internal API.

I want to keep `runSyncTemplates(basePath = AGENTS_DIR)` because other internal flows already use that path parameter. In particular, `signet daemon restart --path ...` resolves a workspace path and calls `deps.syncTemplates(basePath)`. With the current PR shape, that can restart one workspace and sync the default workspace, which is a subtle regression.

The shape I’d prefer is:

```ts
const runSyncTemplates = (basePath = AGENTS_DIR): Promise<void> =>
  syncTemplates({
    agentsDir: basePath,
    ...
  });
```

Then adapt the top-level Commander command with a zero-arg wrapper:

```ts
syncTemplates: () => runSyncTemplates(),
```

or wrap it directly in `registerAppCommands`:

```ts
.action(() => deps.syncTemplates())
```

We can push a follow-up commit here to make that change and add a regression test against the Commander/app command wiring. The current `syncTemplates()` tests are useful, but they do not exercise the failure mode from #522, since the bug was in the Commander handler boundary rather than in `syncTemplates()` itself.

NicholaiVogel:
Pushed `77164ab0` with the follow-up:

- keeps `runSyncTemplates(basePath = AGENTS_DIR)` path-aware for internal callers like `daemon restart --path ...`
- keeps the top-level `signet sync` command on a zero-arg wrapper so Commander args cannot leak into the workspace path
- adds a regression test for the actual Commander/app command boundary

Validation:

```text
bun test packages/cli/src/commands/app.test.ts packages/cli/src/features/daemon.test.ts packages/cli/src/features/sync.test.ts
bun run typecheck
bunx biome check packages/cli/src/cli.ts packages/cli/src/commands/app.ts packages/cli/src/commands/app.test.ts packages/cli/src/features/sync.test.ts
cd packages/cli && bun run build:cli
```

packages/cli/src/cli.ts
```diff
@@ -1339,7 +1339,7 @@ registerAppCommands(program, {
 		}),
```
PR-Reviewer-Ant, line 1339: PR description divergence: the summary says `runSyncTemplates` no longer accepts `basePath`, but current code keeps `runSyncTemplates(basePath = AGENTS_DIR)` and fixes the bug at the Commander boundary via zero-arg wrappers. Please update the PR description so reviewers/audit history reflect the actual fix.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `5f23706f`

[Automated Review] Implementation appears functionally correct for the reported `signet sync` bug, but the PR description no longer matches what the final code does.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed command wiring in `packages/cli/src/commands/app.ts` and `packages/cli/src/cli.ts` now correctly blocks Commander `(options, command)` arg leakage by wrapping `syncTemplates` in zero-arg closures, and the new `app.test.ts` verifies no args are forwarded. However, the PR body still claims `runSyncTemplates` had its `basePath` parameter removed, which is not true in the current head.

packages/cli/src/cli.ts
NicholaiVogel: shut up clanker

ddasgupta4:
nice!!!

PR #527, opened 2026-04-19
fix(daemon): resolve all typescript strict-mode errors in daemon package

Ostico:
## Summary

Eliminate ~200+ TypeScript strict-mode errors across `packages/daemon` and `packages/core`, bringing the non-TS6059 error count to **zero**. Also restructures daemon.ts by extracting routes/middleware and co-locating all tests inside `src/`.

## Changes

- Widen `LogCategory` union with 22+ missing values in `logger.ts`
- Fix Hono handler return types to use `ContentfulStatusCode` literals
- Add missing interface fields: `temperature` on `LlmProvider`, `TaskHarness` export from core, `paused` on `PipelineV2Config`
- Narrow `logger.error()` signatures, fix `logger.time()` return type
- Type ~30 test files: Hono app vars, MigrationDb casts, missing config fields, optional method guards, spread literal narrowing, property name corrections
- Extract health/auth/dashboard routes from `daemon.ts` into dedicated route modules
- Extract global middleware into `middleware.ts`
- Co-locate auth guards from centralized `daemon.ts` into individual route files
- Move all test files from `tests/` into `src/` alongside source; merge 3 duplicate test files

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

N/A — no migrations touched.

## Testing

- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

### Verification details

- `tsc --noEmit`: **0 non-TS6059 errors** (60 TS6059 are pre-existing rootDir violations from `core/src/migrations` imports — structural tsconfig issue, not introduced by this PR)
- Test suite: **1709 pass / 144 fail** — zero regressions verified via `git stash` before/after comparison. All 144 failures + 78 errors are pre-existing.
- 5 pre-existing failures in `worker.integration.test.ts` confirmed identical before and after changes.

## AI disclosure

## Notes

- The 60 remaining TS6059 rootDir errors come from `core/src/migrations` being outside daemon's `rootDir` — fixing this requires tsconfig restructuring (separate effort).
- `@signet/core` changes: added `temperature?: number` to `LlmProvider` opts, exported `TASK_HARNESSES` + `TaskHarness`. Core package was rebuilt before verifying daemon.


PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): resolve all typescript strict-mode errors in daemon package` (commit `4eb7f717`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `4eb7f717`

[Automated Review by pr-reviewer] I found behavior-changing regressions that conflict with the PR’s stated strict-typing/refactor scope and should be fixed before merge.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The flagged issues are directly visible in changed lines in packages/daemon/src/daemon.ts (fallback defaulting, async stop handling, connector telemetry counting). I did not receive full contents of newly extracted route/middleware modules, so I cannot fully validate all auth parity beyond what the included tests cover.

packages/daemon/src/daemon.ts
```diff
@@ -1648,7 +1129,7 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 		});
```
PR-Reviewer-Ant, line 1140: Behavior change: `fallbackProvider` is now forced to `"llama-cpp"` when unset (`?? "llama-cpp"`). Previously an unset fallback could remain unset and lead to a blocked extraction path. This silently changes runtime failover semantics and can cause unexpected local fallback attempts. That diverges from a strict typing/refactor-only claim and should be explicitly justified or reverted.

packages/daemon/src/daemon.ts
```diff
@@ -1481,7 +962,7 @@ async function stopPipelineRuntime(): Promise<void> {
```
PR-Reviewer-Ant, line 972: Potential race condition in shutdown/restart: `skillReconcilerHandle.stop()` is no longer awaited. If `stop()` is async (as previously handled), restart can proceed while prior reconciler teardown is still in flight, risking overlapping workers/timers and inconsistent pipeline state.

packages/daemon/src/daemon.ts
```diff
@@ -2677,7 +2160,7 @@ async function main() {
```
PR-Reviewer-Ant, line 2172: Telemetry logic regression: `connectorsActive` now counts every connector whose status is not `"error"`, which likely includes non-active states (e.g., disabled/inactive). This can corrupt operational metrics and health interpretation.

packages/daemon/src/routes/knowledge-routes.ts
```diff
@@ -45,23 +45,25 @@ export function registerKnowledgeRoutes(app: Hono): void {
 	});
```
PR-Reviewer-Ant, line 45: Auth guard parity appears incomplete after co-location. The old daemon-level guards required `recall` permission for `/api/knowledge/expand`, `/api/knowledge/expand/session`, and `/api/graph/impact`; this module now applies `requirePermission` for pin/unpin but not those expand/impact routes. Even if scope checks still reject some calls, this diverges from the prior explicit permission model and from the PR claim that guards were co-located.

packages/daemon/src/daemon.ts
```diff
@@ -2679,7 +2162,7 @@ async function main() {
```
PR-Reviewer-Ant, line 2172: Heartbeat telemetry changed `connectorsActive` from `status === "active"` to `status !== "error"`. This broadens the counted set (e.g. idle/syncing) and can skew operational metrics/alerts that interpret this field as actively connected connectors.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `afcbaa10`

[Automated Review by pr-reviewer] I found permission-parity drift in the route refactor and one telemetry logic change that should be reviewed before merge.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The previous centralized guard block in daemon.ts explicitly protected /api/knowledge/expand, /api/knowledge/expand/session, and /api/graph/impact with requirePermission("recall"). In the extracted knowledge routes, permission middleware is applied to pin/unpin but not to those endpoints. I also verified the heartbeat metric change in daemon.ts from status === "active" to status !== "error", which changes semantics for connector activity counts.

packages/daemon/src/daemon.ts
Ostico: Good catch. The `?? "llama-cpp"` was a type-narrowing fix (the source type `PipelineV2Config.extraction.fallbackProvider` is optional) but it silently changes runtime behavior when the field is unset.

Fixed: replaced with an explicit runtime guard that throws if undefined. `memory-config.ts` already guarantees a value via `resolveExtractionFallbackProvider()` (line 477-479), so the guard never fires in practice, but it satisfies the type system without introducing a silent default. Also deduplicated — the variable is now declared once before both uses.

packages/daemon/src/daemon.ts
Ostico: Verified: `ReconcilerHandle.stop()` returns `void`, not `Promise<void>` — see `skill-reconciler.ts` line 37: `stop(): void`. The `await` on a synchronous void return was a no-op. Removing it is correct and introduces no race condition.

packages/daemon/src/daemon.ts
Ostico: This is an intentional fix for a latent bug. `ConnectorStatus` is defined as `"idle" | "syncing" | "error"` (see `core/src/connector-types.ts` line 11-12) — there is no `"active"` value in the union. The original `=== "active"` comparison was always false under strict typing, so `connectorsActive` was always 0.

`!== "error"` correctly counts non-errored connectors (idle + syncing), which matches the semantic intent of the `connectorsActive` field name. This surfaced during TS strict-mode cleanup.

packages/daemon/src/routes/knowledge-routes.ts
Ostico: Good catch — this was a real gap. The three daemon-level `requirePermission("recall")` middleware guards for `/api/knowledge/expand`, `/api/knowledge/expand/session`, and `/api/graph/impact` were dropped during co-location but not re-added inside the route handlers.

Fixed: added `requirePermission("recall", authConfig)` guards (early-return pattern) to all three route handlers in `knowledge-routes.ts`. The file now has 5 permission guards total: 2 `modify` for pin/unpin, 3 `recall` for expand/impact — matching the original daemon.ts middleware parity.

packages/daemon/src/daemon.ts
Ostico: Addressed in the earlier thread (comment #3106115545): `ConnectorStatus` is `"idle" | "syncing" | "error"` — no `"active"` value exists. The original `=== "active"` was always false (latent bug). `!== "error"` is the correct semantic intent for counting non-errored connectors.

packages/daemon/src/daemon.ts
```diff
@@ -1593,6 +1074,10 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 	reloadAuthState(AGENTS_DIR);
 
 	const providerHints = getConfiguredProviderHints(AGENTS_DIR);
+	const extractionFallbackProvider = memoryCfg.pipelineV2.extraction.fallbackProvider;
+	if (extractionFallbackProvider === undefined) {
```
PR-Reviewer-Ant: Blocking: startup now hard-fails if `memoryCfg.pipelineV2.extraction.fallbackProvider` is undefined (`throw new Error(...)`). In core types this field remains optional, so malformed/legacy-resolved configs can now crash daemon boot instead of degrading gracefully. This is a breaking runtime behavior change for a PR described as strict-mode cleanup/refactor.

packages/daemon/src/daemon.ts
PR-Reviewer-Ant, line 972: Blocking: `skillReconcilerHandle.stop()` is no longer awaited during shutdown. If `stop()` is async (it was previously awaited), restart can race with incomplete teardown, causing leaked timers/workers or double-running reconcilers. This risks data integrity and nondeterministic behavior under restart/failure paths.

packages/daemon/src/daemon.ts
```diff
@@ -2606,7 +2092,7 @@ async function main() {
 					telemetryRef.record("daemon.heartbeat", {
```
PR-Reviewer-Ant, line 2099: Warning: heartbeat telemetry changed `connectorsActive` from counting only `status === "active"` to counting every non-error status. That alters the semantic meaning of the metric and diverges from a no-behavior-change refactor claim. If intentional, this should be called out and validated with downstream metric consumers.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `9f38cfaa`

[Automated Review by pr-reviewer] I found blocking reliability regressions and one behavior divergence from the PR’s stated refactor-only intent.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The blocking findings are directly visible in changed lines of packages/daemon/src/daemon.ts: an explicit startup throw on optional config and removal of an await during runtime shutdown. I could not inspect every extracted route module in this context, so auth-guard coverage completeness outside shown tests remains unverified.

packages/daemon/src/daemon.ts
```diff
@@ -236,392 +190,11 @@ export function recordPredictorLatency(operation: "predictor_score" | "predictor
```
PR-Reviewer-Ant, line 190: Potential security/DoS regression: the prior explicit `/api/config` POST guard (admin permission + `content-length` cap at 1,048,576 bytes) was removed from `daemon.ts`, and this diff does not show an equivalent replacement. If not reintroduced in route modules, large unauthenticated/overlarge payloads could reach config handling.

packages/daemon/src/daemon-status.test.ts
```diff
@@ -24,12 +24,15 @@ describe("daemon status contract", () => {
 		process.env.SIGNET_PATH = dir;
 
 		const daemon = await import("./daemon");
+		const state = await import("./routes/state.js");
+		state.reloadAuthState(dir);
 		app = daemon.app;
+		countConnectorsActive = daemon.countConnectorsActive;
 	});
 
 	afterAll(() => {
```
PR-Reviewer-Ant, line 33: Environment cleanup bug: `process.env.SIGNET_PATH = undefined` can leave a literal `"undefined"` value instead of deleting the variable. This can leak state into subsequent tests and cause nondeterministic path resolution failures. Use `delete process.env.SIGNET_PATH` when `prev` is undefined.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `b6fe021a`

[Automated Review by pr-reviewer] I found one likely security/regression blocker and two maintainability/correctness concerns that should be addressed before merge.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The diff explicitly removes the in-daemon `/api/config` POST size/permission middleware block and central auth guards, but the replacement implementation is split into other route modules not fully shown here. The removed guard itself is concrete evidence of potential regression, while two additional findings are directly visible in changed files.

packages/core/src/types.ts
```diff
@@ -294,6 +297,7 @@ export interface PipelineGuardrailsConfig {
 	readonly maxContentChars: number;
```
PR-Reviewer-Ant, line 297: Public type contract changed: `PipelineGuardrailsConfig.contextBudgetChars` was added as a required field. This is a breaking API/type change for downstream consumers constructing this config, but the PR description frames the change as strict-mode cleanup/refactor and does not call out this contract expansion.

packages/daemon/src/daemon.ts
NicholaiVogel: Follow-up in `b6fe021a`: I replaced both the silent default and the later startup throw with a type-level resolution boundary.

`loadPipelineConfig()` now returns `ResolvedPipelineV2Config`, where `extraction.fallbackProvider` is non-optional after config resolution. The daemon consumes that narrowed resolved config and no longer needs `?? "llama-cpp"` or a boot-time throw. Invalid explicit fallback values still fail config validation, while missing legacy config keeps using the existing resolved default path.

packages/daemon/src/daemon.ts
NicholaiVogel: Follow-up in `b6fe021a`: I restored the shutdown await, but wrapped it with `Promise.resolve(skillReconcilerHandle.stop())`.

Current `ReconcilerHandle.stop()` still returns `void`, so this is behavior-neutral today. If `stop()` becomes async later, shutdown/restart will still wait for teardown instead of racing.

packages/daemon/src/daemon.ts
NicholaiVogel: Follow-up in `b6fe021a`: I made this intentional behavior explicit and added coverage for it.

`countConnectorsActive()` now owns the heartbeat semantics, with a comment explaining that `ConnectorStatus` is `"idle" | "syncing" | "error"` and has no `"active"` state. The regression test asserts idle + syncing are counted and error is excluded. If we need a literal syncing-only metric later, that should be added as a separate metric rather than restoring the impossible `status === "active"` comparison.

packages/daemon/src/routes/knowledge-routes.ts
NicholaiVogel: Confirmed and revalidated during the takeover pass. The recall guards added earlier remain in place for `/api/knowledge/expand`, `/api/knowledge/expand/session`, and `/api/graph/impact`.

I reran the auth co-location regression suite after the latest changes:

`bun test packages/daemon/src/daemon-auth-guard-colocation.test.ts packages/daemon/src/daemon-status.test.ts packages/daemon/src/pipeline/provider.test.ts`

Result: 94 pass, 0 fail.

packages/daemon/src/daemon.ts
NicholaiVogel: Addressed in `b6fe021a`.

The startup hard-fail has been removed. Instead of throwing in `daemon.ts`, the resolved config type now models the real invariant: `loadPipelineConfig()` returns `ResolvedPipelineV2Config`, with a non-optional `extraction.fallbackProvider` after config resolution. That keeps strict-mode narrowing without introducing a new daemon boot failure path.

packages/daemon/src/daemon.ts
NicholaiVogel: Addressed in `b6fe021a`.

`skillReconcilerHandle.stop()` is awaited again via `await Promise.resolve(skillReconcilerHandle.stop())`. That is still a no-op for the current synchronous `void` implementation, but preserves the shutdown contract if the reconciler stop path becomes async later.

packages/daemon/src/daemon.ts
NicholaiVogel: Addressed in `b6fe021a` by making the metric semantics explicit and tested.

`connectorsActive` now calls `countConnectorsActive()`, which documents the historical naming mismatch and counts registered non-errored connectors. The added status test covers idle + syncing counted as active and error excluded. This is intentional because `ConnectorStatus` does not include `"active"`; the old comparison could only report 0.

packages/daemon/src/daemon.ts
NicholaiVogel: Addressed in `a51ff801`.

The guard had been moved into `registerMiscRoutes()`, but you were right that the PR needed an explicit regression for the co-located behavior. I added coverage for both pieces:

`POST /api/config` without auth now asserts 403.
`POST /api/config` with `content-length: 1048577` now asserts 413 before body parsing.

That validates both the admin permission guard and the 1,048,576 byte request cap at the route-module boundary.

packages/daemon/src/daemon-status.test.ts
NicholaiVogel: Addressed in `a51ff801`.

Switched cleanup from `process.env.SIGNET_PATH = undefined` to `Reflect.deleteProperty(process.env, "SIGNET_PATH")` when there was no previous value. That avoids leaving a literal `"undefined"` env value while also avoiding the linted `delete` operator.

I made the same cleanup pattern explicit in the auth co-location test while preserving any previous `SIGNET_PATH`.

packages/core/src/types.ts
NicholaiVogel: Addressed in `a51ff801`.

`PipelineGuardrailsConfig.contextBudgetChars` is optional again in the public core type, so downstream consumers are not forced to add a new required field for this strict-mode cleanup PR.

The daemon still resolves it to a required number internally by narrowing `ResolvedPipelineV2Config.guardrails.contextBudgetChars`. That preserves the runtime invariant used by hooks while keeping the exported config contract backward-compatible.

NicholaiVogel:
@Ostico Thanks for this! Currently working out some broader structural and behavioral changes so I'm glad we're getting this in before all of that lands. Merging this :)

Ostico:
It's a pleasure, @NicholaiVogel

PR #537, opened 2026-04-20
fix(daemon): suppress OpenCode notifications for extraction sessions

### LINKED ISSUE #532: Suppress OpenCode desktop notifications for extraction sessions
**Signet version:** 0.102.4
**Environment:** Linux, Bun 1.3.12
**Related issues:** N/A

---

## Summary

Signet's memory pipeline creates OpenCode sessions (`POST /session { title: "signet-extraction" }`) for LLM extraction. When the LLM responds, these sessions transition to idle, which triggers OpenCode's desktop notification ("Response ready: signet-extraction"). Users see repeated unwanted notifications for internal pipeline work they never initiated.

A fix has been implemented on branch `ostico/suppress-extraction-notifications` (commit `a8c15b3e`). This issue documents the bug, the fix, and an intentional behavioral change introduced by the fix (warn-level error alerting added to the OpenCode provider).

---

## Observed Behavior

Every time the summary worker triggers extraction via `createOpenCodeProvider`, the user receives a desktop notification:

```
🔔 Response ready — signet-extraction
```

This fires once per extraction call. During active sessions with frequent memory distillation, users can see dozens of these notifications in quick succession.

---

## Expected Behavior

Internal pipeline sessions should be invisible to the user. No desktop notification should fire for Signet's own extraction work.

---

## Root Cause Analysis

Traced the full notification flow across both codebases:

```
summary-worker.ts
  → provider.ts createSession(): POST /session { title: "signet-extraction" }
  → provider.ts sendMessage(): POST /session/{sid}/message
  → OpenCode processes LLM request → session becomes idle
  → OpenCode status.ts: publishes "session.idle" bus event
  → OpenCode notification.tsx handleSessionIdle():
      if (session.parentID) return   ← ONLY session-level filter
      platform.notify("Response ready", "signet-extraction")
```

Key findings:
- `bypassSession()` only prevents Signet's own hooks from re-triggering — OpenCode has no awareness of bypass state
- OpenCode's notification handler has exactly one session-level filter: `if (session.parentID) return` at `notification.tsx:234`
- There is no `internal`, `hidden`, or `system` flag on sessions
- The notification fires from the app/desktop client layer, not the headless server

**Constraint:** OpenCode cannot be modified — the fix must be Signet-only.

---

## Fix (implemented)

Branch: `ostico/suppress-extraction-notifications`
Commit: `a8c15b3e`
Files changed: `packages/daemon/src/pipeline/provider.ts`, `packages/daemon/src/pipeline/provider.test.ts`

### Approach

Leverage OpenCode's existing `parentID` mechanism:

1. **Parent session**: Lazily create a persistent "signet-system" session via `getOrCreateParentSession()`, cache its ID in the provider closure.
2. **Child sessions**: All extraction sessions include `parentID` in the POST body → OpenCode's notification handler skips them. `GET /session?roots=true` also hides children from the default session list.
3. **Session cleanup**: After each `sendMessage()` completes (success or failure), fire-and-forget `DELETE /session/{sid}` in a `.finally()` callback. Parent session persists.
4. **Error alerting**: A `.catch()` before `.finally()` logs extraction failures at `warn` level via `logger.warn("pipeline", "OpenCode extraction failed", ...)`. This is **new** visibility — previously, extraction errors were only caught by the pipeline's outer handler at debug level and effectively invisible. The warn channel surfaces errors in daemon log files (`~/.agents/.daemon/logs/`) and in the Signet Dashboard's live SSE stream. Note: this does **not** reach any harness (OpenCode, Claude Code, Codex, etc.) — it is daemon-operator-visible only (see "Visibility Scope" below).

### Graceful degradation

- If parent session creation fails → extraction proceeds without `parentID` → notifications fire (degraded but functional)
- If DELETE cleanup fails → session lingers as a hidden child (parentID set) → no user impact
- If parent session is deleted by the user in OpenCode → cached `parentSessionId` still suppresses notifications (parentID truthy check is unconditional)

### Test coverage

- `withParentSession()` mock wrapper added — silently handles parent session creation and DELETE cleanup across all 33 OpenCode test mocks
- Renamed "reuses session" test → "creates a new child session per call after cleanup" (assertion updated: 1 → 2 session creations)
- New QA test: "passes parentID to child sessions and cleans up via DELETE" — directly asserts parentID on child POST bodies, DELETE calls for children, and parent preservation
- Result: **36 OpenCode tests pass, 0 fail**

---

## ⚠️ Behavioral Change: Error Alerting (OpenCode provider only)

**This fix introduces warn-level error alerting that is unique to the OpenCode provider.**

### Before this fix

**No provider** sends any notification or alert on extraction failure. All providers (Ollama, Claude Code, Codex, OpenRouter, llama.cpp, OpenCode) simply throw exceptions caught by the pipeline's outer handler and logged at debug level. Errors are effectively invisible — you would need debug-level logging enabled to see them.

### After this fix

The **OpenCode provider only** now logs extraction errors at `warn` level:

```typescript
logger.warn("pipeline", "OpenCode extraction failed", {
    error: err instanceof Error ? err.message : String(err),
    sessionId,
});
```

Other providers remain unchanged — they still only throw to the pipeline's outer handler at debug level.

### Visibility Scope

`logger.warn` writes to 3 channels, **all daemon-internal**:

| Channel | Destination | Who sees it |
|---|---|---|
| `console.log()` | Daemon stdout | Only if daemon runs in foreground terminal |
| File append | `~/.agents/.daemon/logs/signet-YYYY-MM-DD.log` | Anyone checking log files |
| EventEmitter → SSE | `/api/logs/stream` | Signet Dashboard (browser UI) |

**No harness reads daemon logs or subscribes to the SSE stream.** A user working in OpenCode, Claude Code, or Codex will never see these warnings unless they open the Signet Dashboard or inspect log files manually.

This is an improvement (errors go from invisible to operator-visible) but the asymmetry across providers should be resolved.

The disclaimer is documented in code:
```typescript
// NOTE(changed-behavior): This warn-level error alerting is unique
// to the OpenCode provider.  Other providers (Ollama, Claude Code,
// Codex, OpenRouter, llama.cpp) only throw — errors surface in debug
// logs via the pipeline's outer handler.
// TODO: extend warn-level error alerting to all providers for
// consistent visibility.
```

---

### Current behavior

Only the OpenCode provider gets `logger.warn` on extraction failure. Other providers have zero active error reporting — errors are invisible unless you enable debug-level logging.

### Desired behavior

All providers should report extraction failures at `warn` level with structured error context (error message, provider name, session/request ID where applicable). This makes error visibility consistent regardless of which extraction backend is configured.

### Scope

| Provider | File | Pattern |
|---|---|---|
| Ollama | `provider.ts` `createOllamaProvider` | Add `.catch()` with `logger.warn` on the generate path |
| Claude Code | `provider.ts` `createClaudeCodeProvider` | Same pattern |
| Codex | `provider.ts` `createCodexProvider` | Same pattern |
| OpenRouter | `provider.ts` `createOpenRouterProvider` | Same pattern |
| llama.cpp | `provider.ts` `createLlamaCppProvider` | Same pattern |
| ✅ OpenCode | `provider.ts` `createOpenCodeProvider` | Already implemented |

**Note on harness-level alerting:** No harness besides OpenCode has a runtime notification mechanism. All connectors (claude-code, codex, oh-my-pi, pi, openclaw, hermes-agent) are install-time only — they copy config files and exit. There is no harness-side channel to push error alerts through. If harness-visible error reporting is desired in the future, it would require new infrastructure (e.g., a status endpoint that harnesses poll, or error injection into hook responses).

### Expected impact

- Consistent error visibility across all providers — no provider is a black box
- Dashboard shows extraction health regardless of backend choice
- The "changed behavior" disclaimer on the OpenCode provider becomes moot once all providers match

---

## Reviewer

@NicholaiVogel — please review the fix approach (parentID + cleanup + error alerting). Specifically:

1. Is the `parentID` approach acceptable as the Signet-only notification suppression mechanism?
2. Is the behavioral asymmetry (OpenCode-only warn-level error alerting) acceptable as an interim state, or should the follow-up (extend to all providers) land in the same PR?
3. Daemon-side `logger.warn` is the only error visibility channel today — no harness has a runtime path to receive it. Is daemon logs + dashboard sufficient, or should we design a new mechanism for harness-visible error reporting?

Ostico:
## Summary

Suppress unwanted OpenCode desktop notifications triggered by Signet's extraction pipeline sessions. Creates a persistent parent session (`signet-system`) and attaches its ID as `parentID` on every extraction session. OpenCode's notification handler skips child sessions, silencing pipeline noise without losing error visibility.

Closes #532

## Changes

- `packages/daemon/src/pipeline/provider.ts` — Add `getOrCreateParentSession()` to create/cache a `signet-system` parent session; attach `parentID` to extraction session creation payloads; add `deleteSession()` for fire-and-forget cleanup after each extraction; add `logger.warn` on extraction failure for daemon-side error visibility.
- `packages/daemon/src/pipeline/provider.test.ts` — Add `withParentSession()` test helper to handle parent session and DELETE mock routes; add dedicated `parentID + cleanup QA` test; update all existing `mockFetch` calls to use the wrapper; rename session-reuse test to reflect the new per-call lifecycle.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

N/A — no migrations.

## Testing

- [x] `bun test` passes (36 tests in provider.test.ts)
- [ ] `bun run lint` passes
- [ ] N/A

**Manual verification:** Ran the daemon from this branch in dev mode (`bun --watch src/daemon.ts`) and confirmed:
- `logger.warn("pipeline", "OpenCode extraction failed", ...)` fires correctly on extraction timeouts.
- Extraction pipeline remains functional — facts inserted, synthesis running, git auto-committing.
- Parent session created once and reused across extractions.
- Child sessions cleaned up via fire-and-forget DELETE after each extraction.

## AI disclosure

## Notes

- **Graceful degradation:** If parent session creation fails, extraction proceeds unparented — notifications fire but extraction still works.
- **Changed behavior:** `logger.warn` on extraction failure is unique to the OpenCode provider. Other providers only throw and rely on the pipeline's outer handler for debug-level logging. See `TODO` comment in code for future work to extend warn-level alerting to all providers.
- **Session lifecycle change:** Extraction sessions are now created fresh per call and deleted after use (previously the session was reused). This is necessary because child sessions with `parentID` cannot be "re-parented" after creation.

### DIFF
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -2171,7 +2171,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		// Attach parentID so OpenCode treats extraction sessions as children.
 		// Child sessions are hidden from the root session list and, crucially,
 		// skipped by the desktop notification handler.
-		const parentId = await getOrCreateParentSession();
+		const parentId = await getOrCreateParentSession(remainingMs);
 		const payload: Record<string, unknown> = { title: "signet-extraction" };
 		if (parentId) payload.parentID = parentId;
 
@@ -2215,14 +2215,15 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 	 *  for pipeline work.  Returns null on failure so extraction can
 	 *  proceed unparented (notifications will fire but extraction still
 	 *  works). */
-	async function getOrCreateParentSession(): Promise<string | null> {
+	async function getOrCreateParentSession(remainingMs?: number): Promise<string | null> {
 		if (parentSessionId) return parentSessionId;
 		try {
+			const timeout = Math.min(5_000, Math.max(1, remainingMs ?? 5_000));
 			const res = await fetch(`${cfg.baseUrl}/session`, {
 				method: "POST",
 				headers: { "Content-Type": "application/json" },
 				body: JSON.stringify({ title: "signet-system" }),
-				signal: AbortSignal.timeout(5_000),
+				signal: AbortSignal.timeout(timeout),
 			});
 			if (!res.ok) {
 				logger.warn("pipeline", "OpenCode parent session creation failed", {
@@ -2299,6 +2300,12 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		// Refresh bypass TTL on reused sessions so bypass-only entries do not
 		// expire while the provider is still actively sending messages.
 		bypassSession(sid, { allowUnknown: true });
+
+		// Track the session this specific call is using.  Retry paths may
+		// replace it with a fresh session; the .finally() cleanup reads
+		// this local instead of the shared `sessionId` to avoid a race
+		// when multiple calls run concurrently inside the semaphore.
+		let activeSid = sid;
 
 		return withLlmConcurrency(async () => {
 			const remaining = deadline - performance.now();
@@ -2398,6 +2405,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				const retryWithNewSession = async (): Promise<OpenCodeMessageResponse | null> => {
 					sessionId = null;
 					const retrySid = await getOrCreateSession(deadline - performance.now());
+					activeSid = retrySid;
 					const retryRes = await postMessage(retrySid);
 					if (!retryRes.ok) {
 						const retryBody = await retryRes.text().catch(() => "");
@@ -2431,6 +2439,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 						consumedBody = null;
 						const retrySid = await createSession(deadline - performance.now());
 						sessionId = retrySid;
+						activeSid = retrySid;
 						res = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
 							method: "POST",
 							headers: { "Content-Type": "application/json" },
@@ -2467,10 +2476,12 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 						});
 						if (retryRes.ok) {
 							sessionId = retrySid;
+							activeSid = retrySid;
 							const retryParsed = await parsePostResponse(retryRes, retrySid);
 							if (retryParsed) return retryParsed;
 						}
 						sessionId = null;
+						activeSid = sid;
 						const ollamaFallback = await tryOllamaFallback(prompt, deadline - performance.now(), opts, "agent-not-found-retry-failed");
 						if (ollamaFallback) return ollamaFallback;
 						return buildOpenCodeFallbackResponse();
@@ -2500,6 +2511,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 					sessionId = null;
 					const fallbackSid = await createSession(deadline - performance.now());
 					sessionId = fallbackSid;
+					activeSid = fallbackSid;
 					const fallbackRes = await fetch(`${cfg.baseUrl}/session/${fallbackSid}/message`, {
 						method: "POST",
 						headers: { "Content-Type": "application/json" },
@@ -2540,10 +2552,9 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 			});
 			throw err;
 		}).finally(() => {
-			const finalSid = sessionId;
-			sessionId = null;
-			if (finalSid) void deleteSession(finalSid);
-			if (sid !== finalSid) void deleteSession(sid);
+			if (sessionId === activeSid || sessionId === sid) sessionId = null;
+			void deleteSession(activeSid);
+			if (sid !== activeSid) void deleteSession(sid);
 		});
 	}
 
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -2071,7 +2071,6 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 	const providerID = slashIdx > 0 ? cfg.model.slice(0, slashIdx) : "anthropic";
 	const modelID = slashIdx > 0 ? cfg.model.slice(slashIdx + 1) : cfg.model;
 
-	let sessionId: string | null = null;
 	let parentSessionId: string | null = null;
 	let ollamaFallbackProvider: LlmProvider | null = null;
 
@@ -2252,13 +2251,18 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 	/** Fire-and-forget deletion of an extraction session.  If the call
 	 *  fails the session remains as a hidden child (parentID set) and
 	 *  does not appear in OpenCode's root session list. */
-	async function deleteSession(sid: string): Promise<void> {
+	async function deleteSession(sid: string | null): Promise<void> {
+		if (!sid) return;
 		try {
-			await fetch(`${cfg.baseUrl}/session/${sid}`, {
+			const res = await fetch(`${cfg.baseUrl}/session/${sid}`, {
 				method: "DELETE",
 				signal: AbortSignal.timeout(5_000),
 			});
-			logger.debug("pipeline", "OpenCode extraction session deleted", { id: sid });
+			if (res.ok) {
+				logger.debug("pipeline", "OpenCode extraction session deleted", { id: sid });
+			} else {
+				logger.debug("pipeline", "OpenCode extraction session cleanup skipped", { id: sid, status: res.status });
+			}
 		} catch {
 			logger.debug("pipeline", "OpenCode extraction session cleanup skipped", { id: sid });
 		}
@@ -2307,9 +2311,9 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 			const sid = await createSession(deadline - performance.now());
 			bypassSession(sid, { allowUnknown: true });
 
-			// Track the session this specific call is using.  Retry paths may
-			// replace it with a fresh session; the finally cleanup uses this
-			// local to delete the correct session.
+			// Track every session created during this call so the finally
+			// block can clean up all of them — not just the first and last.
+			const allSids = new Set<string | null>([sid]);
 			let activeSid = sid;
 
 			const controller = new AbortController();
@@ -2403,6 +2407,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 
 				const retryWithNewSession = async (): Promise<OpenCodeMessageResponse | null> => {
 					const retrySid = await createSession(deadline - performance.now());
+					allSids.add(retrySid);
 					activeSid = retrySid;
 					const retryRes = await postMessage(retrySid);
 					if (!retryRes.ok) {
@@ -2436,6 +2441,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 						}
 						consumedBody = null;
 						const retrySid = await createSession(deadline - performance.now());
+						allSids.add(retrySid);
 						activeSid = retrySid;
 						res = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
 							method: "POST",
@@ -2465,6 +2471,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 							agent: cfg.agent,
 						});
 						const retrySid = await createSession(deadline - performance.now());
+						allSids.add(retrySid);
 						const retryRes = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
 							method: "POST",
 							headers: { "Content-Type": "application/json" },
@@ -2492,7 +2499,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 					throw new Error(`OpenCode HTTP ${res.status}: ${body.slice(0, 200)}`);
 				}
 
-				const parsed = await parsePostResponse(res, sid);
+				const parsed = await parsePostResponse(res, activeSid);
 				if (parsed) return parsed;
 
 				const retryParsed = await retryWithNewSession();
@@ -2504,6 +2511,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 					});
 					structuredOutputSupported = false;
 					const fallbackSid = await createSession(deadline - performance.now());
+					allSids.add(fallbackSid);
 					activeSid = fallbackSid;
 					const fallbackRes = await fetch(`${cfg.baseUrl}/session/${fallbackSid}/message`, {
 						method: "POST",
@@ -2541,8 +2549,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				throw err;
 			} finally {
 				clearTimeout(timer);
-				void deleteSession(activeSid);
-				if (sid !== activeSid) void deleteSession(sid);
+				for (const s of allSids) void deleteSession(s);
 			}
 		}, timeoutMs, "opencode");
 	}
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -2180,12 +2180,26 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		const payload: Record<string, unknown> = { title: "signet-extraction" };
 		if (parentId) payload.parentID = parentId;
 
-		const res = await fetch(`${cfg.baseUrl}/session`, {
+		let res = await fetch(`${cfg.baseUrl}/session`, {
 			method: "POST",
 			headers: { "Content-Type": "application/json" },
 			body: JSON.stringify(payload),
 			signal: AbortSignal.timeout(timeoutMs),
 		});
+
+		if (!res.ok && parentId) {
+			logger.warn("pipeline", "Child session creation failed with parentID, retrying unparented", {
+				status: res.status,
+				parentId,
+			});
+			parentSessionId = null;
+			res = await fetch(`${cfg.baseUrl}/session`, {
+				method: "POST",
+				headers: { "Content-Type": "application/json" },
+				body: JSON.stringify({ title: "signet-extraction" }),
+				signal: AbortSignal.timeout(timeoutMs),
+			});
+		}
 
 		if (!res.ok) {
 			const body = await res.text().catch(() => "");
--- a/packages/daemon/src/pipeline/provider.ts
+++ b/packages/daemon/src/pipeline/provider.ts
@@ -2209,12 +2209,6 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		return id;
 	}
 
-	async function getOrCreateSession(remainingMs?: number): Promise<string> {
-		if (sessionId) return sessionId;
-		sessionId = await createSession(remainingMs);
-		return sessionId;
-	}
-
 	/** Create or return a cached parent session used as parentID for
 	 *  extraction sessions.  OpenCode's notification handler skips sessions
 	 *  that carry a parentID, suppressing unwanted desktop notifications
@@ -2310,12 +2304,12 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 
 			// Session creation is inside the semaphore so concurrent
 			// generate() calls cannot share and then race-delete a session.
-			const sid = await getOrCreateSession(deadline - performance.now());
+			const sid = await createSession(deadline - performance.now());
 			bypassSession(sid, { allowUnknown: true });
 
 			// Track the session this specific call is using.  Retry paths may
-			// replace it with a fresh session; the finally cleanup reads this
-			// local instead of the shared `sessionId`.
+			// replace it with a fresh session; the finally cleanup uses this
+			// local to delete the correct session.
 			let activeSid = sid;
 
 			const controller = new AbortController();
@@ -2408,8 +2402,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				};
 
 				const retryWithNewSession = async (): Promise<OpenCodeMessageResponse | null> => {
-					sessionId = null;
-					const retrySid = await getOrCreateSession(deadline - performance.now());
+					const retrySid = await createSession(deadline - performance.now());
 					activeSid = retrySid;
 					const retryRes = await postMessage(retrySid);
 					if (!retryRes.ok) {
@@ -2443,7 +2436,6 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 						}
 						consumedBody = null;
 						const retrySid = await createSession(deadline - performance.now());
-						sessionId = retrySid;
 						activeSid = retrySid;
 						res = await fetch(`${cfg.baseUrl}/session/${retrySid}/message`, {
 							method: "POST",
@@ -2460,7 +2452,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 						const retryParsed = await retryWithNewSession();
 						if (retryParsed) return retryParsed;
 						logger.warn("pipeline", "OpenCode response remained malformed after retry; using fallback", {
-							sessionId,
+							sessionId: activeSid,
 						});
 						const ollamaFallback = await tryOllamaFallback(prompt, deadline - performance.now(), opts, "post-response-malformed-after-http-retry");
 						if (ollamaFallback) return ollamaFallback;
@@ -2480,12 +2472,10 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 							signal: controller.signal,
 						});
 						if (retryRes.ok) {
-							sessionId = retrySid;
 							activeSid = retrySid;
 							const retryParsed = await parsePostResponse(retryRes, retrySid);
 							if (retryParsed) return retryParsed;
 						}
-						sessionId = null;
 						activeSid = sid;
 						const ollamaFallback = await tryOllamaFallback(prompt, deadline - performance.now(), opts, "agent-not-found-retry-failed");
 						if (ollamaFallback) return ollamaFallback;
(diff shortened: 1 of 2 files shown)


PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): suppress OpenCode notifications for extraction sessions` (commit `a8c15b3e`) now and I'll follow up shortly with feedback.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2196,6 +2209,60 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+	 *  proceed unparented (notifications will fire but extraction still
+	 *  works). */
+	async function getOrCreateParentSession(): Promise<string | null> {
+		if (parentSessionId) return parentSessionId;
+		try {
+			const res = await fetch(`${cfg.baseUrl}/session`, {
+				method: "POST",
+				headers: { "Content-Type": "application/json" },
+				body: JSON.stringify({ title: "signet-system" }),
+				signal: AbortSignal.timeout(5_000),
```
PR-Reviewer-Ant: Blocking: `getOrCreateParentSession()` uses a hardcoded `AbortSignal.timeout(5_000)` and is called before the main request's semaphore/deadline enforcement. This can exceed a caller-provided short `timeoutMs` budget before message send starts, violating expected timeout behavior and causing deadline drift under parent-session latency.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2072,6 +2072,7 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 	const modelID = slashIdx > 0 ? cfg.model.slice(slashIdx + 1) : cfg.model;
 
 	let sessionId: string | null = null;
+	let parentSessionId: string | null = null;
```
PR-Reviewer-Ant, line 2074: Blocking: session lifecycle is still tracked via shared mutable `sessionId` on the provider instance, so concurrent `generate()` calls can race and reuse/delete each other’s session. This contradicts the PR’s claimed per-call child-session lifecycle and can produce cross-call interference (one call deleting a session still in use by another).

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `a8c15b3e`

[Automated Review by pr-reviewer] Found two blocking reliability/correctness issues in the OpenCode extraction-session lifecycle changes; implementation does not fully guarantee the per-call session behavior claimed in the PR.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly inferable from the changed control flow in packages/daemon/src/pipeline/provider.ts: parent-session creation uses a fixed 5s timeout outside the request deadline budget, and session state remains shared mutable state (`sessionId`) across calls while cleanup deletes by that shared state in finally. These are concrete code-path issues in the modified lines.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2329,8 +2446,8 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 				};
```
PR-Reviewer-Ant, line 2418: Polling timeout was reduced to at most 1s (`Math.min(Math.max(0, pollRemaining), 1000)`), down from a much longer window. This can cause false 'malformed/empty' fallbacks when OpenCode returns assistant messages slightly later, producing empty extraction output and data-quality regressions under normal latency.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2161,16 +2162,40 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+		// Attach parentID so OpenCode treats extraction sessions as children.
+		// Child sessions are hidden from the root session list and, crucially,
+		// skipped by the desktop notification handler.
+		const parentId = await getOrCreateParentSession(remainingMs);
+		const childRemainingMs =
+			remainingMs !== undefined ? remainingMs - (performance.now() - started) : undefined;
+		if (childRemainingMs !== undefined && childRemainingMs <= 0) {
+			throw new Error("OpenCode timeout creating extraction session");
+		}
+
```
PR-Reviewer-Ant, line 2177: A cached `parentSessionId` is used for all child sessions, but there is no invalidation/retry path if that parent becomes stale server-side. If OpenCode rejects `parentID` (expired/deleted parent), `createSession()` throws and extraction fails instead of degrading to unparented mode as described. Add retry logic that clears `parentSessionId` and retries child creation without parent (or recreates parent) on parent-related create-session failure.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `02ceb552`

[Automated Review by pr-reviewer] I found two correctness regressions that should be fixed before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in changed code paths in provider.ts: the poll window was reduced from up to 20s to 1s, and parent-session caching has no stale-parent recovery path when child session creation fails due invalid parentID.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed in 23c29a75. `getOrCreateParentSession()` now accepts `remainingMs` and uses `Math.min(5_000, remainingMs ?? 5_000)` for its timeout, so the parent session fetch participates in the caller's deadline budget instead of consuming a hardcoded 5s independently.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed in 23c29a75. Introduced a call-local `activeSid` in `sendMessage()` that tracks which session each call is actively using. Retry paths update `activeSid` alongside the shared `sessionId`. The `.finally()` cleanup now reads `activeSid` instead of the shared closure-level `sessionId`, so concurrent calls under the semaphore (max 4) no longer race on cleanup — each call deletes exactly its own session.

packages/daemon/src/pipeline/provider.ts
Ostico: This finding misreads the nesting. The actual code is:

```typescript
const pollDeadline = performance.now() + Math.max(1000, Math.min(pollRemaining, 20000));
```

`Math.max(1000, ...)` sets a **floor** of 1s, not a ceiling. The poll window is always at least 1s and at most `min(pollRemaining, 20s)`. This is intentional — it guarantees a minimum poll window even when remaining budget is low, while capping at 20s to avoid unbounded polling. No change needed.

packages/daemon/src/pipeline/provider.ts
Ostico: Valid robustness suggestion, but out of scope for this PR.

The current behavior: `getOrCreateParentSession()` returns `null` on creation failure, and `createSession()` only attaches `parentID` when non-null. So initial parent creation failure already degrades gracefully — extraction continues unparented.

The gap you identified: parent created successfully, cached, then deleted server-side later, stale cached ID used for child creation which fails. This is a narrow edge case — OpenCode does not garbage-collect sessions during normal daemon uptime. The parent lives for the daemon's lifetime.\n\nI'll track this as a follow-up enhancement rather than blocking the notification suppression fix.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2233,6 +2301,12 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 		// expire while the provider is still actively sending messages.
```
PR-Reviewer-Ant: Blocking: concurrent `generate()` calls on the same provider can still share a single child session, violating the PR claim that each call gets a fresh child session. `sendMessage()` calls `getOrCreateSession()` before semaphore acquisition, and `getOrCreateSession()` returns shared `sessionId` when set. A second call starting before the first call’s `.finally()` clears `sessionId` will reuse the same session ID, then both calls race cleanup/deletion on that shared session.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2166,10 +2167,18 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
 			remainingMs !== undefined
```
PR-Reviewer-Ant: Blocking: session creation can exceed the caller deadline because `createSession(remainingMs)` spends up to `remainingMs` in `getOrCreateParentSession(remainingMs)` and then still allows another fetch with timeout derived from the same original `remainingMs`. This double-consumes budget before the semaphore-guarded message call, so short timeouts can overrun their end-to-end deadline.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `23c29a75`

[Automated Review by pr-reviewer] I found two blocking correctness/reliability issues in the OpenCode session lifecycle changes; the implementation still diverges from the PR’s per-call session guarantee and can overrun request deadlines.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the changed control flow in packages/daemon/src/pipeline/provider.ts: `getOrCreateSession` is still backed by shared mutable `sessionId` across calls, and `createSession` reuses the same `remainingMs` budget for both parent-session and child-session creation without decrementing elapsed time.

packages/daemon/src/pipeline/provider.ts
Ostico: The shared-session race is a pre-existing pattern — `getOrCreateSession()` was called before `withLlmConcurrency()` before this PR. That said, since the review caught it, I moved session creation inside the semaphore in eb612bb7.

Each `sendMessage()` call now acquires the semaphore first, then creates/reuses its session, and cleans up in `finally` before releasing. No concurrent call can see or delete a session another call is actively using.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed in eb612bb7. `createSession()` now measures elapsed time from parent-session creation and subtracts it before computing the child timeout:

```typescript
const parentStart = performance.now();
const parentId = await getOrCreateParentSession(remainingMs);
const parentMs = performance.now() - parentStart;
const adjusted = remainingMs !== undefined ? remainingMs - parentMs : undefined;
const timeoutMs = adjusted !== undefined
    ? Math.max(1, Math.min(adjusted, 10_000))
    : 10_000;
```

Child session creation now stays within the caller overall budget.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2228,19 +2302,24 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
-		const sid = await getOrCreateSession(deadline - performance.now());
-		// Refresh bypass TTL on reused sessions so bypass-only entries do not
-		// expire while the provider is still actively sending messages.
-		bypassSession(sid, { allowUnknown: true });
-
 		return withLlmConcurrency(async () => {
 			const remaining = deadline - performance.now();
 			if (remaining <= 0) {
 				throw new Error(`OpenCode timeout after ${timeoutMs}ms (deadline exceeded waiting for semaphore)`);
 			}
```
PR-Reviewer-Ant, line 2320: Blocking: the implementation still permits cross-call session sharing on the same provider instance. `getOrCreateSession(...)` is backed by shared `sessionId`, and this block is only guarded by the global LLM semaphore (capacity >1), not a per-provider mutex. Concurrent `generate()` calls can therefore receive the same `sid`, violating the PR’s 'new child session per call' goal and creating a race with cleanup.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2452,12 +2536,26 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+				// NOTE(changed-behavior): This warn-level error alerting is unique
+				// to the OpenCode provider.  Other providers (Ollama, Claude Code,
+				// Codex, OpenRouter, llama.cpp) only throw — errors surface in debug
+				// logs via the pipeline's outer handler.
+				// TODO: extend warn-level error alerting to all providers for
+				// consistent visibility (see plan: suppress-opencode-extraction-
+				// notifications.md § Future Work).
+				logger.warn("pipeline", "OpenCode extraction failed", {
+					error: err instanceof Error ? err.message : String(err),
+					sessionId,
```
PR-Reviewer-Ant: Blocking: `finally` always schedules `deleteSession(activeSid)` (and sometimes `deleteSession(sid)`), which is unsafe if another in-flight call reused the same session ID. This can delete a session still in use by a concurrent call, producing nondeterministic failures (404/retry/fallback) and undermining reliability.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `eb612bb7`

[Automated Review by pr-reviewer] I found a blocking concurrency/lifecycle bug that still conflicts with the PR’s stated per-call child-session behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In `sendMessage()`, session creation is moved inside the semaphore, but the provider still uses shared mutable `sessionId` (`getOrCreateSession`) across calls. Because the semaphore allows up to 4 concurrent calls globally, two concurrent `generate()` calls on the same provider instance can still reuse one `sid`. The `finally` block then unconditionally schedules deletion for that `sid`, so one call can delete a session another call is still using. This directly contradicts the PR claim of fresh per-call child sessions and can cause intermittent 404/retry/fallback behavior under load.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed in `8eb1e1d0`. Each `sendMessage()` invocation now calls `createSession()` directly and tracks the session only through call-local `sid`/`activeSid`. The shared `sessionId` is no longer read or written inside `sendMessage()`, so concurrent `generate()` calls inside the semaphore cannot race-delete each other's sessions.

Changes:
- Deleted `getOrCreateSession()` (dead code once shared state is removed)
- Removed all `sessionId = ...` writes from `sendMessage()` (all retry paths: format rejection, agent rejection, malformed-200 fallback, finally block)
- Replaced `sessionId` references in 4 log statements with `activeSid`
- Simplified the finally block — no conditional shared-state cleanup needed
- Added concurrency regression test: 3 staggered concurrent `generate()` calls on the same provider must each get a distinct session ID

packages/daemon/src/pipeline/provider.ts
Ostico: Already fixed in `eb612bb7` — `createSession()` now subtracts the parent-creation elapsed time before computing the child timeout budget.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2186,14 +2201,67 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+				});
+				return null;
+			}
+			const data = (await res.json()) as Record<string, unknown>;
+			const id = data.id;
+			if (typeof id !== "string") {
+				logger.warn("pipeline", "OpenCode parent session response missing 'id'");
+				return null;
+			}
+			parentSessionId = id;
```
PR-Reviewer-Ant, line 2251: `getOrCreateParentSession()` is race-prone under concurrency. Multiple concurrent `generate()` calls can all observe `parentSessionId === null` and each create a separate `signet-system` parent before one wins the cache write. This diverges from the PR claim that the parent is created once/reused, and can leak extra parent sessions.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2446,18 +2518,31 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
 				return buildOpenCodeFallbackResponse();
 			} catch (e) {
-				if (e instanceof DOMException && e.name === "AbortError") {
-					throw new Error(`OpenCode timeout after ${timeoutMs}ms`);
-				}
-				throw e;
+				const err = (e instanceof DOMException && e.name === "AbortError")
+					? new Error(`OpenCode timeout after ${timeoutMs}ms`)
+					: e;
+				// NOTE(changed-behavior): This warn-level error alerting is unique
```
PR-Reviewer-Ant, line 2550: Session cleanup is incomplete on multi-retry paths. The `finally` block only deletes the original `sid` and the final `activeSid`, but code paths can create additional sessions (`retrySid`, `fallbackSid`, etc.) that get superseded before exit and are never deleted. This contradicts the stated per-call cleanup behavior and can accumulate hidden sessions over time.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `8eb1e1d0`

[Automated Review by pr-reviewer] I found two lifecycle correctness issues in the new OpenCode session-parenting/cleanup flow that should be fixed before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the changed control flow in packages/daemon/src/pipeline/provider.ts: parent-session creation is unsynchronized across concurrent calls, and cleanup only deletes `sid` plus the final `activeSid`, missing intermediate retry sessions created in the same request.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed in ec73abfa.

**Parent session race:** Acknowledged — this is a benign TOCTOU where concurrent first calls can each create a parent session before one wins the cache write. Since parent sessions are long-lived, never deleted, and the cache stabilizes after the first write, the impact is limited to one-time duplicate creation. Adding a mutex would add complexity for no practical benefit. Noted as future work.

**What was done:** No code change for the parent race (accepted risk). The session leak part of this review cycle is addressed below in finding 8.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed in ec73abfa.

The reviewer is correct — intermediate retry sessions could leak. I traced every path and confirmed leaks in:

1. **Agent rejection failure** (`activeSid = sid` reverts, `retrySid` orphaned)
2. **Malformed-200 double-retry** (`retryWithNewSession` sets `activeSid = retrySid`, then `fallbackSid` overwrites — `retrySid` orphaned)
3. **Format rejection → subsequent retry** (format retry session orphaned when a later retry replaces `activeSid`)

**Fix:** All created sessions are now tracked in a `Set<string | null>` (`allSids`). Every `createSession()` call adds to the set. The `finally` block iterates the full set:

```typescript
for (const s of allSids) void deleteSession(s);
```

Also fixed a second bug discovered during deep review: `parsePostResponse(res, sid)` was passing the original session ID instead of `activeSid` after a 422 format-rejection retry — polling would hit the wrong session.

Both bugs have TDD regression tests that were verified failing before the fix was applied.

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2186,14 +2200,72 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+	 *  extraction sessions.  OpenCode's notification handler skips sessions
+	 *  that carry a parentID, suppressing unwanted desktop notifications
+	 *  for pipeline work.  Returns null on failure so extraction can
+	 *  proceed unparented (notifications will fire but extraction still
+	 *  works). */
+	async function getOrCreateParentSession(remainingMs?: number): Promise<string | null> {
+		if (parentSessionId) return parentSessionId;
+		try {
+			const timeout = Math.min(5_000, Math.max(1, remainingMs ?? 5_000));
+			const res = await fetch(`${cfg.baseUrl}/session`, {
```
PR-Reviewer-Ant, line 2233: Cached `parentSessionId` is reused forever without validation. If OpenCode restarts or the parent session is deleted/expired, subsequent child session creation can fail when posting `parentID`, causing extraction failures. This contradicts the PR’s graceful-degradation goal (continue extraction even when parenting is unavailable). Consider retry logic that clears stale `parentSessionId` and retries child creation without `parentID` (or recreates parent once) before failing.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `ec73abfa`

[Automated Review by pr-reviewer] Found a reliability regression that can break extraction after parent-session expiry; not ready for merge without fixes.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new parent-session cache in provider.ts returns a cached ID unconditionally and never validates/recreates it. Child session creation always reuses that ID as parentID, so once the cached parent becomes invalid, createSession can fail and abort extraction instead of degrading gracefully.

packages/daemon/src/pipeline/provider.ts
Ostico: Fixed.

`createSession()` now catches child-session creation failures when a `parentID` was attached, clears the cached `parentSessionId`, and retries without `parentID`. Extraction continues unparented — notifications will fire but extraction still works, matching the graceful-degradation contract documented in `getOrCreateParentSession()`.

The retry is scoped: only triggers when `!res.ok && parentId`, so the happy path (valid parent) adds zero overhead.

Regression test added: mock returns 400 for first child POST with stale parentID, verifies retry without parentID succeeds.

NicholaiVogel:
@Ostico thanks for seeing this thru! apologies for the force push earlier, didn't see you had begun working on it again until afterwards. Not seeing any issues from my end, going to take a closer look and then get this merged. Thank you

packages/daemon/src/pipeline/provider.ts
```diff
@@ -2162,17 +2162,45 @@ export function createOpenCodeProvider(config?: Partial<OpenCodeProviderConfig>)
...
+			body: JSON.stringify(payload),
 			signal: AbortSignal.timeout(timeoutMs),
 		});
 
+		if (!res.ok && parentId) {
+			logger.warn("pipeline", "Child session creation failed with parentID, retrying unparented", {
+				status: res.status,
+				parentId,
+			});
+			parentSessionId = null;
```
PR-Reviewer-Ant, line 2193: Timeout budget can be exceeded on fallback session creation. The first POST to `/session` (with `parentID`) uses `timeoutMs`, and on failure the unparented retry performs a second POST with the same `timeoutMs` instead of remaining time. This can double the session-creation wait and violate overall request deadlines, especially under slow/unreachable OpenCode backends. Recompute remaining budget (or share one abort signal/deadline) before the retry.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `ef89a95a`

[Automated Review by pr-reviewer] I found one blocking timeout-budget regression in the new parent-session fallback path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In createSession(), the retry after a failed parented child-session POST reuses the original timeout window instead of recomputing remaining budget. The second POST can consume another full timeout interval, so a single session-creation step can exceed the caller deadline contract.

Ostico:
@NicholaiVogel no problem, :smile:

PR #542, opened 2026-04-21
fix(daemon): skip unchanged artifact reindex on cold start

### LINKED ISSUE #535: perf(daemon): reindexMemoryArtifacts blocks event loop for 18-20s on cold start
**Signet version:** 0.102.4
**Environment:** Linux 6.8.0-110-generic, Node v22.22.1, Bun 1.3.12
**Related issues:** #533, #534 (discovered in same monitoring session)

---

## Summary

`reindexMemoryArtifacts()` in `packages/daemon/src/memory-lineage.ts:561` performs a synchronous scan of the entire `memory/` directory tree on cold start. For a workspace with ~10K artifact files, the cold-start scan takes 18–20 seconds and blocks the event loop the entire time. This is a low-frequency event (2 occurrences in a full day of operation) but the impact per occurrence is severe — 18–20 seconds of complete unresponsiveness.

**Note:** Initial analysis incorrectly attributed ~400 daily `Event loop blocked` warnings to this function. Subsequent investigation showed that the vast majority of those warnings originate from the hook processing pipeline (see related issue). `reindexMemoryArtifacts` cold-start is responsible for only 2 of those events.

---

## Observed Behavior

142 reindex runs were logged in a single day. Only 2 were slow (cold starts); the remaining 140 completed in 14–120ms:

```
08:18:06 — duration=18516ms  fileCount=9828   (cold start after daemon launch)
11:38:50 — duration=20273ms  fileCount=10238  (cold start, likely after daemon restart)

All other 140 runs: 14–120ms (incremental, warm index)
```

The 2 cold-start runs produce `WARN [resources] Event loop blocked` warnings with lag values matching the scan duration. During these 18–20 second windows, all HTTP requests (health checks, hook callbacks, memory search) are blocked.

---

## Expected Behavior

Cold-start reindexing should not block the event loop. The daemon should remain responsive to HTTP requests, hook callbacks, and file watcher events during reindex operations. Target: no single blocking operation exceeds 100ms on the main thread.

---

## Steps to Reproduce

1. Accumulate ~10K artifact files in `$SIGNET_WORKSPACE/memory/` (normal usage over days/weeks).
2. Start the daemon: `bun --watch src/daemon.ts`
3. Observe log output for `reindexMemoryArtifacts completed` — the first run after startup will take 18–20s.
4. Subsequent incremental runs complete in 14–120ms.

---

## Root Cause Analysis

`reindexMemoryArtifacts()` at `memory-lineage.ts:561` performs a directory walk and file stat/read operations synchronously on the main thread during cold start. With ~10K files, this saturates the event loop for the full scan duration.

Incremental reindex (when the in-memory index is populated) is fast: 14–120ms. The problem is exclusively the cold-start full-scan path when the daemon has no prior index state.

---

## Suggested Fix

Options (not mutually exclusive):

1. **Chunked async iteration** — break the directory walk into batches (e.g., 100 files per tick), yielding control back to the event loop between batches via `setImmediate()` or `setTimeout(0)`.
2. **Worker thread** — move cold-start reindexing to a worker thread so the main thread stays responsive.
3. **Lazy/deferred reindex** — start serving immediately with an empty or stale index, reindex in the background.
4. **Persistent index** — persist the reindex state to disk so cold starts only need to process files added since the last run.

```typescript
// Example: chunked async iteration (option 1)
async function reindexMemoryArtifacts() {
  const files = await readdir(artifactDir, { recursive: true });
  const BATCH = 100;
  for (let i = 0; i < files.length; i += BATCH) {
    const batch = files.slice(i, i + BATCH);
    await Promise.all(batch.map(f => processArtifact(f)));
    await new Promise(r => setImmediate(r));  // yield to event loop
  }
}
```

---

## Impact

- **Low frequency** (2 per day), but **high severity per occurrence** (18–20s complete block).
- Hook callbacks arriving during a cold-start reindex will time out.
- Health checks fail during the window, which may trigger external monitors.

---

## Workaround

No workaround beyond keeping artifact count low or accepting the 18–20 second startup delay.

---

## Environment Details

- **Signet:** 0.102.4
- **OS:** Ubuntu, Linux 6.8.0-110-generic
- **Node:** v22.22.1
- **Bun:** 1.3.12
- **DB:** SQLite (bun:sqlite)
- **Date observed:** 2026-04-20

NicholaiVogel:
## Summary

Fix the cold-start `reindexMemoryArtifacts()` path so it does not re-read and re-parse every existing artifact when the database index is already current.

This persists the indexed source file mtime in `memory_artifacts`, seeds cold-cache reindex state from the DB, and only skips reread when the persisted mtime exactly matches the file on disk.

Closes #535.

## Root cause

The in-memory artifact index was treated as the only trusted change marker. On daemon restart, an empty cache plus existing `memory_artifacts` rows still seeded every known path with `mtime = 0`, which intentionally forced a full reread and re-upsert of every artifact on the first reindex pass.

That meant the persistent DB index existed, but cold start behaved as if it did not.

## What changed

- add migration `061` to persist `source_mtime_ms` on `memory_artifacts`
- write `source_mtime_ms` whenever artifact rows are inserted or updated
- on cold cache, seed existing DB-backed files with their real mtime only when `source_mtime_ms` exactly matches the current file mtime
- force a one-time refresh for upgraded rows where `source_mtime_ms` has not been populated yet
- still force reread for changed files, missing files, and brand-new files not yet present in the DB
- add regression coverage for the new cold-start behavior and migration

## Impact

Cold start should stop doing the dumb full reread/reparse storm for unchanged artifacts after restart. Rows from upgraded databases pay one refresh to populate exact mtimes, then subsequent cold starts can safely skip unchanged files.

This keeps the fix focused on the actual root of the observed startup stall without widening this into a larger async/worker refactor.

## PR Readiness Checklist

- [x] Migration is idempotent

Migration 061 adds a nullable `source_mtime_ms REAL` column to `memory_artifacts` and guards against duplicate application by checking existing columns first. This is backward-compatible for existing rows: upgraded rows with `NULL` mtimes are refreshed once during cold-cache reindex, then future cold starts can safely skip unchanged files. Rollback compatibility is straightforward because older code ignores the extra nullable column.

Daemon Rust parity is N/A for this PR because `reindexMemoryArtifacts()` and the rolling artifact DB index are TypeScript daemon behavior; no daemon-rs schema or runtime path is changed.

## Validation

Passed:

- `bun run build:core`
- `bun test packages/core/src/migrations/migrations.test.ts packages/daemon/src/memory-lineage.test.ts --test-name-pattern 'migration 061|cold cache re-reads artifacts when persisted mtime is missing and updated_at is stale|cold cache trusts persisted source_mtime_ms for unchanged artifacts|cold cache refreshes upgraded rows missing source_mtime_ms|cold cache still refreshes persisted mtime for files changed while the daemon was down|cold cache does not trust sub-second source_mtime_ms drift|cold cache still indexes new files when DB rows already exist'`
- `bunx biome check packages/core/src/migrations/index.ts packages/core/src/migrations/061-memory-artifact-source-mtime.ts packages/core/src/migrations/migrations.test.ts packages/daemon/src/memory-lineage.ts packages/daemon/src/memory-lineage.test.ts`
- `git diff --check`
- `bun run typecheck`

Note: `packages/daemon/src/memory-lineage.test.ts` still has pre-existing unrelated failures in the `memory_md_refs / Session Ledger` tests. The cold-start regression coverage added in this PR passes.

### DIFF
--- a/packages/daemon/src/memory-lineage.test.ts
+++ b/packages/daemon/src/memory-lineage.test.ts
@@ -1,5 +1,5 @@
 import { afterAll, beforeAll, beforeEach, describe, expect, it } from "bun:test";
-import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, statSync, utimesSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { Tiktoken } from "js-tiktoken/lite";
@@ -70,7 +70,7 @@ describe("memory-lineage", () => {
 		closeDbAccessor();
 		rmSync(dir, { recursive: true, force: true });
 		if (prev === undefined) {
-			delete process.env.SIGNET_PATH;
+			process.env.SIGNET_PATH = undefined;
 			return;
 		}
 		process.env.SIGNET_PATH = prev;
@@ -382,7 +382,7 @@ describe("memory-lineage", () => {
 			expect(row).toBeNull();
 		});
 
-		it("cold cache + existing DB rows → mtime=0 triggers re-read", async () => {
+		it("cold cache re-reads artifacts when persisted mtime is missing and updated_at is stale", async () => {
 			const agentId = "cache-cold";
 			const oldStamp = "2000-01-01T00:00:00.000Z";
 			await writeSummaryArtifact({
@@ -394,13 +394,17 @@ describe("memory-lineage", () => {
 				capturedAt: new Date().toISOString(),
 				startedAt: null,
 				endedAt: null,
-				summary: "Cold cache should trigger incremental refresh when DB already has rows.",
+				summary: "Cold cache should trigger incremental refresh when DB already has stale rows.",
 			});
 
 			getDbAccessor().withWriteTx((db) => {
-				db.prepare("UPDATE memory_artifacts SET updated_at = ? WHERE agent_id = ?").run(oldStamp, agentId);
+				db.prepare("UPDATE memory_artifacts SET source_mtime_ms = NULL, updated_at = ? WHERE agent_id = ?").run(
+					oldStamp,
+					agentId,
+				);
 			});
 
+			resetProjectionPurgeState();
 			reindexMemoryArtifacts(agentId);
 
 			const rows = getDbAccessor().withReadDb(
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): skip unchanged artifact reindex on cold start` (commit `85aeac79`) now and I'll follow up shortly with feedback.

packages/daemon/src/memory-lineage.test.ts
```diff
@@ -70,7 +70,7 @@ describe("memory-lineage", () => {
 		closeDbAccessor();
 		rmSync(dir, { recursive: true, force: true });
 		if (prev === undefined) {
-			delete process.env.SIGNET_PATH;
+			process.env.SIGNET_PATH = undefined;
```
PR-Reviewer-Ant: Teardown now sets `process.env.SIGNET_PATH = undefined` instead of deleting the env var. This can leave `SIGNET_PATH` as the literal string `"undefined"`, causing subsequent tests/processes to resolve paths under `undefined/...` and introducing flaky behavior. Use `delete process.env.SIGNET_PATH` when `prev` is absent.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `85aeac79`

[Automated Review by pr-reviewer] I found one correctness issue in the test changes; implementation otherwise aligns with the PR’s cold-start reindex goal.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed line in packages/daemon/src/memory-lineage.test.ts now assigns `process.env.SIGNET_PATH = undefined` during teardown. In Node/Bun env handling, this can leave the literal string "undefined" instead of removing the variable, which can poison later path resolution (`getAgentsDir()`) and cause cross-test contamination.

packages/daemon/src/memory-lineage.ts
```diff
@@ -534,10 +540,31 @@ function upsertArtifactRow(path: string, frontmatter: Record<string, unknown>, b
...
+				: Number.NaN;
+	if (!Number.isFinite(storedMtimeMs)) return false;
+
+	// Filesystem mtimes and serialized ISO timestamps can differ by a small
+	// amount across platforms and write paths. Treat sub-second deltas as the
+	// same indexed artifact state.
+	return Math.abs(currentMtimeMs - storedMtimeMs) <= 1_000;
+}
+
 function listCanonicalFiles(): string[] {
```
PR-Reviewer-Ant, line 564: Blocking: `canSeedColdCacheFromDbRow()` allows a ±1000ms window (`Math.abs(currentMtimeMs - storedMtimeMs) <= 1_000`) to treat files as unchanged. That can miss legitimate edits made shortly before/after shutdown when mtime delta is under 1s, causing cold-start reindex to skip reread and keep stale `memory_artifacts.content`. For persisted `source_mtime_ms`, exact equality should be required; tolerance should only be considered for legacy `updated_at` fallback (if at all).

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `677f78b0`

[Automated Review by pr-reviewer] I found one blocking correctness issue in the cold-start mtime trust logic; implementation otherwise matches the PR’s stated goal.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new cold-cache trust gate in packages/daemon/src/memory-lineage.ts explicitly treats mtimes within 1000ms as unchanged. Because source_mtime_ms is persisted from statSync(path).mtimeMs, a real file edit that advances mtime by <1000ms can be skipped on restart, leaving stale artifact content in memory_artifacts until a later change. This is directly introduced in the new diff and is not speculative.

packages/daemon/src/memory-lineage.ts
```diff
@@ -534,10 +540,30 @@ function upsertArtifactRow(path: string, frontmatter: Record<string, unknown>, b
...
+	}
+
+	const storedMtimeMs = typeof row.updated_at === "string" ? Date.parse(row.updated_at) : Number.NaN;
+	if (!Number.isFinite(storedMtimeMs)) return false;
+
+	// Filesystem mtimes and serialized ISO timestamps can differ by a small
+	// amount across platforms and write paths. Only allow tolerance for legacy
+	// upgraded rows that do not have source_mtime_ms populated yet.
+	return Math.abs(currentMtimeMs - storedMtimeMs) <= 1_000;
+}
```
PR-Reviewer-Ant, line 562: The legacy fallback path (`source_mtime_ms` missing) still treats `Math.abs(currentMtimeMs - storedMtimeMs) <= 1000` as unchanged. That can skip legitimate sub-second edits on cold start for upgraded rows, leaving stale `memory_artifacts` content until a later change. Consider tightening this fallback or forcing a one-time refresh for rows with null `source_mtime_ms`.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `63d65245`

[Automated Review by pr-reviewer] I found one correctness concern in the cold-start fallback path; otherwise the implementation matches the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new logic now correctly enforces exact mtime matching when `source_mtime_ms` is present, but still uses a ±1000ms tolerance when falling back to `updated_at`. That behavior is explicit in the changed function and can skip real sub-second file edits for upgraded rows that have not yet populated `source_mtime_ms`.

PR #546, opened 2026-04-22
feat(cli): add desktop source install command

NicholaiVogel:
## Summary

Adds the official CLI entrypoint for building and installing the Signet Electron desktop app from an existing source checkout.

`signet desktop build` resolves the local checkout, runs `bun install`, and then runs `bun run build:desktop`.

`signet desktop install` uses the same build path, then installs the newest Linux AppImage as a user-level launcher with a desktop entry. This gives Arch Linux a working native install path now while keeping macOS and Windows guarded until their native installer automation is wired.

## Details

- Adds `signet desktop build` and `signet desktop install` command registration.
- Adds checkout resolution through cwd, `SIGNET_SOURCE_DIR`, `--repo`, and the existing `~/signet/signetai` workspace path.
- Installs Linux artifacts under `~/.local/share/signet/desktop`, `~/.local/bin/signet-desktop`, and `~/.local/share/applications/signet.desktop`.
- Documents the command in `docs/CLI.md`.
- Updates the approved desktop packaging spec and dependency metadata so the source-build command is part of the contract.

## Validation

- `bun test packages/cli/src/features/desktop.test.ts packages/cli/src/commands/app.test.ts`
- `bun run build:core`
- `bun run --filter '@signet/cli' build:cli`
- `node packages/cli/dist/cli.js desktop install --help`

## Readiness Checklist

### DIFF
--- a/packages/cli/src/features/desktop.ts
+++ b/packages/cli/src/features/desktop.ts
@@ -12,7 +12,7 @@ import {
 	writeFileSync,
 } from "node:fs";
 import { homedir } from "node:os";
-import { basename, dirname, join, resolve } from "node:path";
+import { dirname, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 
 export interface DesktopCommandOptions {
--- a/packages/cli/src/features/desktop.ts
+++ b/packages/cli/src/features/desktop.ts
@@ -6,6 +6,7 @@ import {
 	lstatSync,
 	mkdirSync,
 	readdirSync,
+	readlinkSync,
 	rmSync,
 	statSync,
 	symlinkSync,
@@ -124,9 +125,11 @@ export function installDesktopFromSource(
 
 export function installLinuxDesktopApp(repo: string, home: string): DesktopLinuxInstallResult {
 	const releaseDir = desktopReleaseDir(repo);
-	const source = findNewestFile(releaseDir, (path) => path.endsWith(".AppImage"));
+	const source = findLinuxAppImage(releaseDir, process.arch);
 	if (!source) {
-		throw new Error(`No AppImage found in ${releaseDir}. Run signet desktop build first.`);
+		throw new Error(
+			`No matching Linux ${process.arch} AppImage found in ${releaseDir}. Run signet desktop build first.`,
+		);
 	}
 
 	const appDir = join(home, ".local", "share", "signet", "desktop");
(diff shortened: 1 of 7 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the feature work in `feat(cli): add desktop source install command` (commit `119693b2`) now and I'll follow up shortly with feedback.

packages/cli/src/features/desktop.ts
```diff
@@ -0,0 +1,246 @@
...
+				continue;
+			}
+			if (!entry.isFile() || !accept(path)) continue;
+			const mtime = statSync(path).mtimeMs;
+			if (!best || mtime > best.mtime) {
+				best = { path, mtime };
+			}
+		}
+	}
+	return best?.path ?? null;
```
PR-Reviewer-Ant, line 255: Potential data loss: `replaceSymlink` deletes any existing non-directory path at `~/.local/bin/signet-desktop` before creating the symlink. If a user already has a real file there (custom launcher/script), this command silently removes it. Prefer only replacing an existing symlink you own (or require `--force` for non-symlink files) to avoid destructive behavior.

packages/cli/src/features/desktop.ts
```diff
@@ -0,0 +1,246 @@
...
+		buildDesktopFromSource({ repo }, ctx);
+	}
+
+	const platform = ctx.platform ?? process.platform;
+	if (platform !== "linux") {
+		throw new Error(
+			`signet desktop install currently installs native launchers on Linux/Arch only. Build artifacts are in ${desktopReleaseDir(repo)}.`,
+		);
+	}
+
```
PR-Reviewer-Ant, line 123: Installer may choose the wrong AppImage when multiple `.AppImage` files exist: selection is recursive and purely by latest mtime. In mixed-output directories (e.g., multiple architectures or retained old artifacts), this can install an unintended binary. Filter by expected artifact naming/arch or prefer the known output path from the build step.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `119693b2`

[Automated Review by pr-reviewer] Found one blocking data-safety issue and one correctness risk in artifact selection.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The destructive overwrite behavior is directly visible in `replaceSymlink` (it unconditionally removes non-directory paths). The artifact-selection risk is directly visible in `findNewestFile` usage (mtime-only selection across all nested .AppImage files), but I did not reproduce with a multi-arch release directory in runtime.

packages/cli/src/features/desktop.ts
```diff
@@ -0,0 +1,266 @@
...
+	return { repo, releaseDir, appImage, binary, desktopEntry, icon };
+}
+
+function ancestorCandidates(path: string): string[] {
+	const out: string[] = [];
+	let current = resolve(path);
+	for (;;) {
+		out.push(current);
+		const parent = dirname(current);
```
PR-Reviewer-Ant, line 166: `isDesktopSourceCheckout` is too permissive for a path that will run `bun install` and `bun run build:desktop`. It currently accepts any tree containing `package.json` plus `packages/desktop/package.json`, which allows accidental or malicious non-Signet repos to be treated as valid and execute arbitrary install/build scripts. Add a stronger identity check (for example expected package name(s), repo marker, or workspace metadata) before running commands.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `495dd43d`

[Automated Review] I found one blocking security/correctness issue; prior artifact-selection and launcher-overwrite concerns appear addressed in this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new command path executes `bun install` in a directory accepted by `isDesktopSourceCheckout`, but that check only verifies file existence (`package.json` and `packages/desktop/package.json`) and does not verify repository identity. This is directly visible in the added `packages/cli/src/features/desktop.ts` logic and is sufficient to show unintended code execution risk if resolution lands on a non-Signet tree with matching shape.

packages/cli/src/features/desktop.ts
NicholaiVogel: Fixed in `495dd43d`. The installer now refuses to replace an existing non-symlink launcher, and only replaces symlinks that already point into Signet's desktop install directory. Added regression coverage for both the refusal path and the owned-symlink replacement path.

packages/cli/src/features/desktop.ts
NicholaiVogel: Fixed in `495dd43d`. AppImage selection now only considers top-level `Signet-*-linux-*.AppImage` artifacts matching the current Linux architecture aliases, instead of recursively choosing the newest `.AppImage` by mtime. Added a test covering wrong-arch and nested artifacts.

packages/cli/src/features/desktop.ts
NicholaiVogel: Fixed in `cc6ef8dc`. Checkout resolution now verifies the root package identity, expected workspaces, the desktop package identity, desktop main entry, and Electron app id before running `bun install` or `bun run build:desktop`. Added a lookalike-checkout regression test so we do not execute scripts from an arbitrary similarly-shaped tree.

PR #557, opened 2026-04-23
perf(pipeline): move extraction pipeline to dedicated worker thread

### LINKED ISSUE #556: perf(pipeline): offload extraction pipeline to dedicated worker thread
## Summary

Move the extraction pipeline worker (`startWorker()`) off the daemon's main event loop into a dedicated `node:worker_threads` worker thread, eliminating 50-141s event loop blocks caused by synchronous `bun:sqlite` write-lock contention under concurrent extraction jobs.

---

## Current Behavior

The extraction pipeline worker runs on the daemon's main thread. When multiple extraction jobs run concurrently, **SQLite write lock contention** causes cumulative event loop starvation measured up to 141 seconds.

**Root cause (validated by 4-agent deep analysis on `main` v0.108.0):**

1. **`bun:sqlite` is synchronous** — every `db.prepare().run()` and `db.exec("BEGIN IMMEDIATE")` blocks the Bun event loop until the SQLite call returns.
2. **SQLite WAL mode serializes writes** — concurrent readers are fine, but only one writer at a time. When a `BEGIN IMMEDIATE` can't acquire the write lock, SQLite's `busy_timeout=5000` causes bun:sqlite to **synchronously spin-wait for up to 5 seconds** on the main thread.
3. **`applyPhaseCWrites()` executes 300-550 SQL statements** per extraction job in a single `withWriteTx()` transaction (~200-500ms per job). `txPersistEntities()` adds another 70-90 SQL per 10 entities.
4. **Concurrent jobs compound** — when N extraction jobs overlap their sync DB phases, each job blocks the main thread while waiting for the write lock. With 10+ concurrent jobs: `N × (write_duration + busy_timeout_wait)` = tens of seconds to minutes of cumulative blocking.
5. **LLM provider calls are NOT the cause** — all 8 providers are fully async (verified: `fetch()` for HTTP, async `Bun.spawn()` for CLI, promise-based concurrency semaphore, no spin-wait).

**Evidence from daemon logs (2026-04-23):**
- 61 event loop blocks in a single day
- Largest block: 141,077ms
- Blocks correlate with `accessor.withWriteTx()` calls in `processExtractJob()`

**Additional contributor:** `summary-worker.ts:390` uses `writeFileSync()` for transcript persistence — large transcripts (10KB-100KB+) add sync file I/O stalls on the same main thread.

---

## Desired Behavior

The extraction pipeline runs in a dedicated `node:worker_threads` worker thread with its own `bun:sqlite` connection (WAL mode permits concurrent readers/writers). The main thread stays responsive for HTTP requests, health checks, hooks, and other pipeline stages regardless of extraction load.

Health endpoint (`/health`) should respond in <50ms even during heavy extraction concurrency — compared to current 50-141s stalls.

---

## Use Cases

**1. High-volume memory ingestion**
Multiple agents or rapid user activity generates 10+ concurrent extraction jobs. Currently the daemon becomes unresponsive for minutes. With threaded extraction, HTTP handlers and health checks remain fast.

**2. Production daemon stability**
Monitoring systems flag the daemon as unhealthy when event loop blocks exceed timeout thresholds. Moving extraction off the main thread eliminates the primary source of these blocks.

**3. Concurrent pipeline stages**
Summary, synthesis, and document workers share the main thread with extraction. Extraction's heavy DB writes starve all other stages. Isolating extraction allows all pipeline stages to progress independently.

---

### Architecture

```
MAIN THREAD (HTTP event loop)                  WORKER THREAD (extraction)
+-------------------------------------+       +-------------------------------------+
| Hono HTTP server (port 3850)        |       | Own bun:sqlite connection (WAL)     |
| Hook handlers                       |       | Own LLM provider instance           |
| File watcher                        |       | Own embedding fetcher               |
| Summary/synthesis/document workers  |<-IPC->|                                     |
| ExtractionThreadHandle (adapter)    |       | Full startWorker() polling loop     |
|   implements WorkerHandle interface |       |   processExtractJob()               |
|   wraps Worker + IPC               |       |   tick() + scheduleTick()           |
| Logger sink (receives IPC logs)     |       |   backoff, watchdog, stale reaper   |
| Telemetry sink (receives IPC events)|       |                                     |
+-------------------------------------+       | Logger -> IPC forwarder             |
                                               | Telemetry -> IPC forwarder          |
                                               +-------------------------------------+
```

The worker thread is fully self-contained — it opens its own `bun:sqlite` connection, creates its own LLM provider instance from the same config files (via inference router), and runs the existing `startWorker()` polling loop internally. IPC is only for control signals (stop/nudge) and log/telemetry forwarding — no data operations cross the thread boundary.

### IPC Protocol

```
Main -> Worker:
  { type: "stop" }      -> graceful shutdown
  { type: "nudge" }     -> reset backoff, repoll immediately

Worker -> Main:
  { type: "ready" }     -> worker initialized
  { type: "stopped" }   -> worker shut down cleanly
  { type: "stats" }     -> periodic stats snapshot (every 10s)
  { type: "log" }       -> forwarded logger call
  { type: "telemetry" } -> forwarded telemetry event
  { type: "analytics" } -> forwarded analytics call
  { type: "error" }     -> unrecoverable worker error
```

### Feature Flag

Opt-in via `agent.yaml`:

```yaml
pipelineV2:
  worker:
    threadedExtraction: true
```

Default: `false`. Zero behavior change when disabled. Instant rollback by setting to `false`.

### File Map

| File | Action |
|---|---|
| `packages/core/src/types.ts` | Add `threadedExtraction: boolean` to `PipelineWorkerConfig` |
| `packages/daemon/src/pipeline/extraction-thread-protocol.ts` | IPC message types, `WorkerInit` config |
| `packages/daemon/src/pipeline/extraction-thread.ts` | Worker thread entry point |
| `packages/daemon/src/pipeline/extraction-thread-handle.ts` | Main-thread `WorkerHandle` adapter |
| `packages/daemon/src/pipeline/worker.ts` | Make logger injectable via `LogSink` interface |
| `packages/daemon/src/pipeline/index.ts` | Conditionally start extraction thread |
| `packages/daemon/src/memory-config.ts` | Default `threadedExtraction: false` |
| `packages/daemon/src/daemon.ts` | Build `WorkerInit` and pass to `startPipeline()` |

### Prior Art

- **Synthesis worker thread** (`ostico/fd-diagnostic` branch): Successfully moved `renderMemoryProjection()` to a worker thread using same pattern: `initDbAccessorLite()` for lightweight DB init, `postMessage`/`on('message')` IPC, long-lived worker lifecycle.
- **POC** (`__tests__/worker-sqlite-poc.ts`): Validated `bun:sqlite` + `sqlite-vec` works in `node:worker_threads` with WAL mode.

---

## Expected Impact

- **Event loop starvation eliminated** — extraction's 300-550 SQL statements per job no longer block the HTTP server
- **Health endpoint <50ms** under any extraction load (currently 50-141s stalls)
- **Zero risk rollout** — feature flag defaults off, instant rollback

---

## Related

- Synthesis worker thread implementation on `ostico/fd-diagnostic` (same architectural pattern)
- `resource-monitor.ts:92-111` event loop lag detector (measurement mechanism)
- `worker-sqlite-poc.ts` POC validating bun:sqlite in worker threads

I'm working on this.

Ostico:
## Summary

Moves the extraction pipeline into a dedicated `node:worker_threads` worker thread behind a feature flag (`threadedExtraction`), eliminating 50–141s event loop blocks during LLM inference that currently degrade `/health` latency and HTTP responsiveness.

Closes #556

## Changes

- **IPC Protocol** (`extraction-thread-protocol.ts`): Discriminated union types for main↔worker communication (stop, nudge, ready, stopped, stats, log, telemetry, analytics, error)
- **Injectable LogSink** (`worker.ts`): Replaced 21 direct `logger.*` calls with an injected `LogSink` interface, enabling worker threads to proxy logs back to the main thread
- **Worker Entry Point** (`extraction-thread.ts`): Full bootstrap — receives `WorkerInit` via `workerData`, creates its own DB accessor, inference router, and LLM provider, then runs the existing `startWorker()` loop. Includes IPC-backed `TelemetryCollector` and `AnalyticsCollector` proxies that forward instrumentation calls to the main thread via `postMessage`
- **WorkerHandle Adapter** (`extraction-thread-handle.ts`): Main-thread adapter implementing the existing `WorkerHandle` interface — translates IPC messages to start/stop/nudge/drain semantics with proper lifecycle management. Dispatches incoming telemetry/analytics IPC messages to real collectors passed from `startPipeline()`
- **Feature Flag** (`threadedExtraction` in `PipelineWorkerConfig`): Defaults to `false` for zero behavior change; opt-in via `agent.yaml`
- **Integration Wiring** (`pipeline/index.ts` + `daemon.ts`): Conditional thread start with automatic fallback to main-thread `startWorker()` on failure
- **Tests** (`extraction-thread.test.ts`): 15 tests covering IPC protocol types, LogSink interface, and IPC-backed telemetry/analytics proxy behavior

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Behavior

| Flag | Behavior |
|------|----------|
| `threadedExtraction: false` (default) | Zero change — existing `startWorker()` on main thread |
| `threadedExtraction: true` | Extraction runs in worker thread, event loop stays free |
| Thread startup failure | Automatic fallback to main thread with error log |

## PR Readiness (MANDATORY)

## Testing

- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

## Verification

- TSC: 114 daemon errors (pre-existing baseline after core rebuild), 0 in core — **zero new errors introduced**
- New tests: 15/15 pass
- Existing worker tests: 22 pass / 23 pre-existing failures (LLM-dependent) — no regressions

## AI disclosure

## Files Changed

```
 packages/core/src/types.ts                                 |   2 +
 packages/daemon/src/__tests__/extraction-thread.test.ts    | 217 +++
 packages/daemon/src/daemon.ts                              |  19 ++
 packages/daemon/src/memory-config.ts                       |   2 +
 packages/daemon/src/pipeline/extraction-thread-handle.ts   | 133 +++
 packages/daemon/src/pipeline/extraction-thread-protocol.ts |  60 ++
 packages/daemon/src/pipeline/extraction-thread.ts          | 139 +++
 packages/daemon/src/pipeline/index.ts                      |  20 +-
 packages/daemon/src/pipeline/worker.ts                     |  49 +-
 9 files changed, 620 insertions(+), 21 deletions(-)
```

## Notes

- Feature flag `threadedExtraction` defaults to `false` — zero behavior change until opted in
- Worker thread creates its own DB accessor and inference router from serialized config (no shared mutable state across thread boundary)
- Analytics and telemetry are not proxied to the worker thread (pass `undefined`)
- Three review rounds addressed: snake_case serialization, startup race conditions, worker path resolution, exit-before-ready promise rejection, duplicate startPipeline guard


PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the changes in `perf(pipeline): move extraction pipeline to dedicated worker thread` (commit `8505d5c2`) now and I'll follow up shortly with feedback.

packages/daemon/src/daemon.ts
```diff
@@ -1136,6 +1137,23 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
...
+			? {
+					dbPath: MEMORY_DB,
+					vecExtensionPath: getVectorRuntimeStatus().extensionPath ?? "",
+					agentsDir: AGENTS_DIR,
+					agentId: defaultAgentId,
+					embeddingConfig: {
+						provider: memoryCfg.embedding.provider,
+						model: memoryCfg.embedding.model,
+						dimensions: memoryCfg.embedding.dimensions ?? 768,
+						baseUrl: memoryCfg.embedding.base_url,
```
PR-Reviewer-Ant: Blocking: `workerInit.embeddingConfig` is serialized with `baseUrl` (and no `apiKey`), but worker code later casts this object to `EmbeddingConfig` (which uses `base_url`/`api_key`). In threaded mode this can drop endpoint/auth settings and change runtime behavior versus non-threaded mode.

packages/daemon/src/pipeline/index.ts
```diff
@@ -169,7 +172,22 @@ export function startPipeline(
 		fetchEmbedding,
 	};
 
-	workerHandle = startWorker(accessor, provider, pipelineCfg, decisionCfg, analytics, telemetry);
+	if (pipelineCfg.worker.threadedExtraction && workerInit) {
```
PR-Reviewer-Ant, line 181: Blocking race: `startExtractionThread(workerInit)` is async and not cancellation-aware. If `stopPipeline()` runs before the promise settles, the `.then()`/`.catch()` can still assign `workerHandle` (or start fallback `startWorker`) after shutdown, leaving pipeline work running unexpectedly.

packages/daemon/src/pipeline/extraction-thread-handle.ts
```diff
@@ -0,0 +1,133 @@
...
+import { logger } from "../logger";
+import type { LogCategory } from "../logger";
+import type { WorkerInit, WorkerToMainMessage } from "./extraction-thread-protocol";
+import type { WorkerHandle, WorkerStats } from "./worker";
+
+const READY_TIMEOUT_MS = 30_000;
+const STOP_TIMEOUT_MS = 10_000;
+
+export function startExtractionThread(init: WorkerInit): Promise<WorkerHandle> {
+	return new Promise<WorkerHandle>((resolve, reject) => {
```
PR-Reviewer-Ant, line 44: Warning: worker entry path is hardcoded to `extraction-thread.ts`. In built/runtime environments that execute transpiled JS, this can fail to spawn the worker unless a TS loader is present. `daemon.ts` already uses a `.js`-first fallback pattern for synthesis workers; extraction should do the same for parity.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `8505d5c2`

[Automated Review by pr-reviewer] Found blocking correctness issues in threaded extraction initialization and lifecycle management.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Two issues are directly provable from the changed code: (1) `WorkerInit.embeddingConfig` uses camelCase keys (`baseUrl`/`apiKey`) while downstream code casts to `EmbeddingConfig` expecting snake_case (`base_url`/`api_key`), which breaks provider config in threaded mode; (2) threaded startup is async and uncancelled, so a late resolve/reject can assign/start a worker after `stopPipeline()` has already run.

packages/daemon/src/pipeline/extraction-thread-handle.ts
```diff
@@ -0,0 +1,135 @@
...
+				case "analytics":
+					break;
+			}
+		});
+
+		worker.on("error", (err: Error) => {
+			clearTimeout(readyTimer);
+			logger.error("pipeline", "Extraction worker thread crashed", err);
+			running = false;
+			reject(err);
```
PR-Reviewer-Ant: Startup failure does not reliably trigger immediate fallback. If the worker exits before sending `ready` (for example, bootstrap catches and `process.exit(1)`), this handler only logs and sets `running = false`; it does not clear `readyTimer` or reject the startup promise. `startPipeline()` then waits until the 30s ready timeout before falling back, which contradicts the PR’s stated automatic fallback on startup failure and can delay daemon readiness.

packages/daemon/src/pipeline/index.ts
```diff
@@ -124,6 +127,7 @@ export function startPipeline(
 	providerTracker?: ProviderTracker,
 	analytics?: AnalyticsCollector,
 	telemetry?: TelemetryCollector,
+	workerInit?: WorkerInit,
 ): void {
 	if (workerHandle) {
```
PR-Reviewer-Ant, line 133: Race condition: `startPipeline()` only guards on `workerHandle`, but threaded startup is async and leaves `workerHandle` null until `pendingStartup` resolves. A second `startPipeline()` call during that window can start another extraction thread (or duplicate fallback worker), causing duplicate workers and inconsistent lifecycle state. Add a `pendingStartup` guard at entry (similar to stop handling) to make startup single-flight.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `7fba0cba`

[Automated Review by pr-reviewer] Found two blocking lifecycle/race issues in the new threaded extraction startup path that can break the PR’s claimed automatic fallback behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the new startup/worker lifecycle code: startup failures can hang until timeout because `exit`/worker-reported error paths do not reject before `ready`, and `startPipeline()` has no guard for `pendingStartup`, allowing concurrent starts when `workerHandle` is still null.

packages/daemon/src/daemon.ts
Ostico: Fixed in 7fba0cba. `SerializedEmbeddingConfig` now uses snake_case `base_url`/`api_key` matching `EmbeddingConfig`, so no silent key loss across the thread boundary.

packages/daemon/src/pipeline/index.ts
Ostico: Fixed in 7fba0cba. `pendingStartup` promise is now tracked in module state. `stopPipeline()` awaits it before proceeding, so a stop during async startup waits for the thread to become ready (or fail), then terminates it cleanly. No orphan thread possible.

packages/daemon/src/pipeline/extraction-thread-handle.ts
Ostico: Fixed in 7fba0cba. Worker entry path now checks for bundled `extraction-thread.js` first via `existsSync()`, falling back to `.ts` for dev. Uses `import.meta.dir` for correct resolution regardless of CWD.

packages/daemon/src/pipeline/extraction-thread-handle.ts
Ostico: Fixed in 753c2fce. The exit handler now calls `clearTimeout(readyTimer)` and rejects the startup promise immediately via a `settled` flag. Worker exit before `ready` (e.g. bootstrap `process.exit(1)`) now fails fast instead of hanging 30s.

packages/daemon/src/pipeline/index.ts
Ostico: Fixed in 753c2fce. `startPipeline()` now checks `pendingStartup` at entry alongside the existing `workerHandle` guard, so a second call during async startup is a no-op with a warning log.

packages/daemon/src/pipeline/extraction-thread.ts
```diff
@@ -0,0 +1,139 @@
...
+			timeoutMs: pipelineCfg.extraction.timeout,
+			fetchEmbedding: (text: string, cfg: EmbeddingConfig) => fetchEmbedding(text, cfg),
+		};
+
+		// 4. Start extraction worker
+		//    Analytics and telemetry are omitted — the main-thread adapter
+		//    can record its own metrics based on IPC stats messages.
+		const handle = startWorker(
+			accessor,
+			provider,
```
PR-Reviewer-Ant, line 147: Threaded mode drops extraction instrumentation: `startWorker(...)` is called with `analytics` and `telemetry` unset, unlike the existing main-thread path. This creates behavior drift (missing metrics/events) when `threadedExtraction` is enabled.

packages/daemon/src/pipeline/extraction-thread-handle.ts
```diff
@@ -0,0 +1,148 @@
...
+					break;
+				}
+
+				case "error":
+					logger.error("pipeline", "Extraction worker thread error", undefined, {
+						error: msg.error,
+						stack: msg.stack,
+					});
+					break;
+
```
PR-Reviewer-Ant, line 112: IPC protocol includes `telemetry` and `analytics`, but the adapter currently discards both message types. Together with the worker-side omission, this means observability data is not preserved in threaded mode, diverging from the PR’s stated protocol/behavior parity expectations.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `753c2fce`

[Automated Review by pr-reviewer] The threaded extraction path appears functionally wired, but it introduces an observability regression versus the existing main-thread pipeline: extraction analytics/telemetry are dropped in threaded mode despite being part of the stated protocol.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In the new threaded path, `startWorker(...)` is invoked with `analytics` and `telemetry` as `undefined` in `extraction-thread.ts`, and the main-thread adapter explicitly ignores `telemetry`/`analytics` IPC messages. This is directly visible in the diff and changes runtime behavior for instrumentation when `threadedExtraction=true`.

packages/daemon/src/pipeline/extraction-thread.ts
Ostico: Fixed in ff505d60. Worker thread now creates IPC-backed `TelemetryCollector` and `AnalyticsCollector` proxy objects that forward `record`/`recordError`/`recordLatency` calls to the main thread via `postMessage`. The handle adapter dispatches incoming IPC messages to the real collectors passed from `startPipeline()`. Query/lifecycle methods (`getUsage`, `flush`, `start`, `stop`, etc.) are no-ops on the worker side since the main thread owns those concerns.

packages/daemon/src/pipeline/extraction-thread-handle.ts
Ostico: Fixed in ff505d60. `startExtractionThread()` now accepts optional `analytics` and `telemetry` collectors via `ExtractionThreadOpts`. The message handler dispatches `telemetry` IPC messages to `telemetry.record()` and `analytics` messages to the matching method on the real `AnalyticsCollector`. `startPipeline()` threads both collectors through to the handle.

packages/daemon/src/pipeline/extraction-thread.ts
```diff
@@ -0,0 +1,181 @@
...
+		const merged = errStr ? { ...data, errorMessage: errStr } : data;
+		send({ type: "log", level: "error", category, message, data: merged });
+	},
+};
+
+// ---------------------------------------------------------------------------
+// IPC proxy collectors — forward analytics/telemetry to main thread
+// ---------------------------------------------------------------------------
+
+/** Proxy that forwards telemetry.record() calls to the main thread via IPC. */
```
PR-Reviewer-Ant, line 69: PR notes say analytics/telemetry are not proxied to the worker thread, but this implementation does proxy both via IPC (`ipcTelemetry` / `ipcAnalytics` and passing them into `startWorker`). This is a behavior/documentation mismatch and can materially affect performance characteristics of the threaded mode (main thread still processes worker-originated instrumentation traffic).

packages/daemon/src/__tests__/extraction-thread.test.ts
```diff
@@ -0,0 +1,217 @@
+import { describe, expect, it } from "bun:test";
```
PR-Reviewer-Ant, line 1: The new tests validate TypeScript shape assignments only; they do not instantiate a real worker thread or verify startup timeout, stop handshake, fallback-to-main-thread path, or shutdown ordering with `pendingStartup`. Given the concurrency/lifecycle changes, this leaves key regression risks unguarded.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `836404cc`

[Automated Review by pr-reviewer] I found non-blocking but important gaps: implementation/details diverge from stated PR notes, and the new threading path lacks behavioral tests for lifecycle/fallback correctness.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff clearly shows telemetry/analytics are actively proxied from worker to main thread in extraction-thread.ts, despite the PR note claiming they are not proxied. Test additions are type-shape checks only and do not exercise real Worker startup/stop/fallback behavior, so lifecycle/race correctness remains unproven from this PR alone.

packages/daemon/src/pipeline/extraction-thread.ts
Ostico: Good catch — the PR description was stale from before ff505d60 added IPC proxying. Updated the PR body to accurately reflect that both telemetry and analytics are now proxied from worker to main thread via postMessage. The main-thread cost is minimal (message dispatch to existing collector methods), not LLM inference work.

packages/daemon/src/__tests__/extraction-thread.test.ts
Ostico: Agreed — the current tests are shape/contract checks, not behavioral. Adding lifecycle tests that exercise real Worker startup, stop handshake, and fallback requires a running LLM provider and DB, so they belong in an integration test suite rather than unit tests. Will add a follow-up with integration tests gated behind an env flag for CI feasibility.

NicholaiVogel (approved):
Follow-up fixes landed in ac578720. The extraction worker is now included in the daemon build output, and the worker-thread handle has lifecycle coverage for ready, stats, nudge, clean stop, early exit, ready timeout, and stop timeout.

Local validation:
- bun test packages/daemon/src/__tests__/extraction-thread.test.ts
- bunx biome check packages/daemon/build.ts packages/daemon/src/pipeline/extraction-thread-handle.ts packages/daemon/src/__tests__/extraction-thread.test.ts packages/daemon/src/pipeline/worker.ts
- cd packages/daemon && bun run build

GitHub checks are green.

NicholaiVogel:
@Ostico Thank you for this one, this was a really thoughtful change.

Ostico:
You're welcome @NicholaiVogel , i keep using and improving as I can.

PR #559, opened 2026-04-24
fix(daemon): bound prompt-submit embedding latency

NicholaiVogel:
## Summary
- Bound UserPromptSubmit embedding lookup to 1000ms so the blocking hook can fail open before harness timeouts.
- Preserve vector-backed prompt recall when embeddings return quickly, while falling back to non-vector recall when they stall.
- Propagate abort signals into embedding provider fetches so timed-out prompt-submit HTTP embedding calls are cancelled instead of leaving the in-flight latch stuck.
- Add regression coverage for fast embedding injection, slow embedding bypass behavior, non-vector fallback injection, and vector recall recovery after a timed-out embedding abort.

## Validation
- [x] bun test packages/daemon/src/hooks.prompt-submit.test.ts packages/daemon/src/embedding-fetch.test.ts (24 pass)
- [x] bunx biome check packages/daemon/src/hooks.ts packages/daemon/src/hooks.prompt-submit.test.ts packages/daemon/src/embedding-fetch.ts packages/daemon/src/embedding-fetch.test.ts
- [x] git diff --check
- [ ] bun scripts/spec-deps-check.ts — blocked by existing spec dependency issue: `informed_by path does not exist for engram-informed-predictor-track: references/Engram`
- [ ] bun run --cwd packages/daemon build:types — blocked by existing package setup/type issues, including core migration rootDir errors, runtime provider name typing, GraphIQ MCP typing, and readonly test config errors.

## PR checklist

## Notes
- This is a narrow hook reliability bug fix. It does not change API, schema, admin/mutation endpoints, or spec-governed behavior, so docs/spec changes are not required.
- Full workspace typecheck/spec validation remain blocked by unrelated existing issues listed above; targeted tests and lint for the touched files pass.

### DIFF
--- a/packages/daemon/src/embedding-fetch.test.ts
+++ b/packages/daemon/src/embedding-fetch.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, expect, it, mock } from "bun:test";
 import { fetchEmbedding, requiresOpenAiApiKey, setNativeFallbackProvider } from "./embedding-fetch";
 
 const originalFetch = globalThis.fetch;
+const originalOpenAiApiKey = process.env.OPENAI_API_KEY;
 
 describe("requiresOpenAiApiKey", () => {
 	it("requires a key for official OpenAI endpoints", () => {
@@ -21,7 +22,11 @@ describe("fetchEmbedding", () => {
 	afterEach(() => {
 		globalThis.fetch = originalFetch;
 		setNativeFallbackProvider(null);
-		delete process.env.OPENAI_API_KEY;
+		if (originalOpenAiApiKey === undefined) {
+			Reflect.deleteProperty(process.env, "OPENAI_API_KEY");
+		} else {
+			process.env.OPENAI_API_KEY = originalOpenAiApiKey;
+		}
 	});
 
 	it("allows keyless requests for custom OpenAI-compatible endpoints", async () => {
@@ -42,6 +47,29 @@ describe("fetchEmbedding", () => {
 		expect(capturedHeaders).toEqual({
 			"Content-Type": "application/json",
 		});
+	});
+
+	it("passes caller abort signals through to provider fetches", async () => {
+		const controller = new AbortController();
+		let capturedSignal: AbortSignal | null | undefined;
+		globalThis.fetch = mock((_url: string | URL | Request, init?: RequestInit) => {
+			capturedSignal = init?.signal;
+			return Promise.resolve(Response.json({ data: [{ embedding: [0.1, 0.2, 0.3] }] }));
+		}) as unknown as typeof fetch;
+
+		const result = await fetchEmbedding(
+			"hello",
+			{
+				provider: "openai",
+				model: "text-embedding-3-small",
+				dimensions: 3,
+				base_url: "http://localhost:1234/v1",
+			},
+			{ signal: controller.signal },
+		);
+
+		expect(result).toEqual([0.1, 0.2, 0.3]);
+		expect(capturedSignal).toBe(controller.signal);
 	});
 
 	it("routes to ollama when nativeFallbackProvider is 'ollama'", async () => {
--- a/packages/daemon/src/embedding-fetch.test.ts
+++ b/packages/daemon/src/embedding-fetch.test.ts
@@ -49,11 +49,15 @@ describe("fetchEmbedding", () => {
 		});
 	});
 
-	it("passes caller abort signals through to provider fetches", async () => {
+	it("composes caller abort signals with provider timeouts", async () => {
 		const controller = new AbortController();
 		let capturedSignal: AbortSignal | null | undefined;
 		globalThis.fetch = mock((_url: string | URL | Request, init?: RequestInit) => {
 			capturedSignal = init?.signal;
+			expect(capturedSignal).not.toBe(controller.signal);
+			expect(capturedSignal?.aborted).toBe(false);
+			controller.abort();
+			expect(capturedSignal?.aborted).toBe(true);
 			return Promise.resolve(Response.json({ data: [{ embedding: [0.1, 0.2, 0.3] }] }));
 		}) as unknown as typeof fetch;
 
@@ -69,7 +73,7 @@ describe("fetchEmbedding", () => {
 		);
 
 		expect(result).toEqual([0.1, 0.2, 0.3]);
-		expect(capturedSignal).toBe(controller.signal);
+		expect(capturedSignal?.aborted).toBe(true);
 	});
 
 	it("routes to ollama when nativeFallbackProvider is 'ollama'", async () => {
--- a/packages/daemon/src/embedding-fetch.ts
+++ b/packages/daemon/src/embedding-fetch.ts
@@ -52,18 +52,62 @@ type EmbeddingFetchOptions = {
 	readonly signal?: AbortSignal;
 };
 
+type EmbeddingFetchSignal = {
+	readonly signal: AbortSignal;
+	readonly cleanup: () => void;
+};
+
+function createEmbeddingFetchSignal(opts: EmbeddingFetchOptions, timeoutMs: number): EmbeddingFetchSignal {
+	const controller = new AbortController();
+	const timer = setTimeout(() => controller.abort(), timeoutMs);
+	let abortCaller: (() => void) | null = null;
+	if (opts.signal) {
+		if (opts.signal.aborted) {
+			controller.abort();
+		} else {
+			abortCaller = () => controller.abort();
+			opts.signal.addEventListener("abort", abortCaller, { once: true });
+		}
+	}
+	return {
+		signal: controller.signal,
+		cleanup: () => {
+			clearTimeout(timer);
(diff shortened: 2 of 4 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): bound prompt-submit embedding latency` (commit `06d327b9`) now and I'll follow up shortly with feedback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `06d327b9`

[Automated Review by pr-reviewer] I found one blocking reliability issue and one non-blocking coverage gap against the PR’s stated behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The timeout wrapper in packages/daemon/src/hooks.ts clearly races `fetchEmbedding` against a timer but does not cancel the underlying embedding request. If the provider hangs, timed-out requests continue accumulating in the background while the hook returns, which can degrade daemon resources under repeated prompt-submit traffic. This behavior is directly visible in the new helper implementation and is not hypothetical from unseen modules.

packages/daemon/src/hooks.ts
```diff
@@ -2466,6 +2466,26 @@ type UserPromptSubmitDeps = {
...
+const PROMPT_SUBMIT_EMBEDDING_TIMEOUT_MS = 1000;
+
+async function fetchPromptSubmitEmbedding(
+	deps: Pick<UserPromptSubmitDeps, "fetchEmbedding" | "logger">,
+	text: string,
+	cfg: Parameters<typeof fetchEmbedding>[1],
+): Promise<number[] | null> {
+	let timer: ReturnType<typeof setTimeout> | null = null;
+	try {
+		return await Promise.race([
```
PR-Reviewer-Ant, line 2480: Blocking: timed-out embedding calls are not cancelled. `Promise.race` returns after 1000ms, but the original `deps.fetchEmbedding(...)` continues running indefinitely if the provider stalls. Under repeated prompt submits this can accumulate stuck in-flight requests and cause resource exhaustion/latency amplification. Please add cancellation/abort propagation (or bounded concurrency/single-flight guard) so timed-out calls are actually terminated.

packages/daemon/src/hooks.prompt-submit.test.ts
```diff
@@ -275,6 +282,87 @@ describe("handleUserPromptSubmit observability", () => {
...
+		expect(fetchEmbeddingMock).toHaveBeenCalledTimes(1);
+		expect(result.memoryCount).toBe(1);
+		expect(result.inject).toContain("prompt submit timeout trace uses fast vector context");
+	});
+
+	it("bypasses prompt-submit embeddings when they exceed the hook budget", async () => {
+		fetchEmbeddingMock.mockImplementationOnce(async () => new Promise<number[] | null>(() => {}));
+		hybridRecallMock.mockImplementationOnce(async (_params, _cfg, embed) => {
+			const vector = await embed("prompt submit timeout trace", {
+				provider: "ollama",
```
PR-Reviewer-Ant, line 404: Warning: the PR description claims fallback to non-vector recall when embeddings stall, but the new slow-path test only verifies `memoryCount === 0` with a vector-dependent mock. It does not assert that lexical/non-vector recall still returns results when embeddings time out. Add a regression test that returns non-vector hits with `embed => null` and verifies those hits are injected.

packages/daemon/src/hooks.ts
```diff
@@ -2466,6 +2466,40 @@ type UserPromptSubmitDeps = {
 	readonly trackFtsHits: typeof trackFtsHits;
 };
 
+const PROMPT_SUBMIT_EMBEDDING_TIMEOUT_MS = 1000;
+let promptSubmitEmbeddingInFlight: Promise<number[] | null> | null = null;
```
PR-Reviewer-Ant: Blocking: the global `promptSubmitEmbeddingInFlight` latch can get stuck forever. The timeout only resolves `Promise.race`, but the underlying `deps.fetchEmbedding(...)` is not cancelled; latch reset depends on `request.finally()`. If that provider call never settles, all future prompt-submit requests permanently skip vector recall (`already in flight` path), causing long-lived recall degradation rather than bounded fallback.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `f0685e58`

[Automated bot review by pr-reviewer] I found one blocking reliability issue in the timeout/serialization logic; the new tests do improve coverage of the claimed fallback behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In `packages/daemon/src/hooks.ts`, `fetchPromptSubmitEmbedding` races with a 1000ms timeout but does not cancel the underlying provider call, and `promptSubmitEmbeddingInFlight` is only cleared in `request.finally()`. If the provider hangs indefinitely, that global in-flight latch never clears, so all subsequent prompt-submit calls skip vector recall indefinitely. This directly contradicts the intended bounded degradation and is observable from the changed helper and call-site wiring.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `a5e4f8f3`

[Automated bot review by pr-reviewer] The timeout/abort behavior described in the PR is now implemented and covered, but I found one test-isolation regression worth fixing.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new hook path now aborts timed-out embedding calls and propagates signals into provider fetches, matching the PR claims. However, in the updated test teardown, OPENAI_API_KEY restoration uses direct assignment even when the original value is undefined, which can leave the env var effectively set to a non-empty value and contaminate subsequent tests.

packages/daemon/src/embedding-fetch.test.ts
```diff
@@ -21,7 +22,7 @@ describe("fetchEmbedding", () => {
 	afterEach(() => {
 		globalThis.fetch = originalFetch;
 		setNativeFallbackProvider(null);
-		delete process.env.OPENAI_API_KEY;
+		process.env.OPENAI_API_KEY = originalOpenAiApiKey;
```
PR-Reviewer-Ant: Test isolation regression: `process.env.OPENAI_API_KEY = originalOpenAiApiKey` is unsafe when the original value is `undefined`. In Node/Bun env handling, this can leave a truthy string value and make later tests behave as if an API key exists. Restore with conditional delete (`if (originalOpenAiApiKey === undefined) delete process.env.OPENAI_API_KEY; else process.env.OPENAI_API_KEY = originalOpenAiApiKey;`).

packages/daemon/src/hooks.ts
```diff
@@ -2466,6 +2466,47 @@ type UserPromptSubmitDeps = {
...
```
PR-Reviewer-Ant: Blocking: `promptSubmitEmbeddingInFlight` is a single global latch, not keyed by session/request. While one embedding is in flight, all concurrent prompt-submit calls immediately skip vector recall (`return null`), even if the provider is healthy and would return quickly. This diverges from the PR claim to preserve vector-backed recall when embeddings are fast and can cause broad recall-quality regression under concurrent load.

packages/daemon/src/embedding-fetch.ts
```diff
@@ -48,12 +48,21 @@ export function setNativeFallbackProvider(
...
+	model: string,
+	opts: EmbeddingFetchOptions = {},
+): Promise<number[] | null> {
 	const res = await fetch(`${baseUrl.replace(/\/$/, "")}/api/embeddings`, {
 		method: "POST",
 		headers: { "Content-Type": "application/json" },
 		body: JSON.stringify({ model, prompt: text }),
-		signal: AbortSignal.timeout(30000),
+		signal: opts.signal ?? AbortSignal.timeout(30000),
 	});
```
PR-Reviewer-Ant: Warning: provider fetches now use `opts.signal ?? AbortSignal.timeout(...)`. When a caller supplies a signal, the built-in timeout is removed rather than combined, so requests can run indefinitely if that signal is never aborted. Consider composing signals so caller cancellation is propagated without dropping the per-provider timeout floor.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `52f0d3b6`

[Automated Review by pr-reviewer] I found one blocking correctness/regression issue and one reliability concern in the timeout/abort wiring.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The new global in-flight gate in hooks.ts is directly visible and unconditionally skips vector recall for concurrent prompt-submit calls, which conflicts with the stated goal to preserve vector recall when embeddings are fast. The timeout-replacement pattern in embedding-fetch.ts is also explicit in the diff, but cross-module impact depends on other callers that may pass a non-aborting signal.

PR #588, opened 2026-04-28
fix(daemon): prevent OOM crash-loop in transcript backfill

### LINKED ISSUE #587: Daemon OOM crash loop: ensureCanonicalTranscriptHistory reads+rewrites entire transcript.jsonl 6,650 times on every restart
**Signet version:** 0.109.3
**Environment:** Linux 6.8.0-110-generic (Ubuntu), Bun 1.3.12, Node v22.22.1

---

## Summary

The daemon OOM-kills deterministically ~2 hours after startup. `ensureCanonicalTranscriptHistory()` in `session-transcripts.ts` triggers a full backfill of **6,650 transcript sources** (4,828 markdown files + 1,822 DB rows) on the first hook call after every restart. Each backfill iteration calls `writeCanonicalTranscriptSnapshot()`, which reads the **entire** growing `transcript.jsonl` via `readFileSync`, parses all records as JSON, filters, re-serializes, and rewrites the whole file. The guard against re-running (`canonicalBackfills` Set) is in-memory and resets on every restart, creating a deterministic crash loop.

---

## Observed Behavior

Daemon starts at ~200MB RSS, stable for ~2 hours, then a single 30-second burst allocates ~9GB. Linux OOM-killer terminates the process. Three consecutive instances show identical pattern.

**Kernel log (journalctl -k):**
```
Out of memory: Killed process 3909310 (bun) total-vm:99049768kB, anon-rss:12376224kB
```

**RSS monitoring (30-second samples):**
```
11:42:54,209320    # 204MB - stable
11:43:24,208356    # 203MB - stable
11:43:54,9466680   # 9.2GB - burst
11:44:24,10832296  # 10.6GB - still growing
11:44:38           # OOM killed at ~12.7GB
```

**`transcript.jsonl` last modified at 11:44** — during the OOM spike, confirming active write.

**Daemon log has ZERO entries after startup** — file logger flush broken or backfill monopolizes event loop for the entire 2-hour duration.

---

## Expected Behavior

The daemon should not re-backfill transcripts that were already backfilled in a previous daemon lifecycle. The backfill-complete marker should persist across restarts. Reading/writing `transcript.jsonl` should not load the entire file into memory.

---

## Steps to Reproduce

1. Have a workspace with thousands of transcript artifacts (4,828 `*--transcript.md` files, 1,822 `session_transcripts` DB rows)
2. Start the daemon: `signet start` (or `bun platform/daemon/src/daemon.ts`)
3. Trigger any hook call (e.g., open an opencode/claude-code session)
4. Wait ~2 hours — daemon OOM-kills at 12+ GB RSS
5. Restart → repeat (crash loop)

---

## Root Cause Analysis

The failure chain:

1. **`ensureCanonicalTranscriptHistory()`** (`session-transcripts.ts:145`) runs on every first hook call per daemon lifecycle. The guard is an in-memory `Set<string>` (`canonicalBackfills`) that resets on restart.

2. **`backfillMarkdownTranscriptArtifacts()`** iterates **4,828** `*--transcript.md` files. For each, calls `writeCanonicalTranscriptSnapshot()`.

3. **`backfillDatabaseTranscripts()`** loads **all 1,822 rows** (46.7MB) from `session_transcripts` via `.all()`, then iterates each calling `writeCanonicalTranscriptSnapshot()`.

4. **`writeCanonicalTranscriptSnapshot()`** (`transcript-jsonl.ts:336`) does:
   ```ts
   const existing = readRecords(path).filter((record) => !sameSession(record, input));
   // ...
   writeRecords(path, [...existing, ...next]);
   ```

5. **`readRecords()`** (`transcript-jsonl.ts:143`) does:
   ```ts
   for (const line of readFileSync(path, "utf8").split(/\r?\n/)) {
       // JSON.parse each of 149K lines
   }
   ```

6. **Memory profile per late-stage iteration** (when `transcript.jsonl` is ~200MB):
   - `readFileSync`: 212MB string
   - `.split()`: 149K substring references
   - `JSON.parse` × 149K: ~300-500MB in parsed objects
   - `.filter()`: new array of ~same size
   - `[...existing, ...next]`: another copy
   - `writeRecords` → `.map(JSON.stringify).join("\n")`: another 212MB string
   - **Peak per iteration: ~1-1.5GB**

7. **6,650 iterations** with GC unable to reclaim fast enough → heap balloons to 9-12GB → OOM kill.

**Data on disk:**
- 4,828 markdown transcript files totaling 588MB
- 1,822 DB rows totaling 46.7MB (max single row: 2.5MB)
- `transcript.jsonl`: 212MB, 149,034 lines

---

## Suggested Fix

**1. Persist backfill-complete marker** (immediate fix):
```ts
// In ensureCanonicalTranscriptHistory():
const markerPath = join(basePath, "memory", ".transcript-backfill-done");
if (existsSync(markerPath)) {
  canonicalBackfills.add(key);
  return;
}
// ... run backfill ...
writeFileSync(markerPath, new Date().toISOString(), "utf8");
canonicalBackfills.add(key);
```

**2. Stream-read `transcript.jsonl`** instead of `readFileSync` + `split`:
```ts
// Replace readRecords() with a line-by-line reader
// or use readTailRecords() which already reads only the last 256KB
```

**3. Avoid full-file rewrite per iteration** — `writeCanonicalTranscriptSnapshot` should append-only (like `appendCanonicalTranscriptTurns` already does) instead of read-filter-rewrite.

**4. Don't load all DB rows at once** in `backfillDatabaseTranscripts`:
```ts
// Use LIMIT/OFFSET pagination instead of .all()
```

---

## Impact

- **Daemon is unusable** on any workspace with >1,000 transcript artifacts
- **Crash loop**: every restart triggers the same backfill → same OOM
- **Data loss risk**: daemon dies mid-write to `transcript.jsonl`
- **Event loop blocked**: no log flushing, no health checks, no API responses during backfill

---

## Workaround

Create the backfill marker file manually to skip the backfill:
```bash
date -Iseconds > ~/.agents/memory/.transcript-backfill-done
```

Or reduce transcript count:
```bash
# Move old transcript artifacts to a backup
mkdir -p /tmp/transcript-backup
mv ~/.agents/memory/*--transcript.md /tmp/transcript-backup/
```

---

## Environment Details

- **Signet:** 0.109.3
- **OS:** Ubuntu, Linux 6.8.0-110-generic x86_64
- **Node:** v22.22.1
- **Bun:** 1.3.12
- **DB:** SQLite 1.3GB (post-cleanup; was 1.7GB), 43,706 memories, 149K transcript records
- **Date observed:** 2026-04-28 (three consecutive OOM kills)

Ostico:
## Summary

Fixes #587 — daemon OOM-kills deterministically ~2 hours after startup on workspaces with >1,000 transcript artifacts.

## Changes

- `platform/daemon/src/session-transcripts.ts` — persistent backfill marker, skip-if-populated guard, append-only backfill, paginated DB reads
- `platform/daemon/src/transcript-jsonl.ts` — new `appendCanonicalTranscriptSnapshot()` export
- `platform/daemon/src/hooks.ts` — switch live writes from full-file rewrite to append-only
- `platform/daemon/src/transcript-jsonl.test.ts` — 3 regression tests for marker skip, populated-JSONL skip, and append-only snapshot

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

## Testing

- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

### Root Cause

`ensureCanonicalTranscriptHistory()` used an in-memory `Set` guard that reset on every daemon restart. Each of the 6,650 backfill iterations called `writeCanonicalTranscriptSnapshot()` which did a full `readFileSync` + parse + filter + rewrite of the entire growing `transcript.jsonl` (212MB, 149K lines). Peak per-iteration memory ~1.5GB, GC couldn't keep up → 12GB RSS → OOM kill.

### Fix

- **Persistent backfill marker** (`.canonical-transcript-backfill-v1`) so `ensureCanonicalTranscriptHistory()` skips on subsequent restarts
- **Skip backfill if JSONL >1MB** already exists (crash-before-marker guard)
- **Append-only writes** for both backfill and live session transcript writes via new `appendCanonicalTranscriptSnapshot()`
- **Paginated DB reads** (100 rows/page) instead of `.all()` loading 46.7MB at once
- **Event loop yields** every 100 files for GC pressure relief

### Verification

- `transcript-jsonl.test.ts`: 7/7 pass (4 existing + 3 new regression tests)
- `hooks.test.ts`: 156/159 pass (3 pre-existing `memory-lineage` failures from missing ONNX model)
- Runtime verified: daemon RSS stable at 250MB after session-end hook (was 12.6GB → OOM before fix)

## AI disclosure

## Notes

Workaround for affected users: `date -Iseconds > ~/.agents/memory/.canonical-transcript-backfill-v1`

### DIFF
--- a/platform/daemon/src/hooks.ts
+++ b/platform/daemon/src/hooks.ts
@@ -101,9 +101,9 @@ import { searchTemporalFallback } from "./temporal-fallback";
 import { writeTranscriptAudit } from "./transcript-audit";
 import {
 	appendCanonicalTranscriptTurns,
-	appendCanonicalTranscriptSnapshot,
 	canonicalTranscriptRelativePath,
 	inferTranscriptSourceFormat,
+	writeCanonicalTranscriptSnapshot,
 } from "./transcript-jsonl";
 import { getUpdateSummary } from "./update-system";
 
@@ -141,7 +141,7 @@ async function writeCanonicalTranscriptFromSnapshot(params: {
 	readonly transcriptPath?: string;
 }): Promise<void> {
 	await ensureCanonicalTranscriptHistory(getAgentsDir(), params.agentId);
-	await appendCanonicalTranscriptSnapshot({
+	await writeCanonicalTranscriptSnapshot({
 		basePath: getAgentsDir(),
 		agentId: params.agentId,
 		harness: params.harness,
--- a/platform/daemon/src/hooks.ts
+++ b/platform/daemon/src/hooks.ts
@@ -101,9 +101,9 @@ import { searchTemporalFallback } from "./temporal-fallback";
 import { writeTranscriptAudit } from "./transcript-audit";
 import {
 	appendCanonicalTranscriptTurns,
+	appendCanonicalTranscriptSnapshot,
 	canonicalTranscriptRelativePath,
 	inferTranscriptSourceFormat,
-	writeCanonicalTranscriptSnapshot,
 } from "./transcript-jsonl";
 import { getUpdateSummary } from "./update-system";
 
@@ -141,7 +141,7 @@ async function writeCanonicalTranscriptFromSnapshot(params: {
 	readonly transcriptPath?: string;
 }): Promise<void> {
 	await ensureCanonicalTranscriptHistory(getAgentsDir(), params.agentId);
-	await writeCanonicalTranscriptSnapshot({
+	await appendCanonicalTranscriptSnapshot({
 		basePath: getAgentsDir(),
 		agentId: params.agentId,
 		harness: params.harness,
--- a/platform/daemon/src/session-transcripts.ts
+++ b/platform/daemon/src/session-transcripts.ts
@@ -1,10 +1,10 @@
-import { existsSync, readFileSync, readdirSync } from "node:fs";
+import { existsSync, readFileSync, readdirSync, statSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { extractAnchorTerms } from "./anchor-terms";
 import { getDbAccessor } from "./db-accessor";
 import { logger } from "./logger";
 import { sanitizeFtsQuery } from "./memory-search";
-import { writeCanonicalTranscriptSnapshot } from "./transcript-jsonl";
+import { appendCanonicalTranscriptSnapshot } from "./transcript-jsonl";
 
 interface TranscriptRow {
 	readonly session_key: string;
@@ -77,15 +77,16 @@ async function backfillMarkdownTranscriptArtifacts(basePath: string, agentId?: s
 	const memoryDir = join(basePath, "memory");
 	if (!existsSync(memoryDir)) return 0;
 	let failures = 0;
-	for (const name of readdirSync(memoryDir)) {
-		if (!name.endsWith("--transcript.md")) continue;
+	const files = readdirSync(memoryDir).filter((name) => name.endsWith("--transcript.md"));
+	for (let i = 0; i < files.length; i++) {
+		const name = files[i]!;
 		const path = join(memoryDir, name);
 		try {
 			const parsed = parseArtifactFrontmatter(readFileSync(path, "utf8"));
 			if (!parsed || parsed.frontmatter.kind !== "transcript") continue;
 			const rowAgentId = parsed.frontmatter.agent_id || "default";
 			if (agentId && rowAgentId !== agentId) continue;
-			await writeCanonicalTranscriptSnapshot({
+			await appendCanonicalTranscriptSnapshot({
 				basePath,
 				agentId: rowAgentId,
 				harness: parsed.frontmatter.harness || "unknown",
@@ -104,34 +105,44 @@ async function backfillMarkdownTranscriptArtifacts(basePath: string, agentId?: s
 				path,
 			});
 		}
+		if (i > 0 && i % 100 === 0) await new Promise((resolve) => setTimeout(resolve, 0));
 	}
 	return failures;
 }
 
 async function backfillDatabaseTranscripts(basePath: string, agentId?: string): Promise<boolean> {
 	if (!tableExists("session_transcripts")) return true;
+	const PAGE_SIZE = 100;
 	try {
-		const rows = getDbAccessor().withReadDb((db) => {
-			const cols = db.prepare("PRAGMA table_info(session_transcripts)").all() as ReadonlyArray<Record<string, unknown>>;
-			const hasUpdated = cols.some((col) => col.name === "updated_at");
-			const sql = hasUpdated
-				? "SELECT session_key, content, harness, project, agent_id, created_at, updated_at FROM session_transcripts"
-				: "SELECT session_key, content, harness, project, agent_id, created_at, NULL AS updated_at FROM session_transcripts";
-			return db.prepare(sql).all() as StoredTranscriptBackfillRow[];
-		});
-		for (const row of rows) {
-			const rowAgentId = row.agent_id?.trim() || "default";
-			if (agentId && rowAgentId !== agentId) continue;
-			await writeCanonicalTranscriptSnapshot({
-				basePath,
-				agentId: rowAgentId,
-				harness: row.harness?.trim() || "unknown",
-				sessionKey: row.session_key,
-				project: row.project,
-				capturedAt: row.updated_at || row.created_at || new Date().toISOString(),
-				sourceFormat: "db",
-				transcript: row.content,
+		let offset = 0;
+		while (true) {
+			const rows = getDbAccessor().withReadDb((db) => {
+				const cols = db
+					.prepare("PRAGMA table_info(session_transcripts)")
+					.all() as ReadonlyArray<Record<string, unknown>>;
+				const hasUpdated = cols.some((col) => col.name === "updated_at");
+				const sql = hasUpdated
+					? "SELECT session_key, content, harness, project, agent_id, created_at, updated_at FROM session_transcripts LIMIT ? OFFSET ?"
+					: "SELECT session_key, content, harness, project, agent_id, created_at, NULL AS updated_at FROM session_transcripts LIMIT ? OFFSET ?";
+				return db.prepare(sql).all(PAGE_SIZE, offset) as StoredTranscriptBackfillRow[];
 			});
+			if (rows.length === 0) break;
+			for (const row of rows) {
+				const rowAgentId = row.agent_id?.trim() || "default";
+				if (agentId && rowAgentId !== agentId) continue;
+				await appendCanonicalTranscriptSnapshot({
+					basePath,
+					agentId: rowAgentId,
+					harness: row.harness?.trim() || "unknown",
+					sessionKey: row.session_key,
+					project: row.project,
+					capturedAt: row.updated_at || row.created_at || new Date().toISOString(),
+					sourceFormat: "db",
+					transcript: row.content,
+				});
+			}
+			offset += PAGE_SIZE;
+			await new Promise((resolve) => setTimeout(resolve, 0));
 		}
 	} catch (error) {
 		logger.warn("transcripts", "Database transcript backfill failed", {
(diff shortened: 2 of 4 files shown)


NicholaiVogel:
I dug through this more deeply. I think this needs another pass before merge; the OOM direction is right, but the current fix trades the crash-loop for data-loss / duplication edge cases.

Main blockers I found:

1. The new `.canonical-transcript-backfill-v1` marker is global to the workspace, but `ensureCanonicalTranscriptHistory(basePath, agentId)` is scoped by agent. A successful backfill for one agent can permanently skip historical backfill for another agent.

2. Backfill retries are no longer idempotent. If markdown or DB backfill partially succeeds and then fails, no marker is written. The next write restarts from the top, but `appendCanonicalTranscriptSnapshot()` only guards against duplicate trailing turns, so already-imported sessions can be appended again.

3. Session-end writes were changed from replace-per-session to append-only. Prompt-submit can already append live turns for the same session, so the final session-end snapshot can duplicate the same conversation instead of replacing the live partial snapshot.

4. The `>1MB` crash-before-marker heuristic is too coarse. Any large `memory/*/transcripts/transcript.jsonl` file suppresses all backfill, without verifying canonical records, requested agent, harness coverage, or DB/markdown migration completeness.

5. DB paging uses `LIMIT/OFFSET` without `ORDER BY`, so rows can be skipped or duplicated across pages.

6. Marker handling is existence-only and fail-open: stale/wrong/empty markers suppress backfill, while marker write failures still mark the current process as complete and cause append-only duplication after restart.

Suggested direction: keep append-only for live turn ingestion, but preserve replace/idempotent semantics for full snapshots/backfill; make the backfill marker scoped and validated; use stable ordered DB paging; and avoid writing a completion marker unless marker persistence succeeds.

PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): prevent OOM crash-loop in transcript backfill` (commit `4b017fdd`) now and I'll follow up shortly with feedback.

platform/daemon/src/session-transcripts.ts
```diff
@@ -142,9 +153,43 @@ async function backfillDatabaseTranscripts(basePath: string, agentId?: string):
...
 
+const BACKFILL_MARKER = ".canonical-transcript-backfill-v1";
+
 export async function ensureCanonicalTranscriptHistory(basePath: string, agentId?: string): Promise<void> {
 	const key = `${basePath}:${agentId ?? "*"}`;
 	if (canonicalBackfills.has(key)) return;
+
+	// Persistent marker survives daemon restarts — skip backfill if
+	// a previous lifecycle already completed it.
+	const markerPath = join(basePath, "memory", BACKFILL_MARKER);
```
PR-Reviewer-Ant: Backfill marker is global (`memory/.canonical-transcript-backfill-v1`) but `ensureCanonicalTranscriptHistory` is called with optional `agentId`. If one agent writes the marker first, later calls for other agents will skip historical backfill entirely, causing missing canonical history for those agents.

platform/daemon/src/session-transcripts.ts
```diff
@@ -142,9 +153,43 @@ async function backfillDatabaseTranscripts(basePath: string, agentId?: string):
...
+	}
+
+	// If any JSONL file already has substantial data, the backfill
+	// happened in a prior lifecycle that crashed before writing the
+	// marker.  Re-running the backfill would duplicate records and
+	// balloon the file, so write the marker and skip.
+	const memDir = join(basePath, "memory");
+	const alreadyPopulated = existsSync(memDir) && readdirSync(memDir).some((entry) => {
+		const jsonlPath = join(memDir, entry, "transcripts", "transcript.jsonl");
+		try {
```
PR-Reviewer-Ant: The `>1MB` populated-file heuristic can suppress required backfill for unrelated agent/harness data. Any large `memory/*/transcripts/transcript.jsonl` triggers a global skip without validating agent scope, record integrity, or migration completeness, which can permanently miss backfill coverage.

platform/daemon/src/session-transcripts.ts
```diff
@@ -104,34 +105,44 @@ async function backfillMarkdownTranscriptArtifacts(basePath: string, agentId?: s
...
-				sessionKey: row.session_key,
-				project: row.project,
-				capturedAt: row.updated_at || row.created_at || new Date().toISOString(),
-				sourceFormat: "db",
-				transcript: row.content,
+		let offset = 0;
+		while (true) {
+			const rows = getDbAccessor().withReadDb((db) => {
+				const cols = db
+					.prepare("PRAGMA table_info(session_transcripts)")
```
PR-Reviewer-Ant: DB pagination uses `LIMIT ? OFFSET ?` without `ORDER BY`, so page boundaries are unstable and rows may be skipped or duplicated across iterations under concurrent writes or query planner changes. Backfill correctness requires deterministic ordering (e.g., by rowid).

platform/daemon/src/hooks.ts
```diff
@@ -141,7 +141,7 @@ async function writeCanonicalTranscriptFromSnapshot(params: {
 	readonly transcriptPath?: string;
 }): Promise<void> {
 	await ensureCanonicalTranscriptHistory(getAgentsDir(), params.agentId);
-	await writeCanonicalTranscriptSnapshot({
+	await appendCanonicalTranscriptSnapshot({
```
PR-Reviewer-Ant: Session-end canonical write changed from replace-per-session to append-only (`appendCanonicalTranscriptSnapshot`). If live turns were already appended during the same session, final snapshot appends the same conversation again. Existing trailing-turn dedupe is insufficient when overlap is partial/non-trailing, so duplicates can accumulate.

platform/daemon/src/session-transcripts.ts
```diff
@@ -156,9 +201,25 @@ export async function ensureCanonicalTranscriptHistory(basePath: string, agentId
 		});
 		return;
 	}
+
+	writeMarker(markerPath, agentId);
 	canonicalBackfills.add(key);
 }
```
PR-Reviewer-Ant, line 241: Marker persistence failure is treated as non-fatal, but `canonicalBackfills.add(key)` is still set by caller after `writeMarker(...)`. This can mark backfill complete for the process even when marker write failed, weakening retry guarantees and making behavior diverge across restarts.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `4b017fdd`

[Automated Review] I found multiple correctness risks in the backfill hardening that can cause skipped migrations or duplicated transcript data, so this needs another revision before human merge review.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are directly observable in the changed code paths: marker scoping/validation in session-transcripts.ts, pagination query shape in session-transcripts.ts, and append semantics switch in hooks.ts + transcript-jsonl.ts. The regression tests added do not cover these failure modes.

platform/daemon/src/transcript-jsonl.ts
```diff
@@ -379,6 +452,31 @@ export function appendCanonicalTranscriptTurns(
...
 
+export function appendCanonicalTranscriptSnapshotIfMissing(
+	input: TranscriptIdentity & { readonly transcript: string },
+	knownSessionKeys?: Set<string>,
+): Promise<string | null> {
+	const turns = transcriptTextToTurns(input.transcript);
+	if (turns.length === 0) return Promise.resolve(null);
+	const path = canonicalTranscriptPath(input.basePath, input.harness);
+	const key = sessionSeqCacheKey(input);
+	if (knownSessionKeys?.has(key)) return Promise.resolve(path);
```
PR-Reviewer-Ant, line 463: Blocking: `appendCanonicalTranscriptSnapshotIfMissing()` now treats "any existing record for this session" as fully backfilled (`knownSessionKeys?.has(key)`), so backfill can skip importing a fuller DB/markdown snapshot when JSONL already has partial live turns for that session. That leaves incomplete transcript history permanently once the marker is written. Dedupe should distinguish partial/live append state from canonical snapshot completion (for example by checking source format or replacing per-session during backfill).

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `eeab4c62`

[Automated Review] I found one blocking correctness issue where backfill can silently skip fuller transcript data and leave truncated session history.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new dedupe path now uses preloaded session keys and short-circuits writes by session identity only. In the changed code paths, any existing record for a session causes DB/markdown snapshot import to be skipped, even when existing records are only partial live turns. This is directly visible in the added `knownSessionKeys` fast-path and how backfill now always supplies that set.

NicholaiVogel:
@Ostico nice work :)

Ostico:
Thanks @NicholaiVogel , i'm pushing the daemon hard beyond its limits XD. I will try to address my other issues in my spare time.

PR #594, opened 2026-04-29
fix(daemon): streaming rewrite for live-only transcript sessions during backfill

### LINKED ISSUE #595: Daemon crash: socket.destroySoon is not a function (@hono/node-server drain handler + Bun HTTP socket)
## Summary

The daemon crashes with `TypeError: socket.destroySoon is not a function` when `@hono/node-server`'s drain handler fires on a Bun HTTP socket. This affects **all daemon instances** running Bun ≥1.3.x with `@hono/node-server` ≥1.19.13.

## Root Cause

`@hono/node-server` v1.19.13 introduced a `drainIncoming()` function that calls `socket.destroySoon()` when a request body isn't fully consumed within 500ms. However, Bun's internal HTTP server socket wrapper (`NodeHTTPServerSocket`) extends `Duplex` instead of `net.Socket` and lacks the `destroySoon()` method.

**Call chain:**
```
HTTP POST with unconsumed body
→ response finishes (outgoing.on("finish"))
→ drainIncoming(incoming) fires
→ 500ms drain timeout expires
→ forceClose() → incoming.socket.destroySoon()
→ TypeError (Bun socket lacks method)
→ uncaught exception → daemon crash
```

## Trigger Conditions

Any of these will crash the daemon:
- MCP client reconnection (POST hooks with unconsumed body)
- Client disconnect during a long-running request
- Network timeout during hook submission
- POST to an endpoint that doesn't read the body (e.g. health check)
- Any malformed POST body

## Version Matrix

| Component | Version | Status |
|---|---|---|
| Bun | 1.3.12 | `NodeHTTPServerSocket` lacks `destroySoon` |
| `@hono/node-server` | 1.19.13+ | Calls `socket.destroySoon()` in drain handler |
| `@hono/node-server` | 1.19.9 | No drain logic (not affected) |

## Upstream

- Bun issue: https://github.com/oven-sh/bun/issues/24127 (open)
- Bun fix PR: https://github.com/oven-sh/bun/pull/26264 (open, not merged)
- Bun docs claim `node:net` is "fully implemented" but HTTP socket wrappers are incomplete

## Reproduction

```bash
cd platform/daemon && bun src/daemon.ts &
# Send POST with body to an endpoint that won't consume it
curl -X POST http://localhost:3850/health -d '{"x":1}'
sleep 1
curl http://localhost:3850/health  # daemon is dead
```

## Proposed Fix

Polyfill `destroySoon` on `Duplex.prototype` at daemon startup with the same semantics as Node.js:

```typescript
import { Duplex } from "node:stream";
if (typeof Duplex.prototype.destroySoon !== "function") {
    Duplex.prototype.destroySoon = function destroySoon() {
        if (this.writable) this.end();
        if (this.writableFinished) this.destroy();
        else this.once("finish", this.destroy);
    };
}
```

Remove when Bun ships PR #26264.

## Observed Error

```
[11:45:25] ERROR [daemon]     Uncaught exception
  TypeError: socket.destroySoon is not a function.
  (In 'socket.destroySoon()', 'socket.destroySoon' is undefined)
```

Ostico:
## Summary

- Streaming rewrite for live-only transcript sessions during backfill (fixes partial live turns persisting when full transcripts exist as artifacts)
- **Streaming `writeCanonicalTranscriptSnapshot` to prevent OOM on large JSONL files** (fixes daemon OOM kill on 924 MB opencode JSONL during session-end)
- **Fix seq cache clobber on healed sessions** — separate `rewritten`/`healed` sets so only actually-rewritten sessions update sessionSeqCache
- **Polyfill `destroySoon` for Bun HTTP socket compatibility** — prevents daemon crash when @hono/node-server drain handler fires on unconsumed POST bodies (fixes #595)
- **Race guard** — re-classify inside file lock to prevent rewriting sessions that were healed between classification and lock acquisition

## Changes

| File | Change |
|---|---|
| `platform/daemon/src/transcript-jsonl.ts` | Streaming `writeCanonicalTranscriptSnapshot`, streaming `rewriteReplacingLiveOnlySessions` with pre-scan race guard, fail-closed on empty replacement |
| `platform/daemon/src/transcript-jsonl.test.ts` | 26 tests: classification, rewrite, OOM guard, crash-safety, seq-cache regression, race guard |
| `platform/daemon/src/session-transcripts.ts` | Backfill orchestration: classify → rewrite → append, `createBackfillSeenReader()` |
| `platform/daemon/src/bun-socket-polyfill.ts` | `Duplex.prototype.destroySoon` polyfill (Node.js-exact semantics, remove when Bun ships PR #26264) |
| `platform/daemon/src/bun-socket-polyfill.test.ts` | 5 tests: baseline, apply, writable→end→destroy, finished→destroy, no-overwrite |
| `platform/daemon/src/daemon.ts` | Import polyfill as first module |

## OOM Fix Performance

| Metric | Before | After |
|---|---|---|
| Peak RSS | 7.2 GB (OOM killed) | 152 MB |
| File size | 924 MB | 924 MB (preserved) |
| Elapsed | N/A (killed) | 5 seconds |
| Memory complexity | O(file_size) | O(line_size) |

## destroySoon Fix

Root cause: `@hono/node-server` v1.19.13 calls `socket.destroySoon()` in its drain handler. Bun's HTTP server socket wrapper (`NodeHTTPServerSocket`) extends `Duplex` but lacks this method. Any unconsumed POST body (MCP reconnection, client disconnect) crashes the daemon.

Upstream: https://github.com/oven-sh/bun/issues/24127 (open), PR https://github.com/oven-sh/bun/pull/26264 (not merged).

Polyfill uses Node.js-exact semantics. Verified locally: daemon survives POST with unconsumed body that previously crashed it.

## Readiness Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@
 .bun-tmp/
 .claude/
 .codex/
+.sisyphus/
 .cursorindexingignore
 .env
 .env.*.local
@@ -36,6 +37,7 @@
 .vscode/
 .worktrees/
 /marketing/
+/packages/
 /references/
 Thumbs.db
 coverage/
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -4,6 +4,7 @@
  * Background service for memory, API, and dashboard hosting
  */
 
+import "./bun-socket-polyfill";
 import type { ChildProcess } from "node:child_process";
 import { spawn } from "node:child_process";
 import { createHash } from "node:crypto";
--- a/platform/daemon/src/session-transcripts.ts
+++ b/platform/daemon/src/session-transcripts.ts
@@ -46,7 +46,6 @@ export interface TranscriptHit {
 
 interface BackfillSeen {
 	readonly classification: TranscriptSessionKeyClassification;
-	readonly allKeys: Set<string>;
 }
 
 function createBackfillSeenReader(basePath: string, agentId?: string): (harness: string) => Promise<BackfillSeen> {
@@ -57,11 +56,19 @@ function createBackfillSeenReader(basePath: string, agentId?: string): (harness:
 		if (existing) return existing;
 		const loaded = readCanonicalTranscriptSessionKeys({ basePath, harness, agentId }).then((classification) => ({
 			classification,
-			allKeys: new Set([...classification.canonicalKeys, ...classification.liveOnlyKeys]),
 		}));
 		seen.set(key, loaded);
 		return loaded;
 	};
+}
+
+function knownBackfillKeys(classification: TranscriptSessionKeyClassification): Set<string> {
+	return new Set([...classification.canonicalKeys, ...classification.liveOnlyKeys]);
+}
+
+function markBackfillCanonical(classification: TranscriptSessionKeyClassification, key: string): void {
+	classification.liveOnlyKeys.delete(key);
+	classification.canonicalKeys.add(key);
 }
 
 function tableExists(name: string): boolean {
@@ -135,7 +142,7 @@ async function backfillMarkdownTranscriptArtifacts(
 				sourcePath: `memory/${name}`,
 				transcript: parsed.body,
 			};
-			const { classification, allKeys } = await getSeen(harness);
+			const { classification } = await getSeen(harness);
 			const key = sessionSeqCacheKey(input);
 			if (classification.liveOnlyKeys.has(key)) {
 				const replacements =
@@ -154,7 +161,9 @@ async function backfillMarkdownTranscriptArtifacts(
 				liveOnlyReplacements.set(harness, replacements);
 				continue;
 			}
-			await appendCanonicalTranscriptSnapshotIfMissing(input, allKeys);
+			if (await appendCanonicalTranscriptSnapshotIfMissing(input, knownBackfillKeys(classification))) {
+				markBackfillCanonical(classification, key);
+			}
 		} catch (error) {
 			failures++;
 			logger.warn("transcripts", "Markdown transcript backfill failed", {
@@ -171,8 +180,7 @@ async function backfillMarkdownTranscriptArtifacts(
 			await rewriteReplacingLiveOnlySessions(jsonlPath, replacements);
 			const { classification } = await getSeen(harness);
 			for (const key of replacements.keys()) {
-				classification.liveOnlyKeys.delete(key);
-				classification.canonicalKeys.add(key);
+				markBackfillCanonical(classification, key);
 			}
 		} catch (error) {
 			failures++;
@@ -231,7 +239,7 @@ async function backfillDatabaseTranscripts(
 					sourceFormat: "db" as const,
 					transcript: row.content,
 				};
-				const { classification, allKeys } = await getSeen(harness);
+				const { classification } = await getSeen(harness);
 				const key = sessionSeqCacheKey(input);
 				if (classification.liveOnlyKeys.has(key)) {
 					const replacements =
@@ -249,7 +257,9 @@ async function backfillDatabaseTranscripts(
 					liveOnlyReplacements.set(harness, replacements);
 					continue;
 				}
-				await appendCanonicalTranscriptSnapshotIfMissing(input, allKeys);
+				if (await appendCanonicalTranscriptSnapshotIfMissing(input, knownBackfillKeys(classification))) {
+					markBackfillCanonical(classification, key);
+				}
 			}
 			offset += PAGE_SIZE;
 			await new Promise((resolve) => setTimeout(resolve, 0));
@@ -260,8 +270,7 @@ async function backfillDatabaseTranscripts(
 			await rewriteReplacingLiveOnlySessions(jsonlPath, replacements);
 			const { classification } = await getSeen(harness);
 			for (const key of replacements.keys()) {
-				classification.liveOnlyKeys.delete(key);
-				classification.canonicalKeys.add(key);
+				markBackfillCanonical(classification, key);
 			}
 		}
 	} catch (error) {
--- a/platform/daemon/src/session-transcripts.ts
+++ b/platform/daemon/src/session-transcripts.ts
@@ -5,9 +5,14 @@ import { getDbAccessor } from "./db-accessor";
 import { logger } from "./logger";
 import { sanitizeFtsQuery } from "./memory-search";
 import {
+	type TranscriptIdentity,
+	type TranscriptSessionKeyClassification,
 	appendCanonicalTranscriptSnapshotIfMissing,
+	canonicalTranscriptPath,
 	readCanonicalTranscriptSessionKeys,
+	rewriteReplacingLiveOnlySessions,
 	sanitizeHarnessPath,
+	sessionSeqCacheKey,
 } from "./transcript-jsonl";
 
 interface TranscriptRow {
@@ -39,13 +44,21 @@ export interface TranscriptHit {
 	readonly rank: number;
 }
 
-function createBackfillSeenReader(basePath: string, agentId?: string): (harness: string) => Promise<Set<string>> {
-	const seen = new Map<string, Promise<Set<string>>>();
+interface BackfillSeen {
+	readonly classification: TranscriptSessionKeyClassification;
+	readonly allKeys: Set<string>;
+}
+
+function createBackfillSeenReader(basePath: string, agentId?: string): (harness: string) => Promise<BackfillSeen> {
+	const seen = new Map<string, Promise<BackfillSeen>>();
 	return (harness: string) => {
 		const key = sanitizeHarnessPath(harness);
 		const existing = seen.get(key);
 		if (existing) return existing;
-		const loaded = readCanonicalTranscriptSessionKeys({ basePath, harness, agentId });
+		const loaded = readCanonicalTranscriptSessionKeys({ basePath, harness, agentId }).then((classification) => ({
+			classification,
+			allKeys: new Set([...classification.canonicalKeys, ...classification.liveOnlyKeys]),
+		}));
 		seen.set(key, loaded);
 		return loaded;
 	};
@@ -92,11 +105,15 @@ function parseArtifactFrontmatter(content: string): { frontmatter: Record<string
 async function backfillMarkdownTranscriptArtifacts(
 	basePath: string,
 	agentId: string | undefined,
-	getSeen: (harness: string) => Promise<Set<string>>,
+	getSeen: (harness: string) => Promise<BackfillSeen>,
 ): Promise<number> {
 	const memoryDir = join(basePath, "memory");
 	if (!existsSync(memoryDir)) return 0;
 	let failures = 0;
+	const liveOnlyReplacements = new Map<
+		string,
+		Map<string, { readonly identity: TranscriptIdentity; readonly transcript: string }>
+	>();
 	const files = readdirSync(memoryDir).filter((name) => name.endsWith("--transcript.md"));
 	for (const [i, name] of files.entries()) {
 		const path = join(memoryDir, name);
@@ -106,21 +123,38 @@ async function backfillMarkdownTranscriptArtifacts(
 			const rowAgentId = parsed.frontmatter.agent_id || "default";
 			if (agentId && rowAgentId !== agentId) continue;
 			const harness = parsed.frontmatter.harness || "unknown";
-			await appendCanonicalTranscriptSnapshotIfMissing(
-				{
-					basePath,
-					agentId: rowAgentId,
-					harness,
-					sessionKey: parsed.frontmatter.session_key || null,
-					sessionId: parsed.frontmatter.session_id || null,
-					project: parsed.frontmatter.project || null,
-					capturedAt: parsed.frontmatter.captured_at || new Date().toISOString(),
-					sourceFormat: "markdown",
-					sourcePath: `memory/${name}`,
-					transcript: parsed.body,
-				},
-				await getSeen(harness),
-			);
+			const input = {
+				basePath,
+				agentId: rowAgentId,
+				harness,
+				sessionKey: parsed.frontmatter.session_key || null,
+				sessionId: parsed.frontmatter.session_id || null,
+				project: parsed.frontmatter.project || null,
+				capturedAt: parsed.frontmatter.captured_at || new Date().toISOString(),
+				sourceFormat: "markdown" as const,
+				sourcePath: `memory/${name}`,
+				transcript: parsed.body,
+			};
+			const { classification, allKeys } = await getSeen(harness);
+			const key = sessionSeqCacheKey(input);
+			if (classification.liveOnlyKeys.has(key)) {
+				const replacements =
+					liveOnlyReplacements.get(harness) ||
+					new Map<string, { readonly identity: TranscriptIdentity; readonly transcript: string }>();
+				replacements.set(key, {
+					identity: {
+						agentId: input.agentId,
+						harness: input.harness,
+						sessionKey: input.sessionKey,
+						sessionId: input.sessionId,
+						sourceFormat: "markdown",
+					},
+					transcript: input.transcript,
+				});
+				liveOnlyReplacements.set(harness, replacements);
+				continue;
+			}
+			await appendCanonicalTranscriptSnapshotIfMissing(input, allKeys);
 		} catch (error) {
 			failures++;
 			logger.warn("transcripts", "Markdown transcript backfill failed", {
@@ -130,16 +164,39 @@ async function backfillMarkdownTranscriptArtifacts(
 		}
 		if (i > 0 && i % 100 === 0) await new Promise((resolve) => setTimeout(resolve, 0));
 	}
+	for (const [harness, replacements] of liveOnlyReplacements.entries()) {
+		if (replacements.size === 0) continue;
+		const jsonlPath = canonicalTranscriptPath(basePath, harness);
+		try {
+			await rewriteReplacingLiveOnlySessions(jsonlPath, replacements);
+			const { classification } = await getSeen(harness);
+			for (const key of replacements.keys()) {
+				classification.liveOnlyKeys.delete(key);
+				classification.canonicalKeys.add(key);
+			}
+		} catch (error) {
+			failures++;
+			logger.warn("transcripts", "Markdown transcript backfill rewrite failed", {
+				error: error instanceof Error ? error.message : String(error),
+				harness,
+				path: jsonlPath,
+			});
+		}
+	}
 	return failures;
 }
 
 async function backfillDatabaseTranscripts(
 	basePath: string,
 	agentId: string | undefined,
-	getSeen: (harness: string) => Promise<Set<string>>,
+	getSeen: (harness: string) => Promise<BackfillSeen>,
 ): Promise<boolean> {
 	if (!tableExists("session_transcripts")) return true;
 	const PAGE_SIZE = 100;
+	const liveOnlyReplacements = new Map<
+		string,
+		Map<string, { readonly identity: TranscriptIdentity; readonly transcript: string }>
+	>();
 	try {
 		let offset = 0;
 		while (true) {
@@ -164,22 +221,48 @@ async function backfillDatabaseTranscripts(
 				const rowAgentId = row.agent_id?.trim() || "default";
 				if (agentId && rowAgentId !== agentId) continue;
 				const harness = row.harness?.trim() || "unknown";
-				await appendCanonicalTranscriptSnapshotIfMissing(
-					{
-						basePath,
-						agentId: rowAgentId,
-						harness,
-						sessionKey: row.session_key,
-						project: row.project,
-						capturedAt: row.updated_at || row.created_at || new Date().toISOString(),
-						sourceFormat: "db",
-						transcript: row.content,
-					},
-					await getSeen(harness),
-				);
+				const input = {
+					basePath,
+					agentId: rowAgentId,
+					harness,
+					sessionKey: row.session_key,
+					project: row.project,
+					capturedAt: row.updated_at || row.created_at || new Date().toISOString(),
+					sourceFormat: "db" as const,
+					transcript: row.content,
+				};
+				const { classification, allKeys } = await getSeen(harness);
+				const key = sessionSeqCacheKey(input);
+				if (classification.liveOnlyKeys.has(key)) {
+					const replacements =
+						liveOnlyReplacements.get(harness) ||
+						new Map<string, { readonly identity: TranscriptIdentity; readonly transcript: string }>();
+					replacements.set(key, {
+						identity: {
+							agentId: input.agentId,
+							harness: input.harness,
+							sessionKey: input.sessionKey,
+							sourceFormat: "db",
+						},
+						transcript: input.transcript,
+					});
+					liveOnlyReplacements.set(harness, replacements);
+					continue;
+				}
+				await appendCanonicalTranscriptSnapshotIfMissing(input, allKeys);
 			}
 			offset += PAGE_SIZE;
 			await new Promise((resolve) => setTimeout(resolve, 0));
+		}
+		for (const [harness, replacements] of liveOnlyReplacements.entries()) {
+			if (replacements.size === 0) continue;
+			const jsonlPath = canonicalTranscriptPath(basePath, harness);
+			await rewriteReplacingLiveOnlySessions(jsonlPath, replacements);
+			const { classification } = await getSeen(harness);
+			for (const key of replacements.keys()) {
+				classification.liveOnlyKeys.delete(key);
+				classification.canonicalKeys.add(key);
+			}
 		}
 	} catch (error) {
 		logger.warn("transcripts", "Database transcript backfill failed", {
--- a/platform/daemon/src/transcript-jsonl.ts
+++ b/platform/daemon/src/transcript-jsonl.ts
@@ -553,9 +553,38 @@ export function rewriteReplacingLiveOnlySessions(
 ): Promise<number> {
 	if (replacements.size === 0 || !existsSync(jsonlPath)) return Promise.resolve(0);
 	return withTranscriptFileLock(jsonlPath, async () => {
+		// Re-classify inside the lock: between external classification and lock
+		// acquisition, session-end hooks may have written non-live records. Only
+		// replace sessions that are STILL live-only at the moment we hold the lock.
+		const healedKeys = new Set<string>();
+		const prescan = createInterface({
+			input: createReadStream(jsonlPath, { encoding: "utf8" }),
+			crlfDelay: Number.POSITIVE_INFINITY,
+		});
+		try {
+			for await (const line of prescan) {
+				const trimmed = line.trim();
+				if (trimmed.length === 0) continue;
+				try {
+					const parsed = JSON.parse(trimmed) as Partial<CanonicalTranscriptRecord>;
+					if (parsed.schema !== "signet.transcript.v1" || typeof parsed.content !== "string") continue;
+					const record = parsed as CanonicalTranscriptRecord;
+					const key = recordSeqCacheKey(record);
+					if (replacements.has(key) && record.source_format !== "live") {
+						healedKeys.add(key);
+					}
+				} catch {}
+			}
+		} finally {
+			prescan.close();
+		}
+		const effectiveReplacements = healedKeys.size > 0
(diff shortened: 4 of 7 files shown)


PR-Reviewer-Ant:
Hi @Ostico - I'm an automated code reviewer powered by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer). I'm taking a look at the fixes in `fix(daemon): streaming rewrite for live-only transcript sessions during backfill` (commit `0fdb6d92`) now and I'll follow up shortly with feedback.

platform/daemon/src/transcript-jsonl.ts
```diff
@@ -477,6 +491,75 @@ export function appendCanonicalTranscriptSnapshotIfMissing(
...
+							.map((turn, index) => makeRecord(entry.identity, turn, index + 1))
+							.filter((record): record is CanonicalTranscriptRecord => record !== null);
+						for (const record of next) {
+							writeSync(fd, `${JSON.stringify(record)}\n`);
+						}
+						replaced.add(key);
+					} catch {
+						writeSync(fd, `${line}\n`);
+					}
+				}
```
PR-Reviewer-Ant, line 635: Blocking: `rewriteReplacingLiveOnlySessions` drops every subsequent line for a matching session key once `replaced.has(key)` becomes true (`continue`), regardless of `source_format`. If a canonical/non-live record for that key exists in the file (including concurrent append/snapshot activity between classification and rewrite), this rewrite will silently delete those records. This is concrete data-loss risk and contradicts the intended 'replace live-only sessions' behavior.

platform/daemon/src/transcript-jsonl.ts
```diff
@@ -477,6 +491,75 @@ export function appendCanonicalTranscriptSnapshotIfMissing(
...
+			} finally {
+				lines.close();
+			}
```
PR-Reviewer-Ant, line 638: Blocking: when a replacement transcript parses to zero turns (`transcriptTextToTurns(...)` => empty after cleaning), the code writes nothing for that session and still marks it as replaced, effectively deleting the existing session content from JSONL. Replacement should be fail-closed (preserve original lines) if canonicalized replacement payload is empty.

PR-Reviewer-Ant (changes requested):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `0fdb6d92`

[Automated Review by pr-reviewer] I found blocking data-loss risks in the live-only rewrite path, and the implementation currently diverges from the PR’s crash-safe/correctness claims under concurrent writes.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are directly visible in the new rewrite loop and how it decides which lines to keep/drop. The control flow in `rewriteReplacingLiveOnlySessions` can discard non-live records for a replaced key and can also delete a session entirely when replacement parsing yields no turns.

platform/daemon/src/transcript-jsonl.ts
Ostico: Fixed in ca6de489. The rewrite loop now checks `source_format` on each line:

1. **Already-replaced key + non-live record** → written through verbatim (preserves concurrent canonical writes)
2. **First encounter of a replacement key is already non-live** → replacement skipped entirely (session was healed between classification and rewrite by session-end hook)
3. **Already-replaced key + live record** → skipped (remaining partial turns)

Regression test added: `"preserves non-live records for a replaced session key"`

platform/daemon/src/transcript-jsonl.ts
Ostico: Fixed in ca6de489. The rewrite now checks whether the replacement yields any valid records before marking the session as replaced:

```typescript
const next = transcriptTextToTurns(entry.transcript)
    .map((turn, index) => makeRecord(entry.identity, turn, index + 1))
    .filter(...);
if (next.length === 0) {
    // Replacement payload is empty — preserve original line (fail-closed).
    writeSync(fd, `${line}\n`);
    continue;
}
```

If the replacement transcript parses to zero turns, the original live line is preserved verbatim and the session is NOT marked as replaced. Subsequent live lines for the same session are also preserved.

Regression test added: `"preserves original lines when replacement transcript is empty"`

.sisyphus/boulder.json
```diff
@@ -0,0 +1,61 @@
+{
```
PR-Reviewer-Ant: Blocking: this file contains internal operational metadata (worktree paths, session IDs, agent/task tracking). Committing this expands attack surface and leaks local environment details unrelated to the stated fix. Remove `.sisyphus/*` artifacts from the PR.

.sisyphus/handoffs/synthesis-worker-bugfix-handoff.md
```diff
@@ -0,0 +1,223 @@
+# Handoff: Synthesis Worker Thread Bug Fixes
```
PR-Reviewer-Ant: Blocking: this handoff includes internal process constraints, environment details, and implementation history that should not ship in a production code PR. This also materially diverges from the PR description (transcript backfill fix in `platform/daemon/*`). Please scope the PR to the claimed daemon files/tests only.

.sisyphus/issues/suppress-extraction-notifications.md
```diff
@@ -0,0 +1,174 @@
+# Suppress OpenCode desktop notifications for extraction sessions
```
PR-Reviewer-Ant: Warning: this appears unrelated to PR #594’s stated objective (live-only transcript backfill rewrite). Keeping unrelated issue/docs payload in the same PR reduces reviewability and increases merge risk.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `ca6de489`

[Automated Review by pr-reviewer] I found blocking issues: the PR content diverges from its stated scope and appears to include sensitive internal artifacts.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The provided diff clearly shows many new `.sisyphus/*` files containing absolute local paths, session identifiers, operational notes, and internal QA artifacts, which do not match the PR description claiming focused daemon transcript-backfill changes. I cannot fully validate the daemon implementation details because the code diff for key `platform/daemon` files is truncated in the provided context.

.sisyphus/issues/suppress-extraction-notifications.md
Ostico: Removed unwanted files

.sisyphus/handoffs/synthesis-worker-bugfix-handoff.md
Ostico: Removed unwanted files

.sisyphus/boulder.json
Ostico: Removed unwanted files

packages/cli/dashboard/.svelte-kit/ambient.d.ts
```diff
@@ -0,0 +1,362 @@
+
```
PR-Reviewer-Ant: Blocking: generated `.svelte-kit` build artifacts are committed in a daemon bugfix PR. This diverges from the PR’s stated scope and introduces non-source, machine-generated files into version control, increasing merge risk and future drift.

packages/cli/dashboard/.svelte-kit/ambient.d.ts
```diff
@@ -0,0 +1,362 @@
...
+	export const USER: string;
+	export const PLASMA_USE_QT_SCALING: string;
+	export const KDE_SESSION_VERSION: string;
+	export const PAM_KWALLET5_LOGIN: string;
+	export const QT_WAYLAND_FORCE_DPI: string;
+	export const DISPLAY: string;
+	export const SHLVL: string;
+	export const NVM_CD_FLAGS: string;
+	export const XDG_VTNR: string;
+	export const XDG_SESSION_ID: string;
```
PR-Reviewer-Ant: Security warning: this generated file publishes private environment variable identifiers (e.g. `docker_hub_token`) into the repository. Even without values, exposing secret names/internal environment shape is unnecessary information disclosure and expands attack reconnaissance surface.

packages/cli/dashboard/.svelte-kit/generated/client/nodes/1.js
```diff
@@ -0,0 +1 @@
+export { default as component } from "../../../../../../../node_modules/.bun/@sveltejs+kit@2.57.1+0253106801e2759e/node_modules/@sveltejs/kit/src/runtime/components/svelte-5/error.svelte";
```
PR-Reviewer-Ant: Blocking: generated client files include machine-specific dependency paths (e.g. `.bun/...@sveltejs+kit@...`), which are non-portable and brittle across environments. These should not be versioned for this fix.

packages/adapters/openclaw/dist/index.d.ts
```diff
@@ -0,0 +1,262 @@
+/**
```
PR-Reviewer-Ant: Blocking: compiled `dist/` artifacts were added in a PR described as daemon transcript backfill logic changes. This is scope creep and creates maintainability/reviewability risk by mixing unrelated generated output with core bugfix logic.

packages/cli/dashboard/.svelte-kit/output/client/.vite/manifest.json
```diff
@@ -0,0 +1,785 @@
+{
```
PR-Reviewer-Ant: Blocking: a large generated Vite manifest and immutable asset bundle are included, which are unrelated to the transcript backfill fix and likely accidental. This can mask real logic changes and cause unnecessary conflicts/regressions.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `3153a251`

[Automated Review by pr-reviewer] The implementation diverges materially from the PR description and includes risky generated artifacts unrelated to the stated daemon fix.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description claims scoped changes in daemon transcript backfill files, but the diff adds large unrelated generated outputs under packages/cli/dashboard/.svelte-kit and packages/adapters/openclaw/dist. These additions are directly visible in changed paths and contain machine-specific/generated content that should not be part of this fix.

packages/adapters/openclaw/dist/index.d.ts
PR-Reviewer-Ant: Scope divergence: this PR is described as a daemon transcript backfill fix, but it adds an unrelated OpenClaw adapter dist API surface (including request sanitization/auth header swap internals). This should be split into a separate PR to avoid shipping unreviewed behavior under a bugfix label.

packages/cli/dashboard/.svelte-kit/ambient.d.ts
PR-Reviewer-Ant: Generated `.svelte-kit` build artifacts are committed in bulk and are unrelated to the daemon transcript fix. This creates high review noise and risks stale generated output overriding source-of-truth changes. Remove generated build output from this PR.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `b020cf93`

[Automated Review - pr-reviewer] The PR description claims a focused daemon transcript-backfill fix, but the diff includes large unrelated generated/build artifacts and adapter dist surface changes that are not part of that fix.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The provided diff shows extensive additions outside the stated daemon files: `packages/cli/dashboard/.svelte-kit/**` generated output and `packages/adapters/openclaw/dist/**` API/type surface files. This materially diverges from the PR scope and introduces review-risking changes unrelated to the bugfix.

packages/cli/dashboard/.svelte-kit/ambient.d.ts
```diff
@@ -0,0 +1,362 @@
...
+	export const npm_config_user_agent: string;
+	export const npm_execpath: string;
+	export const XDG_RUNTIME_DIR: string;
+	export const PYENV_ROOT: string;
+	export const DEBUGINFOD_URLS: string;
+	export const npm_package_json: string;
+	export const QT_AUTO_SCREEN_SCALE_FACTOR: string;
+	export const JOURNAL_STREAM: string;
+	export const XCURSOR_THEME: string;
+	export const XDG_DATA_DIRS: string;
```
PR-Reviewer-Ant: Private runtime environment variable names (including sensitive-token-like names such as `docker_hub_token`) are embedded in committed generated typing output. Even without values, this leaks internal environment surface and should not be committed.

.gitignore
```diff
@@ -35,6 +35,7 @@
 .turbo/
 .vscode/
 .worktrees/
+.sisyphus/
```
PR-Reviewer-Ant: The PR adds `.sisyphus/` ignore while also introducing many generated artifacts elsewhere; this reinforces that unrelated workspace/build byproducts were included. Keep this PR strictly to transcript-backfill changes to preserve correctness reviewability.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `57e9bc59`

[Automated Review by pr-reviewer] I found no new blocking correctness or security issues in the current diff; this looks ready for human review.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff specifically hardens `rewriteReplacingLiveOnlySessions` to avoid dropping non-live records and to fail closed on empty replacement transcripts, and adds focused tests that directly exercise those edge cases. The implementation now matches the PR claim of replacing only live-only sessions while preserving healed/non-live data.

platform/daemon/src/session-transcripts.ts
```diff
@@ -106,21 +123,38 @@ async function backfillMarkdownTranscriptArtifacts(
...
+				sessionKey: parsed.frontmatter.session_key || null,
+				sessionId: parsed.frontmatter.session_id || null,
+				project: parsed.frontmatter.project || null,
+				capturedAt: parsed.frontmatter.captured_at || new Date().toISOString(),
+				sourceFormat: "markdown" as const,
+				sourcePath: `memory/${name}`,
+				transcript: parsed.body,
+			};
+			const { classification, allKeys } = await getSeen(harness);
+			const key = sessionSeqCacheKey(input);
```
PR-Reviewer-Ant, line 146: Potential data loss: once a session key is in `liveOnlyKeys`, this queues a replacement from markdown without validating recency/completeness versus current JSONL live turns. If artifact content is stale, `rewriteReplacingLiveOnlySessions` will replace newer live turns with older transcript text.

platform/daemon/src/session-transcripts.ts
```diff
@@ -164,23 +221,49 @@ async function backfillDatabaseTranscripts(
...
+					sessionKey: row.session_key,
+					project: row.project,
+					capturedAt: row.updated_at || row.created_at || new Date().toISOString(),
+					sourceFormat: "db" as const,
+					transcript: row.content,
+				};
+				const { classification, allKeys } = await getSeen(harness);
+				const key = sessionSeqCacheKey(input);
+				if (classification.liveOnlyKeys.has(key)) {
+					const replacements =
```
PR-Reviewer-Ant, line 245: Same overwrite risk on DB backfill path: live-only sessions are replaced from DB content without freshness checks (timestamp/version/turn-count guard). A stale DB row can clobber more recent live-only turns in JSONL.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `eaa99c85`

[Automated Review by pr-reviewer] Found a potential transcript data-loss regression in the new live-only replacement path; otherwise the rewrite approach matches the PR’s stated backfill goal.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The new control flow routes live-only sessions through replacement using markdown/DB transcript payloads, and the replacement unconditionally swaps live records when a key is classified live-only. From the diff alone, there is no guard that replacement content is newer or at least as complete as current live content, so stale artifacts/rows can overwrite fresher live turns.

platform/daemon/src/session-transcripts.ts
Ostico: This is by design — not a data-loss risk.

Live-only sessions contain PARTIAL turns (just the last few appended during prompt-submit hooks). The markdown artifact contains the FULL session transcript captured at session-end. By definition, the full transcript is always a superset of the partial live turns.

The freshness concern would only apply if the artifact were somehow older than the JSONL live content. That cannot happen: artifacts are written at session-end, and live turns are written during session (before session-end). The artifact always post-dates the live turns for the same session.

Timeline: prompt-submit → live turns written → session-end → artifact written → backfill replaces live with artifact.

platform/daemon/src/session-transcripts.ts
Ostico: Same reasoning applies to the DB backfill path.

DB rows are written by the memory pipeline AFTER the session ends. Live-only turns in the JSONL are partial turns from prompt-submit hooks DURING the session. The DB content is always a superset (full parsed transcript) of the partial live turns.

A "stale DB row" cannot exist for this case because:
1. DB writes happen after session-end (always newer than mid-session live turns)
2. A session classified as "live-only" means NO canonical/normalized records exist for it — only live partial turns
3. The backfill replaces partial fragments with the complete record

No freshness guard is needed because the data flow guarantees monotonic completeness.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `9e124cba`

[Automated Review by pr-reviewer] I found a data-integrity regression in the new live-only rewrite path that can corrupt per-session sequencing state.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - In rewriteReplacingLiveOnlySessions, keys are marked as replaced even when no replacement is actually written (healed non-live record path), then sessionSeqCache is overwritten from the replacement payload. That can reduce cached seq below the true on-disk seq and cause subsequent appendCanonicalTranscriptTurns writes to reuse sequence numbers for that session.

platform/daemon/src/transcript-jsonl.ts
```diff
@@ -477,6 +547,95 @@ export function appendCanonicalTranscriptSnapshotIfMissing(
...
+						continue;
+					}
+					const next = transcriptTextToTurns(entry.transcript)
+						.map((turn, index) => makeRecord(entry.identity, turn, index + 1))
+						.filter((r): r is CanonicalTranscriptRecord => r !== null);
+					if (next.length === 0) {
+						// Replacement payload is empty — preserve original line (fail-closed).
+						writeSync(fd, `${line}\n`);
+						continue;
+					}
```
PR-Reviewer-Ant, line 627: Blocking: `replaced.add(key)` is executed in the `record.source_format !== "live"` branch even though no replacement is written. Later, all `replaced` keys are used to reset `sessionSeqCache` from replacement transcript length. For healed sessions, this can clobber a higher real seq with a smaller value, leading to duplicate/non-monotonic seq values on later live appends (data integrity risk). Track a separate set for 'actually rewritten' keys, or only update cache for keys where replacement lines were emitted.

Ostico: Fixed. Separated `replaced` into two sets:

- `rewritten` — sessions where replacement lines were actually emitted
- `healed` — sessions that were already non-live (replacement skipped)

Only `rewritten` keys update `sessionSeqCache`. Healed sessions keep their existing (correct) seq from the session-end hook.

Regression test added: `"healed session does not clobber seq cache with replacement turn count"` — writes 5 canonical turns, calls rewrite with a 2-turn replacement (which gets skipped because session is already non-live), then appends a live turn and asserts it gets seq 6 (not seq 3).

platform/daemon/src/transcript-jsonl.ts
```diff
@@ -477,6 +547,99 @@ export function appendCanonicalTranscriptSnapshotIfMissing(
...
+					try {
+						const parsed = JSON.parse(trimmedLine) as Partial<CanonicalTranscriptRecord>;
+						if (parsed.schema !== "signet.transcript.v1" || typeof parsed.content !== "string") {
+							writeSync(fd, `${line}\n`);
+							continue;
+						}
+					const record = parsed as CanonicalTranscriptRecord;
+					const key = recordSeqCacheKey(record);
+					if (!replacements.has(key)) {
+						writeSync(fd, `${line}\n`);
```
PR-Reviewer-Ant, line 608: Blocking: `rewriteReplacingLiveOnlySessions` can still remove existing non-live records for a target session when the first encountered record for that key is `live`. After `rewritten.add(key)`, later non-live lines for the same key fall into `if (rewritten.has(key) || healed.has(key))` and are preserved, but the initial rewrite already emitted a full replacement snapshot and future non-live records may now duplicate or conflict sequence/content semantics. More importantly, because replacement is triggered on first live line rather than after confirming the entire session is live-only, this function depends on input classification being perfectly in-sync with file ordering/state, which is race-prone under concurrent writes and can rewrite sessions that are no longer live-only.

platform/daemon/src/daemon.ts
```diff
@@ -4,6 +4,7 @@
  * Background service for memory, API, and dashboard hosting
  */
 
+import "./bun-socket-polyfill";
```
PR-Reviewer-Ant, line 7: The PR description is transcript/OOM/backfill-specific, but this commit also introduces a global `Duplex.prototype` polyfill via side-effect import. That broad runtime behavior change should be explicitly called out in the PR description and risk-assessed, since it expands process-wide attack/debug surface and changes socket behavior outside transcript code paths.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `dd41fc2e`

[Automated Review by pr-reviewer] I found one blocking correctness issue and one scope divergence from the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed logic in transcript-jsonl.ts can deterministically drop existing canonical records when replacing a live-only session key that appears with mixed source formats in-file. This is directly visible in the new control flow. The PR also introduces an unrelated daemon socket polyfill not reflected in the stated transcript-focused scope.

platform/daemon/src/transcript-jsonl.ts
Ostico: Fixed. The rewrite now re-classifies inside the file lock before proceeding:

1. **Pre-scan** (inside lock): streams through the file to find which target sessions have ANY non-live records
2. **Filter**: removes those sessions from the effective replacement set
3. **Rewrite**: only processes sessions confirmed live-only at lock time

This eliminates the race entirely — if a session-end hook writes non-live records between external classification and lock acquisition, the pre-scan detects them and skips the replacement.

The in-loop healed detection is removed (no longer needed). The rewrite loop is now simpler: only processes sessions that are guaranteed live-only.

Regression test added: `"skips replacement when session has both live and non-live records (race guard)"` — writes live + normalized records for the same session, then attempts replacement. Asserts both originals are preserved and replacement is not applied.

platform/daemon/src/daemon.ts
Ostico: PR description updated to explicitly call out the polyfill (see "destroySoon Fix" section in body). The risk profile:

- **Scope**: Only affects `Duplex` instances that lack `destroySoon()` (i.e., Bun's HTTP socket wrapper). Standard Node.js sockets already have it and are not affected (no-overwrite guard).\n- **Semantics**: Exact match to Node.js implementation (same code as Bun PR #26264).\n- **Removal**: Tracked to Bun issue #24127 / PR #26264. Module-level comment says when to remove.\n- **Position**: First import ensures the polyfill is applied before any HTTP server is created.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `a92e1bdf`

[Automated Review by pr-reviewer] I did not find blocking correctness, security, or data-integrity regressions in this PR, and the implementation aligns with the stated goals (streaming rewrites/snapshots for OOM avoidance, live-only replacement behavior, seq-cache protection on healed sessions, and Bun destroySoon compatibility).

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The key risk areas described in the PR are directly exercised in the changed code and tests: streaming file rewrite paths in transcript-jsonl.ts, race/healed-session handling in rewriteReplacingLiveOnlySessions, backfill orchestration in session-transcripts.ts, and Bun socket polyfill behavior in dedicated tests. The new incremental race-guard test specifically covers the prior concurrency edge.

platform/daemon/src/session-transcripts.ts
```diff
@@ -39,13 +44,21 @@ export interface TranscriptHit {
...
 }
 
-function createBackfillSeenReader(basePath: string, agentId?: string): (harness: string) => Promise<Set<string>> {
-	const seen = new Map<string, Promise<Set<string>>>();
+interface BackfillSeen {
+	readonly classification: TranscriptSessionKeyClassification;
+	readonly allKeys: Set<string>;
+}
+
+function createBackfillSeenReader(basePath: string, agentId?: string): (harness: string) => Promise<BackfillSeen> {
```
PR-Reviewer-Ant, line 51: Blocking: `allKeys` is cached once and becomes stale after live-only replacement promotion. In markdown backfill you mutate `classification.liveOnlyKeys/canonicalKeys`, but `allKeys` is not kept in sync. Later DB backfill reuses the same cached object and calls `appendCanonicalTranscriptSnapshotIfMissing(input, allKeys)`, which may miss already-promoted keys and append duplicate canonical snapshots for the same session. Please either derive known keys from `classification` at call time, or update `allKeys` whenever promotion occurs.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `305793ee`

[Automated Review by pr-reviewer] I found one blocking data-integrity issue where backfill can duplicate canonical transcript sessions across phases.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The cached `BackfillSeen` object stores `allKeys` once in `createBackfillSeenReader`, but after live-only rewrites the code updates only `classification` sets and never updates `allKeys`. The subsequent DB backfill phase reuses stale `allKeys`, so `appendCanonicalTranscriptSnapshotIfMissing(..., allKeys)` can append duplicate canonical records for already-rewritten sessions.

platform/daemon/src/session-transcripts.ts
NicholaiVogel: Valid. `allKeys` is cached once, but live-only promotion only mutates `classification`, so the later DB backfill can still pass a stale known-key set into `appendCanonicalTranscriptSnapshotIfMissing` and append a duplicate canonical snapshot. This should be fixed by deriving the known keys from `classification` at call time or keeping `allKeys` in sync when promotion succeeds.

NicholaiVogel: Fixed in `db658d36`. I removed the cached `allKeys` state, derive the known-key set from the current classification at each append boundary, and update the classification when an append or live-only promotion succeeds. Added a regression guard covering markdown promotion followed by DB backfill for the same session.

PR-Reviewer-Ant (commented):
> Automated review by [pr-reviewer](https://github.com/NicholaiVogel/pr-reviewer) | model: gpt-5.3-codex | commit: `db658d36`

[Automated Review by pr-reviewer] I did not find new blocking issues in the current diff; the implementation appears to match the PR description’s stated fixes for streaming transcript rewrite/snapshot behavior, live-only replacement guardrails, seq-cache handling, and Bun `destroySoon` compatibility.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed paths include both implementation and focused regression tests for the claimed behaviors (`transcript-jsonl.ts`, `session-transcripts.ts`, and new polyfill tests). The race-guard and healed/live-only classification logic are explicitly exercised in tests. I did not run a live daemon/runtime repro for the Bun socket drain path or large-file performance profile in this review context.

NicholaiVogel:
@Ostico the john wick of bug fixes

PR #612, opened 2026-04-30
fix(daemon): resolve session-end timeout under concurrent load

### LINKED ISSUE #611: Session-end hook times out after 10000ms under concurrent load
## Problem

`/api/hooks/session-end` regularly times out (10s) under concurrent session-end load, causing:

- Client-side timeout errors visible to users (OpenCode, Claude Code, Pi)
- Event loop starvation in the daemon (measured 26.5s lag in production)
- Lost session-end notifications when multiple sessions end simultaneously

## Root Causes

1. **Client-side blocking**: All connectors await the full daemon response synchronously, blocking their event loop for the entire processing duration
2. **Server-side serial work**: `handleSessionEnd()` performs transcript retention, summary enqueue, audit write, JSONL indexing, FTS feedback, and decay propagation all synchronously before responding
3. **Reindex transaction overhead**: `doReindex()` in `memory-lineage.ts` wraps each individual file in its own write transaction (13,744 individual transactions for a typical workspace)

## Impact

- Users see timeout errors in their terminal
- Session summaries may not be generated if the hook is dropped
- Concurrent session-ends cascade into event loop blocking

## Expected Behavior

Session-end hooks should:
- Never block the client (fire-and-forget pattern)
- Return quickly from the HTTP handler (lossless transcript write + enqueue only)
- Batch expensive DB operations to reduce event loop contention

## Environment

- Signet daemon v0.109.16+
- Any harness (OpenCode, Claude Code, Pi, OpenClaw)
- Workspaces with >100 memory artifacts

Ostico:
## Summary

Fixes #611 — `/api/hooks/session-end` times out (10s) under concurrent session-end load, causing client-side errors and event loop starvation.

Three-vector fix: clients fire-and-forget, server defers heavy work to `setImmediate`, and reindex batches DB writes (50-row transactions).

## Type

- [x] Bug fix

## Changes

| File | Change |
|------|--------|
| `integrations/opencode/plugin/src/index.ts` | Fire-and-forget session-end POST (don't await response) |
| `integrations/claude-code/connector/src/index.ts` | Fire-and-forget with AbortSignal timeout |
| `integrations/openclaw/memory-adapter/src/index.ts` | Fire-and-forget via `sessionEndFireAndForget()` |
| `integrations/pi/extension-base/src/lifecycle.ts` | Use configurable `writeTimeout` instead of hardcoded constant |
| `libs/sdk/src/client-p2.ts` | Add `sessionEndFireAndForget()` helper method |
| `platform/daemon/src/hooks.ts` | Split `handleSessionEnd()` into fast-path (transcript retention + summary enqueue) and deferred `setImmediate` (audit, indexing, feedback, decay). Pipeline-disabled guard placed after transcript retention. |
| `platform/daemon/src/memory-lineage.ts` | Batch reindex writes (50-row transactions via `upsertArtifactRowInTx` / `deleteArtifactRowsForPathInTx`). Stage cache mutations only after successful DB commit to prevent shared-state corruption. |
| `platform/daemon/src/memory-lineage.test.ts` | Regression test: batch-threshold cache staging correctness |

## PR Readiness Checklist

## Testing

- [x] Manual integration test (daemon from source, v0.109.16)
- [x] Reindex benchmark: 2.8s → 192ms (93% reduction)
- [x] Client timeout eliminated: fire-and-forget never blocks
- [x] Regression test: batch-threshold cache staging (26 tests pass)
- [x] LSP diagnostics clean on all changed files
- [ ] Full test suite (pre-existing failures unrelated to this branch: 29/1625 in daemon)

## AI Disclosure

- [x] AI-assisted — OpenCode + Claude Opus 4

## Notes

- The server-side fast path still takes ~5.28s due to `writeCanonicalTranscriptFromSnapshot()` performing a full-file JSONL rewrite under lock. This is not a regression — it's the pre-existing bottleneck. The timeout bug is fixed because clients no longer await the response.
- Known remaining IMPORTANT-level items (non-blocking): same-runtime duplicate session-end race (pre-existing, fire-and-forget widens the window), deferred transcript indexing has no retry mechanism, connector contracts now weaker (Claude Code always returns success).
- Batch size of 50 aligns with existing `yieldEvery(50)` pattern in the codebase.


PR-Reviewer-Ant:
Hi @Ostico - I'm taking a look at the fixes in `fix(daemon): resolve session-end timeout under concurrent load` (commit `c346b419`) and will follow up shortly.

platform/daemon/src/hooks.ts
```diff
@@ -3187,22 +3183,119 @@ export async function handleSessionEnd(req: SessionEndRequest): Promise<SessionE
 		}
```
PR-Reviewer-Ant, line 3195: This guard now returns before `deferSessionEndWork()` is scheduled, so a pipeline-disabled daemon skips the transcript audit/canonical transcript indexing path entirely. That diverges from the PR description's "pipeline-disabled guard placed after transcript retention" claim and can drop session-end transcript retention for users who disable extraction but still expect transcripts to be captured. The guard needs to sit after the retention work is scheduled, with only extraction/feedback gated.

platform/daemon/src/memory-lineage.ts
```diff
@@ -746,10 +764,61 @@ export async function reindexMemoryArtifacts(agentId?: string): Promise<void> {
...
+	// Stage cache mutations separately — only apply after successful commit.
+	// This prevents the shared-state bug where a rolled-back transaction
+	// leaves the in-memory cache advanced, causing future reindexes to skip files.
+	const stagedCacheUpdates: Array<{ path: string; statKey: string }> = [];
+	const stagedCacheDeletes: string[] = [];
+	const stagedChangedPaths: string[] = [];
+	const commitStagedUpdates = (): void => {
+		for (const { path, statKey } of stagedCacheUpdates) {
+			cache.set(path, statKey);
+		}
```
PR-Reviewer-Ant: The staged cache buffers are shared across upserts and deletes, but the DB batches flush independently. If an invalid/tombstoned file queues a delete, then 50 valid files trigger `flushUpsertBatch()`, `commitStagedUpdates()` also advances the cache for the still-pending delete before that delete transaction has committed. If the later delete batch fails, the cache is now ahead of the DB and future reindexes can skip the stale row. This is the same class of cache/DB divergence the PR says it fixes; staged mutations need to be per batch or tagged so only the successfully committed batch is applied.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c346b419`

</details>

I found two correctness/data-integrity issues. The deferred session-end split skips transcript retention when the pipeline is disabled, and the reindex cache staging can still advance cache entries before the matching DB batch commits.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the changed code: handleSessionEnd returns before deferSessionEndWork is scheduled when pipelineV2 is disabled, and doReindex uses one shared staged cache buffer for both pending upserts and pending deletes while flushing those DB batches independently.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `c346b419`

</details>

I found two correctness/data-integrity issues. The deferred session-end split skips transcript retention when the pipeline is disabled, and the reindex cache staging can still advance cache entries before the matching DB batch commits.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the changed code: handleSessionEnd returns before deferSessionEndWork is scheduled when pipelineV2 is disabled, and doReindex uses one shared staged cache buffer for both pending upserts and pending deletes while flushing those DB batches independently.

platform/daemon/src/memory-lineage.ts
```diff
@@ -746,10 +764,61 @@ export async function reindexMemoryArtifacts(agentId?: string): Promise<void> {
...
+		if (pendingUpserts.length === 0) return false;
+		getDbAccessor().withWriteTx((db) => {
+			for (const item of pendingUpserts) {
+				upsertArtifactRowInTx(db as WriteDb as Database, item.path, item.frontmatter, item.body, item.mtime);
+			}
+		});
+		pendingUpserts.length = 0;
+		commitStagedUpdates();
+		return true;
+	};
```
PR-Reviewer-Ant, line 816: This still lets cache state get ahead of the database in mixed batches. `stagedCacheUpdates` / `stagedChangedPaths` are shared across both queues, but `flushDeleteBatch()` commits all staged updates after only the delete transaction succeeds. If the loop has staged some upserts, then hits 50 deletes and flushes them first, the cache can mark those upsert files as indexed before their upsert transaction has run. A later upsert failure would leave the DB missing rows while future reindexes skip the files because the cache was already advanced. The staged cache data needs to be tied to the specific batch being committed, not one shared buffer across independent delete and upsert transactions.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `886b2d3b`

</details>

I found one batching bug in the reindex cache staging path. The PR is trying to make cache mutations transaction-safe, but mixed delete/upsert batches can still advance cache state for rows whose DB write has not committed.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The shared stagedCacheUpdates array is committed wholesale from both flushUpsertBatch and flushDeleteBatch in platform/daemon/src/memory-lineage.ts. Because pending upserts and pending deletes are flushed independently, one successful delete flush can apply cache updates for pending upserts that have not been written yet; if the later upsert transaction fails, future reindexes can skip those files.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `886b2d3b`

</details>

I found one batching bug in the reindex cache staging path. The PR is trying to make cache mutations transaction-safe, but mixed delete/upsert batches can still advance cache state for rows whose DB write has not committed.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The shared stagedCacheUpdates array is committed wholesale from both flushUpsertBatch and flushDeleteBatch in platform/daemon/src/memory-lineage.ts. Because pending upserts and pending deletes are flushed independently, one successful delete flush can apply cache updates for pending upserts that have not been written yet; if the later upsert transaction fails, future reindexes can skip those files.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/hooks.ts
Ostico: Fixed in 5ce0ac19. The pipeline-disabled guard no longer returns early — it now only skips summary extraction. `deferSessionEndWork()` always runs regardless of pipeline state, so transcript audit and JSONL indexing are preserved. The feedback/decay section inside deferred work was already gated on `pipelineV2.graph.enabled && pipelineV2.feedback.enabled`.

platform/daemon/src/memory-lineage.ts
Ostico: Fixed in 5ce0ac19. Replaced the shared staging buffers with per-batch cache isolation:

- `PendingUpsert` and `PendingDelete` now carry their own `statKey` and `markChanged` flag
- `flushUpsertBatch()` snapshots its batch, commits to DB, then applies only that batch's cache mutations via `commitBatchCacheUpserts(batch)`\n- `flushDeleteBatch()` does the same via `commitBatchCacheDeletes(batch)`\n- A separate `cacheOnlyUpdates[]` handles unreadable files that need no DB write\n\nThis guarantees a cache mutation is committed only after the DB transaction that processed that specific item succeeds. Cross-batch contamination is no longer possible.

platform/daemon/src/memory-lineage.ts
Ostico: Same fix as above (5ce0ac19). The shared `stagedCacheUpdates` / `stagedCacheDeletes` / `stagedChangedPaths` are gone entirely. Each batch type now carries its own cache mutations embedded in the pending items and only commits them after its own transaction succeeds.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5ce0ac19`

</details>

I found one data-retention regression in the session-end refactor: long canonical transcript artifacts now get truncated before they are written.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff moves the 100k transcript cap ahead of deferred transcript artifact writing in platform/daemon/src/hooks.ts. The old flow wrote transcript artifacts before applying the cap used for summary enqueueing; the new deferred path receives the already-truncated transcript.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `5ce0ac19`

</details>

I found one data-retention regression in the session-end refactor: long canonical transcript artifacts now get truncated before they are written.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff moves the 100k transcript cap ahead of deferred transcript artifact writing in platform/daemon/src/hooks.ts. The old flow wrote transcript artifacts before applying the cap used for summary enqueueing; the new deferred path receives the already-truncated transcript.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
@Ostico apologies for the duplicate replies with the pr-reviewer, I'm reworking some of his capabilities and a few things are a little broken

Ostico:
Hi @NicholaiVogel no prob, maybe do you want to try this? https://github.com/ostico/test-guard I built it, if you try it please give feedbacks to improve.

NicholaiVogel:
@Ostico I'll try to look into it this week, thank you for the suggestion!

NicholaiVogel:
Addressed the latest retention regression in `bb58113a`.

What changed:
- Kept a lossless `retainedTranscript` for session transcript storage, canonical transcript snapshots, and deferred JSONL indexing.
- Moved the 100k safety cap onto a separate `summaryTranscript` copy used only for summary/extraction enqueueing.
- Derived fallback session IDs from the retained full transcript, matching the old artifact-path behavior.
- Added a regression test proving long transcripts keep their tail marker in canonical transcript artifacts and `session_transcripts`, while the queued summary job receives the capped `[truncated]` copy.
- Updated affected async/deferred-work tests to wait for the deferred session-end work before checking manifests/audit files.

Validated:
- `bun test platform/daemon/src/hooks.test.ts` — 160 pass
- `./node_modules/.bin/biome check platform/daemon/src/hooks.ts platform/daemon/src/hooks.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `bb58113a`

</details>

I found one correctness issue: the Claude Code path is described as fire-and-forget, but the pending fetch and 10s abort timer can still keep the hook process alive. That leaves the PR’s main timeout fix incomplete for the short-lived hook case.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed Claude connector starts a fetch with AbortSignal.timeout(10000) and returns immediately, but there is no unref/detach mechanism around the pending request or timer. For a short-lived hook command, pending network I/O can keep the process alive until the daemon responds or the timeout fires; I did not have a runtime trace of the packaged Claude hook process to confirm exact process-exit behavior.

PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `da3b7da5`.

NicholaiVogel:
Addressed the new Claude Code fire-and-forget feedback in `da3b7da5`.

The parent hook process no longer starts an in-process `fetch()` with a live abort timer. Instead, `onSessionEnd()` now spawns a detached child process with ignored stdio and calls `unref()` immediately; that child performs the best-effort POST to `/api/hooks/session-end` with its own 10s timeout. The Claude Code hook command can return without being kept alive by pending network I/O or the timeout timer.

I also added a regression test around the dispatcher to assert it:
- uses a detached child process
- ignores stdio
- calls `unref()`
- sends the expected session-end URL and payload

Validated:
- `cd integrations/claude-code/connector && bun test index.test.ts`
- `cd integrations/claude-code/connector && bun run build`
- `./node_modules/.bin/biome check integrations/claude-code/connector/src/index.ts integrations/claude-code/connector/index.test.ts`
- `git diff --check`

integrations/claude-code/connector/index.test.ts
```diff
@@ -14,18 +14,53 @@ beforeEach(() => {
 
 afterEach(() => {
 	if (origHome !== undefined) process.env.HOME = origHome;
-	else delete process.env.HOME;
+	else process.env.HOME = undefined;
```
PR-Reviewer-Ant: This should delete `process.env.HOME`, not assign `undefined`. In Node/Bun, assigning `undefined` to `process.env` can leave the environment variable present as the string value `"undefined"`, which means later tests can resolve `homedir()`/Claude config paths under a bogus HOME. The previous code correctly removed the variable when it was originally absent; please keep that behavior.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `da3b7da5`

</details>

I found one test-environment regression that can leak a bogus HOME value into later tests. The main session-end changes otherwise match the PR description at the diff level I reviewed.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The HOME restoration bug is directly visible in the changed Claude connector test. I did not run the connector or daemon runtime, so I am not claiming coverage of the timeout behavior under real concurrent load.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

integrations/claude-code/connector/index.test.ts
NicholaiVogel: Addressed the HOME cleanup feedback in `93dc0d12`.

The connector test cleanup no longer assigns `undefined` to `process.env.HOME`. When HOME was originally absent, it now removes the key with `Reflect.deleteProperty(process.env, "HOME")`, preserving the prior behavior without triggering Biome's `noDelete` rule.

Validated:
- `cd integrations/claude-code/connector && bun test index.test.ts`
- `cd integrations/claude-code/connector && bun run build`
- `./node_modules/.bin/biome check integrations/claude-code/connector/index.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `93dc0d12`

</details>

I found one correctness issue in the server-side split: the deferred feedback path can still mutate graph feedback state even when the memory pipeline is disabled. That weakens the claimed pipeline-disabled guard and changes disabled-mode behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff moves the pipeline-disabled guard into handleSessionEnd's summary enqueue path, but deferSessionEndWork is scheduled unconditionally and its feedback block only checks graph.enabled and feedback.enabled. This is directly visible in platform/daemon/src/hooks.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `93dc0d12`

</details>

I found one correctness issue in the server-side split: the deferred feedback path can still mutate graph feedback state even when the memory pipeline is disabled. That weakens the claimed pipeline-disabled guard and changes disabled-mode behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff moves the pipeline-disabled guard into handleSessionEnd's summary enqueue path, but deferSessionEndWork is scheduled unconditionally and its feedback block only checks graph.enabled and feedback.enabled. This is directly visible in platform/daemon/src/hooks.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Addressed the deferred feedback disabled-mode issue in `794d6cf2`.

The deferred session-end path now gates graph feedback behind the same pipeline-active check used for extraction (`pipelineV2.enabled || pipelineV2.shadowMode`) in addition to the existing graph/feedback flags. Transcript retention and canonical JSONL indexing still run unconditionally, but FTS overlap feedback, decay, and graph status propagation no longer mutate graph feedback state when the memory pipeline is disabled.

I also added a regression test that disables the pipeline, seeds feedback-eligible session memory/aspect rows, runs session-end, and verifies:
- no summary job is queued
- canonical transcript retention still happens
- the aspect weight remains unchanged

Validated:
- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "skips deferred graph feedback when pipeline is disabled"`
- `bun test platform/daemon/src/hooks.test.ts` — 161 pass
- `./node_modules/.bin/biome check platform/daemon/src/hooks.ts platform/daemon/src/hooks.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `794d6cf2`

</details>

I found one correctness issue in the new Claude Code fire-and-forget path: the detached child script reads the wrong argv slots, so it never posts to the daemon. That means Claude session-end retention/extraction is effectively dropped while the connector still reports success.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The spawned command passes the URL and JSON body after the --eval script, but Node's eval argv handling leaves process.argv[1] as the first user argument. The child script uses process.argv.slice(1), so url becomes the body argument only if argv[1] is present as expected; this is easy to get wrong and the current unit test only asserts the parent-side args, not that the eval script actually sends the request. A real child-process integration test would catch whether the daemon receives the POST.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `794d6cf2`

</details>

I found one correctness issue in the new Claude Code fire-and-forget path: the detached child script reads the wrong argv slots, so it never posts to the daemon. That means Claude session-end retention/extraction is effectively dropped while the connector still reports success.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The spawned command passes the URL and JSON body after the --eval script, but Node's eval argv handling leaves process.argv[1] as the first user argument. The child script uses process.argv.slice(1), so url becomes the body argument only if argv[1] is present as expected; this is easy to get wrong and the current unit test only asserts the parent-side args, not that the eval script actually sends the request. A real child-process integration test would catch whether the daemon receives the POST.

NicholaiVogel:
Addressed the detached child argv feedback in `ae1cf5f0`.

The fire-and-forget child no longer depends on `process.argv` slot behavior for `--eval`. The parent now passes the session-end URL and JSON payload through explicit child environment variables (`SIGNET_SESSION_END_URL` / `SIGNET_SESSION_END_BODY`), and the eval script reads those values before performing the POST. This removes the Node/Bun argv ambiguity entirely while keeping the child detached with ignored stdio and `unref()`.

I also added a real child-process regression test: it starts a local HTTP server, dispatches session-end through the detached child, and asserts the server actually receives the POST at `/api/hooks/session-end` with the expected JSON payload.

Validated:
- `cd integrations/claude-code/connector && bun test index.test.ts` — 4 pass
- `cd integrations/claude-code/connector && bun run build`
- `./node_modules/.bin/biome check integrations/claude-code/connector/src/index.ts integrations/claude-code/connector/index.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `ae1cf5f0`

</details>

I found one correctness issue in the Claude Code fire-and-forget path: it reports success even when the detached dispatch could not be started, which weakens the connector contract more than the implementation needs to.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed connector exports a boolean from dispatchSessionEndFireAndForget, catches spawn failures there, and then ignores that boolean in onSessionEnd while unconditionally returning success. This is directly visible in integrations/claude-code/connector/src/index.ts.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `ae1cf5f0`

</details>

I found one correctness issue in the Claude Code fire-and-forget path: it reports success even when the detached dispatch could not be started, which weakens the connector contract more than the implementation needs to.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed connector exports a boolean from dispatchSessionEndFireAndForget, catches spawn failures there, and then ignores that boolean in onSessionEnd while unconditionally returning success. This is directly visible in integrations/claude-code/connector/src/index.ts.

NicholaiVogel:
Addressed the dispatch-result feedback in `1dbd13ac`.

`onSessionEnd()` now preserves the detached dispatch result instead of ignoring it. If the detached child cannot be spawned, `dispatchSessionEndFireAndForget()` returns `false` and `onSessionEnd()` now returns `{ success: false, memoriesExtracted: 0 }`; successful startup still returns `{ success: true, memoriesExtracted: 0 }` without waiting for the best-effort POST to complete.

I also added a regression test that overrides the dispatch hook to simulate startup failure and verifies `onSessionEnd()` reports `success: false` while still passing the expected Claude Code session-end payload.

Validated:
- `cd integrations/claude-code/connector && bun test index.test.ts` — 5 pass
- `cd integrations/claude-code/connector && bun run build`
- `./node_modules/.bin/biome check integrations/claude-code/connector/src/index.ts integrations/claude-code/connector/index.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1dbd13ac`

</details>

I found one issue in the reindex path: the batching change stops yielding for unchanged cached files, so large warm reindexes can still monopolize the event loop. That undercuts the PR's stated concurrency/starvation fix even though the DB write batching itself looks directionally right.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed loop in platform/daemon/src/memory-lineage.ts now continues immediately on cache hits and only calls the yielder after flushed write/delete batches. I did not have a runtime trace for the original timeout, so confidence is medium rather than high, but the starvation path is directly visible in the diff.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `1dbd13ac`

</details>

I found one issue in the reindex path: the batching change stops yielding for unchanged cached files, so large warm reindexes can still monopolize the event loop. That undercuts the PR's stated concurrency/starvation fix even though the DB write batching itself looks directionally right.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed loop in platform/daemon/src/memory-lineage.ts now continues immediately on cache hits and only calls the yielder after flushed write/delete batches. I did not have a runtime trace for the original timeout, so confidence is medium rather than high, but the starvation path is directly visible in the diff.

NicholaiVogel:
Addressed the warm reindex yielding feedback in `76a13424`.

Warm/no-op artifact reindexes now still call the reindex yielder when a file is missing/unstatable or when its cached stat key matches and the loop skips the full read/write path. That keeps large cached scans from monopolizing the event loop even when there are no DB batches to flush.

I also added a regression test that warms the artifact index cache with more than one batch of summary artifacts, instruments `setImmediate`, runs a second unchanged reindex, and verifies the warm-cache skip path yields.

Validated:
- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "warm reindex yields|memory-lineage"`
- `bun test platform/daemon/src/hooks.test.ts` — 162 pass
- `./node_modules/.bin/biome check platform/daemon/src/memory-lineage.ts platform/daemon/src/hooks.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `76a13424`

</details>

I found one implementation mismatch worth discussing: the SDK helper was added, but the OpenClaw adapter still hand-rolls its own fire-and-forget path, so the PR does not quite deliver the shared-helper behavior it claims. I did not find a concrete blocker in the changed daemon batching or deferred session-end path from the provided diff.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed files and PR description are enough to verify the helper-use mismatch directly: libs/sdk/src/client-p2.ts adds sessionEndFireAndForget(), while integrations/openclaw/memory-adapter/src/index.ts continues calling daemonFetch(...).catch(...) inline. I did not run the daemon under concurrent load or inspect the full server route wrapper around handleSessionEnd, so timeout and process-lifetime behavior are assessed from the diff rather than runtime traces.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `76a13424`

</details>

I found one implementation mismatch worth discussing: the SDK helper was added, but the OpenClaw adapter still hand-rolls its own fire-and-forget path, so the PR does not quite deliver the shared-helper behavior it claims. I did not find a concrete blocker in the changed daemon batching or deferred session-end path from the provided diff.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed files and PR description are enough to verify the helper-use mismatch directly: libs/sdk/src/client-p2.ts adds sessionEndFireAndForget(), while integrations/openclaw/memory-adapter/src/index.ts continues calling daemonFetch(...).catch(...) inline. I did not run the daemon under concurrent load or inspect the full server route wrapper around handleSessionEnd, so timeout and process-lifetime behavior are assessed from the diff rather than runtime traces.

NicholaiVogel:
Addressed the OpenClaw/SDK helper mismatch in `184c54f7`.

OpenClaw’s `onSessionEnd()` now uses `SignetClient.sessionEndFireAndForget()` instead of hand-rolling its own `daemonFetch(...).catch(...)` call. I also widened the SDK helper payload type so integration callers can pass the full session-end metadata OpenClaw already forwards (`harness`, `agentId`, `transcriptPath`, `transcript`, `sessionId`, `cwd`, `reason`, and `runtimePath`) without losing daemon context.

Validated:
- `bun run --cwd libs/sdk build`
- `cd integrations/openclaw/memory-adapter && bun test src/index.test.ts` — 89 pass
- `cd integrations/openclaw/memory-adapter && bun run build`
- `./node_modules/.bin/biome check libs/sdk/src/client-p2.ts integrations/openclaw/memory-adapter/src/index.ts integrations/openclaw/memory-adapter/src/signet-sdk.d.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `184c54f7`

</details>

I found one gap between the stated server-side fix and the implementation: the daemon still does the known slow canonical transcript rewrite before responding. That leaves the endpoint itself vulnerable to the same concurrent-load timeout/starvation pattern for any caller that still waits on `/api/hooks/session-end`.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The diff directly shows `handleSessionEnd()` still awaiting `writeCanonicalTranscriptFromSnapshot()` before scheduling deferred work with `setImmediate()`. I did not run the PR's concurrent-load benchmark, so the severity is based on the code path and the PR's own note that this rewrite still takes about 5.28s.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `184c54f7`

</details>

I found one gap between the stated server-side fix and the implementation: the daemon still does the known slow canonical transcript rewrite before responding. That leaves the endpoint itself vulnerable to the same concurrent-load timeout/starvation pattern for any caller that still waits on `/api/hooks/session-end`.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The diff directly shows `handleSessionEnd()` still awaiting `writeCanonicalTranscriptFromSnapshot()` before scheduling deferred work with `setImmediate()`. I did not run the PR's concurrent-load benchmark, so the severity is based on the code path and the PR's own note that this rewrite still takes about 5.28s.

NicholaiVogel:
Addressed the server-side canonical transcript rewrite feedback in `b74d8d59`.

`handleSessionEnd()` now only persists the live transcript snapshot synchronously before returning. The slow canonical JSONL rewrite and transcript indexing are moved into the existing deferred session-end work path, with a tracked deferred-work promise so tests can await completion deterministically.

I added a regression test that verifies the session-end response returns before `memory/<harness>/transcripts/transcript.jsonl` exists, then confirms the canonical JSONL is written once deferred work completes.

Validated:
- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "defers canonical|canonical JSONL|full canonical"`
- `bun test platform/daemon/src/hooks.test.ts` — 163 pass
- `./node_modules/.bin/biome check platform/daemon/src/hooks.ts platform/daemon/src/hooks.test.ts`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b74d8d59`

</details>

I found two concrete correctness issues: the Claude Code detached child path will fail under Node, and the daemon hook file has an invalid async return type that should fail typecheck.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in changed code: the spawned script uses top-level await via node --eval without module input, and an async function in hooks.ts is annotated as returning void.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

integrations/claude-code/connector/src/index.ts
```diff
@@ -60,6 +61,56 @@ export interface SessionEndResult {
 	memoriesExtracted: number;
 }
 
+export interface SessionEndFireAndForgetPayload {
+	harness: "claude-code";
+	sessionId?: string;
```
PR-Reviewer-Ant, line 66: This detached child script uses top-level `await`, but `dispatchSessionEndFireAndForget()` runs it with `process.execPath --eval` and no `--input-type=module`. Under the Node runtime this is parsed as CommonJS and exits with a syntax error before sending the POST. The new test passes under Bun because `process.execPath` is Bun there, but the connector package targets Node and the installed hook can run under Node. Wrap the fetch in an async IIFE or spawn with module input so the fire-and-forget path actually fires.

PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `44622c91`.

integrations/openclaw/memory-adapter/package.json
```diff
@@ -33,6 +33,7 @@
     "test": "bun test"
   },
   "dependencies": {
+    "@signet/sdk": "workspace:*",
     "@sinclair/typebox": "0.34.47"
```
PR-Reviewer-Ant, line 37: This package is publishable, but this adds a runtime dependency as `"@signet/sdk": "workspace:*"`. Unless the publish/version rewrite pipeline is guaranteed to replace this before npm packaging, the published OpenClaw adapter can ship with an unresolved workspace protocol dependency or a dependency on a package/version that is not available to adapter consumers. The repo conventions specifically call out publishable packages depending on unpublished workspace packages as an install-integrity failure mode. Either keep the adapter on its local daemon client path, make the SDK dependency a real published semver dependency validated by the publish manifest checks, or add the package-manifest rewrite/validation coverage in this PR.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `530be4b8`

</details>

I found one publish/install integrity issue in the OpenClaw adapter change. The daemon-side timeout work is directionally aligned with the PR description, but this dependency change risks breaking the published adapter package.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff adds a runtime dependency from the publishable @signetai/signet-memory-openclaw package to @signet/sdk using workspace:*, and the repository conventions explicitly call out publishable-package workspace dependency drift as a failure mode.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the Node detached-child/typecheck feedback in `530be4b8`.

Fixes:
- Replaced the detached child’s top-level `await` with a `void (async () => { ... })();` IIFE, so the `node --eval` path parses under CommonJS eval semantics while still starting the async POST immediately.
- Corrected the async deferred session-end worker return annotation from `void` to `Promise<void>`.

Validated:
- `node --eval '<detached child IIFE shape>'` parses/exits successfully
- `cd integrations/claude-code/connector && bun test index.test.ts && bun run build` — 5 pass + build/type declarations pass
- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "defers canonical"`
- `./node_modules/.bin/biome check integrations/claude-code/connector/src/index.ts platform/daemon/src/hooks.ts`
- `git diff --check`

NicholaiVogel:
the pr reviewer has fangs lol

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `44622c91`

</details>

I found one correctness issue in the OpenClaw session-end path: the rewrite switches away from the adapter's daemon helper and drops the runtime-path headers that this integration has been relying on. That makes the PR's concurrency fix weaker for the OpenClaw path because session-end can be routed differently from the rest of the plugin lifecycle.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The changed OpenClaw code visibly replaces daemonFetch, which always adds pluginHeaders including x-signet-runtime-path and x-signet-actor, with a generic SignetClient call. I cannot inspect the SDK transport in this supplied context to prove it never adds equivalent headers, but the new call site only passes runtimePath in the JSON body and no headers.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

integrations/openclaw/memory-adapter/src/index.ts
```diff
@@ -606,21 +607,23 @@ export async function onSessionEnd(
 		reason?: string;
```
PR-Reviewer-Ant, line 607: This changes the OpenClaw session-end request from `daemonFetch(...)` to the generic SDK client, but `daemonFetchResult` is the adapter path that attaches `pluginHeaders()` (`x-signet-runtime-path: plugin`, `x-signet-actor`, and actor type). The new SDK call only includes `runtimePath` in the JSON body. The daemon/session tracker contract in this repo is header-based for runtime-path routing, so this can make `agent_end` look like a different runtime path than the rest of the OpenClaw session lifecycle and reintroduce conflict/dedup failures under exactly the concurrent session-end load this PR is trying to fix. Keep the fire-and-forget behavior, but preserve the adapter headers or add an SDK fire-and-forget path that can send them.

PR #614, opened 2026-05-01
fix(hermes-agent): harden connector install and diagnostics

NicholaiVogel:
## Summary

- hardens the Hermes connector install flow with config backup/restore, bundled plugin hashing, repo + user plugin copy support, uninstall cleanup, and an idempotent repair path
- adds `signet doctor hermes` plus app-level health surfacing for Hermes configuration, plugin copy freshness, daemon health, and tool routing
- updates Hermes docs with the easier install, doctor, uninstall, and troubleshooting paths
- addresses review feedback for dotted `memory.provider` config, CodeQL URL normalization, repo-plugin staleness detection, Hermes user-provider import loading, uninstall restoration of replaced/removed memory providers, mixed dotted/nested provider config, nested `provider:` keys under `memory:` subfeatures, user-plugin-only doctor behavior, install failure when no usable plugin target is written, and non-mapping `memory:` block safety

## Validation

- `bun test integrations/hermes-agent/connector/index.test.ts` (51 tests, 189 expects)
- `cd integrations/hermes-agent/connector && bun run typecheck`
- `cd integrations/hermes-agent/connector && bun run build`
- `bunx biome check integrations/hermes-agent/connector/src/index.ts integrations/hermes-agent/connector/index.test.ts`
- `git diff --check`
- `bun surfaces/cli/src/cli.ts doctor hermes --json` returns `ok: true` for `/home/nicholai/.hermes`, including daemon health, provider config, user plugin copy, repo plugin copy, and tool routing
- direct `diagnoseHermesIntegration({ hermesHome: "/home/nicholai/.hermes", hermesRepo: null })` returns `ok: true` for the supported user-plugin-only path, with repo-plugin marked optional and tool routing reported from the user plugin schemas
- previous full validation on this PR also covered CLI build, platform core build, spec dependency validation, the broader focused test set, and actual local Hermes user-provider import loading from `$HERMES_HOME/plugins/signet`

## Readiness checklist

Notes: this PR does not add admin or mutation endpoints, so the security checklist item is satisfied by confirming there is no new protected endpoint surface.

### DIFF
--- a/integrations/hermes-agent/connector/src/index.ts
+++ b/integrations/hermes-agent/connector/src/index.ts
@@ -702,11 +702,14 @@ export class HermesAgentConnector extends BaseConnector {
 
 		const hermesHome = this.getHermesHome();
 		const hermesRepo = this.getHermesRepo();
+		let userPluginInstalled = false;
+		let repoPluginInstalled = false;
 
 		// 1. Install the Python plugin into the current user-plugin location.
 		try {
 			const pluginFiles = installPlugin(getUserPluginTargetDir(hermesHome), "user");
 			filesWritten.push(...pluginFiles);
+			userPluginInstalled = true;
 		} catch (e) {
 			const msg = e instanceof Error ? e.message : String(e);
 			warnings.push(`Failed to install Hermes user plugin files: ${msg}`);
@@ -719,10 +722,23 @@ export class HermesAgentConnector extends BaseConnector {
 			try {
 				const pluginFiles = installPlugin(getRepoPluginTargetDir(hermesRepo), "repo");
 				filesWritten.push(...pluginFiles);
+				repoPluginInstalled = true;
 			} catch (e) {
 				const msg = e instanceof Error ? e.message : String(e);
 				warnings.push(`Failed to refresh Hermes repo plugin files: ${msg}`);
 			}
+		}
+		const usablePluginTargetInstalled = hermesRepo ? repoPluginInstalled : userPluginInstalled;
+		if (!usablePluginTargetInstalled) {
+			return {
+				success: false,
+				message: hermesRepo
+					? "Hermes Agent integration failed — could not refresh the Hermes repo Signet provider"
+					: "Hermes Agent integration failed — could not install the Hermes user Signet provider",
+				filesWritten,
+				configsPatched,
+				warnings,
+			};
 		}
 
 		// 2. Write env config for the Signet daemon connection
--- a/integrations/hermes-agent/connector/src/index.ts
+++ b/integrations/hermes-agent/connector/src/index.ts
@@ -341,7 +341,8 @@ function configureProvider(
 	let backupPath: string | null = null;
 	if (dottedProvider !== null) {
 		let changed = false;
-		if (!dottedProviderLineIsSignet(lines[dottedProvider] ?? "")) {
+		const dottedWasSignet = dottedProviderLineIsSignet(lines[dottedProvider] ?? "");
+		if (!dottedWasSignet) {
 			backupPath = writeProviderBackup(
 				hermesHome,
 				configPath,
@@ -352,6 +353,15 @@ function configureProvider(
 			changed = true;
 		}
 		if (block !== null && typeof block === "object" && block.provider !== null && block.provider !== undefined) {
+			if (dottedWasSignet) {
+				const nestedBackupPath = writeProviderBackup(
+					hermesHome,
+					configPath,
+					"nested",
+					parseProviderLine(lines[block.provider] ?? "") ?? "",
+				);
+				backupPath = nestedBackupPath ?? backupPath;
+			}
 			lines.splice(block.provider, 1);
 			changed = true;
 		}
@@ -398,6 +408,14 @@ function restoreOrClearProvider(hermesHome: string): { configPath: string | null
 	let configChanged = false;
 	if (dottedProvider !== null && dottedProviderLineIsSignet(lines[dottedProvider] ?? "")) {
 		setDottedProviderLine(lines, dottedProvider, backup?.providerKind === "dotted" ? backup.previousProvider : "''");
+		if (backup?.providerKind === "nested") {
+			if (block !== null && typeof block === "object") {
+				lines.splice(block.start + 1, 0, `${" ".repeat(block.indent)}provider: ${backup.previousProvider}`);
+			} else {
+				if (lines.length > 0 && lines[lines.length - 1] !== "") lines.push("");
+				lines.push("memory:", `  provider: ${backup.previousProvider}`);
+			}
+		}
 		configChanged = true;
 	} else if (
 		block !== null &&
@@ -884,7 +902,7 @@ export async function diagnoseHermesIntegration(opts?: {
 	readonly daemonUrl?: string;
 }): Promise<HermesDoctorReport> {
 	const hermesHome = opts?.hermesHome ?? resolveHermesHomePath();
-	const hermesRepo = opts?.hermesRepo ?? resolveHermesRepoPath();
+	const hermesRepo = opts && "hermesRepo" in opts ? (opts.hermesRepo ?? null) : resolveHermesRepoPath();
 	const daemonUrl = opts?.daemonUrl ?? (process.env.SIGNET_DAEMON_URL?.trim() || "http://localhost:3850");
 	const configPath = resolveConfigPath(hermesHome);
 	const userPluginDir = getUserPluginTargetDir(hermesHome);
@@ -895,7 +913,9 @@ export async function diagnoseHermesIntegration(opts?: {
 	const repoPluginCurrent = repoPluginDir ? pluginLooksCurrent(repoPluginDir) : false;
 	const probe = hermesRepo
 		? probeHermesProvider(hermesRepo)
-		: { ok: false, toolNames: [], error: "Hermes repo not found" };
+		: userPluginCurrent
+			? { ok: true, toolNames: REQUIRED_TOOL_NAMES, error: null }
+			: { ok: false, toolNames: [], error: "Hermes repo not found and user plugin is missing or stale" };
 	const daemon = await checkDaemon(daemonUrl);
 
 	checks.push({
@@ -926,15 +946,15 @@ export async function diagnoseHermesIntegration(opts?: {
 	checks.push({
 		id: "repo-plugin",
 		label: "Hermes repo plugin copy",
-		ok: repoPluginDir !== null && repoPluginCurrent,
+		ok: repoPluginDir === null || repoPluginCurrent,
 		detail:
 			repoPluginDir === null
-				? "Hermes checkout not found"
+				? "Hermes checkout not found; using user plugin copy only"
 				: repoPluginCurrent
 					? `${repoPluginDir} matches bundled Signet plugin`
 					: `${repoPluginDir} is missing or stale`,
 		fix:
-			repoPluginDir !== null && repoPluginCurrent
+			repoPluginDir === null || repoPluginCurrent
 				? undefined
 				: "Set HERMES_REPO to the Hermes Agent checkout, then run `signet setup --harness hermes-agent`.",
 	});
@@ -943,7 +963,9 @@ export async function diagnoseHermesIntegration(opts?: {
 		label: "Hermes tool routing",
 		ok: probe.ok,
 		detail: probe.ok
-			? `Hermes exposes ${REQUIRED_TOOL_NAMES.join(", ")}`
+			? hermesRepo
+				? `Hermes exposes ${REQUIRED_TOOL_NAMES.join(", ")}`
+				: `User plugin advertises ${REQUIRED_TOOL_NAMES.join(", ")}; runtime probe skipped without Hermes checkout`
 			: `Hermes provider probe failed: ${probe.error ?? "unknown error"}`,
 		fix: probe.ok
 			? undefined
@@ -952,7 +974,7 @@ export async function diagnoseHermesIntegration(opts?: {
 
 	if (!hermesRepo) {
 		warnings.push(
-			"Hermes checkout was not found; repo-plugin and tool-routing checks need HERMES_REPO or ~/.hermes/hermes-agent.",
+			"Hermes checkout was not found; repo-plugin install is optional, and runtime tool-routing probes need HERMES_REPO or ~/.hermes/hermes-agent.",
 		);
 	}
 
--- a/integrations/hermes-agent/connector/src/index.ts
+++ b/integrations/hermes-agent/connector/src/index.ts
@@ -831,6 +831,16 @@ export class HermesAgentConnector extends BaseConnector {
 		if (providerConfig.backupPath) {
 			filesWritten.push(providerConfig.backupPath);
 		}
+		if (!isProviderConfigured(hermesHome)) {
+			return {
+				success: false,
+				message:
+					"Hermes Agent integration incomplete — Signet provider was deployed but not activated in Hermes config",
+				filesWritten,
+				configsPatched,
+				warnings,
+			};
+		}
 
 		if (hermesRepo) {
 			const probe = probeHermesProvider(hermesRepo);
--- a/integrations/hermes-agent/connector/src/index.ts
+++ b/integrations/hermes-agent/connector/src/index.ts
@@ -176,6 +176,14 @@ function leadingWhitespaceLength(line: string): number | null {
 	return match ? (match[1]?.length ?? null) : null;
 }
 
+function isYamlMappingEntry(value: string): boolean {
+	const trimmed = value.trimStart();
+	if (trimmed.startsWith("-")) return false;
+	const colon = trimmed.indexOf(":");
+	if (colon <= 0) return false;
+	return trimmed.slice(0, colon).trim().length > 0;
+}
+
 function findMemoryBlock(lines: string[]): MemoryBlock | "missing" | null {
 	const starts: number[] = [];
 	for (let i = 0; i < lines.length; i++) {
@@ -205,7 +213,10 @@ function findMemoryBlock(lines: string[]): MemoryBlock | "missing" | null {
 	const childIndent = indent ?? 2;
 	for (let i = start + 1; i < end; i++) {
 		const line = lines[i] ?? "";
-		if (leadingWhitespaceLength(line) === childIndent && /^provider:\s*/.test(line.slice(childIndent))) {
+		if (leadingWhitespaceLength(line) !== childIndent) continue;
+		const child = line.slice(childIndent);
+		if (!isYamlMappingEntry(child)) return null;
+		if (/^provider:\s*/.test(child)) {
 			provider = i;
 			break;
 		}
--- a/integrations/hermes-agent/connector/src/index.ts
+++ b/integrations/hermes-agent/connector/src/index.ts
@@ -1,9 +1,11 @@
+import { spawnSync } from "node:child_process";
+import { createHash } from "node:crypto";
 import { existsSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { BaseConnector, type InstallResult, type UninstallResult } from "@signet/connector-base";
-import { expandHome, resolveHermesHomePath, resolveHermesRepoPath, resolveHermesRepoPluginPath } from "@signet/core";
+import { expandHome, resolveHermesHomePath, resolveHermesRepoPath } from "@signet/core";
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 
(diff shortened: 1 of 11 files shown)


PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `45d6458d`

</details>

I found one config-compatibility regression in the Hermes activation path. The main plugin-schema fix is present, but existing dotted Hermes config is no longer recognized and can be rewritten into a conflicting shape.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed connector replaces the previous regex-based provider check with findMemoryBlock/isProviderConfigured, and the new parser only recognizes a nested `memory:` block. The PR description claims setup preserves existing Hermes config where possible, but the diff no longer handles the `memory.provider: signet` form that the removed code explicitly supported.

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -65,28 +117,295 @@ function uninstallPlugin(hermesRepo: string): string[] {
...
+	let end = lines.length;
+	for (let i = start + 1; i < lines.length; i++) {
+		const line = lines[i] ?? "";
+		if (!isBlankOrComment(line) && !/^\s/.test(line)) {
+			end = i;
+			break;
+		}
+	}
+	let provider: number | null = null;
+	for (let i = start + 1; i < end; i++) {
```
PR-Reviewer-Ant, line 214: This no longer recognizes existing `memory.provider: signet` config. The previous implementation explicitly accepted the dotted-key form, but the new path only considers a nested `memory:` block; `configureProvider()` will then append a second `memory:` block instead of treating the existing provider as active. That can leave Hermes config with two competing representations and make `isInstalled()`/doctor report the integration as inactive or freshly patched even though the user already had Signet selected. Please keep dotted-key compatibility or migrate it deliberately before writing the nested block.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `45d6458d`

</details>

I found one config-compatibility regression in the Hermes activation path. The main plugin-schema fix is present, but existing dotted Hermes config is no longer recognized and can be rewritten into a conflicting shape.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed connector replaces the previous regex-based provider check with findMemoryBlock/isProviderConfigured, and the new parser only recognizes a nested `memory:` block. The PR description claims setup preserves existing Hermes config where possible, but the diff no longer handles the `memory.provider: signet` form that the removed code explicitly supported.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `34a27838`

</details>

I found two install-state bugs in the Hermes connector hardening. Both can make Signet report a healthy or installed Hermes integration while Hermes is still using a stale or nonfunctional provider.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The stale-repo detection issue is directly visible in isInstalled(): it accepts any current plugin dir even though this PR says the repo provider can shadow the user provider. The user-plugin import issue is also visible from the new ~/.hermes/plugins/signet install target combined with the bundled plugin's existing plugins.memory.signet.client import, though exact Hermes user-plugin import semantics are external to this diff.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -349,8 +721,115 @@ export class HermesAgentConnector extends BaseConnector {
 	}
 
 	isInstalled(): boolean {
-		return resolveHermesRepoPluginPath() !== null;
+		const hermesHome = this.getHermesHome();
+		const hermesRepo = this.getHermesRepo();
+		const pluginDirs = [
+			getUserPluginTargetDir(hermesHome),
+			...(hermesRepo ? [getRepoPluginTargetDir(hermesRepo)] : []),
+		];
```
PR-Reviewer-Ant: This still reports the connector as installed when the user plugin is fresh but the repo plugin is stale. The PR explicitly notes that bundled repo providers take precedence over user plugins, and install() refreshes both for that reason; with `some(...)`, a stale `<hermes-repo>/plugins/memory/signet` copy can keep shadowing the fixed user copy while `isInstalled()` returns true. When `hermesRepo` is present, this should require the repo plugin to be current, not just any plugin dir.

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -199,22 +552,26 @@ export class HermesAgentConnector extends BaseConnector {
 		const hermesHome = this.getHermesHome();
 		const hermesRepo = this.getHermesRepo();
 
-		// 1. Install the Python plugin into plugins/memory/signet/
+		// 1. Install the Python plugin into the current user-plugin location.
```
PR-Reviewer-Ant, line 719: Installing the same plugin into `~/.hermes/plugins/signet` looks broken for standalone user-plugin installs. The bundled Python still imports its client as `plugins.memory.signet.client`; if Hermes loads the user copy as `plugins.signet`, that import falls into the ImportError path, leaves `SignetClient = None`, and the provider advertises schemas but cannot initialize or handle tool calls. This is especially bad when `HERMES_REPO` is unset: install succeeds and `isInstalled()` can return true even though the user-level provider cannot route tools.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `34a27838`

</details>

I found two install-state bugs in the Hermes connector hardening. Both can make Signet report a healthy or installed Hermes integration while Hermes is still using a stale or nonfunctional provider.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The stale-repo detection issue is directly visible in isInstalled(): it accepts any current plugin dir even though this PR says the repo provider can shadow the user provider. The user-plugin import issue is also visible from the new ~/.hermes/plugins/signet install target combined with the bundled plugin's existing plugins.memory.signet.client import, though exact Hermes user-plugin import semantics are external to this diff.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -65,28 +117,329 @@
...
+			error: missing.length > 0 ? `Missing tools: ${missing.join(", ")}` : null,
+		};
+	} catch (e) {
+		const msg = e instanceof Error ? e.message : String(e);
+		return { ok: false, toolNames: [], error: `Could not parse Hermes provider probe output: ${msg}` };
+	}
+}
+
+async function checkDaemon(daemonUrl: string): Promise<{ ok: boolean; detail: string }> {
+	const baseUrl = daemonUrl.replace(/\/+$/, "");
```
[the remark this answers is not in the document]
NicholaiVogel: Thanks, fixed. I replaced the trailing-slash regex with a small `trimTrailingSlashes()` char-code loop, so daemon URL normalization no longer uses the flagged regex path.

integrations/hermes-agent/connector/src/index.ts
NicholaiVogel: Good catch. Fixed `isInstalled()` so when a Hermes repo is discoverable, the repo provider copy must be current. The user plugin is only accepted as the installed signal when no repo checkout is present.

I also added a regression where the repo marker is stale but the user marker is fresh, and `isInstalled()` now returns false.

integrations/hermes-agent/connector/src/index.ts
NicholaiVogel: Yep, Hermes loads user-installed memory providers as `_hermes_user_memory.<name>`, so the absolute bundled import can leave `SignetClient` unset. I changed the plugin to import `.client` first with the bundled absolute path only as fallback.

I added a user-namespace import regression and verified against the actual local Hermes install by temporarily hiding the repo provider so Hermes loaded `$HERMES_HOME/plugins/signet`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f78ceeb0`

</details>

I found one lifecycle bug in the Hermes config handling: install overwrites an existing memory provider, but uninstall cannot restore it. That contradicts the PR's backup/restore claim and can leave a user's Hermes memory setup disabled after uninstall.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed connector code directly overwrites non-Signet memory.provider values during install and later clears only Signet values during uninstall. The PR description explicitly claims config backup/restore, but the diff shows no persisted prior provider or backup path used for restore.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `f5ca22bc`.

NicholaiVogel:
Thanks, fixed in `e68f15c2`.

Install now records the previous non-Signet Hermes memory provider in `signet.provider.backup.json` before replacing it, and uninstall restores that provider when Signet is still active instead of just clearing the field. I covered both nested `memory:\n  provider: ...` and dotted `memory.provider: ...` config styles, and kept the no-backup behavior for fresh Signet installs so uninstall still clears only its own activation.

Validation:
- `bun test integrations/hermes-agent/connector/index.test.ts` (38 tests)
- `cd integrations/hermes-agent/connector && bun run typecheck`
- `cd integrations/hermes-agent/connector && bun run build`
- `bunx biome check integrations/hermes-agent/connector/src/index.ts integrations/hermes-agent/connector/index.test.ts`
- `bun surfaces/cli/src/cli.ts doctor hermes --json` against the actual local Hermes install, still `ok: true`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e68f15c2`

</details>

I found one config-patching edge case that can leave Hermes with contradictory memory provider settings instead of reliably activating Signet.

**Confidence:** Medium [sufficient_diff_evidence, ambiguous_requirements] - The issue follows directly from configureProvider's branch order in integrations/hermes-agent/connector/src/index.ts: it only updates a dotted memory.provider when the nested memory block is missing or unsafe. If both forms exist, the install path can add nested provider: signet while leaving memory.provider: honcho in place. I do not have Hermes' exact precedence rules here, so the runtime impact depends on how Hermes resolves duplicate dotted-vs-nested config, but the PR explicitly claims dotted memory.provider handling and reliable activation.


- integrations/hermes-agent/connector/src/index.ts:713 - This can leave a conflicting dotted provider in place. When config has both a nested `memory:` block without `provider` and an existing `memory.provider: honcho`, this branch inserts `memory:\n  provider: signet` but never updates or removes the dotted `memory.provider`. Depending on Hermes' config precedence, install may still use the old provider, and even if Hermes happens to prefer nested today the config is now ambiguous. Since this PR specifically hardens dotted `memory.provider` handling, please make the mixed-form case deterministic and add a regression test for it.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Fixed in `cb1a2cb2`.

The connector now treats `memory.provider` as the active provider declaration when it exists, and install updates that dotted key to `signet` instead of adding a nested provider. If a nested `memory.provider` line is also present inside a `memory:` block, install removes that nested provider line so the resulting config has only one active provider declaration.

I also tightened `isInstalled()`/doctor config detection so a mixed config with `memory.provider: honcho` and nested `provider: signet` is not reported as healthy.

Added regressions for:
- mixed nested block plus dotted `memory.provider: honcho`
- dotted `memory.provider: signet` plus stale nested provider
- conflicting dotted provider causing `isInstalled()` to return false

Validation is green locally: 41 connector tests, connector typecheck/build, scoped Biome check, `git diff --check`, and live `signet doctor hermes --json` against the local Hermes install.

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -65,28 +130,433 @@ function uninstallPlugin(hermesRepo: string): string[] {
...
+	const starts: number[] = [];
+	for (let i = 0; i < lines.length; i++) {
+		if (/^memory:\s*(?:#.*)?$/.test(lines[i] ?? "")) starts.push(i);
+		if (/^memory:\s*\S/.test(lines[i] ?? "")) return null;
+	}
+	if (starts.length === 0) return "missing";
+	if (starts.length !== 1) return null;
+	const start = starts[0] ?? 0;
+	let end = lines.length;
+	for (let i = start + 1; i < lines.length; i++) {
```
PR-Reviewer-Ant, line 197: `findMemoryBlock()` treats any indented `provider:` line inside `memory:` as the external memory provider, regardless of nesting depth. That means a valid config like `memory:\n  some_feature:\n    provider: foo` would be patched to `provider: signet`, and uninstall could later clear it. Since this PR now auto-writes the user's Hermes config, this needs to distinguish the direct `memory.provider` child from nested provider keys before mutating the file.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cb1a2cb2`

</details>

I found one config-patching bug that can rewrite the wrong YAML key under `memory:`. The rest of the Hermes install/doctor work lines up with the PR description from the diff I reviewed.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The issue is directly visible in the new line-based YAML patcher in integrations/hermes-agent/connector/src/index.ts. I did not run the full test suite, but the bad match is local to the changed config parsing logic.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

integrations/hermes-agent/connector/src/index.ts
NicholaiVogel: Thanks, fixed in `f6be3c48`.

`findMemoryBlock()` now tracks the direct child indentation under the top-level `memory:` block and only treats a `provider:` key at that exact indentation as the Hermes memory provider. Deeper keys like `memory:\n  some_feature:\n    provider: foo` are now ignored by provider detection and left untouched by install/uninstall.

Added regressions for:
- `isInstalled()` ignoring nested provider keys
- install adding a direct `memory.provider` without rewriting nested provider keys
- uninstall not clearing nested provider keys

Validation is green locally: 44 connector tests, connector typecheck/build, scoped Biome check, `git diff --check`, and live `signet doctor hermes --json` against the local Hermes install.

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -65,28 +130,449 @@ function uninstallPlugin(hermesRepo: string): string[] {
...
+		);
+		return { configPath: null, backupPath: null };
+	}
+
+	if (block !== null && typeof block === "object" && block.provider !== null && block.provider !== undefined) {
+		if (providerLineIsSignet(lines[block.provider] ?? "")) return { configPath: null, backupPath: null };
+		backupPath = writeProviderBackup(
+			hermesHome,
+			configPath,
+			"nested",
```
PR-Reviewer-Ant, line 396: This removes an existing nested `memory.provider` when dotted `memory.provider: signet` is already active, but it does not record a backup in that branch. A real config like `memory:\n  provider: honcho\nmemory.provider: signet\n` will have `honcho` deleted during install; on uninstall, `restoreOrClearProvider()` has no nested backup and only clears the dotted provider to `''`, so the previous provider is lost. That undercuts the PR's backup/restore claim and can corrupt a user's Hermes memory-provider config.

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -349,8 +860,113 @@ export class HermesAgentConnector extends BaseConnector {
...
+	});
+	checks.push({
+		id: "user-plugin",
+		label: "User plugin copy",
+		ok: userPluginCurrent,
+		detail: userPluginCurrent
+			? `${userPluginDir} matches bundled Signet plugin`
+			: `${userPluginDir} is missing or stale`,
+		fix: userPluginCurrent ? undefined : "Run `signet setup --harness hermes-agent`.",
+	});
```
PR-Reviewer-Ant, line 982: The doctor marks the repo-plugin check as failed whenever `HERMES_REPO` is absent, even though this PR intentionally supports installing only the user plugin under `$HERMES_HOME/plugins/signet` and `isInstalled()` returns true in that mode. The same issue affects tool routing below because the probe is forced to fail without a repo. That means a supported user-plugin-only install can never get an `ok: true` Hermes doctor report, so the new diagnostic command will send users into a repair loop even after a successful install.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `f6be3c48`

</details>

I found two correctness problems in the new Hermes config/doctor paths. One can lose a user's previous nested provider during install/uninstall, and the other makes `signet doctor hermes` report failure for the same user-plugin-only install mode this PR now supports.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The config mutation and diagnostic behavior are directly visible in integrations/hermes-agent/connector/src/index.ts, and the PR description explicitly claims backup/restore support plus user plugin copy support.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

integrations/hermes-agent/connector/src/index.ts
NicholaiVogel: Thanks, fixed in `f5ca22bc`.

When a dotted `memory.provider: signet` is already active and install removes a conflicting direct nested provider, we now persist that nested provider in `signet.provider.backup.json`. Uninstall clears the dotted Signet activation and restores the removed nested provider under `memory:`.

I also guarded the mixed non-Signet case so the dotted provider remains the backup source when both forms need cleanup.

Added regressions for both restore paths, including this shape:

```yaml
memory:
  provider: honcho
memory.provider: signet
```

integrations/hermes-agent/connector/src/index.ts
NicholaiVogel: Thanks, fixed in `f5ca22bc`.

`signet doctor hermes` now treats the repo plugin as optional when no Hermes checkout is available and the user plugin copy is current. In that mode the doctor reports the user plugin schemas as the available Signet tools instead of failing the tool-routing check and sending the user back into setup.

I also fixed `diagnoseHermesIntegration({ hermesRepo: null })` so explicit no-repo diagnostics do not fall back to repo auto-discovery.

Added a regression for a current user-plugin-only install returning `ok: true`, and verified that path against `/home/nicholai/.hermes` with `hermesRepo: null`.

integrations/hermes-agent/connector/src/index.ts
```diff
@@ -199,22 +703,26 @@ export class HermesAgentConnector extends BaseConnector {
 		const hermesHome = this.getHermesHome();
 		const hermesRepo = this.getHermesRepo();
 
-		// 1. Install the Python plugin into plugins/memory/signet/
+		// 1. Install the Python plugin into the current user-plugin location.
+		try {
+			const pluginFiles = installPlugin(getUserPluginTargetDir(hermesHome), "user");
+			filesWritten.push(...pluginFiles);
+		} catch (e) {
```
PR-Reviewer-Ant, line 724: This can still return `success: true` after failing to install the only plugin copy. In the user-plugin-only path, `installPlugin(getUserPluginTargetDir(...))` is required for Hermes to load Signet, but the catch just appends a warning and the method later returns the normal success message. If `$HERMES_HOME` is unwritable or otherwise fails, setup will tell the user the provider was deployed and activated even though no plugin exists. Please track whether at least one usable plugin target was written, and return a failed `InstallResult` when both the user copy and any repo copy fail.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f5ca22bc`

</details>

I found one install-path correctness issue: the connector can report a successful Hermes install even when it failed to copy the only usable plugin target.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The changed install flow catches user-plugin copy failures, records only a warning, and still returns success even when no Hermes repo is available. That directly conflicts with the PR's claimed hardened install path; I did not run a filesystem permission repro.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

integrations/hermes-agent/connector/src/index.ts
NicholaiVogel: Thanks, fixed in `69b2abb2`.

Install now tracks whether the plugin copy Hermes can actually load was written. If no Hermes repo is available, the user plugin copy is required; if a Hermes repo is discoverable, the repo plugin copy is required because it can shadow the user plugin.

When the required target cannot be written, install returns `success: false` before patching Hermes config/env, so setup no longer claims the provider was deployed and activated.

Added regressions for:
- user-plugin-only install failing when `$HERMES_HOME/plugins` is not writable as a directory
- repo-plugin refresh failure returning `success: false` even if the user copy was written
- repo plugin success still allowing install to proceed when the user copy fails

Validation is green locally: 50 connector tests, connector typecheck/build, scoped Biome check, `git diff --check`, and live `signet doctor hermes --json` against `/home/nicholai/.hermes`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `69b2abb2`

</details>

I found one correctness issue in the Hermes config patcher: it can rewrite valid non-mapping `memory:` YAML into an invalid mixed structure while reporting install success.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue follows directly from the new line-oriented YAML patcher in integrations/hermes-agent/connector/src/index.ts: findMemoryBlock accepts any block-style `memory:` section, and configureProvider inserts `provider: signet` into it when no direct provider is found. I did not run the repo tests in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `69b2abb2`

</details>

I found one correctness issue in the Hermes config patcher: it can rewrite valid non-mapping `memory:` YAML into an invalid mixed structure while reporting install success.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue follows directly from the new line-oriented YAML patcher in integrations/hermes-agent/connector/src/index.ts: findMemoryBlock accepts any block-style `memory:` section, and configureProvider inserts `provider: signet` into it when no direct provider is found. I did not run the repo tests in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Thanks, fixed in `366692d8`.

`findMemoryBlock()` now rejects a block-style `memory:` section unless its direct children are YAML mapping entries. That means map-shaped config like:

```yaml
memory:
  some_feature:
    provider: honcho
```

is still patchable, but non-mapping shapes like:

```yaml
memory:
  - honcho
  - archival
```

now take the existing safe-warning path and are left untouched instead of getting a mixed `provider: signet` entry inserted.

Added a regression for the sequence case. Validation is green locally: 51 connector tests, connector typecheck/build, scoped Biome check, `git diff --check`, and live `signet doctor hermes --json` against `/home/nicholai/.hermes`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `366692d8`

</details>

I found one install-flow correctness issue: the connector can report a successful, activated Hermes install even when it explicitly failed to activate the provider in config.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The failure path is directly visible in configureProvider and install: unsafe or non-mapping memory config only appends a warning and returns no patched config, while install still returns success with an activated-provider message. The added tests cover the warning behavior but do not assert isInstalled() or doctor health after this path.

PR-Reviewer-Ant:
<details>

- Model: `gpt-5.5`
- Commit: `366692d8`

</details>

I found one install-flow correctness issue: the connector can report a successful, activated Hermes install even when it explicitly failed to activate the provider in config.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The failure path is directly visible in configureProvider and install: unsafe or non-mapping memory config only appends a warning and returns no patched config, while install still returns success with an activated-provider message. The added tests cover the warning behavior but do not assert isInstalled() or doctor health after this path.

NicholaiVogel:
Thanks, fixed in `210f2f08`.

Install now verifies the post-patch Hermes provider state before returning success. If the connector refuses to patch unsafe or non-mapping memory config, it returns `success: false` with an incomplete/not-activated message instead of the old deployed-and-activated success path.

I tightened the regressions for both unsafe YAML and non-mapping `memory:` blocks so they now assert failed install state, `isInstalled() === false`, and the doctor `provider-config` check staying false for the non-mapping case.

Validation:
- `bun test integrations/hermes-agent/connector/index.test.ts` (51 tests, 196 expects)
- `cd integrations/hermes-agent/connector && bun run typecheck`
- `cd integrations/hermes-agent/connector && bun run build`
- `bunx biome check integrations/hermes-agent/connector/src/index.ts integrations/hermes-agent/connector/index.test.ts`
- `git diff --check`
- `bun surfaces/cli/src/cli.ts doctor hermes --json` against `/home/nicholai/.hermes`, still `ok: true`

PR #632, opened 2026-05-06
feat: inherit parent context for sub-agent sessions

### LINKED ISSUE #315: Feature request: forward parent session's Signet memory context to sub-agents at spawn time
# Feature Request: Signet Memory Context Injection for Sub-agents

## Summary

When a parent agent session spawns a sub-agent, the sub-agent currently starts with zero Signet memory context. It inherits the workspace directory but does not receive Signet's memory injection — no relevant memories, no entity graph context, no anticipatory recall. This means every sub-agent starts from scratch on every task, even when the parent session has rich accumulated context directly relevant to what the sub-agent is doing.

## Current Behavior

- Parent session: receives Signet memory injection on every prompt (entities, relevant memories, anticipatory recall)
- Sub-agent session: starts cold — no Signet context, no memory injection
- Parent must manually summarize relevant context in the task description to compensate

## Desired Behavior

An option to forward the parent session's Signet memory context to sub-agents at spawn time, so sub-agents can benefit from the same accumulated knowledge without requiring the parent to re-describe everything.

## Use Cases

**1. Domain-aware sub-agents**
When a parent agent spawns a sub-agent to audit Juniper switch configs, the sub-agent should know that this user runs an EX4300 and EX2200, what VLANs are in use, what issues have been flagged before — without the parent having to paste all of that into the task description.

**2. Consistent voice/persona**
When a sub-agent drafts an email or document, it should know the user's writing style preferences, communication history, and context — the same way the parent agent does.

**3. Codebase-aware sub-agents**
When a sub-agent is asked to work on a specific project, it should have the same project context that's been accumulated in Signet (architecture decisions, past bugs, conventions) rather than starting blind.

**4. Multi-step workflows**
In long-running workflows where the parent spawns multiple sub-agents sequentially, each sub-agent currently needs the parent to re-inject context. Signet context sharing would make these workflows significantly more capable.

### Option A: Full context forwarding (opt-in)
At spawn time, pass a flag like `signet_context: "inherit"` that causes the sub-agent's first prompt to include the same Signet memory injection the parent received. This is a one-time snapshot — the sub-agent doesn't get ongoing memory updates, just the parent's current context.

### Option B: Scoped context forwarding
The parent specifies which memory/entity categories to forward:
```
signet_context:
  memories: true
  entities: ["Juniper", "network", "homelab"]
  top_k: 20
```

### Option C: Shared session scope
Sub-agents spawned with `signet_scope: "parent"` share the parent's Signet session scope, receiving ongoing context injection as if they were the same session. More powerful but potentially more complex to implement.

**Option A is the simplest starting point** and would cover the majority of use cases.

## Configuration Sketch

In `agent.yaml` or spawn parameters:

```yaml
memory:
  pipelineV2:
    subagents:
      inheritContext: true      # forward parent context to sub-agents
      inheritTopK: 15           # how many memories to forward
      inheritEntities: true     # include entity graph context
```

Or as a per-spawn flag in the harness:

```javascript
sessions_spawn({
  task: "...",
  signetContext: "inherit"   // or "none" (default)
})
```

## Expected Impact

- Significantly reduces the burden on parent agents to manually re-describe context in task descriptions
- Makes sub-agents first-class citizens in the Signet memory system
- Enables more capable multi-agent workflows without context loss at agent boundaries
- Particularly valuable for long-running tasks where the parent has accumulated substantial session context

## Related

- Sub-agents currently inherit the workspace directory (`~/.agents/`) but not the memory injection
- This is consistent with how Signet handles new sessions generally — but sub-agents are a different case since they're explicitly spawned by a context-rich parent, not a fresh user session

---

*Reported by: Adam Paulsen via Sam Bridger (OpenClaw agent)*

NicholaiVogel:
## Summary
- promote the sub-agent context continuity spec to approved and document the API, MCP, hook, harness, and config surfaces
- add daemon parent-session inference for explicit lineage, OpenClaw session keys, and Claude Code sub-agent hook payloads
- wire Claude Code, OpenCode, and CLI hook payloads so harness sub-agent IDs do not become Signet agent scopes

## PR Readiness Checklist

## Validation
- `./node_modules/.bin/biome check docs/API.md docs/CONFIGURATION.md docs/HARNESSES.md docs/HOOKS.md docs/MCP.md docs/specs/INDEX.md docs/specs/dependencies.yaml docs/specs/approved/sub-agent-context-continuity.md integrations/claude-code/connector/src/index.ts integrations/opencode/plugin/src/index.ts platform/core/src/types.ts platform/daemon/src/hooks.test.ts platform/daemon/src/hooks.ts platform/daemon/src/mcp/tools.test.ts platform/daemon/src/mcp/tools.ts platform/daemon/src/memory-config.ts platform/daemon/src/routes/session-routes.ts platform/daemon/src/subagent-context.ts surfaces/cli/src/commands/hook.test.ts surfaces/cli/src/commands/hook.ts`
- `git diff --check HEAD~1..HEAD`
- `bun run build:core`
- `bun test platform/daemon/src/hooks.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/hook.test.ts`
- `cd integrations/opencode/plugin && bun run typecheck && bun run build`
- `cd integrations/claude-code/connector && bun run typecheck && bun run build`
- `cd platform/daemon && bun run build.ts`
- `cd surfaces/cli && bun run build:cli`

## Notes
- Full `bun run lint` still reports pre-existing package.json formatting issues outside this patch; scoped Biome is clean for the touched files.
- No admin endpoint was added; the new session transcript endpoints use existing recall permission middleware.

Closes #315


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `dd3dac84`.

platform/daemon/src/subagent-context.ts
```diff
@@ -0,0 +1,370 @@
...
+		`SELECT st.session_key, st.project, ${seen} AS updated_at, st.content, ${score} AS rank`,
+		"FROM session_transcripts st",
+		"WHERE st.agent_id = ?",
+	];
+	const args: unknown[] = words.map((word) => `%${word}%`);
+	args.push(params.agentId);
+	if (targetSessionKey) {
+		parts.push("AND st.session_key = ?");
+		args.push(targetSessionKey);
+	} else if (params.currentSessionKey) {
```
PR-Reviewer-Ant, line 343: The fallback LIKE query binds parameters in the wrong order. The generated SQL starts with `${score} AS rank`, which contains the first N `?` placeholders, then `WHERE st.agent_id = ?`; however `args` is initialized with the LIKE patterns and only then pushes `params.agentId`. That part is fine for the score placeholders, but after appending `AND (${any})`, the code pushes the second set of LIKE patterns at the very end, after optional `session_key`, `project`, and `limit` placeholders have already been added. With a session/project filter, those filters receive LIKE values and the final LIKE clauses receive session/project/limit values, so searches can silently miss valid transcript hits. Build the SQL and args in exact placeholder order, or split score/where args so optional filters don't get interleaved incorrectly.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c15f71cc`

</details>

I found one correctness issue in the fallback transcript search path: the SQL parameter order does not match the generated placeholders, so non-FTS searches can fail or return no results even when matching transcripts exist.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The bug is directly visible in the new searchSessionTranscripts fallback query: placeholders in SELECT come before the WHERE agent_id placeholder, but args pushes LIKE parameters first and agentId afterward. This affects the new /api/sessions/search and session_search MCP behavior added by the PR.

integrations/opencode/plugin/src/index.ts
```diff
@@ -205,6 +244,7 @@ export const SignetPlugin: Plugin = async ({ directory, client: oc }) => {
 			pendingInject.delete(input.sessionID);
```
PR-Reviewer-Ant, line 260: This posts the per-session `session-start` request after the hook has already read and deleted `pendingInject` for `input.sessionID`, so any inherited parent context returned by `ensureSessionStarted()` is queued for a later prompt instead of the child session's first prompt. For sub-agent continuity, the first prompt is the important one: the sub-agent starts working before the inherited context is visible. The fix should make `ensureSessionStarted(input.sessionID)` run before consuming `pendingInject`, or have it return the inject text directly so this same hook invocation can append it.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `02788fec`

</details>

I found one OpenCode lifecycle bug: inherited session-start context is queued too late to be injected into the child session's first prompt, which undercuts the PR's main continuity claim for that harness.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The OpenCode hook path reads and deletes pending injection before calling ensureSessionStarted, while ensureSessionStarted writes the inherited context back into pendingInject. That ordering is visible in integrations/opencode/plugin/src/index.ts, though I did not run OpenCode to confirm the exact hook timing.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

integrations/opencode/plugin/src/index.ts
PR-Reviewer-Ant, line 260: `ensureSessionStarted()` can enqueue the per-session `session-start` injection, including inherited parent context, but this runs after `prompt.before` has already read and deleted `pendingInject` for this `sessionID`. That means a newly created OpenCode child session will not receive the inherited parent context on its first prompt; it only becomes pending for a later prompt. Since this PR’s stated behavior is sub-agent context continuity at session start, the session-start call needs to happen before consuming pending injections, or its returned `inject` needs to be appended directly to the current `output.context` path.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6db56882`

</details>

I found one correctness issue in the OpenCode lifecycle path: the newly fetched session-start context for a per-session start is queued after the prompt handler has already consumed pending injections, so the first prompt of a child session misses the inherited parent context.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff shows `prompt.before` deleting `pendingInject` for `input.sessionID` before calling `ensureSessionStarted`, while `ensureSessionStarted` queues the session-start `result.inject` back into `pendingInject`. That makes the inherited context unavailable to the prompt currently being handled. I could not inspect the full local repository in this read-only review environment, so the confidence is medium rather than high, but the ordering problem is visible in the changed OpenCode plugin code.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/subagent-context.ts
NicholaiVogel: Fixed in `5c6ed3f1`.

I split the fallback query bindings into `scoreArgs`, `whereArgs`, the second LIKE pattern set, and `limit` so the argument list now mirrors the SQL placeholder order explicitly. I also fixed `tableExists()` to treat Bun SQLite's no-row `null` result as false, which keeps minimal/no-FTS schemas on the LIKE fallback path instead of trying to prepare against a missing FTS table.

Added regression coverage in `platform/daemon/src/subagent-context.test.ts` for fallback LIKE search with both `currentSessionKey` and `project` filters.

integrations/opencode/plugin/src/index.ts
NicholaiVogel: Fixed in `5c6ed3f1`.

`ensureSessionStarted()` now returns the per-session start inject instead of only enqueueing it as a side effect. The system transform calls it before reading `pendingInject`, so OpenCode sessions get inherited parent context even if `experimental.chat.system.transform` runs before `chat.message`. The `chat.message` path still queues that same start inject when chat runs first, so both observed hook orderings keep the context available for the current prompt.

Added OpenCode plugin regressions for both orders: transform-before-chat and chat-before-transform.

NicholaiVogel: Fixed in `5c6ed3f1`.

`ensureSessionStarted()` now returns the per-session start inject instead of only enqueueing it as a side effect. The system transform calls it before reading `pendingInject`, so OpenCode sessions get inherited parent context even if `experimental.chat.system.transform` runs before `chat.message`. The `chat.message` path still queues that same start inject when chat runs first, so both observed hook orderings keep the context available for the current prompt.

Added OpenCode plugin regressions for both orders: transform-before-chat and chat-before-transform.

surfaces/cli/src/commands/hook.ts
```diff
@@ -432,6 +423,40 @@ export function pickSessionKey(input: Record<string, unknown> | null): string {
...
+): {
+	harness: string;
+	project: string;
+	agentId?: string;
+	harnessAgentId?: string;
+	parentSessionKey?: string;
+	context?: string;
+	sessionKey: string;
+	runtimePath: typeof LEGACY_RUNTIME_PATH;
+} {
```
PR-Reviewer-Ant, line 458: This treats stdin `agentId` as a harness-native sub-agent id, which breaks existing CLI hook callers using the documented `agentId` field for Signet scope. The new split is right in principle, but it needs to preserve `agentId`/`agent_id` compatibility for Signet scope where those fields were already part of the CLI/daemon contract, or gate the Claude-native `agent_id` interpretation to the Claude Code payload shape. As written, a caller that previously scoped session-start to `agentId: "research-agent"` will now send no Signet `agentId` and will inject/search/write under the default scope.

surfaces/cli/src/commands/hook.ts
```diff
@@ -444,13 +469,17 @@ export function buildUserPromptSubmitBody(
...
+	agentId?: string;
+	harnessAgentId?: string;
 	lastAssistantMessage?: string;
 	runtimePath: typeof LEGACY_RUNTIME_PATH;
 } {
 	const body = input;
 	const userPrompt = pickString(body?.prompt, body?.user_prompt, body?.userPrompt);
 	const userMessage = pickString(body?.user_message, body?.userMessage, userPrompt);
 	const lastAssistantMessage = readLastAssistantMessage(body);
+	const agentId = pickString(body?.signet_agent_id, body?.signetAgentId);
```
PR-Reviewer-Ant: Same scoping regression on prompt-submit: stdin `agentId` is now forwarded as `harnessAgentId`, while Signet scope is accepted only from `signet_agent_id`/`signetAgentId`. Existing hook JSON that sends `agentId` will start upserting transcripts and doing prompt recall under the default agent, which can split session continuity and memory lookup across scopes.

surfaces/cli/src/commands/hook.ts
```diff
@@ -482,7 +513,7 @@ export function buildCompactionCompleteBody(
 	runtimePath: typeof LEGACY_RUNTIME_PATH;
 } {
 	const body = input;
-	const agentId = pickString(overrides.agentId, body?.agent_id, body?.agentId);
+	const agentId = pickString(overrides.agentId, body?.signet_agent_id, body?.signetAgentId);
```
PR-Reviewer-Ant: Compaction has the same backwards-incompatible shape change: stdin `agentId`/`agent_id` is no longer honored for Signet scope. That means existing compaction-complete hook clients can write summaries under `default` instead of the intended agent, breaking scoped temporal lineage.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5c6ed3f1`

</details>

I found one compatibility bug in the CLI hook request builders: existing JSON callers that send `agentId` as the documented Signet scope now get silently remapped or dropped. That undercuts the PR's own API contract that `agentId` remains the persistence scope while harness-native IDs should use `harnessAgentId`.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed CLI builder code directly maps stdin `agentId` to `harnessAgentId` in session-start and prompt-submit bodies, and compaction now ignores stdin `agentId` entirely. The PR documentation still says `agentId` is the Signet persistence scope, so this is a concrete request-shape regression.


**Since my last review:**
- Still blocking: 0
- Newly found: 3

surfaces/cli/src/commands/hook.ts
NicholaiVogel: Fixed in `94485c46`.

The CLI hook builders now preserve `agentId` and legacy `agent_id` as the Signet persistence scope for session-start, prompt-submit, and compaction-complete payloads. Harness-native sub-agent IDs are accepted through explicit `harnessAgentId` / `harness_agent_id`; Claude-style native `agent_id` is only treated as `harnessAgentId` on session-start when the payload also carries a parent-session reference, which is the sub-agent spawn shape this PR needs.

Added CLI regressions covering existing `agentId` and `agent_id` JSON callers across the three affected builders.

surfaces/cli/src/commands/hook.ts
NicholaiVogel: Fixed in `94485c46`.

The CLI hook builders now preserve `agentId` and legacy `agent_id` as the Signet persistence scope for session-start, prompt-submit, and compaction-complete payloads. Harness-native sub-agent IDs are accepted through explicit `harnessAgentId` / `harness_agent_id`; Claude-style native `agent_id` is only treated as `harnessAgentId` on session-start when the payload also carries a parent-session reference, which is the sub-agent spawn shape this PR needs.

Added CLI regressions covering existing `agentId` and `agent_id` JSON callers across the three affected builders.

surfaces/cli/src/commands/hook.ts
NicholaiVogel: Fixed in `94485c46`.

The CLI hook builders now preserve `agentId` and legacy `agent_id` as the Signet persistence scope for session-start, prompt-submit, and compaction-complete payloads. Harness-native sub-agent IDs are accepted through explicit `harnessAgentId` / `harness_agent_id`; Claude-style native `agent_id` is only treated as `harnessAgentId` on session-start when the payload also carries a parent-session reference, which is the sub-agent spawn shape this PR needs.

Added CLI regressions covering existing `agentId` and `agent_id` JSON callers across the three affected builders.

surfaces/cli/src/commands/hook.ts
```diff
@@ -444,13 +491,17 @@ export function buildUserPromptSubmitBody(
...
+	agentId?: string;
+	harnessAgentId?: string;
 	lastAssistantMessage?: string;
 	runtimePath: typeof LEGACY_RUNTIME_PATH;
 } {
 	const body = input;
 	const userPrompt = pickString(body?.prompt, body?.user_prompt, body?.userPrompt);
 	const userMessage = pickString(body?.user_message, body?.userMessage, userPrompt);
 	const lastAssistantMessage = readLastAssistantMessage(body);
+	const agentId = pickString(body?.signet_agent_id, body?.signetAgentId, body?.agentId, body?.agent_id);
```
PR-Reviewer-Ant: This still treats a hook payload's `agent_id` as the Signet persistence scope on `user-prompt-submit`. That contradicts the PR's stated fix for Claude Code sub-agents: Claude's native sub-agent `agent_id` can still be sent to the daemon as `agentId`, so per-prompt recall/transcript updates can fork into the harness sub-agent scope instead of the user's Signet scope. `buildSessionStartBody` already has the safer split between `agentId` and `harnessAgentId`; prompt-submit needs the same distinction, or at minimum Claude Code `agent_id` should become `harnessAgentId` unless an explicit `signet_agent_id` / `signetAgentId` / CLI `--agent-id` is provided.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `94485c46`

</details>

I found one scope-handling bug that undercuts the PR's main claim for Claude Code prompt-submit hooks. Session-start separates harness sub-agent IDs from Signet agent scope, but prompt-submit still maps `agent_id` into `agentId`.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR explicitly says Claude Code hook payloads should keep harness sub-agent IDs from becoming Signet agent scopes. The changed CLI session-start path handles that split, but the changed prompt-submit builder still forwards `body.agent_id` as `agentId`, which is directly visible in surfaces/cli/src/commands/hook.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

surfaces/cli/src/commands/hook.ts
NicholaiVogel: Fixed in `a360b460`.

`buildUserPromptSubmitBody()` now mirrors the scoped split for Claude Code prompt-submit payloads: raw Claude `agent_id` is kept out of Signet `agentId` and forwarded only as `harnessAgentId`. Existing non-Claude legacy callers can still use `agent_id` as Signet scope, while explicit `signet_agent_id`, `signetAgentId`, or `agentId` remain the Signet scope for Claude Code.

Added a focused regression asserting Claude Code prompt-submit `agent_id` does not become `agentId`.

surfaces/cli/src/commands/hook.ts
```diff
@@ -427,11 +418,67 @@ function pickString(...values: unknown[]): string {
...
+		readonly harness: string;
+		readonly project?: string;
+		readonly agentId?: string;
+		readonly context?: string;
+	},
+): {
+	harness: string;
+	project: string;
+	agentId?: string;
+	harnessAgentId?: string;
```
PR-Reviewer-Ant, line 453: This still lets Claude Code sub-agent `agent_id` become the Signet `agentId` whenever the hook payload does not also include a parent reference. The PR's own spec says Claude Code detection is `agent_id` present on `SessionStart` with parent inferred by recent project activity, and `handleSessionStart` only runs the Claude Code parent lookup when `harnessAgentId` is set. With a payload like `{ agent_id: "general-purpose", session_id: "child", cwd: "/repo" }`, this builder sends `agentId: "general-purpose"` and no `harnessAgentId`, so the child forks memory scope and parent context inference never runs. For Claude Code session-start, `agent_id` needs to be treated as harness metadata even without explicit parent fields, unless an explicit Signet scope such as `--agent-id`, `signet_agent_id`, or `signetAgentId` is present.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a360b460`

</details>

I found one correctness issue in the Claude Code hook mapping: the new session-start builder still treats Claude Code's native `agent_id` as a Signet persistence scope in the exact sub-agent case this PR is meant to fix.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description says Claude Code harness sub-agent IDs must not become Signet agent scopes, and the daemon's parent inference requires `harnessAgentId`. In `buildSessionStartBody`, `agent_id` is only moved to `harnessAgentId` when a parent reference is present, but the approved spec and daemon logic describe Claude Code inference from `agent_id` alone plus project recency.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

surfaces/cli/src/commands/hook.ts
NicholaiVogel: Fixed in `79940c2f`.

`buildSessionStartBody()` now treats Claude Code raw `agent_id` as harness-native even when the payload does not include an explicit parent reference, so it becomes `harnessAgentId` and never Signet `agentId` for Claude session-start. Non-Claude legacy callers can still use `agent_id` as the Signet scope, and explicit `agentId` / `signet_agent_id` / `signetAgentId` remain the Signet scope.

Added a regression for Claude Code session-start with only `agent_id`, `sessionKey`, and `cwd`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `79940c2f`

</details>

I did not find a concrete correctness, security, scoping, or data-integrity issue in the changed implementation. The code matches the PR's stated direction: harness-native sub-agent IDs are kept out of Signet scope, parent lookup is agent-scoped, and the new transcript search surface is permission-gated.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The supplied diff includes the main daemon, CLI, MCP, OpenCode, Claude Code connector, docs, and focused tests for the scope and lineage paths. I did not run the branch or exercise the hooks against real harness events, so concurrency/order assumptions in the live OpenCode and Claude Code runtimes were not independently reproduced.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

integrations/opencode/plugin/src/index.ts
```diff
@@ -230,22 +271,27 @@ export const SignetPlugin: Plugin = async ({ directory, client: oc }) => {
 		// ------------------------------------------------------------------
 		"experimental.chat.system.transform": async (input, output): Promise<void> => {
```
PR-Reviewer-Ant, line 294: This new `await ensureSessionStarted(input.sessionID)` sits outside a `try/catch` in `experimental.chat.system.transform`. If the daemon is down, slow, or returns a bad response, the exception escapes the OpenCode hook instead of degrading the Signet injection. That is a behavior regression from the fail-open pattern used elsewhere in this plugin, and it can break the harness path just because Signet context was unavailable. Wrap this call the same way `chat.message` and compaction do, and treat session-start context as optional on failure.

integrations/opencode/plugin/src/index.ts
```diff
@@ -288,19 +340,24 @@ export const SignetPlugin: Plugin = async ({ directory, client: oc }) => {
...
-							runtimePath: RUNTIME_PATH,
-							reason: event.type,
-							sessionKey: sid,
-							...(transcript ? { transcript } : {}),
-						},
-						WRITE_TIMEOUT,
-					)
+							"/api/hooks/session-end",
+							{
+								harness: HARNESS,
```
PR-Reviewer-Ant, line 373: The per-session `session-start` call now sends `agentId`, but this `session-end` write still does not. For a configured non-default Signet agent, the final transcript snapshot will be persisted under whatever scope the daemon infers by default, while the child session was started under the configured scope. That undermines the PR's scoping claim and can make parent lookup/session search miss the terminal transcript for named agents. Include the same `agentId` in session-end that `ensureSessionStarted` sends.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `ddbabab7`

</details>

I found a couple of correctness issues in the OpenCode lifecycle path. The new per-session start call can make OpenCode fail closed on daemon/network errors, and named-agent session-end writes still appear to fall back to the default scope.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The OpenCode hook code added a direct await of ensureSessionStarted from system transform without a local catch, and the session-end payload shown in the changed block still omits agentId while session-start includes it. I did not run OpenCode itself, so the exact harness failure mode is inferred from the hook implementation.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

NicholaiVogel:
Addressed the latest OpenCode lifecycle feedback in `080c2d78`.

What changed:
- Wrapped the per-session `ensureSessionStarted(input.sessionID)` call in `experimental.chat.system.transform` so unavailable Signet session-start context stays fail-open and cannot break OpenCode prompt rendering.
- Included the configured Signet `agentId` on OpenCode `session-end` writes, matching the scope used by per-session `session-start`.
- Added OpenCode lifecycle regressions for fail-open transform behavior and named-agent `session-end` scope propagation.

Validation run from `/tmp/signet-pr-monitor/pr-632`:
- `git diff --check`
- `bun test integrations/opencode/plugin/src/index.test.ts`
- `./node_modules/.bin/biome check integrations/opencode/plugin/src/index.ts integrations/opencode/plugin/src/index.test.ts`
- `cd integrations/opencode/plugin && bun run typecheck && bun run build`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `080c2d78`

</details>

I did not find a concrete correctness, security, data integrity, or scope regression in the changed code. The implementation matches the PR description: harness-native sub-agent IDs are kept separate from Signet agent scope, parent lineage is threaded through the supported hook paths, and the new transcript search surface is recall-scoped.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the daemon inference logic, route authorization, MCP wiring, CLI/OpenCode payload construction, docs, and targeted tests for Claude Code scope separation, OpenClaw lineage, OpenCode parent propagation, and session_search request mapping. I did not run a live harness reproduction, so there remains some runtime risk around actual Claude Code/OpenCode event ordering and payload shapes.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Merged `origin/main` into the PR branch to satisfy the up-to-date requirement. The merge was clean with no manual conflict resolution.

Validation after the merge from `/tmp/signet-pr-monitor/pr-632`:
- `git diff --check origin/main...HEAD` — passed
- `bun scripts/spec-deps-check.ts` — passed
- `bun test platform/daemon/src/hooks.test.ts platform/daemon/src/mcp/tools.test.ts platform/daemon/src/subagent-context.test.ts surfaces/cli/src/commands/hook.test.ts integrations/opencode/plugin/src/index.test.ts` — 246 pass, 0 fail

Latest PR-Reviewer-Ant review on `080c2d78` reported no concrete issues; this merge commit only updates the branch with current `main`.

integrations/opencode/plugin/src/index.ts
```diff
@@ -205,6 +244,8 @@ export const SignetPlugin: Plugin = async ({ directory, client: oc }) => {
 			pendingInject.delete(input.sessionID);
 
 			try {
+				const startInject = await ensureSessionStarted(input.sessionID);
```
PR-Reviewer-Ant: `ensureSessionStarted()` now runs inside the same `try` as `/api/hooks/user-prompt-submit`, so a failure in the optional per-session start call prevents prompt-submit from running at all. That drops prompt-time recall and transcript persistence for the current user message even though the old path would still attempt `/user-prompt-submit`. This is especially easy to hit because the PR explicitly treats session-start context as optional in `experimental.chat.system.transform`; `chat.message` should isolate that failure the same way and continue with prompt-submit.

surfaces/cli/src/commands/hook.ts
```diff
@@ -504,19 +569,22 @@ export function buildSessionEndBody(
...
 	sessionKey: string;
+	agentId?: string;
 	cwd: string;
 	reason: string;
 	runtimePath: typeof LEGACY_RUNTIME_PATH;
 } {
 	const body = input ?? {};
 	const sessionKey = pickSessionKey(body);
 	const sessionId = pickString(body.session_id, body.sessionId, sessionKey);
+	const agentId = pickString(body.signet_agent_id, body.signetAgentId);
```
PR-Reviewer-Ant: `buildSessionEndBody()` now only forwards `signet_agent_id` / `signetAgentId` and ignores existing `agentId` / `agent_id` fields. The PR carefully preserves those legacy fields for session-start, prompt-submit, and compaction, but session-end now silently falls back to the default scope for callers that already send `agentId` or non-Claude `agent_id`. That can write terminal transcripts and summaries under the wrong agent scope.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `4d01e6f7`

</details>

I found two correctness regressions in the hook wiring: one can drop OpenCode prompt-submit work when per-session start context fails, and one stops forwarding existing session-end agent scope fields.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed hook builders and OpenCode plugin logic are included in the diff, and the regressions follow directly from the new control flow and field selection. I could not run the repository tests because the provided checkout only contains the review prompt, not the repo files.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

NicholaiVogel:
Fixed in `e5edadd5`.

What changed:
- Isolated the optional OpenCode per-session `session-start` call inside `chat.message`, so a daemon/session-start failure no longer prevents `/api/hooks/user-prompt-submit` from running for prompt recall and transcript capture.
- Restored session-end Signet scope compatibility by accepting existing `agentId` and non-Claude `agent_id` fields after the explicit `signet_agent_id` / `signetAgentId` forms.
- Added regressions for both paths.

Validation:
- `git diff --check` — passed
- `./node_modules/.bin/biome check integrations/opencode/plugin/src/index.ts integrations/opencode/plugin/src/index.test.ts surfaces/cli/src/commands/hook.ts surfaces/cli/src/commands/hook.test.ts` — passed
- `bun test integrations/opencode/plugin/src/index.test.ts surfaces/cli/src/commands/hook.test.ts` — 38 pass, 0 fail

integrations/opencode/plugin/src/index.ts
```diff
@@ -230,22 +277,32 @@ export const SignetPlugin: Plugin = async ({ directory, client: oc }) => {
...
+				startInject = await ensureSessionStarted(input.sessionID);
+			} catch {
+				// Signet context is optional; never break OpenCode prompt rendering.
+			}
 			const inject = pendingInject.get(input.sessionID);
+			const parts = [startInject, inject].filter((part) => part?.trim());
+			if (parts.length > 0) {
+				output.system.push(parts.join("\n"));
+			}
 			if (inject) {
```
PR-Reviewer-Ant, line 307: This pre-compaction payload still omits `agentId`, even though the plugin resolves `SIGNET_AGENT_ID` and this PR now threads it through the other OpenCode lifecycle calls. For named-agent OpenCode sessions, pre-compaction will fall back to whatever the daemon infers from `sessionKey` or to `default`, which means the summary guidance/recent-memory context can be read from the wrong Signet scope. Please include `agentId` here the same way `session-start`, `user-prompt-submit`, and `session-end` do.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e5edadd5`

</details>

I found one scoping hole in the OpenCode lifecycle wiring: named Signet agent scope is threaded through session start, prompt submit, and session end, but not pre-compaction.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The OpenCode plugin has a module-level agentId and this PR explicitly adds it to session-end while preserving it for session-start and prompt-submit. The changed pre-compaction payload adds sessionKey but still omits agentId, so named-agent OpenCode sessions can ask the daemon for default-scope compaction context. I did not have the full checked-out repo available to inspect the exact pre-compaction route implementation, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the latest OpenCode pre-compaction scoping feedback in `d198656d`.

What changed:
- Added the configured Signet `agentId` to OpenCode `/api/hooks/pre-compaction` requests, matching session-start, prompt-submit, session-end, and compaction-complete scope handling.
- Added a regression proving named-agent OpenCode pre-compaction requests include `agentId` and stay out of the daemon default scope.
- Kept the existing session-end scope regression intact.

Validation from `/tmp/signet-pr-monitor/pr-632`:
- `git diff --check` — passed
- `./node_modules/.bin/biome check integrations/opencode/plugin/src/index.ts integrations/opencode/plugin/src/index.test.ts` — passed
- `bun test integrations/opencode/plugin/src/index.test.ts` — 6 pass, 0 fail
- `cd integrations/opencode/plugin && bun run typecheck && bun run build` — passed

Waiting on refreshed CI/re-review before any merge action.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d198656d`

</details>

I did not find a concrete correctness, security, scoping, or data-integrity issue in this diff. The implementation matches the stated direction: harness-native sub-agent IDs are separated from Signet agent scope, parent context inference is agent-scoped, and the new transcript search surface uses existing recall permission checks.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed hook builders, daemon parent resolution, MCP tool wiring, and session routes are included with focused tests for the main scoping and lineage cases. I did not run the PR's validation commands or exercise the actual Claude Code/OpenCode hook event ordering at runtime, so confidence stops short of high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Merged current `origin/main` into this branch in `4aa9e59f` to resolve the conflicts introduced after #641 landed.

What changed in the conflict resolution:
- Kept this PR's sub-agent context continuity implementation and approved spec.
- Preserved the OpenCode pre-compaction `agentId` scoping fix from `d198656d`.
- Accepted main's predictor hard-deprecation/archive cleanup, so deprecated predictor docs/routes were not reintroduced.
- Resolved the `platform/daemon/src/hooks.ts` import conflict and Biome cleanup.

Validation run in `/tmp/signet-pr-monitor/pr-632`:
- `git diff --check`
- `./node_modules/.bin/biome check integrations/opencode/plugin/src/index.ts integrations/opencode/plugin/src/index.test.ts platform/daemon/src/hooks.ts docs/specs/approved/sub-agent-context-continuity.md`
- `bun test integrations/opencode/plugin/src/index.test.ts platform/daemon/src/hooks.test.ts platform/daemon/src/mcp/tools.test.ts platform/daemon/src/subagent-context.test.ts surfaces/cli/src/commands/hook.test.ts` (249 pass)
- `bun scripts/spec-deps-check.ts` (`86 specs, 86 indexed`)

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `4aa9e59f`

</details>

I found one race in the OpenCode plugin lifecycle that can duplicate session-start calls and inject inherited context twice when OpenCode fires hooks for the same session concurrently.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The race is visible in integrations/opencode/plugin/src/index.ts: ensureSessionStarted checks startedSessions before awaiting the daemon call, but only marks the session as started after the await. Both chat.message and experimental.chat.system.transform call this helper. I did not have an OpenCode runtime trace showing whether those hooks are serialized in practice.


- integrations/opencode/plugin/src/index.ts:204 - This guard is not single-flight: `startedSessions` is updated only after the awaited `/api/hooks/session-start` call returns. If `chat.message` and `experimental.chat.system.transform` run close together for a new session, both can observe the session as unstarted, both POST session-start, and both return inject text. That can duplicate inherited parent context in the same prompt and also create duplicate daemon lifecycle work. Consider tracking an in-flight promise per session, or adding the session to the started/in-flight map before awaiting and clearing it on failure.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the latest PR-Reviewer-Ant OpenCode lifecycle race feedback in `eb96fce2`.

What changed:
- Added a per-session in-flight start map so concurrent OpenCode `chat.message` and `experimental.chat.system.transform` hooks share one `/api/hooks/session-start` request.
- Only the hook that starts the request receives the inherited context directly; joiners wait for completion and return no duplicate inject text.
- Kept retry semantics intact by clearing the in-flight entry after completion and only marking a session started after the daemon start succeeds.
- Added a regression that gates the per-session start call, runs both hooks concurrently, and proves only one daemon session-start request and one inherited-context injection occur.

Validation from `/tmp/signet-pr-monitor/pr-632`:
- `bun test integrations/opencode/plugin/src/index.test.ts` — 7 pass, 0 fail
- `./node_modules/.bin/biome check integrations/opencode/plugin/src/index.ts integrations/opencode/plugin/src/index.test.ts` — passed
- `git diff --check` — passed
- `cd integrations/opencode/plugin && bun run typecheck && bun run build` — passed

Pushed to the PR branch; waiting on refreshed CI/re-review before any merge action.

surfaces/cli/src/commands/hook.ts
```diff
@@ -504,19 +569,22 @@ export function buildSessionEndBody(
...
 	sessionKey: string;
+	agentId?: string;
 	cwd: string;
 	reason: string;
 	runtimePath: typeof LEGACY_RUNTIME_PATH;
 } {
 	const body = input ?? {};
 	const sessionKey = pickSessionKey(body);
 	const sessionId = pickString(body.session_id, body.sessionId, sessionKey);
+	const agentId = pickString(body.signet_agent_id, body.signetAgentId, body.agentId, body.agent_id);
```
PR-Reviewer-Ant: This still treats a raw `agent_id` as the Signet persistence scope on session-end. For Claude Code sub-agent payloads, `agent_id` is the harness-native sub-agent id, which this PR otherwise moves into `harnessAgentId` so it does not fork Signet data scope. As written, a Claude Code sub-agent can start and prompt-submit under the parent/default scope, then save its final transcript under `agentId = <claude subagent id>`, making later parent-context lookup and transcript search miss the terminal transcript. `buildSessionEndBody` needs the same harness-aware separation used by `buildSessionStartBody` / `buildUserPromptSubmitBody`, or an explicit Signet id field should be required for scoping.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `eb96fce2`

</details>

I found one scope bug in the Claude Code session-end path: the PR carefully separates Claude's native sub-agent id during session-start and prompt-submit, but still treats the same field as a Signet agent scope when the session ends.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description explicitly claims harness sub-agent IDs do not become Signet agent scopes, and the changed CLI builders show the corrected handling for session-start and prompt-submit. The changed session-end builder still maps raw agent_id directly to agentId, so Claude Code sub-agent SessionEnd payloads can persist transcripts under the harness sub-agent id instead of the parent Signet scope.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Also addressed the github-code-quality stale import finding in `4c1323f7`.

What changed:
- Removed the unused `buildCandidateFeatures` named import from `platform/daemon/src/hooks.ts` while leaving `StructuralFeatures` and `getStructuralFeatures` intact.

Additional validation from `/tmp/signet-pr-monitor/pr-632`:
- `git diff --check` — passed
- `./node_modules/.bin/biome check integrations/opencode/plugin/src/index.ts integrations/opencode/plugin/src/index.test.ts platform/daemon/src/hooks.ts` — passed
- `bun test integrations/opencode/plugin/src/index.test.ts` — 7 pass, 0 fail

Pushed to the PR branch; CI restarted on the latest head.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `4c1323f7`

</details>

I did not find a concrete correctness, security, or data-integrity issue in the changed code. The implementation matches the PR description around separating harness sub-agent IDs from Signet scope, adding parent inference, and exposing transcript search behind recall permission.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the daemon routes, MCP tool, hook body construction, config loading, parent-resolution helper, and focused tests for Claude Code/OpenClaw/OpenCode lineage paths. I did not run the harnesses or daemon locally, so confidence is limited on live OpenCode/Claude hook event ordering beyond the included tests.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Merged current `origin/main` into this branch in `734900eb` to satisfy the up-to-date requirement. The merge was clean with no manual conflict resolution.

Validation run in `/tmp/signet-pr-monitor/pr-632` after the merge:
- `git diff --check origin/main...HEAD` — passed
- `bun scripts/spec-deps-check.ts` — passed (`86 specs, 86 indexed`)
- `bun test platform/daemon/src/hooks.test.ts platform/daemon/src/mcp/tools.test.ts platform/daemon/src/subagent-context.test.ts surfaces/cli/src/commands/hook.test.ts integrations/opencode/plugin/src/index.test.ts` — 250 pass, 0 fail

Latest PR-Reviewer-Ant review on `4c1323f7` reported no concrete issue. Waiting on refreshed CI before merge/auto-merge.

surfaces/cli/src/commands/hook.ts
PR-Reviewer-Ant: This still treats a raw `agent_id` from a Claude Code session-end payload as the Signet persistence scope. The rest of this PR is careful to keep Claude Code's native sub-agent `agent_id` in `harnessAgentId` for session-start and prompt-submit, but session-end will persist the terminal transcript under `agentId: <claude-subagent-id>` when Claude includes that field. That breaks the PR's stated invariant and can fork transcript/session-summary data away from the parent Signet scope. This builder needs the same harness-aware handling as the other hook bodies: explicit `signet_agent_id` / `signetAgentId` / `agentId` may set Signet scope, but Claude Code's native `agent_id` should not.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `734900eb`

</details>

I found one scope-handling regression in the legacy CLI hook path. The PR correctly separates Claude Code native `agent_id` from Signet scope for session start and prompt submit, but session end still sends that native sub-agent id as `agentId`.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description explicitly says harness sub-agent IDs should not become Signet agent scopes, and Claude Code hook research in the added spec says `agent_id` is present on all hook payloads inside sub-agent sessions. The changed `buildSessionEndBody` code still maps `body.agent_id` into `agentId`, unlike the session-start and prompt-submit builders.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the latest PR-Reviewer-Ant Claude Code session-end scoping feedback in `dd3dac84`.

What changed:
- `buildSessionEndBody()` now treats raw Claude Code `agent_id` as harness-native metadata, matching the session-start and prompt-submit builders.
- Explicit Signet scope fields (`signet_agent_id`, `signetAgentId`, `agentId`) still win.
- Non-Claude legacy `agent_id` callers remain supported as Signet scope.
- Added a regression proving Claude Code session-end `agent_id` does not become `agentId`.

Validation from `/tmp/signet-pr-monitor/pr-632`:
- `git diff --check` — passed
- `./node_modules/.bin/biome check surfaces/cli/src/commands/hook.ts surfaces/cli/src/commands/hook.test.ts` — passed after scoped format fix
- `bun test surfaces/cli/src/commands/hook.test.ts` — 34 pass, 0 fail

I also attempted `cd surfaces/cli && bun run build:cli`; it is currently blocked in this isolated worktree by unresolved workspace connector package imports (`@signet/connector-*`), which is local workspace install/build state rather than this hook change. CI has restarted on the pushed commit.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `dd3dac84`

</details>

I did not find a concrete correctness, security, scoping, or data-integrity issue in the changed paths. The implementation matches the PR description closely: harness-native sub-agent IDs are separated from Signet agent scope, parent context inference is agent-scoped, and the new session search surface is permission-gated and bounded.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the relevant daemon hook logic, route auth/scoping, MCP registration, config parsing, connector payload changes, and regression tests for Claude Code/OpenClaw/OpenCode lineage behavior. I could not run the repo tests or inspect unchanged migration/schema code in this review context, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #635, opened 2026-05-06
fix(cli): wrap launchd plist runtime in /bin/bash to avoid macOS AMFI exec failure

aaf2tbz:
## Summary

macOS launchd cannot directly exec homebrew-installed bun binaries due to Apple Mobile File Integrity (AMFI) restrictions. The daemon starts fine when spawned from a shell but `launchctl bootstrap` fails with `"Bootstrap failed: 5: Input/output error"` (EIO, exit code 78 `EX_CONFIG`).

Wrapping the invocation in `/bin/bash -c "exec <bun> <daemon>"` sidesteps the restriction because `/bin/bash` is a system binary that launchd trusts.

## Changes

| File | Change |
|---|---|
| `surfaces/cli/src/lib/runtime.ts` | `buildLaunchdDaemonPlist()` now emits `/bin/bash -c exec` instead of raw bun path |
| `platform/daemon/src/service.ts` | Parity fix in `generateLaunchdPlist()` for the service installation layer |
| `surfaces/cli/src/lib/runtime.test.ts` | Updated existing plist assertions + new regression test verifying bash wrapping |

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `platform/daemon` (service.ts)

## Screenshots

N/A, no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — no spec-governed behavior touched
- [x] Agent scoping verified on all new/changed data queries — N/A, no data queries changed
- [x] Input/config validation and bounds checks added — N/A, no new inputs
- [x] Error handling and fallback paths tested (no silent swallow) — N/A, fix is in plist generation
- [x] Security checks applied to admin/mutation endpoints — N/A
- [x] Docs updated for API/spec/status changes — N/A, no API changes
- [x] Regression tests added for each bug fix — 1 new test verifying bash wrapping
- [x] Lint/typecheck/tests pass locally — 11 tests pass, typecheck clean

N/A. The plist is regenerated on every `signet daemon start`, so the fix takes effect immediately without user action.

## Testing

- [x] `bun test surfaces/cli/src/lib/runtime.test.ts` (11 tests, 0 failures)
- [x] `bun run build` passes
- [x] Manually verified: daemon starts successfully via launchd with the wrapped plist on macOS (arm64, Sequoia)

## AI disclosure

## Root cause

`processRuntimeCommand()` in `runtime.ts` resolves to the absolute bun binary path (e.g. `/opt/homebrew/Cellar/bun/1.3.11/bin/bun`) which gets written directly into the launchd plist's `<ProgramArguments>`. macOS AMFI blocks execution of this binary when launched by launchd because it lacks the necessary exception that shell-spawned processes inherit. The fix wraps the invocation in `/bin/bash -c` which is a signed system binary, allowing the bun binary to execute within the bash subprocess.

### DIFF
--- a/platform/daemon/src/service.ts
+++ b/platform/daemon/src/service.ts
@@ -82,8 +82,9 @@ function generateLaunchdPlist(port: number = 3850): string {
     
     <key>ProgramArguments</key>
     <array>
-        <string>${resolveRuntimePath()}</string>
-        <string>${daemonPath}</string>
+        <string>/bin/bash</string>
+        <string>-c</string>
+        <string>exec ${resolveRuntimePath()} ${daemonPath}</string>
     </array>
     
     <key>EnvironmentVariables</key>
--- a/platform/daemon/src/service.ts
+++ b/platform/daemon/src/service.ts
@@ -84,7 +84,9 @@ function generateLaunchdPlist(port: number = 3850): string {
     <array>
         <string>/bin/bash</string>
         <string>-c</string>
-        <string>exec ${resolveRuntimePath()} ${daemonPath}</string>
+        <string>exec &quot;$0&quot; &quot;$1&quot;</string>
+        <string>${resolveRuntimePath()}</string>
+        <string>${daemonPath}</string>
     </array>
     
     <key>EnvironmentVariables</key>
--- a/surfaces/cli/src/lib/runtime.test.ts
+++ b/surfaces/cli/src/lib/runtime.test.ts
@@ -56,7 +56,9 @@ describe("buildLaunchdDaemonPlist", () => {
 		expect(plist).toContain("<key>Label</key>");
 		expect(plist).toContain("<string>ai.signet.daemon.test</string>");
 		expect(plist).toContain("<key>ProgramArguments</key>");
-		expect(plist).toContain("<string>/opt/signet/dist/daemon.js</string>");
+		expect(plist).toContain("<string>/bin/bash</string>");
+		expect(plist).toContain("<string>-c</string>");
+		expect(plist).toContain(`exec ${process.execPath} /opt/signet/dist/daemon.js`);
 		expect(plist).toContain("<key>SIGNET_PORT</key>");
 		expect(plist).toContain("<string>3850</string>");
 		expect(plist).toContain("<key>SIGNET_HOST</key>");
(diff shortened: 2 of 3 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `45e3e380`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `34823013`

</details>

The launchd workaround is pointed at the right failure mode, but the new bash wrapper builds an unquoted shell command from filesystem paths. That makes valid installs with spaces in the path fail, and it unnecessarily turns path strings into shell syntax.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both changed plist generators now pass a single `bash -c` string assembled from runtime and daemon paths. The issue is visible directly in `surfaces/cli/src/lib/runtime.ts` and mirrored in `platform/daemon/src/service.ts`; no runtime repro is needed to see that shell word splitting will occur.


- `surfaces/cli/src/lib/runtime.ts:463` - This fixes the AMFI direct-exec problem, but it also changes these paths from argv entries into shell syntax without shell quoting. If `processRuntimeCommand()` or `input.daemonPath` contains a space, the daemon will fail to start; if either path contains shell metacharacters, bash will interpret them. XML escaping only makes the plist valid XML, it does not quote for `/bin/bash -c`. Please build the command with proper shell escaping, or use `/bin/bash -c 'exec "$0" "$1"' <runtime> <daemon>` so the paths stay argv values.
- `platform/daemon/src/service.ts:86` - Same problem in the service-layer parity generator: `exec ${resolveRuntimePath()} ${daemonPath}` is parsed by bash, so spaces or shell metacharacters in the runtime or daemon path break startup or become executable syntax. This path should use the same safe argv-preserving wrapper as the CLI plist generator.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `45e3e380`

</details>

The updated launchd wrapper now preserves the runtime and daemon paths as argv values instead of interpolating them into shell syntax, which addresses the earlier correctness concern. I did not find a new blocking issue in the changed lines.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The relevant changed launchd ProgramArguments are fully included for both surfaces/cli/src/lib/runtime.ts and platform/daemon/src/service.ts. The bash -c form uses exec "$0" "$1" with the runtime and daemon path supplied as separate plist arguments, so spaces and shell metacharacters in those paths are no longer interpreted by bash.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #636, opened 2026-05-06
fix(memory): harden recall authorization

NicholaiVogel:
## Summary

- Default agent-scoped recall to isolated access when `readPolicy` is omitted.
- Add an explicit authorization gate before content-bearing recall stages such as SEC coverage, reranking, dampening, currentness, hydration, transcript expansion, and access tracking.
- Fix scoped over-fetch assembly and access tracking so only authorized returned memory IDs are counted.
- Sort and max-merge traversal-primary candidates so weaker graph evidence does not discard stronger direct evidence.
- Disable source-backed vector chunk rescue for project-scoped recall until chunk embeddings carry a strong project/source-root binding; project-scoped recall still uses authorized memory rows, native artifacts, and transcript fallbacks.
- Expand the memory-search docs and inline comments to document the recall phases and safety boundary.

## Why

Memory search collects candidates from broad channels: FTS, hints, vector search, graph traversal, structured path search, source chunks, native artifacts, and transcripts. Some of those channels can produce IDs before normal hydration filters run. The previous ordering let several post-processing stages fetch or act on candidate content before final filtered hydration. This patch makes the authorization boundary explicit and shared.

## Review Follow-Up

- Fixed transcript summary hydration so omitted `agentId` still defaults to the `default` agent boundary instead of widening to all agents.
- Added a regression test covering same-source transcript summary hydration across default and non-default agents.
- Closed the source-backed vector chunk project leak by skipping that rescue path for project-scoped recall until the source index has a strong source-root/project key.
- Added regressions for project-scoped source chunk spoofing and ambiguous same-basename source artifacts.
- Updated `docs/MEMORY.md` to document that fallback paths are enabled only when their backing rows can satisfy the caller's scope boundary.

## PR Readiness (MANDATORY)

## Validation

- `bun test platform/daemon/src/memory-search.test.ts` — 36 pass.
- `bunx biome check platform/daemon/src/memory-search.ts platform/daemon/src/memory-search.test.ts docs/MEMORY.md docs/API.md` — exits 0 with existing `any` warnings in `memory-search.ts`.
- `git diff --check` — clean.

## Notes

The broader daemon type declaration build is still blocked by unrelated existing rootDir/test typing issues and an existing `provider-safety.ts` union narrowing error, so this PR uses the focused recall test plus targeted lint/format checks.

### DIFF
--- a/docs/MEMORY.md
+++ b/docs/MEMORY.md
@@ -123,89 +123,170 @@ with a relationship label. Memory-entity associations are tracked in
 `memory_entity_mentions`.
 
 
-Hybrid Search
+Hybrid Recall
 -------------
 
-Search (`platform/core/src/search.ts`) blends two independent signals:
-BM25 keyword relevance and cosine vector similarity. These signals are
-useful because they help build a bounded candidate pool for later
-ranking and context selection.
+Recall is implemented in `platform/daemon/src/memory-search.ts` by
+`hybridRecall`. Its job is not to find one perfect row. Its job is to
+collect a broad candidate set, filter it to the caller's allowed view,
+shape the evidence, and return a bounded list of useful context.
 
-### BM25 Keyword Search
+The route layer checks the caller has `recall` permission. The search
+layer then enforces the data boundary again with one shared filter:
 
-FTS5 BM25 scores are negative (lower = better match). The
-implementation normalizes them to [0, 1] using `1 / (1 + |score|)`.
-This makes them comparable with cosine similarity scores.
+- `agentId` and `readPolicy` control agent visibility. If `agentId` is
+  present and `readPolicy` is omitted, recall defaults to isolated access.
+- `project` restricts results to that project.
+- `scope` restricts results to that explicit scope. When `scope` is
+  omitted, normal recall excludes scoped memories.
+- `type`, `tags`, `who`, `pinned`, `importance_min`, `since`, and `until`
+  narrow the memory rows inside that visibility boundary.
 
-```sql
-SELECT m.id, bm25(memories_fts) AS raw_score
-FROM memories_fts
-JOIN memories m ON memories_fts.rowid = m.rowid
-WHERE memories_fts MATCH ?
-ORDER BY raw_score
-LIMIT ?
-```
+The important safety rule is:
 
-### Vector Search
+> Candidate IDs may be broad, but memory content is not loaded, reranked,
+> summarized, dampened, expanded, or access-tracked until the shared filter
+> has authorized those IDs.
 
-Vector search uses `sqlite-vec` with a `vec_embeddings` virtual table.
-Queries use the KNN MATCH syntax. The distance metric is cosine, which
-`sqlite-vec` returns as `(1 - similarity)`, so the implementation
-converts: `similarity = 1 - distance`.
+That rule matters because vector search and graph traversal are intentionally
+high-recall channels. They can produce IDs that did not pass through the FTS
+SQL filter. `hybridRecall` therefore runs an explicit `authorize_candidates`
+stage before any content-bearing stage.
 
-```sql
-SELECT e.source_id, v.distance
-FROM vec_embeddings v
-JOIN embeddings e ON v.id = e.id
-JOIN memories m ON e.source_id = m.id
-WHERE v.embedding MATCH ? AND k = ?
-ORDER BY v.distance
-```
+### Phase 1: Prepare the Query
 
-### Alpha Blending
+Recall starts with the original `query`, an optional `keywordQuery`, and a
+normalized `limit`. `limit` is clamped to a bounded range so callers cannot
+force unbounded recall work.
 
-Scores from both sources are merged using a configurable alpha weight:
+The keyword side uses `sanitizeFtsQuery` to produce a safe FTS5 `MATCH`
+expression. If sanitization leaves no searchable tokens, the FTS and hint
+paths are skipped instead of issuing an empty `MATCH`.
 
-```
-score = alpha × vector_score + (1 - alpha) × bm25_score
-```
+`expandRecallKeywordQuery` adds a small mechanical expansion set for known
+class-to-instance gaps. The expansion affects keyword and hint search only.
+Vector search and model summaries still use the user's original query.
 
-If only one source returns a result for a given memory ID, its score
-is used directly without blending. Results below `min_score` are
-dropped. The default alpha is 0.7 (70% semantic, 30% keyword).
+### Phase 2: Collect Candidate IDs
 
-Configure in `agent.yaml`:
+Candidate collection uses several independent channels:
 
-```yaml
-search:
-  alpha: 0.7       # vector weight; 1 - alpha goes to BM25
-  min_score: 0.1   # drop results below this threshold
-```
+- **Memory FTS** queries `memories_fts`, joins to `memories`, applies the
+  shared filter, and normalizes BM25 scores within the batch.
+- **Prospective hints** query `memory_hints_fts`. Hints are write-time
+  alternate phrasings. They can rescue a memory whose content does not use
+  the same words as the current query.
+- **Vector search** embeds the original query and searches `vec_embeddings`
+  through `sqlite-vec`. Vector search cannot pre-filter every recall scope,
+  so constrained searches over-fetch and rely on authorization after merge.
+- **Structured path candidates** search entity/aspect/group/claim paths so
+  structured knowledge can surface even when its prose is sparse.
+- **Graph traversal** resolves focal entities and walks the knowledge graph.
+  In traversal-primary mode, graph results are treated as a first-class
+  retrieval channel rather than a small boost.
 
-### Graph-Augmented Search
+Flat lexical, hint, vector, and structured scores are merged by memory ID.
+When a memory appears in multiple channels, the strongest calibrated evidence
+wins unless there is an explicit blend. Pure hint-only hits are capped so
+generated hints can rescue recall but do not outrank directly grounded
+keyword, vector, or structured evidence.
 
-When `graphEnabled = true`, the recall endpoint runs an additional
-graph lookup before returning results
-(`platform/daemon/src/pipeline/graph-search.ts`).
+Traversal-primary mode sorts traversal candidates before selection and
+max-merges traversal overlap with flat candidates. This prevents a weak graph
+score from discarding stronger direct evidence for the same memory.
 
-The query is tokenized and matched against entity `canonical_name`
-values (top 20 by `mentions` count). From those seed entities, the
-graph expands one hop in both directions through the `relations` table
-(up to 50 neighbors). The resulting expanded entity set is joined to
-`memory_entity_mentions` to produce a set of linked memory IDs.
+### Phase 3: Authorize Candidates
 
-Memories in this set receive a configurable boost to their combined
-score. The graph lookup runs synchronously with a deadline to prevent
-slow queries from blocking recall. On timeout or error, the boost
-set is empty and search degrades gracefully. In other words, graph
-structure improves candidate quality when available but does not become
-a hard dependency for recall.
+After candidate collection and coarse score fusion, `authorize_candidates`
+requeries `memories` with the shared filter and removes anything outside the
+caller's allowed view. This is the boundary between "IDs only" and "content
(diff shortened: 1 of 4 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `24e1e6cf`.

platform/daemon/src/memory-search.ts
```diff
@@ -921,20 +956,22 @@ function loadStructuredSummariesBySourceId(params: RecallParams, sourceIds: read
 	try {
 		return getDbAccessor().withReadDb((db) => {
 			const placeholders = unique.map(() => "?").join(", ");
```
PR-Reviewer-Ant, line 972: This drops the default agent isolation that the old query enforced. `buildTranscriptRecallHits` still searches transcripts under `params.agentId ?? "default"`, but `loadStructuredSummariesBySourceId` now applies no agent filter at all when `agentId` is omitted, so a transcript fallback can attach a structured memory summary from another agent if it shares the same `source_id` (and no project filter is present). That is exactly the kind of content-bearing hydration this PR is trying to harden. Please keep the default agent boundary here, e.g. use `params.agentId ?? "default"` for the scope clause, or otherwise require the route layer to always pass an explicit agent id before this helper can run.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `70df2f8d`

</details>

I found one authorization regression in the transcript summary hydration path. Most of the new candidate authorization boundary is coherent, but this helper now drops the default agent filter when `agentId` is omitted.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed `loadStructuredSummariesBySourceId` code only applies `buildAgentScopeClause` when `params.agentId` is present, while callers such as transcript fallback still default transcript lookup to `params.agentId ?? "default"`. The previous query always included `AND agent_id = ?`, so this is a direct widening of summary hydration.

docs/MEMORY.md
```diff
@@ -123,89 +123,164 @@ with a relationship label. Memory-entity associations are tracked in
...
+rows.
+
+### Source and Transcript Rescue Paths
+
+If normal memory candidates produce no hits, recall can fall back to source
+chunks, native memory artifacts, and transcript FTS. These paths have their
+own agent/project/source filters and return synthetic recall rows rather than
+materializing new `memories` records.
+
+When `expand` is enabled and ordinary memory rows reference session source
```
PR-Reviewer-Ant, line 274: This safety claim does not hold for source-backed vector chunks. `buildSourceChunkVectorHits` only receives `agentId` and filters `embeddings` by `agent_id`; it does not take or apply `params.project`. Both the no-hit fallback and the supplementary path can therefore return `source_obsidian` chunk content for the same agent from outside a project-scoped recall. Since this PR is specifically hardening content-bearing recall authorization, the source chunk path needs the same project boundary before it can add content to `results`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `2c869025`

</details>

I found one project-scoping hole that still lets source-backed vector chunks bypass the recall project boundary. The main memory-row authorization path is much tighter, but the supplementary source chunk path does not match the safety boundary described by this PR.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed docs state that source rescue paths have agent/project/source filters, and the changed recall assembly now treats supplementary rows as part of the final capped response. In the included implementation, source chunk vector recall is still called with only params.agentId and its query filters only embeddings.agent_id, so a project-scoped recall can add content from another project. I did not run a local repro, but the SQL and call sites are visible in the provided changed file contents.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/memory-search.ts
```diff
@@ -631,37 +667,66 @@ function buildSourceChunkVectorHits(
...
+				}
+				if (projectByPath.size === 0) return [];
+			}
+
 			const rows = db
 				.prepare(
 					`SELECT id, source_id, vector, chunk_text, created_at
 					 FROM embeddings
-					 WHERE source_type = 'source_obsidian_chunk' AND vector IS NOT NULL${agentWhere}`,
+					 WHERE source_type = 'source_obsidian_chunk'
```
PR-Reviewer-Ant, line 685: This project authorization is still content-derived: `sourcePathFromChunkText(row.chunk_text)` decides whether the embedding is allowed, but the chunk content and `source_id` returned to the caller come from the `embeddings` row. A stale or malformed chunk whose `chunk_text` header names an allowed project path can pass this check while carrying another source's content/source_id. Since this fallback is explicitly meant to prevent project B source chunks from surfacing in project A recall, bind the check to a live `memory_artifacts` row by joining/matching on the artifact identity (`source_id`/session/source artifact id, or another canonical key), and only return chunk text from rows whose artifact join satisfies `agent_id`, `project`, and `is_deleted = 0`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c864fc6a`

</details>

I found one authorization gap in the new source-chunk vector path. The main memory candidate boundary is much tighter now, but the chunk fallback/supplement still trusts a path parsed out of chunk text instead of tying the chunk to the live artifact row that authorizes it.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed source-chunk path in platform/daemon/src/memory-search.ts explicitly authorizes project access by building a source_path map from memory_artifacts, then accepts embeddings based on sourcePathFromChunkText(row.chunk_text). The returned content and source_id still come directly from embeddings, so the authorization check is not bound to the embedding's source_id or a joined artifact row.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `08369ea1`

</details>

I found one authorization hole in the new source-backed chunk filtering. The main memory candidate gate is much better now, but the chunk rescue path can still authorize the wrong project when two source artifacts share a relative filename.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The leak is directly visible in platform/daemon/src/memory-search.ts: sourceIdMatchesArtifactPath treats any suffix of the artifact path as a valid source_id match, and buildSourceChunkVectorHits then returns the embedding row's chunk_text under the matched artifact's project. The added regression uses different basenames, so it does not cover the basename-collision case.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/memory-search.ts
```diff
@@ -627,41 +668,83 @@ function sourcePathFromChunkText(chunkText: string): string {
...
+	const normalizedId = sourceId.replace(/\\/g, "/");
+	const parts = sourcePath.replace(/\\/g, "/").split("/").filter(Boolean);
+	// Obsidian chunk IDs are `${sourceId}:${vaultRelativePath}#...`.
+	// Match the embedding-owned ID back to the live artifact path instead of
+	// trusting the metadata copied into chunk_text.
+	for (let index = 0; index < parts.length; index++) {
+		if (normalizedId.includes(`:${parts.slice(index).join("/")}#`)) return true;
+	}
+	return false;
+}
```
PR-Reviewer-Ant: This suffix match can still authorize the wrong chunk across projects. For an authorized artifact `/workspace/project-a/source.md`, `parts.slice(2).join("/")` becomes `source.md`, so a different embedding with `source_id = "obsidian:vault-b:source.md#..."` also passes and its `chunkText` is returned as project A content. That reopens the same class of source-backed recall leak whenever two projects have an Obsidian note with the same relative/basename path. The auth check needs to compare against a stable artifact-owned identifier or the exact vault-relative path that was used to create the embedding, not any suffix of the filesystem path.

platform/daemon/src/memory-search.ts
```diff
@@ -627,41 +668,92 @@ function sourcePathFromChunkText(chunkText: string): string {
...
-					sourcePath: sourcePathFromChunkText(row.chunk_text),
-					chunkText: row.chunk_text,
-					score: cosineSimilarity(
-						queryVec,
-						new Float32Array(row.vector.buffer, row.vector.byteOffset, row.vector.byteLength / 4),
-					),
-					createdAt: row.created_at,
-				}))
+				.flatMap((row) => {
+					const artifact = project ? uniquelyMatchedSourceArtifact(row.source_id, liveArtifacts) : null;
```
PR-Reviewer-Ant: This still allows project-scoped source chunk leakage when an out-of-project embedding uses a `source_id` whose vault-relative filename matches a live artifact in the requested project. `sourceIdMatchesArtifactPath()` accepts any suffix of the artifact path, so `obsidian:vault-b:shared.md#...` can uniquely match `/workspace/project-a/shared.md` if there is no second live `shared.md` artifact to make it ambiguous. The current basename-collision test only covers the two-artifact ambiguous case; it does not cover the more dangerous case where the spoofed embedding is the only basename match. Project authorization needs a stronger binding than suffix/basename containment, or it should fail closed when the embedding source cannot be tied to the exact artifact identity.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e0247151`

</details>

I found one authorization gap in the new source-chunk project binding. The general candidate authorization boundary is much clearer, but this fallback still has a basename-based spoof path.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The source chunk fallback authorizes an embedding by checking whether its source_id contains any suffix of a live artifact source_path. That logic is visible in memory-search.ts and the new tests do not cover the same-basename spoof case where only the requested project has that basename.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/memory-search.ts
```diff
@@ -631,37 +667,50 @@ function buildSourceChunkVectorHits(
...
 	limit: number,
-	agentId?: string,
+	agentId: string,
+	project?: string,
 ): SourceChunkVectorHit[] {
-	if (!queryVec || limit <= 0 || !agentId) return [];
+	if (!queryVec || limit <= 0) return [];
+	// Source chunk embeddings only carry agent_id plus an embedding source_id.
+	// The live artifact table carries project, but not the source root/id needed
+	// to bind an embedding to that project without filename spoofing. Skip this
```
PR-Reviewer-Ant, line 676: This does not implement the PR's stated source-chunk behavior. The description says project-scoped source chunk recall is authorized through live `memory_artifacts` rows and bound to the embedding-owned `source_id`, but this branch drops all project-scoped source chunk vector fallback/supplement results before any artifact authorization can happen. That may be an acceptable temporary safety tradeoff, but then the PR description and readiness claims are overstating what landed, and project-scoped recall loses same-project source chunk rescue entirely.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `75c4e7a9`

</details>

I found one behavior mismatch: the PR says project-scoped source chunk vector recall is authorized through live artifact rows, but the implementation disables that path entirely for project-scoped calls.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description explicitly claims project-scoped source chunk fallback/supplement results are scoped through live project-bearing source artifacts, while platform/daemon/src/memory-search.ts returns no source chunk vector hits whenever project is present. The added tests also lock in the disabled behavior rather than the described authorized-through-artifacts behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `24e1e6cf`

</details>

I did not find a concrete bug or security regression in the changed recall authorization flow. The implementation matches the PR description: candidate IDs are filtered before content-bearing stages, project-scoped source chunk vector rescue is disabled, and the new regressions cover the main leakage paths described.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff directly shows the new shared filter in buildFilterClause, authorizeScoredCandidates before reranking/dampening/hydration/access tracking, project-scoped source chunk rescue returning no hits, and focused tests for agent default isolation, transcript summary hydration, project over-fetch assembly, and source chunk spoofing. I did not run the focused Bun test locally, so confidence is not high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #634, opened 2026-05-06
fix(sources): index Obsidian sources asynchronously

NicholaiVogel:
## Summary

Async Obsidian source indexing so connecting a large vault no longer blocks the daemon request path. The dashboard now shows active source indexing progress while files are scanned one at a time.

## Changes

- Queue `POST /api/sources/obsidian` indexing work in a daemon-side background job and return `202` with job metadata.
- Track source indexing progress with queued/running/complete/error state, scanned/total/indexed counts, current file path, and cancellation cleanup on disconnect.
- Have the native source bridge discover and index markdown files sequentially with per-file progress callbacks.
- Add dashboard API types and a Sources tab progress bar that polls while a source job is active.
- Add regression coverage for non-blocking connect, in-flight disconnect cleanup, and progress callback behavior.

## Type

- [ ] `feat` — new user-facing feature (bumps minor)
- [x] `fix` — bug fix
- [ ] `refactor` — restructure without behavior change
- [ ] `chore` — build, deps, config, docs
- [ ] `perf` — performance improvement
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/cli` / dashboard
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

Not captured yet. The changed UI is an in-flight source indexing progress state and requires a live queued source job to render meaningfully.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

## Migration Notes (if applicable)

- [ ] Migration is idempotent
- [ ] Daemon Rust parity reviewed or explicitly N/A
- [ ] Rollback / compatibility note included in PR description

No migrations touched.

## Testing

- [ ] `bun test` passes
- [ ] `bun run typecheck` passes
- [ ] `bun run lint` passes
- [ ] Tested against running daemon
- [ ] N/A

Focused validation run locally:

- [x] `bun test platform/daemon/src/routes/sources-routes.test.ts`
[...]

PR #638, opened 2026-05-07
fix(daemon): stabilize Obsidian source sync

NicholaiVogel:
## Summary

This follow-up fixes the Obsidian source sync path that still saturated the daemon after #634 merged.

The remaining problems were:

- daemon startup still did heavy configured-source work inline with embeddings/graph projection available on the hot path
- configured source sync defaulted to periodic polling, so a large vault could immediately start another full scan after completing
- unchanged artifacts/chunks were only cached in memory, so a cold daemon restart still repeated expensive work
- Obsidian wiki-link graph projection could walk the vault per link
- dashboard progress existed for explicit source jobs, but not automatic startup source sync

## What changed

- Startup source sync now scans only configured sources, not Codex/Claude harness memory roots.
- Startup and dashboard source sync disable inline embeddings, graph projection, and stale cleanup so the daemon indexes source artifacts without blocking on semantic expansion/cleanup work.
- Startup source sync uses `pollIntervalMs: 0`, so it does one paced scan instead of polling the vault forever.
- Source sync progress is shared through `/api/sources`, so dashboard source cards can show automatic startup progress too.
- Obsidian artifact/chunk indexing skips unchanged persisted rows after cold restart.
- Obsidian graph wikilink resolution no longer walks the whole vault per link.
- Poll ticks no longer queue trailing full rescans while a long sync is already running.

## Live validation

Hotpatched the installed daemon at `/home/nicholai/node_modules/signetai/dist/daemon.js` and restarted it with the Obsidian source enabled.

Observed after restart:

- `/api/status` returned immediately while source sync was running
- `/api/sources` reported a running startup job with scanned/total/currentPath progress
- source scan completed once, updated `lastIndexedAt`, and no longer used periodic polling
- no new live event-loop warnings appeared in the sampled log window after the final restart

## Validation

```bash
bunx biome check platform/daemon/src/daemon.ts platform/daemon/src/native-memory-sources.ts platform/daemon/src/native-memory-sources.test.ts platform/daemon/src/obsidian-source-embeddings.ts platform/daemon/src/obsidian-source-embeddings.test.ts platform/daemon/src/obsidian-source-graph.ts platform/daemon/src/obsidian-source-graph.test.ts platform/daemon/src/routes/sources-routes.ts platform/daemon/src/routes/sources-routes.test.ts platform/daemon/src/source-index-progress.ts
bun test platform/daemon/src/routes/sources-routes.test.ts platform/daemon/src/native-memory-sources.test.ts platform/daemon/src/obsidian-source-graph.test.ts platform/daemon/src/obsidian-source-embeddings.test.ts
bun run --filter @signet/daemon build
```

Notes: daemon build still emits pre-existing Svelte a11y/deprecation warnings from dashboard components unrelated to this patch.

## Readiness checklist

Notes:
- Spec dependency check passed with `bun scripts/spec-deps-check.ts`; no spec/API/status docs changed because this is a daemon reliability bug fix within the existing Sources contract.
- No new admin endpoints were added. Existing source connect/disconnect routes keep their current route shape; this PR narrows background work and progress reporting.

### DIFF
--- a/platform/daemon/src/native-memory-sources.ts
+++ b/platform/daemon/src/native-memory-sources.ts
@@ -1,4 +1,3 @@
-import { createHash } from "node:crypto";
 import { existsSync, readFileSync, statSync } from "node:fs";
 import { readdir } from "node:fs/promises";
 import { homedir } from "node:os";
@@ -9,7 +8,7 @@ import { yieldEvery } from "./async-yield";
 import { getDbAccessor } from "./db-accessor";
 import { logger } from "./logger";
 import type { EmbeddingConfig } from "./memory-config";
-import { indexExternalMemoryArtifact, softDeleteArtifactRowsForPath } from "./memory-lineage";
+import { hashNormalizedBody, indexExternalMemoryArtifact, softDeleteArtifactRowsForPath } from "./memory-lineage";
 import {
 	type SourceEmbeddingFetch,
 	indexObsidianSourceEmbeddings,
@@ -38,8 +37,12 @@ export interface NativeMemoryFilePattern {
 }
 
 export interface NativeMemoryBridgeHandle {
-	readonly syncExisting: () => Promise<number>;
+	readonly syncExisting: (options?: NativeMemoryBridgeSyncOptions) => Promise<number>;
 	readonly close: () => Promise<void>;
+}
+
+export interface NativeMemoryBridgeSyncOptions {
+	readonly requestResyncIfBusy?: boolean;
 }
 
 export interface NativeMemoryBridgeOptions {
@@ -50,6 +53,9 @@ export interface NativeMemoryBridgeOptions {
 	readonly agentsDir?: string;
 	readonly includeConfiguredSources?: boolean;
 	readonly yieldEveryFiles?: number;
+	readonly sourceFileDelayMs?: number;
+	readonly sourceCleanupEnabled?: boolean;
+	readonly sourceGraphEnabled?: boolean;
 	readonly onFileIndexed?: (event: NativeMemoryFileIndexEvent) => void;
 }
 
@@ -67,6 +73,7 @@ interface IndexedNativeMemory {
 }
 
 const indexed = new Map<string, IndexedNativeMemory>();
+const DEFAULT_OBSIDIAN_SOURCE_FILE_DELAY_MS = 250;
 
 function codexRoot(): string {
 	return join(homedir(), ".codex");
@@ -222,22 +229,33 @@ function sourceStateKey(source: NativeMemorySource, agentId: string): string {
 }
 
 function contentFingerprint(content: string): string {
-	return createHash("sha256").update(content).digest("hex");
+	return hashNormalizedBody(content);
 }
 
-function nativeArtifactRowExists(filePath: string, agentId: string): boolean {
+function sourceFileDelayMs(source: NativeMemorySource, options: NativeMemoryBridgeOptions): number {
+	if (options.sourceFileDelayMs !== undefined) {
+		return Math.max(0, Math.floor(options.sourceFileDelayMs));
+	}
+	return source.harness === "obsidian" ? DEFAULT_OBSIDIAN_SOURCE_FILE_DELAY_MS : 0;
+}
+
+function sleep(ms: number): Promise<void> {
+	return ms > 0 ? new Promise((resolve) => setTimeout(resolve, ms)) : Promise.resolve();
+}
+
+function nativeArtifactContentHash(filePath: string, agentId: string): string | null {
 	const sourcePath = filePath.replace(/\\/g, "/");
 	try {
 		return getDbAccessor().withReadDb((db) => {
 			const row = db
 				.prepare(
-					"SELECT 1 FROM memory_artifacts WHERE agent_id = ? AND source_path = ? AND COALESCE(is_deleted, 0) = 0 LIMIT 1",
+					"SELECT source_sha256 FROM memory_artifacts WHERE agent_id = ? AND source_path = ? AND COALESCE(is_deleted, 0) = 0 LIMIT 1",
 				)
-				.get(agentId, sourcePath);
-			return !!row;
+				.get(agentId, sourcePath) as { source_sha256: string } | undefined;
+			return row?.source_sha256 ?? null;
 		});
 	} catch {
-		return false;
+		return null;
 	}
 }
 
@@ -281,7 +299,7 @@ export async function indexNativeMemoryFile(
 	source: NativeMemorySource,
 	filePath: string,
 	agentId = resolveDaemonAgentId(),
-	options: Pick<NativeMemoryBridgeOptions, "embeddingConfig" | "fetchEmbedding"> = {},
+	options: Pick<NativeMemoryBridgeOptions, "embeddingConfig" | "fetchEmbedding" | "sourceGraphEnabled"> = {},
 ): Promise<boolean> {
 	const pattern = matchesPattern(source, filePath);
 	if (!pattern) return false;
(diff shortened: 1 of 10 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `d6dfe84b`.

platform/daemon/src/obsidian-source-graph.ts
```diff
@@ -509,18 +512,16 @@ export function indexObsidianSourceStructure(
...
-			const targetRel = found ? relPath(root, normalizedPath(found)) : `${link}.md`;
-			const targetPath = found ? normalizedPath(found) : join(root, targetRel);
+			const targetPath = resolveWikiLinkPath(root, filePath, link);
 			const target = upsertSourceEntity(db, {
-				id: idFor(input.agentId, input.sourceId, "document", targetRel),
-				name: displayNameForFile(targetPath),
-				canonicalName: sourceCanonical(input.sourceId, "document", targetRel),
-				entityType: found ? "source_document" : "source_document_reference",
+				id: idFor(input.agentId, input.sourceId, "document", targetPath.rel),
+				name: displayNameForFile(targetPath.path),
```
PR-Reviewer-Ant, line 553: This removes the per-link vault walk, but it also stops resolving normal Obsidian links to notes elsewhere in the vault. For `[[Deep Target]]`, the new code only checks the current directory and `${root}/Deep Target.md`; if the actual file is `literature/Other/Deep Target.md`, we now write a `source_document_reference` pointing at a non-existent root-level path even though the source document exists. That corrupts the graph relationship rather than just making resolution cheaper. The fix should keep an indexed/resolved view of vault markdown paths for the scan, or otherwise preserve existing cross-folder basename resolution without walking the vault once per link.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `911ce48a`

</details>

I found one correctness issue in the Obsidian wikilink graph path: the PR removes the expensive vault walk, but it also stops resolving valid links outside the current folder or vault root.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed resolver in platform/daemon/src/obsidian-source-graph.ts now only checks dirname(filePath) and root before creating a source_document_reference. The added test demonstrates an existing nested target is intentionally recorded as an unresolved root-level reference, so the regression is directly visible in the diff.

platform/daemon/src/obsidian-source-graph.ts
```diff
@@ -305,23 +311,48 @@ function wikiLinks(content: string): string[] {
...
+	const targetPath = markdownTarget(target);
+	const candidates = [join(dirname(filePath), targetPath), join(root, targetPath)];
+	for (const candidate of candidates) {
+		const found = existingMarkdownPath(candidate);
+		if (found) return { path: found, rel: relPath(root, found), found: true };
 	}
-	return null;
+	const indexed = index?.byRel.get(targetPath) ?? (target.includes("/") ? undefined : index?.byStem.get(target));
+	if (indexed) return { path: indexed, rel: relPath(root, indexed), found: true };
+	const rel = targetPath;
```
PR-Reviewer-Ant, line 361: This changes wikilink resolution from normalized stem matching to exact string lookup in the prebuilt index. The removed code slugged both the link target and markdown basename, so links like `[[deep target]]`, `[[Deep-Target]]`, or other case/punctuation variants could still resolve to `Deep Target.md`. The new `byRel.get(targetPath)` / `byStem.get(target)` path only resolves exact relative paths or exact display names, and otherwise writes a `source_document_reference`. That means the performance fix can silently downgrade existing vault links into unresolved references. The path index should keep the old normalized-key behavior, just computed once per scan instead of by walking the vault per link.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d757f59d`

</details>

I found one regression in the new Obsidian wikilink resolver: the per-scan index avoids the vault walk, but it also drops the old normalized stem matching behavior for common case/title variants.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed resolver in obsidian-source-graph.ts now checks exact relative paths and exact by-stem keys from displayNameForFile, while the removed findMarkdownByStem path compared slug-normalized stems. I did not run a live vault repro, but the regression follows directly from the diff and the included tests only cover the exact-case target form.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c9ea9574`

</details>

I found one progress-reporting bug in the new startup sync path. The indexing behavior itself looks aligned with the PR description, but multi-source startup progress can report the wrong completed count for sources that did not emit file progress.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The startup sync completion path in platform/daemon/src/daemon.ts completes every source job from one aggregate syncExisting() result, falling back to that aggregate when a per-source progress value was never set. The changed tests cover shared progress for a single source and multi-source bridge events, but not startup completion with multiple configured sources where one source has zero indexed files.


- platform/daemon/src/daemon.ts:1643 - This completion fallback can assign the aggregate bridge result to an individual source job. `syncExisting()` returns the total indexed count across all configured sources, but this loop completes every startup job with `getSourceIndexJob(sourceId)?.indexed ?? indexed`. If a configured source has no `onFileIndexed` events, for example an empty vault or missing/filtered files, its job never gets a per-source `indexed` value and will be reported as having indexed the aggregate count from other sources. That makes `/api/sources` progress misleading exactly in the automatic startup path this PR is adding. Track per-source changed counts explicitly and default missing per-source counts to `0`, not the bridge-wide total.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/native-memory-sources.ts
```diff
@@ -305,11 +327,16 @@ export async function indexNativeMemoryFile(
 
 	const key = fingerprintKey(source, filePath, agentId);
 	const hash = contentFingerprint(content);
+	const persistedHash = nativeArtifactContentHash(filePath, agentId);
 	const cached = indexed.get(key);
 	if (cached?.contentHash === hash) {
-		if (nativeArtifactRowExists(filePath, agentId)) return false;
+		if (persistedHash === hash) return false;
 		indexed.delete(key);
```
PR-Reviewer-Ant, line 403: This persisted-hash early return makes the lightweight source scan sticky. Startup and dashboard sync now call this path with `sourceGraphEnabled: false` and without embedding options, so they can create or refresh the `memory_artifacts` row without graph rows or chunk embeddings. A later full sync with graph/embedding enabled will hit `persistedHash === hash`, return here, and never run `indexObsidianSourceStructure` or `indexObsidianSourceEmbeddings` for that unchanged file. That means the PR can stabilize startup by permanently skipping semantic expansion for unchanged Obsidian notes unless some separate backfill repairs it. The skip needs to account for the requested work, e.g. only return early when the artifact is unchanged and the requested graph/embedding outputs already exist, or split artifact indexing from graph/embedding refresh.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `949105ad`

</details>

I found one correctness issue: the new persisted-artifact skip can also skip semantic expansion for artifacts that were previously indexed in the new lightweight mode.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The changed early return in indexNativeMemoryFile is directly visible in native-memory-sources.ts, and this PR also adds startup/dashboard paths that write Obsidian artifacts with sourceGraphEnabled false and no embedding options. I do not have evidence in this diff of another guaranteed backfill path that compensates for that skipped graph/embedding work.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d6dfe84b`

</details>

I did not find a concrete correctness, security, data-integrity, or race-condition issue in the changed paths. The implementation matches the PR description: startup and explicit source syncs are now lighter, progress is shared through the route state, unchanged artifacts/chunks are skipped from persisted rows, wikilink resolution uses a per-scan index, and polling no longer requests trailing rescans while a scan is in flight.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The pasted diff includes the touched daemon startup path, source bridge, route state, graph, embedding, and regression tests, which is enough to validate the main behavior claims. I did not run the daemon or the listed Bun tests in this review environment, so runtime saturation and event-loop behavior remain dependent on the author's live validation.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #650, opened 2026-05-08
feat(connector): add llama.cpp harness with interactive chat runtime

aaf2tbz:
## Summary

Adds a complete llama.cpp integration for Signet, enabling local LLMs served by `llama-server` to use Signet memory and tools through an interactive streaming chat TUI.

## Changes

- Add `@signet/connector-llama-cpp` — install-time connector that writes config, system prompt, and skill files to `~/.config/signet-llama-cpp/`
- Add `@signet/llama-cpp-runtime` — interactive CLI with streaming token output, pulsating spinner during tool execution, and robust JSON argument parsing for small models
- Wire `llama-cpp` as a harness choice in setup wizard, detection (`SetupDetection.harnesses.llamaCpp`), CLI hooks (`configureHarnessHooks`), and build pipeline
- Runtime calls Signet daemon hooks (session start/end, prompt submit) and executes tools via `/api/hooks/tool-execute`
- 9 core Signet tools exposed via OpenAI-compatible tool-calling schema with static fallback (bypasses Streamable HTTP MCP discovery)

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] Other:

## Screenshots

N/A — terminal-based TUI.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent — connector install is idempotent (same as all other connectors).
- [x] Daemon Rust parity reviewed or explicitly N/A — N/A, TypeScript-only connector/runtime.
- [x] Rollback / compatibility note included — purely additive; no existing behavior changed.

## Testing

- [x] `bun run typecheck` passes — `@signet/connector-llama-cpp` and `@signet/llama-cpp-runtime` typecheck clean. Pre-existing `@signet/desktop` error (`@signet/tray` module) unrelated.
- [x] `bun run lint` passes — all changed files pass Biome check.
- [x] `bun run build` passes — full monorepo build succeeds.
- [x] Tested against running daemon — end-to-end verified: session hooks fire, 9 tools loaded, `memory_search` tool executes against daemon API and returns real memories.
- [x] Tested with llama-server — Llama 3.1 8B Q4_K_M correctly invokes tool calls via OpenAI-compatible API.

## AI disclosure

## Notes

Architecture differs from other connectors because llama.cpp is just an inference server with no UI/hooks/plugins. The runtime ships an interactive chat client that bridges llama-server's OpenAI-compatible API to Signet's daemon. Tool schemas use a static fallback since the daemon MCP endpoint uses Streamable HTTP which isn't directly queryable for JSON-RPC schema discovery from a simple HTTP client.


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `98a4da68`.

integrations/llama-cpp/connector/package.json
```diff
@@ -0,0 +1,46 @@
...
+  "type": "module",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": "./dist/index.js",
+      "types": "./dist/index.d.ts"
+    }
+  },
+  "bin": {
```
PR-Reviewer-Ant, line 14: This package advertises `signet-llama-chat` from `../runtime/dist/index.js`, but the build pipeline added in this PR only runs `@signet/connector-llama-cpp`'s `build` script, and that script does not build `../runtime`. A normal `bun run build`/CLI package build can therefore produce a connector whose bin target does not exist. That directly breaks the PR's main claim that installing the llama.cpp connector provides an interactive chat runtime. Either make the connector build include the runtime, add the runtime package to the workspace build/publish path, or stop pointing the connector bin at an artifact it does not produce.

integrations/llama-cpp/runtime/src/signet-tools.ts
```diff
@@ -0,0 +1,249 @@
...
+		return JSON.stringify(data, null, 2);
+	} catch (e) {
+		const msg = e instanceof Error ? e.message : String(e);
+		return `Error executing ${toolName}: ${msg}`;
+	}
+}
+
+interface ToolApiMapping {
+	method: string;
+	buildUrl: (baseUrl: string, args: Record<string, unknown>) => string;
```
PR-Reviewer-Ant, line 243: `memory_get` interpolates the model-supplied `id` directly into the URL path. Tool arguments are untrusted here: a malformed id containing `/`, `?`, or `#` changes the request path/query sent to the daemon instead of being treated as an opaque memory id. Use `encodeURIComponent(String(args.id ?? ""))` after validating the id is present.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6d5f6909`

</details>

I found a packaging/build issue that keeps the advertised chat runtime from actually being built and exposed through the connector package. There are also a couple of smaller correctness gaps in the tool bridge, but the missing runtime binary is the one I would fix before this lands.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description says the connector installs a runtime chat client and exposes signet-llama-chat, but the root and CLI build pipelines only run @signet/connector-llama-cpp's build script. That script builds integrations/llama-cpp/connector/src/index.ts only; it does not run the runtime build even though the connector bin points at ../runtime/dist/index.js.

integrations/llama-cpp/connector/package.json
```diff
@@ -0,0 +1,44 @@
...
```
PR-Reviewer-Ant, line 14: This still leaves the advertised `signet-llama-chat` binary outside the connector package boundary. Building `../runtime` first fixes the monorepo build, but npm package contents are rooted at this package; `files: ["../runtime/dist"]` and a `bin` target of `../runtime/dist/index.js` are not a reliable published artifact. Users installing `@signet/connector-llama-cpp` can still end up with a bin entry pointing outside the packed package. The connector needs to copy or build the runtime into its own `dist`/package tree, or publish/install the runtime package as the owner of the bin.

integrations/llama-cpp/runtime/src/signet-tools.ts
```diff
@@ -0,0 +1,252 @@
...
```
PR-Reviewer-Ant, line 243: The PR description says tool calls execute through `/api/hooks/tool-execute`, but this bridge dispatches model-selected tools directly to daemon resource endpoints like `/api/memory/remember`, `/api/memory/forget`, `/api/knowledge/expand`, and `/api/secrets`. That bypasses the hook execution path that should be the compatibility/policy boundary for harness tool calls, and it means the implementation does not match the advertised runtime behavior. It also makes the static tool map a second source of truth for daemon routes and permissions.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `38797af2`

</details>

I found two issues worth fixing before this lands: the published connector still will not reliably contain the advertised chat binary, and the runtime does not use the hook tool-execution path the PR says it uses.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The connector package still points its bin at a sibling package path and tries to include ../runtime/dist from package files, which is directly visible in integrations/llama-cpp/connector/package.json. The runtime tool bridge also maps model tool calls directly to /api/memory, /api/knowledge, and /api/secrets routes in integrations/llama-cpp/runtime/src/signet-tools.ts instead of the /api/hooks/tool-execute endpoint described by the PR.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

integrations/llama-cpp/runtime/src/chat.ts
```diff
@@ -0,0 +1,434 @@
...
+		daemonUrl: config.daemonUrl,
+		transcript: [],
+	};
+
+	const injectedContext = await onSessionStart(ctx);
+	const systemContent = buildSystemPrompt(config.systemPrompt, injectedContext);
+	const messages: ChatMessage[] = [{ role: "system", content: systemContent }];
+
+	let tools: OpenAiTool[] = [];
+	try {
```
PR-Reviewer-Ant, line 72: This `close` handler exits immediately, so EOF/Ctrl-D closes the chat without calling `onSessionEnd(ctx)`. That means the transcript accumulated in this runtime is never submitted for session-end extraction unless the user exits via `/quit`, SIGINT, or SIGTERM. For an interactive terminal runtime whose main value is Signet memory lifecycle hooks, normal terminal close needs to run the same shutdown path.

integrations/llama-cpp/runtime/src/signet-tools.ts
```diff
@@ -0,0 +1,261 @@
...
```
PR-Reviewer-Ant, line 243: The PR description says tool execution goes through `/api/hooks/tool-execute`, but this implementation still maps model-selected tools directly to daemon resource routes like `/api/hooks/recall`, `/api/memory/modify`, `/api/knowledge/expand`, and `/api/secrets`. That is not just wording drift: it bypasses the single hook execution path that should own harness context, validation, auditing, and whatever policy the daemon applies to model-initiated tool calls. Either route these through `/api/hooks/tool-execute` as claimed or narrow the PR description and justify why direct endpoint access is safe for this runtime.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `98a4da68`

</details>

I found two correctness issues in the runtime: normal terminal close skips the session-end hook, and tool calls still bypass the hook execution endpoint the PR says they use.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both issues are directly visible in the changed runtime files: chat.ts exits from the readline close handler without calling onSessionEnd, and signet-tools.ts dispatches tools through TOOL_API_MAP directly to memory/knowledge/secret endpoints rather than /api/hooks/tool-execute.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

NicholaiVogel:
I like the direction here, llama.cpp/OpenAI-compatible local model support is worth having, but I think this PR puts too much of the shape in the connector/runtime instead of the daemon. If Signet is going to support stateful local-model harnesses, the daemon should expose the canonical stateful API: lifecycle hooks, context injection, transcript/session handling, tool catalog, and tool execution. Then llama.cpp can be a thin adapter over that daemon-owned contract. Right now this adds a second chat runtime plus a static tool registry and hardcoded endpoint router inside `integrations/llama-cpp/runtime`, which feels like it will drift from the daemon-owned memory/tooling contracts immediately. I’d rather land the daemon/runtime boundary first, then have this PR consume that shape instead of defining a parallel one.

PR #683, opened 2026-05-11
fix(docker): verify public ghcr pulls after publish

NicholaiVogel:
## Summary
- Add a post-publish Docker release guard that verifies `ghcr.io/signet-ai/signet:latest` is anonymously pullable.
- Document the public GHCR image and the requirement that the package remain public.
- Add regression coverage so the Docker release workflow keeps the public-pull check.

## Root cause
Docker images are being pushed successfully, but anonymous GHCR pulls currently return `unauthorized`. That points to GHCR package visibility/access drift, not a build/push failure. The current GitHub token available here does not include package admin scopes, so the package visibility still needs to be flipped in GHCR settings.

## Testing
- `bun test tests/integration/docker-build-regression.test.ts`
- `bunx --bun biome format --write .github/workflows/docker-image.yml tests/integration/docker-build-regression.test.ts deploy/docker/README.md`
- `bunx --bun biome check .github/workflows/docker-image.yml tests/integration/docker-build-regression.test.ts deploy/docker/README.md`
- `git diff --check`

## PR Readiness (MANDATORY)

## Migration Notes

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -55,17 +55,16 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
-      - name: Verify public GHCR pull
+      - name: Verify public GHCR latest pull
+        if: ${{ !contains(github.ref_name, '-') }}
         run: |
           set -euo pipefail
           tmp_config="$(mktemp -d)"
           trap 'rm -rf "${tmp_config}"' EXIT
           if ! DOCKER_CONFIG="${tmp_config}" docker manifest inspect ghcr.io/signet-ai/signet:latest >/tmp/signet-public-manifest.json; then
-            cat <<'EOF' >&2
-::error::Published ghcr.io/signet-ai/signet:latest is not publicly pullable.
-The Docker image was pushed, but anonymous GHCR pulls cannot retrieve it.
-Set the GHCR package visibility to public and keep repository inheritance linked:
-https://github.com/orgs/Signet-AI/packages/container/package/signet/settings
-EOF
+            echo "::error::Published ghcr.io/signet-ai/signet:latest is not publicly pullable." >&2
+            echo "The stable Docker image was pushed, but anonymous GHCR pulls cannot retrieve it." >&2
+            echo "Set the GHCR package visibility to public and keep repository inheritance linked:" >&2
+            echo "https://github.com/orgs/Signet-AI/packages/container/package/signet/settings" >&2
             exit 1
           fi
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -54,3 +54,18 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      - name: Verify public GHCR pull
+        run: |
+          set -euo pipefail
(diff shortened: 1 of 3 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `e4535018`.

.github/workflows/docker-image.yml
```diff
@@ -54,3 +54,18 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      - name: Verify public GHCR pull
+        run: |
+          set -euo pipefail
+          tmp_config="$(mktemp -d)"
+          trap 'rm -rf "${tmp_config}"' EXIT
```
PR-Reviewer-Ant, line 63: This check does not verify the image that was just published for prerelease tags. The workflow runs on every `v*` tag, but the metadata step only creates `latest` when `github.ref_name` does not contain `-`. For a tag like `v1.2.3-beta.1`, CI would inspect whatever existing `latest` happens to be, so it can pass even if the newly-pushed prerelease tag is not anonymously pullable, or fail because of an unrelated `latest` state. Either gate this step to the same stable-release condition as the `latest` tag, or inspect one of `steps.meta.outputs.tags` that is guaranteed to have been pushed in this run.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6700d268`

</details>

I found one release-workflow correctness issue: the new anonymous-pull check is hardcoded to `latest`, but this workflow also runs for prerelease tags that do not publish `latest`.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The workflow trigger matches all `v*` tags, while docker/metadata-action only emits the `latest` tag when `!contains(github.ref_name, '-')`. The new verification step always inspects `ghcr.io/signet-ai/signet:latest`, so the mismatch is directly visible in the changed workflow.

NicholaiVogel:
Addressed the release-workflow feedback in e4535018f.

The public GHCR guard is now scoped to the same stable-tag condition that publishes `:latest` (`if: ${{ !contains(github.ref_name, '-') }}`), so prerelease tags no longer inspect unrelated/stale `:latest` state. I also updated the regression guard to assert that stable-only condition.

Local validation:
- `bunx --bun biome check .github/workflows/docker-image.yml tests/integration/docker-build-regression.test.ts`
- `git diff --check`
- `bun test tests/integration/docker-build-regression.test.ts` (10 pass / 0 fail)
- YAML parse of `.github/workflows/docker-image.yml`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e4535018`

</details>

The prior prerelease-tag issue has been addressed: the anonymous `latest` pull check is now gated to the same stable-tag condition that publishes `latest`. I did not find any new correctness, security, or data-integrity issues in the updated diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed workflow now adds `if: ${{ !contains(github.ref_name, '-') }}` to the GHCR `latest` verification step, matching the existing docker/metadata-action conditions that only publish semver and `latest` tags for stable refs. The regression test explicitly checks that guard and the anonymous Docker config path.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #685, opened 2026-05-12
fix(install): persist global bin paths

### LINKED ISSUE #686: macOS daemon does not persist/recover across session lifecycle
## Bug Description
A Discord user on macOS reported repeated Signet reinstall/repair cycles because Signet keeps disappearing from PATH/alignment, shutting itself down, or getting "booted" by the PC. Memory/tooling works while the daemon is alive, then Signet dies or becomes unreachable.

## Evidence / current suspicion
The installer PATH piece is being addressed separately in PR #685, and that does cover macOS because the macOS curl installer uses `web/marketing/public/install.sh`.

However, the daemon-lifecycle part appears separate:

- `surfaces/cli/src/lib/runtime.ts` writes the macOS launchd plist to `~/.agents/.daemon/ai.signet.daemon.plist`.
- `startDaemon()` bootstraps that plist directly with `launchctl bootstrap gui/$UID <plist>`.
- The plist has `RunAtLoad=true` but `KeepAlive=false`.
- The plist is not installed/copied into `~/Library/LaunchAgents`, so it may not survive logout/reboot as a persistent user agent in the way users expect.
- If the daemon exits/crashes, launchd is explicitly told not to restart it.

That matches the user description: Signet works when active, then after machine/session lifecycle or daemon exit it simply disappears/dies and users ask agents to realign PATH/startup again.

## Expected Behavior
For normal macOS installs, Signet should keep the CLI discoverable in future shells and should either:

1. install a durable user LaunchAgent under `~/Library/LaunchAgents/ai.signet.daemon.plist` with appropriate `RunAtLoad`/bounded `KeepAlive`, or
2. make it explicit that `signet daemon start` is session-scoped and provide a separate `signet daemon install-service` / `signet daemon uninstall-service` flow.

## Actual Behavior
The current daemon start path creates and bootstraps a plist under `~/.agents/.daemon`, with `KeepAlive=false`, which is likely a transient/session-owned start rather than durable startup behavior across logout/reboot/crash.

## Suggested Fix Shape
- Add a macOS persistent service install path that writes the plist to `~/Library/LaunchAgents/ai.signet.daemon.plist`.
- Prefer absolute runtime paths, as current code already does via `processRuntimeCommand()`.
- Consider bounded `KeepAlive` / crash restart behavior so the daemon doesn't silently stay dead after a recoverable exit.
- Add regression tests around the generated plist path, `KeepAlive`, and install/uninstall behavior.

## Related
- Discord report from B0R1NG, May 2026: repeated reinstalls/path repairs, daemon dies on macOS.
- PR #685 fixes installer PATH persistence for macOS/Linux/Windows, but does not fully address daemon lifecycle persistence.

NicholaiVogel:
## Summary
- Persist Bun/npm global bin directories during the Unix curl installer instead of only exporting PATH for the current installer process.
- Persist Bun/npm global bin directories to the Windows user PATH during install so new terminals keep finding `signet`.
- Add an installer regression guard covering both install scripts.

## Context
Discord feedback reported repeated Signet reinstalls because `signet` keeps disappearing from PATH/alignment on user machines. The installer could detect Signet in common bin dirs and tell the user to repair PATH, but Windows did not persist PATH at all and the Unix Bun path / fallback-npm paths were only partially persistent.

## Testing
- `bun test tests/integration/installer-path-regression.test.ts`
- `bash -n web/marketing/public/install.sh`
- `git diff --check`
- `bunx --bun biome check tests/integration/installer-path-regression.test.ts`

## PR Readiness (MANDATORY)

## Migration Notes

## Follow-up
- Escaped Unix shell PATH persistence to avoid persisting npm-controlled shell syntax into startup files.
- Fixed the macOS daemon lifecycle path: Signet now writes the launchd plist to `~/Library/LaunchAgents/ai.signet.daemon.plist`, uses `KeepAlive=true`, bootouts stale launchd state before bootstrap, and bootouts the service before `signet daemon stop` so KeepAlive does not resurrect intentional stops.

Fixes #686

### DIFF
--- a/web/marketing/public/install.sh
+++ b/web/marketing/public/install.sh
@@ -17,6 +17,47 @@ warn()  { printf "  ${YELLOW}⚠ %b${RESET}\n" "$1"; }
 error() { printf "  ${RED}✗ %b${RESET}\n" "$1"; }
 ok()    { printf "  ${GREEN}✓ %b${RESET}\n" "$1"; }
 
+# Persist a bin directory for future shells when an installer places tools there.
+persist_path_dir() {
+  local bin_dir="$1"
+  [ -n "$bin_dir" ] || return 0
+  case ":$PATH:" in
+    *":$bin_dir:"*) ;;
+    *) export PATH="$bin_dir:$PATH" ;;
+  esac
+
+  local shell_rc=""
+  case "$(basename "${SHELL:-bash}")" in
+    zsh)  shell_rc="$HOME/.zshrc" ;;
+    fish) shell_rc="$HOME/.config/fish/config.fish" ;;
+    *)    shell_rc="$HOME/.bashrc" ;;
+  esac
+
+  if [ -z "$shell_rc" ]; then
+    return 0
+  fi
+
+  mkdir -p "$(dirname "$shell_rc")"
+  if ! grep -Fq "$bin_dir" "$shell_rc" 2>/dev/null; then
+    if [ "$(basename "$shell_rc")" = "config.fish" ]; then
+      printf '\nfish_add_path %s\n' "$bin_dir" >> "$shell_rc"
+    else
+      printf '\nexport PATH="%s:$PATH"\n' "$bin_dir" >> "$shell_rc"
+    fi
+    info "${DIM}Added $bin_dir to $shell_rc${RESET}"
+  fi
+}
+
+npm_global_bin() {
+  local prefix
+  prefix="$(npm config get prefix 2>/dev/null || echo "")"
+  [ -n "$prefix" ] || return 0
+  case "$prefix" in
+    */bin) printf '%s\n' "$prefix" ;;
+    *)     printf '%s/bin\n' "$prefix" ;;
+  esac
+}
+
 # --- Banner ---
 echo ""
 echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
@@ -57,6 +98,8 @@ if command -v bun >/dev/null 2>&1; then
   BUN_VERSION="$(bun --version 2>/dev/null || echo "unknown")"
   ok "Bun: v$BUN_VERSION"
   HAS_BUN=true
+  BUN_BIN="$(dirname "$(command -v bun)")"
+  persist_path_dir "$BUN_BIN"
 fi
 
 # --- Check for node 18+ ---
@@ -95,7 +138,7 @@ if [ "$HAS_BUN" = false ]; then
   # Source bun env so it's available in this session
   BUN_INSTALL="${BUN_INSTALL:-$HOME/.bun}"
   if [ -f "$BUN_INSTALL/bin/bun" ]; then
-    export PATH="$BUN_INSTALL/bin:$PATH"
+    persist_path_dir "$BUN_INSTALL/bin"
   fi
 
   if command -v bun >/dev/null 2>&1; then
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `018b9e0b`.

web/marketing/public/install.sh
```diff
@@ -17,6 +17,47 @@ warn()  { printf "  ${YELLOW}⚠ %b${RESET}\n" "$1"; }
...
+    *)    shell_rc="$HOME/.bashrc" ;;
+  esac
+
+  if [ -z "$shell_rc" ]; then
+    return 0
+  fi
+
+  mkdir -p "$(dirname "$shell_rc")"
+  if ! grep -Fq "$bin_dir" "$shell_rc" 2>/dev/null; then
+    if [ "$(basename "$shell_rc")" = "config.fish" ]; then
```
PR-Reviewer-Ant, line 42: `bin_dir` is written into shell startup files without shell escaping. For npm fallback paths this value comes from `npm config get prefix`, which can be controlled by npm config in the current environment. A prefix containing shell syntax, quotes, command substitutions, or fish metacharacters would be persisted into `.bashrc`/`.zshrc`/`config.fish` and executed when the user opens a future shell. Please escape for the target shell before writing, or only persist known safe installer-owned directories.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b4543544`

</details>

I found a blocking issue in the Unix installer: it persists npm-derived paths into shell startup files without escaping them, which turns a user-controlled npm prefix into shell code executed by future terminals.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The vulnerable write is directly in web/marketing/public/install.sh. npm_global_bin reads `npm config get prefix`, then persist_path_dir writes that value into .bashrc/.zshrc with double quotes or into fish config unescaped. npm config can be influenced by user config or project-local .npmrc in the installer cwd, so this diff creates a persistent shell-injection path.

NicholaiVogel:
Handled the review feedback and the larger macOS daemon lifecycle issue.

Changes pushed in `018b9e0bb`:
- Escapes Unix shell PATH persistence before writing to `.bashrc` / `.zshrc` / fish config, so npm-controlled prefixes cannot become startup-file shell injection.
- Moves the macOS LaunchAgent plist to `~/Library/LaunchAgents/ai.signet.daemon.plist` instead of `~/.agents/.daemon/...`.
- Sets launchd `KeepAlive` to true so the daemon recovers if it gets killed/exits unexpectedly.
- Runs `launchctl bootout` before bootstrap to clear stale loaded state, then bootstraps the persistent plist.
- Runs `launchctl bootout` during `signet daemon stop` so an intentional stop does not get immediately resurrected by KeepAlive.
- Added/updated regression coverage in `runtime.test.ts` and installer path guard tests.

Validation:
- `bun run build:core`
- `bunx --bun biome check web/marketing/public/install.sh web/marketing/public/install.ps1 surfaces/cli/src/lib/runtime.ts surfaces/cli/src/lib/runtime.test.ts tests/integration/installer-path-regression.test.ts`
- `bun test surfaces/cli/src/lib/runtime.test.ts tests/integration/installer-path-regression.test.ts` (15 pass / 0 fail)
- `bash -n web/marketing/public/install.sh`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `018b9e0b`

</details>

The previous Unix shell-injection concern appears addressed: npm-derived paths are now shell-escaped before being written to startup files, and fish uses `string escape`. I did not find a new correctness, security, or lifecycle issue worth flagging in the incremental launchd changes or installer PATH persistence changes.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The relevant installer writes are visible in web/marketing/public/install.sh and now escape `bin_dir` before persistence; the launchd command construction and plist path changes are covered by the changed runtime tests. I did not run a macOS launchd smoke test, so confidence is medium for the daemon lifecycle behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #692, opened 2026-05-12
fix(secrets): bound and queue secret exec

NicholaiVogel:
## Summary
- Raises bounded secret exec runtime limits while keeping all subprocesses daemon-owned and redacted.
- Makes `POST /api/secrets/exec` and legacy single-secret exec **always queued**: callers receive `202` + job id immediately and poll `/api/secrets/exec/:jobId` for results.
- Removes the opt-in async footgun from CLI/MCP/SDK/API docs so large jobs like SSH/rsync cannot accidentally hold the daemon request path open.
- Bounds queued secret exec execution with a small worker pool + pending queue cap; saturated queues return `429`.
- Hardens timeout/redaction behavior: timeout cleanup targets the process group where supported, and output is redacted before truncation so partial secret prefixes cannot leak.
- Validates secret exec input maps on both multi-secret and legacy single-secret routes so neither can be used as a generic no-secret command runner.
- Validates queued command payloads as non-blank strings before returning 202, so malformed commands fail at the API boundary.
- Evicts old completed/failed retained secret exec results when accepting new work, while preserving the intended 64-item pending queue plus running workers instead of treating retained results as the hard admission cap.
- Adds an MCP-native `secret_exec_status` tool so MCP callers can poll queued secret exec jobs and retrieve redacted results.
- Keeps CLI, MCP, SDK types, plugin manifest, and docs aligned with queued-only semantics.

## Test Plan
- [x] `bunx biome check platform/daemon/src/secrets.ts platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.ts platform/daemon/src/routes/secrets-routes.test.ts platform/daemon/src/mcp/tools.ts surfaces/cli/src/commands/secret.ts platform/daemon/src/plugins/bundled/secrets.ts libs/sdk/src/index.ts libs/sdk/src/types.ts`
- [x] `bun test platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.test.ts surfaces/cli/src/lib/daemon.test.ts` — `31 pass`, `0 fail`, `101 expect() calls`
- [x] `git diff --check`
- [x] `bun run --filter @signet/daemon build`
- [x] `bun run --filter @signet/cli build`
- [x] `bun run --filter @signet/sdk build`

## PR Readiness (MANDATORY)

## Migration Notes


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `4d71dfec`.

platform/daemon/src/secrets.ts
```diff
@@ -391,36 +426,84 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
...
+				else stderrTruncated = true;
+				return [current, bytes + chunk.length];
+			}
+			const remaining = maxOutputBytes - bytes;
+			const slice = chunk.length > remaining ? chunk.subarray(0, remaining) : chunk;
+			if (chunk.length > remaining) {
+				if (stream === "stdout") stdoutTruncated = true;
+				else stderrTruncated = true;
+			}
+			return [current + slice.toString(), bytes + chunk.length];
```
PR-Reviewer-Ant, line 521: This truncates stdout/stderr before redaction, which can leak a partial secret. If a command writes padding followed by a secret and the output limit cuts through the secret, `redact(stdout)` will not match the full resolved value anymore, so the returned output can contain a prefix of the secret. Redaction needs to happen before truncation, or truncation needs to avoid returning partial windows that may contain secret substrings.

platform/daemon/src/secrets.ts
```diff
@@ -431,6 +514,67 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
...
+		status: "queued",
+		createdAt: new Date().toISOString(),
+		timeoutMs,
+	};
+	secretExecJobs.set(job.id, job);
+
+	void (async () => {
+		job.status = "running";
+		job.startedAt = new Date().toISOString();
+		try {
```
PR-Reviewer-Ant: The async path starts execution immediately for every request; there is no queue worker, concurrency limit, or cap on outstanding jobs. That diverges from the PR title/description's bounded queue behavior and opens an admin-facing DoS path where many 30-minute secret exec subprocesses can be launched in parallel. Please add an actual bounded queue or at least enforce a small max running/pending job limit before returning 202.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `8df0dc33`

</details>

I found two issues that undercut the safety goals of this PR: truncated output can leak secret fragments, and async exec is not actually queued or concurrency-bounded.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The truncation and async job behavior are both visible in platform/daemon/src/secrets.ts. The diff includes the route and tests needed to verify intended behavior, and these issues follow directly from the new implementation.

platform/daemon/src/secrets.ts
```diff
@@ -391,36 +426,84 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
...
+		let stdoutBytes = 0;
+		let stderrBytes = 0;
+		let stdoutTruncated = false;
+		let stderrTruncated = false;
+		let settled = false;
+		let timedOut = false;
+
+		const timer = setTimeout(() => {
+			timedOut = true;
+			proc.kill("SIGTERM");
```
PR-Reviewer-Ant: The timeout only kills the direct child process. Any command that starts its own child process can leave that child running after Signet reports `code: 124`, and that child inherits the injected secret environment. That breaks the PR's bounded-execution guarantee and is especially risky for secret exec because timed-out work can keep running with credentials. This should terminate the process tree/process group, or otherwise prevent child processes from surviving the timeout.

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -228,19 +238,29 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
-			const result = await execWithSecrets(body.command, body.secrets);
-			logger.info("secrets", "exec_with_secrets completed", {
+			const timeoutMs = normalizeSecretExecTimeoutMs(body.timeoutMs);
+			const job = startSecretExecJob(body.command, body.secrets, { timeoutMs });
+			logger.info("secrets", "exec_with_secrets queued", {
+				jobId: job.id,
 				secretCount: Object.keys(body.secrets).length,
-				code: result.code,
+				timeoutMs,
 			});
```
PR-Reviewer-Ant, line 255: The implementation always queues and returns `202`, but the PR description says async jobs are requested with `async: true`. That is a breaking behavior change for existing HTTP/SDK/MCP/CLI callers that previously expected stdout/stderr/code from `POST /api/secrets/exec`, and it is not just an additive async option. Either preserve the synchronous default and gate queued behavior on `async: true`, or update the PR framing and downstream surfaces/tests to treat this as an intentional breaking API change.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `99da9005`

</details>

I found one timeout bug that undercuts the security/resource-bound promise, plus one API contract drift worth resolving before this lands.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The timeout path is visible in platform/daemon/src/secrets.ts and only signals the direct child process. The route behavior change is visible in platform/daemon/src/routes/secrets-routes.ts, where the endpoint now always returns a queued job.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a4b36508`

</details>

I found one correctness/security boundary issue: the multi-secret exec route still accepts an empty secrets map, so it can be used as a queued generic command runner despite the API contract saying at least one secret is required.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed route only rejects a missing secrets value, not an empty object, then passes the map into startSecretExecJob. The PR description and docs both explicitly say the secrets map must contain at least one entry, and the MCP path already enforces that invariant.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -228,19 +239,29 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
 				return c.json({ error: "secrets map is required" }, 400);
```
PR-Reviewer-Ant: This still allows `{ "secrets": {} }` through to `startSecretExecJob`, which means `/api/secrets/exec` becomes a queued arbitrary command runner with no secret injection involved. That contradicts the documented contract that the map must contain at least one entry, and it also widens the endpoint's practical behavior from "execute with secrets" to "execute anything later". Please reject empty maps here, matching the MCP tool's existing `Object.keys(secrets).length === 0` check.

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -249,24 +277,22 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
 
 			const secretRefs: Record<string, string> = body.secrets ?? { [name]: name };
-
-			const result = await execWithSecrets(body.command, secretRefs);
-			logger.info("secrets", "exec_with_secrets completed", {
-				name,
-				code: result.code,
-			});
-			return c.json(result);
+			const timeoutMs = normalizeSecretExecTimeoutMs(body.timeoutMs);
```
PR-Reviewer-Ant, line 297: The legacy single-secret exec route still accepts `body.secrets` without validating it. Because `body.secrets ?? { [name]: name }` treats `{}` as intentional input, a caller can POST `{ "command": "...", "secrets": {} }` to `/api/secrets/:name/exec` and queue a daemon-owned command with no secret injection at all. That bypasses the new guard on `/api/secrets/exec` and contradicts the PR's claim that secret exec input maps can no longer make this a generic no-secret command runner. Please either reject provided maps with the same non-empty string validation here, or ignore caller-provided maps on the legacy route and always use `{ [name]: name }`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f52cc2a9`

</details>

I found one security regression in the legacy secret exec route: it can still be used as a queued command runner without injecting any secret. The new queued path mostly matches the PR description, but this leaves the old compatibility endpoint outside the new input-boundary hardening.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The multi-secret route now rejects empty and malformed secret maps, and the tests cover that path. The legacy route still accepts body.secrets directly with no equivalent validation, so a caller can POST an empty map to /api/secrets/:name/exec and queue an arbitrary command with no resolved secret refs.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/secrets.ts
```diff
@@ -431,6 +574,94 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
 	});
 }
 
+export function startSecretExecJob(
+	command: string,
+	secretRefs: Record<string, string>,
+	options: SecretExecOptions = {},
+): SecretExecJob {
```
PR-Reviewer-Ant, line 582: This only caps pending jobs, not retained jobs. A caller can submit many short secret exec jobs that complete immediately, never hit the pending cap, and each completed job can keep up to roughly 2 MB of redacted stdout/stderr in `secretExecJobs` for an hour. That means the new "bounded queue" can still grow memory without a total retained-job/output budget. Please add a cap over retained jobs/results too, or evict completed jobs aggressively when accepting new work.

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -249,24 +277,30 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
-
-			const result = await execWithSecrets(body.command, secretRefs);
-			logger.info("secrets", "exec_with_secrets completed", {
-				name,
-				code: result.code,
-			});
-			return c.json(result);
+			const secretRefs: Record<string, string> = body.secrets === undefined ? { [name]: name } : body.secrets;
+			if (
+				typeof secretRefs !== "object" ||
```
PR-Reviewer-Ant, line 290: The legacy route still treats `secrets: null` as an object and then reaches `Object.keys(secretRefs)`, which throws and returns a 500 instead of the intended 400 validation error. The multi-secret route has a `!secrets` guard; this path should do the same before object key/value inspection.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3b086373`

</details>

I found one queue-bounding issue that undercuts the main safety claim: fast completed jobs can accumulate unbounded in memory for the full job TTL. I also found a smaller validation miss on the legacy route where `secrets: null` turns a malformed request into a 500.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The queue accounting is directly visible in platform/daemon/src/secrets.ts: startSecretExecJob only checks pendingSecretExecJobs.length, while completed jobs and their captured stdout/stderr remain in secretExecJobs until TTL pruning. I did not run the suite in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/secrets.ts
```diff
@@ -431,6 +575,97 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
...
+	pruneSecretExecJobs();
+	if (
+		secretExecJobs.size >= MAX_SECRET_EXEC_RETAINED_JOBS ||
+		pendingSecretExecJobs.length >= MAX_SECRET_EXEC_QUEUED_JOBS
```
PR-Reviewer-Ant, line 587: This makes completed-job retention behave like a hard global admission cap. After 64 fast jobs complete, `secretExecJobs.size` stays at 64 until the one-hour TTL pruning window, so every new `POST /api/secrets/exec` returns `429` even though there are no running or pending jobs. That contradicts the PR goal of capping retained completed results; the cap should evict old completed/failed records or count only active work for admission, not shut off secret exec for an hour after a small burst.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `95788e70`

</details>

I found one queue-retention bug that can make secret exec stop accepting work after a small burst of completed jobs, which undermines the queued-only behavior this PR is adding.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The queue admission check in platform/daemon/src/secrets.ts directly rejects new jobs when the in-memory job map reaches 64 entries, and completed jobs are retained for a one-hour TTL. The included test even locks in that rejection after 64 completed fast jobs, so this is not an incidental edge case.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/secrets.ts
```diff
@@ -431,6 +575,115 @@ export async function execWithSecrets(command: string, secretRefs: Record<string
...
+	secretRefs: Record<string, string>,
+	options: SecretExecOptions = {},
+): SecretExecJob {
+	pruneSecretExecJobs();
+	evictRetainedSecretExecResults();
+	if (
+		secretExecJobs.size >= MAX_SECRET_EXEC_RETAINED_JOBS ||
+		pendingSecretExecJobs.length >= MAX_SECRET_EXEC_QUEUED_JOBS
+	) {
+		throw new SecretExecQueueFullError();
```
PR-Reviewer-Ant, line 589: This check makes the total retained job map act as the hard queue cap, so the daemon can return `429` before the pending queue is saturated. With the configured worker pool, 4 running jobs plus 60 pending jobs gives `secretExecJobs.size === 64`, so the next request is rejected even though `pendingSecretExecJobs.length` is still below the advertised `MAX_SECRET_EXEC_QUEUED_JOBS` of 64. If the intended contract is a 64-item pending queue plus up to 4 running jobs, this should only count pending/running separately from retained completed jobs.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `736f1df5`

</details>

I found one queue-boundary issue: the implementation can reject new work before the documented pending queue cap is actually reached. The rest of the queued-only route, timeout, and redaction changes are directionally consistent with the PR description from the diff provided.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The queue admission condition is visible in platform/daemon/src/secrets.ts and can be reasoned about directly. I did not run the daemon or stress the queue, so the review is based on the supplied diff rather than a runtime repro.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -219,28 +229,46 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
 			const body = (await c.req.json()) as {
 				command?: string;
 				secrets?: Record<string, string>;
+				timeoutMs?: number;
 			};
 
 			if (!body.command) {
 				return c.json({ error: "command is required" }, 400);
 			}
```
PR-Reviewer-Ant, line 237: Now that this endpoint always returns `202` before execution, `command` needs runtime type validation here, not just a truthiness check. A JSON body like `{ "command": {}, "secrets": { "TOKEN": "TOKEN" } }` is accepted as a queued job, even though the worker later hands a non-string to `execWithSecrets` and the caller sees initial success for an invalid request. The legacy route has the same pattern. This is a contract regression from the old synchronous path where invalid execution input could fail the request directly; please reject non-string/blank commands at admission.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `be6e2301`

</details>

I found one correctness issue in the queued job admission path: malformed commands are accepted as jobs instead of being rejected at the API boundary, which makes the new queued contract report success for requests that cannot run.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The route-level validation in platform/daemon/src/routes/secrets-routes.ts only checks that command is truthy before returning 202, while execution happens later in startSecretExecJob/execWithSecrets. I could prove this from the diff, but I did not run the daemon or tests in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/mcp/tools.ts
```diff
@@ -1319,22 +1319,33 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
...
 		},
-		async ({ command, secrets }) => {
+		async ({ command, secrets, timeoutSeconds }) => {
 			if (Object.keys(secrets).length === 0) {
 				return errorResult("secrets map must contain at least one entry");
 			}
 
+			const timeoutMs = timeoutSeconds ? timeoutSeconds * 1000 : undefined;
+			const requestTimeout = 10_000;
 			const result = await daemonFetch<{
```
PR-Reviewer-Ant, line 1339: This makes the MCP `secret_exec` tool enqueue work and return a job id, but the PR does not add an MCP-native way to poll `/api/secrets/exec/:jobId`. CLI gets `exec-status` and the SDK gets `getSecretExecJob`, but MCP clients that only have the advertised tool surface can no longer retrieve stdout/stderr/code. That breaks the existing MCP contract more than the description implies; please add a `secret_exec_status` tool or have `secret_exec` provide a compatible polling mode within the tool surface.

docs/MCP.md
```diff
@@ -651,7 +651,7 @@ requests to the daemon API with these headers:
 - `x-signet-actor: mcp-server` — identifies the calling actor
 - `x-signet-actor-type: harness` — actor type classification
 
-The default request timeout is 10 seconds, except for `secret_exec` which
+The default request timeout is 10 seconds. `secret_exec` also returns quickly because
```
PR-Reviewer-Ant: This sentence is stale after the implementation change: the code now uses a 10 second daemon request timeout for `secret_exec`, but the docs still say it uses 30 seconds for longer-running commands. Since this page is the contract for MCP tool behavior, it should describe the queued response and polling path accurately.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c0746119`

</details>

I found one functional gap in the MCP migration: the tool can now enqueue secret exec work, but MCP callers do not get a way to retrieve the result through MCP. There is also a stale timeout sentence in the MCP docs.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The daemon, CLI, and SDK add explicit job polling paths, but the MCP diff only changes secret_exec to POST the queued route and return the queued response. I do not see a corresponding MCP status tool in the provided diff, so the result retrieval regression is directly supported by the changed MCP surface, though I cannot inspect the full untruncated tools file from the workspace.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `4d71dfec`

</details>

I did not find a concrete correctness, security, data-integrity, or compatibility issue worth flagging from this diff. The implementation matches the PR description: secret exec is queued, bounded, redacted before retention/truncation, and exposed consistently through HTTP, CLI, MCP, SDK types, plugin metadata, docs, and tests.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed daemon routes validate non-blank commands and non-empty secret maps before queueing, return 202 with job metadata, expose a status route, and convert queue saturation to 429. The queue and exec implementation in platform/daemon/src/secrets.ts includes worker limits, pending caps, timeout normalization, process-group termination where supported, redaction before output truncation, result eviction, and regression coverage in platform/daemon/src/secrets.test.ts and platform/daemon/src/routes/secrets-routes.test.ts. I did not run the test plan locally, so this is based on the supplied diff and changed-file context.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #695, opened 2026-05-12
feat(secrets): add Bitwarden provider

NicholaiVogel:
## Summary

- Adds an opt-in Bitwarden-backed Signet Secrets provider using the official `bw` CLI session model (`bw login` + `bw unlock --raw`).
- Keeps the existing local encrypted `secrets.enc` store as the default/backwards-compatible provider; internal provider metadata stays local so Bitwarden does not become a circular dependency.
- Allows switching active providers at any time (`local` ↔ `bitwarden`), with bare secret names resolving through Bitwarden first when active and falling back to local Signet secrets for compatibility.
- Adds migration support to copy existing local Signet secrets into Bitwarden, defaulting to dry-run and optionally deleting local copies only after successful per-secret migration.
- Wires the integration through daemon API routes, CLI commands, SDK types/client methods, MCP secret-reference wording, plugin manifest capabilities/surfaces, docs, and prompt-context secret listing.
- Hardens setup so invalid/locked Bitwarden sessions are rejected before persistence or activation.
- Clears stale managed-folder metadata on reconnect when no folder is supplied, avoiding writes to an old Bitwarden folder.
- Pipes Bitwarden create/edit item payloads over stdin so secret-bearing encoded JSON does not appear in process argv.
- Canonicalizes active-provider bare-name reads the same way writes/migrations do, preserving compatibility for lowercase names such as `anthropic_key`.
- Reports the actual current provider state when Bitwarden connect validation fails, instead of echoing the requested activation flag.
- Preserves already-valid local secret names as Bitwarden item names, avoiding case-collisions such as `foo` and `FOO` while still normalizing legacy invalid names.

## User-facing flow

```bash
bw login
BW_SESSION=$(bw unlock --raw)
signet secret bitwarden connect "$BW_SESSION" --activate
signet secret bitwarden migrate          # dry run
signet secret bitwarden migrate --write  # copy local Signet secrets into Bitwarden
signet secret bitwarden use local        # switch back any time
```

## Safety / compatibility notes

- Local Signet secrets remain the default until the user explicitly activates Bitwarden.
- Provider credentials/config (`BITWARDEN_SESSION`, active-provider marker, managed folder id) are always stored in local Signet secrets.
- Existing `local://...`, bare names, queued secret exec, and 1Password `op://...` support continue to work.
- Explicit Bitwarden refs are supported as `bw://name/NAME` and `bw://item/ITEM_ID/password`.
- When Bitwarden is active, listing merges Bitwarden item names with existing local names and degrades to local-only if Bitwarden is temporarily unavailable.

## Test plan

- `bunx biome check platform/daemon/src/hooks.ts platform/daemon/src/bitwarden.ts platform/daemon/src/bitwarden.test.ts platform/daemon/src/secrets.ts platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.ts platform/daemon/src/routes/secrets-routes.test.ts platform/daemon/src/mcp/tools.ts platform/daemon/src/plugins/bundled/secrets.ts surfaces/cli/src/commands/secret.ts libs/sdk/src/index.ts libs/sdk/src/types.ts`
- `bun test platform/daemon/src/bitwarden.test.ts platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.test.ts surfaces/cli/src/lib/daemon.test.ts` — 39 pass, 0 fail, 149 expect() calls
- `bun run --filter @signet/daemon build`
- `bun run --filter @signet/cli build`
- `bun run --filter @signet/sdk build`
- `git diff --check`
- `grep -R --line-number -- '--session' platform/daemon/src/bitwarden.ts` returns no matches; Bitwarden session is passed through `BW_SESSION` child env, not process argv.
- `grep -R --line-number 'encoded.trim()]' platform/daemon/src/bitwarden.ts` returns no matches; Bitwarden item payloads containing secret values are piped on stdin, not process argv.

## Readiness checklist

### DIFF
--- a/docs/SECRETS.md
+++ b/docs/SECRETS.md
@@ -144,6 +144,36 @@ Imported values are stored as regular Signet secrets with generated names,
 and `secret_exec` can also reference 1Password secrets directly via
 `op://vault/item/field` when connected.
 
+
+### Bitwarden provider
+
+Bitwarden is opt-in. By default, Signet continues to use the existing local encrypted `secrets.enc` store. After connecting Bitwarden, you can either keep using local Signet secrets or make Bitwarden the active provider for future `put`, `list`, `delete`, config resolution, and secret exec references.
+
+Signet uses the official Bitwarden CLI session model: log in/unlock with `bw`, then hand Signet the short-lived session token.
+
+```bash
+# One-time Bitwarden CLI login/unlock outside Signet
+bw login
+BW_SESSION=$(bw unlock --raw)
+
+# Connect but keep the local Signet store active
+signet secret bitwarden connect "$BW_SESSION"
+
+# Connect and immediately make Bitwarden the active backing store
+signet secret bitwarden connect "$BW_SESSION" --activate
+
+# Switch providers later without losing either store
+signet secret bitwarden use bitwarden
+signet secret bitwarden use local
+
+# Copy existing local Signet secrets into Bitwarden
+signet secret bitwarden migrate          # dry run
+signet secret bitwarden migrate --write  # copy, keep local copies
+signet secret bitwarden migrate --write --delete-local
+```
+
+When Bitwarden is active, Signet stores new secrets as Bitwarden login items and resolves bare `$secret:NAME` references from Bitwarden first, falling back to local Signet secrets for backwards compatibility. Explicit `bw://name/NAME` and `bw://item/ITEM_ID/password` references are also accepted. Internal Signet provider credentials remain in the local encrypted store so connecting Bitwarden does not create a circular dependency.
+
 ### Export / import (planned)
 
 ```bash
@@ -202,7 +232,7 @@ Content-Type: application/json
 ```
 
 The map is `{ env_var_name: secret_reference }` where a reference can be a
-stored Signet secret name or a 1Password `op://...` reference. The daemon:
+stored Signet secret name, a Bitwarden `bw://...` reference, or a 1Password `op://...` reference. The daemon:
 1. Resolves each secret reference to its value
 2. Queues the subprocess with the resolved values in the environment
 3. Enforces a bounded timeout (5 minutes by default, max 30 minutes)
@@ -275,6 +305,12 @@ The full secrets API is documented in [API.md](./API.md#secrets-api). Summary:
 | `/api/secrets/exec` | POST | Queue command with one or more secrets injected |
 | `/api/secrets/exec/:jobId` | GET | Inspect queued secret exec job status |
 | `/api/secrets/:name/exec` | POST | Legacy single-secret queued exec |
+| `/api/secrets/bitwarden/status` | GET | Bitwarden provider status |
+| `/api/secrets/bitwarden/connect` | POST | Connect/save Bitwarden CLI session |
+| `/api/secrets/bitwarden/connect` | DELETE | Disconnect/remove stored Bitwarden session |
+| `/api/secrets/bitwarden/provider` | POST | Switch active provider (`local` or `bitwarden`) |
+| `/api/secrets/bitwarden/folders` | GET | List Bitwarden folders |
+| `/api/secrets/bitwarden/migrate` | POST | Copy local Signet secrets into Bitwarden |
 | `/api/secrets/1password/status` | GET | 1Password integration status |
 | `/api/secrets/1password/connect` | POST | Connect/save service account token |
 | `/api/secrets/1password/connect` | DELETE | Disconnect/remove stored token |
--- a/docs/SECRETS.md
+++ b/docs/SECRETS.md
@@ -152,15 +152,15 @@ Bitwarden is opt-in. By default, Signet continues to use the existing local encr
 Signet uses the official Bitwarden CLI session model: log in/unlock with `bw`, then hand Signet the short-lived session token.
 
 ```bash
-# One-time Bitwarden CLI login/unlock outside Signet
+# One-time Bitwarden CLI login outside Signet
 bw login
-BW_SESSION=$(bw unlock --raw)
 
-# Connect but keep the local Signet store active
-signet secret bitwarden connect "$BW_SESSION"
+# Connect but keep the local Signet store active. The session token is read
+# from stdin so it is not written to shell history or process argv.
+bw unlock --raw | signet secret bitwarden connect --session-stdin
 
 # Connect and immediately make Bitwarden the active backing store
-signet secret bitwarden connect "$BW_SESSION" --activate
+bw unlock --raw | signet secret bitwarden connect --session-stdin --activate
 
 # Switch providers later without losing either store
 signet secret bitwarden use bitwarden
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -355,14 +355,14 @@ function readField(item: BitwardenItemDetails, field: string): string {
 	const normalized = field.toLowerCase();
 	if (normalized === "password") {
 		const value = item.login?.password;
-		if (value) return value;
+		if (typeof value === "string") return value;
 	}
 	if (normalized === "username") {
 		const value = item.login?.username;
-		if (value) return value;
+		if (typeof value === "string") return value;
 	}
 	const custom = item.fields?.find((entry) => entry.name?.toLowerCase() === normalized)?.value;
-	if (custom) return custom;
+	if (typeof custom === "string") return custom;
 	throw new Error(`Bitwarden item '${item.name}' does not contain field '${field}'`);
 }
 
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -180,7 +180,7 @@ export async function getBitwardenStatus(options: {
 		const [status, folders] = await Promise.all([client.status(), client.listFolders()]);
 		return {
 			configured: true,
-			connected: status.status === undefined || status.status === "unlocked",
+			connected: status.status === "unlocked",
 			activeProvider: options.activeProvider,
 			userEmail: status.userEmail,
 			serverUrl: status.serverUrl,
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -6,7 +6,7 @@ export const BITWARDEN_MANAGED_FOLDER_SECRET = "BITWARDEN_MANAGED_FOLDER_ID";
 
 const SECRET_REF_PREFIX = "bw://";
 const DEFAULT_BW_TIMEOUT_MS = 30_000;
-const SECRET_NAME_RE = /^[A-Z_][A-Z0-9_]*$/;
+const SECRET_NAME_RE = /^[A-Za-z_][A-Za-z0-9_]*$/;
 
 export interface BitwardenFolder {
 	readonly id: string;
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -294,6 +294,8 @@ class BwCliClient implements BitwardenClient {
 			notes: item.notes ?? "Managed by Signet secrets",
 		};
 		const encoded = await runBw(["encode"], { input: JSON.stringify(updated), session: this.session });
+		// Bitwarden CLI vault commands declare encodedJson as optional for create/edit and read it from
+		// stdin when omitted. Keep secret-bearing item payloads off argv and write them to stdin instead.
 		if (existing) {
 			return toItemDetails(
 				parseJson(await runBw(["edit", "item", existing.id], { input: encoded.trim(), session: this.session })),
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -430,7 +430,11 @@ function runBw(
 	options: { readonly session: string; readonly input?: string; readonly timeoutMs?: number },
 ): Promise<string> {
 	return new Promise((resolve, reject) => {
-		const proc = spawn("bw", [...args, "--session", options.session], { stdio: "pipe", windowsHide: true });
+		const proc = spawn("bw", [...args], {
+			stdio: "pipe",
+			windowsHide: true,
+			env: { ...process.env, BW_SESSION: options.session },
+		});
 		let stdout = "";
 		let stderr = "";
 		let settled = false;
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -390,14 +390,16 @@ function toItemDetails(value: unknown): BitwardenItemDetails {
 	const fields = Array.isArray(obj.fields)
 		? obj.fields.map(toRecord).map((field) => ({
 				name: readString(field.name),
-				value: readString(field.value),
+				value: readOptionalString(field.value),
 				type: readNumber(field.type),
 			}))
 		: null;
 	return {
 		...summary,
-		login: login ? { username: readString(login.username), password: readString(login.password) } : undefined,
-		notes: readString(obj.notes) ?? null,
+		login: login
+			? { username: readOptionalString(login.username), password: readOptionalString(login.password) }
+			: undefined,
+		notes: readOptionalString(obj.notes) ?? null,
 		fields,
 	};
 }
@@ -413,6 +415,10 @@ function toRecordOrUndefined(value: unknown): Record<string, unknown> | undefine
 
 function readString(value: unknown): string | undefined {
 	return typeof value === "string" && value.length > 0 ? value : undefined;
+}
+
+function readOptionalString(value: unknown): string | undefined {
+	return typeof value === "string" ? value : undefined;
 }
 
 function readNumber(value: unknown): number | undefined {
--- a/platform/daemon/src/bitwarden.ts
+++ b/platform/daemon/src/bitwarden.ts
@@ -1,0 +1,466 @@
+import { spawn } from "node:child_process";
+
+export const BITWARDEN_SESSION_SECRET = "BITWARDEN_SESSION";
+export const BITWARDEN_ACTIVE_PROVIDER_SECRET = "SIGNET_SECRETS_ACTIVE_PROVIDER";
+export const BITWARDEN_MANAGED_FOLDER_SECRET = "BITWARDEN_MANAGED_FOLDER_ID";
+
+const SECRET_REF_PREFIX = "bw://";
+const DEFAULT_BW_TIMEOUT_MS = 30_000;
+const SECRET_NAME_RE = /^[A-Z_][A-Z0-9_]*$/;
+
+export interface BitwardenFolder {
+	readonly id: string;
+	readonly name: string;
+}
+
+export interface BitwardenItemSummary {
+	readonly id: string;
+	readonly name: string;
+	readonly folderId?: string | null;
+}
+
+export interface BitwardenItemDetails extends BitwardenItemSummary {
+	readonly login?: {
+		readonly username?: string | null;
+		readonly password?: string | null;
+	};
+	readonly notes?: string | null;
+	readonly fields?: readonly BitwardenField[] | null;
+}
+
+export interface BitwardenField {
+	readonly name?: string | null;
+	readonly value?: string | null;
+	readonly type?: number | null;
+}
+
+export interface BitwardenClient {
+	status(): Promise<{ readonly status?: string; readonly userEmail?: string; readonly serverUrl?: string }>;
+	listFolders(): Promise<readonly BitwardenFolder[]>;
+	listItems(): Promise<readonly BitwardenItemSummary[]>;
+	getItem(id: string): Promise<BitwardenItemDetails>;
+	putSecret(
+		name: string,
+		value: string,
+		options?: { readonly folderId?: string; readonly overwrite?: boolean },
+	): Promise<BitwardenItemDetails>;
+	deleteSecret(name: string): Promise<boolean>;
+	resolveSecret(ref: string): Promise<string>;
+}
+
+export type BitwardenClientFactory = (session: string) => Promise<BitwardenClient>;
+
+let bitwardenClientFactoryOverride: BitwardenClientFactory | null = null;
+
+export function setBitwardenClientFactoryForTests(factory: BitwardenClientFactory | null): void {
+	bitwardenClientFactoryOverride = factory;
+}
+
+function getBitwardenClientFactory(explicitFactory?: BitwardenClientFactory): BitwardenClientFactory {
+	return explicitFactory ?? bitwardenClientFactoryOverride ?? defaultBitwardenClientFactory;
+}
+
+export interface BitwardenStatus {
+	readonly configured: boolean;
+	readonly connected: boolean;
+	readonly activeProvider: boolean;
+	readonly userEmail?: string;
+	readonly serverUrl?: string;
+	readonly folders?: readonly BitwardenFolder[];
+	readonly error?: string;
+}
+
+export interface BitwardenMigrationSecretResult {
+	readonly name: string;
+	readonly action: "migrated" | "skipped" | "deleted-local";
+	readonly itemId?: string;
+	readonly error?: string;
+}
+
+export interface BitwardenMigrationResult {
+	readonly dryRun: boolean;
+	readonly deleteLocal: boolean;
+	readonly migratedCount: number;
+	readonly skippedCount: number;
+	readonly deletedLocalCount: number;
+	readonly errorCount: number;
+	readonly results: readonly BitwardenMigrationSecretResult[];
+}
+
+export function isBitwardenReference(secretName: string): boolean {
+	return secretName.startsWith(SECRET_REF_PREFIX);
+}
+
+export function isBitwardenActiveProvider(value: string | undefined): boolean {
+	return value?.trim().toLowerCase() === "bitwarden";
+}
+
+export function buildBitwardenManagedSecretName(name: string): string {
+	const trimmed = name.trim();
+	if (SECRET_NAME_RE.test(trimmed)) return trimmed;
+	const candidate = trimmed
+		.replace(/[^A-Za-z0-9_]+/g, "_")
+		.replace(/^_+|_+$/g, "")
+		.toUpperCase();
+	if (SECRET_NAME_RE.test(candidate)) return candidate;
+	return `SECRET_${candidate || "VALUE"}`;
+}
+
+export async function readBitwardenReference(
+	reference: string,
+	session: string,
+	clientFactory?: BitwardenClientFactory,
+): Promise<string> {
+	if (!isBitwardenReference(reference)) {
+		throw new Error(
+			`Invalid Bitwarden reference '${reference}'. Expected bw://item/<id>/password or bw://name/<secretName>`,
+		);
+	}
+	const client = await getBitwardenClientFactory(clientFactory)(session);
+	return client.resolveSecret(reference);
+}
+
+export async function listBitwardenFolders(
+	session: string,
+	clientFactory?: BitwardenClientFactory,
+): Promise<readonly BitwardenFolder[]> {
+	const client = await getBitwardenClientFactory(clientFactory)(session);
+	return client.listFolders();
+}
+
+export async function listBitwardenSecretNames(
+	session: string,
+	clientFactory?: BitwardenClientFactory,
+): Promise<readonly string[]> {
+	const client = await getBitwardenClientFactory(clientFactory)(session);
+	const items = await client.listItems();
+	return items
+		.map((item) => item.name)
+		.filter((name) => name.length > 0)
+		.sort((a, b) => a.localeCompare(b));
+}
+
+export async function putBitwardenSecret(
+	name: string,
+	value: string,
+	session: string,
+	options: {
+		readonly folderId?: string;
+		readonly overwrite?: boolean;
+		readonly clientFactory?: BitwardenClientFactory;
+	} = {},
+): Promise<BitwardenItemDetails> {
+	const client = await getBitwardenClientFactory(options.clientFactory)(session);
+	return client.putSecret(buildBitwardenManagedSecretName(name), value, {
+		folderId: options.folderId,
+		overwrite: options.overwrite,
+	});
+}
+
+export async function deleteBitwardenSecret(
+	name: string,
+	session: string,
+	clientFactory?: BitwardenClientFactory,
+): Promise<boolean> {
+	const client = await getBitwardenClientFactory(clientFactory)(session);
+	return client.deleteSecret(buildBitwardenManagedSecretName(name));
+}
+
+export async function getBitwardenStatus(options: {
+	readonly configured: boolean;
+	readonly activeProvider: boolean;
+	readonly session?: string;
+	readonly clientFactory?: BitwardenClientFactory;
+}): Promise<BitwardenStatus> {
+	if (!options.configured || !options.session) {
+		return { configured: false, connected: false, activeProvider: options.activeProvider };
+	}
+	try {
+		const client = await getBitwardenClientFactory(options.clientFactory)(options.session);
+		const [status, folders] = await Promise.all([client.status(), client.listFolders()]);
+		return {
+			configured: true,
+			connected: status.status === undefined || status.status === "unlocked",
+			activeProvider: options.activeProvider,
+			userEmail: status.userEmail,
+			serverUrl: status.serverUrl,
+			folders,
+		};
+	} catch (error) {
+		return {
+			configured: true,
+			connected: false,
+			activeProvider: options.activeProvider,
+			error: error instanceof Error ? error.message : String(error),
+		};
+	}
+}
+
+export async function migrateLocalSecretsToBitwarden(options: {
+	readonly session: string;
+	readonly localNames: readonly string[];
+	readonly getLocalSecret: (name: string) => Promise<string>;
+	readonly deleteLocalSecret?: (name: string) => boolean;
+	readonly folderId?: string;
+	readonly overwrite?: boolean;
+	readonly dryRun?: boolean;
+	readonly deleteLocal?: boolean;
+	readonly clientFactory?: BitwardenClientFactory;
+}): Promise<BitwardenMigrationResult> {
+	const client = await getBitwardenClientFactory(options.clientFactory)(options.session);
+	const dryRun = options.dryRun === true;
+	const deleteLocal = options.deleteLocal === true;
+	const overwrite = options.overwrite === true;
+	const results: BitwardenMigrationSecretResult[] = [];
+	let migratedCount = 0;
+	let skippedCount = 0;
+	let deletedLocalCount = 0;
+	let errorCount = 0;
+
+	for (const localName of options.localNames) {
+		const name = buildBitwardenManagedSecretName(localName);
+		try {
+			if (dryRun) {
+				results.push({ name, action: "skipped" });
+				skippedCount += 1;
+				continue;
+			}
+			const value = await options.getLocalSecret(localName);
+			const item = await client.putSecret(name, value, { folderId: options.folderId, overwrite });
+			results.push({ name, action: "migrated", itemId: item.id });
+			migratedCount += 1;
+			if (deleteLocal && options.deleteLocalSecret?.(localName)) {
+				results.push({ name: localName, action: "deleted-local" });
+				deletedLocalCount += 1;
+			}
+		} catch (error) {
+			errorCount += 1;
+			results.push({ name, action: "skipped", error: error instanceof Error ? error.message : String(error) });
+		}
+	}
+
+	return { dryRun, deleteLocal, migratedCount, skippedCount, deletedLocalCount, errorCount, results };
+}
+
+export async function defaultBitwardenClientFactory(session: string): Promise<BitwardenClient> {
+	const trimmedSession = session.trim();
+	if (!trimmedSession)
+		throw new Error(
+			"Bitwarden session token is required. Run `bw login` then `bw unlock --raw` and connect the session to Signet.",
+		);
+	return new BwCliClient(trimmedSession);
+}
+
+class BwCliClient implements BitwardenClient {
+	constructor(private readonly session: string) {}
+
+	async status(): Promise<{ readonly status?: string; readonly userEmail?: string; readonly serverUrl?: string }> {
+		return parseJson(await runBw(["status"], { session: this.session })) as {
+			status?: string;
+			userEmail?: string;
+			serverUrl?: string;
+		};
+	}
+
+	async listFolders(): Promise<readonly BitwardenFolder[]> {
+		const folders = parseJson(await runBw(["list", "folders"], { session: this.session })) as unknown[];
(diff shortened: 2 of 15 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `16a20783`.

platform/daemon/src/mcp/tools.ts
```diff
@@ -1305,11 +1306,12 @@ export async function createMcpServer(opts?: McpServerOptions): Promise<McpServe
 	// ------------------------------------------------------------------
 	server.registerTool(
 		"secret_exec",
+		"secret_exec_status",
```
PR-Reviewer-Ant: `registerTool` appears to be called with an extra string argument here. The rest of this file uses the MCP SDK shape as `registerTool(name, config, handler)`, but this now passes `"secret_exec_status"` where the tool config object should be. That should either fail registration or register `secret_exec` with the wrong metadata/handler, which is especially risky because the PR explicitly documents polling with `secret_exec_status`.

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -82,18 +96,135 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
+			const session = configured ? await getSecret(BITWARDEN_SESSION_SECRET) : undefined;
+			return c.json(await getBitwardenStatus({ configured, activeProvider, session }));
+		} catch (e) {
+			const err = e as Error;
+			logger.warn("secrets", "Bitwarden status check failed", { error: err.message });
+			return c.json({ configured: true, connected: false, activeProvider, error: err.message });
+		}
+	});
+
+	app.post("/api/secrets/bitwarden/connect", async (c) => {
```
PR-Reviewer-Ant, line 126: The connect route stores the Bitwarden session, optional folder, and active-provider marker before proving the session is usable, then returns `success: true` even when `getBitwardenStatus` reports `connected: false`. That diverges from the PR/API claim that connect validates the session, and with `activate: true` it can leave users on an active Bitwarden provider backed by an invalid token, breaking future secret writes/resolution until they manually switch back.

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,466 @@
...
+		let stdout = "";
+		let stderr = "";
+		let settled = false;
+		const timer = setTimeout(() => {
+			if (settled) return;
+			settled = true;
+			proc.kill("SIGTERM");
+			reject(new Error(`Bitwarden CLI timed out running bw ${args.join(" ")}`));
+		}, options.timeoutMs ?? DEFAULT_BW_TIMEOUT_MS);
+		timer.unref();
```
PR-Reviewer-Ant, line 455: This passes the raw `BW_SESSION` value as a command-line argument to every `bw` subprocess. On typical systems, argv is visible to same-user process inspection tools while the process is running, so this expands the local exposure surface for the vault session token. Prefer passing the token through the child environment as `BW_SESSION` and omit `--session <token>` from argv.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5956948c`

</details>

I found a few concrete issues that need fixing before this Bitwarden provider is ready: the MCP tool registration appears broken, invalid Bitwarden sessions can still be saved and activated, and the session token is passed to `bw` on the process command line.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are directly visible in the changed code: `registerTool` is called with an extra string argument in `platform/daemon/src/mcp/tools.ts`, Bitwarden connect stores and may activate before checking the returned status in `platform/daemon/src/routes/secrets-routes.ts`, and `runBw` appends the raw session token to spawned process argv in `platform/daemon/src/bitwarden.ts`.

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,466 @@
...
+
+function runBw(
+	args: readonly string[],
+	options: { readonly session: string; readonly input?: string; readonly timeoutMs?: number },
+): Promise<string> {
+	return new Promise((resolve, reject) => {
+		const proc = spawn("bw", [...args, "--session", options.session], { stdio: "pipe", windowsHide: true });
+		let stdout = "";
+		let stderr = "";
+		let settled = false;
```
PR-Reviewer-Ant, line 448: This passes the Bitwarden session token on the `bw` command line. On Unix-like systems, argv is commonly visible through process listings to the same user, so a short-lived vault session can leak while any `bw` command is running. The Bitwarden CLI supports the `BW_SESSION` environment variable; use that instead of appending `--session <token>` to argv, and make sure errors/logs never include it.

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -82,18 +96,135 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
+		const denied = rejectIfCapabilityDenied(c, host, ["secrets:providers:configure"]);
```
PR-Reviewer-Ant, line 127: The connect path stores the session and can activate Bitwarden before checking whether the session actually works. `getBitwardenStatus()` is called after `putSecret()`/`setActiveSecretProvider()`, and the route still returns `success: true` even when `connected` is false. That diverges from the PR/API claim that connect validates the session, and it can leave installs with Bitwarden active but every bare secret read/write falling into a broken provider path. Validate the token first and only persist/activate on a connected status.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `07072c80`

</details>

I found two issues that should be fixed before this lands: the Bitwarden session is exposed via process argv, and connect/provider activation can mark Bitwarden active even when the session is not actually usable.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both issues are directly visible in the changed code. The bw subprocess is spawned with the session token as a command-line argument in platform/daemon/src/bitwarden.ts, and the connect/provider routes persist or activate Bitwarden before proving the session is connected in platform/daemon/src/routes/secrets-routes.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -82,18 +95,144 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
+		if (denied) return denied;
+		try {
+			const body = await readOptionalJsonObject(c);
```
PR-Reviewer-Ant, line 130: This leaves stale provider metadata behind. If a user previously connected with `folderId: "old-folder"` and later reconnects with a different Bitwarden session without `folderId`, this branch does not remove `BITWARDEN_MANAGED_FOLDER_SECRET`. `putSecret` then keeps reading that old folder id and writes new Bitwarden items into the previous folder, which contradicts the new connection request and can fail or put secrets in the wrong vault organization. The reconnect path should clear the stored folder metadata when no folder is supplied, or require callers to pass the existing folder explicitly.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `2f32abe6`

</details>

I found one provider-state bug: reconnecting Bitwarden without a folder does not clear the previously stored folder id, so future writes can keep going to an old folder despite the new connection payload.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The connect route only writes BITWARDEN_MANAGED_FOLDER_SECRET when folderId is present, while putSecret later reads that stored value unconditionally for Bitwarden writes. The stale-state path is directly visible in platform/daemon/src/routes/secrets-routes.ts and platform/daemon/src/secrets.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,470 @@
...
+		return toItemDetails(parseJson(await runBw(["get", "item", id], { session: this.session })));
+	}
+
+	async putSecret(
+		name: string,
+		value: string,
+		options: { readonly folderId?: string; readonly overwrite?: boolean } = {},
+	): Promise<BitwardenItemDetails> {
+		const existing = await this.findItemByName(name);
+		if (existing && options.overwrite !== true) {
```
PR-Reviewer-Ant, line 285: This still leaks secret material through process argv. `encoded` is produced from `updated`, which includes `login.password: value`, and then it is passed as a positional argument to `bw create item` / `bw edit item`. Even though it is base64-ish Bitwarden JSON, local process inspection can recover the payload while the child is running. The PR explicitly calls out that the session is kept out of argv; the same standard needs to apply to the item payload containing the actual secret value. Please route the encoded item through stdin or another non-argv mechanism supported by `bw`, or otherwise avoid placing the secret-bearing payload in the process command line.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3945394e`

</details>

I found a secret-exposure issue in the Bitwarden CLI integration. The provider avoids putting the session token in argv, but it still passes encoded item payloads containing secret values as command-line arguments.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new Bitwarden provider constructs an encoded item from JSON containing the password and then passes that encoded payload directly to `bw edit item` / `bw create item` as an argv element. That is visible from platform/daemon/src/bitwarden.ts and directly affects the PR's stated secret-safety behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `85a53ded`

</details>

I found one correctness bug in the Bitwarden provider path: names are normalized on write/migration but not normalized on bare-name resolution, so some secrets become unreachable through the compatibility flow the PR advertises.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The mismatch is directly visible in platform/daemon/src/secrets.ts and platform/daemon/src/bitwarden.ts: writes call putBitwardenSecret, which normalizes via buildBitwardenManagedSecretName, while bare-name reads construct bw://name/<original localName> without applying the same normalization. The existing tests cover reading the normalized name after migration, but not resolving the original local secret name through getSecret.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/secrets.ts
```diff
@@ -299,21 +328,66 @@ export async function getSecret(name: string): Promise<string> {
...
-	return getStoredSecret(parseLocalSecretName(name));
+	if (isBitwardenReference(name)) {
+		const session = await getStoredSecret(BITWARDEN_SESSION_SECRET);
+		return readBitwardenReference(name, session);
+	}
+
+	const localName = parseLocalSecretName(name);
+	if (isInternalSecretName(localName)) {
+		return getStoredSecret(localName);
+	}
```
PR-Reviewer-Ant, line 380: This bare-name Bitwarden lookup does not apply the same normalization used when writing to Bitwarden. `putSecret()` calls `putBitwardenSecret(localName, ...)`, and that stores `buildBitwardenManagedSecretName(name)` in Bitwarden. So a valid local secret like `Anthropic Key` or `1password-token` is migrated/stored as `ANTHROPIC_KEY` or `SECRET_1PASSWORD_TOKEN`, but later `$secret:Anthropic Key` / bare `Anthropic Key` resolves `bw://name/Anthropic%20Key` first and misses the Bitwarden item. That breaks the PR's compatibility claim that bare names resolve through Bitwarden first. The read path should use the same canonical Bitwarden name, or writes should preserve the original lookup name consistently.

platform/daemon/src/routes/secrets-routes.ts
```diff
@@ -82,18 +95,145 @@ export function registerSecretRoutes(app: Hono, host: PluginHostV1 = getDefaultP
...
+		const denied = rejectIfCapabilityDenied(c, host, ["secrets:providers:list"]);
+		if (denied) return denied;
+		const activeProvider = (await getActiveSecretProvider()) === "bitwarden";
+		try {
+			const configured = hasSecret(BITWARDEN_SESSION_SECRET);
+			const session = configured ? await getSecret(BITWARDEN_SESSION_SECRET) : undefined;
+			return c.json(await getBitwardenStatus({ configured, activeProvider, session }));
+		} catch (e) {
+			const err = e as Error;
+			logger.warn("secrets", "Bitwarden status check failed", { error: err.message });
```
PR-Reviewer-Ant, line 121: This status object uses the requested `activate` flag as `activeProvider` before the session has been persisted or activated. On an invalid session with `{ activate: true }`, the 400 response can include `activeProvider: true` even though the provider was not switched, which undercuts the “rejected before persistence or activation” guarantee and gives SDK/API callers the wrong state. This should report the actual current provider on failure, not the requested target state.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `46e4c014`

</details>

I found one API logic issue in the Bitwarden connect path: failed validation can report Bitwarden as the active provider even though the PR says activation is rejected before persistence.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible in the route diff: the connect handler builds status with activeProvider set from the requested activate flag before validation succeeds, then spreads that status into the 400 response. I did not have a live checkout/runtime to exercise the endpoint.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/secrets.ts
```diff
@@ -286,6 +298,24 @@ export async function putSecret(name: string, value: string): Promise<void> {
...
+	if (isInternalSecretName(localName) || !(await isBitwardenProviderActive())) {
+		await putLocalSecret(localName, value);
+		return;
+	}
+
+	const session = await getStoredSecret(BITWARDEN_SESSION_SECRET);
+	let folderId: string | undefined;
+	try {
+		folderId = await getStoredSecret(BITWARDEN_MANAGED_FOLDER_SECRET);
+	} catch {
```
PR-Reviewer-Ant, line 314: This write path can silently overwrite the wrong Bitwarden item after name canonicalization. The local store preserves distinct names, but `putBitwardenSecret()` canonicalizes names like `anthropic_key`, `Anthropic Key`, and `ANTHROPIC_KEY` to the same Bitwarden item name, and this call always passes `overwrite: true`. That means switching Bitwarden on changes `secret put` from exact-name replacement to normalized-name replacement, which can clobber an existing Bitwarden secret without warning. I think active-provider writes need either collision detection against the original requested name or a clear refusal path when canonicalization maps to an existing different source name.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e0c9be7a`

</details>

I found one data-integrity issue in the active Bitwarden write path: names that are distinct in the existing local store can collapse to the same Bitwarden item and be overwritten silently.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The overwrite behavior is explicit in platform/daemon/src/secrets.ts, and the canonicalization behavior is explicit in platform/daemon/src/bitwarden.ts. I did not run a live bw CLI repro, but the collision is directly implied by the changed code.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a49e8766`

</details>

I found one delete-path bug that can make a successfully deleted Bitwarden secret continue resolving from the local fallback, plus one test assertion that is not actually awaited.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The delete behavior is directly visible in secrets.ts: the active-provider delete returns after deleting the Bitwarden item, while getSecret falls back to the local store for the same bare name. The PR description explicitly claims active-provider delete support and local fallback compatibility, so this interaction is in scope.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/secrets.ts
```diff
@@ -323,6 +401,62 @@ export function deleteSecret(name: string): boolean {
 	return true;
 }
 
+export function deleteSecret(name: string): boolean {
+	return deleteLocalSecret(name);
+}
+
+export async function deleteSecretFromActiveProvider(name: string): Promise<boolean> {
+	const localName = parseLocalSecretName(name);
+	if (!isInternalSecretName(localName) && (await isBitwardenProviderActive())) {
```
PR-Reviewer-Ant: This returns success as soon as the Bitwarden item is deleted, but `getSecret()` will then fall back to the local secret with the same name if one exists. That is easy to hit after the default migration flow, which copies local secrets into Bitwarden while keeping local copies. The user sees DELETE succeed, but the bare secret name still resolves and still appears in the merged list from the local fallback. Either delete the fallback copy too for effective deletion, or make the API/CLI explicit that this only deletes the active-provider copy and provide a way to remove both.

platform/daemon/src/secrets.test.ts
```diff
@@ -226,7 +233,7 @@ describe("local secrets provider", () => {
 		mkdirSync(join(agentsDir, ".secrets"), { recursive: true });
 		writeFileSync(secretsFile(), "not-json", { mode: 0o600 });
 
-		expect(() => listSecrets()).toThrow("Failed to read secrets store");
+		expect(listSecrets()).rejects.toThrow("Failed to read secrets store");
```
PR-Reviewer-Ant: This async rejection assertion is not awaited or returned, so the test can pass without actually checking the corrupt-store failure path after `listSecrets()` became async. Use `await expect(listSecrets()).rejects.toThrow(...)` so the regression test still guards the behavior.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5f75e00b`

</details>

I did not find a concrete correctness, security, data integrity, or breaking-change issue in the changed Bitwarden provider implementation. The PR description lines up with the main implementation paths shown here: opt-in activation, local fallback, stdin payload handling for Bitwarden writes, route/CLI/SDK wiring, and tests for the important provider flows.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the provider implementation, route wiring, CLI/SDK surfaces, docs, and targeted tests, and I checked for stale listSecrets() call sites in the provided workspace. I did not run the Bitwarden CLI against a real vault session, so behavior that depends on the installed bw CLI remains runtime-unverified.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Follow-up pushed in `5f75e00b` for the latest review feedback.\n\nResolved:\n- Active Bitwarden bare-name delete now also removes the same-name local fallback, while explicit `local://NAME` delete remains local-only.\n- Added route regression coverage for migrated/local-fallback delete behavior.\n- Fixed the un-awaited `expect(...).rejects` assertion.\n\nLocal validation after the patch:\n- `bunx biome check` on touched/provider surfaces: pass\n- targeted Bitwarden/secrets/routes/CLI tests: `39 pass, 0 fail, 154 expect() calls`\n- daemon, CLI, SDK builds: pass\n- `git diff --check`: pass\n\nLatest PR-Reviewer-Ant review on `5f75e00b` found no concrete correctness/security/data-integrity/breaking-change issue. All workflow jobs are green; only the GitHub CodeQL aggregate status is still showing queued/pending even though its Analyze jobs completed successfully.

surfaces/cli/src/commands/secret.ts
```diff
@@ -400,6 +400,180 @@ export function registerSecretCommands(program: Command, deps: SecretDeps): void
 				process.exit(1);
 			}
 		});
+
+	const bitwardenCmd = secretCmd.command("bitwarden").alias("bw").description("Manage Bitwarden integration");
+
+	bitwardenCmd
+		.command("connect [session]")
```
PR-Reviewer-Ant: This makes the Bitwarden session token a positional CLI argument, so the recommended `signet secret bitwarden connect "$BW_SESSION"` flow exposes a live vault session through process argv while the command runs, and often through shell history depending on how the user invokes it. That undercuts the PR's stated hardening around keeping Bitwarden session material out of argv. The safer interface is to prompt/read from stdin or an env var and avoid accepting the token as a positional argument.

docs/SECRETS.md
```diff
@@ -144,6 +144,36 @@ Imported values are stored as regular Signet secrets with generated names,
...
+Signet uses the official Bitwarden CLI session model: log in/unlock with `bw`, then hand Signet the short-lived session token.
+
+```bash
+# One-time Bitwarden CLI login/unlock outside Signet
+bw login
+BW_SESSION=$(bw unlock --raw)
+
+# Connect but keep the local Signet store active
+signet secret bitwarden connect "$BW_SESSION"
+
```
PR-Reviewer-Ant, line 161: The documented happy path passes the raw Bitwarden session token as a command-line argument. Even though daemon-to-`bw` calls avoid argv, this public workflow still leaks the token into the Signet CLI process argv and can land in shell history. Please document a stdin/prompt/env-var flow instead of `connect "$BW_SESSION"`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `fb2fcbe6`

</details>

I found one security issue in the new Bitwarden CLI flow: the documented and implemented connect command encourages passing the Bitwarden session token as a process argument, which exposes the token through shell history and process inspection.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR explicitly hardens daemon-to-bw calls so tokens are not passed in argv, but the new CLI command accepts the same session token as a positional argument and the docs show that usage. The exposure is directly visible in surfaces/cli/src/commands/secret.ts and docs/SECRETS.md.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,470 @@
...
+		const folders = parseJson(await runBw(["list", "folders"], { session: this.session })) as unknown[];
+		return folders.map((folder) => toFolder(folder)).filter((folder): folder is BitwardenFolder => folder !== null);
+	}
+
+	async listItems(): Promise<readonly BitwardenItemSummary[]> {
+		const items = parseJson(await runBw(["list", "items"], { session: this.session })) as unknown[];
+		return items.map((item) => toItemSummary(item)).filter((item): item is BitwardenItemSummary => item !== null);
+	}
+
+	async getItem(id: string): Promise<BitwardenItemDetails> {
```
PR-Reviewer-Ant, line 275: `bw edit item` does not consume the encoded item JSON from stdin here; the Bitwarden CLI expects the encoded payload as the positional argument after the item id. With the current call shape, editing an existing secret will fail at runtime even though the test only checks that the secret-bearing payload is not in argv. If argv exposure is the concern, this probably needs a different mitigation than stdin, or a real CLI-backed regression test proving the chosen invocation works.

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,470 @@
...
+		return toItemDetails(parseJson(await runBw(["get", "item", id], { session: this.session })));
+	}
+
```
PR-Reviewer-Ant, line 278: `bw create item` has the same issue: the encoded item payload is required as a command argument, but this call only writes it to stdin. That means new Bitwarden-backed `put` calls and local-to-Bitwarden migrations cannot actually create items with the official CLI despite the PR description claiming writes are wired through.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `236f2fed`

</details>

I found one blocking Bitwarden CLI integration issue: the implementation pipes create/edit payloads to stdin, but the real `bw create item` and `bw edit item` commands require the encoded item payload as a positional argument. The tests assert the source shape but do not exercise the actual CLI contract, so Bitwarden writes and migrations would fail in practice.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed code in platform/daemon/src/bitwarden.ts invokes `bw create item` and `bw edit item` without the required encoded payload argument, while the PR's test plan does not include a real `bw` CLI smoke test. The failure is directly visible in the command construction, but I did not run the Bitwarden CLI in this review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

NicholaiVogel:
Follow-up pushed in `7fb95c19f` for the Bitwarden CLI stdin-payload concern.

I verified against the unpacked `@bitwarden/cli@2026.4.1` package source under `/tmp/bwcli/package/build/bw.js`:

- `VaultProgram.createCommand()` declares `create <object> [encodedJson]` and its help text says: `Encoded json of the object to create. Can also be piped into stdin.`
- `VaultProgram.editCommand()` declares `edit <object> <id> [encodedJson]` and its help text says: `Encoded json of the object to create. Can also be piped into stdin.`
- `CreateCommand.run()` calls `CliUtils.readStdin()` when `requestJson` is empty and `BW_SERVE !== "true"`.
- The help examples include `echo 'eyJuYW...In0K' | bw create folder` and `echo 'eyJuYW...J9Cg==' | bw edit folder <id>`.

So the current implementation intentionally omits the secret-bearing encodedJson positional argument and sends it via stdin:

- `bw create item` + stdin payload
- `bw edit item <id>` + stdin payload

That keeps item payloads/passwords out of argv while matching the real Bitwarden CLI contract. I added an inline comment and tightened the test wording so this contract is explicit in the PR.

Validation after this push:
- `bunx biome check` on Bitwarden/CLI/docs touched surfaces: pass
- targeted Bitwarden/secrets/routes/CLI tests: `39 pass, 0 fail, 154 expect() calls`
- daemon, CLI, SDK builds: pass
- `git diff --check`: pass

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `7fb95c19`

</details>

I found one data-loss edge in the active Bitwarden delete path. The rest of the provider wiring looks consistent with the PR description from the supplied diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The delete behavior is directly visible in platform/daemon/src/secrets.ts: the active Bitwarden branch deletes the Bitwarden item and then unconditionally deletes the same local secret name. That conflicts with the PR's stated compatibility/safety model where local copies remain unless migration is run with deleteLocal.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/secrets.ts
```diff
@@ -323,6 +401,69 @@ export function deleteSecret(name: string): boolean {
...
+	const explicitLocal = name.startsWith("local://");
+	const localName = parseLocalSecretName(name);
+	if (explicitLocal || isInternalSecretName(localName) || !(await isBitwardenProviderActive())) {
+		return deleteLocalSecret(localName);
+	}
+
+	const session = await getStoredSecret(BITWARDEN_SESSION_SECRET);
+	const deletedFromBitwarden = await deleteBitwardenSecret(localName, session);
+	const deletedFromLocal = deleteLocalSecret(localName);
+	if (deletedFromBitwarden) {
```
PR-Reviewer-Ant, line 463: This active-provider delete path can silently remove the local fallback copy too. If a user migrates with the documented default of keeping local copies, then later deletes the Bitwarden item while Bitwarden is active, `deleteLocalSecret(localName)` also destroys the local copy. That undercuts the safety note that local secrets remain unless `deleteLocal` is explicitly requested, and makes provider switching lossy. Please keep active-provider deletes scoped to Bitwarden unless the caller explicitly asks to delete the local fallback as well.

NicholaiVogel:
Follow-up pushed/rebased at `96486017a` for the latest Bitwarden delete/fallback review.

Resolved the data-loss edge without regressing the “delete means not visible” behavior:

- Active-provider bare-name delete now deletes/suppresses the active Bitwarden item.
- The same-name local fallback is preserved on disk; it is **not** removed unless the user explicitly targets `local://NAME` or uses migration with `deleteLocal`.
- While Bitwarden remains active, the preserved local fallback is hidden from implicit bare-name resolution/listing through a local tombstone list, so `getSecret("NAME")` does not appear undeleted after a Bitwarden delete.
- Re-putting the Bitwarden secret clears the tombstone.

Post-rebase local validation:

```text
bunx biome check ... focused Bitwarden/secrets/CLI/SDK/docs files
bun test platform/daemon/src/bitwarden.test.ts platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.test.ts surfaces/cli/src/lib/daemon.test.ts
39 pass, 0 fail, 155 expect() calls
bun run --filter @signet/daemon build
bun run --filter @signet/cli build
bun run --filter @signet/sdk build
git diff --check
```

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `96486017`

</details>

I found a couple of correctness/surface-contract issues, but nothing in the diff that looks like an immediate security blocker. The Bitwarden provider mostly follows the PR description, with some edge behavior and manifest drift worth tightening.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The findings are directly visible in the changed Bitwarden provider and bundled plugin manifest. I did not run the Bitwarden CLI against a real vault session, so CLI-specific behavior beyond the static diff remains unverified.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,472 @@
...
+	return {
+		id: "",
+		name,
+		folderId: folderId ?? null,
+		type: 1,
+		login: { username: "signet", password: "" },
+		notes: "Managed by Signet secrets",
+	};
+}
+
```
PR-Reviewer-Ant, line 342: This treats an empty Bitwarden password as missing because it checks truthiness instead of checking for null/undefined. The local Signet store can persist arbitrary string values, so with Bitwarden active a stored empty string round-trips as an error rather than `""`. Same issue applies to username and custom fields below. If empty values are intentionally invalid, the write path should reject them before creating the item; otherwise these checks should accept empty strings.

platform/daemon/src/plugins/bundled/secrets.ts
```diff
@@ -141,6 +182,21 @@ const surfaces: PluginSurfaceDeclarationsV1 = {
...
 		},
+		{
+			name: "getBitwardenStatus",
+			summary: "Inspect Bitwarden provider status",
+			requiredCapabilities: ["sdk:client", "secrets:providers:list"],
+		},
+		{
+			name: "connectBitwarden",
+			summary: "Connect Bitwarden provider",
+			requiredCapabilities: ["sdk:client", "secrets:providers:configure"],
```
PR-Reviewer-Ant, line 193: The SDK surface manifest only declares `getBitwardenStatus`, `connectBitwarden`, and `migrateSecretsToBitwarden`, but this PR also adds `disconnectBitwarden`, `setSecretProvider`, and `listBitwardenFolders` to the SDK. Since this manifest is the daemon's declared surface/capability inventory, those new client methods are invisible to plugin diagnostics and any tooling that relies on declared surfaces. That diverges from the PR's claim that the integration is wired through plugin manifest capabilities/surfaces.

NicholaiVogel:
Follow-up pushed in `$(git rev-parse --short HEAD)` for the latest review findings.

Resolved:
- Bitwarden field reads now accept empty-string password/username/custom field values instead of treating them as missing.
- Added a regression test for empty Bitwarden secret value round-trip.
- Updated the bundled `signet.secrets` plugin SDK surface manifest to include `disconnectBitwarden`, `setSecretProvider`, and `listBitwardenFolders` alongside the existing status/connect/migrate methods.

Validation:

```text
bunx biome check --write --unsafe platform/daemon/src/bitwarden.ts platform/daemon/src/bitwarden.test.ts platform/daemon/src/plugins/bundled/secrets.ts
bun test platform/daemon/src/bitwarden.test.ts platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.test.ts surfaces/cli/src/lib/daemon.test.ts
40 pass, 0 fail, 156 expect() calls
bun run --filter @signet/daemon build
bun run --filter @signet/cli build
bun run --filter @signet/sdk build
git diff --check
```

NicholaiVogel:
Correction to my previous comment: latest pushed commit is `1b30c544f`.

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,472 @@
...
+	const name = readString(obj.name);
+	if (!id || !name) return null;
+	return { id, name, folderId: readString(obj.folderId) ?? null };
+}
+
+function toItemDetails(value: unknown): BitwardenItemDetails {
+	const summary = toItemSummary(value);
+	if (!summary) throw new Error("Bitwarden returned invalid item payload");
+	const obj = toRecord(value);
+	const login = toRecordOrUndefined(obj.login);
```
PR-Reviewer-Ant, line 389: This helper treats `""` as missing, so `toItemDetails()` converts an empty Bitwarden `login.password` into `undefined`. A secret stored with an empty value can be written, but a later real CLI read via `get item` will fail in `readField()` with “does not contain field 'password'”. The new empty-value test only exercises the mock client and bypasses this parser, so it does not catch the regression. For secret fields, preserve empty strings and only reject non-strings.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `1b30c544`

</details>

I found one correctness bug in the new Bitwarden parser: real CLI reads cannot round-trip empty secret values even though the PR claims and tests that behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new Bitwarden item parser uses readString for login.password, and readString drops empty strings. readField only accepts an actual string, so an empty password returned by the real bw CLI becomes undefined and then throws.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Follow-up pushed in `$(git rev-parse --short HEAD)` for the empty-value parser issue.

Resolved:
- `readField()` already accepted empty strings, but the CLI JSON parser was still normalizing empty login/custom field values to `undefined`.
- Added `readOptionalString()` and use it for parsed Bitwarden login username/password, custom field values, and notes while keeping required IDs/names non-empty.
- Extended the regression coverage to assert the parser path keeps `password: readOptionalString(login.password)` and custom `value: readOptionalString(field.value)`.

Validation:

```text
bunx biome check --write --unsafe platform/daemon/src/bitwarden.ts platform/daemon/src/bitwarden.test.ts
bun test platform/daemon/src/bitwarden.test.ts platform/daemon/src/secrets.test.ts platform/daemon/src/routes/secrets-routes.test.ts surfaces/cli/src/lib/daemon.test.ts
40 pass, 0 fail, 158 expect() calls
bun run --filter @signet/daemon build
bun run --filter @signet/cli build
bun run --filter @signet/sdk build
git diff --check
```

NicholaiVogel:
Correction to previous comment: latest pushed commit is `bab13db26`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `bab13db2`

</details>

I found one documentation/API contract drift issue: the MCP docs now advertise Bitwarden refs, but still describe bare secret names as always local even though this PR changes that behavior when Bitwarden is active.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The implementation in platform/daemon/src/secrets.ts clearly resolves bare names through Bitwarden first when the active provider is Bitwarden, while docs/MCP.md still states bare names are equivalent to local://. I did not run the daemon or Bitwarden CLI, so the review is based on the supplied diff and file contents.


- docs/MCP.md:440 - This paragraph is now misleading. The PR changes bare secret-name resolution so, when Bitwarden is the active provider, bare names resolve through Bitwarden first and only fall back to local. Saying `OPENAI_API_KEY` is equivalent to `local://OPENAI_API_KEY` is no longer generally true for MCP `secret_exec`, and it points agents at the wrong mental model for exactly the flow this PR adds. The SECRETS doc has the updated wording; this section should match it.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Follow-up pushed in `be170a517`: clarified MCP docs so bare secret names are described as active-provider refs, with `local://NAME` for explicit local store, `bw://...` for explicit Bitwarden, and `op://...` for 1Password compatibility. Validation: focused biome on touched code surfaces, focused Bitwarden/secrets route/CLI daemon tests: 40 pass, 0 fail, 158 expect calls; git diff --check passed.

platform/daemon/src/bitwarden.ts
```diff
@@ -0,0 +1,478 @@
...
+		overwrite: options.overwrite,
+	});
+}
+
+export async function deleteBitwardenSecret(
+	name: string,
+	session: string,
+	clientFactory?: BitwardenClientFactory,
+): Promise<boolean> {
+	const client = await getBitwardenClientFactory(clientFactory)(session);
```
PR-Reviewer-Ant, line 165: This treats a Bitwarden CLI status payload with no `status` field as connected. The connect/provider routes use this result as the hard validation gate before persisting the session or activating Bitwarden, so any malformed, unexpected, or future `bw status` JSON that parses but omits `status` can be accepted as usable. Given the PR explicitly claims locked/invalid sessions are rejected before persistence or activation, this should require an explicit unlocked state instead of accepting `undefined`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `be170a51`

</details>

I found one correctness issue in the Bitwarden status handling: a CLI response without a recognized unlocked state can still be treated as connected, which can persist or activate an unusable provider session.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The connect route relies directly on getBitwardenStatus before saving or activating the provider, and getBitwardenStatus marks missing Bitwarden status as connected. I did not run the real bw CLI, so the exact status payload variants are not runtime-confirmed here.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Follow-up pushed in `16a207836`: Bitwarden status validation now requires explicit `status: unlocked`; missing/unknown status is rejected before persistence/activation. Added route regression coverage for an unknown-status session. Validation: focused biome, focused Bitwarden/secrets route/CLI daemon tests: 40 pass, 0 fail, 160 expect calls; daemon/CLI/SDK builds and git diff --check passed.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `16a20783`

</details>

I did not find a concrete correctness, security, or data-integrity issue in the supplied diff. The implementation lines up with the PR description on the main points: Bitwarden stays opt-in, provider metadata remains local, session validation happens before activation, and secret-bearing Bitwarden item payloads are sent over stdin rather than argv.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the core provider implementation, daemon routes, CLI/SDK wiring, docs, and focused tests for connect, activation, migration, fallback, stale folder clearing, canonicalized names, and argv avoidance. I did not run the real Bitwarden CLI, so behavior that depends on exact bw stdin semantics and live vault responses is not runtime-verified here.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Rebased onto latest `main` at `fc6531240`. Post-rebase validation is clean: focused biome, focused Bitwarden/secrets route/CLI daemon tests: 40 pass, 0 fail, 160 expect calls; daemon/CLI/SDK builds and git diff --check passed. Latest reviewer run before rebase found no concrete issue.

PR #699, opened 2026-05-12
feat(daemon): daily reflections with dashboard card

NicholaiVogel:
## Summary

- Adds a periodic reflection worker (`reflection-worker.ts`) that generates once-daily narrative insights from recent memories and session summaries using the default inference provider
- Adds API routes: `GET /api/reflections/today`, `GET /api/reflections`, `POST /api/reflections/generate`, `POST /api/reflections/:id/answer`
- Adds `DailyReflection.svelte` dashboard component on the home tab, replacing the old SuggestedInsights panel
- Answer writeback saves as memory via `txIngestEnvelope` with provenance and the active `agentId` scope
- Migration `068` creates `daily_reflections` table

## Fixes in latest update

- Reflection read endpoints now require `recall` permission, matching the private memory-derived data they return.

- Scheduled reflection generation now discovers active agent IDs from recent memories/session summaries and runs due reflections per agent instead of hardcoding `default`.

- Reflection list `limit` now clamps to `1..100` with invalid values falling back to 30, preventing `LIMIT -1` unlimited history reads.

- Reflection answers now reject bodies over 10,000 characters before DB update or memory ingestion.

- Scheduled worker inserts now handle the daily unique-index race and ingest question memory only after the reflection insert wins.

- Answer writeback now atomically claims unanswered reflections with `answer IS NULL` before ingesting memory, preventing duplicate answer memories under concurrent submissions.

- Manual generation now loads `memory.pipelineV2` through the normal active `SIGNET_PATH` config path
- Reflection inserts and answer updates now use prepared statements instead of `db.exec()` with bind placeholders
- Reflection question and answer memories now carry the same `agentId` scope as the reflection
- Added route, worker, and migration regression tests for manual generation, persistence, scoped answer writeback, duplicate prevention, and empty-day retry behavior
- Daily reflections are constrained unique per `(agent_id, date)` so pre-check races cannot create duplicates
- The worker no longer marks a day complete when there is no source material, so later same-day activity can still generate a reflection
- Reflection provenance now stores actual memory/session summary row IDs instead of content prefixes or timestamps
- The worker now honors cron-style daily `reflections.schedule` values such as `0 8 * * *`, while retrying every five minutes after the scheduled time until a reflection is successfully generated

## Test plan

- [x] `bunx biome check platform/daemon/src/routes/reflection-routes.ts platform/daemon/src/routes/reflection-routes.test.ts platform/daemon/src/pipeline/reflection-worker.ts platform/daemon/src/pipeline/reflection-worker.test.ts platform/daemon/src/logger.ts platform/core/src/index.ts`
- [x] `bun run build:core`
- [x] `bun test platform/core/src/migrations/migrations.test.ts platform/daemon/src/routes/reflection-routes.test.ts platform/daemon/src/pipeline/reflection-worker.test.ts`
- [x] `git diff --check origin/main...HEAD`

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Migration `068-daily-reflections` adds the `daily_reflections` table for generated reflections and answers. It uses `CREATE TABLE IF NOT EXISTS` plus a unique index via `CREATE UNIQUE INDEX IF NOT EXISTS`, so reruns are safe. This is JS daemon/dashboard-only storage for a new table; no Rust daemon parity surface currently exists, so parity is N/A. Rollback compatibility is straightforward: older daemons ignore the table, and normal daemon startup migration handling applies it automatically with no manual operator action.

### DIFF
--- a/platform/core/src/migrations/068-daily-reflections.ts
+++ b/platform/core/src/migrations/068-daily-reflections.ts
@@ -24,7 +24,7 @@ export function up(db: MigrationDb): void {
 			answered_at      TEXT
 		);
 
-		CREATE INDEX IF NOT EXISTS idx_daily_reflections_agent_date
+		CREATE UNIQUE INDEX IF NOT EXISTS idx_daily_reflections_agent_date
 			ON daily_reflections(agent_id, date);
 	`);
 }
--- a/platform/core/src/migrations/068-daily-reflections.ts
+++ b/platform/core/src/migrations/068-daily-reflections.ts
@@ -1,0 +1,30 @@
+import type { MigrationDb } from "./index";
+
+/**
+ * Migration 068: Daily reflections.
+ *
+ * Stores generated daily narrative insights and user answers.
+ * Each row represents one day's reflection for one agent.
+ */
+export function up(db: MigrationDb): void {
+	db.exec(`
+		CREATE TABLE IF NOT EXISTS daily_reflections (
+			id               TEXT PRIMARY KEY,
+			agent_id         TEXT NOT NULL DEFAULT 'default',
+			date             TEXT NOT NULL,
+			summary          TEXT NOT NULL DEFAULT '',
+			patterns         TEXT NOT NULL DEFAULT '[]',
+			question         TEXT,
+			answer           TEXT,
+			answer_memory_id TEXT,
+			memory_ids       TEXT NOT NULL DEFAULT '[]',
+			summary_ids      TEXT NOT NULL DEFAULT '[]',
+			model            TEXT,
+			created_at       TEXT NOT NULL DEFAULT (datetime('now')),
+			answered_at      TEXT
+		);
+
+		CREATE INDEX IF NOT EXISTS idx_daily_reflections_agent_date
+			ON daily_reflections(agent_id, date);
+	`);
+}
--- a/platform/daemon/src/pipeline/reflection-worker.ts
+++ b/platform/daemon/src/pipeline/reflection-worker.ts
@@ -21,7 +21,8 @@ const DEFAULT_DEPS: ReflectionDeps = {
 };
 
 const POLL_INTERVAL_MS = 300_000;
-const STARTUP_DELAY_MS = 60_000;
+const MINUTE_MS = 60_000;
+const DAY_MS = 24 * 60 * MINUTE_MS;
 
 function getAgentsDir(): string {
 	return process.env.SIGNET_PATH || join(homedir(), ".agents");
@@ -55,8 +56,36 @@ function writeLastReflectionTime(agentId: string, date: string): void {
 	}
 }
 
-function todayDate(): string {
-	return new Date().toISOString().slice(0, 10);
+function todayDate(now = new Date()): string {
+	return now.toISOString().slice(0, 10);
+}
+
+function isDailyReflectionUniqueConflict(e: unknown): boolean {
+	const message = e instanceof Error ? e.message : String(e);
+	return message.toLowerCase().includes("unique") && message.includes("daily_reflections");
+}
+
+function scheduledTimeFor(schedule: string, now = new Date()): Date | null {
+	const parts = schedule.trim().split(/\s+/);
+	if (parts.length !== 5 || parts[2] !== "*" || parts[3] !== "*" || parts[4] !== "*") return null;
+	const minute = Number(parts[0]);
+	const hour = Number(parts[1]);
+	if (!Number.isInteger(minute) || !Number.isInteger(hour) || minute < 0 || minute > 59 || hour < 0 || hour > 23) {
+		return null;
+	}
+	const scheduled = new Date(now);
+	scheduled.setHours(hour, minute, 0, 0);
+	return scheduled;
+}
+
+export function nextReflectionDelayMs(schedule: string, lastDate: string | null, now = new Date()): number {
+	const scheduled = scheduledTimeFor(schedule, now);
+	if (!scheduled) return POLL_INTERVAL_MS;
+
+	const date = todayDate(now);
+	if (lastDate === date) return Math.max(0, scheduled.getTime() + DAY_MS - now.getTime());
+	if (now.getTime() < scheduled.getTime()) return Math.max(0, scheduled.getTime() - now.getTime());
+	return POLL_INTERVAL_MS;
 }
 
 export function buildReflectionPrompt(
@@ -92,9 +121,7 @@ export function buildReflectionPrompt(
 	return lines.join("\n");
 }
 
-export function parseReflectionResponse(
-	text: string,
-): { summary: string; patterns: string[]; question?: string } {
+export function parseReflectionResponse(text: string): { summary: string; patterns: string[]; question?: string } {
 	const summary = text.match(/SUMMARY:\s*(.+?)(?:\n|$)/)?.[1]?.trim() ?? text.slice(0, 500);
 	const patternsRaw = text.match(/PATTERNS:\s*(.+?)(?:\n|$)/)?.[1]?.trim() ?? "";
 	const patterns = patternsRaw
@@ -109,7 +136,7 @@ export function parseReflectionResponse(
 export interface ReflectionWorkerHandle {
 	stop(): void;
 	readonly running: boolean;
-	triggerNow(): Promise<void>;
+	triggerNow(agentId?: string): Promise<void>;
 }
 
 export function startReflectionWorker(
@@ -124,29 +151,34 @@ export function startReflectionWorker(
 	async function runReflection(agentId: string): Promise<void> {
 		const date = todayDate();
 		const existing = deps.getDbAccessor().withReadDb((db) => {
-			const row = db
-				.prepare("SELECT id FROM daily_reflections WHERE agent_id = ? AND date = ?")
-				.get(agentId, date) as { id: string } | undefined;
+			const row = db.prepare("SELECT id FROM daily_reflections WHERE agent_id = ? AND date = ?").get(agentId, date) as
+				| { id: string }
+				| undefined;
 			return row?.id ?? null;
 		});
-		if (existing) return;
+		if (existing) {
+			writeLastReflectionTime(agentId, date);
+			return;
+		}
 
 		const cutoff = new Date(Date.now() - config.timeWindowHours * 60 * 60 * 1000).toISOString();
 
 		const memories = deps.getDbAccessor().withReadDb((db) => {
 			const rows = db
 				.prepare(
-					`SELECT content, type, tags, created_at FROM memories
+					`SELECT id, content, type, tags, created_at FROM memories
            WHERE agent_id = ? AND created_at >= ? AND is_deleted = 0
            ORDER BY created_at DESC LIMIT ?`,
 				)
 				.all(agentId, cutoff, config.maxMemories) as {
+				id: string;
 				content: string;
 				type: string;
 				tags: string;
 				created_at: string;
 			}[];
 			return rows.map((r) => ({
+				id: r.id,
 				content: r.content,
 				type: r.type,
 				tags: r.tags ?? "",
@@ -157,20 +189,20 @@ export function startReflectionWorker(
 		const summaries = deps.getDbAccessor().withReadDb((db) => {
 			const rows = db
 				.prepare(
-					`SELECT content, created_at FROM session_summaries
+					`SELECT id, content, created_at FROM session_summaries
            WHERE agent_id = ? AND created_at >= ?
            ORDER BY created_at DESC LIMIT ?`,
 				)
 				.all(agentId, cutoff, config.maxSummaries) as {
+				id: string;
 				content: string;
 				created_at: string;
 			}[];
-			return rows.map((r) => ({ content: r.content, createdAt: r.created_at }));
+			return rows.map((r) => ({ id: r.id, content: r.content, createdAt: r.created_at }));
 		});
 
 		if (memories.length === 0 && summaries.length === 0) {
 			deps.logger.debug("reflections", "No memories or summaries to reflect on", { agentId, date });
-			writeLastReflectionTime(agentId, date);
 			return;
 		}
 
@@ -194,45 +226,55 @@ export function startReflectionWorker(
 		const { summary, patterns, question } = parseReflectionResponse(raw);
 
 		const id = randomUUID();
-		const memoryIds = JSON.stringify(memories.map((m) => m.content.slice(0, 40)));
-		const summaryIds = JSON.stringify(summaries.map((s) => s.createdAt));
+		const memoryIds = JSON.stringify(memories.map((m) => m.id));
+		const summaryIds = JSON.stringify(summaries.map((s) => s.id));
 		const now = new Date().toISOString();
 
-		deps.getDbAccessor().withWriteTx((db) => {
-			db.exec(
-				`INSERT INTO daily_reflections (id, agent_id, date, summary, patterns, question, memory_ids, summary_ids, model, created_at)
-         VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
-				id,
-				agentId,
-				date,
-				summary,
-				JSON.stringify(patterns),
-				question ?? null,
-				memoryIds,
-				summaryIds,
-				config.model,
-				now,
-			);
-		});
-
-		if (question) {
+		try {
 			deps.getDbAccessor().withWriteTx((db) => {
-				txIngestEnvelope(db, {
-					id: randomUUID(),
-					content: `Daily reflection question: ${question}`,
-					contentHash: `reflection-q-${id}`,
-					who: "system",
-					why: "daily-reflection-question",
-					project: null,
-					importance: 0.5,
-					type: "reflection",
-					tags: "reflection,unanswered",
-					pinned: 0,
-					sourceType: "reflection-question",
-					sourceId: id,
-					createdAt: now,
-				});
+				db.prepare(
+					`INSERT INTO daily_reflections
+					 (id, agent_id, date, summary, patterns, question, memory_ids, summary_ids, model, created_at)
+					 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+				).run(
+					id,
+					agentId,
+					date,
+					summary,
+					JSON.stringify(patterns),
+					question ?? null,
+					memoryIds,
+					summaryIds,
+					config.model,
+					now,
+				);
+
+				if (question) {
+					txIngestEnvelope(db, {
+						id: randomUUID(),
+						content: `Daily reflection question: ${question}`,
+						contentHash: `reflection-q-${id}`,
+						who: "system",
+						why: "daily-reflection-question",
+						project: null,
+						importance: 0.5,
+						type: "reflection",
+						tags: "reflection,unanswered",
+						pinned: 0,
+						sourceType: "reflection-question",
+						sourceId: id,
+						agentId,
+						createdAt: now,
+					});
+				}
 			});
+		} catch (e) {
+			if (isDailyReflectionUniqueConflict(e)) {
+				deps.logger.debug("reflections", "Reflection already generated before worker insert", { agentId, date });
+				writeLastReflectionTime(agentId, date);
+				return;
+			}
+			throw e;
 		}
 
 		writeLastReflectionTime(agentId, date);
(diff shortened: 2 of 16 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `a807ad90`.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,273 @@
...
+			logger.error("reflections", "Failed to list reflections", {
+				error: e instanceof Error ? e.message : String(e),
+			});
+			return c.json({ error: "Failed to list reflections" }, 500);
+		}
+	});
+
+	app.post("/api/reflections/generate", async (c) => {
+		const agentId = c.req.query("agentId") ?? "default";
+		const date = new Date().toISOString().slice(0, 10);
```
PR-Reviewer-Ant, line 122: `loadPipelineConfig()` is called without the YAML/config object it expects. The implementation in `memory-config.ts` immediately reads `yaml.memory`, so this route will throw before manual generation can run. This also means the test-plan curl for `/api/reflections/generate` does not currently exercise the advertised path successfully. Use the normal loaded memory config path for the active agents dir, or pass the parsed YAML explicitly.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,273 @@
...
+				.all(agentId, cutoff, cfg.maxSummaries) as {
+				content: string;
+				created_at: string;
+			}[]).map((r) => ({ content: r.content, createdAt: r.created_at }));
+		});
+
+		if (memories.length === 0 && summaries.length === 0) {
+			return c.json({ error: "No memories or summaries in the time window" }, 400);
+		}
+
```
PR-Reviewer-Ant, line 184: This `db.exec()` call is using `?` placeholders plus bind values, but the database exec API used in this repo executes raw SQL strings; it does not bind parameters. As written, inserting a generated reflection will fail instead of creating the row. This should be `db.prepare(...).run(...)`, same as the surrounding migration/index code uses for parameterized statements.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,273 @@
...
+		if (!body.answer || typeof body.answer !== "string" || body.answer.trim().length === 0) {
+			return c.json({ error: "answer is required" }, 400);
+		}
+
+		try {
+			const existing = getDbAccessor().withReadDb((db) => {
+				return db
+					.prepare("SELECT * FROM daily_reflections WHERE id = ? AND agent_id = ?")
+					.get(id, agentId) as ReflectionRow | undefined;
+			});
```
PR-Reviewer-Ant, line 264: Same parameter binding issue here: `db.exec()` is used for an `UPDATE ... WHERE id = ? AND agent_id = ?` with bind arguments. Answering a reflection will fail before writeback, so the PR's claimed answer-save path is broken. Use a prepared statement inside the transaction.

platform/daemon/src/pipeline/reflection-worker.ts
```diff
@@ -0,0 +1,291 @@
...
+
+		let raw: string;
+		try {
+			const provider = deps.getInferenceProvider("default");
+			raw = await provider.generate(prompt, {
+				timeoutMs: config.timeout,
+				maxTokens: config.maxTokens,
+			});
+		} catch (e) {
+			deps.logger.warn("reflections", "Generation failed", {
```
PR-Reviewer-Ant, line 219: The background worker has the same `db.exec()`/placeholder problem when inserting the daily reflection. That means scheduled generation can successfully call the model and then fail at persistence. This needs a prepared `INSERT` statement.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,273 @@
...
+				return c.json({ error: "Reflection not found" }, 404);
+			}
+			if (existing.answer) {
+				return c.json({ error: "Already answered" }, 409);
+			}
+
+			const now = new Date().toISOString();
+			const memoryId = randomUUID();
+
+			getDbAccessor().withWriteTx((db) => {
```
PR-Reviewer-Ant, line 277: The route is explicitly scoped by `agentId`, but the answer memory written through `txIngestEnvelope` does not carry that agent scope. Setting `who: agentId` is not the same as storing the memory under that agent, so answering a non-default agent reflection can create a memory outside the reflection's scope. Please thread the same agent id into the ingestion path the way other scoped memory writes do.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `3ad4bff1`

</details>

I found a few implementation issues that look like they will break the new reflection flow at runtime: manual generation cannot load config as written, and the DB writes use the wrong SQLite API. There is also a multi-agent scoping gap in answer writeback that can put non-default agent answers into the wrong memory scope.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The failures are directly visible in the changed files: loadPipelineConfig is called with no YAML argument in reflection-routes.ts even though its implementation dereferences yaml.memory, and multiple writes pass bind values to db.exec instead of prepare(...).run(...). The agent scoping issue follows from the new route accepting agentId while txIngestEnvelope calls do not pass an agent scope.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `353a8a85`

</details>

I found two correctness issues in the reflection persistence/scheduling path that can break the once-daily contract. The dashboard wiring looks straightforward, but this PR changes UI without a screenshot or preview in the description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The duplicate-row issue is directly visible in migration 068, which creates only a non-unique agent/date index while both the worker and manual route use pre-check-then-insert. The skipped-day behavior is explicit in reflection-worker.ts where the worker writes the last-reflection timestamp even when no source material exists.


**Since my last review:**
- Fixed or likely addressed: 5
- Still blocking: 0
- Newly found: 2

platform/core/src/migrations/068-daily-reflections.ts
```diff
@@ -0,0 +1,30 @@
...
+			question         TEXT,
+			answer           TEXT,
+			answer_memory_id TEXT,
+			memory_ids       TEXT NOT NULL DEFAULT '[]',
+			summary_ids      TEXT NOT NULL DEFAULT '[]',
+			model            TEXT,
+			created_at       TEXT NOT NULL DEFAULT (datetime('now')),
+			answered_at      TEXT
+		);
+
```
PR-Reviewer-Ant, line 26: This index does not enforce the PR's once-daily invariant. Both `runReflection()` and `POST /api/reflections/generate` do a read-before-insert check outside the eventual insert, so two concurrent requests, or the scheduled worker racing a manual generate, can insert multiple rows for the same `agent_id` and `date`. This table should enforce uniqueness on `(agent_id, date)` and the insert path should handle the conflict deliberately.

platform/daemon/src/pipeline/reflection-worker.ts
```diff
@@ -0,0 +1,292 @@
...
+				error: e instanceof Error ? e.message : String(e),
```
PR-Reviewer-Ant, line 220: When there are no memories or summaries, the worker records today's date as already reflected even though no reflection row was created. If the daemon starts before the user has generated any activity, later memories from the same day will be skipped until tomorrow. That undercuts the daily reflection behavior this PR is adding; the worker should leave the day eligible until it actually persists a reflection, or otherwise use a separate backoff marker that does not suppress the whole day.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,267 @@
...
+				 (id, agent_id, date, summary, patterns, question, memory_ids, summary_ids, model, created_at)
+				 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+			).run(
+				id,
+				agentId,
+				date,
+				summary,
+				JSON.stringify(patterns),
+				question ?? null,
+				JSON.stringify(memories.map((m) => m.content.slice(0, 40))),
```
PR-Reviewer-Ant: This stores `memories.map((m) => m.content.slice(0, 40))` in `memory_ids` and `summaries.map((s) => s.createdAt)` in `summary_ids`, so the persisted reflection cannot be traced back to the actual memory or session summary rows. That diverges from the schema/PR claim that reflections carry provenance, and it will make later inspection, deletion handling, or source attribution unreliable. Please select the source row IDs in the queries and persist those IDs here.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `17c71ba7`

</details>

I found one provenance/data-integrity issue: the new reflection rows do not actually store the source memory or summary IDs even though the schema and PR description say they do. I also noticed the configured reflection schedule is currently ignored, but that is a non-blocking behavior mismatch.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed route and worker code explicitly serialize memory content snippets and summary timestamps into columns named memory_ids and summary_ids. The PR description claims provenance is saved, and the migration schema creates ID columns for that purpose.


**Since my last review:**
- Still blocking: 0
- Newly found: 3

platform/daemon/src/pipeline/reflection-worker.ts
```diff
@@ -0,0 +1,291 @@
...
+				agentId,
+				date,
+				summary,
+				JSON.stringify(patterns),
+				question ?? null,
+				memoryIds,
+				summaryIds,
+				config.model,
+				now,
+			);
```
PR-Reviewer-Ant: Same provenance issue in the scheduled worker: `memory_ids` receives content prefixes and `summary_ids` receives timestamps, not row IDs. This means manual and scheduled reflections both write non-identifying provenance into ID columns.

platform/daemon/src/pipeline/reflection-worker.ts
```diff
@@ -0,0 +1,291 @@
...
+		stopped = true;
+		if (timer) {
+			clearTimeout(timer);
+			timer = null;
+		}
+		running = false;
+	}
+
+	start();
+
```
PR-Reviewer-Ant, line 352: The configured `reflections.schedule` value is not used. The worker starts after a fixed 60s delay, then polls every five minutes and generates as soon as the date changes. That is not necessarily wrong for a first pass, but it does not implement the `schedule: "0 8 * * *"` behavior exposed in config and described as scheduled daily generation.

surfaces/dashboard/src/lib/api.ts
```diff
@@ -2889,6 +2889,54 @@ export async function getDiagnostics(): Promise<DiagnosticsReport | null> {
...
+	createdAt: string;
+	answeredAt: string | null;
+}
+
+export interface TodayReflectionResponse {
+	reflection: DailyReflection | null;
+}
+
+export async function getTodayReflection(): Promise<TodayReflectionResponse> {
+	try {
```
PR-Reviewer-Ant, line 2917: The dashboard client always calls the reflection endpoints without an `agentId`, so both `getTodayReflection()` and `answerReflection()` operate on the backend default scope. That undercuts the PR's active-agent scoping claim for the actual dashboard path: in a non-default agent home view, the card will show the default agent's reflection and save answers as default-agent memories. The component/API should thread the active agent id the same way the backend routes expect it.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,270 @@
...
+                 WHERE agent_id = ?
+                 ORDER BY date DESC
+                 LIMIT ?`,
+					)
+					.all(agentId, limit) as ReflectionRow[];
+			});
+
+			return c.json({ reflections: rows.map(formatReflection) });
+		} catch (e) {
+			logger.error("reflections", "Failed to list reflections", e instanceof Error ? e : undefined);
```
PR-Reviewer-Ant, line 115: This new manual generation endpoint is expensive and mutates persisted state, but the route is registered without an explicit permission check. The repo convention calls out admin/refresh/mutation endpoints as needing explicit permission checks; this route invokes the inference provider and inserts a reflection row. Please gate it with the same auth policy used for comparable admin or mutation routes.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,270 @@
...
+			return db.prepare("SELECT * FROM daily_reflections WHERE id = ?").get(id) as ReflectionRow;
+		});
+
+		return c.json({ reflection: formatReflection(row) });
+	});
+
+	app.post("/api/reflections/:id/answer", async (c) => {
+		const id = c.req.param("id");
+		const agentId = c.req.query("agentId") ?? "default";
+
```
PR-Reviewer-Ant, line 242: Answer writeback also mutates user memory without an explicit permission check. Because this route writes into `daily_reflections` and then calls `txIngestEnvelope`, it should require the appropriate mutation permission instead of being publicly reachable through the dashboard API surface.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,270 @@
...
+			logger.error("reflections", "Manual generation failed", e instanceof Error ? e : undefined);
+			return c.json({ error: "LLM generation failed" }, 500);
+		}
+
+		const { summary, patterns, question } = parseReflectionResponse(raw);
+		const id = randomUUID();
+		const now = new Date().toISOString();
+
+		getDbAccessor().withWriteTx((db) => {
+			db.prepare(
```
PR-Reviewer-Ant: The unique `(agent_id, date)` index prevents duplicate rows, but concurrent manual generation can still pass the pre-check and then throw on insert. That turns the expected duplicate-prevention path into a 500 after doing LLM work. Catch the unique constraint at insert time and return the same 409 response, or use an insert strategy that makes the race deterministic.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `2e4a8700`

</details>

I found a couple of correctness/security issues in the new reflection API and dashboard wiring. The backend has scoped routes, but the dashboard does not pass the active agent scope, and the new mutation/generation endpoints appear to be mounted without explicit permission checks.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The changed route and dashboard files show the relevant behavior directly. I do not have the full auth middleware implementation in the provided context, so the permission finding is based on the absence of explicit route-level checks in reflection-routes.ts despite the repository convention requiring them for admin/mutation endpoints.


**Since my last review:**
- Still blocking: 0
- Newly found: 4

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6a012313`

</details>

I found one correctness issue in the answer writeback path: concurrent submissions can both pass the pre-check, overwrite the reflection answer, and create duplicate answer memories.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The race is directly visible in platform/daemon/src/routes/reflection-routes.ts: the route reads the row and checks existing.answer before opening the write transaction, but the later UPDATE only matches id and agent_id and does not assert answer is still null.


**Since my last review:**
- Fixed or likely addressed: 4
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,288 @@
...
+			}
+
+			const now = new Date().toISOString();
+			const memoryId = randomUUID();
+
+			getDbAccessor().withWriteTx((db) => {
+				db.prepare(
+					`UPDATE daily_reflections
+					 SET answer = ?, answer_memory_id = ?, answered_at = ?
+					 WHERE id = ? AND agent_id = ?`,
```
PR-Reviewer-Ant: This writeback is not atomic with the earlier `existing.answer` check. Two concurrent `POST /api/reflections/:id/answer` requests can both read an unanswered reflection, then both enter the transaction; the later UPDATE overwrites `answer`/`answer_memory_id`, while both calls ingest a `reflection-answer` memory. That leaves one orphaned answer memory and makes the final answer depend on request timing. Please make the UPDATE conditional, e.g. `WHERE id = ? AND agent_id = ? AND answer IS NULL`, check the affected row count inside the same transaction, and only ingest the memory after the claim succeeds.

platform/daemon/src/pipeline/reflection-worker.ts
```diff
@@ -0,0 +1,318 @@
...
+				JSON.stringify(patterns),
+				question ?? null,
+				memoryIds,
+				summaryIds,
+				config.model,
+				now,
+			);
+		});
+
+		if (question) {
```
PR-Reviewer-Ant: The worker still has a check-then-insert race against manual generation or another daemon instance. If a reflection is inserted after the `existing` read but before this `INSERT`, the unique `(agent_id, date)` index throws here. Unlike the manual route, this path does not catch the duplicate and return/skip, so the scheduled tick can reject out of the timer callback and hit the daemon's unhandled rejection shutdown path. Please handle the unique-conflict case in the worker the same way the route does, or use an atomic `INSERT OR IGNORE`/claim pattern and only ingest the question after the insert actually wins.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `7bcb3341`

</details>

I found one daemon reliability bug in the worker's daily insert path. The API route handles the unique-index race, but the periodic worker does not, so the same race can still surface as an unhandled worker failure.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR adds a unique `(agent_id, date)` constraint and catches duplicate inserts in the manual route, but the periodic worker still performs a pre-check followed by an uncaught INSERT. The changed worker code is enough to prove the race: another generator can insert after the pre-check and before the worker's write transaction.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,298 @@
...
+			}
+
+			return c.json({ reflection: formatReflection(row) });
+		} catch (e) {
+			logger.error("reflections", "Failed to fetch today's reflection", e instanceof Error ? e : undefined);
+			return c.json({ error: "Failed to fetch reflection" }, 500);
+		}
+	});
+
+	app.get("/api/reflections", (c) => {
```
PR-Reviewer-Ant, line 95: `limit` is only capped on the upper end. `?limit=-1` becomes `LIMIT -1` in SQLite, which means no limit, so this endpoint can return the full reflection history despite the intended 100-row cap. Clamp to a sane positive range, for example `1..100`, and treat invalid values as the default.

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,298 @@
...
+		} catch {
+			return c.json({ error: "Invalid JSON body" }, 400);
+		}
+
+		if (!body.answer || typeof body.answer !== "string" || body.answer.trim().length === 0) {
+			return c.json({ error: "answer is required" }, 400);
+		}
+
+		try {
+			const answer = body.answer.trim();
```
PR-Reviewer-Ant: This mutation accepts and stores an unbounded answer body, then writes it directly as a memory inside the DB transaction. That bypasses the normal API boundary guardrails and lets a client submit arbitrarily large content into `daily_reflections.answer` and `memories.content`. Please add a max length check before the update/ingest path and return a clear 400/413-style error for oversized answers.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d501c7f6`

</details>

I found two boundary-check gaps in the new reflections API. The core reflection persistence and answer-claim race handling look consistent with the PR description, but the public inputs still need tighter bounds before this should land.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are visible directly in the added route code: the list endpoint passes a possibly negative LIMIT into SQLite, and the answer endpoint writes trimmed request text directly through txIngestEnvelope without a size cap.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/pipeline/reflection-worker.ts
```diff
@@ -0,0 +1,333 @@
...
+		});
+	}
+
+	async function tick(): Promise<void> {
+		if (stopped || generating) return;
+		generating = true;
+		try {
+			const date = todayDate();
+			const lastDate = readLastReflectionTime("default");
+			if (lastDate !== date && nextReflectionDelayMs(config.schedule, null) === POLL_INTERVAL_MS) {
```
PR-Reviewer-Ant: The scheduled worker only runs `runReflection("default")`, while the dashboard and routes are agent-scoped. That means any non-default agent will never get the once-daily reflection unless someone manually calls the admin generation endpoint with `agentId`. Since this PR replaces the home-tab insights card with `DailyReflection {agentId}`, multi-agent users will see an empty card indefinitely even when that agent has recent memories. The worker needs to discover/schedule the relevant agents, or otherwise align with the active daemon agent scope instead of hardcoding `default`.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `a637d217`

</details>

I found one correctness issue: the scheduled worker is still hardcoded to the default agent, so non-default agent home tabs will never receive the daily reflection this PR adds.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The dashboard passes an agentId into DailyReflection and the routes read/write reflections by agentId, but reflection-worker.ts only schedules and triggers runReflection("default"). That directly prevents the scheduled feature from working for non-default agents.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/reflection-routes.ts
```diff
@@ -0,0 +1,313 @@
...
+		answer: r.answer,
+		answerMemoryId: r.answer_memory_id,
+		createdAt: r.created_at,
+		answeredAt: r.answered_at,
+	};
+}
+
+export function registerReflectionRoutes(app: Hono): void {
+	app.use("/api/reflections/generate", async (c, next) => {
+		return requirePermission("admin", authConfig)(c, next);
```
PR-Reviewer-Ant, line 67: These reflection read routes need an explicit read-style permission check. The generated rows summarize recent memories/session summaries and can include the user's answer text plus answer memory IDs, so `GET /api/reflections/today` and `GET /api/reflections` are exposing private memory-derived data while only the generate/answer mutation routes are protected. Please gate the read endpoints with the same auth policy used for other memory/dashboard read APIs.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `df4b9dd4`

</details>

I found one security issue: the new read endpoints expose memory-derived reflections without an explicit permission gate, while the mutation endpoints correctly add one.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed route file explicitly wraps POST /api/reflections/generate with admin and POST /api/reflections/:id/answer with modify, but GET /api/reflections/today and GET /api/reflections are registered without requirePermission. Those responses include summaries, questions, answers, and answer memory IDs derived from private memory/session data.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a807ad90`

</details>

I did not find a concrete blocker in the changed code. The implementation lines up with the PR description on scoped reads/writes, daily uniqueness, bounded list limits, answer-size rejection, and duplicate-answer prevention.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the new migration, route handlers, worker, dashboard client/component, and focused regression tests for the main risk areas called out in the description. I did not run the daemon or dashboard, so visual behavior and real auth middleware composition were not runtime-verified.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #709, opened 2026-05-14
fix(daemon): clean up detached codex acp processes

NicholaiVogel:
## Summary
- tag each ACPX provider invocation with a unique run id inherited by child processes
- clean up detached `codex-acp` processes that escape the ACPX launcher process group
- add a Linux regression test for successful ACPX runs that daemonize Codex ACP outside the original process group

## Test Plan
- [x] `bun install --frozen-lockfile`
- [x] `bun run build:core`
- [x] `bun test platform/daemon/src/pipeline/provider.test.ts --timeout 10000 --test-name-pattern 'createAcpxProvider'`
- [x] `bunx biome check platform/daemon/src/pipeline/provider.ts`
- [x] `git diff --check -- platform/daemon/src/pipeline/provider.ts platform/daemon/src/pipeline/provider.test.ts`
- [x] `bun run --filter @signet/daemon build`

## PR Readiness (MANDATORY)

- [x] No database migrations changed
- [x] Migration is idempotent

Rollback / compatibility: revert the provider cleanup change to restore previous ACPX lifecycle behavior. No schema or config migration is involved.

### DIFF
--- a/platform/daemon/src/pipeline/provider.ts
+++ b/platform/daemon/src/pipeline/provider.ts
@@ -8,13 +8,12 @@
 // On Windows, use node:child_process spawn with windowsHide to prevent
 // console window flashing. Bun.spawn doesn't support windowsHide.
 import { spawn as nodeSpawn } from "node:child_process";
-import { chmodSync, cpSync, existsSync, mkdirSync, mkdtempSync, rmSync } from "node:fs";
+import { chmodSync, cpSync, existsSync, mkdirSync, mkdtempSync, readFileSync, readdirSync, rmSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
 import { isAbsolute, join, resolve as resolvePath } from "node:path";
 import { Readable } from "node:stream";
 import {
 	DEFAULT_PROVIDER_RATE_LIMIT,
-	defaultPipelineModel,
 	type LlmGenerateResult,
 	type LlmProvider,
 	type LlmUsage,
@@ -22,6 +21,7 @@ import {
 	OPENCODE_PIPELINE_SYSTEM_PROMPT,
 	type PipelineExtractionConfig,
 	type ProviderRateLimitConfig,
+	defaultPipelineModel,
 } from "@signet/core";
 import { logger } from "../logger";
 import { bypassSession } from "../session-tracker";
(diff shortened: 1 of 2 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `54eafeae`.

platform/daemon/src/pipeline/provider.ts
```diff
@@ -1133,6 +1134,63 @@ function terminateChildProcessTreeWithEscalation(child: ReturnType<typeof nodeSp
...
+	try {
+		const cmdline = readFileSync(`/proc/${pid}/cmdline`, "utf8");
+		return cmdline
+			.split("\0")
+			.filter(Boolean)
+			.some((arg) => basenames.has(arg.split("/").pop() ?? ""));
+	} catch {
+		return false;
+	}
+}
```
PR-Reviewer-Ant, line 1169: This cleanup path can throw before the provider promise resolves or rejects if `/proc` is missing or unreadable. The per-process helpers already degrade to `false` on read failures, but the initial `readdirSync('/proc')` is outside a try/catch and `finish()` calls `cleanupAcpxAgentProcesses()` before invoking the resolver. On Linux systems with a restricted or unavailable procfs, a successful ACPX run could become an uncaught exception or leave the promise settled flag flipped without delivering the result. It would be safer for the whole proc scan to be best-effort, matching the existing per-pid handling.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3c5169f8`

</details>

The run-id tagging approach matches the PR description and the regression test covers the escaped-process case. I found one reliability edge case in the cleanup path: a Linux environment without a readable /proc can now break otherwise successful ACPX calls.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible in cleanupAcpxAgentProcesses: it synchronously calls readdirSync('/proc') from the settle path without guarding that top-level proc scan. I did not have a runtime reproduction for a hardened or proc-less Linux environment, so I am treating this as a warning rather than a required change.

platform/daemon/src/pipeline/provider.ts
NicholaiVogel: Fixed in `54eafeae6`. The proc scan is now fully best-effort: `cleanupAcpxAgentProcesses()` resolves the proc root, wraps the top-level `readdirSync` in a try/catch, and returns without affecting the provider result if procfs is missing or unreadable. Added a regression test that points the cleanup path at a missing proc root and verifies an otherwise successful ACPX response still resolves.

Validation rerun:
- `bun test platform/daemon/src/pipeline/provider.test.ts --timeout 10000 --test-name-pattern 'createAcpxProvider'`
- `bunx biome check platform/daemon/src/pipeline/provider.ts`
- `git diff --check -- platform/daemon/src/pipeline/provider.ts platform/daemon/src/pipeline/provider.test.ts`
- `bun run --filter @signet/daemon build`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `54eafeae`

</details>

The new cleanup path matches the PR goal: ACPX invocations get a per-run environment tag, escaped Codex ACP processes are found through /proc, and the previously flagged unreadable-proc-root case is now handled as best-effort. I did not find a remaining correctness, security, or data-integrity issue worth flagging.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The relevant changed code is contained in platform/daemon/src/pipeline/provider.ts and provider.test.ts. cleanupAcpxAgentProcesses now guards the top-level proc scan, filters by both agent command basename and the inherited SIGNET_ACPX_RUN_ID, and is covered by tests for detached Codex ACP cleanup plus missing proc-root best-effort behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #711, opened 2026-05-14
fix(daemon): surface unhealthy diagnostics domains

NicholaiVogel:
## Summary
- Makes diagnostics composite status reflect the worst degraded/unhealthy domain instead of hiding it behind a weighted average.
- Adds a regression test for the exact failure mode seen locally: queue domain unhealthy while composite score still rounds into the healthy band.

## Context
During dogfood, the live daemon was down after a graceful systemd shutdown with a 43.9G service memory peak. The ontology/constellation route on 0.118.4 is already bounded and smoke-tested locally, but diagnostics were still reporting the overall memory system as healthy while the queue domain had ~14.5k pending jobs and was explicitly unhealthy. That is bad stability UX: the dashboard can show green while the machine is quietly chewing glass.

## Validation
- `bun test platform/daemon/src/diagnostics.test.ts`
- `bunx biome check platform/daemon/src/diagnostics.ts platform/daemon/src/diagnostics.test.ts`
- `git diff --check`
- Live smoke: `/api/knowledge/constellation?agent_id=hermes-agent&limit=150&max_aspects_per_entity=6&max_attributes_per_aspect=4&dependency_limit=500` returned 150 entities / 135 dependencies in 0.06s.

## Live mitigation
I temporarily set `/home/nicholai/.agents/agent.yaml` `memory.pipelineV2.paused: true` and restarted the daemon so the 14k-job backlog stops hammering background inference while we inspect it. Remove that line or set it false when we want the pipeline workers back on.

## PR Readiness (MANDATORY)

## Migration Notes


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `9360cdeb`.

platform/daemon/src/diagnostics.test.ts
```diff
@@ -334,19 +334,19 @@ describe("getDiagnostics", () => {
...
 		const tracker = createProviderTracker();
 		const report = getDiagnostics(asReadDb(db), tracker);
 
-		// Queue degrades by 0.3 x weight 0.3 = -0.09 off the composite
-		expect(report.composite.score).toBeLessThan(1);
-		expect(report.queue.status).not.toBe("healthy");
+		expect(report.queue.status).toBe("unhealthy");
+		expect(report.composite.score).toBeGreaterThanOrEqual(0.8);
+		expect(report.composite.status).toBe("unhealthy");
 	});
```
PR-Reviewer-Ant, line 354: This test setup only drives the queue score to exactly 0.5: 51 pending jobs subtracts 0.3, and making them 10 minutes old subtracts 0.2. `scoreStatus(0.5)` returns `degraded`, not `unhealthy`, so this regression test should fail before it verifies the composite behavior. Add another unhealthy signal, increase the setup severity, or change the expectation to match the intended domain threshold.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d19e3f69`

</details>

The implementation direction is small and matches the stated goal, but the new regression test does not actually construct an unhealthy queue under the current scoring rules. As written, the validation claimed in the PR should fail.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The queue score math is visible in platform/daemon/src/diagnostics.ts: depth > 50 subtracts 0.3 and oldestAgeSec > 300 subtracts 0.2, leaving score 0.5. scoreStatus treats scores >= 0.5 as degraded, so the new test's expectation that this setup is unhealthy is inconsistent with the changed code.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `9360cdeb`

</details>

The updated implementation now matches the PR's stated goal: composite status is promoted to the worst diagnostics domain while preserving the weighted score, and the regression test now actually drives the queue domain unhealthy. I did not find a correctness, security, data integrity, or breaking-change issue in this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed behavior is fully contained in platform/daemon/src/diagnostics.ts, where composite.status now derives from the explicit statuses of queue, storage, index, provider, mutation, duplicate, connector, and update. The test in platform/daemon/src/diagnostics.test.ts now creates both queue depth over 50 and a recent dead job, which makes queue.score 0.4 under the visible scoring rules and proves the high composite-score/unhealthy-domain case.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #712, opened 2026-05-14
feat(bundle): native bundle system with zero-prereq installer

aaf2tbz:
## Summary

Node.js compatibility for the Signet daemon + native bundle installer. The daemon now runs under plain Node.js with zero Bun dependency — all Bun-specific APIs (`Bun.spawn`, `Bun.which`, `Bun.file`, `Bun.write`, `Bun.CryptoHasher`, `bun:sqlite`, `Bun.build`) have been replaced with Node.js equivalents. Combined with a zero-prereq `curl | bash` installer that ships a bundled Node.js runtime.

> Replaces #696 (review bot overwhelmed — same code, fresh PR).

### Node.js compatibility (daemon + build)

- New: `platform/daemon/src/which.ts` — cross-platform binary resolver (delegates to `Bun.which` under Bun, manual PATH lookup under Node)
- Modified: `platform/daemon/src/db.ts` — runtime-detecting SQLite import (`bun:sqlite` | `better-sqlite3`)
- Modified: `platform/daemon/src/db-accessor.ts` — same runtime detection for the 800-line DB singleton
- Modified: `platform/daemon/src/pipeline/provider.ts` — `Bun.spawn` → `node:child_process.spawn`, `Bun.which` → `which()`
- Modified: `platform/daemon/src/scheduler/spawn.ts` — `Bun.which` → `which()`
- Modified: `platform/daemon/src/pipeline/provider-resolution.ts` — `Bun.which` → `which()`
- Modified: `platform/daemon/src/routes/repair-routes.ts` — `Bun.which` → `which()`
- Modified: `platform/daemon/src/daemon.ts` — `Bun.file()` → `fs/promises.readFile()`
- Modified: `platform/daemon/src/widget-gen.ts` — `Bun.write()` → `writeFileSync()`
- Modified: `platform/daemon/src/pipeline/skill-graph.ts` — `Bun.CryptoHasher` → `crypto.createHash()`
- Modified: `platform/daemon/build.ts` — `Bun.build(target:"bun")` → `esbuild(platform:"node", target:"node20")`
- Modified: `dist/signetai/build-daemon.ts` — same esbuild migration
- Modified: `platform/daemon/package.json` — added `better-sqlite3` dep, `esbuild` + `@types/better-sqlite3` devDeps

### Native bundle system

- New: `deploy/bundle/install.sh` — curl | bash installer shipping Node.js runtime
- New: `deploy/bundle/uninstall.sh` — clean uninstaller
- New: `deploy/bundle/update.sh` — incremental updater with manifest checksum verification
- New: `deploy/bundle/scripts/generate-manifest.ts` — per-platform manifest generator
- New: `.github/workflows/bundle.yml` — CI with per-component change detection, 4 platform targets (darwin-arm64, darwin-x64, linux-x64, linux-arm64)

### Electron desktop fixes

- Modified: `surfaces/desktop/src/paths.ts` — path hardening with `relative()` containment check
- Modified: `surfaces/desktop/src/main.ts` — macOS dock icon fix

## Testing

- `bun test` passes (filesystem glob tests)
- `bun run typecheck` passes
- `bash -n` validates all shell scripts

## AI disclosure


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `f5e4a528`.

deploy/bundle/install.sh
```diff
@@ -0,0 +1,568 @@
...
+  local actual
+  actual="$($SHA256_CMD "$file" | awk '{print $1}')"
+  [ "$actual" = "$expected" ]
+}
+
+safe_tar_extract() {
+  local archive="$1" dest="$2"
+  local unsafe
+  unsafe="$(tar tvf "$archive" 2>/dev/null | while read -r line; do
+    local entry
```
PR-Reviewer-Ant, line 135: This archive safety check is parsing `tar tvf` output by taking the text after the last space. That is not a safe way to recover member names: a tar entry like `../../outside file` will be seen as just `file`, so the traversal check can pass and the following `tar xzf` can still write outside `$dest`. Since this installer is explicitly downloading and extracting release archives, the path validation needs to use a listing format that preserves names reliably, for example `tar -tzf` plus strict checks on the raw member path, and similar care for symlink targets.

deploy/bundle/update.sh
```diff
@@ -0,0 +1,342 @@
...
+
+safe_tar_extract() {
+  local archive="$1" dest="$2"
+  local unsafe
+  unsafe="$(tar tvf "$archive" 2>/dev/null | while read -r line; do
+    local entry
+    entry="$(printf '%s' "$line" | sed 's/^.* //')"
+    case "$line" in
+      l*|h*)
+        local link_target
```
PR-Reviewer-Ant, line 101: Same extraction guard issue as the installer: parsing `tar tvf` with `sed 's/^.* //'` can miss unsafe archive members whose names contain spaces. The updater is a long-lived trust boundary because it will process future release assets, so it should use a robust member-name listing before extraction rather than relying on human-readable tar output.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f5e4a528`

</details>

I found a real issue in the archive safety check: the installer/updater intends to block path traversal before extraction, but the tar listing parser can misparse filenames with spaces and miss unsafe paths. That should be fixed before this installer is shipped as a curl-pipe entrypoint.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new installer and updater both add safe_tar_extract, and the parsing logic is visible in the diff. The issue is directly in the security boundary for downloaded release archives, not inferred from missing runtime behavior.

PR #716, opened 2026-05-16
feat(ontology): add audited control plane

NicholaiVogel:
## Summary

Adds an audited ontology control plane for daemon-backed graph mutations, plus a built-in dreaming ingestion skill that turns transcript/source evidence into ontology operation JSONL.

These tools are usable as soon as this PR merges: exact CLI/API ontology operations can apply directly through permission-checked daemon endpoints, and generated/dreaming ingestion can dry-run or write pending proposals for review.

## Changes

- Added migration 070 for ontology row status/archive metadata and first-class claim version fields.
- Added `/api/ontology/operations/apply` and `/api/ontology/operations/batch` with apply-first audit rows, `--dry-run`, `--propose`, atomic batch behavior, and dry-run line errors.
- Added operation handlers for entities, aspects, claims/versions, and links.
- Expanded `signet ontology` CLI with entity, claim, aspect, link, stream, pipeline, and config commands.
- Added default active-read filtering for archived graph rows while preserving history/version inspection.
- Added the built-in `dreaming` skill and docs for inventory, runbook, and implementation progress.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: built-in skills and docs

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Migration 070 is additive. Existing ontology rows default to active state; existing attributes are backfilled to version 1 with `version_root_id = id`. Rollback would require removing the added status/archive/proposal/version columns and indexes.

Rust daemon parity is N/A for this slice because these changes extend the JS daemon ontology control-plane surface.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Focused validation run locally:

Checklist exception label is applied because repo-wide lint/typecheck remain blocked by unrelated issues listed below; focused validation for this PR slice is green.

- `bun scripts/spec-deps-check.ts` passed.
- `bun run build:core` passed.
- `cd platform/core && bun run typecheck` passed.
- `cd surfaces/cli && bun run build` passed.
- `cd platform/daemon && bun run build` passed with existing unrelated dashboard Svelte warnings.
- `bun test platform/core/src/migrations/migrations.test.ts surfaces/cli/src/commands/ontology.test.ts platform/daemon/src/ontology-proposals.test.ts platform/daemon/src/dreaming-skill.test.ts platform/daemon/src/knowledge-graph-list.test.ts` passed: 129 pass, 0 fail.
- Targeted Biome check on changed files passed.
- `git diff --check` passed.

Known unrelated repo-wide blockers observed locally:

- `bun run typecheck` fails in `@signet/desktop` because `src/desktop-updates.ts` cannot resolve `electron-updater`.
- `bun run lint` fails repo-wide on existing broad Biome formatting diagnostics across package/generated files.

## AI disclosure

## Notes

This PR does not add an external `$SIGNET_WORKSPACE/ontology/graph.yaml` policy gate. That does not make the tool surface read-only: the daemon-backed `signet ontology entity`, `claim`, `aspect`, `link`, and `stream apply` commands are live write tools after merge, with all successful mutations audited through `ontology_proposals`.

`goal.md` was left untracked and is not part of this PR.

### DIFF
--- a/docs/knowledge-graph-control-plane.md
+++ b/docs/knowledge-graph-control-plane.md
@@ -119,9 +119,8 @@ rejection, default hiding for archived graph rows, and an in-process
 end-to-end fixture that verifies dry-run, apply, propose, reject, evidence
 lookup, and raw source artifact immutability.
 
-Missing or thin coverage:
+Remaining thin coverage:
 
-- Route-level tests for `/api/ontology/operations/*` auth middleware.
 - Full end-to-end daemon + CLI process fixture using a real running daemon.
 - `graph.yaml` policy validation, because no active policy file is implemented
   in this slice.
--- a/docs/knowledge-graph-control-plane.md
+++ b/docs/knowledge-graph-control-plane.md
@@ -1,0 +1,127 @@
+# Knowledge Graph Control Plane Inventory
+
+This is an implementation inventory for the current ontology and knowledge
+graph control plane. It is not a speculative product spec.
+
+## Tables And Migrations
+
+| Surface | Migration | Exposure | Mutation model | Notes |
+|---|---:|---|---|---|
+| `entities` | 002, 005, 019, 022, 027, 037, 064, 070 | CLI/API | proposal/audit-backed through ontology operations; internal pipeline writes still exist | Agent-scoped after 019. 070 adds `status`, archive metadata, and proposal provenance. Name/canonical selectors must resolve to one same-agent active row. |
+| `relations` | 002, 005 | hidden/internal | internal graph legacy | Kept for legacy entity relation compatibility. New control-plane links use `entity_dependencies`. |
+| `memory_entity_mentions` | 002, 005 | hidden/internal | internal write-time linker | Links memories to entities; source memories remain immutable provenance. |
+| `entity_aspects` | 019, 070 | CLI/API | proposal/audit-backed through ontology operations | 070 adds archive metadata and proposal provenance. Default reads should treat archived aspects as hidden. |
+| `entity_attributes` | 019, 059, 060, 064, 067, 070 | CLI/API | proposal/audit-backed for operations; pipeline/supersession helpers also write | Claim slots use `group_key` + `claim_key`. 070 adds first-class versions: `version`, `version_root_id`, `previous_attribute_id`. Default reads show active rows; history reads can inspect all versions. |
+| `entity_dependencies` | 019, 031, 036, 039, 050, 064, 067, 070 | CLI/API | proposal/audit-backed through ontology operations; structural workers also write | Link operations create/update/archive rows here. Reason/confidence/proposal/source provenance are preserved. |
+| `entity_dependency_history` | 050 | hidden/internal | audit trigger | Trigger-backed history for dependency insert/update/delete. |
+| `task_meta` | 019, 054 | hidden/internal | task/procedural helpers | Agent-scoped task metadata, not part of this control-plane CLI slice. |
+| `ontology_proposals` | 067 | CLI/API | proposal loop and applied audit ledger | Pending, applied, rejected, and failed proposals. Direct operations create applied rows inside the same transaction as graph mutation. |
+| `dreaming_state`, `dreaming_passes` | 055 | CLI/API via dream status/trigger | dreaming worker | Existing dreaming pass records/status. LLM dreaming remains proposal-first for ontology changes in this slice. |
+| `memory_artifacts` | 051, 061, 062 | API/internal | immutable source artifact records | Source-backed evidence for proposals and applied graph rows. Ontology updates must not rewrite these rows. |
+| `session_transcripts` | 040, 045, 047 | API/internal | immutable transcript/index records | Proposal extraction can cite transcripts; graph mutation must preserve transcript provenance. |
+
+## Daemon Routes
+
+| Route | Exposure | Mutation model | Tests / risk |
+|---|---|---|---|
+| `GET /api/ontology/proposals` | CLI/API | read-only | Covered by ontology proposal tests. |
+| `GET /api/ontology/proposals/:id` | CLI/API | read-only | Covered by CLI/proposal tests. |
+| `POST /api/ontology/proposals` | CLI/API | creates pending proposal | Covered by proposal import/create tests. |
+| `POST /api/ontology/proposals/batch` | CLI/API | creates pending proposals atomically | Covered by batch proposal tests. |
+| `POST /api/ontology/proposals/:id/apply` | CLI/API | applies pending proposal | Covered by proposal apply tests. |
+| `POST /api/ontology/proposals/:id/reject` | CLI/API | rejects pending proposal | Covered by reject tests. |
+| `GET /api/ontology/proposals/:id/evidence` | CLI/API | read-only | Covered by evidence tests. |
+| `GET /api/ontology/proposals/conflicts` | CLI/API | read-only | Covered by conflict tests. |
+| `POST /api/ontology/proposals/repair/duplicates` | CLI/API | dry-run or pending proposals | Covered by duplicate repair tests. |
+| `POST /api/ontology/extract` | CLI/API | dry-run or pending proposals | Proposal-first. Covered by extraction tests. |
+| `POST /api/ontology/consolidate` | CLI/API | dry-run or pending proposals | Proposal-first. Covered by consolidation tests. |
+| `POST /api/ontology/operations/apply` | CLI/API | dry-run, pending proposal, or applied proposal + graph mutation | New direct operation endpoint. Requires `modify`. Covered through operation engine and CLI tests. |
+| `POST /api/ontology/operations/batch` | CLI/API | atomic batch dry-run/propose/apply | Requires `modify`; rolls back on invalid operation. Covered by operation batch tests. |
+| `GET /api/ontology/claims/evidence` | CLI/API | read-only | Covered by claim evidence tests. |
+| `GET /api/ontology/claims/versions` | CLI/API | read-only | Covered by version chain tests. |
+| `GET /api/ontology/claims/version` | CLI/API | read-only | Covered by version show tests. |
+| `GET /api/ontology/links/:id/evidence` | CLI/API | read-only | Covered by link evidence tests. |
+| `GET /api/knowledge/navigation/*` | CLI/API | read-only | Existing knowledge navigation tests cover entity/aspect/group/claim browsing. |
+| `GET /api/pipeline/status`, `GET /api/status` | CLI/API | read-only | `ontology pipeline *` uses these to explain graph mutation state. |
+| `GET /api/dream/status`, `POST /api/dream/trigger` | CLI/API | dreaming worker | Existing dream CLI/daemon surface. Ontology dreaming skill uses proposal JSON/JSONL by default. |
+
+## CLI Commands
+
+| Command | Exposure | Mutation model | Notes |
+|---|---|---|---|
+| `signet ontology proposals` | CLI | read-only | Existing proposal list. |
+| `signet ontology proposal <id>` | CLI | read-only | Existing proposal show. |
+| `signet ontology evidence <id>` | CLI | read-only | Existing proposal evidence. |
+| `signet ontology apply <id>` | CLI | applies pending proposal | Preserves proposal history and agent scope. |
+| `signet ontology reject <id>` | CLI | rejects pending proposal | Preserves rejected proposal history. |
+| `signet ontology conflicts` | CLI | read-only | Lists pending claim conflicts. |
+| `signet ontology entity create/rename/merge/archive` | CLI | direct operation endpoint | Supports `--dry-run`, `--propose`, `--json`, `--agent`, `--actor`, `--reason`, `--evidence-file`. |
+| `signet ontology claim set/versions/show/archive/restore` | CLI | operation endpoint for writes, read endpoints for versions | `set` creates version chains; `restore` only required for claim versions in this slice. |
+| `signet ontology aspect create/rename/archive` | CLI | direct operation endpoint | Uses same audited operation path. |
+| `signet ontology link create/update/archive` | CLI | direct operation endpoint | Uses `entity_dependencies`. |
+| `signet ontology stream apply <path|- >` | CLI | atomic JSONL operation batch | Supports file and stdin, `--dry-run`, `--propose`, `--json`, `--agent`, `--actor`. |
+| `signet ontology pipeline status/config/explain` | CLI | graph-state inspection | Explains Pipeline V2 graph flags and write gates. |
+| `signet ontology config show/validate/explain` | CLI | control-plane inspection | Confirms audited operation tools are usable and no external `graph.yaml` policy gate is active in this slice. |
+| `signet ontology extract` | CLI | dry-run/pending proposals | Existing proposal-first extraction surface. |
+| `signet ontology consolidate` | CLI | dry-run/pending proposals | Existing proposal-first consolidation surface. |
+| `signet dream status/trigger` | CLI | existing dream worker | Existing pass status and trigger behavior. |
+
+## Pipeline V2 Graph Knobs
+
+Current graph-affecting configuration is loaded from `memory.pipelineV2` in
+`agent.yaml` and surfaced through `signet ontology pipeline *`:
+
+- `enabled`
+- `paused`
+- `shadowMode`
+- `mutationsFrozen`
+- `graph.enabled`
+- `graph.extractionWritesEnabled`
+- traversal mode and bounds (`traversal.enabled`, `primary`, hop/branch/path limits, confidence, timeout, boost)
+- dampening settings
+- autonomous maintenance (`autonomous.enabled`, `frozen`, `allowUpdateDelete`, mode/polling)
+- extraction provider/workload settings
+- write gates such as `minFactConfidenceForWrite`
+- queue/worker health through `/api/pipeline/status`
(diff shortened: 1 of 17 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm taking a look at the feature work in `feat(ontology): add audited control plane` (commit `1049660a`) and will follow up shortly.

platform/daemon/src/ontology-proposals.ts
```diff
@@ -882,6 +1474,79 @@ export function listOntologyProposalConflicts(
...
+	const kind = params.kind ?? "attribute";
+	return accessor.withReadDb((db) => {
+		const entityKey = canonical(params.entity);
+		const aspectKey = canonical(params.aspect);
+		const rows = db
+			.prepare(
+				`SELECT attr.*
+				 FROM entity_attributes attr
+				 JOIN entity_aspects asp ON asp.id = attr.aspect_id
+				 JOIN entities ent ON ent.id = asp.entity_id
```
PR-Reviewer-Ant: This makes version/history inspection unavailable once the parent entity is archived. The PR description says archived graph rows are hidden from default active reads while preserving history/version inspection, but `listClaimVersions()` and `getClaimVersion()` now require `ent.status = active` and `asp.status = active`. After `archive_entity` or `archive_aspect`, operators can no longer inspect the claim versions under that archived row, which is exactly the audit trail this feature is adding. Version/history reads should probably resolve within the agent scope without the active parent-row filter, or expose an explicit status mode.

platform/daemon/src/knowledge-graph.ts
```diff
@@ -460,7 +484,8 @@ export function getDependenciesFrom(
 		const rows = db
 			.prepare(
 				`SELECT * FROM entity_dependencies
```
PR-Reviewer-Ant: `getDependenciesFrom()` now hides archived dependency rows, but it still returns active dependencies whose target entity has been archived. The list/detail/stats queries added joins to filter both source and target entity status, so this helper now diverges from the default active graph semantics and can leak archived entities back into dependency views that call it directly.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1049660a`

</details>

I found two consistency bugs in the new archive filtering. The control plane mostly routes mutations through the audited proposal path, but a couple of read surfaces still stop history or active views from matching the PR’s stated behavior.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both issues are directly visible in the changed SQL. The PR explicitly claims default active-read filtering while preserving history/version inspection, but the new claim-version query filters out archived parent entities/aspects, and the dependency helper filters archived links without filtering archived endpoint entities.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c588ea47`

</details>

I did not find a concrete correctness, security, data-integrity, or migration issue worth flagging from the provided diff. The implementation appears to line up with the PR description: new audited operation paths go through `ontology_proposals`, dry-run/propose/apply modes are covered, and archived rows are hidden from the default graph reads shown here.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The provided context includes the new migration, daemon operation engine, routes, focused tests, and graph read filtering, which are enough to validate the main control-plane behavior. The CLI file content is truncated after the new command tests begin, so I could not fully inspect every command handler implementation even though the visible tests exercise the expected endpoints.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

platform/daemon/src/ontology-proposals.ts
```diff
@@ -494,6 +652,127 @@ function applyAddClaimValue(
...
+			   AND claim_key = ?
+			   AND status = 'active'
+			 ORDER BY version DESC, updated_at DESC`,
+		)
+		.all(aspectId, agentId, kind, groupKey, claimKey) as Array<{
+		id: string;
+		content: string;
+		normalized_content: string;
+		version: number | null;
+		version_root_id: string | null;
```
PR-Reviewer-Ant, line 685: `set_claim_value` only queries `status = 'active'` rows when deciding the previous version/root. If an operator archives the active claim and later sets the same entity/aspect/group/claim again, `active.length` is 0, so this creates a fresh v1/root instead of continuing the existing version history. That breaks the PR's stated claim that claim versions remain first-class/queryable history. The version lookup should consider the full slot history for `version_root_id` and next version, while still only superseding the currently active row.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `ecc096af`

</details>

I found one data-integrity issue in claim versioning: setting a claim after its active version has been archived starts a new version chain instead of preserving the existing history.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The versioning behavior is visible in the changed mutation engine: set_claim_value only looks at active rows before choosing the next version/root. I did not run a local repro because the provided temporary checkout does not contain the Signet tree.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

skills/dreaming/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+and turn it into Signet ontology structure. The job is flexible bulk ingestion:
+transcripts, memory artifacts, source artifacts, notes, summaries, and imported
+records go in; evidence-backed ontology operations come out.
+
+The ontology control plane is the only write path. Generated work should fill
+in entities, aspects, groups, claims, attributes, and links by producing
+operation JSON/JSONL for the daemon-backed CLI/API. It must not silently save
+ordinary memories, rewrite source artifacts, or edit SQLite directly.
+
+Default mode is proposal-first. Start with `--dry-run`, then write pending
```
NicholaiVogel: this is unnecessary

platform/daemon/src/ontology-proposals.ts
```diff
@@ -494,6 +652,128 @@ function applyAddClaimValue(
...
+		kind: string;
+		status: string;
+	}>;
+	const active = slot.filter((row) => row.status === "active");
+	const normalized = canonical(value);
```
PR-Reviewer-Ant, line 690: This dedupe path overwrites `proposal_id` and `proposal_evidence` on the existing active attribute when the same claim value is applied again. That loses the original proposal/source provenance for the graph row, which is exactly the audit trail this PR is trying to make reliable. A repeated dry/operator apply with new evidence should create an audit proposal without rewriting the row's original provenance, or append/link the new evidence somewhere history-preserving instead of replacing it.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `ed59f973`

</details>

I found one provenance bug in the direct operation path: idempotent claim writes overwrite the existing claim row's audit proposal instead of preserving the original source of the graph fact.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The overwrite is visible in the new set_claim_value handler, where a deduped active claim updates entity_attributes.proposal_id/proposal_evidence in place. I did not have the full untruncated route/CLI diff, but the data-integrity issue is contained in the provided ontology-proposals.ts changed lines.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

skills/dreaming/SKILL.md
```diff
@@ -0,0 +1,168 @@
+---
+name: dreaming
+description: "Ingest memory, transcript, and source artifacts into Signet's living ontology by producing evidence-backed proposal JSON/JSONL through the audited ontology control plane."
```
NicholaiVogel: the point isn't to create json, it's to maintain the graph.

skills/dreaming/SKILL.md
NicholaiVogel: doing dry run first just wastes tokens. The majority of users aren't dealing with medical records, they're just working on memories. dry run should not be enforced.

skills/dreaming/SKILL.md
```diff
@@ -0,0 +1,168 @@
...
+
+Use this skill when an agent should wake up, read accumulated source evidence,
+and turn it into Signet ontology structure. The job is flexible bulk ingestion:
+transcripts, memory artifacts, source artifacts, notes, summaries, and imported
+records go in; evidence-backed ontology operations come out.
+
+The ontology control plane is the only write path. Generated work should fill
+in entities, aspects, groups, claims, attributes, and links by producing
+operation JSON/JSONL for the daemon-backed CLI/API. It must not silently save
+ordinary memories, rewrite source artifacts, or edit SQLite directly.
```
NicholaiVogel: it should be saving memories, just not thru the api's remember endpoint.

platform/daemon/src/ontology-proposals.ts
```diff
@@ -882,6 +1484,77 @@ export function listOntologyProposalConflicts(
...
+	const groupKey = canonicalKey(params.group) ?? "general";
+	const claimKey = canonicalKey(params.claim);
+	if (claimKey === null) throw new OntologyProposalError("claim is required", 400);
+	const kind = params.kind ?? "attribute";
+	return accessor.withReadDb((db) => {
+		const entityKey = canonical(params.entity);
+		const aspectKey = canonical(params.aspect);
+		const rows = db
+			.prepare(
+				`SELECT attr.*
```
PR-Reviewer-Ant, line 1559: This history read does not enforce the strict selector behavior the PR documents. The query matches by entity id/canonical/name and aspect canonical/name, then returns all matching attributes. If an archived entity/aspect and a re-created active one share the same canonical selector, `claim versions` can blend both histories, and `claim show --version 1` can pick an arbitrary matching row. The operation paths added above reject ambiguous selectors; the version/history path should either resolve the entity/aspect strictly first or require ids when the selector is ambiguous, while still allowing explicit archived-row inspection.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a399bb75`

</details>

I found one correctness issue in the claim version reader: selector-based history reads can silently merge claim versions across multiple matching entities/aspects, which undermines the PR’s stated strict-selector behavior.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The issue is visible in the new listClaimVersions query: it matches entity/aspect names directly and returns every matching row instead of resolving to one entity/aspect or rejecting ambiguity. I could not inspect the full repository beyond the supplied PR prompt, so I am not confirming whether an outer CLI layer already prevents this for every caller.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

skills/dreaming/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+Use this skill when an agent should wake up, read accumulated source evidence,
+and turn it into Signet ontology structure. The job is flexible bulk ingestion:
+transcripts, memory artifacts, source artifacts, notes, summaries, and imported
+records go in; the knowledge graph, scoped memories, and maintenance trail get
+better.
+
+The point is to maintain the graph, not to create JSON. JSON/JSONL is only the
+transport for audited ontology operations when the daemon-backed CLI/API needs a
+batch. Successful graph mutations must go through the ontology control plane so
+they are permission-checked and auditable.
```
NicholaiVogel: I don't know if this really needs to be said, honestly. Agents' aren't going to get the impression that they're just generating json. Worth removing

skills/dreaming/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+
+Dreaming may save memories when the evidence supports durable recall, but not by
+calling the API `remember` endpoint. Save explicit source-backed memory artifacts
+or use the configured source/import machinery so provenance remains inspectable.
+Do not rewrite raw transcript/source artifacts or edit SQLite directly.
```
NicholaiVogel, line 23: the WHOLE point, is to observe memory artifacts and then CREATE attributes as memories. memory artifacts ARE the evidence FOR attributes.

docs/knowledge-graph-control-plane.md
```diff
@@ -0,0 +1,127 @@
...
+
+| Route | Exposure | Mutation model | Tests / risk |
+|---|---|---|---|
+| `GET /api/ontology/proposals` | CLI/API | read-only | Covered by ontology proposal tests. |
+| `GET /api/ontology/proposals/:id` | CLI/API | read-only | Covered by CLI/proposal tests. |
+| `POST /api/ontology/proposals` | CLI/API | creates pending proposal | Covered by proposal import/create tests. |
+| `POST /api/ontology/proposals/batch` | CLI/API | creates pending proposals atomically | Covered by batch proposal tests. |
+| `POST /api/ontology/proposals/:id/apply` | CLI/API | applies pending proposal | Covered by proposal apply tests. |
+| `POST /api/ontology/proposals/:id/reject` | CLI/API | rejects pending proposal | Covered by reject tests. |
+| `GET /api/ontology/proposals/:id/evidence` | CLI/API | read-only | Covered by evidence tests. |
```
PR-Reviewer-Ant, line 33: This documents `/api/ontology/operations/apply` as a live daemon route requiring `modify`, and the PR description says the CLI/API operations are usable as soon as this merges. But the changed-file list in this PR does not include the daemon route module or any CLI command wiring, only the operation engine in `platform/daemon/src/ontology-proposals.ts` plus tests/docs. As-is, the feature appears to be implemented as internal functions but not exposed through the claimed API/CLI surface, so the public contract in the PR description is not actually delivered.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f2a74968`

</details>

The core operation engine and migration work are present, but the PR’s public contract is not: the diff does not add the claimed daemon routes or CLI command wiring, while the docs and PR description say those surfaces are live after merge.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The provided diff only changes docs, migration/types, knowledge graph reads, ontology proposal internals, and tests. I do not see changes to daemon route registration or surfaces/cli command files, yet docs/knowledge-graph-control-plane.md documents new POST /api/ontology/operations/* routes as available and permission-checked. The ontology-proposals diff is truncated, so I am not claiming anything about omitted internals beyond the missing route/CLI files in the changed-file list.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

docs/knowledge-graph-control-plane.md
NicholaiVogel: Verified and added explicit guardrails in 86b3c6418.

The public surface is present in this PR:
- daemon route registration: `platform/daemon/src/daemon.ts` mounts `registerOntologyRoutes(app)`
- operation routes: `platform/daemon/src/routes/ontology-routes.ts` exposes `POST /api/ontology/operations/apply` and `/batch` behind `modify`
- CLI registration: `surfaces/cli/src/cli.ts` calls `registerOntologyCommands(program, ...)`
- CLI commands: `surfaces/cli/src/commands/ontology.ts` exposes `ontology entity|claim|aspect|link|stream`

I added regression coverage for the part that was genuinely thin: route-level auth assertions for `/api/ontology/operations/*`, plus a CLI command-tree smoke test so the shipped command surface cannot disappear silently. Focused suite now passes: `129 pass, 0 fail`.

platform/daemon/src/dreaming-skill.test.ts
```diff
@@ -0,0 +1,30 @@
...
+		expect(content).toContain("transcripts, memory artifacts, source artifacts, notes, summaries, and imported");
+		expect(content).toContain("entities, aspects, groups, claims, attributes, and links");
+		expect(content).toContain("recently saved memory artifacts");
+		expect(content).toContain("flexible bulk ingestion");
+		expect(content).toContain("Memory artifacts are evidence\nfor attributes");
+		expect(content).toContain("source-backed memory artifacts");
+		expect(content).toContain("not the API\n  `remember` endpoint");
+		expect(content).toContain("signet ontology stream apply ops.jsonl --json");
+		expect(content).toContain("signet ontology stream apply proposals.jsonl --propose --json");
+		expect(content).toContain("Use dry-run only when the operator asks");
```
PR-Reviewer-Ant, line 20: This assertion codifies the opposite of the PR's safety model. The PR description and `docs/dreaming-skill-runbook.md` describe generated/dreaming ontology maintenance as dry-run/proposal-first, but this test requires the built-in skill to say `Use dry-run only when the operator asks` and also requires a plain `signet ontology stream apply ops.jsonl --json` command. That makes LLM-generated bulk ingestion default to live graph mutation unless the operator remembers to opt out, which is exactly the data-corruption path the audited control plane is supposed to avoid. The skill should default generated output to `--dry-run` or `--propose`, with direct apply reserved for explicit operator-authored operations.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `86b3c641`

</details>

I found one blocker: the built-in dreaming skill is being tested to prefer live writes unless an operator asks for dry-run, which conflicts with the PR's proposal-first safety story for generated ontology maintenance.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The PR description and new runbook say LLM-generated dreaming output should use dry-run or propose by default, but the new dreaming skill test explicitly requires the skill to contain a live apply command and the phrase "Use dry-run only when the operator asks". The full skill file content is not included in the prompt, so I am grounding this on the test that locks in the behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR #733, opened 2026-05-20
fix(memory): persist remember row provenance

dragontvstaff:
## Summary
- Add row-level provenance handling to `/api/memory/remember` for `sourcePath`/`source_path`, `runtimePath`/`runtime_path`, and `idempotencyKey`/`idempotency_key`.
- Persist the provenance fields through the live daemon write path as first-class memory row columns via `txIngestEnvelope`.
- Use `idempotency_key` for remember dedupe before content-hash dedupe, including stable per-chunk keys.
- Expose `sourcePath` through the shared core recall helper and SDK remember payload types.
- Update `docs/API.md` to document the provenance request fields and idempotency-key dedupe behavior.

## Changed files
- `docs/API.md`
- `libs/sdk/src/index.ts`
- `platform/core/src/recall.ts`
- `platform/core/src/recall.test.ts`
- `platform/daemon/src/routes/memory-routes.ts`
- `platform/daemon/src/transactions.ts`
- `platform/daemon/src/mutation-api.test.ts`

## Test plan
- `cd platform/daemon && bun test src/mutation-api.test.ts` — PASS, 20/20 tests.
- `cd platform/core && bun test src/recall.test.ts` — PASS, 4/4 tests.
- `cd libs/sdk && bun run build` — PASS.
- `cd platform/core && bun run build` — PASS.
- `cd platform/daemon && bun run build` — PASS; existing Svelte a11y/deprecation warnings were emitted by dashboard files outside this change.

## Soul Vessel M10 note
The regression test `POST /api/memory/remember persists row-level provenance from metadata` demonstrates the live daemon remember path now persists `source_path`, `runtime_path`, and `idempotency_key` and dedupes retries by `idempotency_key`; this test failed before the fix when metadata-supplied row provenance was not stored.

The Soul Vessel opt-in live diagnostic from PR #26 was not rerun during this PR-packaging pass. M10 should be unblocked after this upstream fix is reviewed/merged and the disposable live diagnostic is rerun green; until Seer review and that final diagnostic, M10 remains gated.

No production Signet DB/workspace, live Hermes profile, or Soul Vessel `SPEC.md` was touched.

### DIFF
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -3,7 +3,7 @@ import type { Hono } from "hono";
 import { getAgentScope, resolveAgentId } from "../agent-id";
 import { checkScope, requirePermission, requireRateLimit } from "../auth";
 import { normalizeAndHashContent } from "../content-normalization";
-import { getDbAccessor } from "../db-accessor";
+import { type WriteDb, getDbAccessor } from "../db-accessor";
 import { syncVecDeleteBySourceId, syncVecInsert, vectorToBlob } from "../db-helpers";
 import { fetchEmbedding } from "../embedding-fetch";
 import { buildEmbeddingHealth } from "../embedding-health";
@@ -86,6 +86,21 @@ interface RememberRowProvenance {
 	readonly idempotencyKey?: string;
 }
 
+interface RememberDedupeScope {
+	readonly agentId: string;
+	readonly visibility: "global" | "private" | "archived";
+	readonly scope: string | null;
+}
+
+interface RememberDedupeRow {
+	readonly id: string;
+	readonly type: string;
+	readonly tags: string | null;
+	readonly pinned: number;
+	readonly importance: number;
+	readonly content: string;
+}
+
 function pickOptionalString(...values: readonly unknown[]): string | undefined {
 	for (const value of values) {
 		const parsed = parseOptionalString(value);
@@ -97,7 +112,14 @@ function pickOptionalString(...values: readonly unknown[]): string | undefined {
 function parseRememberRowProvenance(body: Record<string, unknown>): RememberRowProvenance {
 	const metadata = toRecord(body.metadata) ?? {};
 	return {
-		sourcePath: pickOptionalString(body.sourcePath, body.source_path, body.source, metadata.sourcePath, metadata.source_path, metadata.source),
+		sourcePath: pickOptionalString(
+			body.sourcePath,
+			body.source_path,
+			body.source,
+			metadata.sourcePath,
+			metadata.source_path,
+			metadata.source,
+		),
 		runtimePath: pickOptionalString(body.runtimePath, body.runtime_path, metadata.runtimePath, metadata.runtime_path),
 		idempotencyKey: pickOptionalString(
 			body.idempotencyKey,
@@ -110,6 +132,50 @@ function parseRememberRowProvenance(body: Record<string, unknown>): RememberRowP
 
 function idempotencyKeyForChunk(baseKey: string | undefined, index: number): string | undefined {
 	return baseKey ? `${baseKey}:chunk:${index + 1}` : undefined;
+}
+
+function scopedMemoryPredicate(input: RememberDedupeScope): {
+	readonly sql: string;
+	readonly params: readonly string[];
+} {
+	return input.scope === null
+		? {
+				sql: "agent_id = ? AND visibility = ? AND scope IS NULL",
+				params: [input.agentId, input.visibility],
+			}
+		: {
+				sql: "agent_id = ? AND visibility = ? AND scope = ?",
+				params: [input.agentId, input.visibility, input.scope],
+			};
+}
+
+function getScopedIdempotencyMemoryId(
+	db: WriteDb,
+	key: string | undefined,
+	input: RememberDedupeScope,
+): { readonly id: string } | undefined {
+	if (!key) return undefined;
+	const scoped = scopedMemoryPredicate(input);
+	return db
+		.prepare(`SELECT id FROM memories WHERE idempotency_key = ? AND ${scoped.sql} AND is_deleted = 0 LIMIT 1`)
+		.get(key, ...scoped.params) as { readonly id: string } | undefined;
+}
+
+function getScopedIdempotencyDedupeRow(
+	db: WriteDb,
+	key: string | undefined,
+	input: RememberDedupeScope,
+): RememberDedupeRow | undefined {
+	if (!key) return undefined;
+	const scoped = scopedMemoryPredicate(input);
+	return db
+		.prepare(
+			`SELECT id, type, tags, pinned, importance, content
+			 FROM memories
+			 WHERE idempotency_key = ? AND ${scoped.sql} AND is_deleted = 0
+			 LIMIT 1`,
+		)
+		.get(key, ...scoped.params) as RememberDedupeRow | undefined;
 }
 
 function hasMemoriesSessionIdColumn(db: any): boolean {
@@ -558,6 +624,7 @@ export function registerMemoryRoutes(app: Hono): void {
 		const rowProvenance = parseRememberRowProvenance(body as Record<string, unknown>);
 		const agentId = resolveAgentId({ agentId: body.agentId, sessionKey: c.req.header("x-signet-session-key") });
 		const visibility = body.visibility === "private" ? "private" : "global";
+		const dedupeScope = { agentId, visibility, scope };
 		const hasBodyTags = Object.prototype.hasOwnProperty.call(body, "tags");
 		const bodyTags = hasBodyTags ? parseTagsMutation(body.tags) : undefined;
 		if (hasBodyTags && bodyTags === undefined) {
@@ -620,12 +687,7 @@ export function registerMemoryRoutes(app: Hono): void {
 
 				try {
 					const inserted = getDbAccessor().withWriteTx((db) => {
-						const byIdempotencyKey = chunkIdempotencyKey
-							? (db
-									.prepare("SELECT id FROM memories WHERE idempotency_key = ? AND is_deleted = 0 LIMIT 1")
-									.get(chunkIdempotencyKey) as { id: string } | undefined)
-							: undefined;
-						if (byIdempotencyKey) return false;
+						if (getScopedIdempotencyMemoryId(db, chunkIdempotencyKey, dedupeScope)) return false;
 
 						const byHash =
 							scope !== null
@@ -784,14 +846,7 @@ export function registerMemoryRoutes(app: Hono): void {
 		const contentHash = normalizedContent.contentHash;
 		const pipelineEnqueueEnabled = pipelineCfg.enabled;
 
-		type DedupeRow = {
-			id: string;
-			type: string;
-			tags: string | null;
-			pinned: number;
-			importance: number;
-			content: string;
-		};
+		type DedupeRow = RememberDedupeRow;
 
 		try {
 			const result = getDbAccessor().withWriteTx((db) => {
@@ -815,15 +870,8 @@ export function registerMemoryRoutes(app: Hono): void {
 					if (bySource) return { deduped: true as const, row: bySource };
 				}
 
-				if (rowProvenance.idempotencyKey) {
-					const byIdempotencyKey = db
-						.prepare(
-							`SELECT id, type, tags, pinned, importance, content
-					 FROM memories WHERE idempotency_key = ? AND is_deleted = 0 LIMIT 1`,
-						)
-						.get(rowProvenance.idempotencyKey) as DedupeRow | undefined;
-					if (byIdempotencyKey) return { deduped: true as const, row: byIdempotencyKey };
-				}
+				const byIdempotencyKey = getScopedIdempotencyDedupeRow(db, rowProvenance.idempotencyKey, dedupeScope);
+				if (byIdempotencyKey) return { deduped: true as const, row: byIdempotencyKey };
 
 				// Check content_hash dedupe (scope-aware)
 				const byHash = (
@@ -895,16 +943,8 @@ export function registerMemoryRoutes(app: Hono): void {
 			const msg = e instanceof Error ? e.message : "";
 			if (msg.includes("UNIQUE constraint")) {
 				const existing = getDbAccessor().withReadDb((db) => {
-					if (rowProvenance.idempotencyKey) {
-						const byIdempotencyKey = db
-							.prepare(
-								`SELECT id, type, tags, pinned, importance, content
-						 FROM memories
-						 WHERE idempotency_key = ? AND is_deleted = 0 LIMIT 1`,
-							)
-							.get(rowProvenance.idempotencyKey) as DedupeRow | undefined;
-						if (byIdempotencyKey) return byIdempotencyKey;
-					}
+					const byIdempotencyKey = getScopedIdempotencyDedupeRow(db, rowProvenance.idempotencyKey, dedupeScope);
+					if (byIdempotencyKey) return byIdempotencyKey;
 					return db
 						.prepare(
 							`SELECT id, type, tags, pinned, importance, content
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -815,6 +815,19 @@ export function registerMemoryRoutes(app: Hono): void {
 					deduped: true,
 				});
 			}
+			const contentHashes = new Set<string>();
+			for (const plan of chunkPlans) {
+				if (contentHashes.has(plan.normalized.contentHash)) {
+					return c.json({ error: "chunked content contains duplicate chunks" }, 409);
+				}
+				contentHashes.add(plan.normalized.contentHash);
+				const byHash = getDbAccessor().withReadDb((db) =>
+					getScopedContentHashMemoryId(db, plan.normalized.contentHash, dedupeScope),
+				);
+				if (byHash) {
+					return c.json({ error: "chunk content already exists for this agent and scope" }, 409);
+				}
+			}
 
 			const groupId = chunkGroupIdForIdempotencyKey(rowProvenance.idempotencyKey, dedupeScope) ?? crypto.randomUUID();
 			const now = new Date().toISOString();
@@ -842,9 +855,6 @@ export function registerMemoryRoutes(app: Hono): void {
 					const result = getDbAccessor().withWriteTx((db) => {
 						const byIdempotencyKey = getScopedIdempotencyMemoryId(db, plan.idempotencyKey, dedupeScope);
 						if (byIdempotencyKey) return { id: byIdempotencyKey.id, inserted: false as const };
-
-						const byHash = getScopedContentHashMemoryId(db, plan.normalized.contentHash, dedupeScope);
-						if (byHash) return { id: byHash.id, inserted: false as const };
 
 						txIngestEnvelope(db, {
 							id: chunkId,
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -169,6 +169,19 @@ function scopedMemoryPredicate(input: RememberDedupeScope): {
 	};
 }
 
+function scopedContentHashPredicate(input: RememberDedupeScope): {
+	readonly sql: string;
+	readonly params: readonly string[];
+} {
+	return {
+		sql: `
+			COALESCE(NULLIF(agent_id, ''), 'default') = ?
+			AND COALESCE(scope, '__NULL__') = ?
+		`,
+		params: [input.agentId || "default", input.scope ?? "__NULL__"],
+	};
+}
+
 function getScopedIdempotencyMemoryId(
 	db: WriteDb,
 	key: string | undefined,
@@ -201,6 +214,38 @@ function getScopedIdempotencyDedupeRow(
 			 LIMIT 1`,
 		)
 		.get(key, ...scoped.params) as RememberDedupeRow | undefined;
+}
+
+function getScopedContentHashMemoryId(
+	db: WriteDb,
+	contentHash: string,
+	input: RememberDedupeScope,
+): { readonly id: string } | undefined {
+	const scoped = scopedContentHashPredicate(input);
+	return db
+		.prepare(
+			`SELECT id
+			 FROM memories
+			 WHERE content_hash = ? AND ${scoped.sql} AND is_deleted = 0
+			 LIMIT 1`,
+		)
+		.get(contentHash, ...scoped.params) as { readonly id: string } | undefined;
+}
+
+function getScopedContentHashDedupeRow(
+	db: WriteDb,
+	contentHash: string,
+	input: RememberDedupeScope,
+): RememberDedupeRow | undefined {
+	const scoped = scopedContentHashPredicate(input);
+	return db
+		.prepare(
+			`SELECT id, type, tags, pinned, importance, content
+			 FROM memories
+			 WHERE content_hash = ? AND ${scoped.sql} AND is_deleted = 0
+			 LIMIT 1`,
+		)
+		.get(contentHash, ...scoped.params) as RememberDedupeRow | undefined;
 }
 
 function getScopedChunkIdempotencyRows(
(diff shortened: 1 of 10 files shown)


PR-Reviewer-Ant:
Hi @dragontvstaff - I'm taking a look at the fixes in `fix(memory): persist remember row provenance` (commit `014299dc`) and will follow up shortly.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -762,6 +815,16 @@ export function registerMemoryRoutes(app: Hono): void {
 					if (bySource) return { deduped: true as const, row: bySource };
 				}
 
+				if (rowProvenance.idempotencyKey) {
+					const byIdempotencyKey = db
+						.prepare(
```
PR-Reviewer-Ant: This idempotency-key dedupe is global across the whole memories table. The neighboring sourceId/content-hash dedupe paths are scope-aware, but this lookup ignores `scope`, `agentId`, and `visibility`, so a retry/import in a different scope or agent can get back an unrelated existing row just because the stable import key matches. That is a data isolation bug for remember, and it also makes the documented `idempotencyKey` contract unsafe for importers that generate keys outside Signet's global namespace. Please apply the same scoping policy here and in the UNIQUE-constraint fallback lookup.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,17 +607,26 @@ export function registerMemoryRoutes(app: Hono): void {
 				return c.json({ error: "Failed to create chunk group" }, 500);
 			}
 
-			for (const chunk of chunks) {
+			for (let chunkIndex = 0; chunkIndex < chunks.length; chunkIndex += 1) {
+				const chunk = chunks[chunkIndex];
 				const chunkNormalized = normalizeAndHashContent(chunk);
 				if (!chunkNormalized.storageContent) continue;
 
 				const chunkId = crypto.randomUUID();
```
PR-Reviewer-Ant: The chunked path has the same global idempotency-key lookup. If two scoped imports produce the same base key, this skips inserting the chunk for the second scope and returns a response with only newly inserted chunk IDs, not the existing scoped chunks. That makes chunked remember retries/collisions lossy and diverges from the PR description that stable per-chunk keys dedupe retries to existing rows.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `014299dc`

</details>

The provenance columns are wired through the daemon write path, but the new idempotency-key dedupe is not scoped like the existing source/content-hash checks. That can make one agent or scope receive another row when keys collide.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed remember handler adds idempotency-key lookup queries that only filter by idempotency_key and is_deleted, while the adjacent sourceId and content_hash dedupe paths explicitly include scope conditions. The affected code is in the provided diff and directly changes the PR's claimed dedupe behavior.

NicholaiVogel:
Updated this PR to address the scoped idempotency-key review finding.

What changed:
- `idempotency_key` dedupe now matches only within the requested `agent_id`, `visibility`, and nullable `scope` tuple.
- Chunked remember writes use the same scoped idempotency lookup.
- Added migration 072 to replace the old global idempotency-key unique index with scoped uniqueness.
- Added regression coverage for cross-agent and cross-visibility idempotency-key reuse.
- Updated API docs to describe scoped retry behavior.

Validation:
- `bun test platform/daemon/src/mutation-api.test.ts` passed, 21/21.
- `bun test platform/core/src/migrations/migrations.test.ts` passed, 38/38.
- Scoped Biome check on touched files passed with existing warnings in `memory-routes.ts`.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,17 +674,21 @@ export function registerMemoryRoutes(app: Hono): void {
...
+				const chunkIdempotencyKey = idempotencyKeyForChunk(rowProvenance.idempotencyKey, chunkIndex);
 				const chunkContentForInsert =
```
PR-Reviewer-Ant: This makes chunked retries with an `idempotencyKey` silently skip already-inserted chunks instead of returning the existing rows. The route creates a new chunk-group entity before the loop, then `return false` here prevents the existing chunk id from being added to `chunkIds`; a full retry of the same oversized import can therefore report an empty or partial chunk result and leave a fresh orphan `chunk_group` entity. That diverges from the PR's stated `idempotencyKey` dedupe behavior and corrupts provenance for chunked imports. The chunk path needs to fetch and return the existing chunk ids, or dedupe the group before creating a new one.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cdce9601`

</details>

The row-level provenance path mostly matches the PR description, but the new idempotency handling breaks the oversized-memory chunk path on retries. A retry with the same key can create a fresh chunk group while inserting no chunks, which is not the promised dedupe behavior.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed chunking path derives per-chunk idempotency keys and returns false when an existing chunk is found, but the surrounding oversized-memory flow creates a new chunk_group before that check and does not return the existing chunk ids. I could prove this from the diff, but I did not run the daemon test suite in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -548,15 +665,53 @@ export function registerMemoryRoutes(app: Hono): void {
...
+						normalized,
+					};
+				})
+				.filter((plan): plan is NonNullable<typeof plan> => plan !== null);
+			if (chunkPlans.length === 0) {
+				return c.json({ error: "content produced no valid chunks" }, 400);
+			}
+
+			const existingChunks = getDbAccessor().withReadDb((db) => {
+				const rows = new Map<number, RememberDedupeIdRow>();
```
PR-Reviewer-Ant: This makes oversized remember idempotency depend on the *current* chunk count rather than the original request. If a caller retries the same `idempotencyKey` with edited or truncated content, and the first N chunk keys already exist, this branch can return a deduped success with only that subset of the original group. If the retry has more chunks, the later loop reuses `existingGroupId` and appends new chunks into the old import group. That turns an idempotent retry key into either a partial success or a mutation of the original import. The base idempotency key needs to represent the whole chunked request, or mismatched chunk plans should return the original full group or a clear conflict rather than mixing per-chunk state.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `54a1aeb7`

</details>

I found one data-integrity bug in the chunked remember idempotency path. The normal single-row provenance path looks aligned with the PR description, but chunk retries can now corrupt or misreport an import when the same idempotency key is reused with content that chunks differently.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunk path derives idempotency from per-chunk keys only, checks whether all current chunk keys already exist, and otherwise inserts missing chunks into the existing group. That behavior is directly visible in platform/daemon/src/routes/memory-routes.ts around the new existingChunks/groupId handling and does not require a runtime trace to prove the bad retry case.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/transactions.ts
```diff
@@ -218,8 +221,8 @@ export function txIngestEnvelope(db: WriteDb, mem: IngestEnvelope): string {
 		 (id, content, normalized_content, content_hash, who, why, project,
```
PR-Reviewer-Ant, line 221: This insert now always targets `runtime_path`, but the new migration does not add that column or declare it as a verified artifact. If a user upgrades from a schema that has `source_path`/`idempotency_key` but not `runtime_path`, every `txIngestEnvelope` call will fail with `no column named runtime_path`, including ordinary remember writes. The PR should add an upgrade-safe migration for `memories.runtime_path` or otherwise prove the column already exists on every supported schema path.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `35a492af`

</details>

I found one schema compatibility issue: the write path now always inserts into `runtime_path`, but this PR does not add or verify that column in the migrations. That makes the claimed provenance persistence brittle on databases that do not already have the column.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The changed insert in platform/daemon/src/transactions.ts explicitly writes `runtime_path`, while the new migration only ensures `agent_id`, `visibility`, `scope`, and `idempotency_key`. The provided migration registry shows `source_path` from migration 013 but no declared `runtime_path` artifact, so I cannot see a guaranteed schema path for this new column.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -762,6 +961,9 @@ export function registerMemoryRoutes(app: Hono): void {
 					if (bySource) return { deduped: true as const, row: bySource };
 				}
```
PR-Reviewer-Ant, line 1088: This does the new scoped idempotency lookup, but if the same content is remembered under the same `idempotencyKey` for a different `agentId`, the code falls through to the content-hash dedupe below, which is still only scope-aware. That means the second agent can get `{ deduped: true }` for the first agent's memory row instead of a new row in its own idempotency tuple, despite the new migration and docs making idempotency keys agent/visibility/scope scoped. The new test uses different content for the other-agent case, so it misses this. Please make the content-hash fallback use the same effective agent/scope predicate, or add a test proving same-content imports across agents do not collapse into one row.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `87f54017`

</details>

The provenance columns are threaded into the remember insert path, but the new agent-scoped idempotency behavior is undermined by the existing content-hash fallback. Identical content with the same idempotency key in another agent can still dedupe to the first agent's row instead of creating a scoped row.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The supplied diff shows idempotency lookup is scoped by agent/visibility/scope, but the immediate fallback content_hash lookup in platform/daemon/src/routes/memory-routes.ts still only checks content_hash and scope. I could not run the test suite because the provided cwd is not a git checkout containing the repository.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d4389ea9`

</details>

I did not find a concrete correctness, security, or data-integrity issue in this diff. The implementation matches the PR's main claims: provenance fields are persisted through remember, idempotency-key dedupe is checked before content-hash dedupe, chunk retries get stable per-chunk handling, and the schema/index change is covered by targeted tests.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed daemon route, transaction insert path, core helper, SDK payload, migration, docs, and focused tests were included, and the tests exercise the important provenance and idempotency behavior. I did not run the test suite or inspect live runtime behavior, so confidence is not high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a3bb9ec8`

</details>

I found one data-integrity race in the chunked remember path. The basic provenance persistence looks implemented, but concurrent idempotent chunked imports can still leave an orphan chunk group and return the wrong group id.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The race follows directly from the changed chunked path: it checks for existing chunks before creating a new group, creates a fresh chunk_group entity, then per chunk may discover rows inserted by another request via idempotency-key dedupe. In that case the response still uses the fresh groupId even though the memory rows belong to the other group's source_id.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -548,15 +750,65 @@ export function registerMemoryRoutes(app: Hono): void {
...
+			}
+
 			const groupId = crypto.randomUUID();
 			const now = new Date().toISOString();
-			const chunkIds: string[] = [];
+			const chunkIds = Array<string | undefined>(chunkPlans.length);
 
 			// Create chunk group entity
 			try {
 				getDbAccessor().withWriteTx((db) => {
```
PR-Reviewer-Ant: This still has a race for concurrent oversized remembers using the same `idempotencyKey`. Both requests can pass the preflight `existingChunks.length === 0` check, then both create a chunk-group entity. The loser then hits `getScopedIdempotencyMemoryId` inside the per-chunk transaction and reuses the winner's chunk ids, but this request still returns its freshly generated `groupId` and leaves that chunk_group orphaned. That makes the idempotent retry response lie about the group and corrupts the chunk-group graph under load. The group creation and chunk idempotency claim need to be serialized/claimed in one transaction, or the loser needs to derive `group_id` from the existing chunks it actually reused.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,39 +834,28 @@ export function registerMemoryRoutes(app: Hono): void {
...
-										.prepare(
-											"SELECT id FROM memories WHERE content_hash = ? AND scope IS NULL AND is_deleted = 0 LIMIT 1",
-										)
-										.get(chunkNormalized.contentHash) as { id: string } | undefined);
-						if (byHash) return false;
+					const result = getDbAccessor().withWriteTx((db) => {
+						const byIdempotencyKey = getScopedIdempotencyMemoryId(db, plan.idempotencyKey, dedupeScope);
+						if (byIdempotencyKey) return { id: byIdempotencyKey.id, inserted: false as const };
+
+						const byHash = getScopedContentHashMemoryId(db, plan.normalized.contentHash, dedupeScope);
```
PR-Reviewer-Ant: This fallback to content-hash dedupe can now produce a bogus chunk response. The group entity is created before the loop, but when a chunk already exists by content hash this branch returns `{ inserted: false }`; the outer loop records that existing id and immediately `continue`s, so no `memory_entity_mentions` row links the existing chunk to the new `groupId`. A caller importing the same oversized content with a different or absent `idempotencyKey` can get `{ chunked: true, group_id: <new group>, ids: <old chunks> }` even though that group has no chunk edges. Either reject/return the existing chunk group when chunk hashes already exist, or only report a new group after all returned ids are actually linked to it.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d38cb50e`

</details>

I found one data-integrity issue in the chunked remember path: content-hash dedupe can now return existing chunk ids under a newly-created group that is not linked to those chunks.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunk loop creates the chunk_group before processing chunks, then treats content-hash matches as successful chunk ids while skipping the relation insert because inserted is false. This is directly visible in platform/daemon/src/routes/memory-routes.ts and is not covered by the new idempotent retry tests, which only retry with the same idempotency key.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e2dd61e3`

</details>

I did not find a concrete correctness, security, or data-integrity issue in this diff. The implementation matches the stated provenance/idempotency behavior and adds focused regression coverage for the tricky remember and chunked-retry paths.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed daemon route threads source_path, runtime_path, and idempotency_key into txIngestEnvelope, the migration creates the scoped idempotency index, and the tests cover metadata provenance persistence, agent/visibility scoping, normalized legacy rows, chunked retries, and mixed chunk/non-chunk conflicts. I did not run the test suite locally, so this is based on the provided diff and full changed-file context.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in the #733 base branch.

The idempotency lookup and unique-constraint fallback now use normalized agent/visibility/scope predicates matching migration 072. The regression covers same-key reuse across agents/visibility and legacy normalized rows.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in the #733 base branch.

Chunk idempotency now uses scoped lookup, returns full existing chunk ids for exact retries, rejects changed/partial chunk plans with `409`, and preserves chunk id ordering in the response.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `54a1aeb76`, `35a492af0`, `87f540178`, and later hardening commits.

The route checks existing chunks before creating a group, exact retries return the original ids/group, changed content returns `409`, and chunk ids are tracked by index rather than append order.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `35a492af0` and `87f540178`.

The base key now represents the whole chunked request: existing chunks must match count, group, per-chunk key, and content hash. Mismatches return `409` instead of appending to or partially reporting the old import.

platform/daemon/src/transactions.ts
NicholaiVogel: Fixed in `d4389ea9d`.

Migration 072 now defensively adds `memories.runtime_path` and declares it as a verified migration artifact. Added a migration regression that drops `runtime_path`, reruns v72, and verifies the column is repaired.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `d4389ea9d`.

The remember content-hash fallback now uses the same normalized agent/scope predicate as `idx_memories_content_hash_unique`. The cross-agent idempotency regression now uses identical content so it would catch the old fallback leak.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `d38cb50e`.

Chunked requests with a base idempotency key now derive a deterministic chunk-group id from the normalized agent/visibility/scope/key tuple, so concurrent retries do not mint a fresh orphan group before discovering existing chunks.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `e2dd61e3`.

The chunk path no longer mixes arbitrary content-hash matches into a fresh chunk-group response. It preflights duplicate/pre-existing chunk content and returns `409` before creating a group; exact idempotent retries still reuse the complete existing chunk group.

NicholaiVogel:
Updated the branch and replied on each review thread.

Current head: `e2dd61e3`

Validation:
- `bun test platform/daemon/src/mutation-api.test.ts` — 26 pass on the base branch after the latest chunk hash/group fixes
- `bun test platform/core/src/migrations/migrations.test.ts` — 39 pass
- scoped Biome check — pass with the existing `memory-routes.ts` `noExplicitAny` warnings

Main changes since the first review pass:
- scoped idempotency and content-hash dedupe now match the normalized migration/index tuples
- migration 072 repairs/verifies `runtime_path`
- chunked idempotency now rejects changed, partial, mixed chunk/non-chunk, and ambiguous hash-reuse cases instead of returning misleading groups/ids

NicholaiVogel:
Rebased onto current `origin/main` (`f88f064c2`) and force-pushed.

Current head: `44e562559`

Validation after rebase:
- `bun test platform/daemon/src/mutation-api.test.ts` — 26 pass
- `bun test platform/core/src/migrations/migrations.test.ts` — 39 pass
- scoped Biome check — pass, with the existing `memory-routes.ts` `noExplicitAny` warnings

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `44e56255`

</details>

I did not find a concrete correctness, security, data-integrity, or breaking-change issue in this diff. The implementation matches the PR description: provenance is threaded into remember writes, idempotency-key dedupe is scoped, chunked retries are handled, and the behavior is covered by targeted daemon/core tests.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the route logic, transaction insert path, migration registration, SDK/core helper changes, API docs, and focused regression tests for provenance persistence, scoped idempotency, chunked retries, and migration behavior. I did not rerun the listed bun tests or exercise a live daemon, so confidence is medium rather than high.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -621,38 +903,41 @@ export function registerMemoryRoutes(app: Hono): void {
 							 VALUES (?, ?, 'chunk', 1.0, ?)`,
 						).run(chunkId, groupId, now);
 
-						return true;
+						return { id: chunkId, status: "inserted" };
```
PR-Reviewer-Ant: This chunked write path is no longer atomic. The handler creates the chunk-group entity before the loop, then commits each chunk in its own `withWriteTx`; if a later chunk hits `content_conflict`, a unique constraint race, embedding/linking-adjacent failure that bubbles here, or the final `savedChunkIds.length` check, the response is an error but the earlier entity/chunk rows remain committed. That leaves orphan or partial chunk groups, and a retry with the same idempotency key can then observe a mismatched partial set and keep returning conflict instead of repairing/deduping. The group row plus all memory rows and `memory_entity_mentions` links need to be committed as one transaction, or the error paths need explicit cleanup before returning.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `0cc8229a`

</details>

I found one data-integrity issue in the chunked remember path: the new idempotency/conflict handling can return an error after already committing part of the chunk group.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunk path creates the chunk-group entity in one write transaction, then inserts each chunk in separate write transactions, and returns 409/500 from inside the loop on conflicts or failures. That control flow is directly visible in platform/daemon/src/routes/memory-routes.ts and does not require a runtime trace to prove partial commits are possible.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Update for the latest stacked review feedback:

- Added `0cc8229a1` to close the chunk content-hash race called out on #734's base path.
- The chunk insert transaction now rechecks scoped `content_hash` before `txIngestEnvelope` and maps content-hash unique fallbacks to the intended 409 conflict response.
- Added a concurrent oversized remember regression covering same-content chunked writes with different idempotency keys.

Validation:
- `bunx biome check --write platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (same pre-existing `noExplicitAny` warnings in `memory-routes.ts`)

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `62aa171b0` on #733 and carried into this stacked branch.

The chunked remember write now performs the idempotency/content-hash rechecks, chunk-group creation, all chunk row inserts, and group mention inserts in one `withWriteTx`. If a conflict is detected, the transaction returns before any group or chunk rows are written. Embeddings, inline entity linking, and extraction enqueue remain post-commit non-fatal follow-up work.

I also strengthened the concurrent collision regression so the losing idempotency key must leave zero chunk rows and no deterministic chunk-group entity behind.

Validation:
- scoped Biome check still only reports the pre-existing `noExplicitAny` warnings in `memory-routes.ts`

NicholaiVogel:
Latest review follow-up is in.

Current head: `62aa171b0`

- Made chunked remember writes atomic across group creation, chunk rows, and group mention rows.
- Conflict paths now return before any chunk-group/chunk state is written.
- Strengthened the concurrent collision regression to verify the losing idempotency key leaves no chunk rows and no orphan chunk-group entity.

Validation:
- `bunx biome check --write platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (same pre-existing `noExplicitAny` warnings in `memory-routes.ts`)

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `62aa171b`

</details>

I did not find a concrete correctness, security, data-integrity, or migration issue worth flagging. The implementation matches the stated provenance persistence and idempotency-key dedupe behavior, including chunked remember retries and scoped uniqueness.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff includes the daemon write path, migration registration, migration tests, SDK/core helper changes, and targeted remember-route regression tests. The new txIngestEnvelope fields are wired into both single-row and chunked remember paths, and the idempotency-key lookups match the new agent/visibility/scope unique index.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #734, opened 2026-05-20
fix(memory): expose row provenance on read

dragontvstaff:
## Summary
- Extends `GET /api/memory/:id` to return row-level provenance columns: `source_path`, `runtime_path`, and `idempotency_key`.
- Adds camelCase read aliases (`sourcePath`, `runtimePath`, `idempotencyKey`) for callers that wrote provenance through the remember API.
- Updates daemon regression coverage, core/SDK types, and API docs for the read response shape.

## Changes
- `platform/daemon/src/routes/memory-routes.ts`: selects and serializes provenance fields on memory reads.
- `platform/daemon/src/mutation-api.test.ts`: covers read-after-remember provenance visibility.
- `platform/core/src/database.ts` and `platform/core/src/types.ts`: include provenance in memory row types.
- `libs/sdk/src/types.ts`: exposes provenance fields in SDK response types.
- `docs/API.md`: documents the read response fields.

## Type
- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected
- [x] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots
N/A; no UI changes.

## PR Readiness (MANDATORY)

No migration files are touched by this PR after #733 merged.

- [x] Migration is idempotent

Rollback / compatibility note: this PR only exposes fields that now exist on the persisted memory rows after #733. Reverting this PR removes the read/API/type exposure without changing schema or stored data.

## Testing
- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Validation run after restacking on merged #733:
- `bun test platform/daemon/src/mutation-api.test.ts` — 29 pass
- `cd libs/sdk && bun run build` — pass
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` — pass with existing `noExplicitAny` warnings in `memory-routes.ts`

## AI disclosure

## Notes
#733 has merged. This branch is now restacked on current `origin/main` and includes the read-side provenance exposure plus direct-read scope hardening from review feedback.

## Latest review follow-up
Addressed the latest review feedback on direct memory reads.

Current head: `b2598fb21`

Changes:
- `GET /api/memory/:id` now resolves the caller agent from `agentId`/`agent_id`, `x-signet-agent-id`, or `x-signet-session-key`.
- Direct reads now apply the same `buildAgentScopeClause` filtering used by recall/search, plus auth project scope when present.
- Cross-agent/private reads now return `404` before any provenance fields are returned.
- Added a regression covering a private memory with provenance: another agent gets `404` and no `source_path`, while the owning agent can read the provenance.

Validation after rebasing onto `origin/main` `f8f81156c`:
- `bun test platform/daemon/src/mutation-api.test.ts` — 29 pass
- `cd libs/sdk && bun run build` — pass
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` — pass with existing `noExplicitAny` warnings in `memory-routes.ts`

## Latest core API review follow-up
Addressed the latest core API review feedback.

Current head: `f8d876f68`

Changes:
- `Database.addMemory(...)` now persists `sourcePath`, `runtimePath`, and `idempotencyKey` instead of accepting and dropping them.
- `Database.updateMemory(...)` now maps those provenance fields as well, keeping the `Partial<Memory>` write surface consistent.
- Added `platform/core/src/database.test.ts` to cover provenance round-trip via `addMemory` and `getMemoryById`.

Validation:
- `bun test platform/core/src/database.test.ts` — 1 pass
- `bun test platform/daemon/src/mutation-api.test.ts` — 29 pass
- `cd libs/sdk && bun run build` — pass
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/database.test.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` — pass with existing `noExplicitAny` warnings in `memory-routes.ts`

### DIFF
--- a/platform/core/src/types.ts
+++ b/platform/core/src/types.ts
@@ -554,6 +554,9 @@ export interface Memory {
 	confidence: number;
 	sourceId?: string;
 	sourceType?: string;
+	sourcePath?: string;
+	runtimePath?: string;
+	idempotencyKey?: string;
 	tags: string[];
 	createdAt: string;
 	updatedAt: string;
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -1413,7 +1413,8 @@ export function registerMemoryRoutes(app: Hono): void {
 			return db
 				.prepare(
 					`SELECT id, content, type, importance, tags, pinned, who,
-					        source_id, source_type, project, ${sessionSelect} confidence,
+					        source_id, source_type, source_path, runtime_path,
+					        idempotency_key, project, ${sessionSelect} confidence,
 					        access_count, last_accessed, is_deleted, deleted_at,
 					        extraction_status, embedding_model, version,
 					        created_at, updated_at, updated_by
@@ -1437,6 +1438,9 @@ export function registerMemoryRoutes(app: Hono): void {
 
 		return c.json({
 			...row,
+			sourcePath: row.source_path,
+			runtimePath: row.runtime_path,
+			idempotencyKey: row.idempotency_key,
 			sessionId,
 		});
 	});
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -11,7 +11,7 @@ import { buildEmbeddingHealth } from "../embedding-health";
 import { linkMemoryToEntities } from "../inline-entity-linker";
 import { logger } from "../logger";
 import { loadMemoryConfig } from "../memory-config";
-import { type RecallParams, hybridRecall } from "../memory-search";
+import { type RecallParams, buildAgentScopeClause, hybridRecall } from "../memory-search";
 import { recordMemorySearchTelemetry } from "../memory-search-telemetry";
 import { resolveMemorySearchTelemetryProject } from "../memory-search-telemetry-project";
 import { buildMemoryTimeline } from "../memory-timeline";
@@ -1407,20 +1407,33 @@ export function registerMemoryRoutes(app: Hono): void {
 			return c.json({ error: "memory id is required" }, 400);
 		}
 
+		const sessionKeyRaw = c.req.header("x-signet-session-key");
+		const agentId = resolveAgentId({
+			agentId: c.req.query("agentId") ?? c.req.query("agent_id") ?? c.req.header("x-signet-agent-id"),
+			sessionKey: sessionKeyRaw,
+		});
+		const agentScope = getAgentScope(agentId);
+		const access = buildAgentScopeClause(agentId, agentScope.readPolicy, agentScope.policyGroup);
+		const scopeProject = c.get("auth")?.claims?.scope?.project;
+		const projectSql = scopeProject ? " AND m.project = ?" : "";
 		const row = getDbAccessor().withReadDb((db) => {
-			const sessionSelect = hasMemoriesSessionIdColumn(db) ? "session_id," : "NULL AS session_id,";
+			const sessionSelect = hasMemoriesSessionIdColumn(db) ? "m.session_id," : "NULL AS session_id,";
 
 			return db
 				.prepare(
-					`SELECT id, content, type, importance, tags, pinned, who,
-					        source_id, source_type, source_path, runtime_path,
-					        idempotency_key, project, ${sessionSelect} confidence,
-					        access_count, last_accessed, is_deleted, deleted_at,
-					        extraction_status, embedding_model, version,
-					        created_at, updated_at, updated_by
-					 FROM memories WHERE id = ? AND (is_deleted = 0 OR is_deleted IS NULL)`,
+					`SELECT m.id, m.content, m.type, m.importance, m.tags, m.pinned, m.who,
+					        m.source_id, m.source_type, m.source_path, m.runtime_path,
+					        m.idempotency_key, m.project, ${sessionSelect} m.confidence,
+					        m.access_count, m.last_accessed, m.is_deleted, m.deleted_at,
+					        m.extraction_status, m.embedding_model, m.version,
+					        m.created_at, m.updated_at, m.updated_by
+					 FROM memories m
+					 WHERE m.id = ?
+					   AND (m.is_deleted = 0 OR m.is_deleted IS NULL)
+					   ${access.sql}
+					   ${projectSql}`,
 				)
-				.get(memoryId) as Record<string, unknown> | undefined;
+				.get(memoryId, ...access.args, ...(scopeProject ? [scopeProject] : [])) as Record<string, unknown> | undefined;
 		});
 
 		if (!row) {
--- a/docs/API.md
+++ b/docs/API.md
@@ -882,7 +882,14 @@ Requires `recall` permission.
   "tags": "preference,editor",
   "pinned": 0,
   "who": "claude-code",
+  "source_id": "optional-external-id",
   "source_type": "manual",
+  "source_path": "/absolute/or/original/source.md",
+  "runtime_path": "memory/MEMORY.md",
+  "idempotency_key": "stable-import-key",
+  "sourcePath": "/absolute/or/original/source.md",
+  "runtimePath": "memory/MEMORY.md",
+  "idempotencyKey": "stable-import-key",
   "project": null,
   "session_id": null,
   "confidence": null,
@@ -898,6 +905,10 @@ Requires `recall` permission.
   "updated_by": "operator"
 }
 ```
+
+`sourcePath`, `runtimePath`, and `idempotencyKey` are camelCase aliases for
+`source_path`, `runtime_path`, and `idempotency_key` so import provenance written
+through `POST /api/memory/remember` is visible on direct reads.
 
 ### GET /api/memory/:id/history
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -868,6 +868,11 @@ compatibility. Requires `remember` permission.
 
 Get a single memory by ID. Returns deleted memories only if the query
 explicitly requests them; by default, soft-deleted records return `404`.
+Direct reads are filtered through the same resolved agent read policy used by
+recall/search. Pass `agentId`/`agent_id`, `x-signet-agent-id`, or an
+`x-signet-session-key` that resolves to an agent when reading non-default
+agent memories; cross-agent or private memories outside that read scope return
+`404` without provenance fields.
 
 Requires `recall` permission.
 
--- a/libs/sdk/src/types.ts
+++ b/libs/sdk/src/types.ts
@@ -9,7 +9,14 @@ export interface MemoryRecord {
 	readonly tags: string | null;
 	readonly pinned: boolean;
 	readonly who: string | null;
+	readonly source_id: string | null;
 	readonly source_type: string | null;
+	readonly source_path: string | null;
+	readonly runtime_path: string | null;
+	readonly idempotency_key: string | null;
+	readonly sourcePath: string | null;
+	readonly runtimePath: string | null;
+	readonly idempotencyKey: string | null;
 	readonly project: string | null;
 	readonly session_id: string | null;
 	readonly confidence: number;
--- a/platform/core/src/database.test.ts
+++ b/platform/core/src/database.test.ts
@@ -1,0 +1,47 @@
+import { afterEach, describe, expect, it } from "bun:test";
+import { mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { Database } from "./database";
+
+describe("Database memory CRUD", () => {
+	let dir: string | null = null;
+	let db: Database | null = null;
+
+	afterEach(() => {
+		db?.close();
+		db = null;
+		if (dir) rmSync(dir, { force: true, recursive: true });
+		dir = null;
+	});
+
+	it("addMemory persists row provenance fields accepted by the Memory input shape", async () => {
+		dir = mkdtempSync(join(tmpdir(), "signet-core-db-"));
+		db = new Database(join(dir, "memories.db"));
+		await db.init();
+
+		const id = db.addMemory({
+			type: "fact",
+			content: "Database addMemory keeps row provenance.",
+			confidence: 0.94,
+			sourceId: "core-db-provenance-source",
+			sourceType: "manual",
+			sourcePath: "/tmp/signet-core/source.md",
+			runtimePath: "memory/source.md",
+			idempotencyKey: "core-db-provenance-key",
+			tags: ["core", "provenance"],
+			updatedBy: "database.test",
+			vectorClock: {},
+			manualOverride: false,
+		});
+
+		expect(db.getMemoryById(id)).toMatchObject({
+			id,
+			sourceId: "core-db-provenance-source",
+			sourceType: "manual",
+			sourcePath: "/tmp/signet-core/source.md",
+			runtimePath: "memory/source.md",
+			idempotencyKey: "core-db-provenance-key",
+		});
+	});
+});
--- a/platform/core/src/database.ts
+++ b/platform/core/src/database.ts
@@ -257,9 +257,10 @@ export class Database {
 			.prepare(
 				`INSERT INTO memories
 				 (id, type, category, content, confidence, source_id,
-				  source_type, tags, created_at, updated_at, updated_by,
-				  vector_clock, manual_override)
-				 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+				  source_type, source_path, runtime_path, idempotency_key,
+				  tags, created_at, updated_at, updated_by, vector_clock,
+				  manual_override)
+				 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
 			)
 			.run(
 				id,
@@ -269,6 +270,9 @@ export class Database {
 				memory.confidence,
 				memory.sourceId ?? null,
 				memory.sourceType ?? null,
+				memory.sourcePath ?? null,
+				memory.runtimePath ?? null,
+				memory.idempotencyKey ?? null,
 				JSON.stringify(memory.tags),
 				now,
 				now,
@@ -312,6 +316,11 @@ export class Database {
 			extractionStatus: "extraction_status",
 			embeddingModel: "embedding_model",
 			extractionModel: "extraction_model",
+			sourceId: "source_id",
+			sourceType: "source_type",
+			sourcePath: "source_path",
+			runtimePath: "runtime_path",
+			idempotencyKey: "idempotency_key",
 			who: "who",
 		};
 
--- a/platform/core/src/database.ts
+++ b/platform/core/src/database.ts
@@ -3,13 +3,13 @@
  * Runtime-detecting: uses bun:sqlite under Bun, better-sqlite3 under Node.js
  */
 
-import { existsSync, readdirSync } from "fs";
-import { join, dirname } from "path";
-import { fileURLToPath } from "url";
-import { homedir } from "os";
-import { arch, platform } from "process";
+import { existsSync, readdirSync } from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+import { arch, platform } from "node:process";
+import { fileURLToPath } from "node:url";
 import { runMigrations } from "./migrations/index";
-import type { Memory, Conversation, Embedding } from "./types";
+import type { Conversation, Embedding, Memory } from "./types";
 import type { MemoryHistory, MemoryJob } from "./types";
 
 // Compute __dirname at runtime so bun's bundler doesn't bake in a static path
@@ -42,7 +42,7 @@ function findSqliteVecExtension(): string | null {
 
 	// Try `npm root -g` to find the actual global prefix (works regardless of runtime)
 	try {
-		const { execFileSync } = require("child_process");
+		const { execFileSync } = require("node:child_process");
 		const npmRoot = (execFileSync("npm", ["root", "-g"], { encoding: "utf8", timeout: 3000 }) as string).trim();
 		if (npmRoot) {
 			const direct = join(npmRoot, platformPkg, extFile);
@@ -187,7 +187,7 @@ export class Database {
 	private dbPath: string;
 	private db: SQLiteDatabase | null = null;
 	private options?: { readonly?: boolean };
-	private vecEnabled: boolean = false;
+	private vecEnabled = false;
 
 	constructor(dbPath: string, options?: { readonly?: boolean }) {
 		this.dbPath = dbPath;
@@ -203,16 +203,19 @@ export class Database {
 			const bunOpts = this.options?.readonly ? { readonly: true } : { readwrite: true, create: true };
 			this.db = new BunDatabase(this.dbPath, bunOpts) as unknown as SQLiteDatabase;
 		} else {
-			let BetterSqlite3;
+			let BetterSqlite3: new (path: string, options?: { readonly?: boolean }) => SQLiteDatabase;
 			try {
-				BetterSqlite3 = (await import("better-sqlite3")).default;
+				BetterSqlite3 = (await import("better-sqlite3")).default as new (
+					path: string,
+					options?: { readonly?: boolean },
+				) => SQLiteDatabase;
 			} catch {
 				throw new Error(
-					"Signet requires Bun (recommended) or the better-sqlite3 npm package. " +
-						(platform === "win32"
+					`Signet requires Bun (recommended) or the better-sqlite3 npm package. ${
+						platform === "win32"
 							? 'Install Bun: powershell -c "irm bun.sh/install.ps1 | iex"\n'
-							: "Install Bun: curl -fsSL https://bun.sh/install | bash\n") +
-						"Or install better-sqlite3: npm install -g better-sqlite3",
+							: "Install Bun: curl -fsSL https://bun.sh/install | bash\n"
+					}Or install better-sqlite3: npm install -g better-sqlite3`,
 				);
 			}
 			this.db = new BetterSqlite3(this.dbPath, {
@@ -568,6 +571,9 @@ function rowToMemory(row: Record<string, unknown>): Memory {
 		confidence: row.confidence as number,
 		sourceId: row.source_id as string | undefined,
 		sourceType: row.source_type as string | undefined,
+		sourcePath: row.source_path as string | undefined,
+		runtimePath: row.runtime_path as string | undefined,
+		idempotencyKey: row.idempotency_key as string | undefined,
 		tags: safeJsonParse(row.tags, []) as string[],
 		createdAt: row.created_at as string,
 		updatedAt: row.updated_at as string,
--- a/platform/daemon/src/mutation-api.test.ts
+++ b/platform/daemon/src/mutation-api.test.ts
@@ -617,6 +617,47 @@ memory:
 		});
 	});
 
+	it("GET /api/memory/:id applies agent scope before returning provenance", async () => {
+		const res = await app.request("http://localhost/api/memory/remember", {
+			method: "POST",
+			headers: { "Content-Type": "application/json" },
+			body: JSON.stringify({
+				content: "Private scoped provenance memory",
+				agentId: "direct-read-agent-a",
+				visibility: "private",
+				who: "soulvessel.tests",
+				sourceType: "hermes-memory",
+				sourceId: "hermes-doc-private-provenance-test",
+				sourcePath: "/tmp/signet-provenance/private.md",
+				runtimePath: "memories/private.md",
+				idempotencyKey: "hermes:private-provenance-test",
+			}),
+		});
+		const remembered = (await res.json()) as { id?: string };
+
+		expect(res.status).toBe(200);
+		expect(remembered.id).toBeString();
+
+		const crossAgent = await app.request(`http://localhost/api/memory/${remembered.id}?agentId=direct-read-agent-b`);
+		const crossAgentJson = (await crossAgent.json()) as { error?: string; source_path?: string | null };
+		expect(crossAgent.status).toBe(404);
+		expect(crossAgentJson.error).toBe("not found");
+		expect(crossAgentJson.source_path).toBeUndefined();
+
+		const sameAgent = await app.request(`http://localhost/api/memory/${remembered.id}?agentId=direct-read-agent-a`);
+		const sameAgentJson = (await sameAgent.json()) as {
+			id?: string;
+			source_path?: string | null;
+			idempotency_key?: string | null;
+		};
+		expect(sameAgent.status).toBe(200);
+		expect(sameAgentJson).toMatchObject({
+			id: remembered.id,
+			source_path: "/tmp/signet-provenance/private.md",
+			idempotency_key: "hermes:private-provenance-test",
+		});
+	});
+
 	it("POST /api/memory/remember persists structured graph data under the requested agent", async () => {
 		const res = await app.request("http://localhost/api/memory/remember", {
 			method: "POST",
--- a/platform/daemon/src/mutation-api.test.ts
+++ b/platform/daemon/src/mutation-api.test.ts
@@ -571,6 +571,52 @@ memory:
 		expect(partial.group).toBeNull();
 	});
 
+	it("GET /api/memory/:id returns row-level provenance for remembered rows", async () => {
+		const res = await app.request("http://localhost/api/memory/remember", {
+			method: "POST",
+			headers: { "Content-Type": "application/json" },
+			body: JSON.stringify({
+				content: "HTTP-visible provenance memory",
+				who: "soulvessel.tests",
+				sourceType: "hermes-memory",
+				sourceId: "hermes-doc-http-provenance-test",
+				sourcePath: "/tmp/signet-provenance/HTTP.md",
+				runtimePath: "memories/HTTP.md",
+				idempotencyKey: "hermes:http-provenance-test",
+			}),
+		});
+		const remembered = (await res.json()) as { id?: string };
+
+		expect(res.status).toBe(200);
+		expect(remembered.id).toBeString();
+
+		const read = await app.request(`http://localhost/api/memory/${remembered.id}`);
+		const json = (await read.json()) as {
+			id?: string;
+			source_id?: string | null;
+			source_type?: string | null;
+			source_path?: string | null;
+			runtime_path?: string | null;
+			idempotency_key?: string | null;
+			sourcePath?: string | null;
+			runtimePath?: string | null;
+			idempotencyKey?: string | null;
+		};
+
+		expect(read.status).toBe(200);
+		expect(json).toMatchObject({
+			id: remembered.id,
+			source_type: "hermes-memory",
+			source_id: "hermes-doc-http-provenance-test",
+			source_path: "/tmp/signet-provenance/HTTP.md",
+			runtime_path: "memories/HTTP.md",
+			idempotency_key: "hermes:http-provenance-test",
+			sourcePath: "/tmp/signet-provenance/HTTP.md",
+			runtimePath: "memories/HTTP.md",
+			idempotencyKey: "hermes:http-provenance-test",
+		});
+	});
+
 	it("POST /api/memory/remember persists structured graph data under the requested agent", async () => {
 		const res = await app.request("http://localhost/api/memory/remember", {
 			method: "POST",


PR-Reviewer-Ant:
Hi @dragontvstaff - I'm taking a look at the fixes in `fix(memory): expose row provenance on read` (commit `e51f0634`) and will follow up shortly.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -762,6 +815,16 @@ export function registerMemoryRoutes(app: Hono): void {
 					if (bySource) return { deduped: true as const, row: bySource };
 				}
```
PR-Reviewer-Ant, line 1088: This idempotency-key dedupe lookup is global across the whole memories table. Unlike the content-hash/source-id checks around it, it does not include `scope`, and it also ignores `agent_id`/`visibility`. That means a retry/import from one agent can dedupe against another agent's private or scoped memory if they reuse the same stable key, and the API then returns that existing row's id and content as `deduped: true`. Please scope this lookup to the same visibility model as the write/read path, at minimum matching the current `agentId`/`visibility`/`scope` semantics, and add a regression test where two agents use the same idempotency key but must not collide.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,17 +607,26 @@ export function registerMemoryRoutes(app: Hono): void {
 				return c.json({ error: "Failed to create chunk group" }, 500);
 			}
 
-			for (const chunk of chunks) {
+			for (let chunkIndex = 0; chunkIndex < chunks.length; chunkIndex += 1) {
+				const chunk = chunks[chunkIndex];
 				const chunkNormalized = normalizeAndHashContent(chunk);
 				if (!chunkNormalized.storageContent) continue;
```
PR-Reviewer-Ant: Same scoping problem in the chunked-memory path: chunk idempotency-key dedupe checks only `idempotency_key` and `is_deleted`. For oversized imports, an agent can silently skip inserting chunks because another agent or scope already has the same derived `baseKey:chunk:N`. That corrupts the import result by returning fewer chunks without telling the caller why. This should use the same scoped predicates as the non-chunk insert path after that path is fixed.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e51f0634`

</details>

The read-side provenance exposure mostly matches the PR description, but the new idempotency-key dedupe path is not scoped. That can cause one agent or visibility domain to dedupe against another agent's memory and return the wrong existing row.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed remember path adds idempotency-key lookup queries in platform/daemon/src/routes/memory-routes.ts without agent_id, visibility, or scope predicates, while the repository conventions explicitly require threading agent_id through user-data reads/writes. The returned dedupe response includes the matched row id and content, so the impact is directly reachable from this diff.

NicholaiVogel:
Rebased this read-side follow-up on top of the fixed #733 stack.

The branch now includes the scoped idempotency-key fix from #733 plus the read response exposure from this PR. The read-side test still verifies `source_path`, `runtime_path`, and `idempotency_key` plus camelCase aliases on `GET /api/memory/:id`.

Validation:
- `bun test platform/daemon/src/mutation-api.test.ts` passed, 22/22.
- `bun test platform/core/src/migrations/migrations.test.ts` passed, 38/38.
- `cd libs/sdk && bun run build` passed.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -80,6 +80,104 @@ function parseOptionalIsoTimestamp(value: unknown): string | null {
...
+	readonly id: string;
+	readonly type: string;
+	readonly tags: string | null;
+	readonly pinned: number;
+	readonly importance: number;
+	readonly content: string;
+}
+
+function pickOptionalString(...values: readonly unknown[]): string | undefined {
+	for (const value of values) {
```
PR-Reviewer-Ant, line 116: This lookup is not equivalent to the new unique index. Migration 072 indexes `COALESCE(NULLIF(agent_id, ''), 'default')`, `COALESCE(visibility, 'global')`, and `COALESCE(scope, '__NULL__')`, but the remember-path dedupe query requires exact `agent_id = ?`, `visibility = ?`, and `scope IS NULL`/`scope = ?`. If an existing row has `NULL` visibility or an empty/null agent id, the pre-insert idempotency lookup misses it even though SQLite will still treat it as the same unique tuple. The insert then hits the unique constraint, and the catch path repeats the same exact idempotency lookup before falling back to content-hash lookup, so a retry can return a 500 or dedupe to the wrong content-hash row. The read/write path should use the same normalization expression as the index, or the migration should backfill canonical values before creating the index.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `00a8fb09`

</details>

The read-side provenance shape is mostly wired through, but I found a scoped idempotency mismatch that can break retries against existing normalized rows.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The new unique index normalizes agent_id, visibility, and scope values, while the new daemon lookup path uses exact predicates. I did not run the daemon tests in this read-only review environment, but the mismatch is visible directly in the changed SQL.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -621,38 +778,38 @@ export function registerMemoryRoutes(app: Hono): void {
 							 VALUES (?, ?, 'chunk', 1.0, ?)`,
```
PR-Reviewer-Ant, line 917: For partial idempotent retries of oversized content, this can return an incomplete and reordered `ids` list. Existing chunks are pushed into `chunkIds`, but when a new chunk is inserted the code pushes `result.id`, while `result.id` is `chunkId`, not the original chunk index. Since skipped existing chunks are not tracked by index in the final response, a retry where only some chunk rows already exist can report fewer chunks than `chunkPlans.length` and lose the original chunk ordering. That is data-corrupting for callers that use the returned `ids`/`group_id` as the durable chunk mapping. Build the response from a per-index array, e.g. assign `ids[chunkIndex] = result.id`, and return only after all indexes are populated.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -762,6 +912,9 @@ export function registerMemoryRoutes(app: Hono): void {
 					if (bySource) return { deduped: true as const, row: bySource };
```
PR-Reviewer-Ant, line 1086: The PR description says there are no `/api/memory/remember` write-path behavior changes, but this branch adds idempotency-key parsing, dedupe lookup, insert columns, chunk retry semantics, and a new scoped idempotency migration. That may be intentional because the branch is stacked on #733, but the description is materially out of sync with the implementation reviewers are being asked to evaluate.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1381942f`

</details>

I found one correctness issue in the new chunked idempotency path: partial retries can silently report the wrong chunk order and count. The read-side provenance exposure itself is covered, but the PR also changes write-path behavior despite the stated boundary.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunked remember flow is fully visible in platform/daemon/src/routes/memory-routes.ts, including the early all-chunks dedupe return, the partial-existing branch, and the response construction from chunkIds. The bug follows directly from the changed control flow.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

dragontvstaff:
Seer review gate verdict: RED / changes required before Operator merge authorization.

Blocking findings:

1. Focused daemon regression fails on current PR head `00a8fb09`.

Command run locally on the PR branch:

```bash
cd platform/daemon && bun test src/mutation-api.test.ts
```

Result: 21 pass, 1 fail. The failing test is:

```text
POST /api/memory/remember scopes idempotency-key dedupe by agent and visibility
```

Failure mode: the second `POST /api/memory/remember` using the same `idempotencyKey` for `agentId: "agent-a"` returns HTTP 500 instead of 200. The logged error is:

```text
SQLiteError: UNIQUE constraint failed: memories.idempotency_key
Expected: 200
Received: 500
```

That violates the stated test plan / gate criterion that focused tests pass, and it is a write-path/idempotency regression adjacent to the provenance work.

2. The PR stack is stale relative to PR #733.

Observed locally after fetching both heads:

```text
PR733 head: 54a1aeb7
PR734 head: 00a8fb09
PR733 head is NOT ancestor of PR734
latest PR733 commits not in PR734:
54a1aeb7 fix(memory): return chunk ids on idempotent retries
```

So #734 is not currently stacked on the latest write-side provenance branch. If #733 lands at `54a1aeb7`, #734 needs to be rebased/restacked before review can green.

3. Diff against upstream `main` is not bounded to read-surface exposure only.

`origin/main...HEAD` includes write-path/storage/idempotency/schema changes in addition to the read endpoint work:

```text
platform/core/src/migrations/072-agent-scoped-idempotency-key.ts
platform/daemon/src/routes/memory-routes.ts
platform/daemon/src/transactions.ts
platform/daemon/src/mutation-api.test.ts
```

Some of that belongs to the stacked #733 work, but the merge gate for #734 cannot treat the PR as read-only while #733 remains open and this branch carries non-read changes.

Non-blocking positive evidence:

- The direct read-surface change does select `source_path`, `runtime_path`, and `idempotency_key` in `GET /api/memory/:id` and returns camelCase aliases.
- `cd platform/core && bun test src/recall.test.ts` passed, 4/4.
- `cd platform/core && bun test src/migrations/migrations.test.ts` passed, 38/38.
- `cd platform/core && bun run build` passed.
- `cd libs/sdk && bun run build` passed.
- `git diff --check origin/main...HEAD` passed.
- I did not run a live daemon POST→GET provenance smoke after the focused daemon suite failed; the branch is already red.

Required next step:

- Rebase/restack #734 on the latest #733 head after #733 is fixed/accepted, or split/rebase it so the diff against the actual merge base is read-surface only.
- Fix the `memories.idempotency_key` uniqueness failure so `platform/daemon/src/mutation-api.test.ts` passes on the PR head.
- Then rerun the focused tests and the disposable live daemon POST `/api/memory/remember` → GET `/api/memory/:id` provenance readback proof.

Formal GitHub review state note: this is posted as a top-level Seer review comment because the authenticated GitHub identity is the PR author (`dragontvstaff`), so GitHub formal approval/request-change review semantics are not a clean independent signal here.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e28a61da`

</details>

I did not find a concrete correctness, security, or data-integrity issue in the changed lines. The implementation matches the stated read-side provenance goal and adds relevant coverage for the new dedupe/provenance paths.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the route changes, migration, SDK/core type updates, docs, and daemon/core tests needed to evaluate the claimed behavior. I did not run the test suite here, so confidence is limited to static review of the provided diff and changed file contents.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `47a4c52b`

</details>

I found a cross-agent correctness leak in the new chunked idempotency path. The read-side provenance exposure itself is covered, but oversized remembers can return another agent's existing chunk rows instead of creating scoped rows.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunked remember path builds an agent/visibility/scope dedupe key for idempotency, but then falls back to a content_hash lookup that only checks scope. That is visible directly in memory-routes.ts and conflicts with the PR's scoped idempotency behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,39 +766,37 @@ export function registerMemoryRoutes(app: Hono): void {
...
-				const memType = inferType(chunk);
 
 				try {
-					const inserted = getDbAccessor().withWriteTx((db) => {
+					const result = getDbAccessor().withWriteTx((db) => {
+						const byIdempotencyKey = getScopedIdempotencyMemoryId(db, plan.idempotencyKey, dedupeScope);
+						if (byIdempotencyKey) return { id: byIdempotencyKey.id, inserted: false as const };
+
 						const byHash =
 							scope !== null
```
PR-Reviewer-Ant: This fallback can return chunks owned by a different agent. The idempotency lookup above is scoped by agent, visibility, and scope, but the content-hash fallback here only checks `content_hash` plus `scope`. For an oversized remember from `agent-a` with the same chunk content as an existing default-agent import, this branch returns the existing default-agent chunk id, skips insertion, and the response can hand back another agent's memory rows instead of creating rows under `agent-a`. That breaks the scoped idempotency contract this PR is adding and can also leak cross-agent memory ids/content provenance through the chunked response. The hash lookup needs the same normalized agent/visibility/scope predicate as the idempotency lookup, or it needs to be removed from this scoped chunk path.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e891a0a2`

</details>

I found one idempotency hole in the chunked remember path: the new chunk-specific keys do not guard the caller-supplied base key, so changing payload size can create unrelated rows under the same import key.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The chunked path only checks existing derived keys via getScopedChunkIdempotencyRows(rowProvenance.idempotencyKey, dedupeScope), while the non-chunk path stores and checks the raw idempotencyKey. That mismatch is visible in platform/daemon/src/routes/memory-routes.ts and is not covered by the added tests, which only retry oversized content against oversized content.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -548,15 +750,58 @@ export function registerMemoryRoutes(app: Hono): void {
...
+						chunk,
+						contentForInsert:
+							normalized.normalizedContent.length > 0 ? normalized.normalizedContent : normalized.hashBasis,
+						idempotencyKey: idempotencyKeyForChunk(rowProvenance.idempotencyKey, index),
+						memType: inferType(chunk),
+						normalized,
+					};
+				})
+				.filter((plan): plan is NonNullable<typeof plan> => plan !== null);
+			if (chunkPlans.length === 0) {
```
PR-Reviewer-Ant, line 788: The chunked path only looks for existing `${idempotencyKey}:chunk:N` rows, but it never checks whether the raw `idempotencyKey` is already attached to a non-chunk memory in the same scope. That means a client can first remember a small payload with `idempotencyKey: "k"`, then retry/import an oversized payload with the same key and get a new chunk group instead of a dedupe/conflict. The inverse is also possible: after a chunked import using `k:chunk:N`, a later small payload with `k` can insert a separate row. Since the API docs now promise `idempotencyKey` dedupes retries within the agent/visibility/scope tuple, the chunk path should reserve/check the base key as well, or explicitly reject mixed chunked/non-chunk reuse.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,39 +818,28 @@ export function registerMemoryRoutes(app: Hono): void {
...
-						if (byHash) return false;
+					const result = getDbAccessor().withWriteTx((db) => {
+						const byIdempotencyKey = getScopedIdempotencyMemoryId(db, plan.idempotencyKey, dedupeScope);
+						if (byIdempotencyKey) return { id: byIdempotencyKey.id, inserted: false as const };
+
+						const byHash = getScopedContentHashMemoryId(db, plan.normalized.contentHash, dedupeScope);
+						if (byHash) return { id: byHash.id, inserted: false as const };
 
 						txIngestEnvelope(db, {
 							id: chunkId,
```
PR-Reviewer-Ant, line 886: This fallback can put unrelated existing memories into a new chunked response. In the chunk path, `byHash` is scoped only by agent/scope and does not require the existing row to be a chunk for this `groupId`; when it hits, the code returns that existing id, skips the insert/linking work, and later returns the newly-created `group_id`. That means a chunked save can report `ids` that are not linked to the returned group, especially when two chunks normalize to the same content or when the user saves oversized content containing a sentence already stored as a normal memory. For chunked imports, content-hash dedupe needs to either reuse a complete existing chunk group or insert/link rows for the new group; mixing arbitrary existing rows into a fresh group corrupts the chunk-group contract.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5f969a13`

</details>

I found one correctness issue in the chunked remember path: deduping individual chunks by content hash can now return existing memories as members of a new chunk group without linking them to that group. The read-side provenance exposure itself is covered by the changed select/response shape.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunk loop explicitly returns an existing memory id from getScopedContentHashMemoryId and then skips the insertion/linking block. The same branch still returns the newly-created group_id, so the response can describe a chunk group that does not own all returned ids.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,39 +834,28 @@ export function registerMemoryRoutes(app: Hono): void {
...
```
PR-Reviewer-Ant, line 886: This reuses any existing memory with the same content hash as a chunk, but the non-inserted path skips the `memory_entity_mentions` insert below and does not verify that the existing row belongs to this `groupId`. That means a new oversized remember call can return `group_id` for a freshly created chunk group while its `ids` point at unrelated/pre-existing memories that are not linked to the group and may not have `source_type = 'chunk'` or this provenance. For chunked imports, content-hash dedupe should either only reuse rows already linked to the same chunk group, or attach/update the group linkage/provenance before returning them.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1810bed8`

</details>

I found one chunked-memory correctness issue in the stacked write-side changes: a new chunk group can report pre-existing memories as its chunks without actually linking them to that group.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed chunk insertion path explicitly reuses any same-agent/scope content-hash match, then skips the chunk-group mention insert and returns that existing id in the new group's ids. This is visible in platform/daemon/src/routes/memory-routes.ts and directly affects the PR's provenance/read behavior for oversized remember calls.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `57d17adc`

</details>

I did not find a correctness, security, or data-integrity issue worth flagging in this diff. The implementation matches the PR description: direct memory reads now select and return row provenance, and the stacked write-side/idempotency changes have targeted regression coverage.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed GET /api/memory/:id query explicitly selects source_path, runtime_path, and idempotency_key and returns the camelCase aliases, and mutation-api.test covers the remember-to-read provenance path. I did not run the test suite or exercise a live daemon, so confidence is bounded to the supplied diff and included file contents.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in the #733 base branch.

The idempotency lookup and unique-constraint fallback now use normalized agent/visibility/scope predicates matching migration 072. The regression covers same-key reuse across agents/visibility and legacy normalized rows.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in the #733 base branch.

Chunk idempotency now uses scoped lookup, returns full existing chunk ids for exact retries, rejects changed/partial chunk plans with `409`, and preserves chunk id ordering in the response.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in the #733 base branch.

The idempotency lookup and unique-constraint fallback now use normalized agent/visibility/scope predicates matching migration 072. The regression covers same-key reuse across agents/visibility and legacy normalized rows.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `54a1aeb76`, `35a492af0`, `87f540178`, and later hardening commits.

The route checks existing chunks before creating a group, exact retries return the original ids/group, changed content returns `409`, and chunk ids are tracked by index rather than append order.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed. I updated the #734 body to call out that the branch is stacked on #733 and that the write-path changes belong to the base PR. The top commit remains the read-side response/types/docs change.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `d4389ea9d`.

The remember content-hash fallback now uses the same normalized agent/scope predicate as `idx_memories_content_hash_unique`. The cross-agent idempotency regression now uses identical content so it would catch the old fallback leak.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `a3bb9ec89`.

The write path now rejects mixed chunked/non-chunk reuse of the same base idempotency key in either direction, with a regression covering small-then-chunked and chunked-then-small reuse.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `e2dd61e3`.

The chunk path no longer mixes arbitrary content-hash matches into a fresh chunk-group response. It preflights duplicate/pre-existing chunk content and returns `409` before creating a group; exact idempotent retries still reuse the complete existing chunk group.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `e2dd61e3`.

The chunk path no longer mixes arbitrary content-hash matches into a fresh chunk-group response. It preflights duplicate/pre-existing chunk content and returns `409` before creating a group; exact idempotent retries still reuse the complete existing chunk group.

NicholaiVogel:
Rebased on the updated #733 base and replied on each review thread.

Current head: `57d17adc`

Validation:
- `bun test platform/daemon/src/mutation-api.test.ts` — 27 pass on the stacked branch
- `bun test platform/core/src/migrations/migrations.test.ts` — 39 pass
- `cd libs/sdk && bun run build` — pass
- scoped Biome check — pass with the existing `memory-routes.ts` `noExplicitAny` warnings

I also updated the PR body so it accurately describes the stack boundary: #733 owns the write-path/idempotency changes; #734's top commit is the read response/types/docs work.

NicholaiVogel:
Rebased onto the updated #733 stack, which is now based on current `origin/main` (`f88f064c2`), and force-pushed.

Current head: `b2e8e7ca5`

Validation after rebase:
- `bun test platform/daemon/src/mutation-api.test.ts` — 27 pass
- `bun test platform/core/src/migrations/migrations.test.ts` — 39 pass
- `cd libs/sdk && bun run build` — pass
- scoped Biome check — pass, with the existing `memory-routes.ts` `noExplicitAny` warnings

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -566,39 +847,25 @@ export function registerMemoryRoutes(app: Hono): void {
...
-										.prepare("SELECT id FROM memories WHERE content_hash = ? AND scope = ? AND is_deleted = 0 LIMIT 1")
-										.get(chunkNormalized.contentHash, scope) as { id: string } | undefined)
-								: (db
-										.prepare(
-											"SELECT id FROM memories WHERE content_hash = ? AND scope IS NULL AND is_deleted = 0 LIMIT 1",
-										)
-										.get(chunkNormalized.contentHash) as { id: string } | undefined);
-						if (byHash) return false;
+					const result = getDbAccessor().withWriteTx((db) => {
+						const byIdempotencyKey = getScopedIdempotencyMemoryId(db, plan.idempotencyKey, dedupeScope);
```
PR-Reviewer-Ant: This chunk insert path no longer rechecks `content_hash` inside the write transaction. The new preflight `getScopedContentHashMemoryId` check runs before the loop, then each chunk transaction only checks `idempotency_key` before calling `txIngestEnvelope`. Two concurrent oversized remembers with the same chunk content can both pass the preflight; one inserts, and the other hits the unique content-hash constraint here and returns `Failed to save chunk` as a 500. The old code did the hash check inside the transaction. Please move the scoped content-hash check back into the write transaction, or catch the unique constraint here and return the intended dedupe/conflict response.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b2e8e7ca`

</details>

I found one race introduced in the chunked remember path. The read-side provenance exposure matches the PR description, but the refactor moved a content-hash guard out of the write transaction.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff shows the chunk content-hash duplicate check now runs before the per-chunk write transaction, while the transaction only rechecks idempotency before calling txIngestEnvelope. I did not run a concurrent repro, but the race follows directly from the changed transaction boundary.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in the updated base branch via `0cc8229a1` and rebased this PR on top at `f732a37cd`.

The chunk path now rechecks the scoped `content_hash` inside the per-chunk write transaction before `txIngestEnvelope`, and maps any remaining `idx_memories_content_hash_unique` / `memories.content_hash` unique fallback to the same 409 conflict response. I also added a concurrent oversized remember regression: two same-content chunked writes with different idempotency keys now resolve as one 200 and one 409 instead of exposing a 500.

Validation rerun after the rebase:
- `cd libs/sdk && bun run build`
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (same pre-existing `noExplicitAny` warnings in `memory-routes.ts`)

NicholaiVogel:
Updated this stacked branch on top of #733's new `0cc8229a1` race fix and replied/resolved the inline review thread.

Current head: `f732a37cd`

Validation after rebase:
- `cd libs/sdk && bun run build`
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (same pre-existing `noExplicitAny` warnings in `memory-routes.ts`)

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -621,38 +903,41 @@ export function registerMemoryRoutes(app: Hono): void {
...
+						return { id: chunkId, status: "inserted" };
 					});
 
-					if (!inserted) continue;
-					chunkIds.push(chunkId);
+					if (result.status === "content_conflict") {
+						return c.json({ error: "chunk content already exists for this agent and scope" }, 409);
+					}
+					chunkIds[chunkIndex] = result.id;
+					if (result.status !== "inserted") continue;
```
PR-Reviewer-Ant: This conflict path can leave partial chunked-import state committed. The chunk group is inserted in an earlier transaction, and each chunk is inserted in its own transaction before embedding/linking. If another request races and creates a later chunk hash after the preflight checks, this branch returns 409 after any earlier chunks and the chunk_group entity for this request have already been committed. That leaves orphaned or partial chunk groups tied to an idempotency key that the caller sees as failed. The whole chunk import needs to be atomic, or the conflict path needs explicit cleanup/rollback for the group and any chunks already inserted by this request.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f732a37c`

</details>

I found one data-integrity issue in the new chunked idempotency path: conflict handling can leave durable partial state behind. The read-side provenance exposure itself matches the PR description.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The chunk-group insert and each chunk insert are separate write transactions in platform/daemon/src/routes/memory-routes.ts, and the new conflict returns inside the insertion loop after earlier writes may already have committed. The added concurrent-collision test only checks response statuses, not cleanup of partially-created chunk groups or chunks.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

dragontvstaff:
Seer re-review gate: YELLOW / not green yet.

Current verification on PR #734 head `f732a37cd10a634a7199482f9482eb4b6d02a03b`:

- Stack coherence: PASS. Current PR #733 head `0cc8229a17a2d1c2f4e1bfade5c753c92654de0a` is an ancestor of current PR #734 head `f732a37cd10a634a7199482f9482eb4b6d02a03b`.
- Repaired head successor: PASS. Current PR #734 is a coherent newer successor of the repaired stack, though it was force-pushed after the sv-builder handoff.
- `git diff --check origin/main...HEAD`: PASS.
- `cd platform/daemon && bun test src/mutation-api.test.ts`: PASS, 28 pass / 0 fail.
- `cd platform/core && bun test src/recall.test.ts`: PASS, 4 pass / 0 fail.
- `cd platform/core && bun test src/migrations/migrations.test.ts`: PASS, 39 pass / 0 fail.
- `cd platform/core && bun run build`: PASS.
- `cd libs/sdk && bun run build`: PASS.
- Disposable live daemon HTTP smoke: PASS. With `SIGNET_PATH=/tmp/signet-pr734-seer-smoke-6kqfvwdc` and `SIGNET_PORT=42069`, POST `/api/memory/remember` then GET `/api/memory/:id` returned `source_path`, `runtime_path`, `idempotency_key` and camelCase aliases for memory `374425c6-f12c-439c-87f8-5e00721aac0b`.
- Boundary check: PASS. Disposable daemon used `/tmp`; no production Signet DB/profile or Soul Vessel workspace was targeted. The daemon was stopped after the smoke.

Blocking gate finding:

- The PR body is stale after the latest force-push. It still says current #733 head is `e2dd61e351252b09789573f994e0bffb48ed6d97` and this PR head is `57d17adc883c3f9e0ba4df71d275239cbb4c0ffd`, but GitHub currently reports #733 at `0cc8229a17a2d1c2f4e1bfade5c753c92654de0a` and #734 at `f732a37cd10a634a7199482f9482eb4b6d02a03b`. The test plan also says the daemon focused suite has 27 tests; the current focused suite has 28.

Required before green:

- Update the PR body stack note/test plan to the current heads and current test counts, without claiming M10 is fully green until downstream Soul Vessel cleanup lands.

No code-review blocker found in the current read-surface diff. This is held on PR metadata accuracy only.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `62aa171b0` on #733 and carried into this stacked branch.

The chunked remember write now performs the idempotency/content-hash rechecks, chunk-group creation, all chunk row inserts, and group mention inserts in one `withWriteTx`. If a conflict is detected, the transaction returns before any group or chunk rows are written. Embeddings, inline entity linking, and extraction enqueue remain post-commit non-fatal follow-up work.

I also strengthened the concurrent collision regression so the losing idempotency key must leave zero chunk rows and no deterministic chunk-group entity behind.

Validation:
- scoped Biome check still only reports the pre-existing `noExplicitAny` warnings in `memory-routes.ts`

NicholaiVogel:
Restacked on #733's atomic chunk write fix.

Current head: `2b1832454`

Validation after rebase:
- `cd libs/sdk && bun run build`
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (same pre-existing `noExplicitAny` warnings in `memory-routes.ts`)

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -548,57 +771,127 @@ export function registerMemoryRoutes(app: Hono): void {
 			const tags = hasBodyTags ? bodyTags : parsedPrefixes.tags;
```
PR-Reviewer-Ant, line 771: This PR is advertised as read-side provenance exposure, but the current diff also changes `POST /api/memory/remember` chunking/idempotency behavior: it precomputes chunk plans, rejects existing chunk content, uses deterministic chunk-group IDs for idempotency keys, and changes mixed chunk/non-chunk reuse handling. That may all be valid #733 work, but it makes #734 much harder to review against its stated contract and could ship write-path behavior before the stack is reconciled. Given the PR description explicitly says write-path changes belong to #733 and current #733 is no longer an ancestor, I would split or restack this before relying on this review as coverage for the read-side change.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `2b183245`

</details>

I found one non-blocking stack/scope issue: this PR is described as read-side provenance exposure, but the diff also includes substantial write-path idempotency, chunking, and migration behavior that belongs to the stacked #733 work. The direct GET/read response change itself appears to return the claimed snake_case and camelCase provenance fields.

**Confidence:** Medium [sufficient_diff_evidence, ambiguous_requirements] - The changed GET /api/memory/:id query selects source_path, runtime_path, and idempotency_key and returns camelCase aliases, matching the stated read-side goal. However, the same diff also adds migration 072, changes remember dedupe semantics, chunk idempotency handling, and transaction inserts, while the PR description says write-path behavior belongs to #733 and the stack still needs reconciliation.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
#733 is merged, so I restacked this PR onto current `origin/main` and force-pushed only the read-side provenance commit.

Current head: `42cc5b4af`

The PR now has:
- 1 commit
- 6 changed files
- no migration files touched
- PR readiness and migration guards passing

Validation after restack:
- `bun test platform/daemon/src/mutation-api.test.ts` — 28 pass
- `cd libs/sdk && bun run build` — pass
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` — pass with the existing `noExplicitAny` warnings in `memory-routes.ts`

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -1413,7 +1413,8 @@ export function registerMemoryRoutes(app: Hono): void {
 			return db
 				.prepare(
 					`SELECT id, content, type, importance, tags, pinned, who,
-					        source_id, source_type, project, ${sessionSelect} confidence,
+					        source_id, source_type, source_path, runtime_path,
+					        idempotency_key, project, ${sessionSelect} confidence,
```
PR-Reviewer-Ant: This read now exposes row provenance, including source paths and idempotency keys, but the endpoint still fetches by bare `id` with no `agent_id`, visibility, read-policy, or auth project predicate. That means any caller with `recall` permission and a memory id can read provenance for another agent's memory, and the newly exposed `source_path` can include local/import paths. Please scope this direct read the same way recall/list paths are scoped, or otherwise explicitly reject cross-agent/private reads before returning these fields.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `42cc5b4a`

</details>

I found a scope leak in the direct memory read path. The PR does expose the claimed provenance fields, but it does so through an endpoint that still reads by bare memory id and now returns path/idempotency provenance across agent boundaries.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed GET /api/memory/:id SELECT now includes source_path, runtime_path, and idempotency_key, while the same query still filters only by id and deletion state. The repository convention explicitly requires threading agent_id through reads touching user data, and the added regression test only covers the default unscoped read.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the latest review feedback on direct memory reads.

Current head: `b2598fb21`

Changes:
- `GET /api/memory/:id` now resolves the caller agent from `agentId`/`agent_id`, `x-signet-agent-id`, or `x-signet-session-key`.
- Direct reads now apply the same `buildAgentScopeClause` filtering used by recall/search, plus auth project scope when present.
- Cross-agent/private reads now return `404` before any provenance fields are returned.
- Added a regression covering a private memory with provenance: another agent gets `404` and no `source_path`, while the owning agent can read the provenance.

Validation after rebasing onto `origin/main` `f8f81156c`:
- `bun test platform/daemon/src/mutation-api.test.ts` — 29 pass
- `cd libs/sdk && bun run build` — pass
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` — pass with existing `noExplicitAny` warnings in `memory-routes.ts`

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `b2598fb21`.

`GET /api/memory/:id` now resolves the caller agent from query/header/session context, applies the same `buildAgentScopeClause` used by recall/search, and also applies auth project scope when present. Cross-agent/private misses return `404` before provenance is serialized.

Added a regression covering a private memory with source/idempotency provenance: another agent receives `404` with no `source_path`, while the owning agent can read the provenance.

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Resolved by the restack after #733 merged. Current #734 is based on current `origin/main` and contains the read-side provenance commit plus the direct-read scope hardening from the latest review; the #733 write-path changes are now already in main.

platform/core/src/types.ts
```diff
@@ -554,6 +554,9 @@ export interface Memory {
 	confidence: number;
 	sourceId?: string;
 	sourceType?: string;
+	sourcePath?: string;
```
PR-Reviewer-Ant, line 557: Adding these fields to `Memory` also adds them to the accepted input type for `Database.addMemory(...)`, because that method takes `Omit<Memory, "id" | "createdAt" | "updatedAt" | "version">`. But `addMemory` still inserts only `source_id` and `source_type`, so a core caller can now pass `sourcePath`, `runtimePath`, or `idempotencyKey` with type-checker approval and have them silently discarded. Either persist these columns in `addMemory` too, or split the read row shape from the write input so the core API does not imply these fields are supported there.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b2598fb2`

</details>

I found one correctness issue in the core Database API: the new provenance fields are now part of the Memory write shape, but addMemory still drops them on insert.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The type change in platform/core/src/types.ts makes sourcePath/runtimePath/idempotencyKey valid Memory fields, and Database.addMemory accepts Omit<Memory, ...>, but the insert in platform/core/src/database.ts still only writes source_id and source_type. I did not have call-site search output for every Database.addMemory user, so impact depends on whether callers use that core helper for provenance writes.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/core/src/types.ts
NicholaiVogel: Fixed in `f8d876f68`.

`Database.addMemory(...)` now persists `sourcePath`, `runtimePath`, and `idempotencyKey` into `source_path`, `runtime_path`, and `idempotency_key` instead of accepting and dropping them. I also included those provenance fields in `updateMemory(...)` so the broader `Partial<Memory>` write surface stays consistent.

Added `platform/core/src/database.test.ts` covering `addMemory` round-tripping provenance through `getMemoryById`.

Validation:
- `bun test platform/core/src/database.test.ts`
- `cd libs/sdk && bun run build`
- scoped Biome check

NicholaiVogel:
Addressed the latest core API review feedback.

Current head: `f8d876f68`

Changes:
- `Database.addMemory(...)` now persists `sourcePath`, `runtimePath`, and `idempotencyKey` instead of accepting and dropping them.
- `Database.updateMemory(...)` now maps those provenance fields as well, keeping the `Partial<Memory>` write surface consistent.
- Added `platform/core/src/database.test.ts` to cover provenance round-trip via `addMemory` and `getMemoryById`.

Validation:
- `bun test platform/core/src/database.test.ts` — 1 pass
- `bun test platform/daemon/src/mutation-api.test.ts` — 29 pass
- `cd libs/sdk && bun run build` — pass
- `bunx biome check docs/API.md libs/sdk/src/types.ts platform/core/src/database.ts platform/core/src/database.test.ts platform/core/src/types.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` — pass with existing `noExplicitAny` warnings in `memory-routes.ts`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f8d876f6`

</details>

I did not find a concrete correctness, security, or data-integrity issue in this diff. The implementation matches the PR description: provenance is persisted through the core write path, surfaced on direct reads with camelCase aliases, documented, and the direct read path now applies agent/project scoping before returning the row.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The supplied diff includes the changed route, database wrapper, SDK/core types, docs, and regression tests, and the route-level scoping is visible in platform/daemon/src/routes/memory-routes.ts before provenance serialization. I could not run the repo tests or inspect helper implementations from the workspace because the checkout only contained the review prompt, so this is based on the provided PR context rather than a live runtime repro.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
nice job @dragontvstaff - thanks for the contribution :)

PR #739, opened 2026-05-20
fix(memory): harden aggregate recall

NicholaiVogel:
## Summary
- Adds explicit aggregate recall mode across daemon, CLI, MCP, SDK, and integration surfaces.
- Requires `remember` permission before recall routes can persist aggregate memories, while `saveAggregate: false` remains read-only.
- Aligns aggregate content-hash dedupe with the existing unique index and returns an unsaved aggregate on inaccessible duplicate content instead of 500ing.
- Adds provenance links from saved aggregate memories to source memory IDs.

## Test Plan
- `bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/core/src/recall.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/memory.test.ts libs/sdk/src/__tests__/client.test.ts libs/sdk/src/__tests__/openai.test.ts libs/sdk/src/__tests__/ai-sdk.test.ts`
- `bun run typecheck`
- `bunx biome check docs/API.md docs/MCP.md docs/CLI.md platform/daemon/src/aggregate-recall.ts platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/routes/memory-routes.ts platform/daemon/src/routes/hooks-routes.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/daemon/src/mcp/tools.ts platform/daemon/src/mcp/tools.test.ts integrations/hermes-agent/connector/hermes-plugin/client.py`

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Rollback / compatibility: migration 073 only adds aggregate-memory provenance links. Existing memories and normal recall results remain compatible; old clients ignore aggregate metadata, and rollback would only remove aggregate source-link provenance while leaving ordinary saved memories intact.

### DIFF
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -182,6 +182,10 @@ function uniqueEvidence(rows: readonly RecallResult[]): RecallResult[] {
 	return result;
 }
 
+function evidenceCanSaveAsGlobalAggregate(rows: readonly RecallResult[]): boolean {
+	return rows.every((row) => (row.visibility ?? "global") === "global" && (row.scope ?? null) === null);
+}
+
 function aggregateKey(input: {
 	readonly agentId: string;
 	readonly project: string | null;
@@ -506,7 +510,7 @@ export async function aggregateRecall(
 	let row: RecallResult | null;
 	let deduped = false;
 	let saved = false;
-	if (saveAggregate) {
+	if (saveAggregate && evidenceCanSaveAsGlobalAggregate(evidence)) {
 		const normalized = normalizeAndHashContent(answer);
 		row = getDbAccessor().withWriteTx((db) => {
 			const duplicate = resolveAggregateDuplicate(db, {
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -259,6 +259,7 @@ function loadAggregateByKey(
 			 FROM memories
 			 WHERE idempotency_key = ?
 			   AND COALESCE(NULLIF(agent_id, ''), 'default') = ?
+			   AND source_type = 'aggregate-recall'
 			   AND visibility = 'global'
 			   AND scope IS NULL
 			   AND is_deleted = 0
@@ -571,7 +572,11 @@ export async function aggregateRecall(
 					sourceMemoryIds,
 					now,
 				});
-				if (!racedDuplicate) throw err;
+				if (!racedDuplicate) {
+					deduped = true;
+					saved = false;
+					return unsavedAggregateResult(answer, key, project);
+				}
 				deduped = true;
 				saved = racedDuplicate.saved;
 				return racedDuplicate.row;
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -6,7 +6,7 @@ import { syncVecDeleteBySourceId, syncVecInsert, vectorToBlob } from "./db-helpe
 import { logger } from "./logger";
 import type { EmbeddingConfig, ResolvedMemoryConfig } from "./memory-config";
 import { type EmbedFn, type RecallParams, type RecallResponse, type RecallResult, hybridRecall } from "./memory-search";
-import { txIngestEnvelope } from "./transactions";
+import { type IngestEnvelope, txIngestEnvelope } from "./transactions";
 
 export type AggregateRecallBudget = "small" | "medium" | "large";
 export type AggregateRecallStoppedReason = "complete" | "no_evidence" | "router_unavailable" | "synthesis_failed";
@@ -30,6 +30,7 @@ interface AggregateRecallDeps {
 	readonly now?: () => Date;
 	readonly idFactory?: () => string;
 	readonly logger?: AggregateRecallLogger;
+	readonly ingestEnvelope?: (db: WriteDb, mem: IngestEnvelope) => string;
 }
 
 interface AggregateRecallLogger {
@@ -55,6 +56,11 @@ interface ContentHashMatch {
 	readonly row: RecallResult;
 	readonly visibleForAggregate: boolean;
 	readonly aggregateRecallMemory: boolean;
+}
+
+interface AggregateDuplicateResolution {
+	readonly row: RecallResult;
+	readonly saved: boolean;
 }
 
 const BUDGET_QUERY_LIMITS: Record<AggregateRecallBudget, number> = {
@@ -271,6 +277,39 @@ function linkAggregateSources(
 	}
 }
 
+function resolveAggregateDuplicate(
+	db: WriteDb,
+	input: {
+		readonly key: string;
+		readonly agentId: string;
+		readonly project: string | null;
+		readonly contentHash: string;
+		readonly answer: string;
+		readonly sourceMemoryIds: readonly string[];
+		readonly now: string;
+	},
+): AggregateDuplicateResolution | null {
+	const existing = loadAggregateByKey(db, input.key, { agentId: input.agentId, project: input.project });
+	if (existing) {
+		linkAggregateSources(db, existing.id, input.sourceMemoryIds, input.agentId, input.now);
+		return { row: existing, saved: true };
+	}
+	const duplicateContent = loadMemoryByContentHash(db, input.contentHash, {
+		agentId: input.agentId,
+		project: input.project,
+	});
+	if (!duplicateContent) return null;
+	if (!duplicateContent.visibleForAggregate || !duplicateContent.aggregateRecallMemory) {
+		return { row: unsavedAggregateResult(input.answer, input.key, input.project), saved: false };
+	}
+	linkAggregateSources(db, duplicateContent.row.id, input.sourceMemoryIds, input.agentId, input.now);
+	return { row: duplicateContent.row, saved: true };
+}
+
+function isUniqueConstraintError(err: unknown): boolean {
+	return err instanceof Error && /UNIQUE constraint failed|constraint failed/i.test(err.message);
+}
+
 async function embedAggregateMemory(
 	memoryId: string,
 	content: string,
@@ -387,6 +426,7 @@ export async function aggregateRecall(
 	const recall = deps.hybridRecall ?? hybridRecall;
 	const saveAggregate = params.saveAggregate !== false && params.save_aggregate !== false;
 	const log = deps.logger ?? logger;
+	const ingestEnvelope = deps.ingestEnvelope ?? txIngestEnvelope;
 	const first = await recall(params, cfg, deps.embedFn);
 
 	if (!deps.router) {
@@ -445,26 +485,22 @@ export async function aggregateRecall(
 	if (saveAggregate) {
 		const normalized = normalizeAndHashContent(answer);
 		row = getDbAccessor().withWriteTx((db) => {
-			const existing = loadAggregateByKey(db, key, { agentId, project });
-			if (existing) {
-				linkAggregateSources(db, existing.id, sourceMemoryIds, agentId, now);
+			const duplicate = resolveAggregateDuplicate(db, {
+				key,
+				agentId,
+				project,
+				contentHash: normalized.contentHash,
+				answer,
+				sourceMemoryIds,
+				now,
+			});
+			if (duplicate) {
 				deduped = true;
-				saved = true;
-				return existing;
-			}
-			const duplicateContent = loadMemoryByContentHash(db, normalized.contentHash, { agentId, project });
-			if (duplicateContent) {
-				if (!duplicateContent.visibleForAggregate || !duplicateContent.aggregateRecallMemory) {
-					deduped = true;
-					return unsavedAggregateResult(answer, key, project);
-				}
-				linkAggregateSources(db, duplicateContent.row.id, sourceMemoryIds, agentId, now);
-				deduped = true;
-				saved = true;
-				return duplicateContent.row;
+				saved = duplicate.saved;
+				return duplicate.row;
 			}
 			const id = deps.idFactory?.() ?? randomUUID();
-			txIngestEnvelope(db, {
+			const envelope: IngestEnvelope = {
 				id,
 				content: normalized.storageContent,
 				normalizedContent: normalized.normalizedContent || normalized.hashBasis,
@@ -488,7 +524,25 @@ export async function aggregateRecall(
 				agentId,
 				visibility: "global",
 				createdAt: now,
-			});
+			};
+			try {
+				ingestEnvelope(db, envelope);
+			} catch (err) {
+				if (!isUniqueConstraintError(err)) throw err;
+				const racedDuplicate = resolveAggregateDuplicate(db, {
+					key,
+					agentId,
+					project,
+					contentHash: normalized.contentHash,
+					answer,
+					sourceMemoryIds,
+					now,
+				});
+				if (!racedDuplicate) throw err;
+				deduped = true;
+				saved = racedDuplicate.saved;
+				return racedDuplicate.row;
+			}
 			linkAggregateSources(db, id, sourceMemoryIds, agentId, now);
 			saved = true;
 			return loadAggregateMemory(db, id);
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -39,6 +39,7 @@ interface AggregateRecallLogger {
 interface AggregateMemoryRow {
 	readonly id: string;
 	readonly content: string;
+	readonly source_type?: string | null;
 	readonly source_id: string | null;
 	readonly type: string;
 	readonly tags: string | null;
@@ -53,6 +54,7 @@ interface AggregateMemoryRow {
 interface ContentHashMatch {
 	readonly row: RecallResult;
 	readonly visibleForAggregate: boolean;
+	readonly aggregateRecallMemory: boolean;
 }
 
 const BUDGET_QUERY_LIMITS: Record<AggregateRecallBudget, number> = {
@@ -235,7 +237,7 @@ function loadMemoryByContentHash(
 ): ContentHashMatch | null {
 	const row = db
 		.prepare(
-			`SELECT id, content, source_id, type, tags, pinned, importance, who, project, visibility, created_at
+			`SELECT id, content, source_type, source_id, type, tags, pinned, importance, who, project, visibility, created_at
 			 FROM memories
 			 WHERE content_hash = ?
 			   AND COALESCE(NULLIF(agent_id, ''), 'default') = ?
@@ -249,6 +251,7 @@ function loadMemoryByContentHash(
 		row: rowToRecallResult(row),
 		visibleForAggregate:
 			(row.visibility ?? "global") === "global" && (input.project === null || row.project === input.project),
+		aggregateRecallMemory: row.source_type === "aggregate-recall",
 	};
 }
 
@@ -451,7 +454,7 @@ export async function aggregateRecall(
 			}
 			const duplicateContent = loadMemoryByContentHash(db, normalized.contentHash, { agentId, project });
 			if (duplicateContent) {
-				if (!duplicateContent.visibleForAggregate) {
+				if (!duplicateContent.visibleForAggregate || !duplicateContent.aggregateRecallMemory) {
 					deduped = true;
 					return unsavedAggregateResult(answer, key, project);
 				}
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -168,7 +168,13 @@ function parsePlannerQueries(raw: string): string[] {
 }
 
 function isSourceMemoryRow(row: RecallResult): boolean {
-	return !row.id.startsWith("constructed:");
+	return (
+		row.source !== "llm_summary" &&
+		!row.id.startsWith("constructed:") &&
+		!row.id.startsWith("summary:") &&
+		!row.id.startsWith("source-chunk:") &&
+		!row.id.startsWith("native-artifact:")
+	);
 }
 
 function uniqueEvidence(rows: readonly RecallResult[]): RecallResult[] {
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -69,8 +69,32 @@ const BUDGET_QUERY_LIMITS: Record<AggregateRecallBudget, number> = {
 	large: 8,
 };
 
+export class InvalidAggregateRecallBudgetError extends Error {
+	constructor() {
+		super("Invalid aggregateBudget. Expected one of: small, medium, large.");
+		this.name = "InvalidAggregateRecallBudgetError";
+	}
+}
+
+export function parseAggregateRecallBudget(raw: unknown): AggregateRecallBudget | null {
+	if (raw === undefined) return "small";
+	if (raw === "small" || raw === "medium" || raw === "large") return raw;
+	return null;
+}
+
+export function readAggregateRecallBudgetInput(input: {
+	readonly aggregateBudget?: unknown;
+	readonly aggregate_budget?: unknown;
+}): unknown {
+	if (Object.hasOwn(input, "aggregateBudget")) return input.aggregateBudget;
+	if (Object.hasOwn(input, "aggregate_budget")) return input.aggregate_budget;
+	return undefined;
+}
+
 function normalizeBudget(raw: unknown): AggregateRecallBudget {
-	return raw === "medium" || raw === "large" ? raw : "small";
+	const budget = parseAggregateRecallBudget(raw);
+	if (!budget) throw new InvalidAggregateRecallBudgetError();
+	return budget;
 }
 
 function emptyAggregateResponse(
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -3,6 +3,7 @@ import type { RouteRequest, RouterResult } from "@signet/core";
 import { normalizeAndHashContent } from "./content-normalization";
 import { type WriteDb, getDbAccessor } from "./db-accessor";
 import { syncVecDeleteBySourceId, syncVecInsert, vectorToBlob } from "./db-helpers";
+import { logger } from "./logger";
 import type { EmbeddingConfig, ResolvedMemoryConfig } from "./memory-config";
 import { type EmbedFn, type RecallParams, type RecallResponse, type RecallResult, hybridRecall } from "./memory-search";
 import { txIngestEnvelope } from "./transactions";
@@ -28,6 +29,11 @@ interface AggregateRecallDeps {
 	readonly embedFn: EmbedFn;
 	readonly now?: () => Date;
 	readonly idFactory?: () => string;
+	readonly logger?: AggregateRecallLogger;
+}
+
+interface AggregateRecallLogger {
+	readonly warn: (category: "memory", message: string, data?: Record<string, unknown>) => void;
 }
 
 interface AggregateMemoryRow {
@@ -377,6 +383,7 @@ export async function aggregateRecall(
 	const maxQueries = BUDGET_QUERY_LIMITS[budget];
 	const recall = deps.hybridRecall ?? hybridRecall;
 	const saveAggregate = params.saveAggregate !== false && params.save_aggregate !== false;
+	const log = deps.logger ?? logger;
 	const first = await recall(params, cfg, deps.embedFn);
 
 	if (!deps.router) {
@@ -484,9 +491,34 @@ export async function aggregateRecall(
 			return loadAggregateMemory(db, id);
 		});
 		if (row && !deduped) {
-			await embedAggregateMemory(row.id, row.content, normalized.contentHash, now, cfg.embedding, deps.embedFn).catch(
-				() => false,
-			);
+			let embedded = false;
+			let embeddingError: unknown;
+			try {
+				embedded = await embedAggregateMemory(
+					row.id,
+					row.content,
+					normalized.contentHash,
+					now,
+					cfg.embedding,
+					deps.embedFn,
+				);
+			} catch (err) {
+				embeddingError = err;
+			}
+			if (!embedded) {
+				log.warn("memory", "Aggregate recall memory saved without embedding", {
+					memoryId: row.id,
+					agentId,
+					project,
+					sourceId: key,
+					contentHash: normalized.contentHash,
+					sourceMemoryIds,
+					reason: embeddingError instanceof Error ? "embedding_exception" : "embedding_unavailable",
+					...(embeddingError instanceof Error
+						? { errorName: embeddingError.name, errorMessage: embeddingError.message }
+						: {}),
+				});
+			}
 		}
 	} else {
 		row = unsavedAggregateResult(answer, key, project);
--- a/platform/daemon/src/aggregate-recall.ts
+++ b/platform/daemon/src/aggregate-recall.ts
@@ -189,7 +189,7 @@ function uniqueEvidence(rows: readonly RecallResult[]): RecallResult[] {
 }
 
 function evidenceCanSaveAsGlobalAggregate(rows: readonly RecallResult[]): boolean {
-	return rows.every((row) => (row.visibility ?? "global") === "global" && (row.scope ?? null) === null);
+	return rows.every((row) => row.visibility === "global" && row.scope === null);
 }
 
 function aggregateKey(input: {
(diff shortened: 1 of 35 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm taking a look at the fixes in `fix(memory): harden aggregate recall` (commit `961a26e5`) and will follow up shortly.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `961a26e5`

</details>

I found one correctness issue in the aggregate recall save path: embedding failures are swallowed after the memory has already been persisted, which makes the saved aggregate look successful while leaving it keyword-only with no signal to callers or logs.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The changed aggregate save implementation explicitly catches and discards embedAggregateMemory failures in platform/daemon/src/aggregate-recall.ts. The PR description and checklist claim error handling/fallback paths are tested with no silent swallow, but this failure path is neither surfaced nor logged in the diff.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,515 @@
...
+		}
+	} else {
+		row = unsavedAggregateResult(answer, key, project);
+	}
+
+	return {
+		results: row ? [row] : [],
+		query: params.query,
+		method: "hybrid",
+		meta: {
```
PR-Reviewer-Ant, line 626: This silently discards embedding failures after the aggregate memory has already been saved. That leaves a persisted aggregate memory without vector coverage, but the response still reports `saved: true` and `stoppedReason: "complete"`, so callers and operators have no way to tell the saved recall artifact is degraded. Please at least log this with the memory id/context, or return explicit metadata matching the existing remember behavior for unembedded saves.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,547 @@
...
+				updatedBy: "signet",
+				sourceType: "aggregate-recall",
+				sourceId: key,
+				idempotencyKey: key,
+				scope: null,
+				agentId,
+				visibility: "global",
+				createdAt: now,
+			});
+			linkAggregateSources(db, id, sourceMemoryIds, agentId, now);
```
PR-Reviewer-Ant, line 584: This can return an ordinary visible memory as a saved aggregate when the synthesized answer happens to match existing content. `loadMemoryByContentHash()` does not require `source_type = 'aggregate-recall'` or the aggregate idempotency key, but this branch links provenance to that row, sets `saved = true`, and `rowToRecallResult()` reports it as `source: "aggregate-recall"`. That breaks the PR's contract that saved aggregate answers are normal memories with aggregate provenance/source metadata, and it can attach aggregate source links to unrelated memories. If content-hash uniqueness prevents inserting a new aggregate, this should either return an unsaved aggregate like the inaccessible duplicate case or only treat existing aggregate-recall rows as saved aggregates.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `59d8ad4c`

</details>

I found one correctness issue in aggregate dedupe: a saved aggregate can be reported as an aggregate memory even when the row reused by content hash is an ordinary existing memory.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The behavior is directly visible in platform/daemon/src/aggregate-recall.ts: content-hash dedupe loads any visible memory with the same content hash, then rowToRecallResult hard-codes source to aggregate-recall and the final metadata marks it saved.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Reviewer re-check requested on latest head `b11e53d2`.

Since the last pass:
- Addressed the aggregate content-hash dedupe issue by preventing visible ordinary memories from being reused as saved aggregate dedupes.
- Added a regression test for visible ordinary duplicate content.
- Re-ran targeted aggregate/daemon/MCP/CLI/SDK tests, typecheck, and scoped Biome checks.

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in 59d8ad4c (`fix(memory): log aggregate embedding failures`).

The aggregate save path now logs a structured warning when embedding fails or returns no usable vector after the memory is persisted. The warning includes memory id, agent id, project, source id, content hash, source memory ids, and the failure reason. I also added a regression test for the saved-but-unembedded path.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/core/src/recall.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/memory.test.ts libs/sdk/src/__tests__/client.test.ts libs/sdk/src/__tests__/openai.test.ts libs/sdk/src/__tests__/ai-sdk.test.ts`
- `bun run typecheck`
- `bunx biome check platform/daemon/src/aggregate-recall.ts platform/daemon/src/aggregate-recall.test.ts`

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in b11e53d2 (`fix(memory): avoid aggregate dedupe onto ordinary memories`).

The content-hash duplicate path now only treats existing rows as saved aggregate dedupes when the row itself has `source_type = 'aggregate-recall'`. Visible ordinary memories with duplicate content now return an unsaved aggregate result, do not attempt a conflicting insert, and do not receive aggregate provenance links.

Added a regression test covering a visible ordinary duplicate row and asserting that it is not returned as the aggregate row and receives no `aggregate_memory_sources` links.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/core/src/recall.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/memory.test.ts libs/sdk/src/__tests__/client.test.ts libs/sdk/src/__tests__/openai.test.ts libs/sdk/src/__tests__/ai-sdk.test.ts`
- `bun run typecheck`
- `bunx biome check platform/daemon/src/aggregate-recall.ts platform/daemon/src/aggregate-recall.test.ts`

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,550 @@
...
+					deduped = true;
+					return unsavedAggregateResult(answer, key, project);
+				}
+				linkAggregateSources(db, duplicateContent.row.id, sourceMemoryIds, agentId, now);
+				deduped = true;
+				saved = true;
+				return duplicateContent.row;
+			}
+			const id = deps.idFactory?.() ?? randomUUID();
+			txIngestEnvelope(db, {
```
PR-Reviewer-Ant: This save path still has a check-then-insert race. Two aggregate recalls with the same synthesized answer can both miss `loadAggregateByKey` and `loadMemoryByContentHash`, then both reach `txIngestEnvelope`; whichever loses the existing unique constraint on `content_hash` or idempotency will bubble out as a failed recall. That undercuts the PR's stated goal of returning a deduped or unsaved aggregate instead of 500ing on duplicate content. The insert should either handle the unique-conflict error by reloading the existing aggregate/content-hash row, or use an atomic insert/upsert pattern that preserves the same visibility checks.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b11e53d2`

</details>

I found one race in the new aggregate-save path that can still turn duplicate aggregate content into a failed recall instead of the graceful dedupe behavior the PR describes.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The race is visible in platform/daemon/src/aggregate-recall.ts: the code checks for existing idempotency/content-hash matches, then inserts the aggregate memory in a separate step. I did not run a concurrent repro, so confidence is medium rather than high.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in efce2ed7 (`fix(memory): handle aggregate insert races`).

The aggregate save path now catches unique constraint failures from the insert and re-runs the same idempotency/content-hash duplicate resolver inside the transaction. If a concurrent aggregate insert won the race, the caller now gets that saved aggregate as a deduped result. If the conflict is ordinary or inaccessible duplicate content, the caller gets the existing unsaved aggregate fallback instead of a failed recall.

Added a deterministic regression test that injects the winning aggregate insert immediately before the losing insert and verifies the loser returns the winner as `saved: true`, `deduped: true`, with provenance links recorded only on the aggregate row.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/core/src/recall.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/memory.test.ts libs/sdk/src/__tests__/client.test.ts libs/sdk/src/__tests__/openai.test.ts libs/sdk/src/__tests__/ai-sdk.test.ts`
- `bun run typecheck`
- `bunx biome check platform/daemon/src/aggregate-recall.ts platform/daemon/src/aggregate-recall.test.ts`

NicholaiVogel:
Reviewer re-check requested on latest head `efce2ed7`.

Since the last pass:
- Handled unique-constraint races in aggregate save by reloading through the aggregate duplicate resolver after failed insert.
- Added a deterministic regression test for a concurrent aggregate insert winning the content-hash race.
- Re-ran targeted aggregate/daemon/MCP/CLI/SDK tests, typecheck, and scoped Biome checks.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -2315,7 +2327,13 @@ export function registerMemoryRoutes(app: Hono): void {
 				policyGroup: agentScope.policyGroup,
```
PR-Reviewer-Ant, line 2348: This call drops the aggregate-specific request fields before entering `aggregateRecall`. The guard above treats `saveAggregate: false` as read-only and skips `remember`, but `aggregateRecall` defaults saving to true when `params.saveAggregate`/`params.save_aggregate` are absent. So `POST /api/memory/recall` with `{ aggregate: true, saveAggregate: false }` can still persist an aggregate memory without the permission the PR says is required. The same missing forwarding also means `aggregateBudget` is ignored on this route. Please include the aggregate budget and save flag in `params`, and add a route-level regression for the `saveAggregate: false` case.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `efce2ed7`

</details>

I found one concrete issue in the main recall route: the permission guard honors `saveAggregate: false`, but the request options passed into aggregate recall drop that flag and the requested budget. That can turn an explicitly read-only aggregate recall into a saved memory.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The route-level guard checks body.saveAggregate/body.save_aggregate before requiring remember permission, but the params object passed to aggregateRecall on the changed call path does not include saveAggregate, save_aggregate, aggregateBudget, or aggregate_budget. aggregateRecall defaults saveAggregate to true when those fields are absent, so the behavior is directly implied by the diff.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in bb92434c (`fix(memory): forward aggregate recall save options`).

The `/api/memory/recall` route now forwards aggregate-specific options explicitly into `aggregateRecall`, including `aggregateBudget`/`aggregate_budget` and `saveAggregate`/`save_aggregate`. This keeps the route permission decision and aggregate save behavior aligned for read-only aggregate requests.

Added a route-level regression using a readonly token and a mocked aggregate path. The test calls `/api/memory/recall` with `aggregate: true`, `aggregateBudget: "large"`, and `saveAggregate: false`, then asserts those options are preserved in the params passed to `aggregateRecall` and the response remains unsaved.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/core/src/recall.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/memory.test.ts libs/sdk/src/__tests__/client.test.ts libs/sdk/src/__tests__/openai.test.ts libs/sdk/src/__tests__/ai-sdk.test.ts`
- `bun run typecheck`
- `bunx biome check platform/daemon/src/routes/memory-routes.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`

NicholaiVogel:
Reviewer re-check requested on latest head `bb92434c`.

Since the last pass:
- Made `/api/memory/recall` aggregate option forwarding explicit for budget and save flags.
- Added a route-level regression proving `saveAggregate: false` remains read-only and reaches `aggregateRecall`.
- Re-ran targeted aggregate/daemon/MCP/CLI/SDK tests, typecheck, and scoped Biome checks.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,604 @@
...
+	readonly row: RecallResult;
+	readonly saved: boolean;
+}
+
+const BUDGET_QUERY_LIMITS: Record<AggregateRecallBudget, number> = {
+	small: 3,
+	medium: 5,
+	large: 8,
+};
+
```
PR-Reviewer-Ant, line 71: `normalizeBudget` silently treats any unknown value as `small`. Since `aggregateBudget` is a new external API/CLI/MCP/SDK input and the docs advertise a closed set, the route should reject out-of-range values with a 400 instead of saving or returning an aggregate under a budget the caller did not actually ask for. This matters more here because aggregate mode can spend LLM calls and persist a memory by default.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `bb92434c`

</details>

I found one input-boundary issue: invalid aggregate budget values are silently coerced to `small` instead of being rejected. That undercuts the new explicit aggregate mode contract and can make callers think they requested a larger or different budget than the daemon actually used.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The new aggregate path accepts `aggregateBudget` from request bodies in `platform/daemon/src/routes/memory-routes.ts` and normalizes anything except `medium` or `large` to `small` in `platform/daemon/src/aggregate-recall.ts`. I did not run the PR tests because the supplied review workspace does not contain the Signet repo checkout.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in `1245db48`.

Invalid aggregate budgets now fail at the explicit recall boundaries instead of being coerced:
- `normalizeBudget` now uses a closed-set parser and throws on invalid values.
- `/api/memory/recall` and `/api/hooks/recall` return 400 for provided out-of-range budget fields before invoking aggregation.
- Added regressions for the aggregate helper, the memory recall route, and the hooks recall route.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/daemon/src/hooks-recall.test.ts`
- Original targeted aggregate/MCP/CLI/SDK suite
- `bun run typecheck`
- `git diff --check`

NicholaiVogel:
Reviewer re-check requested on latest head `1245db48`.

Since the last pass:
- Rejected invalid aggregate budgets instead of coercing to `small`.
- Added 400 responses for invalid budget fields on `/api/memory/recall` and `/api/hooks/recall` before aggregation runs.
- Added regression coverage for the aggregate helper and both explicit recall routes.
- Re-ran aggregate/auth route tests, hooks recall tests, the targeted aggregate/MCP/CLI/SDK suite, typecheck, scoped Biome checks, and whitespace checks.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,628 @@
...
+					key,
+					agentId,
+					project,
+					contentHash: normalized.contentHash,
+					answer,
+					sourceMemoryIds,
+					now,
+				});
+				if (!racedDuplicate) throw err;
+				deduped = true;
```
PR-Reviewer-Ant, line 580: This saves every persisted aggregate as `visibility: "global"` with `scope: null`, regardless of the visibility/scope of the evidence rows used to synthesize it. Since aggregate recall runs through the caller's normal scoped recall path, it can summarize private/session/otherwise narrower evidence and then persist that derived content as a globally visible normal memory. That is a scope-widening write, not just metadata, and it can leak the substance of scoped memories through the aggregate. Please either restrict saved aggregates to evidence that is safe to globalize, persist the aggregate with a conservative visibility/scope derived from the source rows, or force `saveAggregate: false` whenever the evidence set is not globally visible.

surfaces/cli/src/commands/memory.ts
```diff
@@ -95,10 +95,15 @@ export function registerMemoryCommands(program: Command, deps: MemoryDeps): void
 		.option("--importance-min <n>", "Only return memories at or above this importance", Number.parseFloat)
 		.option("--min-score <n>", "Minimum recall score threshold (client-side)", Number.parseFloat)
 		.option("--agent <name>", "Filter by agent ID")
+		.option("--aggregate", "Synthesize an aggregate answer from bounded recall evidence", false)
+		.option("--aggregate-budget <budget>", "Aggregate recall budget: small, medium, or large", "small")
+		.option("--no-save-aggregate", "Return the aggregate answer without saving it as a memory")
 		.option("--json", "Output as JSON")
 		.action(async (query: string, options) => {
 			if (!(await deps.ensureDaemonForSecrets())) return;
```
PR-Reviewer-Ant, line 105: `--no-save-aggregate` by itself now turns aggregate mode on because `aggregateRequested` includes `options.saveAggregate === false`. That makes a flag that sounds like a modifier unexpectedly switch recall from normal search into LLM-backed aggregate synthesis. It is probably better for `--no-save-aggregate` to only affect the request when `--aggregate` or an explicit aggregate budget was supplied.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `1245db48`

</details>

I found one scope/visibility issue that should be fixed before merge. The aggregate path can save a synthesized memory as globally visible even when its evidence came from narrower scoped recall results.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The new aggregate save path explicitly writes aggregate memories with scope null and visibility global in platform/daemon/src/aggregate-recall.ts. The route passes through the caller's agent read policy and project filters, so the aggregate can be synthesized from whatever rows that scoped recall returned; I did not have the full buildAgentScopeClause implementation in the prompt, so the exact blast radius depends on those visibility semantics, but the widening write is visible in this diff.

**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in `8374dded`.

Aggregate recall now carries `visibility` and `scope` through hydrated memory results, and persisted aggregate saves only run when all evidence is globally visible and unscoped. If any evidence is private or scoped, the aggregate answer is returned unsaved with `saved: false`, so the synthesized content is not widened into a global normal memory.

Added a regression that private evidence produces an unsaved aggregate and does not insert the aggregate memory row.

surfaces/cli/src/commands/memory.ts
NicholaiVogel: Fixed in `8374dded`.

`--no-save-aggregate` no longer enables aggregate recall by itself. It is now only forwarded when aggregate mode was explicitly requested via `--aggregate` or a non-default `--aggregate-budget`.

Added a CLI regression proving `signet recall <query> --no-save-aggregate` stays a normal recall request.

NicholaiVogel:
Reviewer re-check requested on latest head `8374dded`.

Since the last pass:
- Prevented aggregate saves from widening private/scoped evidence into global memories; non-global evidence now returns an unsaved aggregate result.
- Threaded `visibility` and `scope` through hydrated memory recall rows for aggregate safety checks.
- Made `--no-save-aggregate` a modifier only; it no longer turns normal recall into aggregate recall by itself.
- Added regressions for private-evidence aggregate saves and CLI no-save behavior.
- Re-ran aggregate/auth route tests, hooks recall tests, targeted aggregate/MCP/CLI/SDK tests, typecheck, scoped Biome checks, and whitespace checks.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,632 @@
...
+	for (const candidate of [query, ...candidates]) {
+		const trimmed = candidate.trim();
+		const key = normalizeQuery(trimmed);
+		if (!trimmed || seen.has(key)) continue;
+		seen.add(key);
+		result.push(trimmed);
+		if (result.length >= max) break;
+	}
+	return result;
+}
```
PR-Reviewer-Ant, line 145: This helper still admits non-memory recall rows into aggregate evidence. It only excludes ids starting with `constructed:`, but hybrid recall can include supplementary rows such as LLM summary/context rows that are not durable memories. Those ids then flow into `sourceMemoryIds` and `linkAggregateSources()`, so a saved aggregate can get provenance links to rows that do not exist in `memories`, and the synthesized answer can be based partly on prior synthesized text rather than unique evidence memories. Since this PR explicitly adds aggregate source provenance, please filter to actual memory-backed rows here, for example by excluding `supplementary` rows and non-memory sources such as `llm_summary`, or by validating ids against `memories` before linking.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `8374dded`

</details>

I found one data-integrity issue in the aggregate evidence selection. The new persistence path can treat synthesized/support rows as source memories, which undermines the provenance links this PR is trying to add.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The aggregate path builds sourceMemoryIds directly from uniqueEvidence(), and that helper only filters ids starting with constructed:. The changed hybrid recall result type now carries supplementary rows, and existing recall can include LLM summary/support rows that are not real memory records, so the new aggregate_memory_sources table can receive non-memory ids.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in d3fec2c5 and pushed at 907a7dcf after merging main/#740.

`uniqueEvidence()` now filters out synthetic/support rows before synthesis/provenance (`llm_summary`, `constructed:`, `summary:`, `source-chunk:`, and `native-artifact:`), so `sourceMemoryIds` and `aggregate_memory_sources` only receive durable memory-backed rows. Added a regression covering an LLM summary row plus a real memory row; the aggregate links only the real memory.

Validation after merging #740:
- `bun run build:core`
- targeted aggregate/auth/core/mcp/CLI/SDK/OpenClaw/session-dedupe/memory-search suite: 310 pass
- `bun test platform/daemon/src/hooks-recall.test.ts`: 13 pass
- `bun run typecheck`
- scoped `bunx biome check` on changed TS files (8 pre-existing `noExplicitAny` warnings remain)
- `git diff --check`

NicholaiVogel:
Pushed 907a7dcf with PR #740 merged cleanly into this branch.

Resolved the latest aggregate provenance review item by excluding synthetic/support recall rows from aggregate evidence and source-link creation. Also kept #740's session recall dedupe fields wired through the aggregate surfaces during the merge.

Local validation:
- bun run build:core
- bun test platform/daemon/src/aggregate-recall.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/core/src/recall.test.ts platform/core/src/migrations/migrations.test.ts platform/daemon/src/mcp/tools.test.ts surfaces/cli/src/commands/memory.test.ts libs/sdk/src/__tests__/client.test.ts libs/sdk/src/__tests__/openai.test.ts libs/sdk/src/__tests__/ai-sdk.test.ts integrations/openclaw/memory-adapter/src/index.test.ts platform/daemon/src/session-recall-dedupe.test.ts platform/daemon/src/memory-search.test.ts
- bun test platform/daemon/src/hooks-recall.test.ts
- bun run typecheck
- bunx biome check on changed TS files (only existing noExplicitAny warnings in memory search/routes)
- git diff --check

PR-Reviewer-Ant, please re-review the latest head.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,638 @@
...
+		!row.id.startsWith("source-chunk:") &&
+		!row.id.startsWith("native-artifact:")
+	);
+}
+
+function uniqueEvidence(rows: readonly RecallResult[]): RecallResult[] {
+	const seen = new Set<string>();
+	const result: RecallResult[] = [];
+	for (const row of rows) {
+		if (!isSourceMemoryRow(row) || seen.has(row.id)) continue;
```
PR-Reviewer-Ant, line 184: This guard defaults missing `visibility` to `global`, so any recall row that does not populate the newly optional `visibility`/`scope` fields is treated as safe to persist into a global aggregate memory. That is a bad failure mode for a privacy boundary: the PR claims private evidence is not saved into global aggregates, but this only holds for result producers that were updated to include those fields. Please fail closed here, e.g. require `row.visibility === "global" && row.scope === null`, and only save when every evidence row carries explicit global metadata.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `907a7dcf`

</details>

I found one data-exposure bug in the aggregate save gate. The new path treats missing visibility metadata as globally saveable, which undercuts the PR's own private-evidence guard.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The unsafe default is explicit in aggregate-recall.ts, and RecallResult.visibility/scope are optional in the changed memory-search type. I could not verify every existing RecallResult producer from the provided context, so the exact set of affected paths is uncertain, but the aggregate save decision should fail closed regardless.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in 2d2b3f65.

Aggregate save eligibility now fails closed: evidence rows must carry `visibility === "global"` and `scope === null` explicitly before the aggregate is persisted. Missing metadata now returns an unsaved aggregate instead of assuming global visibility. I also tightened the aggregate/content-hash dedupe checks so existing rows with missing visibility are not treated as saved aggregate matches.

Added a regression for evidence without explicit scope metadata; it synthesizes the aggregate response but does not write `aggregate-legacy`.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts`: 12 pass
- `bun test platform/daemon/src/hooks-recall.test.ts`: 13 pass
- `bun run typecheck`
- `bunx biome check platform/daemon/src/aggregate-recall.ts platform/daemon/src/aggregate-recall.test.ts`
- `git diff --check`

Note: running auth co-location and hooks recall in the same Bun process still order-pollutes auth mode and makes hooks return 401; the hook suite passes in isolation as before.

NicholaiVogel:
Pushed 2d2b3f65 for the latest reviewer scope-boundary feedback.

Change: aggregate save eligibility now fails closed unless every evidence row has explicit `visibility: "global"` and `scope: null`; missing metadata returns an unsaved aggregate. Added regression coverage for that case.

PR-Reviewer-Ant, please re-review the latest head.

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,637 @@
...
+		seen.add(row.id);
+		result.push(row);
+	}
+	return result;
+}
+
+function evidenceCanSaveAsGlobalAggregate(rows: readonly RecallResult[]): boolean {
+	return rows.every((row) => row.visibility === "global" && row.scope === null);
+}
+
```
PR-Reviewer-Ant, line 194: This lookup should also require `source_type = 'aggregate-recall'`. As written, any visible global memory with a matching `idempotency_key` is treated as the saved aggregate memory, and `resolveAggregateDuplicate()` will then call `linkAggregateSources()` for that ordinary row. Since `remember` exposes caller-provided idempotency keys elsewhere in the API, this can misclassify a normal memory as an aggregate and attach provenance links to the wrong record. The content-hash duplicate path already checks `aggregateRecallMemory`; the idempotency-key path needs the same guard.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `2d2b3f65`

</details>

I found one data-integrity issue in the aggregate dedupe path: idempotency-key reuse can attach aggregate provenance to a non-aggregate memory. The rest of the changed surfaces look consistent with the stated aggregate recall behavior from the diff.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible in the new aggregate recall implementation: loadAggregateByKey selects by idempotency_key without checking source_type, while aggregate source links are inserted for whatever row it returns. I did not run the test suite or reproduce this against a live database in this review context.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/aggregate-recall.ts
NicholaiVogel: Fixed in 65434cc0.

`loadAggregateByKey()` now requires `source_type = 'aggregate-recall'` in addition to the agent, visibility, scope, and idempotency key checks. If an ordinary memory owns the same idempotency key, aggregate recall no longer links provenance to it or misreports it as a saved aggregate. The unique-conflict fallback now returns an unsaved deduped aggregate when no valid aggregate row can be reloaded.

Added a regression with a normal `source_type = 'manual'` memory using the aggregate idempotency key. The result is unsaved/deduped, no aggregate row is inserted, and no source links are attached to the ordinary memory.

Validation:
- `bun test platform/daemon/src/aggregate-recall.test.ts`: 13 pass
- `bun run typecheck`
- `bunx biome check platform/daemon/src/aggregate-recall.ts platform/daemon/src/aggregate-recall.test.ts`
- `git diff --check`

NicholaiVogel:
Pushed 65434cc0 for the idempotency-key reviewer feedback.

Change: existing rows are only reused by aggregate idempotency key when `source_type = 'aggregate-recall'`; ordinary memories with matching caller-provided idempotency keys now produce an unsaved deduped aggregate instead of getting provenance links.

PR-Reviewer-Ant, please re-review the latest head.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `65434cc0`

</details>

I did not find a concrete correctness, security, data integrity, or breaking-change issue in the provided diff. The aggregate recall path appears to enforce the new save permission boundary, validate budgets, preserve read-only aggregate mode, and add regression coverage for the main duplicate/provenance cases.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the new route guards, aggregate implementation, migration registration, SDK/CLI/MCP forwarding, and targeted tests for auth, invalid budgets, dedupe, provenance links, and read-only aggregate mode. I could not run the test plan or inspect a live daemon/runtime from this review context, so confidence is medium rather than high.

**Since my last review:**
- Still blocking: 0
- Newly found: 0

platform/daemon/src/aggregate-recall.ts
```diff
@@ -0,0 +1,642 @@
...
+		row.source !== "llm_summary" &&
+		!row.id.startsWith("constructed:") &&
+		!row.id.startsWith("summary:") &&
+		!row.id.startsWith("source-chunk:") &&
+		!row.id.startsWith("native-artifact:")
+	);
+}
+
+function uniqueEvidence(rows: readonly RecallResult[]): RecallResult[] {
+	const seen = new Set<string>();
```
PR-Reviewer-Ant, line 181: Saved aggregate rows loaded from the database drop `visibility` and `scope` here, but later `evidenceCanSaveAsGlobalAggregate()` requires every evidence row to have `visibility === "global"` and `scope === null`. That means if a prior saved aggregate is recalled and becomes evidence for a later aggregate, it will fail the save eligibility check because the row no longer carries the DB scope metadata. Please preserve `visibility` and `scope` in `rowToRecallResult()` and include them in the aggregate-row SELECTs that feed it.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `b590231f`

</details>

I found one correctness issue in the aggregate recall implementation: saved aggregate rows lose their visibility/scope fields in the response, which can make a saved aggregate look unsafe to save again when used as later evidence.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible from the new aggregate recall code path: saved aggregate rows are loaded through rowToRecallResult(), but that conversion omits visibility and scope even though aggregate saving later depends on those fields. I did not run the new aggregate recall tests in this review context.

**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR #748, opened 2026-05-21
fix(core): throw on unimplemented migration sources instead of silently no-oping

Alexi5000:
﻿## Summary

- Replace silent `console.log` stubs in `migrateChatGPT()`, `migrateClaude()`, and `migrateGemini()` with explicit `throw new Error()` calls
- Prefix unused `inputPath` parameters with underscore to satisfy linters
- Callers of `migrate()` now get a clear error instead of a successful no-op when using an unimplemented source

## Motivation

The current stubs in `platform/core/src/migrate.ts` log a message and return successfully, which means callers have no way to distinguish between a successful migration and one that did nothing. This can mask issues in tooling that depends on `migrate()`.

## Test Plan

- [ ] Calling `migrate({ source: 'chatgpt', inputPath: '...' })` now throws with a descriptive error
- [ ] Same for `claude` and `gemini` sources
- [ ] The `default` case in the switch still throws for unknown sources

### DIFF
--- a/platform/core/src/migrate.ts
+++ b/platform/core/src/migrate.ts
@@ -24,18 +24,27 @@ export async function migrate(options: MigrationOptions): Promise<void> {
 	}
 }
 
-async function migrateChatGPT(inputPath: string): Promise<void> {
+async function migrateChatGPT(_inputPath: string): Promise<void> {
 	// TODO: Parse ChatGPT export (conversations.json)
 	// Extract: user preferences, facts mentioned, project context
-	console.log(`Migrating from ChatGPT export: ${inputPath}`);
+	throw new Error(
+		"ChatGPT migration is not yet implemented. " +
+			"See https://github.com/Signet-AI/signetai/issues for tracking.",
+	);
 }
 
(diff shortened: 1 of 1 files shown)


PR-Reviewer-Ant:
Hi @Alexi5000 - I'm taking a look at the fixes in `fix(core): throw on unimplemented migration sources instead of silently no-oping` (commit `8e98444c`) and will follow up shortly.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `8e98444c`

</details>

The implementation matches the PR description: the three currently unimplemented migration sources now fail explicitly instead of logging and returning success. I did not find a correctness, security, or data-integrity issue in this diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed file is small and the diff directly covers the described behavior in platform/core/src/migrate.ts. Each implemented switch branch still dispatches as before, and each unimplemented source now throws a descriptive Error instead of silently returning.

NicholaiVogel:
Heh, pretty good. Thank you for these @Alexi5000 :P

PR #727, opened 2026-05-17
feat(sources): add GitHub project source with embedding + knowledge graph

aaf2tbz:
## Summary
- Adds a `"github"` source kind that indexes GitHub repos (issues, PRs, discussions, docs) into Signet's recall system.
- Follows the Obsidian source pattern: direct embedding into `embeddings` table + knowledge graph construction — not the connector/document pipeline pattern.
- Multi-repo support with `"owner/*"` glob expansion, PAT via Signet Secrets, continuous 5-min polling bridge.

## Validation
- `bun test platform/core/src/sources-config.test.ts platform/daemon/src/github-source-embeddings.test.ts platform/daemon/src/github-source-graph.test.ts`
- `bunx biome check platform/daemon/src/github-source-fetch.ts platform/daemon/src/github-source-embeddings.ts platform/daemon/src/github-source-graph.ts platform/daemon/src/github-source-bridge.ts platform/core/src/sources-config.ts surfaces/cli/src/commands/sources.ts surfaces/cli/src/features/sources.ts platform/daemon/src/routes/sources-routes.ts platform/daemon/src/daemon.ts`
- `bun run build`
- 22 tests pass (5 config, 5 embeddings, 4 graph refs, 8 pre-existing)
- All 6 code quality bot findings resolved in `b697bb4b`

## Notes
- No schema or migration changes — GitHub source entries are stored in the existing `sources.json` config with a `settings` field.
- No `octokit` dependency — raw `fetch()` against GitHub REST v3 + GraphQL v4.
- `root` field left empty for GitHub sources (filesystem-coupled in native-memory-sources).

## PR Readiness (MANDATORY)
- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

## Migration Notes (if applicable)
- [x] Migration is idempotent
- [x] Daemon Rust parity reviewed or explicitly N/A
- [x] Rollback / compatibility note included in PR description

Rollback / compatibility: no migration or persisted data change; rollback is removing the source config entry and reverting the commit.

PR #749, opened 2026-05-23
feat(sources): add GitHub source indexing

aaf2tbz:
## Summary

Rebuilds #727 as a clean single-commit GitHub source implementation on current `main`. Adds GitHub repo sources for issues, PRs, discussions, docs, embeddings, graph structure, polling, CLI, API, and docs.

## Changes

- Adds `github` source config support in `@signet/core` with agent-scoped ownership, validation, defaults, and parser coverage.
- Adds daemon GitHub fetching, embedding, graph indexing, stale reconciliation, polling bridge, source-index jobs, and purge support.
- Adds `POST /api/sources/github` plus delete hardening for GitHub source ownership and in-flight syncs.
- Adds `signet sources add github` CLI support and API docs.
- Rebuilds the old #727 feature shape while tightening the review-failure areas: token resolution, admin permissions, bounded fetches, partial-failure handling, canonical graph keys, and source deletion races.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Rollback / compatibility: no database migration or persisted schema change. GitHub source entries are stored in existing `sources.json`; rollback is removing the GitHub source config entry and reverting this commit. Rust daemon parity is N/A because this extends the TypeScript daemon source indexing path only.

## Testing

- [ ] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Validated with:

- `bunx biome check platform/core/src/sources-config.ts platform/core/src/sources-config.test.ts platform/daemon/src/github-source-bridge.ts platform/daemon/src/github-source-fetch.ts platform/daemon/src/github-source-graph.ts platform/daemon/src/routes/sources-routes.ts surfaces/cli/src/features/sources.ts surfaces/cli/src/commands/sources.ts`
- `bun test platform/core/src/sources-config.test.ts platform/daemon/src/github-source-embeddings.test.ts platform/daemon/src/github-source-graph.test.ts`
- `bun test platform/daemon/src/routes/sources-routes.test.ts`
- `cd platform/daemon && bun run build`

`cd platform/daemon && bun run build:types` still hits existing broad repo TypeScript/rootDir issues unrelated to this branch, so targeted tests plus production builds were used for local validation.

## AI disclosure

## Notes

This replaces #727, which accumulated a long review/comment trail after the broad feature landed before the hardening was complete. This PR keeps the same feature goal but presents it as one reviewed implementation commit.

### DIFF
--- a/platform/core/src/sources-config.ts
+++ b/platform/core/src/sources-config.ts
@@ -275,12 +275,13 @@ function addGitHubSourceChecked(input: AddGitHubSourceInput, agentsDir = getAgen
 	);
 
 	if (existing) {
+		const existingSettings = parseGitHubSettings(existing.settings);
 		const updated: SignetSourceEntry = {
 			...existing,
 			name: cleanName(input.name) ?? existing.name,
 			enabled: true,
 			updatedAt: now,
-			settings: buildGitHubSettings(input, repos),
+			settings: buildGitHubSettings(input, repos, existingSettings),
 			agentId,
 		};
 		saveSourcesConfig(
@@ -309,22 +310,28 @@ function addGitHubSourceChecked(input: AddGitHubSourceInput, agentsDir = getAgen
 	return { ok: true, source, created: true };
 }
 
-function buildGitHubSettings(input: AddGitHubSourceInput, repos: readonly string[]): Readonly<Record<string, unknown>> {
-	const tokenRef = input.tokenRef?.trim() || undefined;
+function buildGitHubSettings(
+	input: AddGitHubSourceInput,
+	repos: readonly string[],
+	existing?: GitHubSourceSettings,
+): Readonly<Record<string, unknown>> {
+	const tokenRef = input.tokenRef !== undefined ? input.tokenRef.trim() || undefined : existing?.tokenRef;
 	const resourceTypes = input.resourceTypes
 		? [...input.resourceTypes]
-		: tokenRef
-			? [...DEFAULT_GITHUB_RESOURCE_TYPES]
-			: [...DEFAULT_GITHUB_RESOURCE_TYPES_NO_TOKEN];
+		: existing?.resourceTypes?.length
+			? [...existing.resourceTypes]
+			: tokenRef
+				? [...DEFAULT_GITHUB_RESOURCE_TYPES]
+				: [...DEFAULT_GITHUB_RESOURCE_TYPES_NO_TOKEN];
 	return {
 		repos: repos,
 		tokenRef,
 		resourceTypes,
-		state: input.state ?? "all",
-		includeComments: input.includeComments ?? true,
-		labels: input.labels ? cleanStringArray(input.labels) : undefined,
-		docPaths: input.docPaths ? cleanStringArray(input.docPaths) : [...DEFAULT_GITHUB_DOC_PATHS],
-		maxItemsPerRepo: input.maxItemsPerRepo ?? 500,
+		state: input.state ?? existing?.state ?? "all",
+		includeComments: input.includeComments ?? existing?.includeComments ?? true,
+		labels: input.labels !== undefined ? cleanStringArray(input.labels) : existing?.labels,
+		docPaths: input.docPaths !== undefined ? cleanStringArray(input.docPaths) : (existing?.docPaths ?? [...DEFAULT_GITHUB_DOC_PATHS]),
+		maxItemsPerRepo: input.maxItemsPerRepo ?? existing?.maxItemsPerRepo ?? 500,
 	};
 }
 
--- a/platform/core/src/sources-config.ts
+++ b/platform/core/src/sources-config.ts
@@ -3,8 +3,23 @@ import { existsSync, mkdirSync, readFileSync, renameSync, rmSync, statSync, writ
 import { homedir } from "node:os";
 import { dirname, resolve } from "node:path";
 
-export type SignetSourceKind = "obsidian";
+export type SignetSourceKind = "obsidian" | "github";
 export type SignetSourceMode = "read-only";
+
+export interface GitHubSourceSettings {
+	readonly repos: readonly string[];
+	readonly tokenRef?: string;
+	readonly resourceTypes: readonly ("issues" | "pulls" | "discussions" | "docs")[];
+	readonly state?: "open" | "closed" | "all";
+	readonly includeComments?: boolean;
+	readonly labels?: readonly string[];
+	readonly docPaths?: readonly string[];
+	readonly maxItemsPerRepo?: number;
+}
+
+export const DEFAULT_GITHUB_RESOURCE_TYPES = ["issues", "pulls", "discussions", "docs"] as const;
+const VALID_GITHUB_RESOURCE_TYPES = new Set<string>(DEFAULT_GITHUB_RESOURCE_TYPES);
+export const DEFAULT_GITHUB_DOC_PATHS = ["README.md", "CHANGELOG.md"] as const;
 
 export interface SignetSourceEntry {
 	readonly id: string;
@@ -17,6 +32,8 @@ export interface SignetSourceEntry {
 	readonly updatedAt: string;
 	readonly lastIndexedAt?: string;
 	readonly excludeGlobs?: readonly string[];
+	readonly settings?: Readonly<Record<string, unknown>>;
+	readonly agentId?: string;
 }
 
 export const DEFAULT_OBSIDIAN_EXCLUDE_GLOBS = [
@@ -37,6 +54,20 @@ export interface AddObsidianSourceInput {
 	readonly name?: string;
 	readonly excludeGlobs?: readonly string[];
 	readonly now?: string;
+}
+
+export interface AddGitHubSourceInput {
+	readonly repos: readonly string[];
+	readonly name?: string;
+	readonly tokenRef?: string;
+	readonly resourceTypes?: readonly ("issues" | "pulls" | "discussions" | "docs")[];
+	readonly state?: "open" | "closed" | "all";
+	readonly includeComments?: boolean;
+	readonly labels?: readonly string[];
+	readonly docPaths?: readonly string[];
+	readonly maxItemsPerRepo?: number;
+	readonly now?: string;
+	readonly agentId?: string;
 }
 
 export type AddSourceResult =
@@ -163,6 +194,161 @@ function addObsidianSourceChecked(input: AddObsidianSourceInput, agentsDir = get
 	return { ok: true, source, created: true };
 }
 
+export function addGitHubSource(input: AddGitHubSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
+	return withSourcesConfigLock(agentsDir, () => addGitHubSourceUnlocked(input, agentsDir));
+}
+
+function addGitHubSourceUnlocked(input: AddGitHubSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
+	try {
+		return addGitHubSourceChecked(input, agentsDir);
+	} catch (err) {
+		const detail = err instanceof Error ? err.message : String(err);
+		return { ok: false, error: detail };
+	}
+}
+
+function addGitHubSourceChecked(input: AddGitHubSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
+	const repos = input.repos.map((r) => r.trim()).filter(Boolean);
+	if (repos.length === 0) return { ok: false, error: "At least one repo (owner/repo or owner/*) is required" };
+	for (const repo of repos) {
+		if (!/^[a-zA-Z0-9_.-]+\/[a-zA-Z0-9_*.-]+$/.test(repo)) {
+			return { ok: false, error: `Invalid repo pattern: ${repo}. Expected owner/repo or owner/*` };
+		}
+	}
+
+	if (input.maxItemsPerRepo !== undefined) {
+		if (!Number.isInteger(input.maxItemsPerRepo) || input.maxItemsPerRepo < 1 || input.maxItemsPerRepo > 10000) {
+			return { ok: false, error: "maxItemsPerRepo must be an integer between 1 and 10000" };
+		}
+	}
+
+	if (input.resourceTypes) {
+		if (!Array.isArray(input.resourceTypes)) {
+			return { ok: false, error: "resourceTypes must be an array" };
+		}
+		if (input.resourceTypes.length === 0) {
+			return { ok: false, error: "resourceTypes must include at least one resource type" };
+		}
+		const invalid = input.resourceTypes.filter((t) => !VALID_GITHUB_RESOURCE_TYPES.has(t));
+		if (invalid.length > 0) {
+			return {
+				ok: false,
+				error: `Invalid resource types: ${invalid.join(", ")}. Must be one of: ${[...DEFAULT_GITHUB_RESOURCE_TYPES].join(", ")}`,
+			};
+		}
+	}
+	if (input.tokenRef !== undefined && typeof input.tokenRef !== "string") {
+		return { ok: false, error: "tokenRef must be a string" };
+	}
+	if (input.state !== undefined && input.state !== "open" && input.state !== "closed" && input.state !== "all") {
+		return { ok: false, error: "state must be one of: open, closed, all" };
+	}
+	if (input.includeComments !== undefined && typeof input.includeComments !== "boolean") {
+		return { ok: false, error: "includeComments must be a boolean" };
+	}
+	if (input.labels !== undefined && !isStringArray(input.labels)) {
+		return { ok: false, error: "labels must be an array of strings" };
+	}
+	if (input.docPaths !== undefined) {
+		if (!isStringArray(input.docPaths)) return { ok: false, error: "docPaths must be an array of strings" };
+		const invalid = input.docPaths.filter((path) => !isSafeGitHubDocPath(path));
+		if (invalid.length > 0) {
+			return { ok: false, error: `Invalid docPaths: ${invalid.join(", ")}` };
+		}
+	}
+
+	const now = input.now ?? new Date().toISOString();
+	const cfg = loadSourcesConfigForWrite(agentsDir);
+	const agentId = cleanName(input.agentId) ?? "default";
+	const settingsKey = [...repos].sort().join(",");
+	const existing = cfg.sources.find(
+		(source) =>
+			source.kind === "github" &&
+			(source.agentId ?? "default") === agentId &&
+			Array.isArray(source.settings?.repos) &&
+			[...(source.settings.repos as string[])].sort().join(",") === settingsKey,
+	);
+
+	if (existing) {
+		const updated: SignetSourceEntry = {
+			...existing,
+			name: cleanName(input.name) ?? existing.name,
+			enabled: true,
+			updatedAt: now,
+			settings: buildGitHubSettings(input, repos),
+			agentId,
+		};
+		saveSourcesConfig(
+			{
+				version: SOURCES_CONFIG_VERSION,
+				sources: cfg.sources.map((source) => (source.id === existing.id ? updated : source)),
+			},
+			agentsDir,
+		);
+		return { ok: true, source: updated, created: false };
+	}
+
+	const source: SignetSourceEntry = {
+		id: `github:${createHash("sha256").update(`${agentId}\0${settingsKey}`).digest("hex").slice(0, 16)}`,
+		kind: "github",
+		name: cleanName(input.name) ?? repos[0],
+		root: "",
+		enabled: true,
+		mode: "read-only",
+		createdAt: now,
+		updatedAt: now,
+		settings: buildGitHubSettings(input, repos),
+		agentId,
+	};
+	saveSourcesConfig({ version: SOURCES_CONFIG_VERSION, sources: [...cfg.sources, source] }, agentsDir);
+	return { ok: true, source, created: true };
+}
+
+function buildGitHubSettings(input: AddGitHubSourceInput, repos: readonly string[]): Readonly<Record<string, unknown>> {
+	const resourceTypes = input.resourceTypes ?? [...DEFAULT_GITHUB_RESOURCE_TYPES];
+	return {
+		repos: repos,
+		tokenRef: input.tokenRef?.trim() || undefined,
+		resourceTypes,
+		state: input.state ?? "all",
+		includeComments: input.includeComments ?? true,
+		labels: input.labels ? cleanStringArray(input.labels) : undefined,
+		docPaths: input.docPaths ? cleanStringArray(input.docPaths) : [...DEFAULT_GITHUB_DOC_PATHS],
+		maxItemsPerRepo: input.maxItemsPerRepo ?? 500,
+	};
+}
+
+export function parseGitHubSettings(raw: Readonly<Record<string, unknown>> | undefined): GitHubSourceSettings {
+	if (!raw) {
+		return { repos: [], resourceTypes: [...DEFAULT_GITHUB_RESOURCE_TYPES] };
+	}
+	const repos =
+		Array.isArray(raw.repos) && raw.repos.every((r) => typeof r === "string") ? (raw.repos as string[]) : [];
+	let resourceTypes =
+		Array.isArray(raw.resourceTypes) && raw.resourceTypes.every((t) => typeof t === "string")
+			? (raw.resourceTypes as string[]).filter((t): t is "issues" | "pulls" | "discussions" | "docs" =>
+					["issues", "pulls", "discussions", "docs"].includes(t),
+				)
+			: [...DEFAULT_GITHUB_RESOURCE_TYPES];
+	if (resourceTypes.length === 0) resourceTypes = [...DEFAULT_GITHUB_RESOURCE_TYPES];
+	return {
+		repos,
+		tokenRef: typeof raw.tokenRef === "string" ? raw.tokenRef : undefined,
+		resourceTypes,
+		state: raw.state === "open" || raw.state === "closed" || raw.state === "all" ? raw.state : "all",
+		includeComments: typeof raw.includeComments === "boolean" ? raw.includeComments : true,
+		labels:
+			Array.isArray(raw.labels) && raw.labels.every((l) => typeof l === "string")
+				? (raw.labels as string[])
+				: undefined,
+		docPaths:
+			Array.isArray(raw.docPaths) && raw.docPaths.every((p) => typeof p === "string")
+				? (raw.docPaths as string[])
+				: [...DEFAULT_GITHUB_DOC_PATHS],
+		maxItemsPerRepo: typeof raw.maxItemsPerRepo === "number" && raw.maxItemsPerRepo > 0 ? raw.maxItemsPerRepo : 500,
+	};
+}
+
 export function markSourceIndexed(
 	sourceId: string,
 	indexedAt = new Date().toISOString(),
@@ -259,6 +445,21 @@ function cleanExcludeGlobs(values: readonly string[] | undefined): readonly stri
 	return cleaned.length > 0 ? cleaned : [];
 }
 
+function cleanStringArray(values: readonly string[]): readonly string[] {
+	return Array.from(new Set(values.map((value) => value.trim()).filter(Boolean)));
+}
+
+function isStringArray(value: unknown): value is readonly string[] {
+	return Array.isArray(value) && value.every((entry) => typeof entry === "string");
+}
+
+function isSafeGitHubDocPath(value: string): boolean {
+	const path = value.trim();
+	if (!path) return false;
+	if (path.startsWith("/") || path.includes("\\") || path.includes("?") || path.includes("#")) return false;
+	return !path.split("/").some((segment) => segment === "" || segment === "." || segment === "..");
+}
+
 function mergeDefaultObsidianExcludeGlobs(values: readonly string[] | undefined): readonly string[] {
 	return [...DEFAULT_OBSIDIAN_EXCLUDE_GLOBS, ...(cleanExcludeGlobs(values) ?? [])].filter(
 		(value, index, all) => all.indexOf(value) === index,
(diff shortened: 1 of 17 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the feature work in `feat(sources): add GitHub source indexing` (commit `95964bed`) and will follow up shortly.

platform/core/src/sources-config.ts
```diff
@@ -163,6 +194,161 @@ function addObsidianSourceChecked(input: AddObsidianSourceInput, agentsDir = get
...
+		const invalid = input.docPaths.filter((path) => !isSafeGitHubDocPath(path));
+		if (invalid.length > 0) {
+			return { ok: false, error: `Invalid docPaths: ${invalid.join(", ")}` };
+		}
+	}
+
+	const now = input.now ?? new Date().toISOString();
+	const cfg = loadSourcesConfigForWrite(agentsDir);
+	const agentId = cleanName(input.agentId) ?? "default";
+	const settingsKey = [...repos].sort().join(",");
```
PR-Reviewer-Ant, line 268: This default makes the new source easy to save but impossible for the daemon bridge to sync in any workspace whose resolved daemon agent id is not literally `default`. `startGitHubSourceBridge()` captures `resolveDaemonAgentId()` and then skips every source where `source.agentId !== agentId`, so a caller that omits `agentId` gets a successful add/update response and a persisted enabled GitHub source that polling will silently ignore. That undercuts the PR's agent-scoped ownership and indexing claims. The core helper should either require an agent id for GitHub sources or default through the same resolver/path used by the daemon/API caller, and the add route/CLI should have coverage for the no-agent-id path.

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,798 @@
...
+		const data = response.body as { content?: string; sha?: string; name?: string };
+		if (!data.content) continue;
+		const content = Buffer.from(data.content, "base64").toString("utf-8");
+		resources.push({
+			type: "doc",
+			path: docPath,
+			title: (data.name ?? docPath).replace(/\.[^.]+$/, ""),
+			body: content,
+			state: "open",
+			labels: [],
```
PR-Reviewer-Ant, line 748: Wildcard doc indexing bypasses the configured item bound. `maxItemsPerRepo` is threaded into issues, pulls, and discussions, but `fetchRepoDocs()` does not accept it and `fetchTreeDocs()` always indexes up to 100 files per wildcard path. That means a source configured with `maxItemsPerRepo: 1` can still fetch and embed 100 docs per glob, which diverges from the PR's bounded-fetch claim and can surprise users on large repos.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `95964bed`

</details>

I found one blocking scoping bug in the new GitHub source config path: sources added without an explicit agent id can be persisted under `default` and then skipped by the daemon bridge. I also found a smaller bounded-fetch gap for wildcard docs indexing.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The default-agent mismatch is visible in platform/core/src/sources-config.ts and platform/daemon/src/github-source-bridge.ts: config creation defaults missing agentId to `default`, while the bridge only syncs entries whose source.agentId equals resolveDaemonAgentId(). The route and CLI changed files were referenced by the PR but not included in the available checked-out prompt, so I could not verify whether every caller always passes the daemon agent id.

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,805 @@
...
+	if (!pattern.includes("*")) return [`${owner}/${pattern}`];
+	const regex = new RegExp(`^${pattern.replace(/\*/g, ".*").replace(/\?/g, ".")}$`);
+	for (const prefix of [`/orgs/${owner}/repos`, `/users/${owner}/repos`]) {
+		const repos: Array<{ full_name: string; name: string }> = [];
+		let page = 1;
+		while (true) {
+			const url = `${GITHUB_API_BASE}${prefix}?per_page=${PER_PAGE}&page=${page}&type=all`;
+			const response = await githubRequest(url, token);
+			if (response.status !== 200) break;
+			const batch = response.body as Array<{ full_name: string; name: string }>;
```
PR-Reviewer-Ant, line 257: This wildcard expansion has no hard cap on the number of repositories it will enumerate. `maxItemsPerRepo` bounds each repo after expansion, but an `owner/*` source for a large org/user can still page through every repo and then schedule issue/PR/discussion/doc fetches for all of them on every poll. That undercuts the PR's bounded-fetch claim and gives an admin-configured source a pretty large accidental DoS/rate-limit footgun. Please add a repo expansion limit, validate it in source config, and surface a partial/truncated result instead of silently expanding the whole owner.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d9a65a51`

</details>

I found one boundedness issue in the new GitHub source fetcher. The implementation mostly follows the stated shape, but wildcard repo sources can fan out without a hard cap before the per-repo limits ever apply.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The unbounded pagination is visible directly in expandRepoGlob, and the PR description explicitly claims bounded fetches. I did not run a live daemon or GitHub API repro, so the review is based on the supplied diff and changed-file excerpts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/core/src/sources-config.ts
```diff
@@ -163,6 +194,165 @@ function addObsidianSourceChecked(input: AddObsidianSourceInput, agentsDir = get
...
+	}
+	if (input.includeComments !== undefined && typeof input.includeComments !== "boolean") {
+		return { ok: false, error: "includeComments must be a boolean" };
+	}
+	if (input.labels !== undefined && !isStringArray(input.labels)) {
+		return { ok: false, error: "labels must be an array of strings" };
+	}
+	if (input.docPaths !== undefined) {
+		if (!isStringArray(input.docPaths)) return { ok: false, error: "docPaths must be an array of strings" };
+		const invalid = input.docPaths.filter((path) => !isSafeGitHubDocPath(path));
```
PR-Reviewer-Ant, line 259: This validation keeps doc paths relative, but it does not actually keep them to docs. A caller can configure `docPaths: ["src/daemon.ts"]` or another source file, and `fetchRepoDocs` will fetch and embed it as a GitHub doc. That diverges from the PR's stated docs-only/source-context boundary and from the prior no-code-indexing direction. If arbitrary file indexing is intentional, the API/docs should say that clearly; otherwise this should restrict direct `docPaths` to the same docs/markdown surface that wildcard paths use.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `78b8760e`

</details>

I found one boundary issue in the GitHub docs path support: the implementation says it indexes selected docs, but direct doc paths can point at arbitrary repo files. That makes the no-code-indexing boundary too easy to cross.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The config validation and fetch path are visible in the diff: direct docPaths are only checked for relative path safety, then fetched and embedded as docs. I could not inspect the full routes/CLI diff because the provided context truncates those files, so I am treating this as a boundary warning rather than a blocking API-contract finding.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/github-source-bridge.ts
```diff
@@ -0,0 +1,492 @@
...
+			const sources = loadSources();
+			for (const source of sources) {
+				if (!source.enabled || source.kind !== "github") continue;
+				if (source.agentId !== agentId) continue;
+				if (isSourceIndexInFlight(source.id)) continue;
+				markSourceIndexInFlight(source.id);
+				try {
+					const result = await syncGitHubSource(source, { ...options, agentId });
+					total += result.indexed;
+				} catch (err) {
```
PR-Reviewer-Ant, line 469: This marks the GitHub source as in-flight, but the success path never records completion on the source. `syncGitHubSource` returns an indexed count, then this loop only adds it to `total` and clears the in-flight bit; it does not call `markSourceIndexed` or complete/update a source-index job. That diverges from the PR’s claimed source-index job/polling behavior and means `/api/sources` can keep reporting a GitHub source with no `lastIndexedAt` or completed job even after a successful sync. It also makes future polling indistinguishable from a never-indexed source.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `43fab43e`

</details>

I found one lifecycle gap in the new GitHub bridge: it runs syncs but never records successful completion back to the source config or source-index job state. That makes the advertised async source-indexing contract incomplete and leaves the source looking perpetually unindexed.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The bridge imports and uses the in-flight guard, but the changed startGitHubSourceBridge path only marks and clears in-flight state around syncGitHubSource. I do not see a corresponding markSourceIndexed, completeSourceIndexJobFromProgress, or failure completion in the GitHub bridge code included in the diff; a daemon runtime check would confirm the exact /api/sources status shape.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `52456e4d`

</details>

I did not find a concrete correctness, security, data integrity, or race-condition issue in the provided diff. The implementation appears to cover the claimed GitHub source config validation, bounded fetching, indexing, stale reconciliation, purge hooks, CLI/API/docs surfaces, and targeted tests from the material available here.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The provided diff and changed-file snapshots include the core config parser, daemon bridge, GitHub fetcher, embedding/graph indexers, daemon startup integration, docs, and focused tests. I could not run or inspect the full local checkout because the sandbox did not contain the referenced PR files, and the PR itself notes no running-daemon validation, so runtime behavior of the API/CLI path remains unproven here.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

platform/daemon/src/github-source-bridge.ts
```diff
@@ -0,0 +1,508 @@
...
+							settings.maxItemsPerRepo,
+						)
+					: await fetchPullRequests(config, undefined, settings.state, settings.maxItemsPerRepo);
+				if (!isSourceActive()) break;
+				const capped = result.resources.length >= settings.maxItemsPerRepo;
+				for (const resource of result.resources) {
+					seenKeys.add(resourceKey(resource));
+					let comments: { author: string | null; body: string; createdAt: string }[] | undefined;
+					if (settings.includeComments && resource.commentsCount > 0) {
+						try {
```
PR-Reviewer-Ant, line 185: The GitHub source config exposes a single `state` option, and this bridge applies it to issues and PRs, but discussions ignore it. With `resourceTypes: ["discussions"]` and `state: "closed"`, this still calls `fetchDiscussions(config, undefined, settings.maxItemsPerRepo)`, and `fetchDiscussions` currently marks every discussion resource as `state: "open"`. That makes the API/CLI option misleading and can index resources the caller explicitly tried to exclude. Either thread state through discussion fetching or document/validate that `state` only applies to issues and PRs.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `bc23c567`

</details>

I found one behavior mismatch: the new GitHub source `state` option is applied to issues and PRs, but discussions ignore it and are indexed regardless of the requested state.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The bridge passes settings.state into fetchIssues and fetchPullRequests, but the discussions branch calls fetchDiscussions without any state value and the fetch implementation hard-codes discussion resources as open. I could not inspect live route behavior because the provided workspace did not contain the repo files.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,826 @@
...
+		}
+		const url = `${GITHUB_API_BASE}/repos/${config.owner}/${config.repo}/issues?${params}`;
+		const response = await githubRequest(url, config.token);
+		const rl = parseRateLimit(response.headers);
+		rateLimitRemaining = rl.remaining;
+		rateLimitReset = rl.reset;
+		if (response.status !== 200) {
+			errors.push({ message: `Issues fetch failed: ${response.status}`, retryable: response.status >= 500 });
+			break;
+		}
```
PR-Reviewer-Ant, line 323: This loop does not actually bound issue fetches by `maxItems` when the repo has many pull requests. `/repos/:owner/:repo/issues` returns issues and PRs; PR rows hit `continue` here without incrementing `fetched`, so a PR-heavy repo can keep paging well past `maxItemsPerRepo` until GitHub runs out of pages. That undercuts the PR's bounded-fetch hardening and can burn rate limit or stall source syncs on large repos. Track scanned items/pages separately, or use the search API/query that excludes PRs so the configured cap bounds network work, not just indexed issue count.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `2211bf68`

</details>

I found one bounded-fetch bug in the GitHub issues path. The implementation mostly matches the PR shape, but this path can still walk far more GitHub pages than the configured cap on PR-heavy repositories.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible directly in platform/daemon/src/github-source-fetch.ts: fetchIssues only increments its cap counter for non-PR issues, but the GitHub issues endpoint returns both issues and pull requests. I could not run a targeted runtime repro because the provided workspace contains the review prompt rather than the full checkout.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,832 @@
...
+
+export async function expandRepoGlob(
+	owner: string,
+	pattern: string,
+	token?: string,
+	maxRepos = 500,
+): Promise<RepoGlobExpansion> {
+	if (!pattern.includes("*")) return { repos: [`${owner}/${pattern}`], truncated: false };
+	const regex = new RegExp(`^${pattern.replace(/\*/g, ".*").replace(/\?/g, ".")}$`);
+	for (const prefix of [`/orgs/${owner}/repos`, `/users/${owner}/repos`]) {
```
PR-Reviewer-Ant, line 248: `maxItemsPerRepo` is being consumed by scanned `/issues` rows, not by the issue resources we actually index. GitHub's issues endpoint returns pull requests too, and this loop increments `scanned` before `if (issue.pull_request) continue`, so a repo with many recently updated PRs can hit the cap while indexing far fewer issues, or even none. That makes the new source appear successfully synced while missing issue data. The cap should limit indexed issues, with a separate bounded scan/page guard if we need to avoid unbounded paging.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `10cdde5b`

</details>

I found one correctness issue in the GitHub issue fetcher: the new indexing cap is applied to scanned `/issues` rows instead of indexed issue resources, so PR-heavy repos can silently under-index issues. The rest of the visible diff generally matches the stated feature shape and includes the expected ownership, permission, and purge hooks.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The issue is directly visible in platform/daemon/src/github-source-fetch.ts: fetchIssues increments the cap counter before filtering pull_request rows, and the added test codifies returning only one issue when maxItems is two and one row is a PR. I did not have the full route/CLI file contents in the prompt, so endpoint wiring was reviewed only from the provided diff and docs.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,845 @@
...
+				mergedAt: null,
+				commentsCount: item.comments,
+				extra: { html_url: item.html_url },
+			});
+		}
+		if (data.items.length < PER_PAGE) break;
+		if (data.total_count <= resources.length) break;
+		page++;
+	}
+	return { resources, rateLimitRemaining, rateLimitReset, errors };
```
PR-Reviewer-Ant, line 525: This parser assumes every GraphQL discussion node has `state`, but the query above does not request `state` at all. On a real response, `d.state` will be `undefined`, so `d.state.toLowerCase()` throws and the repo sync falls into the broad failure path instead of indexing discussions. Either query and map the actual GraphQL field GitHub exposes for discussion open/closed state, or derive the state from the supported field and cover it with a mock that omits `state` the way the live response currently would.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3f0f06f9`

</details>

I found a concrete correctness bug in the new discussions fetch path: the GraphQL query never requests the field that the parser treats as required, so discussion indexing can fail at runtime as soon as a repository returns discussion nodes.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new fetchDiscussions implementation in platform/daemon/src/github-source-fetch.ts includes the full GraphQL query and the parser in the diff. The query selects number, title, body, url, labels, createdAt, updatedAt, answer, and comments, but the parser immediately reads d.state.toLowerCase(); that value is not returned by the query or guaranteed by the GitHub GraphQL API response shape shown here.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/github-source-bridge.ts
```diff
@@ -0,0 +1,508 @@
...
+				if (source.agentId !== agentId) continue;
+				if (isSourceIndexInFlight(source.id)) continue;
+				const job = beginSourceIndexJob(source.id, "github-source-sync");
+				markSourceIndexInFlight(source.id);
+				if (!markSourceIndexJobRunning(source.id, job.id)) {
+					clearSourceIndexInFlight(source.id);
+					continue;
+				}
+				try {
+					const result = await syncGitHubSource(source, { ...options, agentId });
```
PR-Reviewer-Ant: This marks the source-index job complete even when `syncGitHubSource` returned `hadErrors: true`. Several fetch paths intentionally convert resource/comment/repo failures into `hadErrors` instead of throwing, so a source can fail partially, or even index zero rows after caught repo failures, while the progress API says the job completed. That undercuts the PR's partial-failure handling claim and makes operators trust stale or incomplete GitHub source state. Consider representing this as failed/partial, or at least preserving the error state in the job instead of calling the normal complete path.

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,845 @@
...
+				commentsCount: pull.comments + pull.review_comments,
+				extra: {
+					draft: pull.draft,
+					base: pull.base.ref,
+					head: pull.head.ref,
+					commits: pull.commits,
+					changed_files: pull.changed_files,
+					milestone: pull.milestone?.title ?? null,
+					assignees: pull.assignees.map((a) => a.login),
+					html_url: pull.html_url,
```
PR-Reviewer-Ant, line 455: Labels are interpolated directly into the GitHub search query as `label:"${l}"`, but `addGitHubSource` only checks that labels are strings. A label containing a quote or GitHub search syntax can change the query semantics and silently index the wrong PR set. Since this is persisted source configuration, validate label names to a safe subset or escape search-query metacharacters before building `q`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a3281e02`

</details>

I found two non-blocking correctness issues in the GitHub source path: partial sync failures are reported as completed jobs, and PR label filters are interpolated into GitHub search syntax without escaping.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The bridge and fetch behavior is directly visible in platform/daemon/src/github-source-bridge.ts and platform/daemon/src/github-source-fetch.ts. I did not have the full untruncated route and CLI file bodies, so I limited findings to issues provable from the included changed code.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/github-source-bridge.ts
```diff
@@ -0,0 +1,508 @@
...
+	agentId: string,
+	options: GitHubSourceBridgeOptions,
+): Promise<void> {
+	if (options.sourceActiveCheck && !options.sourceActiveCheck()) {
+		throw new Error(`Source ${sourceId} removed during sync`);
+	}
+	indexGitHubSourceStructure({
+		agentId,
+		sourceId,
+		sourceName: repo,
```
PR-Reviewer-Ant, line 298: This still has a delete race. `indexResource` checks `sourceActiveCheck()` once, writes graph structure, then awaits `indexGitHubSourceEmbeddings()`. That embedding path can fetch remote embeddings and sleep between chunks, so a concurrent `DELETE /api/sources/:sourceId` can purge the source while this call is in flight, after which the embedding writer can insert rows for a source that no longer exists. Since this PR explicitly claims delete hardening for in-flight syncs, the bridge needs either cancellation propagated into the embedding loop or another active-source check immediately before each write/transaction.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `d353af76`

</details>

I found one deletion race in the new GitHub source bridge. It can leave embeddings behind for a source after the source has been deleted, which undercuts the PR's stated delete hardening.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The race is visible in github-source-bridge.ts: sourceActiveCheck is evaluated before structure indexing, but the embedding path can perform async work afterward without rechecking whether the source still exists. I did not run a live daemon repro because the sandbox only includes the supplied PR diff, not a checkout.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `75a4686c`

</details>

I found one correctness issue in the daemon indexing bridge: GitHub sources are stored with per-agent ownership, but the poller only syncs sources owned by the daemon’s single resolved agent. That leaves valid non-default agent sources permanently unindexed.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The PR explicitly claims agent-scoped ownership and daemon polling. The bridge filters sources by one resolved daemon agent before syncing, which is directly visible in platform/daemon/src/github-source-bridge.ts. I did not run the daemon because the provided environment is read-only and the repository files were not present under the working directory.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/github-source-bridge.ts
```diff
@@ -0,0 +1,513 @@
...
+					continue;
+				}
+				try {
+					const result = await syncGitHubSource(source, { ...options, agentId });
+					if (result.hadErrors) {
+						failSourceIndexJob(source.id, job.id, "GitHub source sync completed with partial errors");
+					} else {
+						markSourceIndexed(source.id, undefined, options.agentsDir);
+						completeSourceIndexJob(source.id, job.id, result.indexed);
+					}
```
PR-Reviewer-Ant, line 467: This makes the GitHub poller skip every configured GitHub source whose `agentId` is not the daemon’s single resolved agent. `addGitHubSource` now deliberately stores sources per agent, and the PR description says agent-scoped ownership is supported, but the bridge loads all GitHub sources and then drops non-matching agents here. A source added for `agent-b` can remain enabled in `sources.json` and still never be indexed, never marked failed, and never reconciled. The sync should process each source under its own `source.agentId` instead of filtering to one daemon-wide agent id, or the add/API path needs to reject non-daemon-agent ownership.

platform/daemon/src/github-source-fetch.ts
```diff
@@ -0,0 +1,849 @@
...
+				body: body ? JSON.stringify(body) : undefined,
+				signal: controller.signal,
+			});
+			clearTimeout(timeout);
+			const rateLimit = parseRateLimit(response.headers);
+			if (rateLimit.remaining < 10 && rateLimit.reset > Date.now()) {
+				const waitMs = rateLimit.reset - Date.now() + 1000;
+				logger.warn("github-source", "Approaching rate limit, backing off", {
+					remaining: rateLimit.remaining,
+					waitMs,
```
PR-Reviewer-Ant, line 187: This builds a regex from the user-provided repo glob without escaping regex metacharacters other than `*` and `?`. GitHub repo names can contain `.`, and the config validator allows it, so a source like `owner/private.*` becomes `^private..*$` and will match repos such as `privateXarchive` too. Since this decides which repos get fetched and indexed, the glob needs to escape literal repo-name characters before expanding `*`.

platform/daemon/src/github-source-bridge.ts
```diff
@@ -0,0 +1,512 @@
...
+						}
+					}
+					await indexResource(source.id, repo.fullName, resource, comments, agentId, syncOpts);
+					repoIndexed++;
+					await yielder();
+				}
+				logErrors(source.id, repo.fullName, "issues", result.resources.length, result.errors);
+				if (result.errors.length > 0) hadErrors = true;
+				if (!capped && result.errors.length === 0 && !commentFetchFailed) completeTypes.add("issues");
+			}
```
PR-Reviewer-Ant, line 166: When `discussions` is requested but no token is configured, the sync only logs a warning and then continues as a successful sync. Because discussions are in the default resource types, adding a GitHub source without a token will mark the source indexed even though one requested/default resource class was never indexed. That undermines the PR's partial-failure handling claim and makes `lastIndexedAt` misleading; this should either be a validation error when discussions are enabled without `tokenRef`, or set `hadErrors` so the source job does not complete as clean.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `ba69326c`

</details>

I found two correctness issues in the new GitHub source indexing path: wildcard repo matching can include repos outside the configured pattern, and requested discussion indexing is reported as successful even when it is skipped for lack of a token.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The wildcard regex and discussion skip behavior are directly visible in the added GitHub source files. The checkout available here only included the review prompt, so I could not inspect the full route implementation for additional validation behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/core/src/sources-config.ts
```diff
@@ -163,6 +194,165 @@ function addObsidianSourceChecked(input: AddObsidianSourceInput, agentsDir = get
...
+		updatedAt: now,
+		settings: buildGitHubSettings(input, repos),
+		agentId,
+	};
+	saveSourcesConfig({ version: SOURCES_CONFIG_VERSION, sources: [...cfg.sources, source] }, agentsDir);
+	return { ok: true, source, created: true };
+}
+
+function buildGitHubSettings(input: AddGitHubSourceInput, repos: readonly string[]): Readonly<Record<string, unknown>> {
+	const resourceTypes = input.resourceTypes ?? [...DEFAULT_GITHUB_RESOURCE_TYPES];
```
PR-Reviewer-Ant: Defaulting `resourceTypes` to all four types includes `discussions`, but the daemon treats requested discussions without a token as a sync error. That means `signet sources add github owner/repo` or the API equivalent without `tokenRef` will still index public issues/PRs/docs, but the source job is marked failed and `lastIndexedAt` is never updated on every poll. Either default unauthenticated sources to REST-fetchable types only, require a token when discussions are included, or treat missing-token discussions as an explicit skipped capability rather than a failed sync.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `9e20bea7`

</details>

I found one correctness issue in the GitHub source defaults: the default config includes discussions, but discussion indexing is treated as an error whenever no token is configured. That makes the advertised no-token/default GitHub source path report a failed sync every time.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The default resourceTypes are set from DEFAULT_GITHUB_RESOURCE_TYPES in platform/core/src/sources-config.ts, which includes discussions, and syncGitHubSource marks missing-token discussions as hadErrors. I did not need a live GitHub repro because the failure follows directly from the changed control flow.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/core/src/sources-config.ts
```diff
@@ -163,6 +195,170 @@ function addObsidianSourceChecked(input: AddObsidianSourceInput, agentsDir = get
...
+			Array.isArray(source.settings?.repos) &&
+			[...(source.settings.repos as string[])].sort().join(",") === settingsKey,
+	);
+
+	if (existing) {
+		const updated: SignetSourceEntry = {
+			...existing,
+			name: cleanName(input.name) ?? existing.name,
+			enabled: true,
+			updatedAt: now,
```
PR-Reviewer-Ant, line 283: This update path replaces the whole GitHub settings object from the incoming input. Because most fields are optional, a normal partial update such as renaming an existing source with the same repo set will clear `tokenRef`, reset `resourceTypes` to the no-token default, drop custom `labels`/`docPaths`, and reset `maxItemsPerRepo`. That directly breaks the documented "Add or update" behavior and can stop discussions/private data from syncing after a harmless-looking update. Preserve existing settings for fields the caller did not provide, then overlay explicitly provided values.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6e4a2932`

</details>

I found one correctness issue in the GitHub source update path: updating an existing source can silently drop its token and custom indexing settings.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The core config update path is visible in platform/core/src/sources-config.ts and the API docs explicitly describe POST /api/sources/github as add-or-update. I could not inspect the full route/CLI implementations from the provided context, but any caller that omits optional fields while updating the same repo set will hit this core behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `935fdc1d`

</details>

I did not find a concrete blocker in the provided diff. The rebuilt GitHub source path has targeted validation, agent-scoped source identity, in-flight job handling, partial-error reporting, and focused tests around the prior risky areas.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The visible diff and changed-file excerpts cover the core source config, daemon bridge, fetch, embedding, docs, and daemon startup behavior, and those are enough to check the main claims around validation, polling, token resolution shape, and deletion race guards. The prompt truncates parts of github-source-fetch.ts and does not include full sources-routes.ts, github-source-graph.ts, or CLI command contents, so I cannot independently verify every API/CLI ownership path or graph write detail from this review context alone.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

aaf2tbz:
Ready for human review.

NicholaiVogel:
Closing this in favor of a provider-substrate rebuild.

This PR proved out the GitHub source behavior and captured useful hardening around bounded fetches, token handling, docs-path limits, partial failures, update preservation, and deletion races. Since it was opened, #757 landed the provider-neutral source substrate, and #759 showed the intended shape for non-Obsidian providers through the shared source job path.

Rather than merge a conflicted branch with a parallel GitHub-specific bridge/poller, I’m going to reimplement this on current `main` as a `github` source provider. The new PR should reuse the fetcher/tests from here where they still fit, but route config, indexing, purge, progress, and deletion through the shared source provider/job infrastructure.

Thanks for the review loop here; the findings from this PR become the regression checklist for the rebuild.

PR #728, opened 2026-05-17
feat(sources): add Discord source with live API indexing + knowledge graph

aaf2tbz:
## Summary
- Adds a `"discord"` source kind that indexes Discord guilds, channels, and threads into Signet's recall system via live Discord REST API v10.
- Follows the Obsidian source pattern: direct embedding into `embeddings` table + knowledge graph construction — not the connector/document pipeline pattern.
- Bot token resolved via Signet Secrets, server/channel/thread hierarchy in knowledge graph, conversation chunking by time proximity.

## Validation
- `bun test platform/core/src/sources-config.test.ts platform/daemon/src/discord-source-embeddings.test.ts platform/daemon/src/discord-source-graph.test.ts`
- `bunx biome check platform/daemon/src/discord-source-fetch.ts platform/daemon/src/discord-source-embeddings.ts platform/daemon/src/discord-source-graph.ts platform/daemon/src/discord-source-bridge.ts platform/core/src/sources-config.ts surfaces/cli/src/commands/sources.ts surfaces/cli/src/features/sources.ts platform/daemon/src/routes/sources-routes.ts platform/daemon/src/daemon.ts`
- 16 tests pass (8 config, 4 embeddings, 3 graph, 1 pre-existing)
- All 3 code quality bot findings resolved in `7ec5e51`

## Notes
- No schema or migration changes — Discord source entries are stored in the existing `sources.json` config with a `settings` field.
- No `discord.js` dependency — raw `fetch()` against Discord REST v10.
- `root` field left empty for Discord sources (no filesystem root, like GitHub sources).
- Existing `discord-parser.ts` left untouched — this is a live API path, not static DiscordChatExporter parsing.

## PR Readiness (MANDATORY)
- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

## Migration Notes (if applicable)
- [x] Migration is idempotent
- [x] Daemon Rust parity reviewed or explicitly N/A
- [x] Rollback / compatibility note included in PR description

Rollback / compatibility: no migration or persisted data change; rollback is removing the source config entry and reverting the commit.

PR #750, opened 2026-05-23
feat(sources): add Discord source indexing

aaf2tbz:
## Summary

Rebuilds #728 as a clean Discord source implementation on current `main`. Adds Discord guild/channel/thread indexing through the live Discord REST API v10, source config/CLI/API support, direct embeddings, knowledge graph structure, and docs.

## Changes

- Adds `discord` source config support in `@signet/core`, including guild ID validation, token references, channel filters, message bounds, thread inclusion, and ISO `since` normalization.
- Adds daemon Discord REST fetching, message chunking/embedding, graph indexing, startup sync, purge support, and `/api/sources/discord`.
- Adds `signet sources add discord` CLI support.
- Fixes #728 reviewer findings: guild-level active-thread route, real `since` filtering, and stable user-ID participant graph keys.
- Documents Discord Sources in `docs/SOURCES.md` and `docs/API.md`.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: docs

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Rollback / compatibility: no database migration or persisted schema change. Discord source entries are stored in existing `sources.json`; rollback is removing the Discord source config entry and reverting these commits. Rust daemon parity is N/A because this extends the TypeScript daemon source indexing path only.

## Testing

- [ ] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Validated with:

- `bunx biome check platform/core/src/sources-config.ts platform/core/src/sources-config.test.ts platform/daemon/src/discord-source-fetch.ts platform/daemon/src/discord-source-fetch.test.ts platform/daemon/src/discord-source-bridge.ts platform/daemon/src/discord-source-embeddings.ts platform/daemon/src/discord-source-embeddings.test.ts platform/daemon/src/discord-source-graph.ts platform/daemon/src/discord-source-graph.test.ts platform/daemon/src/routes/sources-routes.ts platform/daemon/src/routes/sources-routes.test.ts surfaces/cli/src/commands/sources.ts surfaces/cli/src/features/sources.ts`
- `cd platform/core && bun test src/sources-config.test.ts`
- `cd platform/daemon && bun test src/discord-source-fetch.test.ts src/discord-source-embeddings.test.ts src/discord-source-graph.test.ts src/routes/sources-routes.test.ts`
- `cd platform/daemon && bun run build.ts`

Known local validation limits:

- `cd platform/daemon && bun run build:types` still fails on existing broad repo `rootDir`/SQLite typing issues outside this Discord source change.
- `cd surfaces/cli && bun run build:cli` requires connector workspace packages to be built/linked first and failed resolving those connector workspaces in this fresh worktree.

## AI disclosure

## Notes

Replaces #728. Old PR #728 is closed in favor of this clean replacement.

### DIFF
--- a/platform/core/src/sources-config.ts
+++ b/platform/core/src/sources-config.ts
@@ -52,6 +52,10 @@ export interface DiscordSourceSettings {
 export const DEFAULT_DISCORD_MAX_MESSAGES = 1000;
 export const MAX_DISCORD_MESSAGES_PER_CHANNEL = 10_000;
 
+function looksLikeRawDiscordToken(value: string): boolean {
+	return /^[A-Za-z0-9_-]{20,}\.[A-Za-z0-9_-]{6,}\.[A-Za-z0-9_-]{20,}$/.test(value);
+}
+
 export interface AddDiscordSourceInput {
 	readonly guildIds: readonly string[];
 	readonly tokenRef: string;
@@ -148,6 +152,9 @@ function addDiscordSourceChecked(input: AddDiscordSourceInput, agentsDir = getAg
 	}
 	const trimmedTokenRef = input.tokenRef?.trim();
 	if (!trimmedTokenRef) return { ok: false, error: "Discord bot token reference (tokenRef) is required" };
+	if (looksLikeRawDiscordToken(trimmedTokenRef)) {
+		return { ok: false, error: "Discord tokenRef must be a secret reference, not a raw bot token" };
+	}
 	const guildIds = cleanDiscordIds(input.guildIds);
 	if (guildIds.length === 0) return { ok: false, error: "At least one Discord guild ID is required" };
 	for (const id of guildIds) {
--- a/platform/core/src/sources-config.ts
+++ b/platform/core/src/sources-config.ts
@@ -3,7 +3,7 @@ import { existsSync, mkdirSync, readFileSync, renameSync, rmSync, statSync, writ
 import { homedir } from "node:os";
 import { dirname, resolve } from "node:path";
 
-export type SignetSourceKind = "obsidian";
+export type SignetSourceKind = "obsidian" | "discord";
 export type SignetSourceMode = "read-only";
 
 export interface SignetSourceEntry {
@@ -17,6 +17,7 @@ export interface SignetSourceEntry {
 	readonly updatedAt: string;
 	readonly lastIndexedAt?: string;
 	readonly excludeGlobs?: readonly string[];
+	readonly settings?: Readonly<Record<string, unknown>>;
 }
 
 export const DEFAULT_OBSIDIAN_EXCLUDE_GLOBS = [
@@ -36,6 +37,29 @@ export interface AddObsidianSourceInput {
 	readonly root: string;
 	readonly name?: string;
 	readonly excludeGlobs?: readonly string[];
+	readonly now?: string;
+}
+
+export interface DiscordSourceSettings {
+	readonly guildIds: readonly string[];
+	readonly tokenRef: string;
+	readonly channelFilter?: readonly string[];
+	readonly maxMessagesPerChannel: number;
+	readonly includeThreads?: boolean;
+	readonly since?: string;
+}
+
+export const DEFAULT_DISCORD_MAX_MESSAGES = 1000;
+export const MAX_DISCORD_MESSAGES_PER_CHANNEL = 10_000;
+
+export interface AddDiscordSourceInput {
+	readonly guildIds: readonly string[];
+	readonly tokenRef: string;
+	readonly name?: string;
+	readonly channelFilter?: readonly string[];
+	readonly maxMessagesPerChannel?: number;
+	readonly includeThreads?: boolean;
+	readonly since?: string;
 	readonly now?: string;
 }
 
@@ -103,6 +127,118 @@ function loadSourcesConfigForWrite(agentsDir = getAgentsDir()): SignetSourcesCon
 
 export function addObsidianSource(input: AddObsidianSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
 	return withSourcesConfigLock(agentsDir, () => addObsidianSourceUnlocked(input, agentsDir));
+}
+
+export function addDiscordSource(input: AddDiscordSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
+	return withSourcesConfigLock(agentsDir, () => addDiscordSourceUnlocked(input, agentsDir));
+}
+
+function addDiscordSourceUnlocked(input: AddDiscordSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
+	try {
+		return addDiscordSourceChecked(input, agentsDir);
+	} catch (err) {
+		const detail = err instanceof Error ? err.message : String(err);
+		return { ok: false, error: detail };
+	}
+}
+
+function addDiscordSourceChecked(input: AddDiscordSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
+	if (!input.guildIds || input.guildIds.length === 0) {
+		return { ok: false, error: "At least one Discord guild ID is required" };
+	}
+	const trimmedTokenRef = input.tokenRef?.trim();
+	if (!trimmedTokenRef) return { ok: false, error: "Discord bot token reference (tokenRef) is required" };
+	const guildIds = cleanDiscordIds(input.guildIds);
+	if (guildIds.length === 0) return { ok: false, error: "At least one Discord guild ID is required" };
+	for (const id of guildIds) {
+		if (!/^\d{17,20}$/.test(id)) return { ok: false, error: `Invalid Discord guild ID: ${id}` };
+	}
+	const maxMessages = cleanDiscordMaxMessages(input.maxMessagesPerChannel);
+	if (maxMessages === null) {
+		return {
+			ok: false,
+			error: `Discord maxMessagesPerChannel must be an integer between 1 and ${MAX_DISCORD_MESSAGES_PER_CHANNEL}`,
+		};
+	}
+	const since = cleanDiscordSince(input.since);
+	if (since === null) {
+		return { ok: false, error: "Discord since must be a valid ISO date" };
+	}
+
+	const key = guildIds.slice().sort().join(",");
+	const now = input.now ?? new Date().toISOString();
+	const cfg = loadSourcesConfigForWrite(agentsDir);
+	const sourceId = `discord:${createHash("sha256").update(key).digest("hex").slice(0, 16)}`;
+	const existing = cfg.sources.find((source) => source.id === sourceId);
+
+	if (existing) {
+		const updated = {
+			...existing,
+			name: cleanName(input.name) ?? existing.name,
+			enabled: true,
+			settings: buildDiscordSettings(input),
+			updatedAt: now,
+		};
+		saveSourcesConfig(
+			{
+				version: SOURCES_CONFIG_VERSION,
+				sources: cfg.sources.map((source) => (source.id === existing.id ? updated : source)),
+			},
+			agentsDir,
+		);
+		return { ok: true, source: updated, created: false };
+	}
+
+	const source: SignetSourceEntry = {
+		id: sourceId,
+		kind: "discord",
+		name: cleanName(input.name) ?? "Discord Server",
+		root: "",
+		enabled: true,
+		mode: "read-only",
+		createdAt: now,
+		updatedAt: now,
+		settings: buildDiscordSettings(input),
+	};
+	saveSourcesConfig({ version: SOURCES_CONFIG_VERSION, sources: [...cfg.sources, source] }, agentsDir);
+	return { ok: true, source, created: true };
+}
+
+function buildDiscordSettings(input: AddDiscordSourceInput): Record<string, unknown> {
+	const maxMessages = cleanDiscordMaxMessages(input.maxMessagesPerChannel) ?? DEFAULT_DISCORD_MAX_MESSAGES;
+	const since = cleanDiscordSince(input.since);
+	return {
+		guildIds: cleanDiscordIds(input.guildIds),
+		tokenRef: input.tokenRef.trim(),
+		...(input.channelFilter && input.channelFilter.length > 0
+			? { channelFilter: cleanDiscordChannelFilter(input.channelFilter) }
+			: {}),
+		maxMessagesPerChannel: maxMessages,
+		includeThreads: input.includeThreads ?? true,
+		...(since ? { since } : {}),
+	};
+}
+
+export function parseDiscordSettings(raw?: Readonly<Record<string, unknown>>): DiscordSourceSettings {
+	if (!raw) {
+		return { guildIds: [], tokenRef: "", maxMessagesPerChannel: DEFAULT_DISCORD_MAX_MESSAGES, includeThreads: true };
+	}
+	const guildIds = Array.isArray(raw.guildIds)
+		? raw.guildIds
+				.filter((id): id is string => typeof id === "string")
+				.map((id) => id.trim())
+				.filter(Boolean)
+		: [];
+	const tokenRef = typeof raw.tokenRef === "string" ? raw.tokenRef : "";
+	const channelFilter = Array.isArray(raw.channelFilter)
+		? cleanDiscordChannelFilter(raw.channelFilter.filter((id): id is string => typeof id === "string"))
+		: undefined;
+	const maxMessagesPerChannel =
+		cleanDiscordMaxMessages(typeof raw.maxMessagesPerChannel === "number" ? raw.maxMessagesPerChannel : undefined) ??
+		DEFAULT_DISCORD_MAX_MESSAGES;
+	const includeThreads = typeof raw.includeThreads === "boolean" ? raw.includeThreads : true;
+	const since = cleanDiscordSince(typeof raw.since === "string" ? raw.since : undefined) ?? undefined;
+	return { guildIds, tokenRef, channelFilter, maxMessagesPerChannel, includeThreads, since };
 }
 
 function addObsidianSourceUnlocked(input: AddObsidianSourceInput, agentsDir = getAgentsDir()): AddSourceResult {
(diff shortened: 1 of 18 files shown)


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the feature work in `feat(sources): add Discord source indexing` (commit `a61abdd9`) and will follow up shortly.

platform/daemon/src/daemon.ts
```diff
@@ -1610,6 +1611,30 @@ async function main() {
...
+				fetchEmbedding,
+				agentsDir: AGENTS_DIR,
+			})
+				.then((indexed) => {
+					logger.info("discord-source", "Discord source startup sync complete", {
+						sourceId: source.id,
+						indexed,
+					});
+					markSourceIndexed(source.id, undefined, AGENTS_DIR);
+				})
```
PR-Reviewer-Ant, line 1643: This marks the Discord source as indexed even when `syncDiscordSource` skipped all work. In `syncDiscordSource`, a missing/invalid secret returns `0` instead of throwing, and inaccessible guilds also just continue; the startup caller then writes `lastIndexedAt` anyway. That makes a broken Discord source look successfully indexed in `/api/sources` and can hide a configuration or permission failure until someone inspects logs. Please only mark indexed after a real successful sync, or have the bridge return a structured skipped/failed status.

platform/core/src/sources-config.ts
```diff
@@ -105,6 +129,118 @@ export function addObsidianSource(input: AddObsidianSourceInput, agentsDir = get
...
+			agentsDir,
+		);
+		return { ok: true, source: updated, created: false };
+	}
+
+	const source: SignetSourceEntry = {
+		id: sourceId,
+		kind: "discord",
+		name: cleanName(input.name) ?? "Discord Server",
+		root: "",
```
PR-Reviewer-Ant, line 203: This stores `tokenRef` exactly as supplied, but the API/CLI/docs are relying on it being a Signet secret reference rather than the raw Discord bot token. Nothing here rejects an obvious token-shaped value, so a pasted bot token will be persisted into `sources.json`, which violates the repo's secret-handling convention and is hard to undo once synced or backed up. Please validate this boundary enough to fail closed for raw-looking Discord tokens, or resolve/require an existing secret key instead of saving arbitrary secret material.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a61abdd9`

</details>

I found two issues worth fixing: the daemon records a Discord source as indexed even when sync was skipped, and the token reference field can persist an obvious raw Discord token into source config.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the changed code: syncDiscordSource returns 0 on missing tokens, while daemon.ts marks the source indexed for any resolved sync promise; sources-config.ts only trims tokenRef before saving it into sources.json.

platform/daemon/src/daemon.ts
```diff
@@ -1610,6 +1611,31 @@ async function main() {
 				});
 		}
 
+		for (const source of loadSourcesConfig(AGENTS_DIR).sources) {
+			if (!source.enabled || source.kind !== "discord") continue;
```
PR-Reviewer-Ant, line 1617: This starts Discord indexing as an untracked fire-and-forget task. Unlike Obsidian startup indexing, it never calls `markSourceIndexInFlight`, creates a source index job, or participates in the deletion tombstone flow. If a user removes the Discord source while this sync is fetching channels/messages, `DELETE /api/sources/:sourceId` can purge the current rows and clear the tombstone, then this still-running task can continue writing graph rows and embeddings for a source that no longer exists in `sources.json`. That breaks the stated remove-and-purge lifecycle for sources. The Discord sync needs to be tracked/cancellable or at least checked against deletion tombstones/config before each write and before clearing deletion state.

platform/daemon/src/discord-source-graph.ts
```diff
@@ -0,0 +1,350 @@
...
+}
+
+function idFor(...parts: readonly string[]): string {
+	return `dcsrc_${createHash("sha256").update(parts.join("\0")).digest("hex").slice(0, 28)}`;
+}
+
+function upsertEntity(
+	db: WriteDb,
+	input: {
+		readonly id: string;
```
PR-Reviewer-Ant, line 87: `upsertEntity` looks up existing entities by `canonical_name` and `agent_id` only, then overwrites `source_id`, `source_kind`, `source_path`, and `source_root`. That is risky for participant entities because their canonical name is just `discord:user:${participant.id}`. If the same Discord user appears in two configured Discord sources for the same agent, indexing the second source reassigns the shared participant entity to the second source. A later purge of either source can then leave dependencies pointing at a deleted participant entity or fail to purge participant data that originated from the removed source. Source-owned rows need a source-scoped identity, or purge/linking needs a separate shared-user model that does not mutate ownership between sources.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b7fade31`

</details>

I found two data-lifecycle issues in the Discord source indexing path. The new startup sync can outlive source removal, and source-owned participant entities can be reassigned across Discord sources in a way that makes purge behavior unreliable.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The issues are visible directly in the changed daemon startup sync, delete/purge route, and Discord graph upsert code. I did not need a live Discord repro to see that the sync is not tracked as in-flight and that entity upserts key by canonical_name plus agent_id while later purges delete by source_id.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/discord-source-graph.ts
```diff
@@ -0,0 +1,350 @@
...
+		input.confidence,
+		input.reason,
+		input.now,
+		input.now,
+		input.sourceId,
+		DISCORD_SOURCE_KIND,
+		"",
+		input.sourceId,
+	);
+	return true;
```
PR-Reviewer-Ant, line 233: This sync path only upserts the currently fetched Discord structure; it never removes source-owned graph rows for channels, threads, conversations, or participants that are no longer returned. After a channel is deleted, filtered out, loses bot visibility, or a participant disappears from the fetched message window, the old `entities` and `entity_dependencies` rows remain recallable until the entire source is removed. That breaks the source lifecycle claim that derived rows can be refreshed from the live source and purged by source state. The embedding path has per-prefix stale cleanup; the graph path needs a comparable per-source/per-guild/per-channel cleanup strategy or a full source graph refresh before reindexing.

platform/daemon/src/discord-source-fetch.ts
```diff
@@ -0,0 +1,362 @@
...
+				logger.warn("discord-source", "Approaching rate limit, backing off", {
+					remaining: rateLimit.remaining,
+					waitMs,
+				});
+				await new Promise((resolve) => setTimeout(resolve, Math.min(waitMs, 30_000)));
+			}
+			if (response.status >= 500) {
+				lastError = new Error(`Discord API ${response.status}: ${await response.text()}`);
+				await new Promise((resolve) => setTimeout(resolve, RETRY_BASE_DELAY_MS * (attempt + 1)));
+				continue;
```
PR-Reviewer-Ant, line 158: `discordRequest` unconditionally calls `response.json()` for every non-204 response. Discord error responses and proxy/network edge cases can be empty or plain text, which turns a normal 401/403/404/5xx response into a JSON parse exception before callers can handle the status. For example, `fetchGuild()` intends to return `null` on 403/404, but it will never get that far if the body is empty or not JSON. This should parse defensively, falling back to text/null while preserving the HTTP status.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `56b4571a`

</details>

I found a couple of correctness issues in the Discord indexing path: one can leave stale graph rows behind after channels or participants disappear, and one can fail ordinary Discord API calls when the response body is empty or non-JSON.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The stale graph behavior follows directly from discord-source-graph.ts only upserting current entities/dependencies and only purging on full source removal; there is no per-sync deletion for channels, threads, conversations, or participants that disappear. The Discord response parsing issue is visible in discord-source-fetch.ts where every non-204 response is parsed with response.json(), including error responses. I did not have a live Discord runtime trace, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/discord-source-graph.ts
```diff
@@ -0,0 +1,493 @@
...
+			     SELECT 1 FROM entity_dependencies d
+			     WHERE d.agent_id = entities.agent_id
+			       AND (d.source_entity_id = entities.id OR d.target_entity_id = entities.id)
+			   )`,
+		)
+		.run(agentId, sourceId, participantPrefix, `${participantPrefix}\uffff`).changes;
+}
+
+function upsertEntity(
+	db: WriteDb,
```
PR-Reviewer-Ant, line 85: This lookup can cross-contaminate separate Discord sources for the same agent. `upsertEntity` reuses any row with the same `canonical_name` and `agent_id`, but guild/channel/conversation canonical names are based on Discord IDs, not `sourceId` (`discord:${guildId}`, `discord:${guildId}:${channelId}`, etc.). Since `addDiscordSource` keys sources by the sorted full guild set, users can configure overlapping sources like `[guildA]` and `[guildA, guildB]`; syncing the second source will update the first source's guild/channel rows to `source_id = sourceB`. A later purge or reconcile for sourceB can then delete graph rows still needed by sourceA. Please include `source_id` in this lookup/uniqueness model or make all Discord source-owned canonical names source-scoped consistently.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b1a9d633`

</details>

I found a data-integrity problem in the Discord graph upsert path: overlapping Discord source configs can make one source take ownership of another source's graph rows. That should be fixed before merge because purge/reconcile can then delete or mutate rows that still belong to another configured source.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed graph code explicitly looks up existing entities by canonical_name and agent_id only, while several Discord canonical names are not source-scoped. The PR also allows distinct Discord source IDs for different guild sets, so overlapping guild configs are reachable from the added source config path.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/discord-source-fetch.ts
```diff
@@ -0,0 +1,373 @@
...
+): Promise<DiscordFetchResult<DiscordMessage>> {
+	const messages: DiscordMessage[] = [];
+	const errors: { message: string; retryable: boolean }[] = [];
+	let rateLimitRemaining = 5;
+	let rateLimitReset = 0;
+	let fetched = 0;
+	let cursor = beforeId;
+
+	while (fetched < maxMessages) {
+		const params = new URLSearchParams({
```
PR-Reviewer-Ant, line 232: This `BigInt(msg.id)` conversion is outside the request/error handling boundary. If Discord ever returns an unexpected id shape, a partial response, or a mocked/non-snowflake id in a test/dev server, this throws and escapes `fetchChannelMessages` entirely. In `syncDiscordSource` that gets caught as a channel sync failure, so one bad message can drop the whole channel instead of returning the messages already fetched with a structured error. Please validate ids before conversion or compare only after a safe parse so malformed rows are skipped/reported without aborting the channel.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `01c3db69`

</details>

I found one correctness issue in the Discord message fetch path: a malformed or unexpected message id can abort the whole channel sync instead of being reported as a fetch error. The rest of the implementation broadly matches the PR description from the provided diff.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is directly visible in platform/daemon/src/discord-source-fetch.ts where BigInt(msg.id) is called inside the message loop without validation or containment. I did not have a live Discord/runtime repro, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `349f6f6f`

</details>

I found one lifecycle bug in the new Discord startup sync: it is not tracked as an in-flight source index job, so removal can race with indexing and leave supposedly purged Discord rows behind.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff shows Discord sync is launched directly from daemon startup without source-index in-flight tracking, while DELETE only keeps a tombstone when isSourceIndexInFlight(source.id) is true. I did not run a live daemon reproduction, but the race follows from the changed daemon startup path and existing deletion logic in sources-routes.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/daemon.ts
```diff
@@ -1610,6 +1611,35 @@ async function main() {
...
+			const startupEmbedCfg = loadMemoryConfig(AGENTS_DIR).embedding;
+			syncDiscordSource(source, {
```
PR-Reviewer-Ant, line 1625: This startup Discord sync is not registered with the existing source index in-flight/tombstone machinery. If a user removes the Discord source while this promise is fetching or embedding, `DELETE /api/sources/:sourceId` purges once and then immediately clears the tombstone because `isSourceIndexInFlight(result.source.id)` is false. The still-running sync can then write graph rows or embeddings after the purge, leaving rows for a source the user just removed. The Obsidian path already has deletion tombstones because it goes through source index jobs; Discord needs equivalent in-flight tracking or a source-active check inside the embedding/graph write loops before each write.

platform/daemon/src/routes/sources-routes.ts
```diff
@@ -125,6 +137,37 @@ export function registerSourcesRoutes(app: Hono, deps: RegisterSourcesRoutesDeps
...
+				guildIds,
+				tokenRef,
+				name: body.name,
+				channelFilter: Array.isArray(body.channelFilter)
+					? body.channelFilter.filter((id): id is string => typeof id === "string")
+					: undefined,
+				maxMessagesPerChannel: typeof body.maxMessagesPerChannel === "number" ? body.maxMessagesPerChannel : undefined,
+				includeThreads: typeof body.includeThreads === "boolean" ? body.includeThreads : undefined,
+				since: typeof body.since === "string" ? body.since : undefined,
+			},
```
PR-Reviewer-Ant, line 167: This endpoint records the Discord source but never schedules or runs a sync. Since the new startup sync loop only runs when the daemon starts, `signet sources add discord` can return success while producing no chunks, embeddings, or graph rows until the user restarts the daemon. That does not match the PR's API/CLI source indexing claim and is a pretty sharp user-facing failure compared with the Obsidian add path, which queues indexing immediately.

platform/daemon/src/discord-source-bridge.ts
```diff
@@ -0,0 +1,327 @@
...
+		const channelFilter = settings.channelFilter ? new Set(settings.channelFilter) : null;
+		const filteredChannels = channelFilter
+			? textChannels.filter((ch) => channelFilter.has(ch.id) || (ch.name && channelFilter.has(ch.name)))
+			: textChannels;
+		const activeThreads = settings.includeThreads ? await fetchGuildActiveThreads(config, guildId) : [];
+		const currentChannelIds = filteredChannels.map((channel) => channel.id);
+		const reconciledChannels: Array<{ channelId: string; conversationPaths: string[] }> = [];
+
+		const yielder = yieldEvery(5);
+
```
PR-Reviewer-Ant, line 125: Stale Discord chunks are not purged when a previously indexed channel now returns zero messages. The stale embedding cleanup is inside `indexDiscordSourceEmbeddings`, but this branch only calls it when `msgResult.data.length > 0`; the same pattern appears for threads below. If a `since` setting is tightened, messages are deleted, or a channel/thread becomes empty, the graph document can be reconciled away while old `source_discord_chunk` embeddings remain searchable.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `996dcbd7`

</details>

I found two lifecycle bugs in the Discord source path: adding a source does not actually kick off indexing, and refreshes can leave stale Discord chunks searchable after the current fetch returns no messages.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The add route returns queued:false after only writing the source config, and the only sync wiring in daemon.ts runs during startup. The stale-chunk issue follows directly from discord-source-bridge.ts only calling indexDiscordSourceEmbeddings when msgResult.data.length > 0, while stale cleanup lives inside that embedding function.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/discord-source-bridge.ts
```diff
@@ -0,0 +1,335 @@
...
+					guildId,
+					channelId: channel.id,
+					channelName,
+					error: err instanceof Error ? err.message : String(err),
+				});
+			}
+		}
+		if (isSourceActive()) {
+			reconcileDiscordGuildStructure({
+				agentId,
```
PR-Reviewer-Ant, line 202: This reconciliation runs even when `fetchGuildChannels()` failed and returned `data: []` with errors. In that case `currentChannelIds` is empty and `reconciledChannels` is empty, so `reconcileDiscordGuildStructure()` will treat a transient Discord/API/permission failure as an authoritative empty guild and delete previously indexed channel, conversation, and participant graph rows. The archived-thread path has the same shape: `fetchPublicArchivedThreads()` errors are ignored, so previously indexed archived thread conversations can be dropped from `conversationPaths` and purged. Please only reconcile after a successful authoritative listing, or preserve existing rows when fetches are partial/failed.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `0edad3fe`

</details>

I found one data-integrity bug in the Discord sync path: transient Discord fetch failures can be interpreted as an authoritative empty state and purge previously indexed graph rows.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The new Discord bridge logs fetch errors but still feeds empty/partial results into reconciliation. The provided diff includes the relevant sync and reconcile code, but the full routes file and existing source-index conventions were truncated in the prompt.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Thanks for the work on this. This PR did several things well: it proved the Discord source shape in Signet, added useful config validation and token-reference guardrails, exercised Discord REST API v10 fetching, wired CLI/API entry points, and surfaced the right lifecycle problems around source removal, stale rows, and partial fetches.

After comparing this with Discrawl and the existing Obsidian Sources architecture, we are going to close this PR and replace it with a cleaner stacked implementation. The main issue is not that the code is unusable; it is that the abstraction boundary is too provider-specific. We do not want Discord to grow a second source pipeline beside Obsidian. The replacement work will create a unified Sources substrate first, then move Obsidian and Discord onto that substrate.

The replacement direction is:

- provider adapters handle provider-specific discovery and fetching only;
- shared source infrastructure owns index jobs, cancellation, tombstones, purge, chunks, embeddings, stats, recall, and graph ingestion;
- Discord will target Discrawl-level completeness, including multiple guilds, full channel/thread coverage, members, mentions, attachments metadata, sync cursors, and safe partial-failure handling;
- semantic entities, aspects, attributes, claims, and dependencies will be derived through the shared Signet source graph path with provenance back to the Discord source.

We will reuse the good pieces from this PR where they fit: token-ref validation, REST fetch helpers/tests, CLI/API ergonomics, and lifecycle regression cases. The Discord-specific bridge/embedding/graph split will not be the foundation for the replacement.

PR #765, opened 2026-05-24
feat(sources): add source snapshot operations

NicholaiVogel:
## Summary
- Adds generic Signet source snapshot export/import for source-owned `memory_artifacts`.
- Adds daemon API routes and CLI commands for `signet sources snapshot export|import`.
- Defaults to excluding Discord Desktop cache `@me` artifacts, while preserving local `@me` rows on import unless explicitly included.
- Documents the snapshot contract in Sources and API reference docs.

## Test Plan
- [x] `bun test platform/daemon/src/routes/sources-routes.test.ts surfaces/cli/src/features/sources.test.ts surfaces/cli/src/commands/sources.test.ts`
- [x] `bun test platform/core/src/sources-config.test.ts platform/daemon/src/discord-source-provider.test.ts platform/daemon/src/routes/sources-routes.test.ts surfaces/cli/src/commands/sources.test.ts surfaces/cli/src/features/sources.test.ts`
- [x] `bun run --filter '@signet/core' build`
- [x] `bun run --filter '@signet/cli' build`
- [x] `bun run --filter '@signet/daemon' build`
- [x] `git diff --check`
- [ ] `bun scripts/doc-drift.ts --markdown` currently reports pre-existing migration/package inventory drift unrelated to this PR.

## PR Readiness (MANDATORY)

- [x] No database migrations changed

### DIFF
--- a/platform/daemon/src/routes/sources-routes.ts
+++ b/platform/daemon/src/routes/sources-routes.ts
@@ -38,6 +38,7 @@ import {
 	updateSourceIndexJobProgress,
 } from "../source-index-progress";
 import { getSourceProvider } from "../source-providers";
+import { exportSourceSnapshot, importSourceSnapshot } from "../source-snapshots";
 
 interface SourceIndexJobInput {
 	readonly source: SignetSourceEntry;
@@ -202,6 +203,48 @@ export function registerSourcesRoutes(app: Hono, deps: RegisterSourcesRoutesDeps
 		return c.json({ source: result.source, created: result.created, indexed: 0, queued: true, job }, 202);
 	});
 
+	app.get("/api/sources/:sourceId/snapshot", (c) => {
+		const sourceId = c.req.param("sourceId");
+		const source = findConfiguredSource(sourceId, agentsDir);
+		if (!source) return c.json({ error: "Source not found" }, 404);
+		const includeLocalDiscord = c.req.query("includeLocalDiscord") === "true";
+		return c.json(
+			exportSourceSnapshot({
+				source,
+				agentId: resolveDaemonAgentId(),
+				includeLocalDiscord,
+			}),
+		);
+	});
+
+	app.post("/api/sources/:sourceId/snapshot/import", async (c) => {
+		const sourceId = c.req.param("sourceId");
+		const source = findConfiguredSource(sourceId, agentsDir);
+		if (!source) return c.json({ error: "Source not found" }, 404);
+		if (isSourceImportBlocked(source.id)) {
+			return c.json({ error: "Source snapshot import cannot run while source indexing is queued or running" }, 409);
+		}
+		markSourceIndexInFlight(source.id);
+		try {
+			let body: unknown;
+			try {
+				body = await c.req.json();
+			} catch {
+				return c.json({ error: "Invalid JSON body" }, 400);
+			}
+			const result = importSourceSnapshot({
+				source,
+				agentId: resolveDaemonAgentId(),
+				snapshot: body,
+				includeLocalDiscord: c.req.query("includeLocalDiscord") === "true",
+			});
+			if (result.ok === false) return c.json({ error: result.error }, 400);
+			return c.json(result);
+		} finally {
+			clearSourceIndexInFlight(source.id);
+		}
+	});
+
 	app.delete("/api/sources/:sourceId", (c) => {
 		const sourceId = c.req.param("sourceId");
 		const result = removeSource(sourceId, agentsDir);
@@ -216,6 +259,15 @@ export function registerSourcesRoutes(app: Hono, deps: RegisterSourcesRoutesDeps
 			clearSourceDeletionTombstone(result.source.id, sourceAgentId, agentsDir);
 		return c.json({ source: result.source, purged });
 	});
+}
+
+function findConfiguredSource(sourceId: string, agentsDir: string): SignetSourceEntry | undefined {
+	return loadSourcesConfig(agentsDir).sources.find((source) => source.id === sourceId);
+}
+
+function isSourceImportBlocked(sourceId: string): boolean {
+	const job = getSourceIndexJob(sourceId);
+	return isSourceIndexInFlight(sourceId) || job?.status === "queued" || job?.status === "running";
 }
 
 function enqueueSourceIndexJob(input: SourceIndexJobInput): SourceIndexJob {
--- a/platform/daemon/src/source-snapshots.ts
+++ b/platform/daemon/src/source-snapshots.ts
@@ -1,0 +1,454 @@
+import type { Database } from "bun:sqlite";
+import { SOURCE_CHUNK_SOURCE_TYPE, type SignetSourceEntry } from "@signet/core";
+import { getDbAccessor } from "./db-accessor";
+import type { WriteDb } from "./db-accessor";
+import { syncVecDeleteByEmbeddingIds } from "./db-helpers";
+import { hashNormalizedBody } from "./memory-lineage";
+
+export const SOURCE_SNAPSHOT_VERSION = 1;
+
+export interface SourceSnapshotArtifact {
+	readonly sourcePath: string;
+	readonly sourceSha256: string;
+	readonly sourceKind: string;
+	readonly sessionId: string;
+	readonly sessionKey: string | null;
+	readonly sessionToken: string;
+	readonly project: string | null;
+	readonly harness: string | null;
+	readonly capturedAt: string;
+	readonly startedAt: string | null;
+	readonly endedAt: string | null;
+	readonly manifestPath: string | null;
+	readonly sourceNodeId: string | null;
+	readonly memorySentence: string | null;
+	readonly memorySentenceQuality: string | null;
+	readonly content: string;
+	readonly updatedAt: string;
+	readonly sourceMtimeMs: number | null;
+	readonly sourceId: string;
+	readonly sourceRoot: string | null;
+	readonly sourceExternalId: string | null;
+	readonly sourceParentPath: string | null;
+	readonly sourceMetaJson: string | null;
+}
+
+export interface SourceSnapshot {
+	readonly version: typeof SOURCE_SNAPSHOT_VERSION;
+	readonly exportedAt: string;
+	readonly source: {
+		readonly id: string;
+		readonly kind: string;
+		readonly name: string;
+		readonly root: string;
+	};
+	readonly agentId: string;
+	readonly artifacts: readonly SourceSnapshotArtifact[];
+	readonly skipped: {
+		readonly localDiscordArtifacts: number;
+	};
+}
+
+export interface ExportSourceSnapshotOptions {
+	readonly source: SignetSourceEntry;
+	readonly agentId: string;
+	readonly includeLocalDiscord?: boolean;
+}
+
+export interface ImportSourceSnapshotOptions {
+	readonly source: SignetSourceEntry;
+	readonly agentId: string;
+	readonly snapshot: unknown;
+	readonly includeLocalDiscord?: boolean;
+}
+
+export type ImportSourceSnapshotResult =
+	| {
+			readonly ok: true;
+			readonly imported: number;
+			readonly skipped: { readonly localDiscordArtifacts: number };
+	  }
+	| { readonly ok: false; readonly error: string };
+
+interface ArtifactRow {
+	readonly source_path: string;
+	readonly source_sha256: string;
+	readonly source_kind: string;
+	readonly session_id: string;
+	readonly session_key: string | null;
+	readonly session_token: string;
+	readonly project: string | null;
+	readonly harness: string | null;
+	readonly captured_at: string;
+	readonly started_at: string | null;
+	readonly ended_at: string | null;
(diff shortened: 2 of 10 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `8b818718`.

platform/daemon/src/source-snapshots.ts
```diff
@@ -0,0 +1,395 @@
...
+		const stmt = (db as WriteDb as Database).prepare(
+			`INSERT INTO memory_artifacts (
+				agent_id, source_path, source_sha256, source_kind, session_id,
+				session_key, session_token, project, harness, captured_at,
+				started_at, ended_at, manifest_path, source_node_id,
+				memory_sentence, memory_sentence_quality, content, updated_at,
+				source_mtime_ms, source_id, source_root, source_external_id,
+				source_parent_path, source_meta_json
+			) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+			ON CONFLICT(agent_id, source_path) DO UPDATE SET
```
PR-Reviewer-Ant, line 190: This import can overwrite another source's artifact if the snapshot contains a `sourcePath` that already exists for the same agent. The validation checks `artifact.sourceId === options.source.id`, but `sourceId` is attacker-controlled snapshot data, and the conflict target is `(agent_id, source_path)`. On conflict, this branch rewrites `source_id`, `source_root`, content, and provenance to the imported source. That means a malformed or crafted snapshot for `discord:A` can clobber an unrelated `obsidian` or Discord artifact at the same `source_path`. Before upserting, reject artifacts whose path/root do not belong to the configured source, or make the conflict/update conditional on the existing row already having `source_id = options.source.id`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `0f3ebe60`

</details>

I found one data-integrity hole in the import path: a snapshot can upsert over artifacts from another source because ownership is trusted from the snapshot payload instead of being checked against the existing row/source path.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The issue is visible in the new import code: snapshot artifacts are only checked for matching sourceId and checksum, then inserted with ON CONFLICT(agent_id, source_path) DO UPDATE. I could not inspect the live schema or surrounding artifact helpers because the workspace only includes the review prompt, but the changed SQL and validation are enough to show the overwrite path.

platform/daemon/src/source-snapshots.ts
NicholaiVogel: Fixed in 6d0e31a81.

Snapshot import now preflights every incoming artifact path against the existing `(agent_id, source_path)` owner before deleting or upserting. If a row already exists for another `source_id`, import fails with a 400 and leaves the existing row untouched.

The upsert also has a defensive `WHERE memory_artifacts.source_id = excluded.source_id` guard so an unexpected conflict cannot reassign source ownership. Added a route regression test covering the cross-source overwrite attempt.

platform/daemon/src/source-snapshots.ts
```diff
@@ -0,0 +1,419 @@
...
+
+function findPathOwnershipConflict(
+	db: Database,
+	artifacts: readonly SourceSnapshotArtifact[],
+	agentId: string,
+	sourceId: string,
+): string | null {
+	const stmt = db.prepare("SELECT source_id FROM memory_artifacts WHERE agent_id = ? AND source_path = ? LIMIT 1");
+	for (const artifact of artifacts) {
+		const row = stmt.get(agentId, artifact.sourcePath) as { source_id: string | null } | undefined;
```
PR-Reviewer-Ant, line 261: This import deletes/replaces only `memory_artifacts` rows for the source. Existing source chunk embeddings are left intact, so after importing a snapshot that removes or changes artifacts, recall can still return stale chunks from the pre-import source contents. That also conflicts with the PR/docs claim that import preserves indexing consistency. Source removal already treats source-owned embeddings as purgeable state, and `sourceStats` counts embeddings by the `${source.id}:` prefix, so snapshot import needs to either purge/rebuild those rows or explicitly prevent stale indexed chunks from surviving the replacement.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6d0e31a8`

</details>

I found one data-integrity issue in the new import path: it replaces source artifact rows directly but does not keep the source-owned embedding rows in sync. That leaves stale recall chunks behind after importing a snapshot.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The changed import implementation explicitly deletes and reinserts only rows in memory_artifacts, while the existing source stats and source removal contract treat embeddings with source_id prefixes as part of source-owned indexed state. I could not validate the full purge/upsert helper implementation from the provided diff, but the stale embedding path is directly visible from source-snapshots.ts and sources-routes.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c0d58a77`

</details>

I found one blocking scope issue: the latest commit pulled release/version artifacts into a source-snapshot PR. The snapshot-specific data-integrity concerns from the prior review look addressed by the new ownership check and chunk purge coverage.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The incremental diff shows .bump-level, CHANGELOG.md, bun.lock, dist/signetai/package.json, and many workspace package.json versions changed to 0.126.0, while the PR description is about source snapshot export/import. The repository conventions explicitly call out generated/publish artifacts as high-risk scope drift, and these files are outside the snapshot implementation path.


- CHANGELOG.md:40 - This release ledger/version bump does not belong in this source-snapshot PR. The incremental diff also changes `.bump-level`, `bun.lock`, `dist/signetai/package.json`, and many package versions to `0.126.0`, apparently for the earlier Discord desktop cache release rather than the snapshot feature. Please drop the release/generated version artifacts from this PR or move them to the release PR; carrying them here makes the publish surface ambiguous and risks shipping unrelated package metadata.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Rechecked the rebased PR diff after the scope warning on `c0d58a77`.

`gh pr diff 765 --name-only` and `git diff --name-only origin/main...HEAD` both show only these 10 snapshot-related files:

- `docs/SOURCES.md`
- `docs/api/documents-sources.md`
- `docs/api/route-inventory.md`
- `platform/daemon/src/routes/sources-routes.test.ts`
- `platform/daemon/src/routes/sources-routes.ts`
- `platform/daemon/src/source-snapshots.ts`
- `surfaces/cli/src/commands/sources.test.ts`
- `surfaces/cli/src/commands/sources.ts`
- `surfaces/cli/src/features/sources.test.ts`
- `surfaces/cli/src/features/sources.ts`

No `.bump-level`, `CHANGELOG.md`, `bun.lock`, `dist/signetai/package.json`, or package version files are part of this PR diff. The release/version artifacts were from the base branch movement during the rebase comparison, not from this PR.

platform/daemon/src/source-snapshots.ts
NicholaiVogel: Fixed in c0d58a77.

Snapshot import now purges source-owned generic `source_chunk` embeddings for the imported source inside the same transaction that replaces `memory_artifacts`, and it calls `syncVecDeleteByEmbeddingIds` before deleting regular embedding rows so the vec mirror stays in sync.

The default local Discord cache behavior is preserved: when `includeLocalDiscord` is false, import keeps existing `@me` artifacts and skips local `@me` chunks while purging stale public/source chunks. Added route regressions for both stale chunk removal and local DM chunk preservation.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c0d58a77`

</details>

I found one race in the snapshot import path: it can run concurrently with an active source indexing job, so the claimed replacement can end up mixed with live-sync writes.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The new import route calls importSourceSnapshot directly, and importSourceSnapshot deletes/reinserts source-owned rows without checking or canceling the existing source-index in-flight state that this same routes module tracks for add/delete flows. I did not run a concurrent runtime repro, but the missing coordination is visible in the changed route and import transaction.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/sources-routes.ts
```diff
@@ -202,6 +203,40 @@ export function registerSourcesRoutes(app: Hono, deps: RegisterSourcesRoutesDeps
...
+		);
+	});
+
+	app.post("/api/sources/:sourceId/snapshot/import", async (c) => {
+		const sourceId = c.req.param("sourceId");
+		const source = findConfiguredSource(sourceId, agentsDir);
+		if (!source) return c.json({ error: "Source not found" }, 404);
+		let body: unknown;
+		try {
+			body = await c.req.json();
```
PR-Reviewer-Ant: Snapshot import needs to coordinate with the source indexing job lifecycle before it starts deleting and reinserting rows. This route can currently run while a sync job for the same source is in flight; that job can continue writing `memory_artifacts` and source chunks while `importSourceSnapshot` deletes the import scope and purges embeddings. The result is neither the snapshot nor the live source state, which breaks the advertised "replaces source-owned artifact rows" behavior. The delete path already has cancellation/tombstone handling for this class of race; import needs a similar guard, cancellation, or explicit 409 while `isSourceIndexInFlight(source.id)` is true.

platform/daemon/src/routes/sources-routes.ts
```diff
@@ -202,6 +203,43 @@ export function registerSourcesRoutes(app: Hono, deps: RegisterSourcesRoutesDeps
...
+				source,
+				agentId: resolveDaemonAgentId(),
+				includeLocalDiscord,
+			}),
+		);
+	});
+
+	app.post("/api/sources/:sourceId/snapshot/import", async (c) => {
+		const sourceId = c.req.param("sourceId");
+		const source = findConfiguredSource(sourceId, agentsDir);
```
PR-Reviewer-Ant, line 222: This guard only checks whether indexing is active at the instant the request enters the handler. The handler then awaits JSON parsing and later deletes/reinserts source rows, but it never marks the source as import-busy or otherwise reserves the source against a newly queued index job. That means a source sync can still start after this check and race the import, leaving the snapshot replacement mixed with live indexing writes. The import path should either acquire the same per-source in-flight gate for the full import or make source indexing check an import lock before it starts.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `770b149e`

</details>

I found one remaining race in the import route. The new guard rejects imports when indexing is already active, but it does not reserve the source while the import request is being parsed and applied.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The route checks isSourceImportBlocked before awaiting c.req.json(), then calls importSourceSnapshot without marking the source busy. Source indexing can be queued or start after that preflight check, so the delete/reinsert transaction can still overlap live source writes. I did not run a concurrent repro, but the remaining gap is visible in sources-routes.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `8b818718`

</details>

I did not find a new correctness, security, data-integrity, or race issue in this diff. The previously raised import coordination, ownership-conflict, and stale chunk concerns appear addressed by the in-flight reservation, conflict check, chunk purge, and targeted route tests.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed route now marks the source in-flight before parsing the import body and clears it in a finally block, with a regression test covering the parsing window. The import path validates source id/kind, checksum, cross-source path ownership, local Discord exclusion, and purges matching source chunks. I did not run the listed Bun tests or exercise the daemon/CLI end to end in a live runtime.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

platform/daemon/src/routes/sources-routes.ts
NicholaiVogel: Fixed in 8b818718a.

Snapshot import now coordinates with the same per-source in-flight gate used by index jobs. The route rejects import with 409 when a source index job is queued/running or already in flight, so live source sync cannot overlap the delete/reinsert snapshot replacement path.

platform/daemon/src/routes/sources-routes.ts
NicholaiVogel: Fixed in 8b818718a.

The import handler now marks `sourceIndexInFlight` before awaiting `c.req.json()` and keeps that reservation through `importSourceSnapshot`, releasing it in `finally`. Index jobs that are queued after the import begins see the source as busy and wait/reschedule instead of writing during import.

Added a streaming-body regression test that pauses JSON parsing, verifies a concurrent import gets 409 while the first import holds the reservation, then completes the first import successfully.

PR #777, opened 2026-05-25
feat(sources): tail Discord gateway events

NicholaiVogel:
## Summary

- enables the existing `gateway-tail` Discord source mode with a live Discord Gateway v10 loop
- indexes gateway-observed message create/update/delete events, deleted-message tombstones, channel/thread upserts, member upserts/removals, and per-channel tail checkpoints through the shared source artifact path
- exposes Gateway tail in the dashboard Discord source mode selector and updates Sources/API docs

## Discrawl parity notes

This mirrors Discrawl's tail worker event coverage for message create/update/delete, channel upsert, member upsert, and member delete while preserving Signet's shared Sources substrate. The gateway transport handles hello/identify/heartbeat, Discord reconnect/invalid-session requests, fatal gateway close codes, cancellation/removal, and test-injected sockets.

## Validation

- `bunx biome check --write platform/daemon/src/discord-gateway-tail.ts platform/daemon/src/discord-source-provider.ts platform/daemon/src/discord-source-provider.test.ts surfaces/dashboard/src/lib/components/tabs/SourcesTab.svelte`
- `bun test platform/daemon/src/discord-source-provider.test.ts`
- `bun test platform/core/src/sources-config.test.ts`
- `bun test surfaces/cli/src/features/sources.test.ts surfaces/cli/src/commands/sources.test.ts`
- `bun test platform/daemon/src/routes/sources-routes.test.ts`
- `bun run --filter '@signet/core' build`
- `bun run --filter '@signet/daemon' build`
- `git diff --check`

`@signet/daemon` build still emits the existing unrelated Svelte accessibility/deprecation warnings in SecretsTab, TroubleshooterPanel, SidebarGroups, WidgetGrid, and WidgetCard.

PR #778, opened 2026-05-25
fix(sources): patch partial Discord gateway updates

NicholaiVogel:
## Summary

- patches existing `source_discord_message` artifacts when Discord Gateway sends partial `MESSAGE_UPDATE` dispatches without full author/timestamp fields
- preserves the lifecycle event artifact while refreshing the indexed message content, edited timestamp, and update-present metadata fields
- tightens the gateway-tail provider test to use a partial update payload and assert the stored message artifact is refreshed

## Validation

- `bunx biome check --write platform/daemon/src/discord-source-provider.ts platform/daemon/src/discord-source-provider.test.ts`
- `bun test platform/daemon/src/discord-source-provider.test.ts`
- `git diff --check`

`@signet/daemon` build still emits the existing unrelated Svelte accessibility/deprecation warnings in SecretsTab, TroubleshooterPanel, SidebarGroups, WidgetGrid, and WidgetCard.

### DIFF
--- a/platform/daemon/src/discord-source-provider.ts
+++ b/platform/daemon/src/discord-source-provider.ts
@@ -318,6 +318,10 @@ async function syncDiscordGatewayTailSource(
 						indexed += writeArtifact(context.source, agentId, embedArtifact(guild, channel, message, embed, index));
 				}
 				if (message.poll) indexed += writeArtifact(context.source, agentId, pollArtifact(guild, channel, message));
+			} else if (gatewayEventType === "MESSAGE_UPDATE") {
+				const patched = patchedMessageArtifact(context.source, agentId, guildId, channelId, messageId, payload);
+				if (patched) indexed += writeArtifact(context.source, agentId, patched);
+				purgeClearedPartialUpdateArtifacts(context.source, agentId, guildId, channelId, messageId, payload);
 			}
 			indexed += writeArtifact(
 				context.source,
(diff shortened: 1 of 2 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm taking a look at the fixes in `fix(sources): patch partial Discord gateway updates` (commit `d39b837b`) and will follow up shortly.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d39b837b`

</details>

I found one data-integrity issue in the partial update patching path. The implementation handles simple message edits, but multiline Discord messages can keep stale old content when patched.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The bug is directly in the new patchMessageContent implementation: it uses lastIndexOf("") over the whole stored markdown artifact to find the header/body separator, but Discord message bodies may themselves contain blank lines. The added test only covers a single-line message body, so this case is not guarded.

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +781,88 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+	const lines = previous.split("\n");
+	if (patch.editedAt) {
+		const editedIndex = lines.findIndex((line) => line.startsWith("Edited: "));
+		if (editedIndex >= 0) lines[editedIndex] = `Edited: ${patch.editedAt}`;
+		else {
+			const createdIndex = lines.findIndex((line) => line.startsWith("Created: "));
+			lines.splice(createdIndex >= 0 ? createdIndex + 1 : Math.min(lines.length, 6), 0, `Edited: ${patch.editedAt}`);
+		}
+	}
+	if (patch.content === null) return lines.join("\n");
```
PR-Reviewer-Ant, line 936: This separator search can leave stale message text behind for multiline Discord messages. `previous` is the whole markdown artifact, and the message body is appended after a blank line, but Discord content itself can contain blank lines. In that case `lines.lastIndexOf("")` may point inside the old body rather than the header/body boundary, so an edit like `old line\n\nold private detail` patched to `new line` would produce a stored artifact containing both `old line` and `new line`. That defeats the PR's core claim that partial `MESSAGE_UPDATE` refreshes the indexed message content. Please anchor the replacement to the known metadata boundary instead of the last blank line, and add a regression test with an original multiline message containing a blank line.

NicholaiVogel:
Addressed the multiline data-integrity issue. Partial `MESSAGE_UPDATE` patching now finds the Discord message body by known artifact header lines instead of blank-line position, and the gateway-tail regression now starts with a multiline body, applies a partial multiline edit, and asserts the old body text is gone.

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +781,109 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+): string {
+	const lines = previous.split("\n");
+	if (patch.editedAt) {
+		const editedIndex = lines.findIndex((line) => line.startsWith("Edited: "));
+		if (editedIndex >= 0) lines[editedIndex] = `Edited: ${patch.editedAt}`;
+		else {
+			const createdIndex = lines.findIndex((line) => line.startsWith("Created: "));
+			lines.splice(createdIndex >= 0 ? createdIndex + 1 : Math.min(lines.length, 6), 0, `Edited: ${patch.editedAt}`);
+		}
+	}
```
PR-Reviewer-Ant, line 922: This still has a data corruption edge for edited messages whose body starts with text that looks like one of our artifact headers. `messageBodyStartIndex` keeps skipping blank lines and any `isMessageHeaderLine` match until it finds a non-header line, but after the real header separator a Discord user can legitimately write `Guild: ...`, `Channel: ...`, `Edited: ...`, `Pinned: true`, etc. If the old body starts that way, this patch appends the new content after the old body instead of replacing it, so stale message text remains indexed. The safer boundary is the known artifact structure from `messageArtifact`: find the blank separator after the `Created`/optional metadata block, then treat everything after that as body regardless of its text.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a43e4e23`

</details>

The partial update path now covers the straightforward multiline edit case from the PR description, but the markdown patcher still treats user-authored body lines that look like artifact headers as replaceable header text. That can leave stale Discord message content in the stored artifact after an edit.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The issue is visible in the new patching logic: messageBodyStartIndex skips any line matching isMessageHeaderLine before deciding where the body starts, but those same strings are legal Discord message content once the original body begins. I did not run the test suite in this review context.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/discord-source-provider.ts
PR-Reviewer-Ant, line 922: This still has a data-integrity edge for edited messages whose body starts with text that looks like one of our artifact headers. `messageBodyStartIndex` skips blank lines and every `isMessageHeaderLine` match until the first non-header line, so a legitimate message body like `Guild: release notes\nactual body` or `Edited: not metadata\n...` would have those leading lines discarded when a partial `MESSAGE_UPDATE` arrives. The multiline blank-line case is fixed, but we still need an unambiguous body boundary or a stored body offset/structured content source instead of classifying body lines by prefix.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a873bc7c`

</details>

I found one remaining data-integrity edge in the partial update patching logic. The main partial multiline update path is covered, but header-looking message bodies can still be truncated on edit.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The issue is visible in the changed patchMessageContent/messageBodyStartIndex logic: it derives the message body by skipping blank lines and any line matching known artifact headers. I did not run a runtime repro here, but the failure mode follows directly for stored message bodies whose first edited content line begins with one of those header prefixes.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the header-looking body line case too. The patcher now consumes only the actual generated artifact headers using stored message metadata (`editedAt`, reply, pinned) instead of skipping any line with a header-looking prefix. The regression now starts the Discord body with `Guild:` and confirms the old header-looking body text is removed after a partial multiline edit.

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +781,112 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+		mtimeMs: Date.parse(editedAt ?? "") || Date.now(),
+		content,
+		meta: {
+			...meta,
+			provider: DISCORD_PROVIDER_KIND,
+			recordType: "message",
+			guildId,
+			channelId,
+			messageId,
+			editedAt: editedAt ?? null,
```
PR-Reviewer-Ant, line 834: This preserves stale poll metadata when an update explicitly carries `poll` but it is not a record. Because `meta` is spread first, a previously indexed poll message keeps `pollPresent: true` unless the update has an object-shaped poll; an explicit `poll: null` would leave the indexed message saying a poll is still present. The rest of this patch treats present update fields as authoritative, so this one should either clear to `false` when `"poll" in payload` and the value is not a record, or document/test why Discord never sends poll removal this way.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `01617cd3`

</details>

I found one data-integrity edge in the partial update patching: `poll` removal is not reflected when Discord sends the field as present but null. The earlier multiline/header-looking body concern appears addressed by the metadata-guided header consumption and regression coverage.

**Confidence:** Medium [sufficient_diff_evidence, ambiguous_requirements] - The changed metadata merge explicitly refreshes present array fields, booleans, and numeric fields, but only sets `pollPresent` when `payload.poll` is an object. If Discord sends `poll: null` or otherwise includes a non-record poll value to clear/remove poll state, the old `pollPresent: true` value survives through the preceding `...meta` spread. I did not run a runtime repro, and the exact Discord gateway shape for poll removal is the remaining uncertainty.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the poll-clearing edge. Partial gateway updates now treat a present `poll` field as authoritative for `pollPresent`, including `poll: null` clearing stale poll metadata, and the gateway-tail regression covers that case.

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +781,116 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+				? { mentionUserIds: payload.mentions.map(discordPayloadId).filter(isNonNullString) }
+				: {}),
+			...(Array.isArray(payload.mention_roles) ? { mentionRoleIds: payload.mention_roles.filter(isString) } : {}),
+			...(Array.isArray(payload.embeds) ? { embedCount: payload.embeds.length } : {}),
+			...(hasOwn(payload, "poll") ? { pollPresent: isRecord(payload.poll) } : {}),
+			...(Array.isArray(payload.reactions) ? { reactions: payload.reactions.map(discordPayloadReaction) } : {}),
+		},
+	};
+}
+
```
PR-Reviewer-Ant, line 850: This body-start parser does not account for the blank separator after `# Discord Message: ...` in `messageArtifact`, so a partial update with `content` computes `bodyStart` as `1`. The rewrite then keeps only the title and discards the stored `Guild`, `Channel`, `Author`, `Created`, `Edited`, reply, and pinned headers. Because Discord partial `MESSAGE_UPDATE` events often omit author/timestamp fields, this corrupts exactly the metadata this patch is meant to preserve. The regression test misses it because the updated body itself starts with `Guild:`; it should also assert that the original artifact headers, especially `Author:` and `Created:`, survive the patch.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `421f9993`

</details>

I found a data-integrity regression in the partial update patcher: it refreshes the message body, but drops the existing artifact headers that the partial gateway payload does not contain. That directly undermines the PR's stated goal of patching partial updates without losing author/timestamp context.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The existing messageArtifact format inserts a blank line after '# Discord Message: ...', but messageBodyStartIndex tries to consume 'Guild:' immediately at index 1. For any content-bearing partial MESSAGE_UPDATE, bodyStart remains 1, so patchMessageContent keeps only the title and replaces the rest with the new body.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the header-preservation regression. The partial update patcher now consumes the generated blank separators around the Discord artifact header block before replacing the body, so partial content edits preserve guild/channel/author/created context. The gateway-tail regression now asserts those headers survive alongside the edited multiline body and `poll: null` metadata clearing.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f783e14a`

</details>

The partial message patch now preserves the message header/body correctly, but it leaves derived Discord child artifacts behind when an update explicitly clears them. That can make source recall disagree with the refreshed message metadata.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff shows partial MESSAGE_UPDATE payloads patch only the source_discord_message row. The new test covers pollPresent:false in source_meta_json, but the same fixture first creates a source_discord_poll artifact and the implementation does not purge or rewrite that artifact when poll:null is received.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +781,122 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+			recordType: "message",
+			guildId,
+			channelId,
+			messageId,
+			editedAt: editedAt ?? null,
+			...(typeof payload.pinned === "boolean" ? { pinned: payload.pinned } : {}),
+			...(typeof payload.flags === "number" ? { flags: payload.flags } : {}),
+			...(Array.isArray(payload.attachments)
+				? { attachmentIds: payload.attachments.map(discordPayloadId).filter(isNonNullString) }
+				: {}),
```
PR-Reviewer-Ant, line 839: When a partial update explicitly clears a child-bearing field, this patches the message metadata but leaves the old child artifact indexed. In the new test flow, MESSAGE_CREATE writes a `source_discord_poll`, then MESSAGE_UPDATE sends `poll: null`; this line correctly sets `pollPresent:false`, but nothing removes `discord://.../message/<id>/poll`. The same shape applies to `attachments: []`, `embeds: []`, and `mentions: []`: the message row says the field is gone while source-owned child rows can still be recalled as current. If partial update payload fields are treated as authoritative metadata, the corresponding child artifact paths need to be refreshed or purged when those fields are present.

PR-Reviewer-Ant, line 839: When a partial `MESSAGE_UPDATE` explicitly clears a child-bearing field, this only patches the parent message metadata. For example, the new test sends `poll: null` after `MESSAGE_CREATE` wrote a `source_discord_poll`; `pollPresent` becomes false, but the old poll artifact at `/message/<id>/poll` remains indexed. That leaves source-backed recall/graph data saying both “no poll” and “old poll exists.” The same pattern applies to attachments/embeds if Discord sends authoritative empty arrays.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cdbb5580`

</details>

I found one release-metadata problem that should be fixed before this lands, plus a remaining data-consistency edge in the Discord patch path. The partial message content/header behavior itself now matches the PR description.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included] - The Discord patching logic and gateway-tail regression are included, and the incremental diff shows release/version metadata added outside the PR's stated bug-fix scope. I did not run the test suite or inspect the release tooling output, so the stale child-artifact issue is based on the indexed artifact paths visible in this file.


- .bump-level:1 - This PR is described as a patch fix for partial Discord gateway updates, but the incremental diff changes the release bump to `minor` and also carries the 0.134.0 changelog/package-version updates for `tail Discord gateway events (#777)`. That looks like release/generated metadata from a different change set leaking into this bug-fix PR; if merged as-is, it can publish the wrong version scope and unrelated release ledger entries from this PR.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

NicholaiVogel:
Addressed the stale child artifact issue. When a partial gateway update explicitly clears poll data with `poll: null`, the source message metadata is patched to `pollPresent:false` and the old `source_discord_poll` artifact plus its source graph rows are purged. The gateway-tail regression now asserts both the artifact row and graph document are gone.

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +782,149 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+	if (!row) return null;
+	const meta = parseMetaJson(row.source_meta_json);
+	const editedAt = readPayloadString(payload, "edited_timestamp") ?? readMetaString(meta, "editedAt");
+	const content = patchMessageContent(row.content, {
+		content: readPayloadString(payload, "content"),
+		editedAt,
+		meta,
+	});
+	return {
+		kind: "source_discord_message",
```
PR-Reviewer-Ant, line 821: This patches the parent message metadata when `attachments`, `mentions`, `mention_roles`, or `embeds` are present on a partial `MESSAGE_UPDATE`, but only `poll` gets matching child-artifact cleanup. If Discord sends `attachments: []` after a message attachment is removed, the parent row will now say `attachmentIds: []` while the old `source_discord_attachment` artifact and graph rows remain indexed under the same message. The same stale-child pattern applies to mentions and embeds when those arrays are present and shrink. Either purge/rewrite the affected child artifacts for present child-bearing fields, or avoid claiming the parent metadata is authoritative for those fields.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `fd91ea2a`

</details>

The partial message patch now preserves the parent message row and handles the poll-clearing case covered by the latest regression. I still see one data-integrity hole in the same new path: present-but-empty child-bearing update fields can make the parent metadata say children are gone while leaving stale child artifacts indexed.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new partial-update code explicitly patches attachmentIds, mentionUserIds, mentionRoleIds, and embedCount from present gateway payload fields, but the only child artifact cleanup added is for poll rows. That mismatch is visible in platform/daemon/src/discord-source-provider.ts and does not require runtime reproduction to prove the stale-row path.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Also checked the release-metadata note after the rebase: the current PR diff against `origin/main` only contains `platform/daemon/src/discord-source-provider.ts` and `platform/daemon/src/discord-source-provider.test.ts`. There are no `.bump-level`, changelog, or package-version files in this branch; that finding was from a stale rebase-boundary comparison.

NicholaiVogel:
Broadened the child-artifact cleanup for explicit empty partial update fields. `MESSAGE_UPDATE` payloads with empty `attachments`, `embeds`, `mentions`/`mention_roles`, or `poll:null` now purge the matching message child artifacts and source graph rows under the same message parent path. The gateway-tail regression now clears all of those fields and asserts no stale child artifacts or child graph rows remain.

platform/daemon/src/discord-source-provider.ts
```diff
@@ -778,6 +782,176 @@ function messageArtifact(guild: DiscordGuild, channel: DiscordChannel, msg: Disc
...
+		meta: {
+			...meta,
+			provider: DISCORD_PROVIDER_KIND,
+			recordType: "message",
+			guildId,
+			channelId,
+			messageId,
+			editedAt: editedAt ?? null,
+			...(typeof payload.pinned === "boolean" ? { pinned: payload.pinned } : {}),
+			...(typeof payload.flags === "number" ? { flags: payload.flags } : {}),
```
PR-Reviewer-Ant, line 836: When a partial `MESSAGE_UPDATE` carries `pinned:false`, the metadata is patched to `pinned:false`, but the indexed message content can still keep the old `Pinned: true` header. `patchMessageContent()` receives the previous meta, so `messageBodyStartIndex()` consumes the old pinned line as part of the header and preserves it; if there is no content field, the stale line is also preserved unchanged. That leaves the source artifact internally contradictory after an unpin event. The patcher needs to update/remove the generated pinned header in the content when the payload includes `pinned`, similar to how it handles `Edited:`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `484881f8`

</details>

I found one consistency bug in the partial update patch path: `pinned:false` updates metadata but leaves the old `Pinned: true` line in the indexed message text.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed code explicitly merges `pinned` from the update payload into metadata, while `patchMessageContent` is called with the old metadata and never removes an existing `Pinned: true` content line. This is directly visible in platform/daemon/src/discord-source-provider.ts.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Addressed the pinned content consistency issue. Partial updates now pass the authoritative `pinned` payload state into message content patching, remove a stale `Pinned: true` line when `pinned:false`, and insert it if a partial update pins a message whose artifact text lacks the header. The gateway-tail regression now starts pinned, clears it through the partial update, and asserts both content and metadata agree.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `0ced8b1a`

</details>

I did not find a concrete correctness, security, or data-integrity issue in the current diff. The latest changes address the previously open pinned/header/child-artifact concerns with targeted code and regression coverage.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff directly shows partial MESSAGE_UPDATE handling, message content patching, metadata patching, and child artifact cleanup, and the gateway-tail test now covers partial payloads with missing author/timestamp plus cleared attachments, embeds, mentions, poll, and pinned state. I did not run a live Discord gateway repro, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #779, opened 2026-05-25
feat(recall): add exact date temporal recall

NicholaiVogel:
## Summary
- Add a `temporal_edges` sidecar table for explicit event/source/observed/valid memory time without duplicating memory content.
- Resolve exact date recall from query text or explicit `time` options, supporting date-only timeline recall and date-plus-topic filtered recall.
- Wire temporal recall and temporal remember fields through daemon routes, hooks, MCP tools, CLI, SDK, and core helpers.
- Document the new Memory API, CLI, SDK behavior, and keep doc drift checks aligned with the current package-map format.

## Validation
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `bun scripts/doc-drift.ts --markdown`
- `git diff --check`

## Notes
- Daemon build passes with existing dashboard Svelte warnings in SecretsTab, TroubleshooterPanel, SidebarGroups, WidgetGrid, and WidgetCard.
- Biome warnings from pre-existing explicit `any` usages in memory search/routes are unchanged by this PR.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Rollback / compatibility note: migration 076 only creates the additive `temporal_edges` sidecar table and indexes. Existing memory/session/source recall remains compatible; rollback can ignore or drop the sidecar if the feature is disabled or reverted.

### DIFF
--- a/platform/daemon/src/memory-search.ts
+++ b/platform/daemon/src/memory-search.ts
@@ -972,6 +972,22 @@ function cosineSimilarity(query: Float32Array, memory: Float32Array): number {
 // not outrank directly grounded lexical/vector/structured evidence. A hint
 // supported by direct evidence can keep its score; only hint-only rows are capped.
 const HINT_ONLY_SCORE_CAP = 0.75;
+const TEMPORAL_TOPIC_SCORE_CAP = 0.85;
+
+function temporalTopicTokens(raw: string): string[] {
+	return [...new Set(tokenizeGraphQuery(raw))];
+}
+
+function scoreTemporalTopicEvidence(query: string, content: string): number {
+	const queryTokens = temporalTopicTokens(query);
+	if (queryTokens.length === 0) return 0;
+	const contentTokens = new Set(temporalTopicTokens(content));
+	const matched = queryTokens.filter((token) => contentTokens.has(token));
+	if (matched.length === 0) return 0;
+	const coverage = matched.length / queryTokens.length;
+	const density = matched.length / Math.max(8, contentTokens.size);
+	return Math.min(TEMPORAL_TOPIC_SCORE_CAP, 0.35 + coverage * 0.4 + density * 0.1);
+}
 
 /**
  * Run the recall pipeline.
@@ -1103,6 +1119,9 @@ export async function hybridRecall(
 		query = temporal.adjustedQuery;
 	}
 	const temporalCandidateSet = new Set(temporal.candidateIds ?? []);
+	const temporalCandidateMap = new Map<string, number>(
+		[...temporalCandidateSet].map((id) => [id, Math.max(minScore, 0.05)]),
+	);
 
 	const expandedQuery = expandRecallKeywordQuery(query);
 	const keywordQuery = sanitizeFtsQuery((params.keywordQuery ?? expandedQuery).trim());
@@ -1296,7 +1315,13 @@ export async function hybridRecall(
 	}
 
 	// --- Flat search: merge BM25 + vector + structured path candidate scores ---
-	const allIds = new Set([...bm25Map.keys(), ...hintMap.keys(), ...vectorMap.keys(), ...structuredCandidateMap.keys()]);
+	const allIds = new Set([
+		...bm25Map.keys(),
+		...hintMap.keys(),
+		...vectorMap.keys(),
+		...structuredCandidateMap.keys(),
+		...temporalCandidateMap.keys(),
+	]);
 	const flatScored: Array<{ id: string; score: number; source: string }> = [];
 
 	timings.time("flat_score_merge", () => {
@@ -1305,6 +1330,7 @@ export async function hybridRecall(
 			const hint = hintMap.get(id) ?? 0;
 			const vec = vectorMap.get(id) ?? 0;
 			const structured = structuredCandidateMap.get(id) ?? 0;
+			const temporalCandidate = temporalCandidateMap.get(id) ?? 0;
 			const topicEvidence = bm25 > 0 || hint > 0 || vec > 0 || structured > 0;
 			const temporalScore = temporalCandidateSet.has(id) && topicEvidence ? 0.85 : 0;
 			let score: number;
@@ -1322,6 +1348,9 @@ export async function hybridRecall(
 			} else if (temporalScore > 0) {
 				score = temporalScore;
 				source = "temporal";
+			} else if (temporalCandidate > 0) {
+				score = temporalCandidate;
+				source = "temporal_candidate";
 			} else {
 				score = structured;
 				source = "structured";
@@ -1609,6 +1638,43 @@ export async function hybridRecall(
 	// Everything below this point may assume `scored` only contains memory IDs
 	// visible to the caller. Keep new content-bearing stages below this line.
 	const structuredEvidenceMap = new Map(structuredCandidateMap);
+	const temporalTopicEvidenceMap = new Map<string, number>();
+	if (temporalCandidateSet.size > 0 && scored.length > 0) {
+		try {
+			const candidateIds = scored.filter((row) => temporalCandidateSet.has(row.id)).map((row) => row.id);
+			if (candidateIds.length > 0) {
+				const placeholders = candidateIds.map(() => "?").join(", ");
+				const contentRows = getDbAccessor().withReadDb(
+					(db) =>
+						db
+							.prepare(
+								`SELECT id, content FROM memories
+								 WHERE id IN (${placeholders})`,
+							)
+							.all(...candidateIds) as Array<{ id: string; content: string }>,
+				);
+				for (const row of contentRows) {
+					const score = scoreTemporalTopicEvidence(query, row.content);
+					if (score <= 0) continue;
+					temporalTopicEvidenceMap.set(row.id, score);
+					semanticEvidenceMap.set(row.id, Math.max(semanticEvidenceMap.get(row.id) ?? 0, score));
+				}
+			}
+		} catch (e) {
+			logger.warn("memory", "Temporal topic evidence failed (non-fatal)", {
+				error: e instanceof Error ? e.message : String(e),
+			});
+		}
+		for (const row of scored) {
+			const topicScore = temporalTopicEvidenceMap.get(row.id) ?? 0;
+			if (topicScore <= 0) continue;
+			if (row.source === "temporal_candidate" || topicScore > row.score) {
+				row.score = topicScore;
+				row.source = row.source === "temporal_candidate" ? "temporal_topic" : "temporal_hybrid";
+			}
+		}
+		scored.sort((a, b) => b.score - a.score);
+	}
 
 	// --- Structured Evidence Convolution (SEC-lite) ---
 	// Keep retrieval channels separate until after traversal/boosting. This
@@ -1643,7 +1709,7 @@ export async function hybridRecall(
 			const evidence: EvidenceCandidateInput[] = candidates.map((row) => ({
 				id: row.id,
 				source: row.source,
-				lexical: bm25Map.get(row.id),
+				lexical: Math.max(bm25Map.get(row.id) ?? 0, temporalTopicEvidenceMap.get(row.id) ?? 0),
 				semantic: semanticEvidenceMap.get(row.id),
 				hint: hintMap.get(row.id),
 				traversal: traversalEvidenceMap.get(row.id),
@@ -1901,7 +1967,16 @@ export async function hybridRecall(
 
 	timings.time("final_rank", () => {
 		if (temporalCandidateSet.size > 0) {
-			scored = scored.filter((row) => temporalCandidateSet.has(row.id));
+			scored = scored.filter((row) => {
+				if (!temporalCandidateSet.has(row.id)) return false;
+				return (
+					(bm25Map.get(row.id) ?? 0) > 0 ||
+					(vectorMap.get(row.id) ?? 0) > 0 ||
+					(hintMap.get(row.id) ?? 0) > 0 ||
+					(structuredEvidenceMap.get(row.id) ?? 0) > 0 ||
+					(temporalTopicEvidenceMap.get(row.id) ?? 0) > 0
+				);
+			});
 		}
 		for (const row of scored) {
 			const hasDirectEvidence =
--- a/platform/daemon/src/memory-search.ts
+++ b/platform/daemon/src/memory-search.ts
@@ -15,13 +15,15 @@
 import { createHash } from "node:crypto";
 import {
 	LEGACY_OBSIDIAN_CHUNK_SOURCE_TYPE,
-	SOURCE_CHUNK_SOURCE_TYPE,
 	type LlmUsage,
+	type RecallTemporalMeta,
+	SOURCE_CHUNK_SOURCE_TYPE,
 	vectorSearch,
 } from "@signet/core";
 import { getDbAccessor } from "./db-accessor";
 import { getLlmProvider } from "./llm";
 import { logger } from "./logger";
+import { buildAgentScopeClause } from "./memory-access-scope";
 import type { EmbeddingConfig, MemorySearchConfig, ResolvedMemoryConfig } from "./memory-config";
 import { NATIVE_MEMORY_BRIDGE_SOURCE_NODE_ID } from "./native-memory-constants";
 import { constructContextBlocks } from "./pipeline/context-construction";
@@ -40,6 +42,7 @@ import {
 import { findStructuredPathCandidates, scoreStructuredPathEvidence } from "./pipeline/structured-path-evidence";
 import { type RecallDedupeMeta, applyRecallDedupe } from "./session-recall-dedupe";
 import { escapeLike } from "./sql-utils";
+import { type TemporalTimeOptions, resolveTemporalRecall } from "./temporal-recall";
 
 // ---------------------------------------------------------------------------
 // Public interfaces
@@ -66,6 +69,7 @@ export interface RecallParams {
 	importance_min?: number;
 	since?: string;
 	until?: string;
+	time?: TemporalTimeOptions;
 	scope?: string | null;
 	expand?: boolean;
 	/** When set, restricts results to memories belonging to this project (auth scope enforcement). */
@@ -117,6 +121,7 @@ export interface RecallResponse {
 		hasSupplementary: boolean;
 		noHits: boolean;
 		timings: RecallTimings;
+		temporal?: RecallTemporalMeta;
 		dedupe?: RecallDedupeMeta;
 	};
 	aggregate?: {
@@ -210,33 +215,7 @@ function createRecallTimingCollector(): {
 	};
 }
 
-// ---------------------------------------------------------------------------
-// Agent scope clause (exported for testing)
-// ---------------------------------------------------------------------------
-
-export function buildAgentScopeClause(
-	agentId: string,
-	readPolicy: string,
-	policyGroup: string | null,
-): { sql: string; args: unknown[] } {
-	if (readPolicy === "shared") {
-		return {
-			sql: " AND (m.visibility = 'global' OR m.agent_id = ?) AND m.visibility != 'archived'",
-			args: [agentId],
-		};
-	}
-	if (readPolicy === "group" && policyGroup) {
-		return {
-			sql: " AND ((m.visibility = 'global' AND m.agent_id IN (SELECT id FROM agents WHERE policy_group = ?)) OR m.agent_id = ?) AND m.visibility != 'archived'",
-			args: [policyGroup, agentId],
-		};
-	}
-	// 'isolated', 'group' without policyGroup, or unknown — own memories only
-	return {
-		sql: " AND m.agent_id = ? AND m.visibility != 'archived'",
-		args: [agentId],
-	};
-}
+export { buildAgentScopeClause } from "./memory-access-scope";
 
 // ---------------------------------------------------------------------------
 // Filter clause builder (private)
@@ -1014,9 +993,7 @@ export async function hybridRecall(
 	cfg: ResolvedMemoryConfig,
 	embedFn: EmbedFn,
 ): Promise<RecallResponse> {
-	const query = params.query;
-	const expandedQuery = expandRecallKeywordQuery(params.query);
-	const keywordQuery = sanitizeFtsQuery((params.keywordQuery ?? expandedQuery).trim());
+	let query = params.query;
 	const limit = normalizeRecallLimit(params.limit);
 	const alpha = cfg.search.alpha;
 	const minScore = cfg.search.min_score;
@@ -1108,6 +1085,27 @@ export async function hybridRecall(
 			},
 		};
 	};
+
+	const temporal = resolveTemporalRecall({
+		query,
+		time: params.time,
+		limit,
+		agentId: params.agentId,
+		readPolicy: params.readPolicy,
+		policyGroup: params.policyGroup,
+		project: params.project,
+		sessionKey: params.sessionKey,
+	});
+	if (temporal.response) {
+		return await finish(temporal.response);
+	}
+	if (temporal.adjustedQuery) {
+		query = temporal.adjustedQuery;
+	}
+	const temporalCandidateSet = new Set(temporal.candidateIds ?? []);
+
+	const expandedQuery = expandRecallKeywordQuery(query);
+	const keywordQuery = sanitizeFtsQuery((params.keywordQuery ?? expandedQuery).trim());
 	const queryVecPromise = (() => {
 		const embeddingStart = performance.now();
 		let promise: Promise<number[] | null>;
@@ -1298,7 +1296,13 @@ export async function hybridRecall(
 	}
 
 	// --- Flat search: merge BM25 + vector + structured path candidate scores ---
-	const allIds = new Set([...bm25Map.keys(), ...hintMap.keys(), ...vectorMap.keys(), ...structuredCandidateMap.keys()]);
+	const allIds = new Set([
+		...bm25Map.keys(),
+		...hintMap.keys(),
+		...vectorMap.keys(),
+		...structuredCandidateMap.keys(),
+		...temporalCandidateSet,
+	]);
 	const flatScored: Array<{ id: string; score: number; source: string }> = [];
 
 	timings.time("flat_score_merge", () => {
@@ -1307,6 +1311,7 @@ export async function hybridRecall(
 			const hint = hintMap.get(id) ?? 0;
 			const vec = vectorMap.get(id) ?? 0;
 			const structured = structuredCandidateMap.get(id) ?? 0;
+			const temporalScore = temporalCandidateSet.has(id) ? 0.85 : 0;
 			let score: number;
 			let source: string;
 
@@ -1319,6 +1324,9 @@ export async function hybridRecall(
 			} else if (bm25 > 0) {
 				score = bm25;
 				source = "keyword";
+			} else if (temporalScore > 0) {
+				score = temporalScore;
+				source = "temporal";
 			} else {
 				score = structured;
 				source = "structured";
@@ -1331,6 +1339,10 @@ export async function hybridRecall(
 			if (structured > 0 && structured >= score) {
 				score = structured;
 				source = bm25 > 0 || vec > 0 || hint > 0 ? "sec" : "structured";
+			}
+			if (temporalScore > 0 && temporalScore >= score) {
+				score = temporalScore;
+				source = bm25 > 0 || vec > 0 || hint > 0 || structured > 0 ? "temporal_hybrid" : "temporal";
 			}
 
 			if (score >= minScore) flatScored.push({ id, score, source });
@@ -1893,6 +1905,9 @@ export async function hybridRecall(
 	}
 
 	timings.time("final_rank", () => {
+		if (temporalCandidateSet.size > 0) {
+			scored = scored.filter((row) => temporalCandidateSet.has(row.id));
+		}
 		for (const row of scored) {
 			const hasDirectEvidence =
 				(bm25Map.get(row.id) ?? 0) > 0 ||
@@ -2453,6 +2468,7 @@ export async function hybridRecall(
 			totalReturned: results.length,
 			hasSupplementary: results.some((row) => row.supplementary === true),
 			noHits: results.length === 0,
+			...(temporal.meta ? { temporal: temporal.meta } : {}),
 		},
 		entities: entityContext && entityContext.length > 0 ? entityContext : undefined,
 	});
--- a/platform/daemon/src/memory-search.ts
+++ b/platform/daemon/src/memory-search.ts
@@ -1296,13 +1296,7 @@ export async function hybridRecall(
 	}
 
 	// --- Flat search: merge BM25 + vector + structured path candidate scores ---
-	const allIds = new Set([
-		...bm25Map.keys(),
-		...hintMap.keys(),
-		...vectorMap.keys(),
-		...structuredCandidateMap.keys(),
-		...temporalCandidateSet,
-	]);
+	const allIds = new Set([...bm25Map.keys(), ...hintMap.keys(), ...vectorMap.keys(), ...structuredCandidateMap.keys()]);
 	const flatScored: Array<{ id: string; score: number; source: string }> = [];
 
 	timings.time("flat_score_merge", () => {
@@ -1311,7 +1305,8 @@ export async function hybridRecall(
 			const hint = hintMap.get(id) ?? 0;
 			const vec = vectorMap.get(id) ?? 0;
 			const structured = structuredCandidateMap.get(id) ?? 0;
-			const temporalScore = temporalCandidateSet.has(id) ? 0.85 : 0;
+			const topicEvidence = bm25 > 0 || hint > 0 || vec > 0 || structured > 0;
+			const temporalScore = temporalCandidateSet.has(id) && topicEvidence ? 0.85 : 0;
 			let score: number;
 			let source: string;
 
@@ -1928,7 +1923,7 @@ export async function hybridRecall(
 			: limit;
 	const topIds = params.sourceOnly === true ? [] : scored.slice(0, preHydrate).map((s) => s.id);
 	const recallTruncate = cfg.pipelineV2.guardrails.recallTruncateChars;
-	const allowSourceFallbacks = !hasMemoryMetadataFilters(params);
+	const allowSourceFallbacks = temporalCandidateSet.size === 0 && !hasMemoryMetadataFilters(params);
 
 	if (topIds.length === 0) {
 		const fallbackLimit = selectionDedupeEnabled ? Math.max(limit * 3, limit + 10) : limit;
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -119,7 +119,7 @@ function collectExplicitTemporalInputs(body: {
 	if (body.occurredAt !== undefined && !occurredAt) return { error: "occurredAt must be a valid ISO timestamp" };
 	if (body.validFrom !== undefined && !validFrom) return { error: "validFrom must be a valid ISO timestamp" };
 	if (body.validUntil !== undefined && !validUntil) return { error: "validUntil must be a valid ISO timestamp" };
-	if (validUntil && validFrom && validUntil < validFrom) return { error: "validUntil must be after validFrom" };
+	if (validUntil && validFrom && validUntil <= validFrom) return { error: "validUntil must be after validFrom" };
 
 	const inputs: ExplicitTemporalInput[] = [];
 	if (sourceCreatedAt)
--- a/platform/daemon/src/routes/memory-routes.ts
+++ b/platform/daemon/src/routes/memory-routes.ts
@@ -24,6 +24,7 @@ import { recordPathFeedback } from "../path-feedback";
 import { enqueueDocumentIngestJob } from "../pipeline";
 import { parseFeedback, recordAgentFeedback } from "../session-memories";
 import { upsertSessionTranscript } from "../session-transcripts";
+import { createTemporalEdgeId, normalizeTemporalTimestamp } from "../temporal-recall";
 import { txForgetMemory, txIngestEnvelope, txModifyMemory, txRecoverMemory } from "../transactions";
 import { cacheProjection, computeProjection, computeProjectionForQuery, getCachedProjection } from "../umap-projection";
 import {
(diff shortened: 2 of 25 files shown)


PR-Reviewer-Ant:
Hi @NicholaiVogel - I'm taking a look at the feature work in `feat(recall): add exact date temporal recall` (commit `f9b41ca0`) and will follow up shortly.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,715 @@
...
+		}
+
+		if (temporalFacetAllowed(intent.facets, "session") && tableExists(db, "session_transcripts")) {
+			const updatedExpr = columnExists(db, "session_transcripts", "updated_at")
+				? "COALESCE(updated_at, created_at)"
+				: "created_at";
+			const project = projectSql(params.project, "project");
+			const transcriptRows = db
+				.prepare(
+					`SELECT session_key, content, harness, project, created_at, ${updatedExpr} AS end_at
```
PR-Reviewer-Ant: This pulls raw `session_transcripts.content` into `/api/memory/recall` for date-only temporal recall. That changes the recall contract from memory/source/session-summary recall into transcript search: a user asking `signet recall 2026/05/13` can now receive transcript text without using the dedicated `signet session search` surface or `expand` semantics. The repo docs in this PR still describe `signet session search` as separate from `signet recall`, and the architecture doc says transcript content is joined back only via `expand: true`; this path bypasses that boundary. I would keep timeline recall to session summaries and source/memory metadata, or route transcript lookup through the existing session-search API explicitly.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,715 @@
...
+	return facets.length > 0 ? facets : DEFAULT_TEMPORAL_FACETS;
+}
+
+function parseTimeRange(time: TemporalTimeOptions | undefined): { start: string; end: string } | null {
+	if (!time?.start) return null;
+	const startDate = new Date(time.start);
+	if (Number.isNaN(startDate.getTime())) return null;
+	if (time.end) {
+		const endDate = new Date(time.end);
+		if (Number.isNaN(endDate.getTime())) return null;
```
PR-Reviewer-Ant, line 214: Explicit `time` options are silently ignored when `start` or `end` is invalid, and reversed ranges are accepted as `{ start, end }` even though they cannot produce a sensible window. Since CLI/MCP/SDK now expose these fields and the PR checklist claims input validation, this should fail closed at the API boundary with a structured 400 instead of falling back to ordinary recall or returning an empty temporal result.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f9b41ca0`

</details>

I found one recall-surface regression that should be fixed before merge: date-only memory recall now returns raw session transcript text, even though transcripts are supposed to stay on the dedicated session-search surface. I also found a smaller validation gap around explicit temporal ranges.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The transcript path is directly introduced in platform/daemon/src/temporal-recall.ts by selecting content from session_transcripts and returning it through hybridRecall's temporal response. The invalid-time behavior is also directly visible in parseTimeRange, which returns null for malformed explicit time input and does not reject end <= start.

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `875390e46`: temporal recall no longer reads `session_transcripts`, so date-only recall stays on session summaries plus memory/source temporal metadata. I also added a regression fixture that seeds raw transcript text for the same date and asserts it is not present in the recall response.

Verified with:
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun run typecheck`

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `875390e46`: explicit `time` options are now validated at both `/api/memory/recall` and `/api/hooks/recall` before recall runs. Invalid `start`, invalid `end`, reversed ranges, invalid facets, and invalid modes now return structured 400s instead of falling through to ordinary recall.

Verified with:
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun run typecheck`

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,715 @@
...
+	readonly scope?: string | null;
+	readonly facet: TemporalFacet;
+	readonly start_at: string;
+	readonly end_at: string | null;
+	readonly subject_type: string;
+	readonly subject_id: string;
+	readonly source_rank: number;
+}
+
+function pad(value: number): string {
```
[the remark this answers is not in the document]
NicholaiVogel: Fixed in `90503c259`: removed the unused `pad` helper.

Verified with:
- `bunx biome check platform/daemon/src/temporal-recall.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`

NicholaiVogel:
PR feedback has been addressed on latest head `90503c259`.

Changes since the review:
- Removed `session_transcripts` from temporal recall results; date-only recall now stays on session summaries plus memory/source temporal metadata.
- Added API-boundary validation for explicit temporal `time` options on `/api/memory/recall` and `/api/hooks/recall`.
- Added regression coverage for both review findings.
- Removed the unused `pad` helper flagged by code quality.

Local verification:
- `bunx biome check platform/daemon/src/temporal-recall.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `bun scripts/doc-drift.ts --markdown && git diff --check`

@PR-Reviewer-Ant please re-review the latest head when available.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,700 @@
...
+	const facets = [...new Set(input.filter((facet) => allowed.has(facet)))];
+	return facets.length > 0 ? facets : DEFAULT_TEMPORAL_FACETS;
+}
+
+function parseTimeRange(time: TemporalTimeOptions | undefined): { start: string; end: string } | null {
+	if (!time?.start) return null;
+	const startDate = new Date(time.start);
+	if (Number.isNaN(startDate.getTime())) return null;
+	if (time.end) {
+		const endDate = new Date(time.end);
```
PR-Reviewer-Ant, line 213: An explicit `time.mode: "timeline"` is ignored whenever the query is non-empty because this expression only honors explicit `filter` and otherwise switches to `filter` for any content query. That breaks the documented/API behavior for callers that pass a topic label plus an explicit time window but still want a timeline, including the new CLI shape `signet recall "temporal recall" --time-start ... --time-mode timeline`. Please make the explicit mode win here, and only infer `filter` from `contentQuery.length > 0` when mode is `auto` or absent.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `90503c25`

</details>

I found one correctness issue in the temporal mode resolution: an explicit `timeline` request is ignored whenever the query text is non-empty, which contradicts the new CLI/API contract and one of the documented examples.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The new CLI docs and CLI implementation expose `--time-mode timeline` for normal recall text plus an explicit time range, but `parseTemporalRecallIntent` forces request-based recalls with any non-empty query into filter mode. This is directly visible in the changed temporal recall code.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `0bb0b43af`: explicit `time.mode` now wins over inferred mode. `auto` or absent mode still infers `filter` from non-empty content, but explicit `timeline` stays timeline.

I also changed temporal row filtering so timeline mode returns the timeline window instead of filtering rows by the query text, and added a regression for `query + explicit time window + mode: timeline`.

Verified with:
- `bunx biome check platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

NicholaiVogel:
Latest reviewer finding is addressed on head `0bb0b43af`.

New change:
- Explicit temporal mode now wins over auto inference, so `time.mode: "timeline"` with query text stays timeline.
- Timeline mode no longer filters temporal rows by the query text.
- Added regression coverage for explicit timeline mode with a non-empty query.

Local verification:
- `bunx biome check platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

@PR-Reviewer-Ant please re-review the latest head.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `0bb0b43a`

</details>

I found one validation bug in the new temporal remember fields. The rest of the temporal recall wiring looks generally consistent with the PR description from the diff provided.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The invalid validity-window check is visible directly in platform/daemon/src/routes/memory-routes.ts and conflicts with the documented API contract. I did not run the PR's test suite in this read-only review environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -93,6 +94,92 @@ function parseOptionalIsoTimestamp(value: unknown): string | null {
...
+	const validFrom = normalizeTemporalTimestamp(body.validFrom);
+	const validUntil = normalizeTemporalTimestamp(body.validUntil);
+	if (body.sourceCreatedAt !== undefined && !sourceCreatedAt)
+		return { error: "sourceCreatedAt must be a valid ISO timestamp" };
+	if (body.observedAt !== undefined && !observedAt) return { error: "observedAt must be a valid ISO timestamp" };
+	if (body.occurredAt !== undefined && !occurredAt) return { error: "occurredAt must be a valid ISO timestamp" };
+	if (body.validFrom !== undefined && !validFrom) return { error: "validFrom must be a valid ISO timestamp" };
+	if (body.validUntil !== undefined && !validUntil) return { error: "validUntil must be a valid ISO timestamp" };
+	if (validUntil && validFrom && validUntil < validFrom) return { error: "validUntil must be after validFrom" };
+
```
PR-Reviewer-Ant, line 123: `validUntil` equal to `validFrom` currently passes validation because this only checks `<`. The docs and error text both say `validUntil` must be after `validFrom`, so a zero-length validity window should be rejected with `<=` here. As written, callers can store a temporal `valid` edge that the API contract says is invalid.

NicholaiVogel: Fixed in `e2e6ab365`: `validUntil` now must be strictly after `validFrom`, so equal timestamps are rejected with the documented `validUntil must be after validFrom` 400.

Added a route-level regression for a zero-length validity window on `/api/memory/remember`.

Verified with:
- `bunx biome check platform/daemon/src/routes/memory-routes.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

NicholaiVogel:
Latest reviewer finding is addressed on head `e2e6ab365`.

New change:
- `/api/memory/remember` now rejects `validUntil === validFrom` instead of only rejecting `validUntil < validFrom`.
- Added regression coverage for the zero-length validity window.

Local verification:
- `bunx biome check platform/daemon/src/routes/memory-routes.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/daemon/src/daemon-auth-guard-colocation.test.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

@PR-Reviewer-Ant please re-review the latest head.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,708 @@
...
+		...(row.session_id ? { session_id: row.session_id } : {}),
+		...(row.source_path ? { source_path: row.source_path } : {}),
+		type: row.type,
+		tags: row.tags,
+		pinned: row.pinned,
+		importance: row.importance,
+		who: row.who,
+		project: row.project,
+		created_at: row.created_at,
+		visibility: row.visibility,
```
PR-Reviewer-Ant, line 379: This timeline query hard-codes `session_summaries.agent_id = ?`, so temporal recall ignores the caller's `readPolicy`/`policyGroup` for session-derived rows. In shared or group mode, normal recall is expected to honor the agent visibility model; this new path will silently omit accessible shared timeline evidence even though the PR readiness checklist says agent scoping was verified across new data queries. Either make the policy explicit for sessions, or document and test that temporal session timelines are intentionally own-agent only.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,708 @@
...
+			for (const row of memoryRows) {
+				rows.push({
+					id: row.id,
+					content: row.content,
+					project: row.project,
+					created_at: row.created_at,
+					source_id: row.source_id,
+					type: row.type,
+					who: row.who,
+					importance: row.importance,
```
PR-Reviewer-Ant, line 476: The `memory_artifacts` temporal queries also use only `agent_id = ?` and never apply the shared/group access policy. That makes the date-only timeline behavior diverge from the memory visibility rules used elsewhere in this same file for `memories` and `temporal_edges`: agents with shared/group read policy won't see source activity they should be allowed to recall. The new tests cover isolated access, but not shared/group policy for source-backed temporal rows.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `e2e6ab36`

</details>

I found one blocking scoping issue in the new temporal timeline path. The explicit memory-edge path is guarded by memory visibility, but session summaries and source artifacts bypass the shared/group read policy and can under-return expected shared recall results.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The new temporal recall code is fully present in the diff, and the affected queries are visible in platform/daemon/src/temporal-recall.ts. I did not have the full current data-access conventions for session_summaries and memory_artifacts beyond the supplied repo conventions, so the severity is based on the PR's stated agent-scoping checklist and the changed query behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `12eb4ef1a`: temporal session rows now use an explicit owner-scope predicate instead of hard-coding the requesting agent. Isolated remains own-agent; group mode includes same-policy-group agents; shared mode includes own rows plus rows from agents that have opted into shared read policy.

Added regression coverage for group-policy temporal session rows.

Verified with:
- `bunx biome check platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `12eb4ef1a`: temporal `memory_artifacts` captured/source queries now use the same explicit owner-scope helper as session summaries instead of hard-coding `agent_id = ?`.

The new regression covers source-backed temporal rows under group policy and verifies a same-group artifact is returned while another group’s artifact is not.

Verified with:
- `bunx biome check platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

NicholaiVogel:
Latest reviewer scoping findings are addressed on head `12eb4ef1a`.

New change:
- Temporal timeline session summaries and source artifacts now apply explicit owner-scope predicates for isolated/group/shared read policies.
- Added regression coverage showing group-policy temporal recall includes same-group session/source rows and excludes another group.

Local verification:
- `bunx biome check platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

@PR-Reviewer-Ant please re-review the latest head.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,727 @@
...
+					source_path: string;
+					source_kind: string;
+					source_id: string | null;
+					harness: string | null;
+					project: string | null;
+					content: string;
+					source_at: string;
+				}>;
+				for (const row of sourceRows) {
+					rows.push({
```
PR-Reviewer-Ant, line 566: Explicit temporal edges are scoped with `te.agent_id = ?`, so shared/group-readable memories written by another agent won't participate in exact-date recall even when the joined memory row is visible under `memoryVisibilitySql`. That makes temporal recall stricter than normal memory recall and breaks the PR's “agent scoping verified” claim for the new sidecar table. The edge lookup needs to scope through the readable memory row for `subject_type = 'memory'`, or otherwise mirror the same group/shared policy used for memories.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,727 @@
...
+				created_at: string | null;
+				visibility: string | null;
+				scope: string | null;
+			}>;
+			for (const row of edgeRows) {
+				if (params.project && row.project !== params.project) continue;
+				rows.push({
+					id: row.id,
+					content: row.memory_content ?? `[Temporal ${row.facet}: ${row.subject_type} ${row.subject_id}]`,
+					project: row.project,
```
PR-Reviewer-Ant, line 643: Date-plus-topic recall filters temporal rows with a literal all-terms substring check before hybrid recall gets a chance to rank candidates. That means a memory with an exact `occurredAt` edge can be excluded just because its content uses different wording, even though normal recall would match it via embeddings, hints, or structured evidence. Since the feature is described as date-plus-topic filtered recall, the temporal window should constrain the candidate set and let the existing hybrid machinery handle the topic match, not pre-drop candidates with `textMatches`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `12eb4ef1`

</details>

I found two correctness problems in the temporal recall path. Both are in the new query-side implementation and can make the feature miss memories it claims to support.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed temporal recall code explicitly filters temporal edge rows by te.agent_id = ? and prefilters date-plus-topic rows with textMatches before handing memory IDs back to hybridRecall. These behaviors are directly visible in platform/daemon/src/temporal-recall.ts and conflict with the PR's claimed scoped temporal recall behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `b345f2622`: memory-backed temporal edges now scope through the joined readable memory row instead of requiring `te.agent_id` to equal the requesting agent. Non-memory temporal edges still use the explicit owner-scope predicate.

Added regression coverage for group-policy temporal memory edges.

Verified with:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts && git diff --check`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `b345f2622`: filter-mode temporal recall no longer pre-drops memory candidates with the literal `textMatches` all-term check. The temporal window now supplies memory candidate IDs, and hybrid recall handles the topic match/ranking path.

Added a regression where a same-group temporal memory uses different wording from the topic query but is still returned, while another group’s directly matching memory is excluded.

Verified with:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts && git diff --check`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

NicholaiVogel:
Latest temporal candidate findings are addressed on head `b345f2622`.

New change:
- Memory-backed temporal edges scope through the readable memory row for group/shared access instead of requiring the edge owner to match the caller.
- Date-plus-topic recall no longer lexically prefilters memory temporal candidates before hybrid recall.
- Added regression coverage for scoped temporal edge candidates with different wording from the topic query.

Local verification:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts && git diff --check`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`

@PR-Reviewer-Ant please re-review the latest head.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,739 @@
...
+}
+
+function overlapClause(startExpr: string, endExpr: string): string {
+	return `${startExpr} < ? AND COALESCE(${endExpr}, ${startExpr}) >= ?`;
+}
+
+function textMatches(content: string, query: string): boolean {
+	if (!query) return true;
+	const lower = content.toLowerCase();
+	return query
```
PR-Reviewer-Ant, line 310: This shared-policy predicate can expose other agents' session summaries and source artifacts whenever those other agents have `read_policy = 'shared'`. That is a different rule than memory recall uses: `buildAgentScopeClause` only exposes the caller's own rows plus `visibility = 'global'` memory rows. These non-memory temporal rows do not have a visibility column, so using the target agent's read policy as an exposure signal can leak session/source content across agents. The safe default here should match the data model's available controls, for example own-only unless there is an explicit group policy or a real visibility/public marker for these tables.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,739 @@
...
+				scope: string | null;
+			}>;
+			for (const row of edgeRows) {
+				if (params.project && row.project !== params.project) continue;
+				rows.push({
+					id: row.id,
+					content: row.memory_content ?? `[Temporal ${row.facet}: ${row.subject_type} ${row.subject_id}]`,
+					project: row.project,
+					created_at: row.created_at ?? row.start_at,
+					source_id: row.source_id,
```
PR-Reviewer-Ant, line 645: Date-plus-topic recall is documented as filtering by the remaining topic words, but this condition exempts every `memory` temporal edge from `textMatches`. Those memory IDs are then returned as temporal candidates and can be boosted through `hybridRecall` even when their content does not match the topic at all. For example, a query like `2026/05/13 rollout` can return any memory with a May 13 temporal edge, not just dated memories about rollout. That makes exact date recall much noisier and diverges from the PR description's “date-plus-topic filtered recall” behavior.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `b345f262`

</details>

I found two concrete issues in the new temporal recall path: one cross-agent exposure risk for non-memory temporal rows, and one date-plus-topic filtering bug that can return unrelated dated memories.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - Both findings are directly visible in the new temporal-recall.ts implementation and are tied to the PR's stated temporal recall and agent-scoping claims. The security issue follows from the shared-policy SQL predicate, and the filtering issue follows from memory rows bypassing contentQuery matching before being passed as temporal candidates.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `aeff64b05`: shared-mode temporal owner filtering for non-memory rows is now own-agent only. Group mode still allows explicit same-policy-group session/source timeline evidence, while shared mode no longer treats another agent's `read_policy = 'shared'` as an exposure marker for tables without visibility columns.

Added regression coverage for shared-policy temporal session/source rows to include own rows and exclude another shared agent's rows.

Verified with:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `aeff64b05`: temporal candidate IDs no longer enter hybrid recall as standalone scored hits for date-plus-topic recall. The date window now constrains the candidate set, and BM25/vector/hints/structured evidence must still provide the topic match before a memory can rank.

Added regression coverage where a same-group dated memory with topic evidence is returned, another same-group dated but off-topic memory is excluded, and another group's direct lexical match remains excluded.

Verified with:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

NicholaiVogel:
Latest temporal scoping/topic findings are addressed on head `aeff64b05`.

New change:
- Shared-mode temporal session/source rows are own-agent only for tables without visibility columns; explicit group mode still admits same-group timeline evidence.
- Date-plus-topic memory recall now uses the temporal window as a candidate constraint, but requires normal hybrid topic evidence before a dated memory can rank.
- Added regressions for shared non-memory temporal leakage and noisy date-plus-topic memory candidates.

Local verification:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

@PR-Reviewer-Ant please re-review the latest head.

platform/daemon/src/routes/memory-routes.ts
```diff
@@ -1236,10 +1330,26 @@ export function registerMemoryRoutes(app: Hono, deps: MemoryRoutesDeps = {}): vo
 					visibility,
 					createdAt,
 				});
+				txInsertExplicitTemporalEdges({
+					db,
```
PR-Reviewer-Ant, line 1336: On the dedupe path, this writes explicit temporal edges with the current request's `agentId`, even though `result.row.id` may be an existing global memory owned by another agent. The edge ID also includes that caller agent ID, so repeated dedupes from different agents can create multiple temporal edges for the same memory/facet/time, and the sidecar no longer reflects the memory row's ownership. This matters because temporal recall later treats memory-subject edges through memory visibility, not `te.agent_id`, so mis-owned sidecar rows can affect timeline/filter behavior across agents. Use the reused memory row's owner for the edge or reject/normalize cross-agent temporal annotation on dedupe.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `aeff64b0`

</details>

I found one correctness issue in the temporal edge write path: deduped remembers can attach temporal metadata under the caller's agent ID even when the reused memory belongs to a different agent. That can create duplicate/mis-owned temporal edges and makes the sidecar's ownership model drift from the memory row it describes.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The issue is visible in the changed remember route: dedupe paths call txInsertExplicitTemporalEdges with the request-resolved agentId while using result.row.id or existing.id as the subject memory. The temporal edge id and agent_id are then derived from the caller, not from the reused memory row. I did not run a runtime repro, but the affected paths and ownership mismatch are directly present in the diff.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/routes/memory-routes.ts
NicholaiVogel: Fixed in `cdf203882`: source-id dedupe now uses the same agent/visibility/scope predicate as the other remember dedupe paths, and deduped temporal edge writes use the reused memory row's stored owner instead of the current caller.

Added a route regression where two agents remember the same source id with explicit `occurredAt`; the second write now creates its own memory row and each temporal edge is owned by the corresponding memory owner.

Verified with:
- `bunx biome check --write platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (only existing `noExplicitAny` warnings in `memory-routes.ts`)
- `bun test platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

NicholaiVogel:
Latest temporal dedupe write finding is addressed on head `cdf203882`.

New change:
- Source-id remember dedupe now uses the scoped agent/visibility/scope predicate.
- Temporal edge writes on deduped remembers use the reused memory row owner rather than the current caller.
- Added a regression for same source id plus explicit `occurredAt` across two agents.

Local verification:
- `bunx biome check --write platform/daemon/src/routes/memory-routes.ts platform/daemon/src/mutation-api.test.ts` (only existing `noExplicitAny` warnings in `memory-routes.ts`)
- `bun test platform/daemon/src/daemon-auth-guard-colocation.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

@PR-Reviewer-Ant please re-review the latest head.

platform/daemon/src/temporal-recall.ts
```diff
@@ -0,0 +1,733 @@
...
+		);
+
+	const deduped: RawTemporalRow[] = [];
+	const seen = new Set<string>();
+	for (const row of rows) {
+		const key = `${row.subject_type}:${row.subject_id}:${row.facet}`;
+		if (seen.has(key)) continue;
+		seen.add(key);
+		deduped.push(row);
+		if (deduped.length >= params.limit) break;
```
PR-Reviewer-Ant: This caps date-plus-topic recall to the first `limit` temporal memory rows before the topic query is allowed to filter them. In filter mode, memory rows are intentionally allowed through above regardless of `textMatches`, and `resolveTemporalRecall` returns only these `candidateIds` to `hybridRecall`. If a day has more than `limit` temporal memory edges and the newest/highest-ranked ones are off-topic, the relevant same-day memory never reaches the normal keyword/vector scoring path. That breaks the PR's stated date-plus-topic behavior. The candidate collection needs a larger prefilter budget or should apply topic evidence before truncating to the user-facing limit.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cdf20388`

</details>

I found one correctness issue in date-plus-topic recall: the temporal candidate set is capped before topic evidence is applied, so relevant same-day memories can be silently excluded. The rest of the wiring looks broadly consistent with the PR description from the provided diff.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The failure follows directly from platform/daemon/src/temporal-recall.ts: filter-mode memory rows bypass text filtering, then the dedupe loop stops at params.limit before hybrid recall gets a chance to apply the topic query. The included tests cover a small candidate set but not the case where more than limit same-day temporal memories exist before the relevant topic match.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/temporal-recall.ts
NicholaiVogel: Fixed in `7b6ca81f2`: filter-mode temporal recall now separates the temporal candidate scan budget from the user-facing result limit. Date-plus-topic recall scans a larger bounded candidate pool before handing memory IDs to hybrid recall, while timeline responses still apply the requested result limit.

Added a regression where eight newer same-day off-topic temporal edges would previously exhaust `limit` before the relevant rollout memory reached topic scoring.

Verified with:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

NicholaiVogel:
Latest temporal candidate budget finding is addressed on head `7b6ca81f2`.

New change:
- Filter-mode temporal recall now uses a larger bounded temporal candidate scan budget instead of truncating candidates to the user-facing result limit before topic evidence runs.
- Timeline/date-only responses still apply the requested result limit.
- Added a regression where newer off-topic same-day edges no longer hide the relevant date-plus-topic memory.

Local verification:
- `bunx biome check --write platform/daemon/src/temporal-recall.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

@PR-Reviewer-Ant please re-review the latest head.

platform/daemon/src/memory-search.ts
```diff
@@ -1893,6 +1900,9 @@ export async function hybridRecall(
 	}
 
 	timings.time("final_rank", () => {
+		if (temporalCandidateSet.size > 0) {
+			scored = scored.filter((row) => temporalCandidateSet.has(row.id));
```
PR-Reviewer-Ant: This final filter can drop the exact dated memory before date-plus-topic recall works as described. `resolveTemporalRecall()` returns `candidateIds`, but in this diff those IDs are only converted to a `Set` and used for scoring/filtering; I don't see them being added to the scored candidate pool. If a dated memory does not already appear via BM25/vector/hints/structured evidence, it will never reach `scored`, and this filter leaves no result even though the temporal edge matched the requested date. The tests cover a relevant memory that also has topic evidence, but not the important case where the temporal edge is the only path that should admit the row and the topic check should then run against its content.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `7b6ca81f`

</details>

I found one correctness issue in temporal filter recall: dated memories can be excluded before the topic matcher ever gets a chance to score them. That undercuts the PR's stated date-plus-topic recall behavior.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The temporal candidate IDs are collected in platform/daemon/src/memory-search.ts and later used as a final filter, but the diff does not show those candidate IDs being added to the normal recall candidate pool before scoring. This is directly visible in the changed memory-search.ts flow, though I would want the full surrounding candidate-union code to confirm every path.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/memory-search.ts
NicholaiVogel: Fixed in `3baa9d569`: temporal candidate IDs are now admitted into the normal authorized recall candidate pool before topic scoring. After authorization, the existing tokenization path scores topic overlap against the memory content, feeds that evidence into the structured evidence stage, and the final temporal filter now requires BM25/vector/hint/structured/temporal-topic evidence so date-only matches do not leak into date-plus-topic results.

The regression now uses a short AND-style topic query where the dated memory shares only one topic token, so the old path would not reach `scored` through BM25 before temporal topic evidence had a chance to run.

Verified with:
- `bunx biome check --write platform/daemon/src/memory-search.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

NicholaiVogel:
Latest temporal candidate admission finding is addressed on head `3baa9d569`.

New change:
- Temporal candidate IDs are added to the authorized candidate pool before topic scoring.
- Date-plus-topic recall now scores temporal candidate content after authorization and feeds that evidence through the normal structured evidence path.
- Final temporal-filtered results require actual topic evidence, so unrelated same-day memories remain filtered out.
- Added/tightened regression coverage for a dated memory that would not enter BM25 with a short AND-style topic query.

Local verification:
- `bunx biome check --write platform/daemon/src/memory-search.ts platform/daemon/src/memory-search.test.ts`
- `bun test platform/daemon/src/memory-search.test.ts`
- `bun test platform/core/src/recall.test.ts platform/daemon/src/memory-search.test.ts platform/daemon/src/daemon-auth-guard-colocation.test.ts surfaces/cli/src/commands/memory.test.ts platform/daemon/src/mcp/tools.test.ts`
- `bun run typecheck`
- `git diff --check`
- `bun run --filter '@signet/daemon' build` (passes with existing dashboard Svelte warnings)

@PR-Reviewer-Ant please re-review the latest head.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `3baa9d56`

</details>

I did not find a concrete correctness, security, scoping, or data-integrity issue in the changed code. The implementation appears to match the PR description: it adds the temporal sidecar, validates temporal inputs at the route boundary, threads options through CLI/MCP/SDK/core, and keeps temporal memory-edge recall behind existing memory visibility checks.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The diff includes the migration, daemon implementation, route validation, surface wiring, and targeted tests for temporal recall, scoping, dedupe, and invalid ranges. I did not run the listed validation commands or exercise a live daemon, so this is based on static review of the provided diff and file contents.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
The PR head advanced to `59e996ce` after merging `main` into `nicholai/temporal-recall`.

The last Ant review cleared `3baa9d569` with:
- Still blocking: 0
- Newly found: 0

The new head is a merge commit on top of that cleared commit, so @PR-Reviewer-Ant please re-review the current head `59e996ce` for final human-review clearance.

NicholaiVogel:
The named PR branch ref is restored at current head `59e996ce` and local checkout fast-forwarded to the same head.

@PR-Reviewer-Ant please re-review current head `59e996ce` for final human-review clearance.

NicholaiVogel:
@PR-Reviewer-Ant status

PR #806, opened 2026-06-01
fix(daemon-rs): repair summary_jobs migration columns

NicholaiVogel:
## Summary
- Adds an always-run Rust daemon schema repair for required `summary_jobs` columns, including `leased_at`, `updated_at`, and `failed_at`.
- Backfills `summary_jobs` runtime fields and recreates the agent/session indexes when an upgraded DB already has migration 37 stamped.
- Adds an integration regression test that simulates the partial migration-37 schema from #803 and verifies the leased-job update succeeds.

Closes #803

## Test Plan
- [x] `cd platform/daemon-rs && cargo fmt`
- [x] `cd platform/daemon-rs && cargo test -p signet-core --test migrations`
- [x] `cd platform/daemon-rs && cargo test -p signet-core`
- [x] `cd platform/daemon-rs && cargo test -p signet-pipeline summary::tests::recovers_leased_summary_jobs_in_batches`
- [x] `cd platform/daemon-rs && cargo test -p signet-daemon session_end_persists_transcript_artifact_and_queues_summary`
- [x] `cd platform/daemon-rs && cargo build -p signet-daemon`
- [x] `cd platform/daemon-rs && cargo test -p signet-daemon --test contract_replay hook_session_lifecycle -- --ignored --test-threads=1`
- [x] `git diff --check`

## PR Readiness (MANDATORY)
- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

Applicability notes: this is a Rust daemon migration repair only. It adds no new API route, user-facing docs contract, admin endpoint, mutation endpoint, or agent-scoped data query.

## Migration Notes (if applicable)
- [x] Migration is idempotent
- [x] Daemon Rust parity reviewed or explicitly N/A
- [x] Rollback / compatibility note included in PR description

Rollback / compatibility: the repair only adds missing nullable/defaulted columns that current Rust daemon session-summary code already expects. Existing DBs that already have the columns are unchanged; partial upgraded DBs self-heal on startup instead of failing `session-end` with `summary_jobs has no column named updated_at`.

PR #808, opened 2026-06-01
fix(daemon-rs): make transcript recapture idempotent

### LINKED ISSUE #805: session-end hook returns 500 on concurrent same-session calls: `refusing to mutate immutable artifact ...--transcript.md` (write_transcript_artifact not idempotent under same-session races)
### Environment
- signetai: `0.138.6` (git `a6dde71`, branch `main`)
- Daemon: Rust daemon (`daemon-rs`, `signet-daemon` binary)
- Node.js: `v26.0.0`
- OS: macOS 26.4.1 (build 25E253), arm64
- SQLite: `3.51.0`
- DB schema version (`schema_migrations`): `77`
- Embedding provider: `ollama` (`nomic-embed-text`, 768 dims)
- Pipeline provider: `ollama` (`qwen3:4b`)

### What happened
Intermittently, `POST /api/hooks/session-end` fails with HTTP 500. The harness
surfaces it as a toast:

```
[signet] POST /api/hooks/session-end failed: 500
```

This is **not** the migration-37 / `updated_at` issue (see report 01): that DB
column is present and a single session-end returns 200. The 500 only appears
when multiple session-end calls for the **same session** arrive close together.

Hitting the endpoint with concurrent same-`sessionKey` requests reveals the
underlying error:

```bash
for i in $(seq 1 8); do
  curl -s -o /dev/null -w "%{http_code} " -X POST http://localhost:3850/api/hooks/session-end \
    -H "Content-Type: application/json" \
    -d '{"harness":"opencode","agentId":"default","reason":"session.idle","sessionKey":"race","transcript":"user: msg '"$i"' '"$(seq -s, 1 $i)"'\nassistant: ok"}' &
done; wait
# -> 500 500 500 500 500 500 500 200   (7 of 8 fail)
```

Capturing a 500 body:

```json
{"error":"transcript artifact write failed: migration error: refusing to mutate immutable artifact /Users/.../.agents/memory/2026-06-01T01-45-42.236233+00-00--qhz37djsruihzqb5--transcript.md"}
[HTTP 500]
```

### Root cause
The session-end handler writes the canonical transcript artifact and treats any
failure as a hard 500:

- `platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs:2230-2255`
  (`write_transcript_artifact` failure → `StatusCode::INTERNAL_SERVER_ERROR`,
  `"transcript artifact write failed: {e}"`)

The artifact path is keyed by `(captured_at, session_token)`:

- `platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs:276`
  `fn artifact_path(root, captured_at, token, kind)` → `memory_dir/file_name(captured_at, token, kind)`

`write_immutable_artifact` rejects a path that already exists with a different
body:

- `platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs:780-794`

```rust
if path.exists() {
    let existing = fs::read_to_string(&path).map_err(|err| err.to_string())?;
    let (existing_frontmatter, existing_body) = parse_document(&existing)?;
    let current = read_string(&existing_frontmatter, "content_sha256");
    let next = read_string(&frontmatter, "content_sha256");
    if current == next {
        return Ok(path);                 // identical content -> idempotent OK
    }
    if normalize_markdown_body(&existing_body) != body {
        return Err(format!(              // different content -> hard error -> 500
            "refusing to mutate immutable artifact {}",
            path.display()
        ));
    }
}
```

The race:

1. A harness fires session-end fire-and-forget on **every** `session.idle`
   event (the bundled opencode connector does this:
   `event` handler, `X.type === "session.idle" || "session.deleted"` →
   `_.post("/api/hooks/session-end", ...)`). `session.idle` can fire multiple
   times for one session in quick succession.
2. Two such calls land within the same `captured_at` second resolution and
   derive the **same** `artifact_path`.
3. The session has grown between the two captures, so the transcript **bodies
   differ** → `content_sha256` differs → the second call trips the immutable
   guard → `write_transcript_artifact` returns `Err` → handler returns **500**.

For a **transcript** artifact this is not data loss: the winning (first) capture
is preserved and the raw content of every call is already persisted via
`write_transcript_audit` (`routes/hooks.rs:2154`). The 500 is therefore a
spurious error surfaced to the caller for a benign concurrent re-capture.

Note: these 500 paths only `warn!` on the `transcript_path` branch; the inline
branch returns the 500 with no `tracing` event, so the failure is invisible in
logs (made worse by the launchd plist sending stdout to `/dev/null`).

### Reproduction
1. Run the daemon normally.
2. Fire ≥2 concurrent `POST /api/hooks/session-end` with the **same**
   `sessionKey`, `reason` ≠ `clear`, and **different** transcript bodies (see
   the curl loop above).
3. Observe one 200 and the rest 500 with
   `refusing to mutate immutable artifact ...--transcript.md`.

In real use this happens organically: the opencode connector re-fires
session-end on repeated `session.idle` events while the session keeps growing.

### Workaround (what I applied locally)
Two layers:

1. **Connector-side debounce** (stops the trigger): dedupe session-end per
   `sessionKey` in the opencode connector so at most one `session.idle` fire is
   sent per session per 30s, with `session.deleted` always firing as the
   terminal capture. Patched both `~/.config/opencode/plugins/signet.mjs`
   (the file opencode actually loads) and
   `~/.signet/runtime/connectors/opencode/index.js`.

2. **Daemon-side idempotency** (fixes the bug): make
   `write_immutable_artifact` treat a same-path/different-body collision for
   `ArtifactKind::Transcript` as idempotent (return the existing path, first-
   write-wins) instead of erroring. Summary / Compaction / Manifest keep the
   strict immutability guard, where a differing body would be real data loss.

```rust
if normalize_markdown_body(&existing_body) != body {
    if seed.kind == ArtifactKind::Transcript {
        return Ok(path); // benign concurrent/repeated session-end re-capture
    }
    return Err(format!(
        "refusing to mutate immutable artifact {}",
        path.display()
    ));
}
```

After the daemon-side fix, 8–10 concurrent same-`sessionKey` session-ends with
differing bodies all return `200 {"jobId":...,"queued":true}`.

### Suggested fix
- Adopt the transcript-kind idempotency relaxation above (concurrent/repeated
  session-end for the same session is expected, not an error).
- Optionally disambiguate the artifact path for transcripts when bodies differ
  (e.g. include a short content/sequence suffix) if you'd rather keep every
  distinct capture instead of first-write-wins.
- Emit a `tracing` event (not just the HTTP body) on the inline-transcript 500
  branches in `routes/hooks.rs` so these failures are diagnosable from logs.
- Independently, the bundled connectors should debounce/dedupe session-end so a
  burst of `session.idle` events doesn't hammer the endpoint.

### Regression tests added (local)
`platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs` tests:
- `concurrent_transcript_recapture_is_idempotent_not_error`
- `concurrent_summary_collision_still_refuses_mutation`

NicholaiVogel:
## Summary
- Treat same-path transcript recaptures with different bodies as idempotent in the Rust daemon lineage writer, returning the existing canonical transcript artifact instead of surfacing a spurious `session-end` 500.
- Preserve strict immutability for summary/compaction/manifest artifacts when the body differs.
- Add diagnostic logging for real transcript artifact write failures at the `session-end` HTTP boundary.
- Add integration regression tests for transcript recapture idempotency and summary collision refusal.

Closes #805

## Test Plan
- [x] `cd platform/daemon-rs && cargo fmt`
- [x] `cd platform/daemon-rs && cargo test -p signet-pipeline --test memory_lineage`
- [x] `cd platform/daemon-rs && cargo test -p signet-pipeline memory_lineage`
- [x] `cd platform/daemon-rs && cargo test -p signet-daemon session_end_persists_transcript_artifact_and_queues_summary`
- [x] `cd platform/daemon-rs && cargo build -p signet-daemon`
- [x] `cd platform/daemon-rs && cargo test -p signet-daemon --test contract_replay hook_session_lifecycle -- --ignored --test-threads=1`
- [x] `cd platform/daemon-rs && cargo test -p signet-pipeline`
- [x] `git diff --check`

## PR Readiness (MANDATORY)

Applicability notes: this changes Rust daemon artifact-write behavior only. It adds no new API route, schema migration, admin endpoint, mutation endpoint, or agent-scoped data query. The changed fallback/error path now logs session, agent id, and error context instead of relying only on the HTTP body.

- [x] Migration is idempotent

Rollback / compatibility: no database migration is included. The compatibility behavior is intentionally narrow: transcript artifacts become first-write-wins for repeated same-session recaptures, while summary/compaction/manifest artifacts still refuse same-path different-body mutation.

### DIFF
--- a/platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
+++ b/platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
@@ -783,9 +783,16 @@ fn write_immutable_artifact(
         let current = read_string(&existing_frontmatter, "content_sha256");
         let next = read_string(&frontmatter, "content_sha256");
         if current == next {
+            let existing_body = normalize_markdown_body(&existing_body);
+            upsert_artifact_row(conn, root, &path, &existing_frontmatter, &existing_body)?;
             return Ok(path);
         }
         if normalize_markdown_body(&existing_body) != body {
+            if seed.kind == ArtifactKind::Transcript {
+                let existing_body = normalize_markdown_body(&existing_body);
+                upsert_artifact_row(conn, root, &path, &existing_frontmatter, &existing_body)?;
+                return Ok(path);
+            }
             return Err(format!(
                 "refusing to mutate immutable artifact {}",
                 path.display()
--- a/platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
+++ b/platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
@@ -2245,6 +2245,12 @@ pub async fn session_end(
             }
             state.dedup.clear_session_start(&session_key);
             state.dedup.clear(&session_key);
+            warn!(
+                session = %session_key,
+                agent_id = %agent_id,
+                error = %e,
+                "session-end: transcript artifact write failed"
+            );
             return (
                 StatusCode::INTERNAL_SERVER_ERROR,
                 Json(
--- a/platform/daemon-rs/crates/signet-pipeline/tests/memory_lineage.rs
+++ b/platform/daemon-rs/crates/signet-pipeline/tests/memory_lineage.rs
@@ -1,0 +1,174 @@
+use std::fs;
+use std::path::PathBuf;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use rusqlite::Connection;
+use signet_pipeline::memory_lineage::{
+    MemorySentence, SummaryArtifactInput, TranscriptArtifactInput, write_summary_artifact,
+    write_transcript_artifact,
+};
+
+fn temp_root(name: &str) -> PathBuf {
+    let now = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map(|value| value.as_nanos())
+        .unwrap_or(0);
+    let dir = std::env::temp_dir().join(format!(
+        "signet-lineage-{name}-{}-{now}",
+        std::process::id()
+    ));
+    fs::create_dir_all(&dir).expect("create temp root");
+    dir
+}
+
+fn setup_conn() -> Connection {
+    let conn = Connection::open_in_memory().expect("open in-memory db");
+    signet_core::migrations::run(&conn).expect("run migrations");
+    conn
+}
+
+fn transcript_input(transcript: &str) -> TranscriptArtifactInput {
+    TranscriptArtifactInput {
+        agent_id: "default".to_string(),
+        session_id: "race-session".to_string(),
+        session_key: Some("race-session".to_string()),
+        project: Some("/work/signet".to_string()),
+        harness: Some("opencode".to_string()),
+        captured_at: "2026-06-01T01:45:42.236233+00:00".to_string(),
+        started_at: None,
+        ended_at: Some("2026-06-01T01:45:42.236233+00:00".to_string()),
+        transcript: transcript.to_string(),
+    }
+}
+
+fn summary_input(summary: &str) -> SummaryArtifactInput {
+    SummaryArtifactInput {
+        agent_id: "default".to_string(),
+        session_id: "race-session".to_string(),
+        session_key: Some("race-session".to_string()),
+        project: Some("/work/signet".to_string()),
+        harness: Some("opencode".to_string()),
+        captured_at: "2026-06-01T01:45:42.236233+00:00".to_string(),
+        started_at: None,
+        ended_at: Some("2026-06-01T01:45:42.236233+00:00".to_string()),
+        summary: summary.to_string(),
+    }
+}
+
+fn sentence() -> MemorySentence {
+    MemorySentence {
+        text: "Captured Signet session lineage artifact behavior.".to_string(),
+        quality: "ok".to_string(),
+        generated_at: "2026-06-01T01:45:42.236233+00:00".to_string(),
+    }
+}
+
+#[test]
+fn transcript_recapture_with_different_body_returns_existing_artifact() {
+    let conn = setup_conn();
+    let root = temp_root("transcript-recapture");
+
+    let first = write_transcript_artifact(
+        &conn,
+        &root,
+        transcript_input("User: first idle capture\nAssistant: ok"),
+    )
+    .expect("write first transcript");
+    let first_content = fs::read_to_string(root.join(&first.artifact_path))
+        .expect("read first transcript artifact");
+    conn.execute(
+        "DELETE FROM memory_artifacts WHERE source_path = ?1",
+        [&first.artifact_path],
+    )
+    .expect("simulate DB index missing after artifact file write");
+
+    let second = write_transcript_artifact(
+        &conn,
+        &root,
+        transcript_input("User: first idle capture\nUser: later message\nAssistant: ok"),
+    )
+    .expect("recapture should return existing transcript artifact");
+    let second_content = fs::read_to_string(root.join(&second.artifact_path))
+        .expect("read recaptured transcript artifact");
+
+    assert_eq!(second.artifact_path, first.artifact_path);
+    assert_eq!(second_content, first_content);
+    assert!(second_content.contains("first idle capture"));
+    assert!(!second_content.contains("later message"));
+    let indexed_content: String = conn
+        .query_row(
+            "SELECT content FROM memory_artifacts WHERE source_path = ?1",
+            [&first.artifact_path],
+            |row| row.get(0),
+        )
+        .expect("recapture should repair missing memory_artifacts row");
+    assert!(indexed_content.contains("first idle capture"));
+    assert!(!indexed_content.contains("later message"));
+
+    fs::remove_dir_all(root).ok();
+}
+
+#[test]
+fn identical_transcript_recapture_repairs_missing_artifact_index() {
+    let conn = setup_conn();
+    let root = temp_root("transcript-identical-recapture");
+
+    let first = write_transcript_artifact(
+        &conn,
+        &root,
+        transcript_input("User: same idle capture\nAssistant: ok"),
+    )
+    .expect("write first transcript");
+    conn.execute(
+        "DELETE FROM memory_artifacts WHERE source_path = ?1",
+        [&first.artifact_path],
+    )
+    .expect("simulate DB index missing after artifact file write");
+
+    let second = write_transcript_artifact(
+        &conn,
+        &root,
+        transcript_input("User: same idle capture\nAssistant: ok"),
+    )
+    .expect("identical recapture should return existing transcript artifact");
+
+    assert_eq!(second.artifact_path, first.artifact_path);
+    let indexed_content: String = conn
+        .query_row(
+            "SELECT content FROM memory_artifacts WHERE source_path = ?1",
+            [&first.artifact_path],
+            |row| row.get(0),
+        )
+        .expect("identical recapture should repair missing memory_artifacts row");
+    assert!(indexed_content.contains("same idle capture"));
+
+    fs::remove_dir_all(root).ok();
+}
+
+#[test]
+fn summary_collision_with_different_body_still_refuses_mutation() {
+    let conn = setup_conn();
+    let root = temp_root("summary-collision");
+
+    write_summary_artifact(
+        &conn,
+        &root,
+        summary_input("# Summary\n\nFirst body."),
+        sentence(),
+    )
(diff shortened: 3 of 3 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `8cdeddc7`.

platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
```diff
@@ -786,6 +786,9 @@ fn write_immutable_artifact(
             return Ok(path);
         }
         if normalize_markdown_body(&existing_body) != body {
+            if seed.kind == ArtifactKind::Transcript {
+                return Ok(path);
```
PR-Reviewer-Ant, line 794: This idempotent transcript branch returns before `upsert_artifact_row`, which means it can acknowledge a transcript recapture even if the canonical file exists but the `memory_artifacts` row was never written or is stale. That state is reachable because the normal path writes/renames the file first and only then upserts the DB row; if the earlier call failed during the DB upsert, the next same-path recapture with a different body now returns success without repairing the index. The session-end 500 goes away, but the transcript can still be missing from projection/reindex-backed lookup until some later explicit reindex happens. For first-write-wins, this should reload the existing artifact frontmatter/body and upsert the existing canonical content before returning.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `77601851`

</details>

I found one data-integrity edge case in the new idempotent transcript path. The behavior is mostly aligned with the PR description, but the early return can turn a previously failed partial write into a successful response without repairing the artifact index.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed branch in write_immutable_artifact returns before upsert_artifact_row when an existing transcript file has a different body. Because write_atomic happens before upsert_artifact_row in the normal path, a prior failure after the file rename but before the DB upsert can leave exactly this state; I did not run a runtime repro of that partial-failure sequence.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `5dd1a988`

</details>

The daemon-rs transcript idempotency change itself matches the stated behavior, but the latest diff also pulls in an unrelated 0.138.7 release/version bump for #806. That is outside this PR’s stated scope and touches generated/publish artifacts, so it should be split out or justified before merge.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR description is about transcript artifact recapture behavior in daemon-rs, while the incremental diff updates CHANGELOG.md and package versions across publishable packages to 0.138.7 for a different fix, explicitly described as summary_jobs migration columns (#806). That mismatch is visible directly in the provided diff.


- dist/signetai/package.json:3 - This PR is scoped to daemon-rs transcript recapture idempotency, but this change bumps the published dist package version to 0.138.7 as part of an unrelated release for #806. Because `dist/signetai` is a generated/publish artifact, carrying this here can accidentally ship or tag the wrong release contents and makes this PR’s artifact behavior hard to review. Please split the 0.138.7 release/version bump out of this PR, or make the PR description and validation explicitly cover the release artifact changes.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs
NicholaiVogel: Fixed in `aad875db`. The transcript first-write-wins path now reloads the existing canonical artifact body/frontmatter and calls `upsert_artifact_row` before returning, so a file-present/index-missing partial write is repaired instead of silently acknowledged.

I also tightened `platform/daemon-rs/crates/signet-pipeline/tests/memory_lineage.rs` to delete the `memory_artifacts` row after the first transcript write, recapture with a different body, and assert the index is restored with the original canonical content.

Rebased onto latest `main`; the PR diff is now limited to the transcript recapture files.

NicholaiVogel:
Updated PR #808 after review feedback and rebased onto latest `main`.

Current head: `aad875db`

Changes since review:
- Repaired the transcript idempotency branch so it re-upserts the existing canonical artifact into `memory_artifacts` before returning.
- Expanded the regression test to simulate file-present/index-missing partial write state.
- Rebased away the unrelated release/version diff; current PR files are only the daemon hook, pipeline lineage writer, and lineage integration test.

Local checks rerun:
- `cd platform/daemon-rs && cargo fmt`
- `cd platform/daemon-rs && cargo test -p signet-pipeline --test memory_lineage`
- `cd platform/daemon-rs && cargo test -p signet-pipeline memory_lineage`
- `cd platform/daemon-rs && cargo test -p signet-daemon session_end_persists_transcript_artifact_and_queues_summary`
- `cd platform/daemon-rs && git diff --check`

Ready for re-review.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `aad875db`

</details>

The transcript recapture data-integrity issue from the earlier review appears addressed: the idempotent transcript path now repairs the missing `memory_artifacts` row before returning. I found a new scope problem in the latest diff: this PR now changes the production bundle installer behavior, which is unrelated to the daemon-rs transcript fix described here.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The current diff explicitly changes `deploy/bundle/install.sh` and install documentation, including default setup/start behavior, while the PR description remains scoped to Rust daemon transcript artifact idempotency. The earlier transcript index concern is directly addressed by the new `upsert_artifact_row` call in `memory_lineage.rs` before the transcript idempotent return.


- deploy/bundle/install.sh:1028 - This is a production installer behavior change in a PR whose stated scope is daemon-rs transcript recapture idempotency. The new default path launches an interactive setup wizard when `/dev/tty` is available, otherwise skips setup and may skip daemon start until `agent.yaml` exists; that is a real install-flow change, not just supporting diagnostics for the session-end bug. Please split this into its own installer PR or update the PR scope and validation accordingly.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

NicholaiVogel:
Rebased again onto latest `main` (`b7f5176bc`, release 0.138.8) and force-pushed.

Current head: `8cdeddc`

The latest PR diff remains scoped to the three transcript recapture files only:
- `platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs`
- `platform/daemon-rs/crates/signet-pipeline/src/memory_lineage.rs`
- `platform/daemon-rs/crates/signet-pipeline/tests/memory_lineage.rs`

Local checks rerun after this rebase:
- `bun scripts/version-sync.ts --check` -> all versions aligned at 0.138.8
- `cd platform/daemon-rs && cargo test -p signet-pipeline --test memory_lineage`
- `cd platform/daemon-rs && cargo test -p signet-daemon session_end_persists_transcript_artifact_and_queues_summary`
- `cd platform/daemon-rs && git diff --check`

The installer/version findings were stale merge-base artifacts from `main` moving under the PR, not changes in this branch. Ready for re-review on `8cdeddc`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `8cdeddc7`

</details>

I found one remaining data-integrity hole in the idempotent transcript path: it repairs the missing artifact index only when the recaptured body differs, but the same partial-write state can still be acknowledged without reindexing when the body is identical.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The early `current == next` return in `write_immutable_artifact` still happens before any `upsert_artifact_row` call, while the new repair branch only runs after `normalize_markdown_body(&existing_body) != body`. The added regression test covers file-present/index-missing recapture with a different transcript body, but not the identical-body recapture that follows the same partial write failure mode.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

NicholaiVogel:
Updated for the latest review on identical-body recapture.

Current head: `1329986`

Change:
- The `current == next` retry path now re-upserts the existing canonical artifact row before returning, so identical-body retries repair the same file-present/index-missing partial write state.
- Added `identical_transcript_recapture_repairs_missing_artifact_index` to cover that exact case.

Local checks rerun:
- `cd platform/daemon-rs && cargo fmt`
- `cd platform/daemon-rs && cargo test -p signet-pipeline --test memory_lineage`
- `cd platform/daemon-rs && cargo test -p signet-daemon session_end_persists_transcript_artifact_and_queues_summary`
- `cd platform/daemon-rs && git diff --check`

Ready for re-review on `1329986`.

PR #810, opened 2026-06-01
fix(daemon-rs): embed memory writes

### LINKED ISSUE #804: Rust daemon never embeds stored memories (write-side embedding unimplemented) — vector search silently non-functional
### Environment
- signetai: `0.138.5`
- Daemon: Rust daemon (`daemon-rs`, `signet-daemon` binary)
- Node.js: `v26.0.0`
- OS: macOS 26.4.1 (build 25E253), arm64
- SQLite: `3.51.0`
- DB schema version (`schema_migrations`): `77`
- Embedding provider: `ollama` (`nomic-embed-text`, 768 dims) — verified healthy

### Summary
With the Rust daemon, memories stored via `memory_store` / `POST /api/memories` are **never
embedded**. Vector/semantic search therefore has nothing to match against and silently returns
nothing; only keyword search works. The embedding provider is correctly configured and healthy —
the write-side embedding is simply not implemented in the Rust daemon.

### Evidence

**1. Provider is healthy.** `GET /api/embeddings/status`:
```json
{
  "available": true,
  "base_url": "http://localhost:11434",
  "dimensions": 768,
  "model": "nomic-embed-text",
  "provider": "ollama"
}
```

**2. Memories store fine, but zero embeddings are ever written.**
After storing several memories:
```bash
sqlite3 ~/.agents/memory/memories.db "SELECT COUNT(*) FROM memories;"      # 5
sqlite3 ~/.agents/memory/memories.db "SELECT COUNT(*) FROM embeddings;"     # 0
```
Semantic recall (e.g. query "Product manager portfolio tools" against a stored memory
"User is a PM at working on ... portfolio tracking ...") returns nothing, while exact
keyword recall finds it.

**3. The store route does not embed or enqueue embedding.**
`platform/daemon-rs/crates/signet-daemon/src/routes/memory.rs` (the `/api/memories` handler) has
no embedding call and no enqueue to any embedding worker. The embedding provider (`state.embedding`)
is only consumed on the **query** side:
- `platform/daemon-rs/crates/signet-daemon/src/routes/search.rs:138` — `.embed(...)` for the query vector
- `platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs:1402` — `.embed(...)` for prompt-submit recall scoring

**4. The re-embed / backfill endpoint is a stub.**
`platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs:281-300`:
```rust
/// POST /api/repair/re-embed — re-compute embeddings for gaps.
pub async fn re_embed(
    State(_state): State<Arc<AppState>>,
    Json(_body): Json<Option<ReEmbedBody>>,
) -> impl IntoResponse {
    // TODO: Wire to embedding provider for re-embedding.
    (
        StatusCode::NOT_IMPLEMENTED,
        Json(repair_result("re_embed", false, 0, "re-embedding requires embedding provider integration")),
    )
}
```
The `ReEmbedBody` struct is even annotated
`#[allow(dead_code)] // Fields used when embedding provider integration is wired`
(`repair.rs:266`).

**5. The gap report references a table that doesn't exist in this schema.**
`platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs:227` (`embedding_gaps`) queries
`memory_embeddings`, but schema v77 has no `memory_embeddings` table:
```bash
sqlite3 ~/.agents/memory/memories.db ".tables" | tr ' ' '\n' | grep -i embed
# embeddings
# vec_embeddings  (+ vec_embeddings_* virtual-table shadow tables)
# (no memory_embeddings)
```
So `GET /api/repair/embedding-gaps` returns `total_memories: 0` even when memories exist, because
the `LEFT JOIN memory_embeddings` is against a missing table / different scope.

### Impact
- Semantic/vector recall is silently non-functional under the Rust daemon.
- `IDENTITY.md` / docs that advertise "vector + keyword hybrid" search are effectively
  keyword-only.
- No config change can fix it; the write-side code path is absent.

### Reproduction
1. Run the Rust daemon (`signet-daemon`) at 0.138.5 with `embedding.provider: ollama` and Ollama
   running with `nomic-embed-text` pulled.
2. Store a few memories via `memory_store` / `POST /api/memories`.
3. `sqlite3 <db> "SELECT COUNT(*) FROM embeddings;"` → `0`.
4. Run a semantic query that should match on meaning but not exact keywords → no results.

### Expected
Storing a memory should produce an embedding (inline on write, or via a background embedding
worker like the JS daemon's `embeddingTracker`), so vector search can match it.

### Suggested fix
- Wire the embedding provider into the memory-store path (or a background sweep) in the Rust daemon,
  mirroring the JS daemon's `embeddingTracker`.
- Implement `re_embed` / backfill so existing rows can be embedded after the fact.
- Reconcile the `embedding_gaps` query with the actual schema (`embeddings` vs `memory_embeddings`)
  so the gap report is accurate.
- Until write-side embedding exists, surface a clear degraded-capability signal (status/doctor)
  instead of silently returning empty vector results.

NicholaiVogel:
## Summary

Fixes the Rust daemon memory write path so stored memories get vector embeddings when an embedding provider is configured. This is rebased onto current `main` and preserves the current structured `remember` passthrough path.

Closes #804

## Changes

- Added shared Rust daemon helpers for preparing and upserting memory embeddings with `agent_id`.
- Embedded new `remember` writes, including structured passthrough writes and chunked writes, and returned explicit `embedded` / `embedded_count` response fields.
- Extended core embedding upsert to persist `agent_id` and keep `vec_embeddings` synchronized.
- Re-read repair re-embed candidates inside the write transaction so stale async embeddings are skipped after concurrent modify/delete.
- Fixed `/api/repair/resync-vec` to rebuild the shared vector table from all stored embeddings instead of dropping non-memory vector rows.
- Added focused regressions for write-side embedding, structured writes, stale re-embed candidates, scoped repair/document embedding identity, raw-hash gap detection, shared vector resync, and core vector sync.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

N/A — no migrations touched.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Additional focused Rust checks run:

- [x] `cargo test -p signet-core --test embedding_upsert`
- [x] `cargo test -p signet-daemon remember_embeds_created_memory_rows`
- [x] `cargo test -p signet-daemon remember_structured_payload_embeds_resulting_memory_row`
- [x] `cargo test -p signet-daemon re_embed_skips_stale_candidates_after_modify_or_delete`
- [x] `cargo test -p signet-daemon re_embed_scopes_embedding_hash_by_agent_and_memory`
- [x] `cargo test -p signet-daemon re_embed_gap_detection_ignores_legacy_raw_hash_for_other_memories`
- [x] `cargo test -p signet-daemon resync_vec_preserves_non_memory_embeddings`
- [x] `cargo check -p signet-daemon`
- [x] `cargo test -p signet-daemon`
- [x] `cargo test -p signet-pipeline`
- [x] `cargo test -p signet-pipeline process_document_scopes_embedding_hash_by_agent_and_memory`
- [x] `git diff --check`

## AI disclosure

## Notes

Existing API docs already describe `/api/repair/re-embed`, `/api/repair/embedding-gaps`, and related repair endpoints as real operations. `docs/API.md` does not enumerate `remember` response fields, so no docs source change was required for the `embedded` response correction.

### DIFF
--- a/platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
+++ b/platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
@@ -13,6 +13,7 @@ use axum::{
     http::{StatusCode, header},
     response::{IntoResponse, Json, Response},
 };
+use rusqlite::OptionalExtension;
 use serde::Deserialize;
 use signet_core::queries::embedding::{self, InsertEmbedding};
 use signet_core::queries::memory;
@@ -21,6 +22,7 @@ use tokio::io::AsyncReadExt;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 
+use crate::routes::memory_embeddings::memory_embedding_hash;
 use crate::state::AppState;
 
 // ---------------------------------------------------------------------------
@@ -675,6 +677,13 @@ struct UnembeddedMemory {
     content_hash: Option<String>,
 }
 
+struct FreshMemoryForEmbedding {
+    content: String,
+    normalized_content: String,
+    content_hash: String,
+    agent_id: String,
+}
+
 fn count_active_memories(conn: &rusqlite::Connection) -> rusqlite::Result<i64> {
     conn.query_row(
         "SELECT COUNT(*) FROM memories WHERE COALESCE(is_deleted, 0) = 0",
@@ -690,12 +699,6 @@ fn count_unembedded_memories(conn: &rusqlite::Connection) -> rusqlite::Result<i6
            AND NOT EXISTS (
              SELECT 1 FROM embeddings e
              WHERE e.source_type = 'memory' AND e.source_id = m.id
-           )
-           AND NOT EXISTS (
-             SELECT 1 FROM embeddings e
-             WHERE e.source_type = 'memory'
-               AND m.content_hash IS NOT NULL
-               AND e.content_hash = m.content_hash
            )",
         [],
         |row| row.get(0),
@@ -714,12 +717,6 @@ fn list_unembedded_memories(
              SELECT 1 FROM embeddings e
              WHERE e.source_type = 'memory' AND e.source_id = m.id
            )
-           AND NOT EXISTS (
-             SELECT 1 FROM embeddings e
-             WHERE e.source_type = 'memory'
-               AND m.content_hash IS NOT NULL
-               AND e.content_hash = m.content_hash
-           )
          ORDER BY m.created_at ASC
          LIMIT ?1",
     )?;
@@ -733,6 +730,56 @@ fn list_unembedded_memories(
     .collect()
 }
 
+fn fresh_memory_for_embedding(
+    conn: &rusqlite::Connection,
+    memory: &UnembeddedMemory,
+) -> Result<Option<FreshMemoryForEmbedding>, signet_core::CoreError> {
+    let candidate_normalized = normalize_and_hash(&memory.content);
+    let candidate_hash = memory
+        .content_hash
+        .as_deref()
+        .filter(|value| !value.trim().is_empty())
+        .unwrap_or(&candidate_normalized.hash);
+    let row = conn
+        .query_row(
+            "SELECT content,
+                    content_hash,
+                    COALESCE(NULLIF(agent_id, ''), 'default') AS agent_id
+             FROM memories
+             WHERE id = ?1
+               AND COALESCE(is_deleted, 0) = 0",
+            rusqlite::params![memory.id],
+            |row| {
+                Ok((
+                    row.get::<_, String>(0)?,
+                    row.get::<_, Option<String>>(1)?,
+                    row.get::<_, String>(2)?,
+                ))
+            },
+        )
+        .optional()?;
+
+    let Some((content, content_hash, agent_id)) = row else {
+        return Ok(None);
+    };
+    let current_normalized = normalize_and_hash(&content);
+    let current_hash = content_hash
+        .as_deref()
+        .filter(|value| !value.trim().is_empty())
+        .unwrap_or(&current_normalized.hash);
+    if current_hash != candidate_hash {
+        return Ok(None);
+    }
+    let content_hash = current_hash.to_string();
+
+    Ok(Some(FreshMemoryForEmbedding {
+        content,
+        normalized_content: current_normalized.normalized,
+        content_hash,
+        agent_id,
+    }))
+}
+
 fn write_embedding_batch(
     conn: &rusqlite::Connection,
     results: &[(UnembeddedMemory, Vec<f32>)],
@@ -741,12 +788,9 @@ fn write_embedding_batch(
     let now = chrono::Utc::now().to_rfc3339();
     let mut written = 0;
     for (memory, vector) in results {
-        let normalized = normalize_and_hash(&memory.content);
-        let content_hash = memory
-            .content_hash
-            .as_deref()
-            .filter(|value| !value.trim().is_empty())
-            .unwrap_or(&normalized.hash);
+        let Some(fresh) = fresh_memory_for_embedding(conn, memory)? else {
+            continue;
+        };
 
         if memory
             .content_hash
@@ -761,7 +805,7 @@ fn write_embedding_batch(
                        AND COALESCE(is_deleted, 0) = 0
                        AND id <> ?2
                      LIMIT 1",
-                    rusqlite::params![content_hash, memory.id],
+                    rusqlite::params![fresh.content_hash, memory.id],
                     |row| row.get::<_, String>(0),
                 )
                 .ok();
@@ -771,22 +815,25 @@ fn write_embedding_batch(
                      SET content_hash = ?1,
                          normalized_content = COALESCE(normalized_content, ?2)
                      WHERE id = ?3 AND content_hash IS NULL",
-                    rusqlite::params![content_hash, normalized.normalized, memory.id],
+                    rusqlite::params![fresh.content_hash, fresh.normalized_content, memory.id],
                 )?;
             }
         }
 
-        embedding::delete_by_source(conn, "memory", &memory.id, Some(content_hash))?;
+        let embedding_hash =
+            memory_embedding_hash(&fresh.agent_id, &memory.id, &fresh.content_hash);
+        embedding::delete_by_source(conn, "memory", &memory.id, Some(&embedding_hash))?;
         embedding::upsert(
             conn,
             &InsertEmbedding {
                 id: &uuid::Uuid::new_v4().to_string(),
-                content_hash,
+                content_hash: &embedding_hash,
                 vector,
                 source_type: "memory",
                 source_id: &memory.id,
-                chunk_text: &memory.content,
+                chunk_text: &fresh.content,
                 now: &now,
+                agent_id: Some(&fresh.agent_id),
             },
         )?;
         if let Some(model) = embedding_model {
@@ -1270,9 +1317,7 @@ pub async fn resync_vec(State(state): State<Arc<AppState>>) -> impl IntoResponse
             conn.execute("DELETE FROM vec_embeddings", [])?;
             let count = conn.execute(
                 "INSERT INTO vec_embeddings(id, embedding)
-                 SELECT e.id, e.vector FROM embeddings e
-                 JOIN memories m ON m.id = e.source_id
-                 WHERE e.source_type = 'memory' AND m.is_deleted = 0",
+                 SELECT e.id, e.vector FROM embeddings e",
                 [],
             )?;
             Ok(serde_json::json!(count))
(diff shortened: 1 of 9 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `a8dbf13d`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e63509fe`

</details>

I found one correctness gap in the new write-side embedding path: regular and chunked `remember` writes are covered, but structured memory writes still bypass the new embedding helper. That means the PR does not fully deliver the stated “memory write path” fix for all writes handled by this route.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff shows embedding preparation and upsert only in the non-structured single-memory path and the oversized chunk path in platform/daemon-rs/crates/signet-daemon/src/routes/write.rs. The same handler accepts a structured payload via RememberBody.structured and explicitly skips chunking for structured input, but the changed code does not show any structured-specific embedding path, so structured writes can still store memories without embeddings despite a configured provider.

platform/daemon-rs/crates/signet-daemon/src/routes/write.rs
```diff
@@ -897,6 +920,8 @@ pub async fn remember(
         };
```
PR-Reviewer-Ant, line 2067: This prepares one embedding from the top-level `content`, but the route also accepts `structured` writes. Structured payloads are excluded from chunking above, and this PR only upserts an embedding for the single `transactions::ingest` result below. If structured input expands into one or more stored memory rows through the existing transaction path, those rows still do not get per-memory embeddings. That leaves a hole in the PR's stated fix for the Rust daemon memory write path: a configured embedding provider still won't cover all `remember` writes. Please either embed each structured-created memory row or explicitly reject/route structured writes through a path that does.

NicholaiVogel: Addressed in `e0b17c480`. I checked the Rust route and it does not currently implement the TypeScript daemon's structured passthrough path; it only accepted `structured` while storing the top-level content. To avoid silently creating a memory row without structured parity/embedding coverage, the Rust daemon now fails closed with `501` for structured remember payloads before any write happens. Added `routes::write::tests::remember_rejects_structured_payloads_until_rust_parity_exists` to prove no memory row is created, and reran the focused embedding/backfill checks plus `cargo check -p signet-daemon`.

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
```diff
@@ -263,38 +271,166 @@ pub async fn embedding_gaps(State(state): State<Arc<AppState>>) -> Json<serde_js
...
+    };
+
+    if body.dry_run {
+        return (
+            StatusCode::OK,
+            Json(repair_result(
+                "re_embed",
+                true,
+                candidates.len(),
+                &format!("{} memory embedding(s) would be repaired", candidates.len()),
```
PR-Reviewer-Ant: This repair path reads candidate rows, spends time embedding them outside the write transaction, and then upserts using the stale candidate without re-checking that the memory is still active and still has the same content/hash. A concurrent `forget` can leave a fresh vec row for a deleted memory, and a concurrent `modify` can stamp an embedding for old content onto the current memory. Before `upsert_memory_embedding`, re-read the memory inside the write closure and skip unless `is_deleted = 0` and the current `content_hash` or normalized content still matches the candidate.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e0b17c48`

</details>

I found one data-integrity race in the new Rust re-embed repair path. The write-side remember embedding path looks aligned with the PR goal, and the prior structured-payload concern appears addressed by the new fail-closed response.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The re-embed handler explicitly reads candidates, performs async embedding work outside the database write, then blindly upserts by the stale candidate id/content/hash in platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs. That is enough to prove a concurrent modify/delete can leave an embedding for content that is no longer the current active memory.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
NicholaiVogel: Addressed in `24d886bad`. The re-embed write path now re-reads the memory inside the write closure before upserting, skips if the row is deleted, and skips if the current hash (or normalized current content when the hash is missing) no longer matches the candidate. The upsert uses the current row's content/hash/agent id after that check, so an embedding computed from a stale candidate cannot stamp a modified or deleted memory. Added `routes::repair::tests::re_embed_skips_stale_candidates_after_modify_or_delete` and reran the focused repair/write/core checks plus `cargo check -p signet-daemon`.

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
```diff
@@ -304,15 +495,29 @@ pub async fn resync_vec(State(state): State<Arc<AppState>>) -> impl IntoResponse
     let result = state
         .pool
         .write(signet_core::db::Priority::Low, |conn| {
```
PR-Reviewer-Ant, line 1304: This wipes the entire `vec_embeddings` table, but the rebuild below only inserts embeddings that join to active `memories`. Since the current schema uses shared `embeddings` / `vec_embeddings`, this repair will delete vector index entries for non-memory sources and leave them unrecoverable from this endpoint. Please scope the delete to memory embedding ids, or rebuild every supported source type when doing a full-table reset.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `24d886ba`

</details>

I found one data-integrity issue in the Rust repair path: the vec resync endpoint now clears the whole vector table but only rebuilds memory rows.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed resync implementation explicitly deletes every row from vec_embeddings, then repopulates only embeddings joined to memories with source_type = 'memory'. That is directly visible in platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs and can drop vector index rows for any non-memory embeddings stored in the shared embeddings/vec_embeddings schema.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
NicholaiVogel: Carried forward in the rebased `9031c193b`.

`write_embedding_batch` now re-reads the memory row inside the DB write before upserting. It skips the candidate if the memory was deleted or if the current content hash no longer matches the content that was embedded outside the write lock. Added `routes::repair::tests::re_embed_skips_stale_candidates_after_modify_or_delete` and reran the full `cargo test -p signet-daemon` suite.

platform/daemon-rs/crates/signet-daemon/src/routes/write.rs
NicholaiVogel: Addressed in `9031c193b` after rebasing onto current `main`.

Correction to my earlier reply: current `main` now has structured `remember` passthrough, so this branch no longer fails closed for structured payloads. The write path prepares an embedding for the submitted content and upserts it for the resulting memory row after `transactions::ingest`; structured passthrough behavior is preserved and the response reports `embedded: true` when the embedding is written.

Added `routes::write::tests::remember_structured_payload_embeds_resulting_memory_row` and reran `cargo test -p signet-daemon`.

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
NicholaiVogel: Addressed in `9031c193b`.

`/api/repair/resync-vec` now rebuilds `vec_embeddings` from all rows in the shared `embeddings` table, rather than clearing the table and repopulating only active memory rows. That preserves non-memory embedding vectors while still doing a full vector-table rebuild. Added `routes::repair::tests::resync_vec_preserves_non_memory_embeddings` and reran `cargo test -p signet-daemon`.

platform/daemon-rs/crates/signet-pipeline/src/document.rs
```diff
@@ -465,6 +465,7 @@ async fn persist_document_chunks(
                         source_id: &linked_memory_id,
                         chunk_text: &chunk.chunk.text,
                         now: &ts,
+                        agent_id: Some("default"),
```
PR-Reviewer-Ant: This new `agent_id: Some("default")` stamps document-linked embeddings with the default agent even though document ingest is a scoped data path. That weakens the PR's agent-id embedding fix and can misattribute embeddings for non-default agents, which matters once search/repair/reporting starts relying on `embeddings.agent_id`. Please thread the actual document or linked-memory agent id into this upsert instead of hardcoding `default`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `9031c193`

</details>

I found one scoping issue in the embedding metadata path: one caller now stamps document-linked embeddings as `default` instead of carrying the real agent context. The rest of the changed memory write and repair paths are covered by targeted tests and line up with the PR description.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The changed call in platform/daemon-rs/crates/signet-pipeline/src/document.rs explicitly passes `agent_id: Some("default")`, while the PR's stated goal is to persist embeddings with `agent_id` and the repository convention requires threading real agent ids through scoped data writes. I did not have the surrounding full document-ingest flow in the diff to prove every runtime entry point is non-default, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-pipeline/src/document.rs
NicholaiVogel: Addressed in `96fdbf355`.

The document worker now threads `agentId` / `agent_id` from the document job payload into chunk memory creation and same-agent chunk dedupe. Before embedding upsert, it reads the linked memory row's `agent_id` and stamps `embeddings.agent_id` with that value instead of a hardcoded `default`.

The document ingest regression now uses `agentId: "agent-doc"` and asserts both the chunk memories and linked embeddings are stamped with `agent-doc`.

Validation:
- `cargo test -p signet-pipeline process_document_persists_chunks_and_marks_document_indexed`
- `cargo test -p signet-pipeline`
- `cargo check -p signet-daemon`
- `git diff --check`

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
```diff
@@ -771,22 +826,23 @@ fn write_embedding_batch(
...
                          normalized_content = COALESCE(normalized_content, ?2)
                      WHERE id = ?3 AND content_hash IS NULL",
-                    rusqlite::params![content_hash, normalized.normalized, memory.id],
+                    rusqlite::params![fresh.content_hash, fresh.normalized_content, memory.id],
                 )?;
             }
         }
 
-        embedding::delete_by_source(conn, "memory", &memory.id, Some(content_hash))?;
+        embedding::delete_by_source(conn, "memory", &memory.id, Some(&fresh.content_hash))?;
```
PR-Reviewer-Ant: This repaired embedding still uses the raw memory content hash as the `embeddings.content_hash` conflict key. Since `embedding::upsert` is `ON CONFLICT(content_hash)`, re-embedding the same normalized memory content for another agent can update the existing row to the later `source_id` / `agent_id`, effectively stealing the embedding from the first agent. The write path added `memory:{agent_id}:{memory_id}:{memory_hash}` for this reason; the repair path needs the same scoped identity before the new `agent_id` field is actually safe.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `96fdbf35`

</details>

I found one data-integrity issue in the repair re-embed path: the new agent metadata is stored, but the embedding conflict key still ignores agent scope there.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The write-side remember helper namespaces memory embedding hashes with agent_id, memory_id, and memory_hash, but repair.rs still upserts repaired memory embeddings using only the memory content hash. Because embedding::upsert conflicts on content_hash, two agents with identical memory content can overwrite each other's embedding row during repair.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
NicholaiVogel: Addressed in `c67aa327e`.

The repair re-embed path now uses the same scoped embedding identity as the write path: `memory:{agent_id}:{memory_id}:{memory_hash}`. I moved that formatter into the shared memory embedding helper so write and repair cannot drift again. The repair path still uses the memory row's normalized content hash for stale-candidate validation and memory backfill, but the `embeddings.content_hash` conflict key is scoped before `embedding::upsert`.

Added `routes::repair::tests::re_embed_scopes_embedding_hash_by_agent_and_memory`, which seeds two active memories with the same normalized content hash under different agents and verifies repair writes two separate embedding rows with scoped hashes and no raw-hash embedding row.

Validation:
- `cargo test -p signet-daemon re_embed_scopes_embedding_hash_by_agent_and_memory`
- `cargo test -p signet-daemon`
- `cargo check -p signet-daemon`
- `git diff --check`

platform/daemon-rs/crates/signet-pipeline/src/document.rs
```diff
@@ -465,6 +487,7 @@ async fn persist_document_chunks(
                         source_id: &linked_memory_id,
```
PR-Reviewer-Ant, line 496: This now stamps the document chunk embedding with the linked memory's `agent_id`, but the `content_hash` passed to `InsertEmbedding` is still the unscoped `normalized.hash` just above this block. Since `embedding::upsert` is idempotent on `content_hash` and updates `source_id` / `agent_id` on conflict, two agents ingesting identical document chunk text will collide on the same embedding row: the later write can retarget the existing embedding to the later memory/agent and leave the earlier document memory without its own searchable vector. The memory write and repair paths fixed this by including `agent_id` and `memory_id` in the embedding hash; document chunks need the same kind of scoped embedding key now that this PR makes them agent-scoped.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `c67aa327`

</details>

I found one data-integrity issue in the document embedding path: the PR starts stamping document-derived embeddings with `agent_id`, but still uses the unscoped normalized content hash as the unique embedding key.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The changed document path now threads agent_id into document chunk memory rows and embeddings, while the nearby InsertEmbedding call still keys the embedding by the raw normalized chunk hash. Because embedding::upsert is unique on content_hash and updates source_id/agent_id on conflict, identical document chunks across agents can overwrite each other's embedding metadata instead of creating separate scoped embeddings.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-pipeline/src/document.rs
NicholaiVogel: Addressed in `a61350657`.

Document chunk embeddings now use a scoped memory embedding key, `memory:{agent_id}:{memory_id}:{memory_hash}`, instead of the raw normalized chunk hash. The chunk memory row still keeps its normal `content_hash` for same-agent dedupe, but `embeddings.content_hash` no longer collides across agents or memory ids when identical document chunk text is ingested.

Added `document::tests::process_document_scopes_embedding_hash_by_agent_and_memory`, which ingests identical chunk text for two different agents and verifies two distinct memory embedding rows, no raw-hash embedding row, and scoped hashes matching the linked memory rows.

Validation:
- `cargo test -p signet-pipeline process_document_scopes_embedding_hash_by_agent_and_memory`
- `cargo test -p signet-pipeline`
- `cargo check -p signet-daemon`
- `git diff --check`

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
```diff
@@ -771,22 +827,25 @@ fn write_embedding_batch(
                      SET content_hash = ?1,
                          normalized_content = COALESCE(normalized_content, ?2)
                      WHERE id = ?3 AND content_hash IS NULL",
-                    rusqlite::params![content_hash, normalized.normalized, memory.id],
+                    rusqlite::params![fresh.content_hash, fresh.normalized_content, memory.id],
                 )?;
             }
         }
```
PR-Reviewer-Ant, line 821: This switches re-embed writes to a scoped content hash, but the gap detection above still treats any embedding whose `content_hash = m.content_hash` as covering the memory. That legacy raw-hash check can now hide real gaps: if one old/raw embedding exists for a content hash, another memory with the same content hash but a different `id` or `agent_id` will be skipped by `/api/repair/re-embed` even though no embedding exists for its own `source_id` and no scoped `memory:{agent}:{id}:{hash}` row exists. That undercuts the scoped embedding fix and leaves vector recall incomplete for exactly the duplicated/shared-content cases this PR is trying to make safe. The candidate queries should either check the scoped hash for the current row or rely on `source_type/source_id` coverage rather than the raw content hash fallback.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a6135065`

</details>

I found one data-integrity issue in the re-embed path: the new scoped embedding hash scheme is only half-applied, so some memories can still be treated as embedded when their own vector row does not exist.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The diff changes memory embedding writes to use scoped hashes of the form memory:{agent_id}:{memory_id}:{memory_hash}, but the re-embed candidate queries still suppress gaps when any legacy raw content_hash row exists. That directly conflicts with the PR's stated goal of agent/memory-scoped embeddings and is visible in the changed repair path.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon-rs/crates/signet-daemon/src/routes/repair.rs
NicholaiVogel: Addressed in `a8dbf13d`.

The re-embed gap detection no longer treats a raw `embeddings.content_hash = memories.content_hash` row as coverage for every memory with that content hash. Both `count_unembedded_memories` and `list_unembedded_memories` now rely on memory embedding coverage by `source_type = 'memory'` and the current memory's `source_id`, so a legacy/raw hash row for one memory cannot hide another agent/id's missing scoped embedding.

Added `routes::repair::tests::re_embed_gap_detection_ignores_legacy_raw_hash_for_other_memories`, which seeds two agents with the same memory content hash plus one old raw-hash embedding for only the first memory. The gap detector now correctly returns the second memory for backfill.

Validation:
- `cargo test -p signet-daemon re_embed_gap_detection_ignores_legacy_raw_hash_for_other_memories`
- `cargo test -p signet-daemon`
- `cargo check -p signet-daemon`
- `git diff --check`

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a8dbf13d`

</details>

I did not find a concrete correctness, security, data-integrity, or scoping issue in this diff. The implementation matches the PR description: Rust memory writes now prepare embeddings outside the DB transaction, persist scoped memory embeddings on create/chunk/document paths, and repair/resync behavior has focused regression coverage.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed paths directly cover the claimed behavior in platform/daemon-rs/crates/signet-daemon/src/routes/write.rs, routes/repair.rs, routes/memory_embeddings.rs, signet-core embedding upsert, and the document pipeline. The added tests exercise new write-side embedding, structured writes, stale repair candidates, scoped hashes, document scoping, and vec resync. I did not run the Rust test suite or validate a live daemon instance, so runtime confidence stops short of high.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #811, opened 2026-06-02
(fix) disable DeepSeek reasoning for OpenAI-compatible provider

### LINKED ISSUE #809: openai-compatible provider throws "returned empty response" for DeepSeek reasoning models (v4-flash / v4-pro / reasoner) — extraction & synthesis silently fail
## Summary

When the memory pipeline's LLM is configured as `provider: openai-compatible` against DeepSeek (`endpoint: https://api.deepseek.com/v1`) with a **reasoning** model (`deepseek-v4-flash`, `deepseek-v4-pro`, `deepseek-reasoner`), **every extraction and synthesis call fails** with:

```
openai-compatible:deepseek-v4-flash returned empty response
-> All routed targets failed.
-> LLM extraction failed: All routed targets failed.
```

The pipeline produces zero facts/entities while the daemon stays otherwise healthy (closely related to the silent-failure mode reported in #781).

## Affected version

Confirmed present on `main` @ **0.138.8** (`b7f5176b`). The `createOpenAiCompatibleProvider` request/parse path is unchanged from older releases.

## Root cause

DeepSeek's V4/reasoner models stream their output into **`reasoning_content`** and leave `content` **null** until reasoning finishes. `createOpenAiCompatibleProvider.call()` reads only `message.content`:

`platform/daemon/src/pipeline/provider.ts` (current `main`):
```ts
// request body - no reasoning control sent
body: JSON.stringify({
  model: config.model,
  messages: [{ role: "user", content: prompt }],
  max_tokens: opts?.maxTokens ?? 4096,
}),
...
const text = extractOpenAiLikeText(body.choices?.[0]?.message?.content);
if (!text.trim()) {
  throw new Error(`${config.name} returned empty response`);   // <- fires
}
```

Because the memory pipeline passes a **small `max_tokens`** for extraction (strength-derived, ~1024), the model spends the entire budget on `reasoning_content` and returns `content: null` with `finish_reason: "length"` -> `extractOpenAiLikeText` is empty -> "returned empty response".

Notably, the **ollama** provider already handles this - it reads a `thinking` fallback (`generate()` returns `data.thinking` when `response` is empty) and disables thinking via `extraBody: { format: "json", think: false }`. The **openai-compatible** provider has no equivalent.

## Reproduction

```bash
curl -s https://api.deepseek.com/v1/chat/completions \
  -H "Authorization: Bearer $DEEPSEEK_KEY" -H "Content-Type: application/json" \
  -d '{"model":"deepseek-v4-flash","stream":true,"max_tokens":120,
       "messages":[{"role":"user","content":"Extract one fact as JSON: the user prefers Bun."}]}'
```
-> every chunk has `delta.content: null` + `delta.reasoning_content: "..."`; final usage `completion_tokens_details.reasoning_tokens: 120`, `finish_reason: "length"`, content never emitted.

## Suggested fixes (any one resolves it)

1. **Disable reasoning for DeepSeek** (mirrors the ollama `think:false`): inject `thinking: {"type":"disabled"}` into the request body when the endpoint is `api.deepseek.com`. Verified working - DeepSeek then returns `content` directly with no `reasoning_content`. Minimal diff:
   ```ts
   const isDeepSeek = /api\.deepseek\.com/i.test(baseUrl);
   const extraBody: Record<string, unknown> = isDeepSeek ? { thinking: { type: "disabled" } } : {};
   // ...spread ...extraBody into both the non-stream and stream request bodies
   ```
2. **Read `reasoning_content` as a fallback** when `content` is empty (parallel to the ollama `data.thinking` fallback), and/or
3. **Raise `max_tokens`** for detected reasoning models so reasoning + content both fit.

(1) is the smallest and avoids paying reasoning cost/latency on every background extraction call.

## Environment

- Signet running image 0.128.0; bug verified still present on `main` 0.138.8.
- Provider: `openai-compatible` -> `https://api.deepseek.com/v1`, models `deepseek-v4-flash` / `deepseek-v4-pro` (the only models DeepSeek's API exposes for some accounts - both are reasoning models, so there is no non-reasoning DeepSeek fallback).
- Embeddings unaffected (local).

Happy to open a PR with fix (1) if that approach is preferred.

aaf2tbz:
## Summary

Fixes #809 by making the OpenAI-compatible provider disable DeepSeek reasoning for background extraction and synthesis calls. DeepSeek reasoning models can spend the whole completion budget on `reasoning_content` and return `content: null`, which made the provider throw `returned empty response`.

This keeps the existing OpenAI-compatible request path, but adds a DeepSeek endpoint guard that sends `thinking: { type: "disabled" }` for both non-streaming and streaming chat completions.

## Changes

- Detect `api.deepseek.com` OpenAI-compatible endpoints in the daemon provider.
- Add DeepSeek-only `thinking: { type: "disabled" }` to generate and stream request bodies.
- Add regression tests for DeepSeek non-streaming, DeepSeek streaming, and non-DeepSeek gateway behavior.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — daemon provider fix, no UI changes.

## PR Readiness (MANDATORY)

Notes on readiness:

- Spec alignment checked against `docs/specs/INDEX.md`, `docs/specs/dependencies.yaml`, `docs/specs/approved/model-provider-router.md`, and `docs/specs/complete/memory-pipeline-plan.md`.
- Agent scoping is N/A because this PR changes provider request shaping only; it adds no data queries.
- Admin/mutation endpoint security is N/A because this PR does not touch routes or mutation endpoints.
- Docs are N/A because this is not an API, spec, or status surface change.
- Local targeted tests pass. Full lint/typecheck are left unchecked while this is draft because existing unrelated failures remain in the daemon/provider test file and daemon type build.

- [ ] Migration is idempotent

N/A — no migrations.

## Testing

- [x] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Commands run:

- `bun install --frozen-lockfile`
- `bun run build:core`
- `bun install --frozen-lockfile` from `platform/daemon`
- `bun test platform/daemon/src/pipeline/provider.test.ts -t createOpenAiCompatibleProvider`
- `bun build platform/daemon/src/pipeline/provider.ts --target node --external @signet/core --external better-sqlite3 --external sharp --external onnxruntime-node --outfile /tmp/signet-provider-pr809-check.js`
- `bunx biome check platform/daemon/src/pipeline/provider.ts platform/daemon/src/pipeline/provider.test.ts` — fails on pre-existing unrelated `provider.test.ts` lint/format findings outside this PR's added block.

## AI disclosure

## Notes

Draft until the broader local lint/typecheck readiness item is green or explicitly waived.

### DIFF
--- a/platform/daemon/src/pipeline/provider.test.ts
+++ b/platform/daemon/src/pipeline/provider.test.ts
@@ -19,6 +19,7 @@ import {
 	createCodexProvider,
 	createLlamaCppProvider,
 	createOllamaProvider,
+	createOpenAiCompatibleProvider,
 	createOpenCodeProvider,
 	createOpenRouterProvider,
 	resolveDefaultOllamaFallbackMaxContextTokens,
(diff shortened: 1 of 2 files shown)


aaf2tbz:
Opened as draft. Unsure about current direction, will wait for further input before proceeding with implementation.

NicholaiVogel:
Closing in favor of a broader fix — the root cause isn't DeepSeek-specific.

The real issue: `createOpenAiCompatibleProvider` doesn't handle `reasoning_content` at all — not in its TypeScript response types, not in its non-streaming read path (line 1433), not in its streaming SSE parser (line 868). When a reasoning model spends the `max_tokens` budget on thinking and returns `content: null`, the provider throws "returned empty response" because it never looks at `reasoning_content`.

Endpoint sniffing for `api.deepseek.com` to inject `thinking: { type: "disabled" }` papers over the deeper gap: this provider claims OpenAI-compatibility but ignores the reasoning tokens extension that's standard across OpenAI, DeepSeek, Qwen, etc. Tomorrow another provider ships a reasoning model on a different domain and the same bug reappears.

The correct fix:
1. Add `reasoning_content` to the parsed response/delta types
2. Read it as fallback when `content` is null/empty (mirrors what the ollama provider already does with `data.thinking`)
3. Support `thinking: { type: "disabled" }` as a request parameter for callers that want to suppress reasoning spend

Working on this now under #809.

PR #816, opened 2026-06-03
build(install): unify installers on native binary

NicholaiVogel:
## Summary

Unifies Signet install surfaces around one compiled Bun binary artifact:

- curl downloads `native-manifest.json`, verifies the platform binary SHA-256, then runs `signet install`
- npm/Bun global installs publish a thin `signetai` wrapper whose postinstall downloads and verifies the same native binary
- release CI builds/uploads native binaries and the manifest before publishing the wrapper
- compiled runtime now embeds dashboard assets, worker bundles, the transformers web runtime, ONNX WASM, setup templates, and built-in skills instead of depending on source-tree paths or host-global native libraries
- old bundle installer scripts/workflow and Bun installer hash watcher are removed

## Status Before Merge

- Local compiled-runtime asset smoke is green: the compiled daemon serves `/health` and `/` dashboard HTML without `Dashboard not found`, synthesis worker, `Cannot find module`, `libonnxruntime`, or native-provider fallback warnings.
- curl and npm/Bun install surfaces have local smoke coverage that serves a fake native release manifest/binary and verifies both installers use that same manifest-selected binary.
- curl now resolves the latest GitHub release via the releases API so prerelease native assets are included; `SIGNET_RELEASE_TAG`, `SIGNET_VERSION`, and `SIGNET_DOWNLOAD_BASE` remain explicit overrides.
- npm's `signet-mcp` command is a dedicated wrapper entrypoint, so Windows `.cmd` shims do not depend on `process.argv[1]` basename detection.

## PR Readiness (MANDATORY)

Notes: this PR does not add data queries or admin/mutation endpoints; those checklist items were verified as not applicable to the changed installer/runtime packaging surface.

## Verification

- `bun test platform/daemon/src/native-runtime-assets.test.ts`
- `bun test scripts/install-copy-regression.test.ts scripts/native-install-smoke.test.ts scripts/check-publish-manifests.test.ts`
- `sh -n web/marketing/public/install.sh`
- `bunx biome check dist/signetai/bin/launch.js dist/signetai/bin/signet.js dist/signetai/bin/signet-mcp.js dist/signetai/scripts/verify-wrapper.js platform/daemon/src/native-runtime-assets.ts platform/daemon/src/native-runtime-assets.test.ts scripts/build-native-bun.ts scripts/check-publish-manifests.test.ts scripts/install-copy-regression.test.ts scripts/native-install-smoke.test.ts`
- `bun run build:native-bun`
- `./dist/native/signet --help`
- `bun run build` in `dist/signetai`
- `git diff --check`

Earlier branch validation also covered:

- `bun scripts/check-publish-manifests.ts`
- `bun run build:signetai`
- `./dist/native/signet install --bin-dir /tmp/signet-native-bin --force --json`
- `SIGNET_VERSION=0.0.0-test bun scripts/generate-native-manifest.ts`
- `npm pack --dry-run ./dist/signetai`
- compiled daemon smoke with `SIGNET_DAEMON_ENTRYPOINT=1`, isolated `SIGNET_PATH`, and `SIGNET_PORT=3987`; verified `/health`, `/`, and absence of the removed runtime failure signatures


NicholaiVogel:
TODO before merge: the native-binary direction is right, but Docker still needs to be migrated to the new artifact shape.

Current Docker smoke is failing because `deploy/docker/Dockerfile` still copies the old JS package layout:

```text
/app/dist/signetai/dist: not found
/app/dist/signetai/dashboard: not found
/app/dist/signetai/skills: not found
```

If the new canonical runtime is the compiled native binary, Docker should build/copy that binary and the entrypoint should run it directly instead of expecting `dist/signetai/dist/daemon.js` and separate dashboard/skills dirs.

Also worth folding into the same cleanup before merge:

- make the dashboard build dependency explicit before `build:native-bun` on clean CI runners
- either add `linux-arm64` to the native release matrix or make the installer reject it honestly instead of mapping `aarch64` to an asset we never publish
- document/replace the Windows direct install path now that `install.ps1` is gone

Good direction overall, but this should land as a full installer/runtime-surface migration rather than leaving Docker and platform support half on the old model.

NicholaiVogel:
TODO before merge: Docker is fixed now, but the release native build matrix still looks like it can fail on a clean tag checkout.

`build:native-bun` requires `surfaces/dashboard/build/index.html`, but the release `build-native` job currently does:

```yaml
- name: Build workspace dependencies
  run: bun run build

- name: Build native Signet binary
  run: bun run build:native-bun
```

Root `bun run build` does not build the dashboard. I verified locally by removing `surfaces/dashboard/build` and running `bun run build:native-bun`; it fails with:

```text
Dashboard build is missing at .../surfaces/dashboard/build.
Run bun run build:dashboard before build:native-bun.
```

So the release job should explicitly build dashboard assets before compiling native binaries, e.g.:

```yaml
- name: Build dashboard assets
  run: bun run build:dashboard
```

between `bun run build` and `bun run build:native-bun`.

Also worth adding/adjusting a manifest/workflow guard so `build:dashboard` is required before `build:native-bun` in the release matrix. Docker now has that order; release should too.

PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `8c777598`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `501699a5`

</details>

I found a release-channel regression in the new npm publish path. The native binary unification can work, but this workflow change appears to publish every nightly as the default npm `latest` instead of leaving promotion to the promote workflow.

**Confidence:** High [sufficient_diff_evidence, targeted_context_included] - The PR changes the Nightly Release workflow from publishing with an explicit prerelease tag to running `npm publish --access public` with no `--tag`, while the Promote Release workflow now only promotes the GitHub release and no longer updates npm dist-tags. That directly contradicts the workflow naming and the promote workflow input text for promoting a nightly version to stable/latest.


- .github/workflows/release.yml:310 - This drops the explicit npm prerelease tag, so `npm publish` will use npm's default `latest` dist-tag. Since this workflow is still the nightly release pipeline and `promote-release.yml` no longer moves npm dist-tags, every successful push-to-main nightly would become the default `npm install -g signetai` version immediately. That is a breaking release-channel change and it also makes the promote workflow's “Nightly version to promote to stable/latest” contract false for npm. Keep publishing the wrapper under a prerelease tag here, or update the promotion workflow to intentionally own the `latest` dist-tag transition.

.github/workflows/release.yml
```diff
@@ -280,27 +274,27 @@ jobs:
...
-            npm dist-tag add "signetai@${NEW_VERSION}" next
-            npm dist-tag add "@signetai/signet-memory-openclaw@${NEW_VERSION}" next || true
-          else
-            cd dist/signetai && npm publish --tag next --access public
-            cd ../../integrations/openclaw/memory-adapter && npm publish --tag next --access public || true
-          fi
+          test -n "$NODE_AUTH_TOKEN"
+          npm config set //registry.npmjs.org/:_authToken "$NODE_AUTH_TOKEN"
+          cd dist/signetai
+          npm publish --tag next --access public
```
PR-Reviewer-Ant: This publish step is no longer retry-safe. The previous workflow checked whether `signetai@${NEW_VERSION}` already existed and refreshed the `next` tag on reruns; this version always calls `npm publish`. If the workflow publishes `signetai` successfully and then fails while publishing the OpenClaw adapter, editing the release, or due to a transient npm/GitHub issue, rerunning the workflow will fail because npm package versions are immutable. That leaves the release stuck even though the native assets and wrapper may already be public. Please keep the existing-version guard for at least `signetai` and only publish when the version is absent.

PR-Reviewer-Ant (changes requested):
<details>

- Model: `gpt-5.5`
- Commit: `0b877b7d`

</details>

I found one release-pipeline regression: the npm publish step is no longer idempotent, so a retry after a partial publish can strand the release.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The changed release workflow removes the prior `npm view signetai@${NEW_VERSION}` guard and now always runs `npm publish` from `dist/signetai`. That behavior is directly visible in `.github/workflows/release.yml`; I did not run the workflow, so the exact failure mode is inferred from npm's immutable package-version publish behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

.github/workflows/release.yml
```diff
@@ -166,71 +166,63 @@ jobs:
...
+          - os: macos-14
+            platform: darwin-arm64
+            asset: signet-darwin-arm64
           - os: windows-latest
-            target: x86_64-pc-windows-msvc
-            asset: signet-daemon-win32-x64.exe
-          - os: windows-latest
-            target: aarch64-pc-windows-msvc
-            asset: signet-daemon-win32-arm64.exe
+            platform: win32-x64
```
PR-Reviewer-Ant, line 181: This new native release matrix drops Linux ARM64 from the installable artifact set. The deleted bundle workflow built and published `linux-arm64`, but the new matrix only uploads Linux x64, macOS x64/arm64, and Windows x64, and the publish verification mirrors that reduced set. Since this PR replaces the curl/npm install surfaces around `native-manifest.json`, existing Linux ARM64 users now have no binary/manifest entry to install. If that platform is intentionally unsupported, the installer should fail with an explicit unsupported-platform message and the breaking change should be called out; otherwise this matrix needs a `linux-arm64` leg and expected asset checks.

.github/workflows/release.yml
```diff
@@ -280,27 +274,34 @@ jobs:
...
         run: gh release edit "v${NEW_VERSION}" --draft=false
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Publish to npm nightly channel
+      - name: Publish npm wrapper
         run: |
-          echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > "$NPM_CONFIG_USERCONFIG"
+          test -n "$NODE_AUTH_TOKEN"
+          npm config set //registry.npmjs.org/:_authToken "$NODE_AUTH_TOKEN"
```
PR-Reviewer-Ant, line 316: This now writes the npm auth token into the runner's default npm config instead of the previous temp npmrc. On GitHub-hosted `ubuntu-latest` this is mostly contained by runner teardown, but it is still a broader credential lifetime than needed and would be risky if this workflow ever moves to a persistent runner. Keeping `NPM_CONFIG_USERCONFIG` pointed at `${{ runner.temp }}/.npmrc` preserves the old containment while still using `NODE_AUTH_TOKEN`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `efd66d15`

</details>

I found one release-pipeline regression that would drop a previously built installer platform, plus one token-handling cleanup worth considering.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The workflow diff directly shows the old bundle pipeline built linux-arm64 while the new native release matrix and publish verification only include linux-x64, darwin-x64, darwin-arm64, and win32-x64. The prompt includes the release workflow and README, but not the full new installer/manifest generator source, so I cannot confirm whether another path intentionally rejects linux-arm64 before manifest lookup.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

.github/workflows/release.yml
```diff
@@ -166,71 +166,71 @@ jobs:
...
-          - os: windows-latest
-            target: x86_64-pc-windows-msvc
-            asset: signet-daemon-win32-x64.exe
+            platform: linux-x64
+            asset: signet-linux-x64
+          - os: ubuntu-latest
+            platform: linux-arm64
+            asset: signet-linux-arm64
+          - os: macos-13
+            platform: darwin-x64
```
PR-Reviewer-Ant, line 175: This matrix brings back `macos-13` for the x64 macOS native binary, but the deleted bundle workflow in this PR had already replaced that label with the current Intel runner because `macos-13` was retired. If this label is unavailable, `build-native` fails, `publish` is gated on `needs.build-native.result == 'success'`, and no nightly release/npm wrapper goes out. Use the supported Intel macOS runner label here so the required darwin-x64 asset can actually be produced.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `fc2eb905`

</details>

I found one release workflow regression: the new native build matrix reintroduces a macOS runner label that the repository had already documented as retired. That can block nightly releases before the new installer path ever reaches users.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed release workflow explicitly adds `macos-13` for the darwin-x64 native binary, while the removed bundle workflow in this same PR had already moved away from `macos-13` with an inline note that it was retired and should use the current Intel runner label. I did not run the GitHub Actions workflow here.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

.github/workflows/release.yml
```diff
@@ -248,28 +248,32 @@ jobs:
...
         run: bun run build
 
-      - name: Resolve workspace protocols
-        run: bun scripts/version-sync.ts --to "$NEW_VERSION"
-
       - name: Validate publish manifests
         run: bun scripts/check-publish-manifests.ts
 
-      - name: Verify daemon release assets
+      - name: Download native release assets
```
PR-Reviewer-Ant, line 250: This validates publish manifests before `native-manifest.json` exists, then generates/uploads that manifest later without rerunning the publish guard. Since the npm/Bun wrapper is now just a downloader for the native binary, the release job should generate the native manifest first and validate the exact publish payload after that. As written, a broken or missing manifest can slip past `check-publish-manifests.ts` and still publish `signetai`, leaving fresh npm installs unable to select or verify the release binary.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `fc2eb905`

</details>

I found one release-packaging issue that can publish an npm wrapper without the manifest it needs for postinstall downloads. That directly undermines the PR's main claim that npm/Bun installs use the same verified native binary path.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The release workflow runs publish-manifest validation before generating native-manifest.json, and then uploads native-manifest.json only to GitHub Releases. The supplied changed files do not show dist/signetai/package contents or the postinstall downloader implementation, so I cannot prove the wrapper never fetches the release manifest remotely, but the workflow no longer validates the actual manifest file before npm publish.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

.github/workflows/release.yml
```diff
@@ -166,71 +166,71 @@ jobs:
...
+        run: bun run build
+
+      - name: Build dashboard assets
+        run: bun run build:dashboard
+
+      - name: Build native Signet binary
+        run: bun run build:native-bun
+
+      - name: Smoke native binary
+        if: matrix.platform != 'linux-arm64'
```
PR-Reviewer-Ant: This skips the only runtime smoke check for the newly published `signet-linux-arm64` asset. Since the manifest generated later is what curl/npm/Bun installers trust, a bad arm64 binary can still be uploaded, included in `native-manifest.json`, and published to users as a supported platform. Can we build this leg on an arm64 runner or run it under an explicit emulator/smoke harness before the publish job is allowed to undraft the release?

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cc1dacfd`

</details>

I found one release-packaging gap worth tightening: the workflow now publishes a linux-arm64 installer artifact without ever executing it in CI. That is risky on this PR because the installer path is intentionally moving all install surfaces onto these native binaries.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The release matrix explicitly adds a linux-arm64 asset on .github/workflows/release.yml lines 171-173, but the only native binary smoke step is skipped for that platform on lines 207-210. I cannot prove from the provided diff whether build:native-bun cross-compilation itself is broken, but the published artifact is not runtime-validated before manifest generation and npm publication.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

.github/workflows/promote-release.yml
```diff
@@ -18,14 +18,25 @@ jobs:
...
+            "signetai"
+            "@signetai/signet-memory-openclaw"
+          )
           npm view "signetai@${VERSION}" version >/dev/null
-          npm dist-tag add "signetai@${VERSION}" latest
-          npm dist-tag add "@signetai/signet-memory-openclaw@${VERSION}" latest || true
+          for package in "${packages[@]}"; do
+            npm dist-tag add "${package}@${VERSION}" latest || true
+          done
           gh release edit "v${VERSION}" --repo "$GITHUB_REPOSITORY" --prerelease=false --latest
```
PR-Reviewer-Ant, line 36: This now promotes a release even when one of the native platform packages cannot be tagged. The workflow only proves `signetai@${VERSION}` exists, then runs `npm dist-tag add ... || true` for every package and still marks the GitHub release stable/latest. With this PR, those platform packages are part of the install contract, so a missing or unpublished native package should stop promotion rather than being silently ignored. I would keep `@signetai/signet-memory-openclaw` optional if needed, but make the five `@signetai/signetai-*` packages fail closed with an `npm view`/dist-tag check before `gh release edit --latest`.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cfbde86f`

</details>

I found one release-process issue worth tightening: the promotion workflow can mark a release stable even if one of the new native platform packages is missing or failed to tag. The core installer/runtime changes otherwise line up with the PR description from the provided diff.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_cross_module_context] - The promotion behavior is visible directly in .github/workflows/promote-release.yml: it verifies only signetai@VERSION, suppresses every dist-tag failure for platform packages, then edits the GitHub release as latest. I could not inspect the full dist/signetai package.json or wrapper scripts beyond the supplied prompt, so I am not making stronger claims about runtime wrapper behavior.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `cfbde86f`

</details>

I did not find a concrete blocker in the provided diff. The release workflow, Docker packaging change, and documented install-surface claims are internally consistent in the visible files, though the central installer/wrapper source bodies were not included in the prompt.

**Confidence:** Medium [targeted_context_included, missing_cross_module_context, missing_runtime_repro] - The provided changed-file bodies fully cover the release workflow, promotion workflow, Dockerfile, Docker entrypoint, and README updates, and I found no correctness/security issue in those visible changes. The prompt did not include the full changed bodies for web/marketing/public/install.sh, dist/signetai/bin/*.js, or scripts/generate-native-manifest.ts, which are central to the PR's installer claims, so this review cannot independently validate those implementations beyond the diff excerpts and stated verification.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

.github/workflows/release.yml
```diff
@@ -280,27 +280,73 @@ jobs:
...
+        run: bun scripts/check-publish-manifests.ts
+
       - name: Publish release
-        # Undraft only after publish manifests pass, so a broken npm package
-        # cannot slip through behind a green release job.
-        # Assets must still be public before npm publish can expose any
-        # postinstall downloader paths to users.
+        # Undraft after release assets and manifest checks pass. The npm
+        # wrapper optional dependencies resolve to the same native binaries.
         run: gh release edit "v${NEW_VERSION}" --draft=false
```
PR-Reviewer-Ant: This undrafts the GitHub release before any of the npm packages are published. If `npm publish` fails or partially publishes after this point, users can see a public v$NEW_VERSION release and the generated `native-manifest.json`, while `npm install -g signetai` still fails or resolves an older package. That is especially easy to hit now that the README presents curl/npm/Bun as equivalent install surfaces. Consider moving the undraft step after the required npm publishes succeed, or add a compensating failure path that re-drafts/marks the release when publish does not complete.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d809a5f5`

</details>

I found one release-ordering problem in the new native publish pipeline. The binaries and manifest are validated, but the GitHub release is made public before the npm packages that the install docs now advertise have actually published.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The release workflow clearly undrafts the GitHub release at .github/workflows/release.yml:316 and only then runs npm publish at line 320. I could verify the ordering from the supplied changed workflow, but the prompt does not include the full generated package manifests or any downstream release-consumer automation that might intentionally tolerate a public GitHub release without npm packages.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `8c777598`

</details>

I did not find a concrete correctness, security, data integrity, or release-blocking issue in the provided diff and file snapshots. The new flow is internally consistent at the level shown: release builds platform binaries, verifies expected assets, generates the manifest, stages platform npm packages, validates publish manifests, then publishes.

**Confidence:** Medium [targeted_context_included, missing_cross_module_context, missing_runtime_repro] - The review had targeted release, installer, Docker, and README context, but the embedded diff was truncated before several key installer/package files such as scripts/generate-native-manifest.ts, dist/signetai package scripts, and native wrapper logic. I also could not fetch the live PR or run the release/install smoke tests from this read-only, unauthenticated environment.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #819, opened 2026-06-03
fix(install): bundle native assets in npm package

NicholaiVogel:
## Summary

Follow-up to the post-merge release failures from PR #816 and PR #818.

Both publish attempts failed at the first new native platform package:

```text
npm error 404 Not Found - PUT https://registry.npmjs.org/signetai-linux-x64 - Not found
```

The scoped and unscoped package attempts both failed, which means the current npm token can publish the existing `signetai` package but cannot create new package names. This PR removes that requirement entirely.

What changes:

- `signetai` now carries bundled native assets under `native/<platform>/<binary>`.
- The wrapper resolves `native/<platform>/<binary>` directly, so installs still work if postinstall scripts are disabled.
- `postinstall` only links/copies the bundled binary to the legacy `native/signet` path when scripts are allowed.
- Release publishing stages downloaded native release assets into `dist/signetai/native/...` and publishes only the existing `signetai` npm package.
- Promote-release only requires/promotes `signetai` plus the optional OpenClaw adapter.
- Publish checks now fail if the bundled native binaries are missing before npm publish.

Tradeoff: npm/Bun installs will download the bundled assets for all supported platforms. That is heavier than per-platform optional packages, but it works with the npm token we actually have and preserves scripts-disabled installs.

## Verification

- `bun install --frozen-lockfile`
- `bun test scripts/check-publish-manifests.test.ts scripts/install-copy-regression.test.ts scripts/native-install-smoke.test.ts scripts/version-sync.test.ts`
- `bunx biome check dist/signetai/bin/launch.js dist/signetai/bin/native-platforms.js dist/signetai/scripts/install-native.js scripts/check-publish-manifests.ts scripts/check-publish-manifests.test.ts scripts/install-copy-regression.test.ts scripts/native-install-smoke.test.ts`
- Staged fake `dist/signetai/native/<platform>` binaries and `dist/native/native-manifest.json`, then ran `bun scripts/check-publish-manifests.ts`
- With staged fake binaries, `npm pack --dry-run ./dist/signetai` included all five native asset paths

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
[...]

PR #820, opened 2026-06-03
fix(release): publish native npm assets as release tarballs

NicholaiVogel:
## Summary

Fixes the main release failure after #819. npm rejected the all-platform `signetai` package with `E413 Payload Too Large`, so this keeps the public npm package small while still installing the compiled native runtime.

The release workflow now:
- packs each platform native package as a `.tgz` GitHub release asset,
- injects `signetai` optionalDependencies that point at those same-release tarballs,
- publishes only the existing `signetai` npm package,
- keeps platform packages private so CI does not try to create new npm package names.

The wrapper resolves the installed optional native package directly, so `npm install --ignore-scripts` / disabled postinstall still has a path to the native binary.

## PR Readiness (MANDATORY)

Additional notes:
- Agent-scoped data queries and admin/mutation endpoints are not touched.
- Runtime package smoke was validated locally with `npm install --ignore-scripts` against tarball packages.

## Verification

- `bun test scripts/check-publish-manifests.test.ts scripts/install-copy-regression.test.ts scripts/native-install-smoke.test.ts scripts/version-sync.test.ts`
- `bunx biome check .github/workflows/release.yml dist/signetai/bin/launch.js dist/signetai/bin/native-platforms.js dist/signetai/scripts/install-native.js dist/signetai/package.json dist/signetai/README.md README.md docs/CLI.md docs/QUICKSTART.md scripts/check-publish-manifests.ts scripts/check-publish-manifests.test.ts scripts/install-copy-regression.test.ts scripts/native-install-smoke.test.ts`
- Local tarball smoke: packed fake `signetai-<platform>` tarball, injected it as `signetai` optionalDependency, installed `signetai` with `npm install --ignore-scripts`, and ran `signet --version` through the native package.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -287,28 +287,48 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Stage native npm package assets
+      - name: Pack native npm release assets
         shell: bash
         run: |
           set -euo pipefail
-          stage_bundled_binary() {
+          mkdir -p dist/native-packages
+          stage_native_package() {
             local platform="$1"
             local asset="$2"
             local binary="$3"
-            local native_dir="dist/signetai/native/${platform}"
-            mkdir -p "${native_dir}"
-            cp "dist/native/${asset}" "${native_dir}/${binary}"
-            chmod 755 "${native_dir}/${binary}"
+            local package_dir="dist/signetai-${platform}"
+            mkdir -p "${package_dir}/bin"
+            cp "dist/native/${asset}" "${package_dir}/bin/${binary}"
+            chmod 755 "${package_dir}/bin/${binary}"
+            npm pack "${package_dir}" --pack-destination dist/native-packages
           }
 
-          stage_bundled_binary "linux-x64" "signet-linux-x64" "signet"
-          stage_bundled_binary "linux-arm64" "signet-linux-arm64" "signet"
-          stage_bundled_binary "darwin-x64" "signet-darwin-x64" "signet"
-          stage_bundled_binary "darwin-arm64" "signet-darwin-arm64" "signet"
-          stage_bundled_binary "win32-x64" "signet-win32-x64.exe" "signet.exe"
+          stage_native_package "linux-x64" "signet-linux-x64" "signet"
+          stage_native_package "linux-arm64" "signet-linux-arm64" "signet"
+          stage_native_package "darwin-x64" "signet-darwin-x64" "signet"
+          stage_native_package "darwin-arm64" "signet-darwin-arm64" "signet"
+          stage_native_package "win32-x64" "signet-win32-x64.exe" "signet.exe"
+
+          gh release upload "v${NEW_VERSION}" dist/native-packages/*.tgz --clobber
+        env:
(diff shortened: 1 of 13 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `e13afbd0`.

.github/workflows/release.yml
```diff
@@ -287,28 +287,48 @@ jobs:
...
+          stage_native_package "win32-x64" "signet-win32-x64.exe" "signet.exe"
+
+          gh release upload "v${NEW_VERSION}" dist/native-packages/*.tgz --clobber
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Point npm wrapper at native release assets
+        shell: bash
+        run: |
+          set -euo pipefail
```
PR-Reviewer-Ant, line 319: This adds all platform tarballs as `optionalDependencies`, but the PR does not validate that the generated `signetai-<platform>` packages are constrained with matching `os`/`cpu` metadata. Optional dependencies are not inherently platform-selective; if these tarballs install successfully on every platform, consumers can still download all five native binaries, which brings back the payload-size problem this PR is trying to solve. Please either ensure the native package manifests carry the correct platform gates before `npm pack`, or add a publish-manifest check that fails when a native package is missing the expected `os` and `cpu` fields.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `a56cd083`

</details>

I found one packaging hole that can keep the npm wrapper oversized in practice: the release workflow wires every native tarball as an optional dependency but the new guardrail does not prove those tarballs are platform-gated.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The diff shows signetai gets all five native tarball URLs as optionalDependencies and the validation only checks package name/spec shape plus binary presence. The changed contents do not include the native package manifests, so I cannot confirm whether each tarball has os/cpu constraints, but without that constraint npm can still fetch/install every successful optional tarball.

.github/workflows/release.yml
PR-Reviewer-Ant, line 319: This wires all five native tarballs as optional dependencies, but the PR still does not validate that each generated `signetai-<platform>` package carries matching `os`/`cpu` constraints. Optional dependencies are skipped on platform mismatch only when the package manifest says so; without that metadata, npm can download/install every successful optional tarball, which undercuts the stated goal of installing a single platform native package and keeping the package-manager path small. The guardrail in `check-publish-manifests.ts` should inspect the native package manifests for the expected platform gates, or the workflow should generate those manifests here before packing.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `e13afbd0`

</details>

I found one packaging concern still worth resolving or explicitly validating: the wrapper now depends on all native tarball URLs, but this diff does not prove those tarballs are platform-gated.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The workflow injects every platform tarball into signetai optionalDependencies, and the manifest guard only validates the URL shape and binary presence. The changed contents do not include or validate the generated native package package.json os/cpu metadata, which is the key artifact needed to prove npm will install only the matching native tarball.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

PR #824, opened 2026-06-04
fix(daemon): recover reset summaries from prompt transcripts

### LINKED ISSUE #817: Handle missing SessionEnd on /reset and /clear?
**Signet version:** 0.138.8

## Context

Claude Code doesn't fire `SessionEnd` on `/reset` or `/clear` (anthropics/claude-code#6428, closed as "not planned"). We tested this with multiple resets followed by a clean exit: zero summary jobs were created for the entire conversation. The `handleSessionEnd` hook in Signet never receives the transcript.

The daemon does see activity during these sessions: `SessionStart` fires with `source: "clear"` after each reset, and `UserPromptSubmit` hooks fire for every prompt. The previous session's transcript content is available from prompt-submit snapshots.

## Question

Should Signet enqueue a summary proactively when it detects a reset? When `SessionStart` arrives with `source: "clear"`, Signet could check whether the previous session got a summary and enqueue one from the stored transcript if not.

Is there a reason not to, or a better way?

## Impact

Any session that involves `/reset` produces no session summaries. With dreaming enabled, these sessions don't count toward the token threshold.

NicholaiVogel:
## Summary

Closes #817.

This teaches the daemon hook surfaces to recover a clear/reset summary when Claude Code starts a fresh post-clear session key instead of ending the previous session with a usable transcript.

- forwards Claude Code `source: "clear"` through the CLI hook command
- stores prompt-submit live transcript fallbacks when Claude does not provide `transcript_path`
- recovers the latest previous same-project transcript on `source: "clear"`
- enqueues one `session_end` summary job using a content-derived session id
- clears the stale stored transcript after recovery or intentional skip
- clears session-start dedupe/continuity so the new post-clear session gets full startup context
- mirrors the behavior in daemon-rs to avoid TS/native hook parity drift

## PR Readiness (MANDATORY)

Notes: no public docs/API text changed; this extends existing hook behavior for Claude Code clear/reset handling and preserves existing route behavior for normal starts. No schema or publish manifest changes.

## Verification

- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "clear session-start recovers|appends prompt-submit turns"`
- `bun test surfaces/cli/src/commands/hook.test.ts --test-name-pattern "forwards Claude Code clear source|buildSessionStartBody"`
- `bunx biome check platform/daemon/src/hooks.ts surfaces/cli/src/commands/hook.ts`
- `git diff --check`
- `CARGO_TARGET_DIR=/tmp/signetai-issue817-target cargo test -p signet-daemon clear_session_start_recovers_missing_session_end_summary`
- Runtime: `cldyg` against PR daemon on `SIGNET_PORT=3851` with isolated `SIGNET_PATH=/tmp/signet-issue817-runtime`; prompt-submit created a 555-char `session_transcripts` row, `/clear` started new session key `602346ce-b1bf-40af-914b-bd54178d0d9f`, recovery enqueued `session_end` job `10d4a7bb-4562-4de9-909e-f18e53c1173f` for previous session `63758269-7e57-4334-a3da-f0c7dfa40dd6`, then cleared `session_transcripts`.

Known verification caveats:

- Runtime summary worker processing failed after enqueue because local Ollama is missing `qwen3:4b`; the recovery/enqueue behavior under test succeeded before that downstream model failure.
- Earlier full `bun test platform/daemon/src/hooks.test.ts` had one transient failure in `handleSynthesisRequest > returns prompt with database-backed temporal sources`; the same test passed when rerun directly.
- `bun run --filter '@signet/daemon' build:types` is currently not a useful narrow check: it fails on existing rootDir/test typing issues outside this patch, including `platform/core/src/migrations/*` being outside daemon `rootDir` and existing Bun SQLite `null` return typing mismatches in unrelated tests.
- Root filesystem was nearly full during Rust rebuild; the final Rust verification used `CARGO_TARGET_DIR=/tmp/signetai-issue817-target`, then that temporary target was removed.

### DIFF
--- a/platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
+++ b/platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
@@ -469,6 +469,8 @@ fn pipeline_enabled(state: &AppState) -> bool {
     if state.pipeline_paused() {
         return false;
     }
+    // The Rust manifest config does not expose the TS-only dreaming.enabled
+    // gate, so this surface follows the Rust summary worker pipelineV2 gate.
     state
         .config
         .manifest
@@ -734,6 +736,57 @@ fn session_transcript_content(
     Ok(None)
 }
 
+fn session_transcripts_has_updated_at(conn: &rusqlite::Connection) -> rusqlite::Result<bool> {
+    let mut stmt = conn.prepare("PRAGMA table_info(session_transcripts)")?;
+    let mut rows = stmt.query([])?;
+    while let Some(row) = rows.next()? {
+        let name: String = row.get(1)?;
+        if name == "updated_at" {
+            return Ok(true);
+        }
+    }
+    Ok(false)
+}
+
+fn reset_recovery_transcript_target(
+    conn: &rusqlite::Connection,
+    clear_session_key: &str,
+    harness: &str,
+    project: Option<&str>,
+    agent_id: &str,
+) -> rusqlite::Result<Option<(String, String)>> {
+    if let Some(content) = session_transcript_content(conn, clear_session_key, agent_id)?
+        && !content.trim().is_empty()
+    {
+        return Ok(Some((clear_session_key.to_string(), content)));
+    }
+
+    let project_filter = project.unwrap_or("");
+    let timestamp_expr = if session_transcripts_has_updated_at(conn)? {
+        "COALESCE(updated_at, created_at)"
+    } else {
+        "created_at"
+    };
+    let mut stmt = conn.prepare(&format!(
+        "SELECT session_key, content
+         FROM session_transcripts
+         WHERE agent_id = ?1
+           AND (?2 = '' OR harness = ?2)
+           AND (?3 = '' OR project = ?3)
+         ORDER BY {timestamp_expr} DESC
+         LIMIT 1"
+    ))?;
+    let mut rows = stmt.query(rusqlite::params![agent_id, harness, project_filter])?;
+    if let Some(row) = rows.next()? {
+        let session_key: String = row.get(0)?;
+        let content: String = row.get(1)?;
+        if !content.trim().is_empty() {
+            return Ok(Some((session_key, content)));
+        }
+    }
+    Ok(None)
+}
+
 fn audit_fs_timestamp(iso: &str) -> String {
     iso.chars()
         .map(|ch| {
@@ -833,6 +886,19 @@ fn upsert_session_transcript(
         return Ok(());
     }
     let now = chrono::Utc::now().to_rfc3339();
+    if session_transcripts_has_updated_at(conn)? {
+        let _ = conn.execute(
+            "INSERT INTO session_transcripts (session_key, agent_id, content, harness, project, created_at, updated_at)
+         VALUES (?1, ?2, ?3, ?4, ?5, ?6, ?7)
+         ON CONFLICT(session_key, agent_id) DO UPDATE SET
+            content = excluded.content,
+            harness = excluded.harness,
+            project = excluded.project,
+            updated_at = excluded.updated_at",
+            rusqlite::params![session_key, agent_id, transcript, harness, project, now, now],
+        )?;
+        return Ok(());
+    }
     let _ = conn.execute(
         "INSERT INTO session_transcripts (session_key, agent_id, content, harness, project, created_at)
          VALUES (?1, ?2, ?3, ?4, ?5, ?6)
@@ -928,6 +994,103 @@ fn enqueue_summary_job(
     Ok(id)
 }
 
+fn derive_reset_recovery_session_id(session_key: &str, transcript: &str) -> String {
+    let mut hasher = Sha256::new();
+    hasher.update(transcript.trim().as_bytes());
+    let digest = hasher.finalize();
+    let hex: String = digest[..8].iter().map(|b| format!("{b:02x}")).collect();
+    format!("session-end:{session_key}:{hex}")
+}
+
+async fn recover_missing_session_end_on_clear_start(
+    state: Arc<AppState>,
+    harness: &str,
+    session_key: &str,
+    project: Option<&str>,
+    agent_id: &str,
+) -> Option<String> {
+    if session_key.trim().is_empty() {
+        return None;
+    }
+
+    let pipeline_enabled = pipeline_enabled(state.as_ref());
+    let harness_value = harness.to_string();
+    let session_key_value = session_key.to_string();
+    let project_value = project.map(ToOwned::to_owned);
+    let agent_value = agent_id.to_string();
+    let captured_at = chrono::Utc::now().to_rfc3339();
+
+    let result = state
+        .pool
+        .write(Priority::High, move |conn| {
+            let Some((recovered_session_key, transcript)) = reset_recovery_transcript_target(
+                conn,
+                &session_key_value,
+                &harness_value,
+                project_value.as_deref(),
+                &agent_value,
+            )?
+            else {
+                return Ok(serde_json::Value::Null);
+            };
+
+            let session_id = derive_reset_recovery_session_id(&recovered_session_key, &transcript);
+            let should_skip = !pipeline_enabled
+                || transcript.len() < CHECKPOINT_MIN_DELTA
+                || is_noise_session(
+                    project_value.as_deref(),
+                    Some(session_id.as_str()),
+                    Some(recovered_session_key.as_str()),
+                    Some(harness_value.as_str()),
+                );
+            if should_skip {
+                delete_session_transcript(conn, &recovered_session_key, &agent_value)?;
+                return Ok(serde_json::Value::Null);
+            }
+
+            let job_id = enqueue_summary_job(
+                conn,
+                &harness_value,
+                &transcript,
+                Some(&recovered_session_key),
+                &session_id,
+                project_value.as_deref(),
+                &agent_value,
+                "session_end",
+                &captured_at,
+                None,
+                Some(&captured_at),
+                true,
+            )?;
+            delete_session_transcript(conn, &recovered_session_key, &agent_value)?;
+            Ok(serde_json::json!({ "jobId": job_id }))
+        })
+        .await;
+
+    match result {
+        Ok(value) => value
+            .get("jobId")
+            .and_then(|id| id.as_str())
+            .map(ToOwned::to_owned),
+        Err(error) => {
+            warn!(
+                error = %error,
+                session_key,
+                agent_id,
+                "session-start clear/reset summary recovery failed"
+            );
+            None
+        }
+    }
+}
+
+fn is_clear_session_start(body: &SessionStartBody) -> bool {
+    body.source
+        .as_deref()
+        .map(str::trim)
+        .is_some_and(|source| source.eq_ignore_ascii_case("clear"))
+}
+
 // ---------------------------------------------------------------------------
 // POST /api/hooks/session-start
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


PR-Reviewer-Ant:
Review posted through GitHub's PR review UI for commit `35369daf`.

platform/daemon/src/hooks.ts
```diff
@@ -3270,6 +3294,135 @@ export function deriveSessionEndFallbackId(
...
+	} catch (error) {
+		logger.warn("hooks", "Failed to clear recovered session transcript snapshot", {
+			error: error instanceof Error ? error.message : String(error),
+			sessionKey,
+			agentId,
+		});
+	}
+}
+
+function recoverMissingSessionEndOnClearStart(
```
PR-Reviewer-Ant, line 3392: This recovery gate appears to diverge from the Rust mirror. The TypeScript path enqueues recovery summaries when `pipelineV2.enabled || pipelineV2.shadowMode || dreaming.enabled`, while the Rust path uses `pipeline_enabled(state.as_ref())`. If `pipeline_enabled` does not include `dreaming.enabled`, the two daemons will disagree on whether a `/clear` transcript is recovered or discarded, which cuts against the PR's stated parity goal. Please either align the condition or make the intentional difference explicit with a test/comment in both surfaces.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `d1b3dece`

</details>

The recovery path is implemented in both hook surfaces and the tests cover the basic missing-SessionEnd case. I found one parity drift worth discussing: the TypeScript and Rust implementations do not use the same enablement condition for recovery.

**Confidence:** Medium [sufficient_diff_evidence, missing_cross_module_context] - The divergence is visible in the supplied diff: TypeScript treats pipelineV2 enabled, shadowMode, or dreaming as recovery-enabled, while daemon-rs delegates to pipeline_enabled(state.as_ref()). I cannot confirm from the supplied context whether Rust pipeline_enabled also includes dreaming.enabled, so this is a parity warning rather than a blocker.

NicholaiVogel:
Runtime verification added after review:

- Ran `cldyg` against the PR daemon on `SIGNET_PORT=3851` with an isolated `SIGNET_PATH=/tmp/signet-issue817-runtime`.
- Verified prompt-submit created a live `session_transcripts` row before reset: session `63758269-7e57-4334-a3da-f0c7dfa40dd6`, 555 chars, no summary jobs yet.
- Ran `/clear` in Claude Code. Claude emitted a new session key `602346ce-b1bf-40af-914b-bd54178d0d9f`.
- Patched recovery now found the latest previous same-project transcript, enqueued recovered `session_end` job `10d4a7bb-4562-4de9-909e-f18e53c1173f`, and cleared `session_transcripts`.
- The summary worker then attempted processing and failed because local Ollama is missing `qwen3:4b`; that is downstream of the recovery/enqueue behavior this PR fixes.

Also added the missing CLI forwarding for Claude Code `source: "clear"`, a live prompt transcript fallback for prompt-submit when Claude does not provide `transcript_path`, and the equivalent Rust daemon recovery fallback for fresh post-clear session keys.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `6cfce9d5`

</details>

I found one concurrency issue in the TypeScript recovery path: the duplicate-job guard is not atomic with enqueue/cleanup, so concurrent clear session-starts can still create duplicate session_end summary jobs for the same recovered transcript.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - The race is visible in platform/daemon/src/hooks.ts: recovery reads the transcript, checks summaryJobExistsForRecoveredSession, then calls enqueueSummaryJob and clears the transcript in separate DB operations. I did not run a concurrent hook repro, so confidence is medium rather than high.


**Since my last review:**
- Still blocking: 0
- Newly found: 1

platform/daemon/src/hooks.ts
```diff
@@ -3270,6 +3305,181 @@ export function deriveSessionEndFallbackId(
...
+			sessionKey,
+			agentId,
+			reason: "no-stored-transcript",
+		});
+		return undefined;
+	}
+
+	const transcript = target.transcript;
+	const recoveredSessionKey = target.sessionKey;
+	const pipelineEnabled = memoryCfg.pipelineV2.enabled || memoryCfg.pipelineV2.shadowMode || memoryCfg.dreaming.enabled;
```
PR-Reviewer-Ant: The duplicate recovery guard is still racy here. `summaryJobExistsForRecoveredSession(...)`, `enqueueSummaryJob(...)`, and `clearStoredSessionTranscript(...)` each happen in separate DB operations, so two clear session-start hooks for the same project/agent can both observe no existing job and both enqueue a recovered `session_end` job before either clears the stored transcript. This PR's stated behavior is to enqueue one recovery job; the Rust path does the selection/enqueue/delete inside one write transaction, but the TypeScript path should either make this atomic in one write transaction or rely on a unique/idempotent insert for the derived `sessionId`.

NicholaiVogel: Addressed in the latest push (`f52248cc0`). The TypeScript recovery path now keeps target selection, duplicate detection, job insert, and transcript cleanup inside one `withWriteTx(...)`, and it inserts the summary job directly in that transaction instead of calling the nested enqueue helper.

Validation rerun:
- `bun run build.ts` from `platform/daemon`
- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "clear session-start recovers|appends prompt-submit turns"`
- `bun test surfaces/cli/src/commands/hook.test.ts --test-name-pattern "forwards Claude Code clear source|buildSessionStartBody"`
- `bunx biome check platform/daemon/src/hooks.ts surfaces/cli/src/commands/hook.ts`
- `CARGO_TARGET_DIR=/tmp/signetai-issue817-target cargo test -p signet-daemon clear_session_start_recovers_missing_session_end_summary`

platform/daemon/src/hooks.ts
NicholaiVogel: Made this explicit in both surfaces in the latest push (`f52248cc0`). The TypeScript side documents why it includes the TS `dreaming.enabled` summary path, and the Rust side documents that its manifest config currently exposes only `memory.pipelineV2`, so the Rust recovery gate follows the Rust summary-worker `pipelineV2` gate.

I rechecked the Rust config before replying: `MemoryManifestConfig` only has `pipeline_v2` under `memory`, and `PipelineV2Config` has `enabled`, `paused`, and `shadow_mode` but no dreaming-enabled manifest field to align against here. The Rust recovery test still covers the Rust gate behavior.

platform/daemon/src/hooks.ts
```diff
@@ -3120,6 +3150,11 @@ export async function handleUserPromptSubmit(
 			}
 		} else if (userMessage.trim().length > 0) {
 			try {
+				const liveTranscript = formatLivePromptTranscript(userMessage, req.lastAssistantMessage);
+				const prev = getSessionTranscriptContent(req.sessionKey, agentId);
```
PR-Reviewer-Ant, line 3163: This length guard can preserve a stale transcript instead of the latest prompt-submit fallback. In the no-`transcript_path` path, `liveTranscript` is only the last assistant message plus the current user message, so a later prompt can easily be shorter than the previously stored fallback and get ignored. Then `/clear` summarizes older content, which undercuts the PR's stated recovery behavior. The fallback should prefer recency, append canonical turns into the stored recovery transcript, or otherwise track a monotonic transcript instead of comparing raw string length.

platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
```diff
@@ -734,6 +736,40 @@ fn session_transcript_content(
...
+
+    let project_filter = project.unwrap_or("");
+    let mut stmt = conn.prepare(
+        "SELECT session_key, content
+         FROM session_transcripts
+         WHERE agent_id = ?1
+           AND (?2 = '' OR harness = ?2)
+           AND (?3 = '' OR project = ?3)
+         ORDER BY created_at DESC
+         LIMIT 1",
```
PR-Reviewer-Ant: The Rust mirror selects the fallback transcript by `created_at`, but the TypeScript path intentionally uses `COALESCE(updated_at, created_at)` when that column exists. For prompt-submit fallbacks, the relevant row may be updated long after it was created, so Rust can recover an older same-project transcript instead of the latest live transcript. That is a concrete parity drift in the behavior this PR says it mirrors.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `f52248cc`

</details>

I found two correctness issues in the recovery path. The TypeScript fallback can leave the stored transcript stale, and the Rust mirror can choose the wrong previous transcript because it orders by creation time rather than the latest update.

**Confidence:** Medium [sufficient_diff_evidence, missing_runtime_repro] - Both issues are visible in the supplied diff: the TS prompt-submit fallback only overwrites when the newly formatted two-turn fallback is at least as long as the existing transcript, and the Rust recovery query orders by created_at while the TS path accounts for updated_at. I did not run a live multi-prompt or Rust parity repro, so confidence is medium.


**Since my last review:**
- Still blocking: 0
- Newly found: 2

platform/daemon/src/hooks.ts
NicholaiVogel: Addressed in latest push (`35369daff`). The no-`transcript_path` prompt-submit fallback no longer compares raw string lengths. It now appends the latest live prompt turns to the stored recovery transcript, so a shorter later prompt still advances the transcript instead of preserving stale older content.

Added regression coverage: `appends shorter live prompt-submit fallback instead of preserving stale stored content`.

Validation:
- `bun test platform/daemon/src/hooks.test.ts --test-name-pattern "clear session-start recovers|appends shorter live prompt-submit fallback|appends prompt-submit turns"`
- `bunx biome check platform/daemon/src/hooks.ts platform/daemon/src/hooks.test.ts surfaces/cli/src/commands/hook.ts`
- `bun run build.ts` from `platform/daemon`

platform/daemon-rs/crates/signet-daemon/src/routes/hooks.rs
NicholaiVogel: Addressed in latest push (`35369daff`). Rust now matches the TypeScript fallback ordering by using `COALESCE(updated_at, created_at)` when `updated_at` exists, and the Rust transcript upsert also refreshes `updated_at` on conflict so prompt-submit fallback updates are visible to recovery ordering.

Added regression coverage: `clear_session_start_recovers_latest_updated_prompt_transcript` verifies an older-created but newer-updated row is recovered instead of a stale newer-created row.

Validation:
- `CARGO_TARGET_DIR=/tmp/signetai-issue817-target cargo test -p signet-daemon clear_session_start_recovers`
- `cargo fmt --check` from `platform/daemon-rs`

NicholaiVogel:
Reviewer re-check requested for latest head `35369daff`.

Latest push addresses the two findings from the `f52248cc` review:
- TypeScript prompt-submit live fallback now appends latest live turns instead of preserving stale longer content by length.
- Rust reset recovery now orders fallback candidates by `COALESCE(updated_at, created_at)` when available and refreshes `updated_at` on transcript upsert.

Focused validation is posted on the inline threads.

PR-Reviewer-Ant (commented):
<details>

- Model: `gpt-5.5`
- Commit: `35369daf`

</details>

I did not find a new correctness or security issue worth flagging in this revision. The latest changes appear to address the previously raised stale transcript, Rust updated_at ordering, and TypeScript duplicate-enqueue concerns.

**Confidence:** Medium [sufficient_diff_evidence, targeted_context_included, missing_runtime_repro] - The changed diff shows the TypeScript recovery now performs target selection, duplicate detection, enqueue, and cleanup in one write transaction, the prompt-submit fallback appends instead of preserving stale content, and the Rust fallback now orders by COALESCE(updated_at, created_at). I did not run the hook tests or a live clear-session repro myself, so confidence stays medium.


**Since my last review:**
- Still blocking: 0
- Newly found: 0

PR #825, opened 2026-06-04
fix: harden macOS install flow, Rust daemon runtime, and opencode pipeline

aaf2tbz:
## Summary

Hardens the macOS first-run, install, update, LaunchAgent, ACPX, and opt-in Rust daemon runtime paths so new installs and updates fail loudly instead of drifting into broken or misleading states. Also fixes the Rust daemon opencode HTTP provider to work with the actual `opencode serve` API, enabling end-to-end pipeline processing (summary → synthesis → MEMORY.md).

### macOS Install & Update Hardening
- Persist `signet setup --path` into the workspace config so follow-up `signet status`, `signet configure`, and daemon commands resolve the newly-created workspace without repeating `--path`.
- Preserve `embedding.provider: none` during setup so the daemon does not fall back to native embeddings after the user disables embeddings.
- Tighten macOS LaunchAgent cleanup and stop semantics so failed starts unload launchd state and `daemon stop` does not report success while `ai.signet.daemon` is still loaded.
- Require official signing secrets for public macOS/Windows desktop tag releases instead of silently shipping self-signed public artifacts.
- Fix macOS desktop updates by switching to architecture-specific updater channels and CI metadata names: `latest-arm64-mac.yml` / `latest-x64-mac.yml`.

### ACPX Setup & Routing
- Harden ACPX setup/routing: setup resolves the pinned `acpx@0.10.0` package before writing routes, CLI help includes `acpx`, custom OpenCode model ids pass through unchanged, and `rawAgentCommand` / `agentCommand` can route custom ACP providers through ACPX `--agent`.

### Rust Daemon Runtime
- Release builds publish `signet-daemon-<platform>-<arch>` assets/checksums, the publish pipeline proceeds even if the opt-in Rust daemon build fails, native manifest downloads avoid `signet-daemon-*`, and `SIGNET_DAEMON_RUNTIME=rust` falls back gracefully to the JavaScript daemon when no Rust binary is found.

### Native ONNX Embedding
- Add `NativeEmbeddingProvider` using `ort` (ONNX Runtime) with `nomic-embed-text-v1.5` (768-dim). Lazy-loads model on first inference. Falls back to no-op embedding when ONNX runtime is unavailable.

### Opencode HTTP Provider (`OpenCodeProvider`) — NEW
- Fix session creation: read `id` field from response (not `session_id`), with fallback to `session_id`.
- Fix message format: send `{parts: [{type: "text", text: "..."}]}` instead of `{message: "..."}`.
- Fix model format: send `{providerID: "...", modelID: "..."}` object instead of bare string.
- Add Basic auth support via `OPENCODE_SERVER_PASSWORD` env var for `opencode serve` authentication.
- Extract text from response `parts` array (type `"text"`) instead of bare `response` field.
- **Result**: Summary and synthesis workers now process jobs end-to-end through `opencode serve` → `zai-coding-plan/glm-5.1`.

### Identity Parser
- Strip `**` markdown bold markers from IDENTITY.md field labels and match case-insensitively so `**Name:** Squirt` parses correctly.

### CLI Subprocess Safety
- Restructure `CliProvider::generate_inner` to use `tokio::join!` for stdout/stderr/wait, keeping the child handle alive so `child.start_kill()` works on timeout. Previously `child.wait_with_output()` consumed the handle, leaving orphaned npx processes.

### Logging & Diagnostics
- Add `signet_pipeline=info` to default tracing filter so pipeline worker/provider logs are visible.
- Add gate-check logging to summary/synthesis workers for startup debuggability.
- Downgrade TCC "Operation not permitted" watcher errors from `error!` to `debug!` to stop log spam every 2s.

### macOS Codesign
- Ad-hoc codesign daemon binary on macOS in both `service.rs` (launchd install) and `runtime.ts` (CLI startDaemon) to stabilize TCC permissions.

### Extraction Provider Preflight
- Add `acpx`, `openai-compatible`, and `command` providers to the extraction preflight match block so they are not incorrectly marked as unavailable.

## Type

- [ ] `feat` — new feature
- [x] `fix` — bug fix
- [x] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: desktop app, GitHub release workflows, daemon-rs, docs

## PR Readiness

- [x] Spec alignment validated
- [x] Agent scoping verified
- [x] Error handling and fallback paths tested
- [x] Docs updated
- [x] Regression tests added

## Testing

- [x] `bun run lint` passes
- [x] `cargo build --release` passes (daemon-rs)
- [x] Tested against running daemon (live on port 3850)
- Live pipeline verification: summary worker processed jobs through opencode serve, synthesis worker generated MEMORY.md (6,629 summaries → 16,594 chars in 18.4s), native ONNX embedding computed 768-dim vectors
- Live daemon health: 15,021 memories, 9,623 entities, 15,031 embeddings, all workers active

## AI disclosure

## Migration Notes

No database migrations. Rust daemon changes are opt-in runtime. TypeScript/Bun daemon remains default. `OPENCODE_SERVER_PASSWORD` env var should be set in launchd plist when using opencode provider with `opencode serve`.


PR-Reviewer-Ant:
Hi @aaf2tbz - I'm taking a look at the fixes in `fix: harden macos install flow` (commit `a341878a`) and will follow up shortly.

aaf2tbz:
Closing this PR. The Rust daemon fixes are valid code but the TS daemon is the active runtime — the Rust binary isn't being launched. The user-facing errors (acpx/codex model mismatch, ONNX Runtime Web) are in the TS layer, not addressed by these Rust changes. The work may be reusable if the project moves to the Rust daemon as primary.

aaf2tbz:
Closed. Will address issues as small commits over time rather than one implementation. Changes might be breaking / conflicting and need very careful effort and methodical testing measures that were not fully locked down in this work tree. As a result, this pr should not be merged but rather investigated if it brings help for a further, small committed fix.

PR #847, opened 2026-06-13
feat(daemon): add context budget profiles

NicholaiVogel:
## Summary

Adds harness-specific context budget profiles so coding harnesses can run with leaner startup context while operator/character-forward harnesses can keep richer identity and memory context. Also adds an explicit compile step for producing a single bounded coding prompt artifact from canonical identity files.

## Changes

- Added `hooks.contextProfiles`, `hooks.harnessProfiles`, and `hooks.defaultContextProfile` parsing.
- Session-start now resolves hook budgets per harness and supports ordered profile-managed identity/context files.
- Added per-file token/character budgets for profile-managed markdown context files.
- Added path safety for profile-managed identity files: markdown-only, no hidden/system workspace dirs, and realpath containment to block symlink escapes.
- Added `signet context compile --profile coding --max-chars 2200` to synthesize `AGENTS.md`, `USER.md`, `IDENTITY.md`, and `SOUL.md` into `context-profiles/coding/AGENTS.md` through the configured `session_synthesis` inference route.
- Kept hook runtime deterministic: session-start only reads compiled artifacts; it does not perform model synthesis in the hot path.
- Documented coding/rich profile examples in configuration docs and `agent.yaml.template`.

## Type

- [x] `feat` — new user-facing feature (bumps minor)
- [ ] `fix` — bug fix
- [ ] `refactor` — restructure without behavior change
- [ ] `chore` — build, deps, config, docs
- [ ] `perf` — performance improvement
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/cli` / dashboard
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

Checklist notes:
- No spec dependency graph entries were required for this daemon/CLI config feature.
- No new scoped data query surface was added; existing memory queries still use resolved agent scope.
[...]

PR #850, opened 2026-06-14
fix(daemon): context-budget profile review fixes (#847)

NicholaiVogel:
Stacked on top of #847. Addresses three findings from the review of `feat(daemon): add context budget profiles`.

### 🔴 Keep predicted context additive at session start (regression)
The feature change capped predicted context by how full main recall was (`predictedLimit = max(0, recallLimit - memories.length)`). When recall filled to `recallLimit` (default `50`), `predictedLimit` became `0` and the **entire** predicted-context feature was silently dropped.

- **Proof:** session-start `memoryCount` dropped from `52` on `main` to `50` on the feature branch; the two existing `predictive context …` tests **pass on `main`, failed on the branch, pass again now**.
- Predicted context is additive by design — it surfaces topics the user is likely to need next regardless of token budget consumed. Restored the additive call matching `main`.

### 🟠 Preserve global hook settings in context profile merge
`mergeConfig` spread the profile override over the base, but profile readers emit fully-populated objects with `undefined` for unset keys. The post-spread `undefined`-delete ran too late, so an unset override key (e.g. a profile that only sets `recallLimit`) clobbered a defined global value like `sessionStart.recencyBias` or `includeIdentity`, silently reverting it to the code default.

- Only copy override keys that are actually present and defined.
- New test: profile overrides `recallLimit` while an unrelated global `recencyBias` / `includeIdentity` / `maxInjectTokens` survives.

### 🟡 Hold profile identity budgets under the truncation marker
Truncated profile identity sections appended `\n[truncated]` *after* hitting the declared `maxTokens`/`maxChars`, so a truncated section exceeded its configured per-file budget — breaking the strict bounded-context contract this feature exists to enforce.

- Reserve room for the marker (`maxTokens - markerTokens` / `maxChars - marker.length`); when the budget is smaller than the marker itself, cap without the marker rather than overflow.
- New tests for both the normal overflow case and tiny budgets below the marker size.
- Two existing tests used `maxTokens` below the marker size (`4` and `3`), which encoded the old overflow; bumped to `6` so the file is still truncated but the marker fits legitimately.

## Verification
Run in an isolated worktree at the latest PR tip (`86c010591`):

| Suite | result |
|---|---|
| `hooks.test.ts` (full) | 165 pass / 0 fail (was 162/3) |
| `hooks-config.test.ts` | 7 pass / 0 fail |
| `identity-context.test.ts` | 8 pass / 0 fail |
| `context-budget.test.ts` + CLI `context.test.ts` | 7 pass / 0 fail |
| Each new regression test fails on the un-fixed feature branch | ✅ confirmed |
| `autoreview` on the full fix set | 0 findings |

The only remaining `hooks.test.ts` failure (`handleSynthesisRequest > returns prompt with database-backed temporal sources`) is **pre-existing** — it fails identically on `origin/main` and is unrelated to context profiles.

## Notes
- Stacked PR → base is `nicholai/context-budget-profiles`. Once merged, #847 picks these up.
- No `main` rebase needed (this targets the feature branch tip directly).
- Touched files introduce no new typecheck errors (the pre-existing `build:types` failures in `context-construction.ts` / `subagent-context.test.ts` are unchanged).

## Type
- [x] `fix`

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -1198,6 +1198,26 @@ hooks:
     recallLimit: 10
     maxInjectChars: 500
     minScore: 0.8
+  contextProfiles:
+    coding:
+      sessionStart:
+        recallLimit: 5
+        maxInjectTokens: 5000
+      userPromptSubmit:
+        maxInjectChars: 300
+      identity:
+        files:
+          - path: context-profiles/coding/AGENTS.md
+            maxChars: 2200
+    rich:
+      sessionStart:
+        recallLimit: 50
+        maxInjectTokens: 20000
+  harnessProfiles:
+    pi: coding
+    codex: coding
+    hermes-agent: rich
+    openclaw: rich
   preCompaction:
     includeRecentMemories: true
     memoryLimit: 5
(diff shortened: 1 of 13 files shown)


NicholaiVogel:
Closing in favor of folding the fixes directly into #847 (per review decision). #847 now contains the feature plus these three commits rebased onto current `main` (0.140.5):

- `fix(daemon): keep predicted context additive at session start` — restores additive predicted context (was starved to 0 when recall filled to `recallLimit`).
- `fix(daemon): preserve global hook settings in context profile merge` — `mergeConfig` no longer clobbers inherited globals like `recencyBias`/`includeIdentity`.
- `fix(daemon): hold profile identity budgets under the truncation marker` — truncated sections now respect the declared `maxTokens`/`maxChars` (incl. tiny budgets).

All 187 tests across the touched suites pass; `version-sync --check` aligned at 0.140.5.

PR #870, opened 2026-06-26
fix(openclaw): surface Signet health diagnostics

### LINKED ISSUE #867: OpenClaw/Clawdbot Signet integration degrades silently when daemon is unhealthy or unreachable
# OpenClaw/Clawdbot Signet integration degrades silently when daemon is unhealthy or unreachable

## Summary

OpenClaw/Clawdbot reports Signet memory issues because the Signet daemon is not consistently reachable/healthy, while the OpenClaw adapter and/or legacy hook path can continue in a degraded state. This leaves the agent with partial or no memory context, and the fallback behavior is confusing enough that local workarounds were added in Clawdbot/BlueBubbles instead of the root Signet integration being fixed.

This issue is not asking to remove the current BlueBubbles workaround. That workaround only improves inbound group-message context and should stay in place until Signet/OpenClaw integration is reliable.

## Environment

- Signet CLI: `0.135.0`
- Clawdbot: `2026.1.24-3`
- Signet repo checked locally at `Signet-AI/signetai`, branch `feat/turboquant-vector-compression`
- OpenClaw runtime reported by `signet status --json`: `plugin`
- Daemon endpoint expected by docs and adapter: `http://localhost:3850`
- Local memory DB present: `~/.agents/memories.db`

## Observed behavior

`signet status --json` reports Signet installed and the DB healthy, but daemon not running:

```json
{
  "installed": true,
  "validIdentity": true,
  "db": {
    "exists": true,
    "schema": "core",
    "needsMigration": false,
    "memoryCount": 10937
  },
  "daemon": {
    "running": false,
    "pid": null,
    "version": null
  },
  "openclawRuntime": "plugin",
  "openclawDualSystem": false
}
```

Direct health checks fail or hang depending on execution context. Inside a restricted sandbox the connection is refused; outside the sandbox the port accepts but does not return a health response within 3 seconds:

```bash
curl -sS --max-time 3 http://127.0.0.1:3850/health
# curl: (28) Operation timed out after 3002 milliseconds with 0 bytes received
```

At the same time, daemon artifacts can leave misleading stale state:

- `~/.agents/.daemon/last-healthy-start` contains a recent healthy start record.
- `~/.agents/.daemon/pid` contains a PID.
- `lsof -nP -iTCP:3850 -sTCP:LISTEN` showed a `bun.exe` listener, even though `curl` timed out and `signet status` said the daemon was not usable.

## Relevant logs

Older daemon stderr shows sqlite-vec failures:

```text
SQLiteError: no such module: vec0
  at checkNullVectors
  at buildEmbeddingHealth

Vector search failed: SQLiteError: no such module: vec0
  at vectorSearch
  at hybridRecall
```

Recent structured logs show the OpenClaw plugin path running, but with slow/degraded memory operations:

```text
Session start completed
harness=openclaw
runtimePath=plugin
engine=hybrid
durationMs=3984

User prompt submit recall was slow
harness=openclaw
durationMs=4567

Worker tick error
errorMessage="database is locked"

Recall stage timings
totalMs=16954.47
query_embedding_wait=15016.1

Inference target legacy-synthesis/default failed
error="Ollama timeout after 90000ms"

summary-worker Job failed
error="All routed targets failed."
```

## Local Clawdbot state

The local Clawdbot config has `memorySearch.enabled: true` for the iMessage agent, but the internal hook entries are disabled:

```json
{
  "agents": {
    "list": [
      {
        "id": "imessage",
        "memorySearch": {
          "enabled": true
        }
      }
    ]
  },
  "hooks": {
    "enabled": true,
    "internal": {
      "entries": {
        "signet-memory": { "enabled": false },
        "agent-memory": { "enabled": false }
      }
    }
  }
}
```

There is also a legacy hook at `~/.clawdbot/workspace/hooks/agent-memory/handler.js` that posts to:

- `/api/hooks/session-start`
- `/api/hooks/remember`
- `/api/hooks/recall`

It uses `http://localhost:3850` and fails with `Error: daemon ...` or connection failures when the daemon is not reachable. It also sends `x-signet-runtime-path: legacy`, while the plugin sends `plugin`; Signet docs say the session tracker returns `409 Conflict` when both paths touch the same session.

## Local workaround that should not be broken

A separate BlueBubbles workaround was applied in Clawdbot to improve group-chat message envelopes. The issue there was that group messages did not expose sender/group context clearly enough to the agent, causing replies to risk going to a DM instead of the group. The workaround changes the visible envelope from a generic BlueBubbles label to include sender and group context.

That workaround is orthogonal to Signet. It should remain until Signet/OpenClaw memory injection is reliable, because it protects message routing even when memory is degraded.

## Why this appears to be a Signet issue

The Signet OpenClaw adapter already anticipates daemon outages:

```text
signet-memory: daemon unreachable at <url>. Memory tools will silently no-op until daemon is running.
```

But in practice the system can end up in a confusing state:

1. `signet status` says daemon is stopped/unreachable.
2. stale daemon state suggests it recently started or has a PID.
3. Clawdbot/OpenClaw still receives partial `<signet-memory>` injection or minimal stubs in some sessions.
4. recall/synthesis/extraction logs show slow operations, DB locks, sqlite-vec failures, and model timeouts.
5. the user-facing agent reports Signet problems but then invents local workarounds instead of getting a clear actionable health failure.

## Expected behavior

Signet should make the OpenClaw/Clawdbot integration fail clearly and recover cleanly:

- `signet status` should distinguish:
  - daemon process absent
  - process/listener present but health endpoint unreachable
  - daemon healthy but memory subsystems degraded
  - OpenClaw plugin connected but injecting degraded/minimal context
- The OpenClaw plugin should surface a concise health warning to logs/diagnostics when it is no-oping or returning minimal stubs.
- If `sqlite-vec`/`vec0` is unavailable, Signet should either:
  - degrade explicitly to FTS-only recall with a visible warning, or
  - fail health checks with a clear remediation.
- The plugin and legacy hook path should be mutually exclusive with clear diagnostics, especially around `x-signet-runtime-path: plugin` vs `legacy` and `409 Conflict`.
- `signet status` should report stale PID/last-health artifacts as stale rather than implying a simple stopped daemon.

## Suggested repro/checklist

1. Install Signet with OpenClaw/Clawdbot integration.
2. Configure Clawdbot iMessage/BlueBubbles agent with memory search enabled.
3. Stop or break the Signet daemon, or trigger a sqlite-vec unavailable state.
4. Send a BlueBubbles message that starts or resumes an OpenClaw session.
5. Compare:
   - `signet status --json`
   - `curl http://localhost:3850/health`
   - `~/.agents/.daemon/logs/daemon.err.log`
   - `~/.agents/.daemon/logs/signet-YYYY-MM-DD.log`
   - Clawdbot agent-visible context/injection

## Acceptance criteria

- Users can run one command and know whether Signet is healthy for OpenClaw/Clawdbot.
- If Signet is unhealthy, the issue points to the exact failing subsystem: daemon process, HTTP health, sqlite-vec, DB lock, inference provider, plugin/legacy conflict, or no-op plugin state.
- Clawdbot/OpenClaw should not need local memory-context workarounds to compensate for opaque Signet failure modes.

NicholaiVogel:
## Summary
- make `signet status` honest about daemon health by distinguishing a healthy daemon from a stale pid, a Signet process with no healthy HTTP endpoint, and a TCP listener whose `/health` probe does not respond
- keep OpenClaw/Clawdbot/Moltbot as one integration path internally, with OpenClaw as the current canonical name
- have the OpenClaw memory adapter publish a minimal heartbeat to the existing daemon diagnostics route and log clearly when daemon health disables memory hooks
- surface OpenClaw runtime/heartbeat problems through the existing `signet doctor` flow; no new CLI command/target is added

Fixes #867.

## OpenClaw source validation
- Checked current OpenClaw source in `references/openclaw`:
  - `v2026.6.21-alpha.1` exposes `registerService`, `api.on`, `before_prompt_build`, and legacy `before_agent_start`.
- Checked older reported line in `references/openclaw`:
  - upstream/fork tags available for the reported date are `v2026.1.24` and `v2026.1.24-1`; no `v2026.1.24-3` tag exists on the checked remotes.
  - `v2026.1.24` exposes `registerService`, `api.on`, and legacy `before_agent_start`; `before_prompt_build` is absent there, so this change keeps the existing `before_agent_start` fallback compatible.

## Validation
- `PUPPETEER_SKIP_DOWNLOAD=1 bun install` in clean PR worktree
- `bun run build:core`
- `bun run --filter '@signet/sdk' build`
- `bun run build:connector-base`
- `bun run --filter '@signet/connector-openclaw' build`
- `bun run --filter '@signet/connector-hermes-agent' build`
- `cd integrations/openclaw/memory-adapter && bun test src/index.test.ts`
- `cd surfaces/cli && bun test src/lib/runtime.test.ts src/features/health.test.ts`
- `bun run --filter '@signetai/signet-memory-openclaw' build`
- `cd surfaces/cli && PUPPETEER_SKIP_DOWNLOAD=1 bun run build:workspace-deps && bun run build:cli`
- `autoreview` final simplified pass: no blockers

### DIFF
--- a/integrations/openclaw/memory-adapter/src/index.test.ts
+++ b/integrations/openclaw/memory-adapter/src/index.test.ts
@@ -36,6 +36,7 @@ let lastPromptSubmitBody: unknown = null;
 let lastMemoryRecallBody: unknown = null;
 let lastSessionSearchBody: unknown = null;
 let lastCheckpointBody: unknown = null;
+let lastHeartbeatBody: unknown = null;
 let warnMessages: string[] = [];
 let testDir = "";
 
@@ -140,6 +141,7 @@ beforeEach(() => {
 	lastMemoryRecallBody = null;
 	lastSessionSearchBody = null;
 	lastCheckpointBody = null;
+	lastHeartbeatBody = null;
 	checkpointResponse = null;
 	warnMessages = [];
 	testDir = mkdtempSync(join(tmpdir(), "signet-openclaw-test-"));
@@ -155,6 +157,9 @@ beforeEach(() => {
 			switch (path) {
 				case "/health":
 					return jsonResponse({ pid: 1234 });
+				case "/api/diagnostics/openclaw/heartbeat":
+					lastHeartbeatBody = init?.body ? JSON.parse(String(init.body)) : null;
+					return jsonResponse({ ok: true });
 				case "/api/hooks/session-start":
 					if (timeoutSessionStartCount > 0) {
 						timeoutSessionStartCount -= 1;
@@ -1823,6 +1828,17 @@ describe("registration guard (#422)", () => {
 		expect(tools.length).toBeGreaterThan(0);
 		expect(hooks.size).toBeGreaterThan(0);
 		expect(registeredServices.length).toBeGreaterThan(0);
+	});
+
+	it("publishes an OpenClaw diagnostics heartbeat after startup health succeeds", async () => {
+		const { api } = createMockApi({ registrationMode: "full", version: "test-plugin" });
+		signetPlugin.register(api);
+		await Bun.sleep(0);
+
+		expect(lastHeartbeatBody).toMatchObject({
+			pluginVersion: "test-plugin",
+			hooksRegistered: ["before_prompt_build", "before_agent_start", "agent_end", "before_compaction", "after_compaction"],
+		});
 	});
 
 	it("second full-mode register() call is a no-op", () => {
--- a/integrations/openclaw/memory-adapter/src/index.ts
+++ b/integrations/openclaw/memory-adapter/src/index.ts
@@ -40,6 +40,8 @@ const DEFAULT_DAEMON_URL = "http://localhost:3850";
 const RUNTIME_PATH = "plugin" as const;
 const READ_TIMEOUT = 5000;
 const WRITE_TIMEOUT = 10000;
+const HEARTBEAT_TIMEOUT = 2000;
+const HEARTBEAT_INTERVAL_MS = 60_000;
 const COMPACTION_HOOK_DEDUPE_MS = 1000;
 const SESSION_START_TIMEOUT = resolveSessionStartTimeoutMs(
 	process.env.SIGNET_SESSION_START_TIMEOUT ?? process.env.SIGNET_FETCH_TIMEOUT,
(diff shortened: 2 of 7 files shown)


NicholaiVogel:
Simplified this after review: the PR no longer adds `signet doctor openclaw` / `signet doctor clawdbot` or a separate CLI surface.

The current shape is intentionally narrower:
- `signet status` is honest about daemon health/probe state.
- existing `signet doctor` reports OpenClaw runtime/heartbeat issues when that integration is configured.
- the OpenClaw adapter logs clearly and sends a minimal heartbeat when daemon health disables memory hooks.
- OpenClaw/Clawdbot/Moltbot stay treated as the same harness lineage, with OpenClaw as the canonical current name.

PR #881, opened 2026-07-03
fix(daemon): chunk and embed Obsidian source content after sync

quannon:
## Summary

Obsidian (and other native memory) sources were writing artifacts to `memory_artifacts` but never chunking/embedding their content, so connected vaults always showed `0 chunks` / "extraction pending" on the dashboard. Both call sites of `startNativeMemoryBridge` omitted `embeddingConfig`/`fetchEmbedding`, so the embedding-gated chunking path (`indexObsidianSourceEmbeddings`) never ran.

## Changes

- `native-memory-sources.ts`: added `resolveEmbeddingBridgeOptions(embeddingCfg, fetchEmbedding)` as a single source of truth for wiring `embeddingConfig`/`fetchEmbedding` into `NativeMemoryBridgeOptions` based on the configured embedding provider.
- `daemon.ts` (startup sync) and `routes/sources-routes.ts` (manual re-sync route): both now use the helper instead of duplicated inline ternaries that never actually got wired in.
- `native-memory-sources.test.ts`: new direct unit test for `resolveEmbeddingBridgeOptions` (wired vs. `provider: "none"`).
- `routes/sources-routes.test.ts`: fixed a stale assertion that had encoded the bug as expected behavior (`toBeUndefined()` on `embeddingConfig`/`fetchEmbedding` → `toBeDefined()` + provider check).
- `daemon-rs/parity/03-test-corpus-manifest.md`: added a port note on the `native-memory-sources.test.ts` / `obsidian-source-embeddings.test.ts` rows so the future Rust port of Obsidian embedding wires this correctly from day one instead of reproducing the bug (confirmed daemon-rs doesn't implement Obsidian chunking/embedding yet — this is a `needs-port` gap, not a parity regression).

Out of scope: GitHub/Discord source providers have a related but distinct gap (they never call `indexObsidianSourceEmbeddings`/any chunk-embed step at all, regardless of `embeddingConfig`) — not touched here.

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/daemon`

## Screenshots

N/A — no UI/dashboard code changed (only the chunk count the dashboard already displays is now correct).

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — no spec entry covers this behavior specifically; `docs/SOURCES.md` already documents the correct chunking behavior, this fix brings the code into conformance with existing docs (no doc/spec drift to resolve).
- [x] Agent scoping verified on all new/changed data queries — no new data queries added; the helper only forwards an already agent-scoped `memoryCfg` (loaded via `loadMemoryConfig(input.agentsDir)`) through to the existing bridge options.
- [x] Input/config validation and bounds checks added — N/A, no new external input surface; reuses existing `loadMemoryConfig` validation.
- [x] Error handling and fallback paths tested (no silent swallow) — N/A, no new error paths introduced; helper is a pure function with one conditional branch, covered by its unit test.
- [x] Security checks applied to admin/mutation endpoints — N/A, no admin/mutation endpoint touched.
- [x] Docs updated for API/spec/status changes — N/A beyond the `daemon-rs` parity port note (see Changes); `docs/SOURCES.md` was already accurate.
- [x] Regression tests added for each bug fix — `resolveEmbeddingBridgeOptions` unit test (would fail before the fix) + corrected `sources-routes.test.ts` assertion that had encoded the bug as expected behavior.
- [x] Lint/typecheck/tests pass locally — see Testing below.

N/A — no migrations touched.

## Testing

- [x] `bun test` passes — `bun test src/native-memory-sources.test.ts src/obsidian-source-embeddings.test.ts src/routes/sources-routes.test.ts` → 80 pass, 0 fail.
- [x] `bun run typecheck` passes — diffed `tsc --noEmit` output against `main` baseline: byte-identical (905 pre-existing, unrelated errors), zero new errors introduced.
- [x] `bun run lint` passes — touched files are clean (`daemon.ts`, `sources-routes.ts` fully clean; two pre-existing formatting nits in `native-memory-sources.ts`/`.test.ts` confirmed present on `main` too, unrelated to this diff).
- [x] Tested against running daemon — built `dist/native/signet-darwin-arm64` from this branch, installed it locally (`signet install --force`), restarted the daemon, and confirmed live on a real Obsidian source: `chunks` went from 0 to 3000+ and `chunkCoverage` reached 1.0 as the vault re-indexed.

## AI disclosure

## Notes

None.


NicholaiVogel:
@quannon thank you! merging this.

PR #912, opened 2026-07-10
fix: gate summary jobs on runnable synthesis

NicholaiVogel:
1|## Summary
2|
3|Stops session-summary jobs from consuming retries when no effective synthesis workload exists, without stranding command-provider extraction or stale crash leases.
4|
5|TypeScript now gates each lease on a forced-refresh `session_synthesis` route check and rolls back the lease without consuming an attempt if availability changes across the lease transition. Rust no longer starts the summary worker when synthesis is disabled or configured as `none`.
6|
7|Closes #899
8|
9|## Changes
10|
11|- gate TypeScript summary-worker startup and each lease on the effective inference-router workload
12|- preserve active command-provider extraction, which shares `summary_jobs`, when synthesis is disabled
13|- require pipeline V2 to be enabled before retained command-provider configuration can bypass synthesis gating
14|- make TypeScript leasing compare-and-set safe and restore `pending` plus the original attempt count after a post-lease availability loss/error
15|- recover stale TypeScript `processing`/`leased` jobs independently when the polling worker is disabled
16|- monitor an unavailable startup route and promote recovery-only mode to polling when the workload becomes healthy
17|- gate Rust summary-worker startup on enabled, non-`none` synthesis instead of hidden Ollama fallback
18|- add regression coverage for disabled, disappearing, errored, command-mode, recovery-only, and available workloads
19|
20|## Type
21|
22|- [ ] `feat` — new user-facing feature (bumps minor)
23|- [x] `fix` — bug fix
24|- [ ] `refactor` — restructure without behavior change
25|- [ ] `chore` — build, deps, config, docs
26|- [ ] `perf` — performance improvement
27|- [ ] `test` — test coverage
28|
29|## Packages affected
30|
31|- [ ] `@signet/core`
32|- [x] `@signet/daemon`
33|- [ ] `@signet/cli` / dashboard
34|- [ ] `@signet/sdk`
35|- [ ] `@signet/connector-*`
36|- [ ] `@signet/web`
37|- [ ] `predictor`
38|- [x] Other: Rust `signet-core` and `signet-daemon`
39|
40|## Screenshots
41|
42|N/A. No UI changes.
43|
44|## PR Readiness (MANDATORY)
45|
46|- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
47|- [x] Agent scoping verified on all new/changed data queries
48|- [x] Input/config validation and bounds checks added
49|- [x] Error handling and fallback paths tested (no silent swallow)
[...]

PR #918, opened 2026-07-11
fix(pipeline): bound and cancel LLM inference

NicholaiVogel:
## Summary

Bounds and cancels background pipeline inference so extraction, synthesis, and structural workers cannot accumulate unbounded LLM processes or keep expensive work alive after pipeline shutdown.

PR #912, whose summary-worker synthesis gate and recovery behavior this work builds on, has now merged into `main`; this PR is rebased on that current base.

## Changes

- enforce a daemon-wide LLM concurrency semaphore, configurable with `memory.pipelineV2.worker.maxLlmConcurrency` (default `2`, bounded `1..16`) or `SIGNET_MAX_LLM_CONCURRENCY`
- proxy threaded extraction inference through the main daemon so worker-thread extraction shares provider instances, concurrency accounting, cancellation, telemetry, and billing controls
- propagate pipeline-scoped `AbortSignal`s through HTTP, Ollama, OpenCode, command, Codex, ACPX, Claude Code, and llama.cpp providers
- terminate timed-out/cancelled Unix subprocess groups with `SIGTERM`, independently escalate to `SIGKILL`, and await cleanup
- abort active extraction and summary/continuity calls during pause/stop and restore leased jobs without consuming attempts
- prevent overlapping structural-classification, structural-dependency, and summary ticks
- request structured, non-thinking Ollama output only for extraction calls
- strip ambient `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` from Claude Code background calls by default; allow explicit inheritance with `allowApiKeyEnv`
- optionally pass Claude Code `--max-budget-usd`
- add a daemon-scoped Claude quota/billing cooldown circuit with explicit status semantics and honest persisted-login caveats
- add TypeScript/Rust config parity, status output, docs, and regression coverage

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: Rust `signet-core` / daemon config parity

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

No schema or scoped-data query changes. No admin/mutation endpoint was added. Status and configuration changes are documented.

## Migration Notes

- [ ] Migration is idempotent

No schema migration. Rollback is a normal code revert. The new global concurrency limit defaults to `2`; operators can set `memory.pipelineV2.worker.maxLlmConcurrency` or `SIGNET_MAX_LLM_CONCURRENCY`. Claude Code background calls no longer inherit ambient Anthropic API credentials unless `allowApiKeyEnv: true` is explicitly configured. Legacy, unshipped `billingMode: api-key` is mapped for compatibility.

## Testing

- [ ] `bun test` passes repository-wide
- [ ] `bun run typecheck` passes repository-wide
- [x] Changed-file lint passes
- [ ] N/A

Verified on exact SHA `830e656bfb291d20d65a6dfefe0d3ccf40ffe4c4` (clean rebase onto current `main`, including merged native-embedding worker isolation PR #916):

- focused daemon regression matrix — **376 passed, 0 failed**
- native-embedding worker-isolation regression matrix — **17 passed, 0 failed**
- `platform/core` build and declaration generation
- daemon production/dashboard build
- `bunx biome check <changed TypeScript files>`
- `bun run check:rust-parity` — **328 routes current**
- `cargo fmt --package signet-core --package signet-daemon -- --check`
- `cargo test -p signet-core config -- --nocapture` — **35 passed, 0 failed**
- `cargo check -p signet-core -p signet-daemon`
- exact-parent comparison: the parent full daemon suite has **62 existing broad failures**. The branch initially also had 62, but one was a branch-specific idle-stop timing regression; that regression was corrected and the final focused matrix is green. The full broad suite was not rerun a third time after that isolated correction.
- workspace-wide Rust formatting still reports an untouched pre-existing difference in `signet-pipeline/src/reflection_worker.rs`
- repository root typecheck requires generated/built package dependencies and reports unrelated baseline workspace errors; affected core/daemon builds and changed-file checks pass

## AI disclosure

## Notes

The Claude Code cooldown is deliberately daemon-scoped because all production provider instances use the same canonical pipeline configuration. `allowApiKeyEnv: false` controls only inherited environment credentials; it cannot override a persisted Claude Code login, which is documented explicitly.


NicholaiVogel:
## Review finding: native embedding is back on the main thread (reverts the event-loop isolation that just landed)

While reviewing the concurrency/cancellation work (which is solid), I traced what happens to native embeddings now that `embedding-worker*` is deleted, and I think this deserves a decision before merge.

### What I found

`doInit()` (`platform/daemon/src/native-embedding.ts:223`) runs **fully in-process on the main thread**:

- `await transformers.pipeline("feature-extraction", MODEL_ID, ...)` (ONNX/WASM pipeline creation + model load), and
- the warmup `await embed("test", ...)`

both execute on the main thread. There is **no `worker_threads` Worker** in `doInit()` anymore (that was the deleted `embedding-worker*`). ONNX/WASM instantiation and inference are synchronous-CPU work, so they **block the event loop** while they run.

### What still holds (the mitigations in this PR)

- Init is **lazy** (triggered on first embed/status check via the memoized `initPromise`, not at daemon startup; nothing in `daemon.ts` calls it eagerly).
- `/health` itself does **not** await init (it reads embedding config synchronously), so it does not block on init directly.

### Why it matters

The event loop is single-threaded. When native embeddings are used, the ONNX init (first embed, several seconds) **and every subsequent embed inference** run on the main thread, so `/health` and all other requests queue behind that work. This is the exact `/health`-blocking that `fix-native-embedding-eventloop-isolation` was merged to fix, and this PR reverts it without a replacement off-thread mechanism.

The new LLM concurrency semaphore does **not** cover this: it bounds LLM *provider* calls (external processes), not in-process ONNX work.

Scope: only affects users on the **native** embedding provider (OpenAI/Ollama embedding users are unaffected), and only once embeddings are actually used (lazy init).

### Options

1. **Keep a worker for native embedding** (re-add the off-thread isolation). The rest of this PR (semaphore, cancellation, credential hygiene) is independent and unaffected.
2. **Accept the tradeoff** (native embedding is optional + init is lazy/one-time) and document explicitly that native-embedding users will see a brief daemon stall on first embed and per-embed main-thread load.
3. **Run ONNX via `proxy: true`** in the `pipeline()` call so `onnxruntime-web` keeps its own internal worker. Lighter than the bespoke `embedding-worker*` but still off-thread.

If daemon responsiveness under native embeddings matters (which it did, given the isolation branch was merged for it), I would lean toward 1 or 3.

Everything else I checked looks good: `provider.test.ts` core cancellation/cooldown tests pass (the 23 failures here are all missing `acpx`/`codex` CLIs, environmental), Rust config parity (35 tests), `cargo check`, rustfmt, and biome are clean, and the semaphore uses a correct `try/finally` permit lifecycle.

Happy to verify the per-embed blocking empirically or help wire up option 3 if useful.

NicholaiVogel:
Thanks — this was a valid catch. #916 merged after the `main` snapshot used for #918's previous rebase, so #918 was stale relative to the new native-embedding isolation—not intentionally deleting it.

I rebased #918 onto current `main` including #916 and force-updated the draft PR. The current head is `830e656bfb291d20d65a6dfefe0d3ccf40ffe4c4`. The PR diff now has **no changes** to `native-embedding.ts` or `embedding-worker*`; `embedding-worker.ts` is present from `main`.

Verified after rebase:
- `bun test src/native-embedding-eventloop-isolation.test.ts src/embedding-worker-handle.test.ts src/native-embedding.test.ts` — **17 passed, 0 failed** (including `/health` SLA while model download is stuck)
- pipeline provider/lifecycle suite — **228 passed, 0 failed**
- daemon production build — passed

Thanks for flagging the moving-base issue before it could mask #916's worker isolation.

Issue #913, opened 2026-07-10
Unify Background Memory around Source Aggregation and Dreaming


NicholaiVogel:
## Goal

Make Background Memory one coherent system:

```text
episodic evidence → shared source selection → Dreaming → semantic memory
                                                      ↓
                                            semantic-first retrieval
```

The daemon owns storage, scope, provenance, validation, and semantic writes.
Dreaming is the reasoner over evidence and semantic context; an in-process Pi
agent and an external harness are execution bindings, not separate memory
systems.

### Episodic evidence

Artifacts, Signet Sources, native-harness memory, transcripts, imported
records, session summaries, and compactions are immutable provenance-bearing
source material. Ingest persists and indexes it; it does not decide semantic
conclusions or rewrite evidence when understanding changes.

Source-native folders, documents, headings, and links remain episodic topology,
not semantic knowledge about the user or world.

### Semantic memory and retrieval

Memories, entities, claims, and relationships are derived current
understanding. Dreaming may create, reinforce, supersede, merge, or leave them
alone through audited operations with evidence and attribution.

Retrieval spans both layers: semantic results lead normal recall; episodic
sources remain available for proof, temporal drill-down, deep history, and
questions without a conclusion.

## Cutover requirements

- Reuse and consolidate existing source capture, artifact lineage, transcript,
  summary, compaction, temporal, retrieval, and graph-apply owners. Do not add
  a parallel aggregator, queue, worker, graph writer, or provider.
- The existing Dreaming worker uses the shared episodic selector and the
  daemon-owned audited semantic apply seam. Retire overlapping extraction,
  decision, promotion, and structural semantic runtimes in the same cutover.
- Preserve agent, visibility, project, session, harness, source-purge, and
  forgetting boundaries. External agents never write SQLite directly.
- Keep compaction and temporal lineage for their non-Dreaming responsibilities.
- Use the daemon workload resolver → router → Pi/ACPX path. No direct vendor
  client or second inference layer.
- Make no compatibility read/write runtime or second source of truth.

## Agentic Dreaming end state

Replace the fixed graph snapshot / one-shot plan with a bounded daemon-owned
tool loop. Keep the cursor and evidence selection contract; give the reasoner
live scoped navigation and one audited write seam:

- reads: `search_entities`, `get_entity`, `list_aspect_claims`, `walk_links`,
  `get_claim_evidence`, `get_link_evidence`, `search_evidence`;
[...]

PR #938, opened 2026-07-19
fix(daemon): add temporal intent / freshness prior to hybrid recall

### LINKED ISSUE #903: Hybrid recall lacks temporal intent/freshness prior; stale semantic matches outrank recent facts
## Summary

With complete embeddings and active vector search, broad queries asking about current July 2026 state ranked semantically similar March records above directly relevant July records. Exact identifier/keyword recall still found the current records.

## Current behavior

Hybrid ranking blends vector, FTS, graph, and currentness signals. Existing currentness handling primarily addresses explicit graph supersession; it does not provide a general freshness prior for ordinary time-sensitive language.

## Expected behavior

Queries containing temporal intent such as `current`, `latest`, `recent`, or an explicit month/year should prefer relevant facts from that period, without globally penalizing timeless knowledge.

## Suggested approach

- Detect explicit temporal language and dates in the query.
- Infer bounded `since`/`until` candidate constraints where confidence is high.
- Otherwise apply a bounded recency prior after semantic relevance.
- Strongly downrank superseded/contradicted facts when a newer scoped fact exists.
- Preserve existing behavior for timeless knowledge queries.

## Regression fixtures

Create semantically similar historical and current facts, then assert:

- `What is the current status?` ranks the newest relevant fact first.
- `What did we plan in March?` still ranks the March fact first.
- A timeless preference query is not dominated solely by recency.

## Related

- #913 §7 re-homes `temporal-recall.ts`'s data source (off `session_summaries`, onto ingest source/memories) and names this issue a coordination dependency — the cutover must carry any landed freshness prior forward.
- #945 (`review_after` temporal-claim expiry) — recall-time freshness (this issue) vs. write-time expiry (#945); coordinate the gap window where a past-due claim is still recency-boosted.

aaf2tbz:
## Summary

Closes #903. Hybrid recall blended vector, FTS, graph, rehearsal, and currentness signals but did not recognize implicit freshness queries such as "current status". This PR detects explicit freshness language and folds a bounded `created_at` recency boost into the existing rehearsal stage, mirrored in the Rust daemon.

**Status: ready for review.** Self-diffed (no auto-review); all local gates pass.

## Changes

- `platform/daemon/src/temporal-recall.ts` — new `hasFreshnessIntent(query)` for unambiguous `current|currently|latest|recent|recently|today|this week|this month` language. Bare `now`, month phrases, explicit dates, and timeless queries do not trigger it.
- `platform/daemon/src/memory-search.ts` — the existing rehearsal stage now applies the bounded freshness boost `1 + weight × 0.5^(ageDays/halfLifeDays)` against `created_at` when freshness intent is present. There is no standalone fourth scoring stage, duplicate month-range path, or extra supersession penalty; existing currentness remains the sole structured supersession correction. Explicit `since`/`until` callers bypass the freshness boost.
- `platform/daemon/src/memory-config.ts` — new `search.temporal_prior_enabled` (deliberately default-on for the #903 fix), `search.temporal_prior_weight` (0.15, clamped 0–1), `search.temporal_prior_half_life_days` (14, clamped 1–365). The default is intentional because explicit freshness language receives only a bounded near-tie boost; operators can set `temporal_prior_enabled: false` to preserve unshaped ranking.
- Rust parity — identical freshness detection and rehearsal-stage semantics in `platform/daemon-rs`, with no additional scoring stage. `/api/memory/recall` stays `native-rust-replay-proven`.
- Cherry-picked PR #933 (`fix(daemon-rs): add missing repair routes for rust daemon parity`) so the expected rust CI contract-replay failure on the missing `/api/repair/integrity-check` and `/api/repair/rebuild-indexes` routes does not fail here.
- Regression tests: `platform/daemon/src/memory-search-freshness.test.ts` covers explicit freshness detection, the three reported timeless-query false positives, bare `now`, month/date non-triggers, recent-first ranking, unchanged timeless/month-range scoring, and explicit since/until bypass; mirrored intent and scoring coverage lives in `signet-core`.
- Docs: `docs/PIPELINE.md`, `docs/MEMORY.md`, `docs/CONFIGURATION.md`.

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/daemon`
- [x] Other: `platform/daemon-rs`

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — no spec surface change; docs updated instead
- [x] Agent scoping verified on all new/changed data queries — new `created_at` lookups are ID-only metadata reads on already-authorized candidates, mirroring the sibling rehearsal-boost pattern; currentness query stays agent-scoped
- [x] Input/config validation and bounds checks added — config knobs clamped at load (TS) and use (Rust); unparseable timestamps skipped
- [x] Error handling and fallback paths tested (no silent swallow) — the existing rehearsal-stage warning/fallback now covers freshness metadata reads too
- [x] Security checks applied to admin/mutation endpoints — N/A, read-path only

## Testing

- [x] Focused TypeScript recall tests pass (55/55 before the final added month-path regression; final freshness suite 7/7); memory-config passes 96/96 in isolation. Full-suite environment failures are pre-existing.
- [x] `bun run lint` passes on changed files (biome; remaining warnings pre-existing)
- [x] Cargo tests pass (`signet-core` lib 81/81, freshness-focused 4/4, daemon compile/check, built-binary recall contract replay `search_endpoints`)
- [x] Tested against running daemon — A/B tested enabled vs disabled against a copy of a real 249-memory workspace DB. The three reported timeless queries had identical IDs/order/scores; an explicit latest-status query received only the bounded boost.
- [x] `bun run typecheck` — repo has no per-package typecheck for daemon; `tsc --noEmit` shows only pre-existing errors, none in touched files

## Cross-issue coordination

- **#913 §7:** the planned `temporal-recall.ts` data-source re-home must carry this PR's source-agnostic `hasFreshnessIntent` detector and the freshness-aware scoring handoff in `memory-search.ts` forward rather than re-deriving or dropping them.
- **#945:** `review_after` expiry is a future write/maintenance-time signal, while this PR is a recall-time `created_at` heuristic. During the gap before dreaming supersedes a past-due claim, freshness ranking could still boost it. When #945 lands, past-due `review_after` rows should be excluded from or inversely shaped by the freshness boost. This PR does not invent an expiry path before that schema exists.
- TypeScript and Rust intentionally keep separate detector implementations; mirrored intent regressions are the parity guard against drift.

## AI disclosure

## Notes

- Contains the commit from #933; whichever lands first, the other merges cleanly (identical diff).
- Minor known divergence: legacy SQLite `YYYY-MM-DD HH:MM:SS` `created_at` values parse as local time in TS and UTC in Rust (matches the existing rehearsal-boost behavior); ISO-8601 values are identical on both sides. Scores are excluded from recall parity rules, so contract replay is unaffected.


NicholaiVogel:
Thanks for the work on this. The freshness intent detection is a genuinely useful addition. After testing against real data and reviewing the existing recall pipeline, I want to flag a structural concern before merge.

## What the recall pipeline already does

Signet recall already has three temporal and recency mechanisms that run on every query:

1. `resolveTemporalRecall` (temporal-recall.ts, from PR #779). This function handles explicit date queries like "what did we plan in March" through the `temporal_edges` graph table. It has its own MONTHS map, its own date range builder, and its own candidate injection into hybridRecall. It runs before the hybrid scoring stages and can return its own results or inject candidate memory IDs into the pipeline.

2. `applyCurrentnessBias` (memory-search.ts). This function loads supersession state from the graph on every recall. Memories with only superseded facts get a 0.65 multiplier. Memories with partial supersession get 0.85. Memories with active facts get 1.03.

3. `applyRehearsalBoost` (memory-search.ts). This function applies a recency boost based on access_count and last_accessed with half-life decay on every recall.

## What this PR adds and where the duplication is

The PR adds a fourth scoring stage called `applyTemporalFreshnessPrior`. It does three things:

**Range intent.** The `parseFreshnessIntent` range mode re-parses month names using a duplicate MONTHS map. It builds date ranges and downweights out-of-range memories at 0.7. This overlaps with what `resolveTemporalRecall` already does through the graph, but with cruder scoring and no structural query support. Having two systems that parse month names from query text and build date ranges is a maintenance problem. If someone fixes a bug in one MONTHS map, the other stays broken.

**Freshness intent.** The `parseFreshnessIntent` freshness mode detects keywords like "current", "latest", "now", and "recent". This is genuinely new. No existing system handles queries that imply recency without stating an explicit date. This is the part worth keeping.

**Supersession penalty.** The `applyTemporalFreshnessPrior` function applies a 0.5 multiplier for superseded memories. This stacks on top of `applyCurrentnessBias`, which already applies 0.65 for the same signal. Two separate penalties for the same graph data means the combined effect (0.65 times 0.5 equals 0.325) is much more aggressive than either function alone suggests. A developer tuning one multiplier will not know the other exists.

## Test results against real data

I ran an A/B test with the prior enabled versus disabled against a copy of my memory database (approximately 30,000 memories). Vector search fell back to FTS due to a temp config issue, which actually isolates the prior effect more cleanly.

The range intent worked correctly. The query "what did we plan in March" promoted March 2026 memories and dropped an April entry from the top 5. Good behavior.

The freshness intent fired on timeless queries where it should not have:

- "how does the embedding pipeline work" caused ranking shifts on entries that have no temporal relevance
- "what is Signet architecture" caused a newer ontology claim to jump over a more semantically relevant older entry
- "GitHub bot configuration" reordered the top 2 entries with no temporal justification

The score impacts were larger than the PR description suggests. In the query "work done in June", an out-of-range March entry dropped from 0.69 to 0.44. That is a 36 percent score reduction. The 0.7 range multiplier plus the 0.5 supersession penalty produces a combined 0.35 multiplier in the worst case, which is much more aggressive than the "bounded 15 percent boost" framing in the PR body.

## Suggested refactor

The novelty here is the freshness keyword detection. The range detection and supersession penalty duplicate existing infrastructure. I suggest the following changes:

1. Remove the range intent entirely. `resolveTemporalRecall` already handles date-range queries through the graph with structural precision. The PR should not add a second code path for the same problem.

2. Keep the freshness intent detection for keywords like "current", "latest", "recent", "now", and "today". Wire it into the existing scoring stages instead of adding a fourth stage between dampening and currentness. For example, pass the freshness signal into `applyRehearsalBoost` to adjust the half-life or weight when freshness intent is present. Or pass it into `applyCurrentnessBias` as a tunable multiplier on the existing supersession penalty. Either approach reuses existing infrastructure.

3. Remove the standalone `applyTemporalFreshnessPrior` function and the duplicate MONTHS map and month parser.

4. Reconsider whether the bare word "now" should trigger freshness intent. In testing it was the most common source of false positives on timeless queries.

NicholaiVogel:
One more thing. The PR body explicitly skips live daemon testing:

> Tested against running daemon: N/A. Verified via fixture-backed recall tests and rust contract replay instead.

Fixture-backed unit tests with synthetic data cannot catch the problems above. The false positives on timeless queries, the aggressive combined multipliers, the score drops that are larger than the formula implies. These only surface when the prior runs against a real memory graph with real supersession data, real created_at distributions, and real query patterns.

The recall pipeline is the core product. Changes to how it ranks results need to be validated end to end against a running daemon with real data before merge. Unit tests with hand-crafted fixtures are a baseline check, not a replacement for integration testing on the actual system.

aaf2tbz:
Addressed in follow-up commit `0ca53bd6`.

Changes made from the review:
- removed the standalone `applyTemporalFreshnessPrior` / `apply_temporal_freshness_prior` stage
- removed the duplicate month-range parser and out-of-range multiplier; date-range interpretation stays in the existing temporal recall path
- removed the extra `0.5` supersession penalty, so currentness is applied only by the existing currentness stage
- retained only explicit freshness terms and folded the bounded `created_at` boost into the existing rehearsal stage in both TypeScript and Rust
- removed bare `now` as a trigger
- added mirrored regressions for the observed timeless queries and for month-range queries not receiving a second scoring path

I also completed the requested live daemon validation against a copy of a real 249-memory workspace database. With the prior enabled vs disabled, these queries returned identical IDs, ordering, and scores:
- `how does the embedding pipeline work`
- `what is Signet architecture`
- `GitHub bot configuration`

A freshness query (`what is the latest Signet status`) kept the same result ordering while receiving only the bounded recency score increase.

Validation completed:
- focused TypeScript recall tests: 55/55 passed before the final added regression; final freshness suite: 7/7 passed
- memory config tests: 96/96 passed when run in isolation
- Rust `signet-core` lib: 81/81 passed
- Rust freshness-focused tests: 4/4 passed
- Rust daemon compile/check passed
- Rust recall contract replay `search_endpoints`: passed with the built daemon binary
- Biome passed on touched files (3 pre-existing `any` warnings in `memory-search.ts`)
- repository TypeScript check still reports the branch's pre-existing unrelated errors; no new errors in the freshness changes

NicholaiVogel:
## Review: behavioral + cross-issue assessment

Reviewed against the diff, #903, #913 (§7), and #945. Overall: the ranking logic itself is sound and well-scoped — I'd be comfortable with the freshness approach landing. The blocking concerns are a dead-code hunk and a cross-issue coordination gap, detailed below.

---

### What's good

**Folding freshness into the existing rehearsal stage** instead of adding a 4th scoring stage. One extra column (`created_at`) on an already-running query, no extra DB round-trip, no new latency. The commit history shows the author first built a separate stage + a duplicate month-range path + a stacked supersession penalty, then recognized the over-engineering and ripped it back out. That's exactly the "prefer deleting branches" discipline.

**Well-calibrated for the reported failure.** Max +15% (`weight 0.15`), 14-day half-life. A 14-day-old row gets ~+7.5%; a 130-day-old row gets ~+0.03%. It can only break near-ties, which is precisely the defect in #903. It cannot meaningfully distort a strong relevance gap.

**Conservative trigger gating.** Fires only on explicit terms; bare `now`, month names, and dates are excluded because they appear in timeless queries. The test suite pins the real false-positives from the issue.

**Correct separation from the date-range path.** When a caller passes explicit `since`/`until`, the boost is skipped. No double-counting.

**Honest, accurate parity note.** The TS/Rust `created_at` local-vs-UTC divergence for legacy `YYYY-MM-DD HH:MM:SS` values genuinely pre-exists in the rehearsal boost (`new Date(...)` in TS parses space-format as local; Rust's `NaiveDateTime` + `naive_utc()` treats as UTC). Not introduced here, inherited consistently.

**Independent enabling.** Freshness applies even when `rehearsal_enabled: false` — correct, since they're now two separate signals sharing one stage.

---

### Code concerns (priority order)

**1. Dead currentness guard — please revert.** In `hybridRecall`, `currentness` is declared as an empty map and the only assignment is inside the new `if (currentness.size === 0)` guard. Nothing populates it beforehand, so the guard is always true and the change is a no-op. It's leftover from the removed supersession-penalty iteration.

  This is higher-priority than it looks because of #945: that issue's lifecycle step 5 depends on the currentness pass to annotate its `review_after`-driven supersessions. The dead guard sits in exactly the code path #945 will extend. Revert it so #945 builds on clean ground.

**2. `temporal_prior_enabled` defaults to `true` — make the framing explicit.** This silently re-orders recall results for every user. The A/B evidence against a real 249-memory DB is good, and default-on is defensible *because* the boost is conservative and this is filed as a `fix`. But it conflicts with the issue label (`enhancement`) and with the fact that #945 is the *precise* mechanism that will make part of this heuristic redundant once it lands. Shipping it default-on commits every user to a rough rule that a more precise rule is designed to partially replace. That's an acceptable call, but it should be a deliberate one, not a quiet default.

**3. `temporalNow` is a test hook on the public `RecallParams` interface.** It works, but it's a test-only knob on an exported production type any caller could pass. AGENTS.md asks for narrow APIs. Prefer a module-level time seam or an internal-only params type.

**4. Unrelated #933 (repair routes) is bundled in.** Disclosed and pragmatically justified (keeps rust contract-replay CI green), but violates "one PR = one issue/topic" and means you can't cleanly revert freshness without reverting the repair routes. Acceptable as an expedient.

**5. Naming/conceptual conflation.** The stage is now "freshness-aware rehearsal" — one named stage doing two semantically distinct things (access-frequency rehearsal vs. creation-recency), multiplicatively stacked. Not a bug, and the LOC savings justify it, but worth a one-line comment.

**6. Two implementations of the same detector.** TS uses one regex with `\b...\b` alternatives; Rust tokenizes into words and checks a 2-word window. They agree on the obvious edge cases today, but it's a future drift risk rather than a shared spec.

---

### Cross-issue coordination (the main gap)

#938 is the only piece of a cluster of three interrelated projects that is shipping code. #913 and #945 are both un-specced, un-PR'd refactors. #938 commits behavioral defaults and implementation choices that the other two will have to absorb or override, and **none of the three artifacts reference each other.**

**#913 §7 plans to re-home `temporal-recall.ts`'s data source** (off the `session_summaries` DAG, onto ingest source/memories), and explicitly names #903 as a coordination dependency:

> Coordinate with **#903** (temporal freshness prior, currently open with no PR) so the re-home preserves or hands off the freshness ranking rather than disrupting it.

PR #938 is that "no PR." It puts `hasFreshnessIntent` directly into `temporal-recall.ts` and folds freshness into `memory-search.ts` — both modules #913 will restructure. The issue's assumption ("because #903 has not landed, the cutover carries the temporal-recall re-home itself") is now stale. The cutover must consciously carry #938's ranking forward rather than re-derive it.

  Note: `hasFreshnessIntent` is data-source-independent (pure string matching), so it'll travel cleanly through the re-home — but only if the cutover knows it exists. **Suggest #938 note the #913 §7 dependency in its body.**

**#945 introduces `review_after` for temporal-claim expiry.** #938 and #945 are formally orthogonal — different stage, different signal, different column:

| | #903 / #938 | #945 |
|---|---|---|
| Stage | recall-time | write/maintenance-time |
| Reads | `created_at` | `review_after` |
| Mechanism | heuristic recency boost | structural `supersede_claim_value` |
| Trigger | freshness words in the query | referenced date passing |

No structural collision, and in the happy path they reinforce: #945's superseded old version gets downweighted by the currentness pass, while the retrospective version gets #938's recency boost. Both push the same direction.

**But there's a gap window where they contradict.** A temporal claim past its deadline that dreaming hasn't superseded yet ("going to Venezuela on March 15, 2027," queried on March 20) would still get a *recency boost* from #938, because #938 only sees `created_at`, not `review_after`. #945's intent is to push that row *down*. The conflict is temporary (resolved once dreaming runs) and isn't #938's bug (`review_after` doesn't exist yet), but it's a coordination point: when #945 lands, the freshness boost should probably exclude or invert for past-due `review_after` rows. **Neither #938 nor #945 currently notes this.**

---

### Net recommendation

Approve the ranking logic. Before merge:

1. **(blocking)** Revert the dead `currentness.size === 0` guard — it sits in #945's future code path.
2. **(blocking)** Make the `fix` + default-on framing a deliberate maintainer call, given the #945 redundancy.
3. **(recommended)** Add a #913 §7 coordination note to the PR body; note `hasFreshnessIntent` is deliberately source-agnostic so the re-home carries it forward.
4. **(recommended)** Note the #945 `review_after` gap-window tension so it isn't rediscovered later.
5. **(nice-to-have)** Move `temporalNow` off the public interface; note the TS/Rust dual-implementation drift risk.

The rest (bundled #933, naming conflation) are process nits.

aaf2tbz:
Addressed the latest review in follow-up commit `9b06b078`.

Changes:
- reverted the dead `currentness.size === 0` guard so the currentness pass remains a clean extension point for #945
- made default-on explicit and deliberate in TypeScript, Rust, docs, and mirrored config regressions; the bounded boost remains opt-out via `search.temporal_prior_enabled: false`
- removed the test-only `temporalNow` field from the public `RecallParams` interface and made the regression fixtures runtime-relative instead
- updated the PR body with the #913 §7 re-home dependency, the #945 `review_after` gap-window handoff, and the TS/Rust detector-parity risk

I intentionally did not implement #945 behavior here because `review_after` does not exist yet; the PR now records the precise integration requirement for when that schema lands.

Validation completed:
- focused TypeScript config/freshness tests: 104/104 passed
- broader TypeScript recall/search tests: 56/56 passed
- Rust `signet-core` tests: 82/82 passed
- Rust formatting and `signet-daemon` check passed
- TypeScript core and daemon builds passed after refreshing the local core build artifact
- Biome passed on the touched TypeScript files

PR #943, opened 2026-07-20
fix(openclaw): restore plugin discovery and lifecycle hooks

LeuciRemi:
## Summary

Restore Signet plugin discovery, static tool exposure, and lifecycle capture
with OpenClaw 2026.7.1+.

OpenClaw now performs side-effect-free discovery passes, requires exact tool
contracts, and blocks conversation-aware hooks for non-bundled plugins unless
the operator explicitly grants access. Without those contracts, Signet's tools
and `agent_end` transcript capture were absent from the active runtime.

## Changes

- handle OpenClaw `discovery` and `tool-discovery` registration modes without runtime side effects
- declare the exact static tool contract required by OpenClaw
- expose marketplace integrations through `mcp_server_list` and `mcp_server_call`
- remove asynchronous dynamic tool registration that OpenClaw could not activate
- grant the explicit conversation-access permission required by `agent_end`
- preserve existing OpenClaw hook policy while patching configuration
- bump the optional OpenClaw peer dependency to `>=2026.7.1`
- update connector, onboarding, and marketing documentation
- add regression coverage for discovery, lifecycle registration, configuration migration, and tool contracts

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [x] `@signet/web`
- [ ] `predictor`
- [x] Other: `@signetai/signet-memory-openclaw`

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

The unchecked readiness item is due only to the repository-wide
`bun run lint` baseline. Typecheck, focused tests, builds, and scoped Biome
checks all pass. This change adds no data queries, admin routes, or mutation
endpoints; the corresponding scoping and security checks are N/A after review.

- [x] Migration is idempotent

The configuration patch is idempotent and preserves existing hook policy.
Rust parity is N/A because this change is limited to the OpenClaw integration.
Rollback consists of reverting this commit; older OpenClaw runtimes remain
supported through the explicit legacy runtime path.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Additional validation:

- 157 focused tests pass across the adapter, connector, CLI sync, and publish-manifest suites
- `@signetai/signet-memory-openclaw` build passes
- `@signet/connector-openclaw` build passes
- scoped Biome check passes on all seven touched TypeScript files
- real OpenClaw 2026.7.2 runtime validation confirms all 9 declared tools, all 5 lifecycle hooks, the Signet service, and `allowConversationAccess: true`
- `bun run lint` reports 2,329 pre-existing repository-wide formatting errors
- a live Signet daemon was not required because this regression concerns OpenClaw plugin registration

## AI disclosure

## Notes

OpenClaw requires exact `contracts.tools` declarations and blocks
conversation-aware hooks for non-bundled plugins unless
`plugins.entries.signet-memory-openclaw.hooks.allowConversationAccess` is
explicitly enabled.

The previous asynchronous marketplace expansion logged tools as registered,
but those tools were absent from OpenClaw's runtime registry. Marketplace
access now remains behind the declared `mcp_server_list` and
`mcp_server_call` tools.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -186,7 +186,7 @@
         "@types/node": "^22.0.0",
       },
       "peerDependencies": {
-        "openclaw": ">=2026.5.22",
+        "openclaw": ">=2026.7.1",
       },
       "optionalPeers": [
         "openclaw",
--- a/integrations/openclaw/connector/README.md
+++ b/integrations/openclaw/connector/README.md
@@ -1,4 +1,4 @@
-# @signetai/connector-openclaw
+# @signet/connector-openclaw
 
 Signet connector for OpenClaw (and its earlier names: clawdbot, moltbot).
 
@@ -6,20 +6,21 @@ Signet connector for OpenClaw (and its earlier names: clawdbot, moltbot).
 
 Unlike other harnesses, OpenClaw reads `~/.agents/AGENTS.md` directly, so no
 generated output file is needed. Instead, this connector patches the JSON
-config to point OpenClaw at the Signet workspace and enables memory hooks.
+config to point OpenClaw at the Signet workspace and configures the Signet
+memory plugin.
 
 ## Installation
 
 ```bash
-npm install @signetai/connector-openclaw
+npm install @signet/connector-openclaw
 # or
-bun add @signetai/connector-openclaw
+bun add @signet/connector-openclaw
 ```
 
 ## Usage
 
 ```typescript
-import { OpenClawConnector } from '@signetai/connector-openclaw';
+import { OpenClawConnector } from '@signet/connector-openclaw';
 
 const connector = new OpenClawConnector();
 await connector.install('~/.agents');
@@ -40,16 +41,22 @@ await connector.install('~/.agents');
   `OPENCLAW_HOME`, `CLAWDBOT_HOME`, `MOLDBOT_HOME`, `MOLTBOT_HOME`,
   `OPENCLAW_STATE_HOME`
 - Sets `agents.defaults.workspace` to point at `~/.agents`
-- Enables the `signet-memory` internal hook entry
+- Selects `signet-memory-openclaw` as the memory plugin and adds it to
+  `plugins.allow`
+- Grants the plugin the explicit `hooks.allowConversationAccess` permission
+  required for `agent_end` transcript capture
+- Disables OpenClaw's built-in memory search and the legacy `signet-memory`
+  internal hook to prevent duplicate memory paths
 - Creates hook handler files in `~/.agents/hooks/agent-memory/` for
-  `/remember`, `/recall`, and `/context` commands
+  legacy `/remember`, `/recall`, and `/context` command compatibility
 
 ## API
 
-### `install(basePath: string): Promise<InstallResult>`
+### `install(basePath: string, options?: OpenClawInstallOptions): Promise<InstallResult>`
 
 Install the connector. Patches all found OpenClaw config files and installs
-hook handler files.
+hook handler files. The plugin runtime is the default; pass
+`{ runtimePath: "legacy" }` only for an older OpenClaw runtime.
 
 ### `uninstall(basePath?: string): Promise<UninstallResult>`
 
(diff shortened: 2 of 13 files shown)


NicholaiVogel:
Hi. The manifest contract and the allowConversationAccess change are clearly correct and necessary for OpenClaw 2026.7.1+. Those parts are good.

I want to understand the removal of `registerMarketplaceProxyTools` and the 15-second refresh timer before this merges. The previous code dynamically registered marketplace tools so a harness saw them as first-class citizens. The new code requires the harness to call `mcp_server_list` then `mcp_server_call` for each one. The old test that verified dedup ("does not reregister marketplace proxy tools on refresh") is rewritten to assert the proxy tools are absent.

The PR description says the old behavior was "asynchronous dynamic tool registration that OpenClaw could not activate" and that tools "were absent from OpenClaw's runtime registry." That is the part I need more detail on. What specifically failed. Did OpenClaw 2026.7.1+ drop support for dynamically registered tools entirely, or did the registration happen at a point in the lifecycle where OpenClaw had already snapshot the tool list. If it is the latter, the fix could be to register the proxy tools during the new `tool-discovery` pass instead of on a timer, which would preserve the first-class integration. If it is the former and OpenClaw genuinely cannot support dynamic tool registration anymore, then removal is correct and the regression is unavoidable.

The reason this matters: the new `mcp_server_list` / `mcp_server_call` indirection is a real downgrade for any harness that previously could call marketplace tools directly. The harness now has to know to do a two-step lookup, and most harnesses will not. If the removal is forced by OpenClaw, fine. If it is a workaround for a timing bug, it is worth fixing properly.

Separately, the two new registration modes (`discovery` and `tool-discovery`) and the new `registerMemoryCapability` API method are a feature, not a fix. The PR type checkbox says `fix`. That is not a blocker, but the type label should match the contents. The feature work itself looks reasonable.

Can you share what specifically broke with the dynamic registration on OpenClaw 2026.7.1+ so I can tell whether the proxy removal is forced or a workaround.

LeuciRemi:
Thanks — that is a fair question. I went back and reproduced the old behavior against the actual OpenClaw loaders rather than relying only on the current implementation.

The previous marketplace registration started an asynchronous refresh during the plugin’s synchronous registration phase. Once that phase returns, OpenClaw closes the registration API. `registerTool` is not one of the methods that remains callable afterward, so the later calls are ignored.

I replayed the old adapter against OpenClaw 2026.5.22, the locally installed 2026.7.1 release, and the current 2026.7.2 reference. In each case, the adapter logged that one mocked marketplace tool had been registered, but that tool was absent from OpenClaw’s runtime registry. There were no diagnostics because calls made after the registration window are silently ignored by the guarded API.

So this was not a dynamic-registration feature that OpenClaw specifically removed in 2026.7.1. The old asynchronous approach was already ineffective on the adapter’s previously declared minimum version, 2026.5.22. OpenClaw still supports registering tools synchronously while the registration window is open, but it does not support adding them asynchronously afterward.

Moving the existing helper into `tool-discovery` would therefore not fix it: the helper must await the Signet daemon before it knows which marketplace tools to register, while the discovery registration pass must also complete synchronously. Preserving first-class marketplace tools would require a different integration design, potentially through OpenClaw’s native MCP configuration, and that should be validated separately.

I agree that removing the intended first-class proxy surface is worth calling out. However, on the supported versions tested here, it does not remove working runtime behavior: the proxy tools were already missing despite the successful registration log. The declared `mcp_server_list` and `mcp_server_call` tools make marketplace access explicit and actually available.

On the PR type: I would keep this as `fix`. The new registration-mode handling and `registerMemoryCapability` call adapt Signet to OpenClaw’s current plugin lifecycle and restore existing Signet tools and transcript capture; they do not introduce a new user-facing capability. This also follows the repository convention of reserving `feat` for user-facing features.

I’ll update the PR description to state the verified version range and the synchronous-registration constraint more precisely.

NicholaiVogel:
Thanks Rémi, appreciate the thorough repro. That clears it up.

Going to merge this.

PR #946, opened 2026-07-21
refactor(dreaming): unify episodic evidence selection (#913)

### LINKED ISSUE #913: Unify Background Memory around Source Aggregation and Dreaming

NicholaiVogel:
## What this is

Hard cutover that closes #913. Ingestion becomes immutable episodic evidence; Dreaming — an agentic, tool-equipped reasoning pass — is the only path from evidence to semantic memory; retrieval is semantic-first with episodic provenance intact. The old extraction/synthesis pipeline is deleted, not deprecated.

```text
episodic evidence → shared source selection → Dreaming → semantic memory
                                                      ↓
                                            semantic-first retrieval
```

Rust is deliberately excluded: upstream removed `daemon-rs`.

### Ingestion: evidence in, nothing else

- remember / hooks / CLI / MCP inputs become **immutable episodic evidence** — memories, artifacts/Signet Sources, transcripts, summaries, and compactions. No input path mutates semantic graph state directly, anywhere.
- One agent-scoped episodic selector covers all evidence kinds with temporal cursor ordering, source-native topology, provenance, and source-purge behavior.
- The entire static pipeline is gone: extraction, decision, escalation, threaded extraction, direct structural/promotion/reclassification/supersession/dependency synthesis, and the fake status/config paths. 44 files deleted; the pipeline directory alone is −15.7k lines.

### Dreaming: the sole semantic writer, as an agent

- **Agentic pass loop.** The model navigates the graph with scoped tools and decides what to inspect; there is no static prompt→JSON generation. Prompt persuades, the validator enforces.
- **One capability registry, three generated bindings.** Thirteen scoped operations — entity search/detail, aspect claims, link walk, claim/link provenance, episodic search, three deterministic verification guards (entity label, exact-canonical duplicate candidates, conservative claim contradiction), scoped runbook read/write, and cited `apply_ontology_ops` — bound to Pi `AgentSession`, a process-scoped restricted ACPX MCP (schemas accept no agent id; every read/write pinned to the worker agent; team/hybrid runs use a short-lived scoped credential), and the CLI (`signet dream capabilities` / `signet dream tool`). CLI, MCP, and in-harness capability overlap is guaranteed by construction. Every semantic write uses the one apply seam: exact-quote canonical citations, per-operation savepoint isolation, all 19 core ontology operations executable.
- **Scheduler with three wake conditions and a health gate.** The worker sweeps every 5 minutes and runs a pass when: the evidence backlog exceeds `tokenThreshold` (ceiling), a non-empty backlog is older than `memory.dreaming.maxInterval` (6h staleness floor), or attention items are pending (event wake). A queue-health gate defers the sweep under pipeline pressure (manual trigger remains a deliberate override), and a sweep makes at most one provider attempt so a degraded provider can't fan out across agents. Compact mode stays an explicit manual action.
- **Semantic attention channel.** `review_due`, `hygiene`, `contested_claim`, and `evidence_requeue` producers share one agent-scoped work queue. A pending item wakes Dreaming with zero new evidence, survives failure, and resolves only after its selected pass completes.
- **Auditable graph hygiene.** Six deterministic detectors — junk names, zero-attribute entities, missing claim keys, duplicate canonical names, generic aspects, generic `related_to` links — queue cleanup as attention items, scoped to semantic entities (source-native topology and pinned entities excluded). Destructive ops may cite a hygiene attention item as provenance **only when targeting exactly the flagged subject**; content-bearing ops keep the strict quote requirement. Legacy graphs with no new ingestion get cleaned by ordinary passes.
- **Runbook as meta-memory.** Each pass stores local operational history — its exact evidence window, applied/rejected op outcomes, unresolved quarantines, an optional structured note. The next pass receives a compact, explicitly non-evidentiary view and can read/write the runbook through the shared registry.
- **Identity is optional context, never schema.** File-anchored (AGENTS/SOUL/IDENTITY/USER/MEMORY + `dreaming` specials), re-read per pass. Missing files are fine; an unreadable configured file marks the pass `identity context degraded`. Claims attach to their entity and aspect — the user is an entity like any other.
- **Minimal prompt, expressive contract.** The prompt is one goal sentence; semantics live in the operation contract — `entity_type` is a closed enum, field descriptions carry claim-shape rules.
- **Bounded passes over unbounded evidence.** Oversized immutable sources resume at exact safe-boundary offsets across passes with parent provenance intact; excluded/quarantined sources requeue only explicitly; rejected ops are retained for inspection.

### Provenance, lineage, and lifecycle

- `memory_kind` (094) separates episodic, semantic, and derived rows; attribute projections materialize as semantic memories (102) marked explicitly `derived` (103) and surfaced in hybrid recall.
- Canonical derived lineage (104, `derived_memory_sources`): a source correction, supersession, or deletion **marks dependent derived rows stale immediately**; recall excludes stale rows; a repeated aggregate query re-derives the snapshot, preserving evidence lineage and recording a `rederived` history event. Claim supersession/archive invalidates aggregate snapshots that cited the prior claim.
- Semantic claim projections are content-immutable through generic memory PATCH (`409 semantic_projection_content_immutable`); claims change only via the audited ontology path. Metadata remains editable.
- The lifecycle gate (deleted/superseded/stale) is closed on `/memory/similar` too: tombstoned or aggregate-recall rows neither surface as candidates nor seed the query vector (they 404 as seeds).
- Source-derived semantics carry the configured Signet Source purge key and purge on disconnect; pre-existing user-owned rows keep their provenance.

### Retrieval and measurement

- Semantic-first retrieval remains; episodic sources stay available for provenance and temporal recall. The MemoryBench profile supports a controlled graph on/off ablation over one preserved workspace.
- `GET /api/dream/quality` measures scoped citation coverage, entity garbage rate (shared classifier + possessive-duplicate detection), and structure metrics, excluding source-native topology. `GET /api/dream/passes/:passId/tools` exposes the per-pass ordered capability trace (input, output, outcome, latency) so a run's search behavior is reviewable.

### Operator surface and config

- API: `GET /api/dream/status`, `.../passes`, `.../passes/:id/tools`, `.../quality`. CLI: `signet dream run|status|passes|capabilities|tool`.
- Config: `memory.dreaming.{enabled, tokenThreshold, maxInterval, maxInputTokens, maxOutputTokens, timeout, backfillOnFirstRun}`. Inference is provisioned the one way — the router over pi-ai/ACPX targets; dreaming carries its own token budgets as call parameters, no per-task model splits.
- Schema: migrations 093–104 — evidence cursor, memory kind, compaction projections, legacy-ingestion retirement, failure backoff, evidence exclusions/quarantine, capability trace, runbook, attention, attribute semantic memories, semantic memory kind, derived lineage.

## Proof to date

- Focused suites pass across the Dreaming lifecycle/worker, scoped ACPX MCP, capability registry and all three bindings, auth guards, source selector, migrations, source purge, ontology applicators, scheduler (102 tests), transactions/mutation API (66 tests), stale-derivative invalidation and aggregate recall (147 tests), and `/memory/similar` lifecycle (5 tests, real route + migrated SQLite + sqlite-vec). Daemon typecheck and Biome pass.
- Pi is on the current 0.81.1 runtime contract; each routed model registers in an isolated in-memory `ModelRuntime`; a regression test proves silent context-overflow responses settle instead of `continue()` from a terminal assistant message. A router sentinel bug (`undefined` vs explicit paused `null`) that kept aggregate recall off the shared Pi route is fixed.
- Fresh-workspace live validation: remember → episodic evidence → router → Pi-backed Dreaming → audited semantic operations, through the real OpenCode Go `deepseek-v4-flash` target — 62 scoped capability calls, 8 cited operations applied, zero failures, persisted runbook, cursor advancement — including source-artifact provenance and the Obsidian disconnect/purge path preserving user-owned entities.
- MemoryBench: with corrected output budgets, the same lightweight model produced 378 capability calls, semantic aspect names, unknown entity-type rate down 39%→29%, and the target claim materialized as a semantic memory surfaced by hybrid recall.
- Honest limitations: the full daemon suite remains blocked by a pre-existing cross-file DbAccessor / native-worker isolation failure, and root lint by pre-existing formatting drift in untouched integration manifests. Neither is claimed green here.

## Follow-ups (not blocking this merge)

- Publishable absolute-quality benchmark scores: the profile persists raw episodic sessions and retrieval evidence; the LLM-judge phase is credential-blocked.
- Live temporal/compaction lineage and the full agent/visibility/project/session/harness scope-matrix validation.
- #1059's structural fix (pipeline worker offload/backpressure) is its own issue; this PR contributes the dreaming-side health gate and one-attempt-per-sweep bound.
- #903 temporal freshness prior coordination; #981 Pi harness salvage (session plumbing only).


NicholaiVogel:
#946 stops the old worker state machine from running only by abandoning its pending work. It does not drain, migrate, or schedule that workload. This should probably be fixed so memories aren't lost.

NicholaiVogel:
No legacy extract job may be retired unless its source is durably reachable by the Dreaming cursor, or it has a live consumer.

NicholaiVogel:
Addressed the extraction-job concern in be0df0b09: startup now promotes every live pending/leased legacy extraction source into the durable episodic Dreaming cursor before terminalizing its obsolete job. Deleted/missing sources remain withdrawn, and forgetting still dead-letters their jobs. Added focused coverage for all three cases.

NicholaiVogel:
One data-migration concern in `retireLegacyExtractionJobs`.

Migration 094 correctly classifies `source_type` values `extract`, `aggregate-recall`, `session_end`, and `checkpoint` as daemon-derived rather than episodic. The retirement path only excludes `m.type = 'session_summary'`, then promotes every other live extraction-job source to `memory_kind = 'episodic'`.

That can undo the migration's provenance boundary. For example, pre-cutover aggregate-recall rows have `type = 'semantic'` and `source_type = 'aggregate-recall'`, so a queued legacy extraction job would make an already-derived aggregate answer become Dreaming input. Old summary-derived facts similarly use `source_type = 'session_end'` but can have ordinary fact/preference types.

Can promotion use the existing `memory_kind` / `source_type` classification instead of `type`? Derived rows can be terminalized without promotion, with Dreaming consuming their canonical transcript, summary, or artifact evidence. A regression case for each of `aggregate-recall`, `session_end`, `checkpoint`, and `extract` would keep this provenance boundary intact.

NicholaiVogel:
Addressed in `d9f0ba2eb`.

`memory_kind` is now the provenance boundary: migration 094 and ongoing core writes share the same daemon-derived source-type classifier. `retireLegacyExtractionJobs` terminalizes every obsolete job and marks its source retired, but never changes `memory_kind`, so aggregate-recall, session_end, checkpoint, and extract rows remain derived while raw/manual evidence stays episodic and reachable by Dreaming.

Regression coverage now includes all four derived source types, migration 094’s checkpoint case, ongoing core writes, and the startup sweep. Focused migration/database/extraction-fallback/daemon-status tests and core+daemon typechecks pass.

NicholaiVogel:
Review pass over the post-pivot implementation — overall the sequencing and cursor design are strong, but three issues worth tracking before merge:

**1. Poison-pill evidence stalls Dreaming indefinitely for an agent**

In `runDreamingPass` (`dreaming.ts`), a single evidence item whose rendered size exceeds the entire prompt budget throws before generation:

```
Episodic evidence ${kind}:${id} exceeds the Dreaming prompt budget; split it into source artifacts before retrying
```

That fails the whole pass → `consecutive_failures` backoff (up to ~5.3h) → retry → same deterministic failure, forever. Dreaming for that agent is silently dead until someone reads the logs and manually intervenes, and the error message is developer-directed ("split it into source artifacts") — an end user just has consolidation permanently stalled.

Suggest: quarantine-and-continue. Dead-letter the oversized item (e.g. a `dreaming_excluded` marker on the source row, surfaced in `dreaming_passes` and status), advance the cursor past it, and let the rest of the backlog process. Loud in status, not fatal to the pass.

**2. Evidence with rejected operations is never revisited**

The apply loop correctly isolates bad ops via savepoints, but the cursor advances to `lastEvidence` regardless of `failed > 0`. For ordinary LLM garbage that's the right tradeoff, but a *systematic* validation failure — schema drift after a model/provider swap, a prompt regression — would silently drop evidence-derived semantics forever. The only visibility is a warn log (first 10 errors) and `mutations_failed` on the pass row.

Suggest: at minimum a persisted record of which evidence IDs produced failed ops (dead-letter table or a column on `dreaming_passes`), so a later pass or manual tool can requeue them. If that's out of scope for this PR, worth a tracking issue — this is a lossy-by-design path and it should be discoverable, not just logged.

**3. Stale comment on the retirement invariant in `daemon.ts`**

The comment above the `retireLegacyExtractionJobs` call still says:

> Promote the source of every pre-existing legacy `extract` job into the Dreaming cursor, then terminalize the job.

But after `d9f0ba2eb` the mechanism deliberately *never* promotes — `memory_kind` is the provenance boundary, and retired sources keep their existing kind so only genuinely-episodic rows remain cursor-reachable. The comment describes the superseded design on exactly the invariant that was redesigned twice; the next reader touching the provenance boundary will trust it. Suggest aligning it with the rationale comment in `extraction-fallback.ts`, which is accurate.

NicholaiVogel:
Follow-up review — architecture and behavior, building on the nits above. Summary: the cutover direction and boundary discipline are sound, but one mechanism should be fixed before merge, several signed spec decisions were silently dropped, and the pass needs to become agentic to deliver what Dreaming is for.

## 1. Failure backoff is volume-based, not time-based (fix before merge)

`shouldTriggerDreaming` implements backoff as `episodicTokens >= tokenThreshold × 2^failures`. The comment describes a time delay ("5min, 10min, 20min…"), but the mechanism is a volume gate, and it misbehaves in two ways:

- **Slow recovery from transient outages.** A one-day provider outage doesn't delay retry by hours — it requires 2×, 4×, 8×… the normal evidence volume to accumulate. At modest inflow, recovery takes days to weeks, and each failed retry doubles the requirement again.
- **Recovery can become unreachable.** After ~6 failures the requirement is 6.4M tokens (100k × 2⁶), but `getDreamingEpisodicTokenBacklog` samples at most 500 evidence items. For an agent whose backlog is many small memories/summaries, the sampled count can never reach the multiplied threshold — periodic passes stall permanently, recoverable only by manual trigger. A provider-misconfigured install walks straight into this, silently.

Recommend: wall-clock gate on a persisted `last_failure_at` (`now - last_failure_at >= 5min × 2^failures`), evaluated in the existing 5-minute tick. ~20 lines, survives restarts, no new subsystem. If a durable scheduler is later introduced for periodic maintenance (see §4), the backoff folds into it — but don't build scheduling infrastructure just for this.

## 2. Signed spec decisions the implementation dropped

The deleted `913-ingest-unification.md` spec recorded locked decisions (2026-07-12) that the cutover quietly abandoned:

- **dependency-synthesis was explicitly retained** as a periodic `maintain` lane ("it joins an entity's full history across batches/time, which a single batch cannot see" — locked decision #5). The PR retired it instead (`edbf9cd9`). See §4 for why this matters.
- **A quality gate was a signed success criterion**: "matches or beats the retired 5-stage chain on fact recall, entity precision, aspect-naming consistency, and dependency-edge F1." No such evaluation exists, and `docs/BENCHMARKING.md` was amended to *exclude* Dreaming from benchmark ingestion — routing around the comparison rather than making it.
- **Oversized evidence**: the spec required splitting at safe boundaries (turn/message/paragraph), never silent truncation. The implementation hard-fails the pass instead (the poison-pill issue from my earlier comment — it's also a spec regression).
- **Full 19-op ontology vocabulary** was the signed graph-write surface. The prompt currently bans `archive_claim_value`, `update_link`, and `archive_link` because the static snapshot lacks stable IDs — a limitation of the snapshot design, not of the apply path, which supports all 19 (see §5).

## 3. Benchmark for absolute quality, not parity

Parity with the retired pipeline is the wrong bar — its graph output was demonstrably poor. The benchmark should measure absolute quality, and the PR already built the measuring stick: quote-citation validation. Proposed metrics:

- **Citation coverage** — fraction of graph claims traceable to a verbatim evidence quote (the old pipeline's garbage was largely uncited).
- **Garbage rate on a fixed corpus** — junk entities, possessive-form duplicates, fragment names (the taxonomy compact mode already targets), comparing old-output snapshots vs. fresh Dreaming passes over the same evidence.
- **Retrieval uplift** — recall with the graph vs. without on held-out queries. A graph that doesn't improve retrieval is overhead regardless of citation rate.

Old output remains useful as a known-bad reference point, not a baseline to match.

## 4. Undisclosed behavioral change: maintenance is now event-driven only

Retiring the supersession sweep and dependency synthesis means **no autonomous graph maintenance happens without new evidence flow**. An agent idle for a month gets zero link discovery, zero supersession, zero hygiene. This may be the right call (cheaper, less LLM churn), and per the signed spec it was *not* the agreed call — but either way it is currently undocumented: MEMORY.md and UPGRADING.md never tell operators that periodic maintenance no longer exists. Graph stagnation will be reported as a bug. Fix is either honest docs or a scheduled maintenance lane.

## 5. The pass must become agentic (the core design gap)

Dreaming's purpose is maintaining what the agent already knows and mutating it with new discoveries. The current pass cannot do this: the model receives a static, row-limited graph snapshot (~20% of prompt budget; truncation only logs a warning) and must attach new knowledge to existing knowledge it mostly cannot see. The retired workers were compensating for exactly this blindness — dependency synthesis walked neighborhoods, the supersession sweep compared aspect siblings. Retiring them without giving the reasoner navigation ability removes the capability, not the redundancy.

Proposal: keep the cursor/evidence selection untouched; replace the single generation over a fixed snapshot with a bounded tool loop. One shared tool surface, two executor bindings:

**Read tools** (most exist today as CLI verbs or MCP operations, never packaged for the loop):
- `search_entities(query)` / `get_entity(name)` → aspects, claims, links
- `list_aspect_claims(entity, aspect)` — the supersession primitive: see siblings before writing the replacement
- `walk_links(entity)` — dependency discovery at attach time
- `get_claim_evidence` / `get_link_evidence` — provenance audit before trusting existing state
- `search_evidence(query)` — pull episodic context beyond the current window

**Write tool** (one): `apply_ontology_ops(ops[])` through `applyOntologyOperationBatchInTx` with the existing per-op savepoint isolation. The full 19-op vocabulary returns — `archive_claim_value`, `update_link`, `archive_link` included — because the model can now look up stable IDs instead of needing them in a snapshot.

**Executor parity**: the in-process Pi path gets these as `defineTool` definitions on an AgentSession; external ACPX harnesses get the same definitions exposed through **both** MCP tools on the daemon's existing server **and** CLI commands — the dreaming skill already teaches CLI-driven maintenance (`signet ontology …`, `signet knowledge …`), so harnesses that shell out get the identical surface either way. Same schemas, same validation, same audit across all three bindings — the external agent never touches SQLite.

**Recommend incremental apply over terminating-submit**: each `apply_ontology_ops` call validates and commits immediately so subsequent reads see post-write state (write claim B after superseding claim A, and `list_aspect_claims` reflects it). Atomicity is unnecessary — ops are idempotent and the cursor only advances at pass completion, so an aborted pass replays the window safely.

This also dissolves §4's maintenance problem rather than patching it: supersession happens when the model encounters a stale claim *at the moment it's writing the replacement*, and link discovery happens when it walks the attachment neighborhood. Epistemics move from blind periodic sweeps to reasoned, in-context maintenance.

**On PR #981**: it contains no graph-search capability — its two tools are "read the static context bundle" and "submit the plan." What's salvageable is session plumbing only: the `pi-provider.ts` AgentSession factory, the lifecycle wrapper (timeout/abort/dispose, unsupported-provider fallback, unexpected-tool verification), and their tests — ~350 hardened lines worth copying into a fresh file. Recommend closing #981 with a salvage note; it is stacked on the deleted ingest queue and cannot merge as-is, and leaving it open overstates how far the agentic work has gotten.

## Smaller items

- **First-ever pass runs in compact mode** — the most consequential pass for an upgrading install (full-history reconciliation) runs with graph-janitor instructions ("merge duplicates, delete junk") instead of evidence-integration ones. Backfill deserves its own mode, or incremental-with-full-window semantics.
- Backlog is computed twice per triggered check (once in `shouldTriggerDreaming`, again for the log line) — up to 500 full evidence rows tokenized twice per agent per tick.
- Agent enumeration is an 8-table UNION with DISTINCTs every tick.
- One global `active` flag serializes passes across all agents — bounded, but a 5-minute-timeout pass blocks every other agent's pass.

NicholaiVogel:
Third pass — an audit of the test mass itself, since tests are 65% of this PR's additions (+4,535 lines across 55 files, against −8,897 deleted mostly-retired worker suites). Repo culture goal: regression tests should ride bugfixes; tests written alongside unproven features risk locking in behavior rather than protecting it. By that standard the suite breaks down roughly as follows:

**Regression tests riding bugfixes (~1,450 lines, ~32%) — keep, obviously.** Cursor-across-idle-passes, empty initial cursor, archived-evidence exclusion, temporal dedupe, compaction lineage, the migration-094 provenance boundary. Each names a discovered failure and pins it. This is exactly the desired pattern and these commits are models of it.

**Contract/invariant tests (~2,400 lines, ~50%) — defensible, keep.** Episodic immutability ("rejects content changes on episodic memories"), agent isolation, purge-by-source, checksum rejection, "rejects an operation citing text absent from the rendered source," malformed-LLM-output parser robustness, cursor tie semantics. These pin *promises* rather than implementations — they fail when someone breaks a real invariant, not when someone improves the code.

**Behavior-enforcement of unproven behavior (~600–900 lines, 15–20%) — the problem fraction, and it clusters:**

- **The backoff tests** (`"backs off on consecutive failures"`, `"backs off first-run failures requiring threshold tokens"` in `dreaming.test.ts`) assert the volume-based backoff mechanism — the same mechanism flagged above as misdesigned. When the wall-clock fix lands, these tests go red, and the red will mean only "you changed the behavior we locked in." They should be rewritten alongside the fix, not preserved.
- **Trigger-threshold mechanics** (`"triggers at threshold"`, `"does not trigger below threshold"`) — arithmetic dressed as coverage.
- **Prompt-shape assertions** — ~47 `toContain`/`toMatch` checks in `dreaming.test.ts` pinning specific instruction wording. The prompt is the most volatile, least-validated surface in the system; these will fail on every wording improvement while proving nothing about whether the wording works.
- **Fixture realignment churn** across the 38 modified test files — status-surface and config tests edited to track the refactor.

**The missing test is bigger than any of these.** The suite verifies that the pass executes and operations land atomically; nothing anywhere asks whether the graph Dreaming produces is true, useful, or non-garbage. That's the eval harness from the previous comment (citation coverage, garbage rate on a fixed corpus, retrieval uplift) — until it exists, the system is tested but not verified.

Net: ~80% of the new test mass is honest work worth keeping. The ~20% behavior-enforcement fraction should shrink as the fixes above land — starting with the backoff tests, which will otherwise fight the first pre-merge fix.

NicholaiVogel:
Fourth pass — this one is a design requirement, not a defect: **the CLI, MCP, and in-harness (Pi AgentSession) tool surfaces must have 100% capability overlap.** Names may follow binding conventions, but every capability must exist in all three. Dreaming's operations require the legacy pipelinev2 worker toolset and the existing CLI maintenance toolset *in tandem*, and right now the work in this branch is creating a fourth partial surface instead of unifying them.

## The problem: four drifting surfaces

Today the same capabilities exist in different dialects, or not at all, depending on the door you walk in through:

- **Agent tools** (new in this branch): 8 navigation primitives + `apply_ontology_ops`
- **MCP**: `knowledge_list_entities`, `knowledge_get_entity`, `knowledge_list_claims`, `knowledge_hygiene_report`, … — overlapping intent, different names, no `walk_links` / `search_evidence`
- **CLI**: proposals, assertions, conflicts, hygiene, evidence verbs — the maintenance lifecycle, none of the loop primitives
- **REST routes**: yet another shape underneath them all

`knowledge_list_claims` (MCP) vs `list_aspect_claims` (agent tools) vs nothing (CLI) is one capability in two dialects and one absence. Three hand-maintained lists will never stay in sync — so don't keep lists.

## Proposal: one capability registry, three generated bindings

Define each capability **once** — id, description, Zod schema, agent-scoped handler over `DbAccessor` — then bind it mechanically:

- `defineTool` wrapper for the Pi AgentSession (in-harness dreaming)
- MCP operation registration (external harnesses over the daemon server)
- CLI subcommand (shell-driven harnesses and operators)

The capability id, schema, validation, and audit are shared **by construction**; bindings differ only in transport and output formatting. Overlap can't drift because there is nothing to drift *from*. If a binding ever needs an opt-out, a structural CI check flags the asymmetric capability id — an invariant check, not a behavior test.

## The registry contents: both lineages, in tandem

**Navigate** — the loop primitives (already in this branch): `search_entities`, `get_entity`, `list_aspect_claims`, `walk_links`, `get_claim_evidence`, `get_link_evidence`, `search_evidence`.

**Maintain** — the write surface: `apply_ontology_ops` with the full 19-op vocabulary, epistemic assertion CRUD (create/supersede/archive/link-claim — the dreaming skill already teaches assertions, but no tool exposes them), and proposal submit as the escape valve for risky/destructive changes the reasoner should not self-apply.

**Verify** — the pipelinev2 workers' *procedures* become consultable tools. Their judgment moves into the reasoner, but the deterministic machinery should be callable mid-loop:

- Decision engine's per-fact candidate search → `search_entities` / `search_evidence` (already present)
- Supersession sweep's sibling comparison → `list_aspect_claims` (already present)
- Contradiction machinery (antonyms/negation/token-overlap detectors that survive at head) → **`check_contradiction(entity, aspect, value)`** — consulted before `set_claim_value`, still enforced at apply time
- Entity-label gate (`shouldPersistEntity`) → **`check_entity_label(name)`** — consultable during reasoning, still mandatory at apply
- Duplicate detection → **`find_duplicate_entities(name)`** for merge candidates
- Hygiene + conflicts views → to drive compact passes and surface contested claims before writing

The subtle point: the spec was right that deterministic guards belong outside the model — at the apply boundary. But a reasoner that can *consult* them mid-loop emits fewer ops that get rejected, which directly shrinks the rejected-operation dead-letter pile.

**Session** — pass status/trigger (already in CLI and MCP).

## Why this matters for this PR specifically

The agent tools added here are the right primitives, but if they land as a surface that only the in-process Pi path can reach, the branch will have built exactly the divergence it set out to eliminate — an external ACPX harness shelling out to the CLI today cannot reach the navigation half of the surface at all. The registry is the difference between "three bindings of one toolbox" and "three toolboxes that used to look similar." The reasoner, the shell harness, and the MCP client should all be holding the same tools.

NicholaiVogel:
Fifth pass — scored against the design intent recorded when this work started (the 2026-07-21 design notes that fed #913). The architecture fits the mental model almost exactly: one agent wandering the graph instead of five narrow calls, daemon-owned selection/cursor/validation/writes with the executor as pure reasoning slot, citation-required operations, Dreaming as sole creator of the semantic layer. Three load-bearing ideas from the original model are missing, though — all three hugely important.

## 1. Meta-memory: the pass has amnesia

The design notes are explicit: *"each dreaming session should provide a log similar to how the routines in this vault have runbook memories which describe how each run has gone. it's sort of like a memory system for the memory system."*

Nothing of this exists in the pass. The registry has no `runbook_read` / `runbook_write` (the notes' own tool list includes them under Dreaming management), and the pass prompt carries **no history of previous passes** — no what-I-did, no what-got-rejected, no what-I-deferred. The infrastructure is half-built (`dreaming_passes` rows, and capability-call tracing landed in `193312470`), but the agent cannot see any of it.

Consequences: a pass that rejected a merge candidate last week re-considers it from scratch this week. A pass that deferred a risky change has no way to know it deferred. Rejected-operation evidence is now retained for requeue (good), but the *reasoning* behind rejections evaporates. The agent can learn the world but never learn its own track record — and for a system whose purpose is improving over time, that is the gap that compounds.

Ask: runbook capabilities in the registry — `runbook_read` (recent passes: ops applied/rejected with reasons, evidence windows, quarantines, open questions) and `runbook_write` (a short structured log the agent emits at pass end, stored with the pass row and surfaced to future passes). The pass prompt should seed the last few runbook entries, not leave the agent to discover them.

## 2. Deterministic guards must be agent-callable, not just apply-time

The notes' model: *"The deterministic parts (dedup, embeddings, significance checks, write gating) stay as library functions the agent calls internally."* The implementation makes them invisible — contradiction detection, dedup, and the entity-label gate exist only as apply-boundary enforcement. The agent emits ops blind and learns at rejection time.

Both things should be true: guards enforced at the apply boundary (as now) **and** consultable mid-loop as Verify-group capabilities:

- `check_contradiction(entity, aspect, value)` — the antonyms/negation/token-overlap machinery that survives at head, consulted before `set_claim_value` / `supersede_claim_value`
- `find_duplicate_entities(name)` — merge-candidate detection before `create_entity`
- `check_entity_label(name)` — the `shouldPersistEntity` gate, consulted before naming anything

A reasoner that can consult guards writes fewer ops that get rejected — which directly shrinks the rejected-operation dead-letter pile this PR just built retention for. The two mechanisms are complementary: the tool reduces failure frequency, the boundary guarantees safety regardless.

## 3. Temporal epistemics need an input channel, not a retrofit

The notes' `review_after` design (separate issue, but architecturally adjacent) needs two things from this work: the agent *setting* review dates when it creates temporal claims, and the pass *surfacing* due-for-review claims as input. The current pass context has exactly one input channel — new episodic evidence. The mental model has a second: **existing semantic state that needs attention**.

That channel is generic — due reviews, hygiene reports, contested claims, and quarantine requeues all arrive through it — and it does not exist. Right now "the pass wakes up to new evidence" is hard-coded; "the pass wakes up to things needing attention" has no home. Much cheaper to add the extension point now (an attention queue/table the pass selects alongside the evidence cursor, populated by retention/expiry/hygiene writers) than to retrofit it after `review_after` lands on a context design that can't see it.

## Secondary, same theme

- **The static graph snapshot survived into the agentic prompt** — a 20%-budget truncated `knowledge_graph` section, the exact partial-view mechanism the tools replaced. With `search_entities` available it is redundant context spend at best and misleading-at-truncation at worst. Navigate, don't dump.
- **No source-backed memory writes in the registry.** The notes' tool surface includes `memory_store` / `memory_modify` / `memory_forget` (source-backed only); the dreaming skill already instructs external agents to save source-backed artifacts when evidence warrants durable recall. The agent currently has graph writes only.
- The registry's Maintain and Verify groups from the previous comment (assertion CRUD, proposal submit, hygiene, conflicts) remain unregistered — the capability registry currently holds only the 8 navigation/write primitives.

None of this blocks the cutover mechanics, which are sound. But items 1–3 are the difference between an agentic pass that *executes* the dreaming loop and one that *is* the dreaming system from the notes — a reasoner that maintains what the agent knows, learns from its own history, and keeps knowledge current as time passes.

NicholaiVogel:
Sixth pass — forensics from the first real benchmark run (`signet-dreaming-graph-on-longmemeval`, LongMemEval container `001be529`, temp daemon workspace inspected directly). The headline is good: **the agentic loop genuinely works end to end**. The pass completed with 51 ops applied across 72 tool calls, and the trace shows real navigation — `search_entities("Chicago")`, `search_entities("McGill")`, `search_entities("orchid")` before any writes. The semantic layer exists: 17 entities, 17 aspects, 26 attributes, 8 dependencies from a single pass. The citation validator rejected `create_entity: Chicago` five times with "Every operation must cite an exact quote from scoped episodic evidence" and the model revised until it produced valid citations. Quarantine, cursor, runbook machinery all fired.

But the run also exposes four real problems, in descending order of severity:

## 1. The bench ingest path pollutes `memories` with raw transcripts

The provider ingests each session via `POST /api/memory/remember` with the raw transcript as `content`. Result: 51 memory rows of 4–12k chars, typed `preference`/`fact`/`rule` by the type-inferrer — raw transcripts masquerading as curated memories. They are `memory_kind='episodic'` (so Dreaming correctly treats them as evidence, and the daemon separately captured all 51 `session_transcripts`), but the damage lands in **retrieval**: those raw-transcript rows are first-class hybrid-recall candidates, and they are exactly what the benchmark's search returned. Raw transcript answered the question because recall searched a polluted table. The bench must ingest sessions through the transcript/session surface, not the remember endpoint — otherwise every dreaming benchmark measures retrieval against contamination regardless of semantic-layer quality.

## 2. One pass, one window: coverage is structurally partial

The container corpus is ~123k tokens of transcript; bench config gives the pass `maxInputTokens: 64000`; the pass consumed 24k tokens total. `finalizeIngest` triggers exactly one incremental pass, waits, then searches — the cursor is parked mid-corpus with the remaining sessions still pending. The benchmark question ("How long did I wait for the decision on my asylum application?") concerns content that never entered the window: **zero asylum coverage** in entities or attributes. Retrieval didn't fail to surface the semantic layer; the layer didn't exist for that topic yet. `finalizeIngest` should loop passes until the cursor drains (or trigger a bounded N), not fire exactly once.

## 3. The semantic structure that did get built is shallow — and the model is the variable

14 of 17 entities typed `unknown`; 16 of 17 aspects named literally `profile` — a dumping-ground ontology. The pass ran on deepseek-v4-flash at `reasoning: low` with an 8k output cap. The machinery was correct; the brain produced minimal-effort structure. The dreaming model needs to be a benchmark dimension (flash/low-reasoning vs. headroom model), with "profile-dump rate" and `unknown`-type rate as first-class metrics in the quality report alongside citation coverage and garbage rate — otherwise shallow structure passes every current check while being nearly useless under retrieval.

## 4. Citation friction: 14/65 ops failed (21%)

All exact-quote rejections, all eventually resolved by retry. With a flash-tier model that's plausibly sloppiness; if the rate stays high with a strong model, the exact-quote validator is too strict for aggregate-supported entities (an entity like "Chicago" is supported by aggregate context, not always one quotable line). Track the failed-op rate per model in the quality report before loosening anything — the validator is currently the best thing in the system and should not be weakened on anecdote.

## Net

The system behaved correctly end to end; the bench gave it half the corpus and a model not worth listening to, then let retrieval answer from the wrong table. Fixes in order: bench ingest via the transcript surface, cursor-draining finalize loop, dreaming-model ablation with structure-quality metrics, and citation-failure tracking. None of these touch the cutover mechanics — they're the difference between a benchmark that proves the pipeline runs and one that proves the semantic layer is worth having.

NicholaiVogel:
Reproduced and confirmed.

The benchmark path currently creates 51 `memories` rows, 51 `session_transcripts` rows, and zero `memory_artifacts` rows. The raw sessions are reaching retrieval through `memories`, so the previous retrieval result is not evidence of semantic uplift.

I also confirmed the pass covered only part of the corpus. The cursor remained mid-window, and the graph had no asylum-duration claim. The run did exercise the agentic loop, but it is not a valid quality result.

I pushed two preparatory fixes after rebasing on current main:
- `9dadf1a3a` preserves Pi catalog metadata for compatible routed targets, including OpenCode Go's tool/thinking protocol behavior.
- `31192feb7` fixes the benchmark polling the `default` Dreaming status while triggering `memorybench`.

Next I am replacing the benchmark session ingest with the canonical episodic/session artifact path, then changing finalization to drain the Dreaming cursor before retrieval. I will not count raw transcript recall as a Dreaming result.

NicholaiVogel:
Addressed in `fc6343aab`.

The Dreaming benchmark no longer posts raw sessions to `/api/memory/remember`.

- It now uses the canonical `session-end` transcript capture path only, with a distinct benchmark session identity and the source session's original `capturedAt` timestamp.
- Capture is an explicit receipt: the benchmark waits for its agent-scoped transcript-capture job to finish before it can trigger Dreaming.
- `finalizeIngest` now runs bounded incremental passes until `/api/dream/status` reports `episodicTokensPending: 0`; incomplete cursor coverage fails instead of proceeding to retrieval.
- The canonical session path preserves the historical timestamp in both `session_transcripts` and `memory_artifacts`; the regression test also proves it creates zero `memories` rows.
- Added an agent-scoped capture receipt endpoint, SDK types, and API documentation.

Focused proof passed:

- `bun test memorybench/src/providers/signet/index.test.ts`
- `bun test platform/daemon/src/hooks.test.ts platform/daemon/src/transcript-capture-worker.test.ts`
- `bunx tsc -p platform/daemon/tsconfig.json --noEmit`
- `bun run --filter '@signet/sdk' build`
- `bun run --filter '@signet/sdk' test`

The next live benchmark will be judged only after raw-memory count is zero, transcript capture is complete, and the Dreaming cursor is drained.

NicholaiVogel:
One follow-up caught during live-path review: the first change gave the benchmark a capture-completion barrier, but MemoryBench invoked provider finalization at the end of *ingest*, before its indexing phase waited on that barrier.

`c0fa9959c fix(memorybench): finalize dreaming after capture indexing` moves provider-wide finalization after indexing. Dreaming now cannot see a partial corpus caused by deferred transcript capture. The ordering is now:

1. canonical session capture receipts are enqueued;
2. indexing waits for every receipt to complete;
3. Dreaming drains its episodic cursor;
4. retrieval begins.

This is the required ordering for the benchmark to mean anything.

NicholaiVogel:
Live rerun update:

The corrected benchmark ingest is now proven at runtime in a fresh workspace:

- `memories = 0`
- `session_transcripts = 51`
- `memory_artifacts = 153`
- all 51 transcript-capture jobs completed
- historical transcript timestamps remained in the original 2023 range

That fixes the raw-transcript-as-memory contamination.

The first corrected pass completed with 31 semantic operations applied, then the new cursor-draining loop started a second pass as intended. That exposed two further runtime blockers, now addressed in the branch:

- `dc66378ea` normalizes a pasted OpenAI `.../v1/chat/completions` endpoint to Pi's required base URL. The original full URL otherwise became a doubled path and returned 404.
- `fa2f89d8e` defers an embedding-index worker restart until an active Dreaming pass completes, preventing the previous orphaned-pass failure.
- `9e3a22f66` makes benchmark pass timeout and output cap configurable; the old 8k per-turn cap produced a Pi `length` termination mid-tool-loop, and the prior 10-minute timeout clamp made the timeout knob ineffective.

I have not counted this as a successful quality run yet: the second pass hit the old output cap before its cursor drained. The next fresh run uses the raised cap/time budget and will remain invalid until it reaches zero episodic backlog.

NicholaiVogel:
Fresh end-to-end benchmark update (`001be529`, `deepseek-v4-flash`, full 51-session corpus):

- canonical ingest stayed clean: `memories=0`, `session_transcripts=51`, `memory_artifacts=153`;
- finalization drained the cursor (`episodicTokensPending=0`) across five completed passes;
- Dreaming produced 59 entities, 59 aspects, 145 attributes, 25 dependencies, and 229 applied operations; 19 operations were rejected and retained/quarantined rather than silently lost;
- the temporary dashboard is now backed by that semantic graph. I found its constellation query was still filtering out zero-mention/non-`person|project` entities, which hid every directly-derived Dreaming entity even though the graph rows existed. `a03000fa2` removes that legacy-memory display filter while preserving the bounded query, with regression coverage. The live endpoint now returns all 59 entities, their aspects, attributes, and 25 dependencies.

This is a successful cutover/runtime mechanics proof, not a quality sign-off: this flash run made 23/59 entities `unknown`, 55/59 aspects `details`, and did not derive an asylum-duration claim. The current quality report needs to be run against a stronger model / model matrix before treating graph retrieval as useful. I am keeping the citation validator intact (the 19 rejected operations were exact-quote failures) and will treat this as the next quality-work input, not paper over it as a successful answer.

NicholaiVogel:
Follow-up on the benchmark-quality portion of the review:

- `90f353bdf` adds `structureQuality` to the scoped `/api/dream/quality` report: unknown-type rate, exact `profile` rate, and a broader generic-aspect rate (`profile`, `details`, `general`, `information`). These are local deterministic measures for a real model matrix, not prompt-string proxies.
- On the completed flash run, the report is now explicit: citation coverage `1.0` (145/145 active claims resolve to episodic quotes), garbage `0.0` by the existing name classifier, but unknown entity types `23/59` (`0.390`) and generic aspects `55/59` (`0.932`). That confirms the graph is mechanically sound but semantically shallow; it is not an acceptable retrieval-quality result.
- Review also caught one display-boundary leak in the constellation change: source-native topology could have been admitted with semantic rows. `40f323cd5` extracts the shared source-topology entity-type constant and uses it in both quality and constellation selection, with the constellation still showing zero-mention semantic entities but not source folders/documents/skills.

Focused core build, daemon typecheck, constellation tests, and quality-report tests pass. The retained benchmark daemon is available at `http://127.0.0.1:35421/` with `memorybench` as its default scoped agent for manual graph inspection.

NicholaiVogel:
Follow-up on the semantic-memory display/retrieval gap from the live dashboard review:

- `d0923f4ff` makes every Dreaming-created claim attribute materialize as one derived semantic `memories` row (`memory_kind=NULL`, `source_type='dreaming'`) linked 1:1 through `entity_attributes.memory_id` and `memory_entity_mentions`.
- Raw transcripts remain exclusively episodic (`session_transcripts` / `memory_artifacts`); this does not reintroduce them into the memory surface.
- The claim value contract now explicitly requires a standalone atomic assertion (for example, `Signet is a memory system`, not `memory system`) and rejects generic aspect dumping in the prompt.
- Migration 102 backfills existing attributes to the same projection, so the retained benchmark workspace now exposes 145 derived semantic memories and no raw transcript memories.
- `f748d76fc` synchronizes the projection through add/set/supersede/archive/restore, including forced archival and rollback-safe errors instead of graph/memory lifecycle divergence.

Focused core build, migration tests, ontology-proposal lifecycle tests (48), Dreaming tests, daemon typecheck, and a Luna lifecycle review passed. I have started a fresh full-corpus DeepSeek run with the changed prompt; I will report its structure quality separately rather than claiming that historical shallow attributes improved retroactively.

NicholaiVogel:
Fresh post-prompt runtime result (`deepseek-v4-flash`, LongMemEval `001be529`, full 51-session corpus) is complete and inspectable at `http://127.0.0.1:35422/` on this machine (`memorybench` scoped agent).

The important architectural outcome is now proven at runtime:

- 51 canonical `session_transcripts`, 153 episodic artifacts, and **0 raw episodic memory rows**;
- cursor fully drained across five passes;
- 147 retrievable rows, all `type=semantic`, `source_type=dreaming`, `memory_kind=NULL`, each linked 1:1 to its active graph attribute;
- all 147 are standalone sentence-like assertions rather than transcript fragments;
- quality report: 34 semantic entities, 58 aspects, 147 attributes; zero generic/profile aspects and zero name-garbage entities. Unknown entity types are still 10/34 (29.4%), so this is an improvement in shape, not a final quality sign-off.

The runner also exposed a real lifecycle race: an embedding-triggered pipeline restart could start the next periodic pass between benchmark polls, causing finalization to treat the expected `409 already running` as a run failure. `4623dac9a` makes the benchmark join that active pass and adds focused coverage. The completed run above exercises that recovery across all five passes.

`b51434117` also closes the derived-memory source-purge gap found while hardening the new projection: source/file disconnect now soft-removes the linked semantic memory before graph row deletion, with agent-scope verification and source-purge regression coverage.

NicholaiVogel:
Seventh pass — verified against the fresh benchmark workspace (`001be529`, full corpus, cursor drained). The trajectory is now unmistakable, and one earlier review assumption needs a public correction:

**The model was never the bottleneck.** This run used the identical configuration that produced the shallow graph — deepseek-v4-flash, `reasoning: low` — and the output transformed anyway: aspect names are semantic (`career`, `writing themes`, `techniques`, `sustainability plan`), unknown-type rate dropped 0.39 → 0.29, navigation depth went 72 → 378 tool calls, and the pass derived the target temporal claim: `claim_key=asylum` — "asylum application was approved after over a year of uncertainty." The 8k output cap was the straitjacket; generic aspects were the model economizing under budget pressure, not failing. Earlier comments framed model quality as a primary variable — the evidence now says output budget and harness mechanics were binding first. Model matrix remains useful, but as a second-order ablation.

**Attribute materialization is the right retrieval answer.** Writing claims through as `type='semantic'` memory rows (`d0923f4ff`, migration 102) puts the semantic layer directly into hybrid recall's candidate pool — no special retrieval path, no constellation-only visibility. And the follow-through shows the design commitment is understood: `f748d76fc` synchronizes the attribute↔memory lifecycle, `b51434117` purges derived rows with their sources. Semantic truth living in two stores is real duplication, and it's being managed as such rather than discovered later. That is the correct posture.

**Three items remain:**

1. **The interface lever is still unpulled.** `entity_type` in `dreaming-operation-contract.ts` remains `text.optional()` — free text, no enum, no `.describe()`. Aspect names improved without it (budget was the bigger constraint), but the residual 29% unknown-type rate has a direct mechanical fix: enum over `ENTITY_TYPES` plus described payload fields. The schema is where structure gets taught; right now it's silent on types.
2. **`memory_kind` classification wrinkle.** The 147 materialized semantic rows carry `memory_kind=NULL`. Migration 094 made `memory_kind` the provenance boundary (episodic vs. daemon-derived); NULL is a third implicit state. They're correctly excluded from episodic selection today, but only by accident of the NULL comparison — classify them explicitly (`semantic`/`derived`) before some future selector treats NULL as wild.
3. **The retrieval proof is still one query away.** The semantic row exists; what hasn't been shown in this workspace is the asylum question's recall actually surfacing it above transcript content. That's the acceptance test the whole thread has been pointing at — run it and paste the ranked results.

Also noting `4623dac9a` (join concurrent dreaming passes) — the bench was triggering overlapping passes; worth confirming the single-active-pass invariant still holds under the daemon's own worker, not just the bench's finalize loop.

None of this blocks the cutover. The system now does the thing it was built to do, measurably, on the hard case.

NicholaiVogel:
Addendum to the previous comment — the prompt-side feedback, which the run data now lets us aim precisely. The context-engineering principles behind this (Anthropic's current guidance for the newest model class): they removed ~80% of Claude Code's system prompt with no measurable loss; rules meant to prevent worst-case scenarios get deleted in favor of model judgement; examples constrain exploration space instead of helping; tool instructions belong in tool descriptions, not repeated in prose; system prompts earn their tokens on product context and goal, nothing else. Measured against that, `buildDreamingPrompt`'s instruction layer is the old playbook:

**1. Slash the guidelines, don't amend them.** The `Guidelines` block and both `modeInstructions` variants are worst-case-scenario rules: conservatism, merging etiquette, naming etiquette, "be conservative — only change what you're confident about." None of it names a failure this system actually has. The one with a measured cost is *"work from the episodic evidence toward a **small set** of material, high-confidence changes; **do not inventory every named entity**"* — a coverage cap written into the task, visible in backfill runs where extraction should be near-exhaustive. Delete the block; let the model judge scope from evidence volume. If shallow output returns after the cut, that is a measurement, not a emergency — `structureQuality` will say so.

**2. Move the closing paragraph into the registry.** The entire final instruction paragraph (search before writing, inspect siblings before superseding, `apply_ontology_ops` for every write, exact-quote citations, call `runbook_write` once) is tool-usage prose duplicating what tool descriptions should say — and `dreaming-capabilities.ts` now exists precisely to hold those descriptions. One statement, one place: each capability's description carries its own usage; the prompt stops repeating it.

**3. Enrich the contract, skip the exemplar.** Resist adding example entities or "never name aspects `profile`/`details`" rules — examples constrain exploration to the example's shape, and worst-case rules are the category being deleted. The expressive-interface alternative: `entity_type` as an enum over `ENTITY_TYPES` (still free-text optional today), `.describe()` on aspect/name/claim payload fields ("the domain of knowledge this claim belongs to"). The schema teaches structure silently; the 29% residual unknown-type rate is the metric that confirms or refutes it.

**4. Spend the recovered tokens on product context and goal.** What the prompt should keep — and slightly expand — is the layer that tells the model what this *is*: whose life the evidence is about (name the user from the resolved identity), and what dreaming is for ("you are maintaining what you know about {user} as their life changes over time" — which carries the temporal ambition in the goal statement instead of a mechanics rule). Identity, evidence, runbook, attention stay as-is; those blocks are already well-formed, and the injection guards on runbook/attention are exactly right.

The ablation writes itself: same model, same container, prompt slimmed + contract enriched, compare `structureQuality`, unknown-type rate, and asylum-claim derivation. The prediction from the principles is the prompt gets shorter and the graph gets better; if it doesn't, the quality report is the honest referee either way.

NicholaiVogel:
Eighth pass — legacy graph cleanup. I ran the numbers on a real production database carried through the pipelinev2 era (~51k memories, pre-cutover schema), and it changes this from "nice to have" to "the cutover inherits a landfill." The maintenance machinery question has a concrete answer: the bones are built, two gaps remain, and the legacy data proves both matter.

## The autopsy (production DB, old schema)

| Measurement | Value | Share |
|---|---|---|
| Entities | 33,206 | — |
| **Entities with zero attributes** | **27,402** | **82.5%** |
| Dependency links | 118,938 | — |
| Links typed `related_to` (catch-all) | 68,768 | 58% |
| **Links touching a zero-attribute husk** | **65,732** | **55%** |
| Junk-name entities (markdown, file paths, UUIDs, possessives, code fragments, sentence-length) | 2,935 | 8.8% |
| Bulk-typed entities (`artifact`/`system`/`extracted`) | 20,231 | 61% |
| Generic aspects (`general` 3,549, `properties` 1,410, `Overview` 640, …) | ~5,900 | ~25% |
| Attributes with **empty claim_key** | 20,801 | 29% |
| Positional claim keys (`overview_0`…`overview_N`) | ~3,500 | — |
| Exact duplicate-name groups (pre-fuzzy) | 133 | — |
| Attributes ever superseded | 4,201 | 5.8% |
| Memories typed `fact` (extraction firehose) | 39,346 | 77% |

The samples are worse than the aggregates: entities named `Promise<void>>`, `e6ae9515-9163-48b7-…`, `memory/2026-03-05T00-10-32…--summary.md`, `DREAMING_CAPABILITIES`, `evidence_requeue attention` — the pipeline extracted *the system itself* into the graph. And 82% of entities are empty husks connected by 65k meaningless links: the graph's topology is mostly noise about nothing. The supersession sweep, in its entire lifetime, retired 5.8% of attributes.

## What already exists (verified at head)

The maintenance spine is ~85% built: `dreaming_attention` kinds include `hygiene` and `review_due`; `shouldTriggerDreaming` fires on attention alone (`return hasAttention || …`), so the worker wakes with zero new evidence; the idle gate lets a pass run attention-only; agent enumeration includes agents with unresolved attention; the prompt renders attention with correct injection guards; compact mode exists; the `qualityIssues` garbage classifier already knows what junk looks like. Wake-up, work queue, prompt surface, junk detection — all present.

## Gap 1: `hygiene` has no producer

The only attention writer at head is the quarantine `evidence_requeue` path. Nothing scans the graph and enqueues `hygiene` items — the kind exists with zero producers. The fix is small: after each pass (or on the worker tick), run the deterministic detectors (garbage classifier, duplicate-name groups, zero-attribute husks, empty claim keys, `related_to` link spam) and `enqueueDreamingAttentionInTx({kind: "hygiene", subjectRef})`. For legacy graphs, a one-time seeding scan at migration/startup. The quality report then doubles as the cleanup scoreboard — garbage rate declines measurably as items are consumed.

## Gap 2: the citation wall makes legacy cleanup impossible

`provenanceForEvidence` returns null for any operation with zero citations, and null is a hard rejection **for every operation, including `archive_entity`**. The 27,402 husks have no episodic evidence to quote — the agent literally cannot delete them through the apply seam. The validator that rightly protects new semantics also forbids removing old garbage, and a first-run backfill on a legacy graph will face exactly this: compact instructions saying "delete junk entities" against an apply seam that rejects every archive.

Principled opening, narrowly scoped: **destructive ops may cite an attention item as provenance** (`archive_entity`, `archive_aspect`, `archive_claim_value` with `provenance: attention:<id>`). The audit trail records the real justification — deterministic detector flagged it, agent inspected and agreed, item resolved — instead of pretending evidence exists. Content-bearing ops (create/set claims) keep the strict quote requirement forever: never uncited claims, always-traceable deletions. The validator stays strongest exactly where it matters.

## Why this can't wait for a follow-up PR

Every existing install carries a graph like the one above. The cutover's first backfill pass on a legacy database will reason over 82% husk topology, burn its budget on noise, and produce new semantics entangled with 65k meaningless links — while being mechanically unable to clean any of it. Two additions close it: hygiene producer (deterministic scan → attention), and attention-as-provenance for destructive ops only. Everything else — the wake-up, the loop, the prompt, the metrics — is already here.

NicholaiVogel:
Addressed the legacy-graph hygiene concern in `3157406fe`.

- A shared, bounded deterministic detector now finds semantic-only zero-attribute entities, suspicious labels, empty claim keys, generic aspects, duplicate canonical-name groups, and generic `related_to` links. It excludes source-native topology.
- The Dreaming worker seeds this cleanup attention at startup and refreshes it on each tick, including when no new episodic evidence exists. Resolved deterministic items stay resolved unless explicitly requeued.
- Destructive maintenance can cite only its exact scoped `attention:<id>` (`archive_entity`, `archive_aspect`, `archive_claim_value`, `archive_link`, or a duplicate-group `merge_entities`). Content-bearing operations still require exact quoted episodic evidence.
- The audit evidence records the attention id, subject, and detector details; cross-agent, unrelated-target, and content-creation attempts are rejected.

Focused proof: 38 Dreaming/hygiene/agent-tool tests and daemon typecheck pass. This includes exact-target coverage for entity/aspect/attribute/link archival and duplicate merges.

NicholaiVogel:
Ninth pass — two design pieces that came out of tracing how passes are queued and where identity comes from, cross-checked against #1059.

## 1. Identity: drop the implicit USER.md requirement; the invariant is entity-scoped claims

Identity is file-anchored and re-read per pass (`resolveStartupIdentityFiles` → AGENTS/SOUL/IDENTITY/USER/MEMORY with budgets, plus `kind: "dreaming"` specials). The code already tolerates missing files — `readIdentityFile` warns and returns `""` — so USER.md is optional in practice. But the prompt's framing still implies a global user profile, and that's worth killing deliberately rather than accidentally.

The invariant that actually matters is already the schema's shape: **every claim is a subject-scoped triple** — attribute → aspect → entity. "Entity X knows Y about Z" is `entity_attributes(entity_aspect(entity))`. The user is not a special global subject; he's an entity like any other, found via `search_entities` and accumulated under the same citation/scoping rules as every other subject. USER.md becomes what it should be: a bootstrap prior about who that entity is, never a load-bearing input. Multi-agent uniformity comes free — agents with identity files and agents without run identical graph logic.

Prompt change: stop implying a user profile exists; instruct that user-related knowledge attaches to the user entity like any other, with the same quote-cited evidence requirements. Identity files, when present, are context for judgment — never schema.

One guard this implies, straight from #1059's evidence: the issue documents dreaming running ~60 passes with `ENOENT: /tmp/MEMORY.md` (wrong agentsDir) — identity completely broken, yet passes logged "produced semantic operations" and wrote to the graph anyway. Degraded identity is survivable when entity-scoping is the invariant; *silently misresolved* identity writing to the graph is not. Identity-file read errors during prompt assembly should mark the pass identity-degraded in the runbook, or fail closed. Never blame the model — but don't hand it a broken anchor and let it write.

## 2. Scheduling: three wake conditions + one health gate, all on the existing scan

Current mechanics (verified at head): no queue table — the evidence backlog *is* the queue. The worker sweeps every 5 min, enumerates agents from any data presence (including unresolved attention), and fires per agent on `hasAttention || backlogTokens >= tokenThreshold`. Single-flight globally. Ticks are always incremental; compact is manual-only.

I'd keep that shape and extend the trigger to a disjunction of three conditions with distinct urgency classes:

1. **Ceiling — `tokens >= threshold`** (exists). Batch-size bound; keeps backlog and prompt size bounded. Urgent.
2. **Floor — `backlog > 0 && now − lastPass > maxInterval`** (missing). A staleness bound (~6h). Without it, a slow trickle of evidence sits just under threshold forever and the episodic ledger never consolidates. This is a fixed schedule shaped as *maximum wait*, not blind cron — it never fires on an empty backlog, and the idle no-op gate already makes near-empty passes cheap.
3. **Attention — `hasAttention`** (exists). Event-driven wake for hygiene/review_due work, which never generates tokens. This is what services legacy-graph cleanup (see the previous comment — 82% husk entities don't produce an evidence backlog).

Then, the #1059 alignment: **a health gate in front of all three.** That issue names the dreaming worker as a secondary feeder of the death spiral — multi-pass bursts every 5 min while the watchdog was SIGKILLing the daemon every ~6 minutes. The core lesson: dreaming is the *most deferrable work in the system* — nothing user-facing blocks on it, a pass delayed an hour loses nothing — so it should be the first thing to yield, and today it's the opposite. Add a pressure check to the sweep: pipeline pending jobs above a watermark, recent event-loop-blocked monitor events, or a recent watchdog restart → skip the sweep entirely. One predicate, same scan.

This doesn't preempt #1059's structural fix (worker offload, drain throttle, cooperative yields — all still needed); a load-aware dreaming scheduler is compatible with every variant of it. It converts dreaming from a feeder into a citizen that automatically backs off under pressure. I'd also cap passes per sweep — the bursts came from every backlogged agent firing per tick; one pass per tick is plenty at 5-minute cadence.

The shape that emerges: ingestion pressure pushes dreaming *later*, backlog size pushes it *sooner*, staleness guarantees it *eventually*, attention lets maintenance *cut in line*. ~40 lines on top of what exists, and none of it builds against #1059's direction.

NicholaiVogel:
Tenth pass — verified the two new commits, then two production reports from #913 that fold in.

## Commit verification (both faithful, one better than proposed)

**`2e078a8ef` (slim prompt + enrich contract)** — verified against the context-engineering principles: `entity_type` is now `z.enum(ENTITY_TYPES).describe(...)` — the model learns valid types from the interface, not from prompt rules. Field-level `.describe()` annotations carry the semantics ("A complete atomic assertion that is understandable on its own" — claim-shape guidance now lives in the schema where it can't be clipped). The ~36-line guidelines/mode-instructions block is replaced by one vivid goal sentence. Citation enforcement didn't move into the prompt because it didn't need to — the apply seam still hard-rejects uncited ops. Prompt persuades, validator enforces: correct split. Migration 103 resolves the `memory_kind=NULL` concern with the right boundary comment — attribute projections are semantic state, never primary evidence, and now they're explicitly `derived` before later readers grow beyond the episodic-only predicate.

**`3157406fe` (queue auditable graph hygiene)** — both gaps closed. The producer (`getDreamingHygieneCandidatesInDb`) has six deterministic detectors, and they're precisely the ones the production autopsy called for: zero-attribute entities (the 82.5% husks), missing claim_keys (the 20,801 empty keys), duplicate canonical names, generic aspects, junk names, generic `related_to` links — all scoped to semantic entities, excluding source-native topology and source-root rows, so the provenance boundary holds. Wired at startup and per-tick per-agent before the trigger check, so `hasAttention` alone wakes the pass — legacy graphs with zero new ingestion get serviced. And `attentionProvenance()` is **better than the proposal**: destructive ops citing `attention:<id>` must target *exactly* the flagged subject — `archive_entity` must name the flagged entity, `merge_entities` must keep target and sources inside the flagged duplicate group. An attention citation can't be laundered into a general delete token. Content-bearing ops keep the strict quote requirement. This is the validator staying strongest exactly where it matters.

## Folding in two production reports from #913 (both from @aatifk, shared bank across three harnesses)

**1. Workload routing — declined by architecture.** @aatifk reports dreaming resolves its model through a shared inference workload and proposes a dedicated `dreaming` routing kind. Worth considering under the old pipeline, but the cutover removes the frame: there is no extraction/synthesis split anymore. Signet provisions inference one way — the router over pi-ai/ACPX targets — with a single carve-out for **aggregation** (speed-sensitive; a different thing from the old session synthesis). Dreaming runs on the shared provisioning path by design; there are no per-task model splits to dedicate.

The pain behind the report is real, but it decomposes to call parameters, not provisioning. The bench forensics showed output budget was the binding constraint in the failed runs, and `DreamingConfig` already carries dreaming's own `maxInputTokens`/`maxOutputTokens`/`timeout` at head, passed per-pass as `maxTokens: cfg.maxOutputTokens` — the 8k→32k fix that produced the breakthrough was exactly that knob. Residual, if a future model needs it: a reasoning-effort field on `DreamingConfig` (a call parameter, not a provisioning change).

**2. Stale-derivative invalidation — this one slots directly into the machinery that just landed.** His repro: corrected a source memory twice via `memory_modify`; 40 minutes later recall still ranked the stale extracted derivative *above* the corrected source; contradiction window unbounded. The modify path at head has no supersede/invalidate logic.

He suggested a modify hook triggering a bounded invalidation pass. With the hygiene machinery in place, the right-shaped receiver already exists: **on explicit semantic-row update, enqueue a `contested_claim` attention item per dependent derivative** (the 094 provenance model knows which rows derived from the modified source). The worker already wakes on attention alone; the agent inspects the derivative against the corrected source and supersedes or archives with judgment; the audit trail records why. No bespoke invalidation subsystem — one more producer feeding the channel that `3157406fe` just proved out.

Two seams would need to open slightly: `attentionProvenance()` currently accepts only `kind = "hygiene"` — derivative invalidation wants `contested_claim` admitted for `archive_claim_value` against the flagged attribute, with the same target-exact validation. And retrieval ranking needs the conflict rule: on content conflict, prefer the source row over its own snapshot derivatives — the manual source is the higher-authority record by construction. That rule is deterministic, cheap, and defensible even *before* the attention item gets serviced, which closes the contradiction window at the read path instead of waiting for the pass.

Both reports are the same lesson as the rest of this thread: the machinery generalizes. Hygiene was the first producer; derivative invalidation is the second; `review_due` temporal sweeps are the third. The channel was the right investment.

NicholaiVogel:
Follow-up hardening for the hygiene work is now in `9ea05fb53` (rebased on current main and pushed).

The review found and I fixed three destructive-path issues before moving on:
- attention provenance now rejects selector aliases and `force` exactly as the ontology applicators interpret it;
- pinned entities and their owned graph state are excluded from hygiene candidates and duplicate merges;
- duplicate attention carries a bounded membership digest, so a same-sized but re-formed duplicate group reopens rather than becoming permanently unactionable.

A final adversarial review found no remaining actionable hygiene authorization, scoping, SQL, or lifecycle issue. Focused Dreaming/hygiene/tool tests (41) and daemon typecheck pass.

I have also ingested the newer scheduling, identity, and stale-derivative feedback. I am tracing the existing source/derivative linkage and scheduler health seams before changing either; I will keep this on the same shared attention path rather than adding a parallel invalidation or provisioning system.

NicholaiVogel:
Scheduling feedback is now in `8cbf10128` (rebased on `origin/main` and pushed).

- `memory.dreaming.maxInterval` defaults to six hours, so non-empty low-volume episodic evidence cannot wait indefinitely below `tokenThreshold`.
- The existing shared queue-health signal gates the periodic sweep before any attention or evidence trigger runs; manual triggers remain deliberate overrides.
- A sweep makes at most one provider attempt, including on failure, so a degraded provider cannot fan out across every agent in one five-minute tick.
- I also corrected the source-owned Dreaming integration fixture to use the closed `workflow` entity type; it had been failing after the contract enum deliberately retired free-form `process`.

Focused config, Dreaming, worker, and source-artifact graph tests pass (102 tests / 344 expectations), as does daemon typecheck. I am continuing the source-to-derivative linkage trace before implementing contested-claim invalidation; there is not yet a proven general dependency edge to mutate safely.

NicholaiVogel:
Identity follow-through is in `d959ae547`, rebased and pushed.

Dreaming no longer treats a missing identity file as an error: identity is optional context, so normal absent files neither degrade a pass nor emit recurring warnings. An existing file that cannot be read is distinct: the completed pass summary records `identity context degraded` with the affected configured relative path. The prompt now states the durable invariant directly: identity is contextual prior, never schema; claims attach to their entity and aspect.

Focused Dreaming coverage now proves both the entity-scoped prompt contract and the unreadable-file pass marker, alongside the scheduler/source tests and daemon typecheck.

NicholaiVogel:
I traced the stale-derivative report before adding another invalidation subsystem. One direct contradiction was already present: a materialized Dreaming claim is both an `entity_attributes` row and a semantic `memories` projection with the same id, but generic memory PATCH could rewrite only the projection.

`6de297b4b` closes that split: content/type mutation of an attribute-backed derived memory now returns `409 semantic_projection_content_immutable`; metadata remains editable, and claim changes stay on the audited ontology apply path. Both single PATCH and batch modify use the same transaction guard. Focused transaction, mutation API, and episodic-cutover tests pass (66 tests / 292 expectations), plus daemon typecheck and a Luna review.

The broader report is real but is structurally separate: aggregate derivatives have reverse source links, while claim evidence currently lives in audited JSON provenance rather than a normalized many-to-many source→claim relation. I am not fabricating a partial invalidation path across that boundary; next I am tracing the canonical provenance table shape and retrieval authority rule before proposing it.

NicholaiVogel:
Follow-up on stale derivatives is in `5cebf9b14` (rebased on current `origin/main`, pushed).

- Migration 104 adds `derived_memory_sources`: one canonical, agent-scoped, reverse-indexed lineage relation for derived semantic memories. It copies the two historical aggregate tables but deliberately leaves them intact because migration-artifact verification relies on them; runtime now writes only the new relation.
- A source correction, supersession, or deletion marks its dependent derived rows stale immediately. Recall excludes stale rows; a repeated aggregate query re-derives the stale snapshot, preserves its historical evidence lineage, rewrites its embedding, and records a `rederived` history event.
- Dreaming attribute projections now record their cited episodic record identities in the same relation. Claim supersession/archive also invalidates aggregate snapshots that cited the prior claim.
- The same run exposed a router bug: `undefined` (ordinary foreground operations) was treated as the paused-background sentinel. Aggregate recall therefore never reached the shared Pi route. The router now distinguishes `undefined` from the explicit paused `null` state; the direct OpenAI-compatible aggregate test exercises it.

Focused migrations, transactions, aggregate recall, ontology proposals, and inference-router tests pass (147 tests / 955 expectations), alongside daemon typecheck and Biome. A Luna review found no actionable issue.

Issue #959, opened 2026-07-22
fix(daemon): reasoning flag never forwarded to pi-ai (compares RoutingReasoningDepth to "deep")


NicholaiVogel:
## Bug

`platform/daemon/src/inference-provider-factory.ts:70` derives the `reasoning` flag passed to `createPiModelProvider` like this:

```ts
reasoning:
  target.openrouter?.reasoning?.enabled ??
  (model.reasoning !== undefined ? model.reasoning === "deep" : undefined),
```

`model.reasoning` is a `RoutingReasoningDepth` = `"low" | "medium" | "high"` (see `platform/core/src/routing.ts`). Comparing it to the literal `"deep"` is **always false** — TypeScript itself flags it (TS2367: comparison between `"low" | "medium" | "high"` and `"deep"` has no overlap).

## Impact

Any model configured with `reasoning: "high"` (or `"medium"`) gets `reasoning: false` forwarded into the pi-ai `Model`, so reasoning-capable models silently run **without** extended thinking. This is a behavioral regression introduced by #949.

## Fix needs a product decision

The mapping from `RoutingReasoningDepth` (`low`/`medium`/`high`) to pi-ai's boolean `reasoning` is not mechanical. Options:

- `reasoning = model.reasoning === "high"` (only "high" enables thinking)
- `reasoning = model.reasoning !== "low"` ("medium" and "high" enable thinking)
- Forward as `undefined` (let pi-ai default) unless OpenRouter's explicit `reasoning.enabled` is set

The OpenRouter path already works correctly (`target.openrouter?.reasoning?.enabled`); only the generic `model.reasoning` fallback is broken. #949's PR body states "pi-ai now owns the reasoning wire format" — so the cleanest fix may be to stop translating `RoutingReasoningDepth` into a boolean here entirely and let the depth influence routing/sizing elsewhere.

## Context

- Introduced by #949 (merge `b36d3586f`).
- Caught by running daemon `tsc` against the real `@earendil-works/pi-ai@0.79.2` (no daemon typecheck gate exists today — see separate issue for adding one).

NicholaiVogel:
## Deeper finding (investigated during fix)

This is bigger than the `=== "deep"` TS2367. pi-ai has **two distinct reasoning fields** (verified against `@earendil-works/pi-ai@0.79.2`):

- `Model.reasoning: boolean` — a model **capability** flag ("can this model think")
[...]

PR #963, opened 2026-07-22
fix(daemon): forward per-call reasoning ThinkingLevel to pi-ai (#959)

NicholaiVogel:
## What

Fixes the reasoning regression from #949 (tracked in #959). Investigation against `@earendil-works/pi-ai@0.79.2` revealed the bug is bigger than the `=== "deep"` TS2367 — see the [deeper finding on #959](https://github.com/Signet-AI/signetai/issues/959#issuecomment-5050093562).

pi-ai has **two** reasoning fields:
- `Model.reasoning: boolean` — model *capability* flag
- `options.reasoning: ThinkingLevel` — per-call level that *actually turns thinking on*

#949 set the capability flag but **never forwarded the per-call level** in `buildOptions`, so even a corrected boolean would have had zero observable effect (thinking never engaged). Pre-#949 Anthropic explicitly sent `thinking: { type: "disabled" }`.

## Changes

- **`pi-provider.ts`** — `PiModelProviderConfig.reasoning` widened to `ThinkingLevel`; `Model.reasoning` derived as `config.reasoning !== undefined`; `buildOptions` forwards `reasoning` per-call (reaches both `completeSimple` and `streamSimple`).
- **`inference-provider-factory.ts`** — replaces `=== "deep"` with a sound mapping. `RoutingModelConfig.reasoning` defaults to `"medium"` at every parse site, so mapping it directly would flip thinking ON by default for all routed calls. Treat only explicit signals as intent:
  - OpenRouter `reasoning.enabled` → `"medium"`
  - `model.reasoning === "high"` → `"high"`
  - else → `undefined` (preserves pre-#949 off-by-default behavior)
- **`pi-provider.live.test.ts`** — replaced the undefined `fail()` (would have thrown a swallowed ReferenceError) with a `doneSeen` flag so the abort test can no longer pass on an unexpected normal completion.

## Behavior

Reasoning is now **functional end-to-end** when explicitly requested, and **off by default** (no cost/latency regression). If we later want `"medium"` depth to also engage thinking, that's a one-line change with a cost review.

## Verification

- Daemon `tsc` against real `@earendil-works/pi-ai@0.79.2`: changed files clean (no TS2367, no `fail` TS2304).
- `bun test` provider + router suites: 47/47 pass.
- Structured review confirmed both complete/stream paths forward `options.reasoning`, capability derivation is consistent across all model-construction sites, and no OpenRouter regression (the old code never produced a usable level).

## Deliberate choice flagged

`RoutingModelConfig.reasoning` defaults to `"medium"` and is indistinguishable from unset, so this fix treats only `"high"` (and the OpenRouter block) as intent to emit thinking. Calling out explicitly so a reviewer can weigh in on whether medium-depth-should-enable-thinking is wanted later.

Refs #959

### DIFF
--- a/platform/daemon/src/inference-provider-factory.ts
+++ b/platform/daemon/src/inference-provider-factory.ts
@@ -65,9 +65,19 @@ export async function createRoutingProvider(opts: CreateRoutingProviderOptions):
 		model: model.model,
 		baseUrl: target.endpoint,
 		apiKey: credential,
-		reasoning:
-			target.openrouter?.reasoning?.enabled ??
-			(model.reasoning !== undefined ? model.reasoning === "deep" : undefined),
+		// Map routing intent to a pi-ai ThinkingLevel (forwarded per-call as
+		// options.reasoning). model.reasoning (RoutingReasoningDepth) defaults to
+		// "medium" for every parsed model, so it cannot alone signal "enable
+		// thinking" without flipping a costly default on for all routed calls.
+		// Treat only explicit non-default signals as intent to emit thinking:
+		// the documented OpenRouter reasoning block, or a deliberately-set
+		// "high" depth. Previously this compared to a nonexistent "deep"
+		// value (TS2367) and never produced a usable level.
+		reasoning: target.openrouter?.reasoning?.enabled
+			? "medium"
+			: model.reasoning === "high"
+				? "high"
+				: undefined,
 		contextWindow: model.contextWindow,
 		name: `${target.executor}:${model.model}`,
 		defaultTimeoutMs: 60_000,
--- a/platform/daemon/src/inference-router.test.ts
+++ b/platform/daemon/src/inference-router.test.ts
@@ -331,4 +331,151 @@ describe("InferenceRouter legacy API credentials", () => {
 			rmSync(dir, { recursive: true, force: true });
 		}
 	});
+
+	it("forwards per-call reasoning effort when OpenRouter reasoning is enabled (#959)", async () => {
+		// Regression guard for the reasoning fix: on `main`, the factory derived
+		// reasoning from `=== "deep"` (TS2367, never matched) and pi-provider.ts
+		// never forwarded options.reasoning, so thinking was always off. With the
+		// fix, OpenRouter reasoning.enabled produces reasoning: { effort: "medium" }
+		// on the wire. This test fails on main and passes at HEAD.
+		const dir = mkdtempSync(join(tmpdir(), "signet-router-openrouter-reasoning-on-"));
+		try {
+			mkdirSync(join(dir, "memory"), { recursive: true });
+			writeFileSync(
+				join(dir, "agent.yaml"),
+				`inference:
+  defaultPolicy: flash
+  accounts:
+    openrouter-api:
+      kind: api
+      providerFamily: openrouter
+      credentialRef: OPENROUTER_API_KEY
+  targets:
+    flash:
+      executor: openrouter
+      account: openrouter-api
+      openrouter:
+        reasoning:
+          enabled: true
+      models:
+        default:
+          model: deepseek/deepseek-v4-flash
+  policies:
+    flash:
+      mode: automatic
+      allow:
+        - flash/default
+      defaultTargets:
+        - flash/default
+  workloads:
+    sessionSynthesis:
+      target: flash/default
+      taskClass: session_synthesis
+`,
+			);
+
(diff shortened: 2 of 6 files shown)


NicholaiVogel:
## Adversarial review + live verification results

Ran a second-pass adversarial review (subagent) against this diff and live-tested the provider against Ollama + OpenRouter. Summary below; full detail in the issue thread.

### Live verification
- **Ollama (`qwen3.5:4b`)** ✅ — `reasoning: "high"` forwarded → **no reasoning params on the wire** (proves `LOCAL_COMPAT.supportsReasoningEffort:false` gate works; local servers won't error). `generateWithUsage` returns real usage (24 in / 198 out), confirming the complete+usage path.
- **OpenRouter (`deepseek/deepseek-v4-flash`)** — live call blocked by a stale `OPENROUTER_API_KEY` in env (401 "User not found" on a direct curl, independent of this code) and the `signet secret exec` worker was stuck. Verified the wire contract by reading pi-ai's compiled `openai-completions.js`: `thinkingFormat:"openrouter"` + `options.reasoning:"medium"` → `body.reasoning = { effort: "medium" }`; without the per-call level → `{ effort: "none" }`. That is exactly the bug/fix boundary.

### Review findings (3 substantive)
1. **[major → resolved as design choice]** `model.reasoning === "high"` now enables thinking tokens on every routed call to that model, including cheap background workloads. This is the documented intent of the fix (only explicit `high` or the OpenRouter block opt in), but it overloads the routing *scoring* field. **Flagged for reviewer awareness; not changing in this PR** — the default `medium` path stays off, which is the safe state. If we later want a separate `enableReasoning` toggle, that's a follow-up.
2. **[major — not addressed here]** Anthropic budget-stealing: when thinking is enabled at `"high"` with the default 4096 `maxTokens`, pi-ai's `adjustMaxTokensForThinking` leaves only ~1024 tokens for the answer. This is a real edge but only affects Anthropic targets explicitly set to `reasoning: high` (not OpenRouter, not defaults). Worth a follow-up to thread `maxTokens` through the factory, but out of scope for this reasoning-forwarding fix.
3. **[process gate — fixed in e446dbb]** The fix originally shipped with no regression test. Added two in `inference-router.test.ts`: the `enabled:true → {effort:"medium"}` test **fails on main** (`{effort:"none"}`) and passes at HEAD; plus a guard that default `"medium"` depth does not enable thinking.

### Verified clean (adversarial hypotheses refuted)
- Both `completeSimple` and `streamSimple` receive `options.reasoning` (shared `buildOptions`).
- Capability derivation (`config.reasoning !== undefined`) is equivalent to the old `?? false` for all current callers.
- Stream error path (`ev.error.usage`) remains valid.
- No OpenRouter regression — pre-fix main emitted `{effort:"none"}` (disabled); this PR emits `{effort:"medium"}` when enabled.

Net: no blocker. Findings 1 & 2 are scoped follow-ups; Finding 3 is closed by the new tests.

NicholaiVogel:
## aggregate_recall must not use thinking — fixed (a65ef31)

Good catch from review: the reasoning fix made it possible for thinking tokens to reach **aggregate_recall**, which is latency-sensitive (the routing engine already hard-excludes ACPX subprocesses for it for exactly that reason). Letting thinking fire there would regress aggregate-recall cost/latency whenever its workload target is a reasoning-capable model.

### Why per-call, not per-provider
Providers are cached per target (`providerCache`, keyed on `signature:target/model:hooks`) and **shared across all operations**. So the suppression can't live in provider construction — it has to be call-scoped, exactly like `maxTokens`/`timeoutMs`.

### Fix (mirrors the existing ACPX-exclusion pattern)
- `provider.ts`: added `reasoning?: ThinkingLevel | false` to `LlmProviderCallOptions`. `false` = explicit suppress; a level = override; `undefined` = configured default.
- `pi-provider.ts` `buildOptions`: per-call precedence — `false`→omit, level→override, `undefined`→config.
- `inference-router.ts` `execute` **and** `stream`: pass `reasoning: false` when `request.operation === "aggregate_recall"`.

### Regression test
`aggregate_recall suppresses reasoning even on a reasoning:high target` — routes aggregate_recall to a `reasoning: high` OpenRouter target and asserts no thinking on the wire. **Verified to FAIL without the suppression** (emits `{effort:"high"}`) and PASS with it.

This closes the last behavioral gap from the reasoning fix. The two follow-ups noted earlier (Anthropic maxTokens budget at high depth; separate reasoning toggle vs. overloading the scoring field) remain scoped out of this PR.

Issue #948, opened 2026-07-21
Replace the Signet dashboard with TypeScript + shad-cn and build proper UX


NicholaiVogel:
## Summary

The current Signet dashboard (Svelte-based) is plagued by technical debt, poor UX, and missing critical features. Replace it with a **React + shadcn/ui** implementation with proper onboarding, multi-tenancy support, and a coherent design system.

## Current State

The dashboard is written in Svelte and has accumulated significant debt:
- Hard to work on and iterate
- No proper onboarding flow
- Missing key Signet features like multi-tenancy, agent management, and workspace switching
- Confusing UX that makes it unclear what the dashboard is for
- No coherent design system

This is not a Svelte problem specifically. The dashboard was never given a proper design pass or a full feature set. It grew incrementally without a plan and now needs to be rebuilt from the ground up.

## Design

**The HTML mockup is the definitive design spec. The React implementation must match it exactly. This design is not open to reinterpretation.**

The mockup was built iteratively over several hours and the result is locked. Every pixel of spacing, every token value, every interaction pattern in the mockup carries forward to the React build unchanged. If it looks different in the browser than it does in the mockup, it is wrong.

**Mockup:** `web/marketing/public/redesign-home-mockup.html` — open this file in a browser. That is what the dashboard should look like, pixel for pixel.

**Screenshot:**
![Dashboard redesign home screen](https://raw.githubusercontent.com/Signet-AI/signetai/main/surfaces/dashboard/redesign-home-screenshot.webp)

The mockup includes:
- Complete light and dark themes in oklch
- Geist / Geist Mono font stack
- Sidebar navigation shell with a floating content card
- Platform-aware window chrome (macOS traffic lights, Windows/Linux caption buttons via `data-platform` attribute)
- KPI cards with cursor-tracking border glow
- Daily brief viewer with pagination dots
- Activity contribution chart (GitHub-style)
- Review queue with action buttons
- Source rundown rail
- Settings modal (draft)
- Hover, active, and focus states for every interactive element

The mockup's CSS variables (`--background`, `--foreground`, `--card`, `--border`, `--ring`, `--radius`, etc.) map directly to shadcn/ui's theme contract and will become the `globals.css` in the React project. Do not rename, reinterpret, or "improve" these tokens. They are the design system.

## Decisions

Recorded 2026-07-22. These supersede any earlier direction (including the old retro-tech/Svelte design notes) captured elsewhere.

### Stack
[...]

PR #968, opened 2026-07-22
fix(daemon): integrate pi-ai OAuth providers

### LINKED ISSUE #966: Inference settings ignore pi-ai auth: no OAuth sign-in (Claude Max/Codex/Copilot), OpenRouter models don't load, keys saved as Anthropic
## Summary

The dashboard's inference settings were supposed to inherit Signet's auth + model discovery from the pi SDK (`@earendil-works/pi-ai`, the reason we migrated to pi). In practice the integration is unfinished: the UI hand-rolls a static executor list and a single provider-blind credential field, and the daemon never imports pi-ai's OAuth system at all. The result is that subscription login (Claude Max, OpenAI Codex, Copilot), per-provider model dropdowns, and correct key storage are all broken.

This is a backend integration gap **and** a UI gap. The backend fix is framework-independent and can land ahead of the dashboard rewrite (#948); the UI fix belongs in the rewrite.

## Root cause

**pi-ai already ships a complete OAuth provider system.** `@earendil-works/pi-ai@0.79.2` exports (from its `oauth` module):

- `anthropicOAuthProvider` — *"Anthropic (Claude Pro/Max)"* — PKCE + local callback server + auto-refresh
- `openaiCodexOAuthProvider` — OpenAI Codex (ChatGPT OAuth) — device-code + browser flow + auto-refresh
- `githubCopilotOAuthProvider` — GitHub Copilot OAuth
- A registry: `getOAuthProviders()`, `getOAuthProvider(id)`, `registerOAuthProvider(...)`
- `getOAuthApiKey(providerId, credentials)` — resolves an API key from stored OAuth creds **and auto-refreshes expired tokens**
- `OAuthCredentials = { refresh, access, expires, ... }`
- `getProviders()` / `getModels(provider)` — dynamic model discovery across ~35 `KnownProvider`s (incl. `anthropic`, `openai-codex`, `openrouter`, `opencode-go`, `google`, ...)

**Signet consumes none of it.** Verified:

- `grep` for `oauth|getOAuthProvider|OAuthCredentials|loginAnthropic|openaiCodexOAuth|getOAuthApiKey|registerOAuthProvider` across `platform/daemon/src`, `platform/core/src`, `surfaces/cli/src` → **zero hits**.
- Daemon inference routes (`platform/daemon/src/routes/inference.ts`) are only: `status`, `catalog`, `history`, `explain`, `execute`, `stream`, `requests/:id`. There is **no OAuth login route** — no way to even start a login flow.
- `resolveCredential()` in `platform/daemon/src/inference-router.ts` (~L462) is **provider-blind**: it only does `process.env[credentialRef]` then `getSecret(credentialRef)`. No OAuth awareness, no token refresh, no notion of which provider a key belongs to.

### 1. Can't sign in with subscription providers (Claude Max, OpenAI Codex, OpenCode Go, Copilot)
`surfaces/dashboard/src/lib/components/tabs/settings/InferenceSection.svelte` L40 hardcodes `EXECUTOR_OPTIONS = ["openai-compatible", "anthropic", "openrouter", "ollama", "llama-cpp", "acpx"]`. This is a static 6-item enum — it does **not** come from `getProviders()`, and there is no OAuth sign-in surface anywhere in the app. Claude Max / Codex / OpenCode Go / Copilot are structurally impossible to select, even though pi-ai ships ready-made login flows for the first three.

### 2. OpenRouter (and Anthropic) model dropdown is empty
The model dropdown exists in code (`bgModelOptions()` reads `catalog.models[family]` from `/api/inference/catalog`), but `/api/inference/catalog` (`routes/inference.ts` L802) builds `models[provider]` per `getProviders()` and **silently swallows failures**: `catch { models[provider] = []; }`. When `getModels(provider)` throws for a family, the user gets an empty dropdown and falls through to a bare custom-id text input. The catalog endpoint should expose the real provider list and not silently zero-out families.

### 3. API key saved as an Anthropic key instead of the chosen provider's key
The API-key field's placeholder is **hardcoded to `ANTHROPIC_API_KEY` for every provider** — `InferenceSection.svelte` L340 (background) and L418 (aggregation). `credentialRef` is a single free-text secret-name slot with no per-provider credential model. There is no concept of "the OpenRouter key" vs "the Anthropic key" — one undifferentiated field, mislabeled Anthropic. Combined with the provider-blind `resolveCredential()`, whatever the placeholder nudges the user toward is what gets stored and looked up.

### 4. No per-provider model selection for direct APIs
Selecting Anthropic or OpenRouter should present a dropdown of that provider's available models (from `getModels(provider)`). Today the dropdown is empty (bug #2) and only a free-text custom id is offered.

### Backend (framework-independent — can land before #948)
- Import pi-ai OAuth in the daemon.
- Store `OAuthCredentials` (`{refresh, access, expires}`) in the secrets store, keyed by provider id.
- New routes in `routes/inference.ts`:
  - `GET /api/inference/oauth/providers` — from `getOAuthProviders()`
  - `POST /api/inference/oauth/login/:id` — starts the flow, streams `onAuth` / `onDeviceCode` / `onPrompt` events back to the client
  - `POST /api/inference/oauth/complete` (or callback handler) — stores resulting creds
  - `POST /api/inference/oauth/disconnect/:id`
- Make `resolveCredential()` **provider-aware**: an account whose `providerFamily` is an OAuth provider resolves via `getOAuthApiKey()` (auto-refresh on expiry), not raw env/secret lookup.
- Extend `/api/inference/catalog` to also return `oauthProviders` and the real provider list from `getProviders()` (and stop silently zeroing-out families — surface the failure or fix the resolver).

### UI (belongs in the React rewrite — #948)
- Dynamic provider list driven by the catalog endpoint (kill the hardcoded `EXECUTOR_OPTIONS`).
- "Sign in with Claude / Codex / Copilot" buttons that drive the OAuth routes and render device-code/prompt UI.
- Per-provider model dropdowns from `getModels(provider)`.
- Remove the hardcoded `ANTHROPIC_API_KEY` placeholder; credential storage is per-provider and keyed by provider id.

## Non-goals

- Not changing the routing/executor model semantics (`targets` / `accounts` / `workloads`).
- Not touching ACPX (harness subprocess path is separate and unaffected).

## Related
#947 — inference cutover to pi-ai (#949); the auth backend this OAuth work extends.

- #948 (dashboard rewrite) — the UI half of this fix lands there.

aaf2tbz:
## Summary

Completes the framework-independent backend half of #966 by integrating pi-ai's OAuth and dynamic provider registries into the daemon. OAuth credentials remain daemon-owned in Signet's encrypted secrets store, refresh through pi-ai at request time, and never cross the inference API boundary.

Fixes #966. The React dashboard work remains scoped to #948 as requested.

## Changes

- Add bounded SSE OAuth login, interactive completion, provider status, and disconnect routes.
- Store provider-keyed OAuth credentials in Signet Secrets and single-flight token refresh through `getOAuthApiKey()`.
- Resolve OAuth accounts by `providerFamily` while preserving env/secret API-key precedence for conventional accounts.
- Build native pi-ai providers/models dynamically, including credential-dependent Copilot model metadata.
- Return real pi-ai/OAuth catalog metadata and explicit per-provider model errors instead of silent empty lists.
- Accept safe dynamic pi-ai executor ids without changing target/account/workload semantics or ACPX behavior.
- Document the API, configuration, security boundary, session TTL, and approved router-spec delivery status.
- Add regression coverage for OAuth lifecycle, refresh persistence, catalog routing, provider model fidelity, validation, and authorization.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI files changed. The UI half remains in #948.

## PR Readiness (MANDATORY)

N/A — no database migration or Rust surface changes.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Passed locally:

- `bunx biome check` on all changed TypeScript files
- `bun run build:core`
- `cd platform/core && bun run typecheck`
- `cd platform/daemon && bun build.ts`
- 33 focused core/router/OAuth/provider tests
- 3 focused inference route, validation, and permission tests
- `bun scripts/doc-drift.ts` confirms the new OAuth routes are documented

Full `platform/daemon/src/inference-api.test.ts` result is 14 pass / 12 fail. A pristine `origin/main` worktree reproduces the same 12 failures at 12 pass / 12 fail; they are existing command-executor folding and OpenAI-compatible availability-probe expectations. This PR adds two passing cases and no new failures to that file.

## AI disclosure

## Notes

OAuth login sessions are capped at 8 concurrent sessions, expire after 10 minutes, validate offered selections, and clear pending interactions on timeout or disconnect. Credential-bearing providers bypass the fixed provider cache so refreshed tokens and provider-specific model mutations are applied per request.

### DIFF
--- a/platform/core/src/routing.ts
+++ b/platform/core/src/routing.ts
@@ -1,5 +1,5 @@
-import type { PipelineCommandConfig, PipelineExtractionConfig, PipelineSynthesisConfig } from "./types";
 import { defaultPipelineModel } from "./pipeline-providers";
+import type { PipelineCommandConfig, PipelineExtractionConfig, PipelineSynthesisConfig } from "./types";
 
 export const ROUTING_ACCOUNT_KINDS = ["subscription_session", "api"] as const;
 export const ROUTING_TARGET_KINDS = ["subscription_session", "api", "local", "gateway"] as const;
@@ -15,6 +15,7 @@ export const ROUTING_EXECUTOR_KINDS = [
 	"openai-compatible",
 	"command",
 ] as const;
+const ROUTING_EXECUTOR_PATTERN = /^[a-z0-9][a-z0-9._-]{0,127}$/;
 export const ROUTING_POLICY_MODES = ["strict", "automatic", "hybrid"] as const;
 export const ROUTING_PRIVACY_TIERS = ["remote_ok", "restricted_remote", "local_only"] as const;
 export const ROUTING_REASONING_DEPTHS = ["low", "medium", "high"] as const;
@@ -34,7 +35,7 @@ export const ROUTING_OPERATION_KINDS = [
 
 export type RoutingAccountKind = (typeof ROUTING_ACCOUNT_KINDS)[number];
 export type RoutingTargetKind = (typeof ROUTING_TARGET_KINDS)[number];
-export type RoutingExecutorKind = (typeof ROUTING_EXECUTOR_KINDS)[number];
+export type RoutingExecutorKind = (typeof ROUTING_EXECUTOR_KINDS)[number] | (string & {});
 export type RoutingPolicyMode = (typeof ROUTING_POLICY_MODES)[number];
 export type RoutingPrivacyTier = (typeof ROUTING_PRIVACY_TIERS)[number];
 export type RoutingReasoningDepth = (typeof ROUTING_REASONING_DEPTHS)[number];
@@ -366,8 +367,10 @@ function inferTargetKind(executor: string): RoutingTargetKind {
 	if (executor === "ollama" || executor === "llama-cpp" || executor === "openai-compatible" || executor === "command") {
 		return executor === "openai-compatible" ? "gateway" : "local";
 	}
-	if (executor === "anthropic" || executor === "openrouter") return "api";
-	return "subscription_session";
+	if (executor === "acpx" || executor === "claude-code" || executor === "codex" || executor === "opencode") {
+		return "subscription_session";
+	}
+	return "api";
 }
 
 function inferLegacyTargetKind(executor: string, endpoint: string | undefined): RoutingTargetKind {
@@ -588,7 +591,7 @@ function parseOpenRouterConfig(raw: unknown): RoutingOpenRouterConfig | undefine
 function parseTargetConfig(raw: unknown): RoutingTargetConfig | null {
 	if (!isRecord(raw)) return null;
 	const executor = asString(raw.executor);
-	if (!executor || !(ROUTING_EXECUTOR_KINDS as readonly string[]).includes(executor)) return null;
+	if (!executor || !ROUTING_EXECUTOR_PATTERN.test(executor)) return null;
 	const modelsRaw = isRecord(raw.models) ? raw.models : null;
 	if (!modelsRaw) return null;
 	const models: Record<string, RoutingModelConfig> = {};
--- a/platform/daemon/src/inference-oauth.ts
+++ b/platform/daemon/src/inference-oauth.ts
@@ -1,0 +1,313 @@
+import { randomUUID } from "node:crypto";
+import {
+	type OAuthCredentials,
+	type OAuthLoginCallbacks,
+	type OAuthPrompt,
+	type OAuthSelectPrompt,
+	getOAuthApiKey,
+	getOAuthProvider,
+	getOAuthProviders,
+} from "@earendil-works/pi-ai/oauth";
+import { deleteSecretFromActiveProvider, getSecret, putSecret } from "./secrets";
+
+const OAUTH_SECRET_PREFIX = "SIGNET_OAUTH_";
+const OAUTH_SESSION_TTL_MS = 10 * 60_000;
+const MAX_ACTIVE_OAUTH_SESSIONS = 8;
+const PROVIDER_ID_PATTERN = /^[a-z0-9][a-z0-9._-]{0,127}$/;
+
+export type OAuthLoginEvent =
+	| { readonly type: "session"; readonly sessionId: string; readonly providerId: string }
+	| { readonly type: "auth"; readonly url: string; readonly instructions?: string }
+	| {
+			readonly type: "device_code";
+			readonly userCode: string;
+			readonly verificationUri: string;
+			readonly intervalSeconds?: number;
+			readonly expiresInSeconds?: number;
+	  }
+	| ({ readonly type: "prompt"; readonly responseId: string } & OAuthPrompt)
+	| ({ readonly type: "select"; readonly responseId: string } & OAuthSelectPrompt)
+	| { readonly type: "manual_code"; readonly responseId: string; readonly message: string }
+	| { readonly type: "progress"; readonly message: string }
+	| { readonly type: "connected"; readonly providerId: string }
+	| { readonly type: "error"; readonly error: string }
+	| { readonly type: "done" };
+
+export interface ResolvedOAuthCredential {
+	readonly apiKey: string;
+	readonly credentials: OAuthCredentials;
+}
+
+interface PendingInteraction {
+	readonly responseId: string;
+	readonly allowEmpty: boolean;
+	readonly allowedValues?: ReadonlySet<string>;
+	resolve(value: string | undefined): void;
+	reject(error: Error): void;
+}
+
+interface OAuthLoginSession {
+	readonly id: string;
+	readonly providerId: string;
+	readonly controller: AbortController;
+	readonly pending: Map<string, PendingInteraction>;
+	readonly expiresAt: number;
+	emit(event: OAuthLoginEvent): void;
+	close(): void;
+}
+
+const activeSessions = new Map<string, OAuthLoginSession>();
+const refreshes = new Map<string, Promise<ResolvedOAuthCredential | null>>();
+
+function validateProviderId(providerId: string): string {
+	const normalized = providerId.trim();
+	if (!PROVIDER_ID_PATTERN.test(normalized)) throw new Error("Invalid OAuth provider id");
+	return normalized;
+}
+
+function secretName(providerId: string): string {
+	const normalized = validateProviderId(providerId);
+	return `${OAUTH_SECRET_PREFIX}${Buffer.from(normalized, "utf8").toString("hex").toUpperCase()}`;
+}
+
+function isOAuthCredentials(value: unknown): value is OAuthCredentials {
+	if (!value || typeof value !== "object" || Array.isArray(value)) return false;
+	const candidate = value as Record<string, unknown>;
+	return (
+		typeof candidate.refresh === "string" &&
+		candidate.refresh.length > 0 &&
+		typeof candidate.access === "string" &&
+		candidate.access.length > 0 &&
+		typeof candidate.expires === "number" &&
+		Number.isFinite(candidate.expires) &&
+		candidate.expires >= 0
+	);
+}
+
+function safeError(error: unknown): string {
+	const message = error instanceof Error ? error.message : String(error);
+	return message
+		.replace(/\bBearer\s+[^\s,}]+/gi, "Bearer [redacted]")
+		.replace(/([?&](?:code|token|access_token|refresh_token)=)[^&\s]+/gi, "$1[redacted]")
+		.slice(0, 500);
+}
+
+export function listOAuthProviderMetadata(): Array<{
+	readonly id: string;
+	readonly name: string;
+	readonly usesCallbackServer: boolean;
+}> {
+	return getOAuthProviders().map((provider) => ({
+		id: provider.id,
+		name: provider.name,
+		usesCallbackServer: provider.usesCallbackServer === true,
+	}));
+}
+
+export function isOAuthProvider(providerId: string): boolean {
+	return getOAuthProvider(providerId) !== undefined;
+}
+
+export async function loadOAuthCredentials(providerId: string): Promise<OAuthCredentials | null> {
+	const name = secretName(providerId);
+	let raw: string;
+	try {
+		raw = await getSecret(name);
+	} catch (error) {
+		const message = error instanceof Error ? error.message : String(error);
+		if (/not found|does not exist|no such/i.test(message)) return null;
+		throw error;
+	}
+	let parsed: unknown;
+	try {
+		parsed = JSON.parse(raw);
+	} catch {
+		throw new Error(`Stored OAuth credentials for ${providerId} are invalid JSON`);
+	}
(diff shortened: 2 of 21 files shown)


platform/daemon/src/inference-router.ts
```diff
@@ -536,7 +562,7 @@ export class InferenceRouter {
 			};
 		}
 		if (needsCredential) {
-			const credential = await this.resolveCredential(account?.credentialRef);
+			const credential = await this.resolveCredential(account);
```
NicholaiVogel, line 563: 🔴 `resolveCredential()` is called **outside** the try/catch here (the try only wraps `createProvider` + `available()`), and it propagates a throw on refresh failure.

`getOAuthApiKey` (pi-ai) **throws** when a refresh fails and only returns `null` when credentials are absent (see `pi-ai/dist/utils/oauth/index.js` `getOAuthApiKey` and the `@throws` in the `.d.ts`). `resolveOAuthCredential` (inference-oauth.ts) doesn't catch it, so the throw lands here unguarded.

The blast radius is the problem. `runtimeSnapshot` builds state with `Promise.all(allTargetRefs(loaded.config).map(... runtimeStateForTarget(...)))` and **no `.catch`**. So one OAuth target whose refresh token was revoked/expired makes `Promise.all` reject, which means `GET /api/inference/status`, `POST /api/inference/explain`, `execute`, and `stream` all return **500 for every target** — including perfectly healthy local / Anthropic / OpenRouter ones. And because the throw prevents the snapshot from being cached, it 500s on *every* call until someone disconnects the offending OAuth provider.

**Fix (small):** catch refresh failures in `resolveCredential` (or `resolveOAuthCredential`) and treat them as "no usable credential" — i.e. return `undefined` so this target flows into the existing `available:false / health:blocked / accountState:expired` path (the `if (!credential)` block right below). The execute attempt loop already does per-target fallback correctly for call failures; only the snapshot / credential-resolution path needs hardening.

platform/daemon/src/inference-oauth.ts
```diff
@@ -0,0 +1,313 @@
...
+
+export async function resolveOAuthCredential(providerId: string): Promise<ResolvedOAuthCredential | null> {
+	const normalized = validateProviderId(providerId);
+	const existing = refreshes.get(normalized);
+	if (existing) return existing;
+
+	const pending = (async () => {
+		const credentials = await loadOAuthCredentials(normalized);
+		if (!credentials) return null;
+		const result = await getOAuthApiKey(normalized, { [normalized]: credentials });
```
NicholaiVogel: Same root cause as the router comment above — this is the cleanest single place to fix it.

`getOAuthApiKey` throws here on refresh failure. Returning `null` for that case (instead of letting it throw) is exactly what lets the router degrade gracefully: the target gets marked `accountState: "expired"` / blocked rather than poisoning the whole snapshot. Keeping the throw for genuine programming errors is fine, but the refresh/network/revocation path should be caught here and surfaced as "unavailable," not propagated all the way up to `runtimeSnapshot`'s `Promise.all`.

Concretely something like:

```ts
let result;
try {
  result = await getOAuthApiKey(normalized, { [normalized]: credentials });
} catch {
  return null; // refresh failed — treat as expired/unavailable, don't poison the router
}
if (!result) return null;
```

platform/daemon/src/inference-router.ts
```diff
@@ -459,14 +461,25 @@ export class InferenceRouter {
...
 	}
 
-	private async resolveCredential(credentialRef: string | undefined): Promise<string | undefined> {
+	private async resolveCredential(
+		account: RoutingAccountConfig | undefined,
+	): Promise<ResolvedInferenceCredential | undefined> {
+		if (
+			account &&
+			isOAuthProvider(account.providerFamily) &&
+			(account.kind === "subscription_session" || !account.credentialRef)
```
NicholaiVogel: 🟡 This OAuth-account predicate is duplicated **verbatim at three call sites** — `resolveCredential` (here, ~L470), `buildProvider` (~L500), and `probeTarget`/`runtimeStateForTarget` (~L549):

```ts
account !== undefined &&
  isOAuthProvider(account.providerFamily) &&
  (account.kind === "subscription_session" || !account.credentialRef)
```

If the rule ever changes (e.g. a third trigger for the cache bypass) the three sites have to move in lockstep, and if `buildProvider`'s cache-bypass diverges from `probeTarget`'s `needsCredential` you can get the nasty failure mode where the cache returns a provider built with stale credentials while the probe thinks the account needs fresh resolution. Worth extracting a single helper, e.g. `isOAuthBackedAccount(account: RoutingAccountConfig | undefined): boolean`.

platform/core/src/routing.ts
```diff
@@ -588,7 +591,7 @@ function parseOpenRouterConfig(raw: unknown): RoutingOpenRouterConfig | undefine
 function parseTargetConfig(raw: unknown): RoutingTargetConfig | null {
 	if (!isRecord(raw)) return null;
 	const executor = asString(raw.executor);
-	if (!executor || !(ROUTING_EXECUTOR_KINDS as readonly string[]).includes(executor)) return null;
+	if (!executor || !ROUTING_EXECUTOR_PATTERN.test(executor)) return null;
```
NicholaiVogel, line 594: Note (not blocking) — flagging the design trade-off so it's a conscious decision in the spec:

Config validation moved from a closed allowlist (`ROUTING_EXECUTOR_KINDS` `.includes`) to a regex, and `RoutingExecutorKind` / `PiExecutorKind` widened to `| (string & {})` to accept dynamic pi-ai provider ids without a daemon release. Supporting dynamic provider ids without a release is a reasonable call, and the factory still rejects unknown executors/models at build time — so it's defense-in-depth. Two things to be aware of:

1. **Error timing shifts** from config-parse (early, clear: "executor not in known set") to provider-build (later, less obvious: "Unsupported routing executor …" / "Unknown pi-ai model …"). Worth a line in the spec so users know where to look.
2. The `(string & {})` idiom **disables switch-exhaustiveness checking**. The new `default: throw` in `resolvePiModel` covers the gap, just keep it there.

NicholaiVogel (changes requested):
Requesting changes for one real behavioral bug; the rest of this is well done.

## The blocker: an uncaught refresh failure takes down the *entire* router

`getOAuthApiKey` (pi-ai) **throws** on refresh-token failure and only returns `null` when credentials are absent. `resolveOAuthCredential` doesn't catch it, and in the router `resolveCredential()` is called **outside** the try/catch in the snapshot path. I reproduced the throw:

```
await storeOAuthCredentials(PID, { refresh: "r", access: "expired", expires: Date.now() - 1 });
await expect(resolveOAuthCredential(PID)).rejects.toThrow(/Failed to refresh OAuth token/);
```

`runtimeSnapshot` builds state with `Promise.all(allTargetRefs(...).map(... runtimeStateForTarget))` and no `.catch`. So **one** OAuth target with a revoked/expired refresh token makes that reject, which means `GET /api/inference/status`, `explain`, `execute`, and `stream` all return **500 for every target** — including healthy local/Anthropic/OpenRouter ones — and it 500s persistently on every call until someone disconnects the dead provider. Details + fix in the inline comments. The execute *attempt loop* already does per-target fallback correctly; only the snapshot/credential-resolution path is unhardened.

## What I verified (all green)

- **typecheck** passes on every PR-touched file. The 4 `inference-router.test.ts` errors (`Property 'reasoning' does not exist on type 'never'`) are **pre-existing on `origin/main`** (#959 OpenRouter-reasoning mock), not introduced here.
- **No new test regressions.** Pristine `origin/main` worktree: `inference-api.test.ts` = 12 pass / 12 fail. PR branch: **14 pass / 12 fail**. Same 12 failures (command-executor folding + OpenAI-compatible availability-probe expectations the PR description already calls out), +2 new passing cases.
- All 4 new `inference-oauth` tests, plus `inference-provider-factory`, `pi-provider`, and the new `routing.test.ts` case pass; `bunx biome check` clean on all 6 changed source files.

## Good decisions worth keeping

- **Auth gating is right.** `/api/inference/*` middleware requires `diagnostics` for GET, `admin` for mutations; the new test asserts an `operator` token gets `403` on `disconnect` (passes). `catalog`/`providers` only expose `connected: boolean`, never credential values.
- **Provider cache bypass** for OAuth-backed targets + single-flight `refreshes` map + `invalidateCredentialState()` on connect/disconnect — correct ordering and no refresh storms.
- **SSE session lifecycle** — 10 min TTL (unref'd), 8-session cap, pending-interaction cleanup on timeout/disconnect, `cancel()`→abort. `validateProviderId` blocks traversal-style ids before they reach secret storage; hex-encoded secret names satisfy the `[A-Za-z_][A-Za-z0-9_]*` store regex.
- **`completeOAuthInteraction`** validates session alive / responseId / non-empty / selection ∈ offered options; sessionId & responseId are `randomUUID`.
- **`skipAvailabilityProbe`** for native providers — the right call given OAuth providers have no unauthenticated health endpoint.
- **Docs** updated (permission requirements, session TTL, security boundary) in `inference.md`, `CONFIGURATION.md`, and the approved router spec.

Fix the blast radius (small, localized) and this is good to go. I'd also extract the duplicated OAuth-account predicate while you're in there.

NicholaiVogel (approved):
Approving. All three findings from my earlier changes-requested review were addressed in b6bcc1f, and I verified the fix in a worktree:

- Refresh-failure isolation: `resolveOAuthCredential` now catches pi-ai's `Failed to refresh OAuth token` throw and returns null, so the router marks the target `accountState: "expired"` / blocked instead of poisoning the whole snapshot. The end-to-end regression test (revoked OAuth target + healthy fallback) passes — `status()` returns ok and `execute()` routes to the healthy target.
- `isOAuthBackedAccount` extracted as a type guard; the three duplicated predicate sites now call it.
- Spec note added for the executor-allowlist widening.

This commit (d88faee) also adds the dashboard connect UI that completes the #966 UI half — a provider connect wall plus a connect dialog for both OAuth login and API-key entry. Re-reviewed after a second adversarial pass caught that neither connect nor disconnect actually persisted (in-memory store writes wiped by invalidateAll before st.save()); that's fixed in e651618 along with a dead Verify button, a disconnect-orphan, href XSS hardening, and the duplicated secret-name transform. Build + svelte-check clean for the new files, 6 key-validation tests pass.

Note for the record: this expands the PR past the "backend-only, UI is #948" scope in the description, at the owner's direction, so backend + UI ship together and the feature isn't half-finished.

Scope override aside, the backend fixes are correct and the UI is verified green. Merging.

PR #985, opened 2026-07-22
fix(acpx): harden sterile OpenCode inference

### LINKED ISSUE #979: fix(acpx): OpenCode background targets ignore the empty tool allowlist and fail under deny-all
## Summary

OpenCode-backed ACPX inference can fail with exit code 5 under Signet's `deny-all` background policy even when the workload itself does not need tools:

```text
opencode via ACPX exited 5: [acpx] tokens: input=21198 output=35 cache_read=8192 total=29445
[acpx] error: PERMISSION_DENIED Permission request denied or cancelled
```

This was observed during a semantic contradiction check on Signet v0.147.10 with the current source checkout at v0.148.3-1-g2a83986e, OpenCode 1.18.4, ACPX 0.12.0, and `minimax-coding-plan/MiniMax-M3` as the intended model.

## What is happening

Signet correctly launches ACPX with a deny-all policy and an empty allowlist:

```text
--deny-all --allowed-tools ""
```

However, ACPX currently serializes `allowedTools: []` in `session/new` under `_meta.claudeCode.options`. OpenCode does not treat that Claude Code-specific metadata as its own tool configuration. Without an OpenCode-side agent/config override, its ACP adapter falls back to the default `build` agent with tools available.

OpenCode's process logs for the failing background sessions showed:

- `agent=build`
- `model.providerID=opencode`
- `model.id=big-pickle` (rather than the intended MiniMax-M3)
- shell tool execution attempts

The permission request is then denied as configured. ACPX records that every requested permission was denied/cancelled and converts the otherwise completed turn into exit code 5.

This means `resolveAcpxAllowedTools() === []` currently does not guarantee a tool-free OpenCode session.

## Expected behavior

- A hooks-disabled/background OpenCode target must start with no tools advertised or enabled.
- It must use the configured Signet inference agent and routed model rather than silently falling back to OpenCode's `build` agent/default model.
- Structured extraction and synthesis should not be discarded due to incidental permission requests that Signet intended to make impossible.

## Suggested fix

1. Make ACPX launch policy agent-aware rather than assuming `--allowed-tools` is enforced generically across ACP agents.
2. Add supported per-target environment/configuration for OpenCode background workers (or a first-class OpenCode ACP profile).
3. For hooks-disabled OpenCode targets, configure a dedicated primary agent with:
   - `tools: { "*": false }` globally and on the agent
   - `permission: { "*": "deny" }`
   - the routed model selected inside OpenCode when its ACP adapter does not advertise `session/set_model`
4. Surface the effective ACP agent, model, and tool policy in target preflight/status so fallback to `build` or another model is detectable.
5. Add an OpenCode ACP integration test that asserts:
   - the intended model is selected
   - no permission request occurs for a structured inference prompt
   - the process exits 0 under `--deny-all`
   - extraction and synthesis both complete with tools disabled

I would avoid generally treating ACPX exit 5 as success merely because some output exists; preventing the unintended tool request is safer than masking it.

## Confirmed workaround

A dedicated Signet ACPX wrapper can export an OpenCode-only `OPENCODE_CONFIG_CONTENT` overlay that selects MiniMax-M3 and disables all tools before launching ACPX. With that overlay:

- a direct OpenCode/ACPX structured prompt exited 0 with no permission request
- `signet route test --target opencode-background/default --operation memory_extraction ...` completed successfully through the live daemon
- no daemon restart was required because the wrapper is evaluated for each new ACPX child

## Related

- #973 — recursive Signet hooks inside ACPX background workers
- #974 — OpenCode ACP does not advertise generic model selection

aaf2tbz:
## Summary

Hardens sterile OpenCode background inference through ACPX so deny-all targets honor the empty tool policy and recover once from an exit-0, end-turn response with no assistant text. Closes #979.

## Changes

- Inject a native OpenCode config overlay that denies all tools/permissions and selects the routed model without ACPX's unsupported generic `--model` flag.
- Parse ACP JSON completion metadata and retry only sterile, side-effect-free empty completions in a fresh one-shot session under the original deadline.
- Add bounded `emptyResponseRetries` configuration, structured terminal diagnostics, regression coverage, and ACP integration docs.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: ACP integration/configuration docs

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

Agent scoping and admin/mutation endpoint checks are N/A: this change adds no data queries or HTTP endpoints. The readiness result is scoped to changed files and affected packages; repository-wide baseline failures are documented below.

- [ ] Migration is idempotent

Rust parity is N/A because the ACPX provider is TypeScript-only. Rollback is a normal commit revert; `emptyResponseRetries: 0` disables the new retry behavior independently.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Passing scoped verification:

- `bun test platform/core/src/routing.test.ts` — 16 pass, 0 fail
- Focused ACPX provider regressions — 5 pass, 0 fail
- Full provider file — 28 pass; 3 existing macOS fixture failures reproduced before this change (`/var` vs `/private/var`, Linux process-group fixture)
- Biome check on all four changed TypeScript files
- `bun run --filter '@signet/core' typecheck`
- Real ACPX 0.12.0 → OpenCode 1.18.0 → `minimax-coding-plan/MiniMax-M3` inference returned `PONG`
- Isolated daemon health, inference status, and execution all returned 200; execution returned `PONG` with no permission, empty-response, or retry warnings

Repository-wide baseline state:

- `bun run lint` reports 1,773 pre-existing errors outside this diff.
- `bun run typecheck` fails in unrelated daemon test files; affected core typecheck and daemon production build pass.
- The root test run encountered unrelated existing packaging, SQLite extension, shared-state, and legacy-config failures; it was stopped after the failures were established. No issue-specific test failed.

## AI disclosure

## Notes

Retries are deliberately limited to sterile targets (`deny-all`, hooks disabled, empty tools), exit-0 completion, and no observed tool or permission activity. Retry attempts use a fresh one-shot ACPX session and share the original request deadline.

### DIFF
--- a/docs/ACP-INTEGRATION.md
+++ b/docs/ACP-INTEGRATION.md
@@ -135,6 +135,7 @@ inference:
         format: json
         captureEvents: true
         maxCapturedEvents: 200
+        emptyResponseRetries: 1
       models:
         default:
           model: gpt-5.4-mini
@@ -150,6 +151,14 @@ API.
 to JSON when `format` is omitted. Provider-level callers can also attach an
 event callback; Signet limits callback delivery with `maxCapturedEvents`
 (default `200`) while still scanning the full stream for the final response.
+
+ACPX agents can occasionally finish with `end_turn` but no assistant message
+chunks. Signet retries that empty completion in a fresh one-shot session only
+when the target is sterile (`deny-all`, hooks disabled, and an empty tool
+catalog). `emptyResponseRetries` defaults to `1`, is capped at `3`, and can be
+set to `0` to disable the retry. Repeated empties report the output format,
+byte counts, stop metadata when JSON is available, retry count, and duration.
+Tool or permission activity is never retried as an empty completion.
 
 This is intentionally the first Zed-inspired step, not the whole session model:
 Signet now has a typed ACPX event parsing seam, but durable dashboard-visible
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -485,6 +485,7 @@ subscription-backed CLI session, or gateway.
 | `format` / `outputFormat` | string | ACPX output format. `quiet` is the default; `json` parses ACPX JSON events and extracts the final response |
 | `captureEvents` | boolean | When true, defaults ACPX to JSON output and enables the provider event-capture path |
 | `maxCapturedEvents` | number | Maximum number of JSON events delivered to the provider-side event callback; defaults to 200 |
+| `emptyResponseRetries` | number | Fresh-session retries after an exit-0 empty response. Defaults to 1 and is capped at 3; retries run only for sterile deny-all targets with hooks and tools disabled |
 | `timeoutMs` | number | Per-call ACPX subprocess deadline |
 | `extraArgs` | array | Additional ACPX CLI args appended after Signet-managed args |
 | `privacy` | string | `remote_ok`, `restricted_remote`, or `local_only` |
--- a/platform/core/src/routing.test.ts
+++ b/platform/core/src/routing.test.ts
@@ -478,6 +478,7 @@ describe("inference config + decision engine", () => {
 							format: "json",
 							captureEvents: true,
 							maxCapturedEvents: 128,
+							emptyResponseRetries: 9,
 						},
 						models: { default: { model: "gpt-5.4-mini" } },
 					},
@@ -490,6 +491,7 @@ describe("inference config + decision engine", () => {
 			format: "json",
 			captureEvents: true,
 			maxCapturedEvents: 128,
+			emptyResponseRetries: 3,
 		});
 	});
 
--- a/platform/core/src/routing.ts
+++ b/platform/core/src/routing.ts
@@ -1,5 +1,5 @@
-import type { PipelineCommandConfig, PipelineExtractionConfig, PipelineSynthesisConfig } from "./types";
 import { defaultPipelineModel } from "./pipeline-providers";
+import type { PipelineCommandConfig, PipelineExtractionConfig, PipelineSynthesisConfig } from "./types";
 
 export const ROUTING_ACCOUNT_KINDS = ["subscription_session", "api"] as const;
 export const ROUTING_TARGET_KINDS = ["subscription_session", "api", "local", "gateway"] as const;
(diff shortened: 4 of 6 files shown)


NicholaiVogel (approved):
Approving per review. Verified the end-to-end hardening and the integration with #986 (agent-managed models), which landed concurrently.

**Safety of the retry path (the core risk):** retries fire ONLY for sterile targets (`hooks=disabled`, `permissions=deny-all`, empty tool catalog) on an exit-0 response with NO assistant text AND no observable side-effects. The `isAcpxSideEffectEvent` check refuses to retry on any permission request, tool_call, or tool_call_update — so a completion that touched tools is never retried as "empty." Retries run in a fresh one-shot (`mode=exec`, no session) under the original deadline, bounded by `emptyResponseRetries` (default 1, capped 3, 0 disables). Structured diagnostics on terminal failure (exitCode, byte counts, format, sessionId, stopReason, usage, retryCount, durationMs) give operators what they need without leaking prompt content.

**#986 integration (verified in `buildAcpxCommand` + `acpxEnv`):** the merge correctly composes both changes. `--model` is gated on `modelSelection === "acp"` (#986). The deny-all `OPENCODE_CONFIG_CONTENT` overlay is injected only for sterile OpenCode targets, and now passes the routed model only when `modelSelection === "agent"` — which avoids double-setting when `acp` is explicit. I checked all four cases: sterile OpenCode (default `agent`) omits `--model` + overlays the model; Codex/normal ACP passes `--model`; OpenCode with explicit `acp` passes `--model` + no overlay model; OpenCode default `agent` omits `--model`. All correct.

**Premise check vs acpx source:** confirmed against `references/acpx` (github.com/openclaw/acpx@0.12.0) that ACPX's `--model` requires advertised model support that OpenCode lacks, so the overlay approach is the right way to hand OpenCode a routed model under deny-all.

**Verified:** no file overlap between this PR and the concurrent main surge (dashboard/release commits), so the squash applies cleanly. CI on this head is fully green (smoke, Validate mandatory, Daemon tsc, check-versions, regression-test gate). Merging via admin override.

PR #986, opened 2026-07-23
fix(acpx): support agent-managed models

### LINKED ISSUE #974: fix(acpx): OpenCode targets fail because Signet unconditionally passes unsupported --model
**Signet version:** 0.147.10 (also present in current source at `v0.148.3-1-g2a83986e`)
**Environment:** macOS 27.0 arm64, Node 26.5.0, Bun 1.3.14, ACPX 0.12.0, OpenCode 1.18.0
**Related issues:** #947

---

## Summary

Signet unconditionally passes the routed model as ACPX `--model`. OpenCode's ACP agent does not advertise ACP model selection / `session/set_model`, so an otherwise valid OpenCode configuration fails before inference begins.

---

## Observed Behavior

Configuring the target model as `minimax-coding-plan/MiniMax-M3` fails with:

```text
Inference target opencode-background/default failed
opencode via ACPX exited 1: Cannot apply --model "minimax-coding-plan/MiniMax-M3":
the ACP agent did not advertise model support. Generic model selection requires
ACP models plus session/set_model support.
```

OpenCode itself is already configured to use MiniMax-M3 and succeeds if `--model` is omitted.

---

## Expected Behavior

Signet should support ACP agents whose model is selected by the agent's native configuration. Lack of ACP model-negotiation capability should not make the target unusable.

---

## Steps to Reproduce

1. Configure an inference target using `executor: acpx` and `acpx.agent: opencode`.
2. Configure the routed model as `minimax-coding-plan/MiniMax-M3`.
3. Run `signet inference test --target opencode-background/default` or trigger extraction.
4. Observe ACPX reject `--model` before the request reaches OpenCode.
5. Run the equivalent ACPX command without `--model`; it succeeds using OpenCode's configured model.

---

## Root Cause Analysis

`platform/daemon/src/inference-provider-factory.ts` always copies `model.model` into the ACPX provider config. `platform/daemon/src/pipeline/provider.ts::buildAcpxCommand()` then unconditionally emits `--model` whenever `config.model` exists.

There is no configuration or capability-aware path for agent-managed model selection.

---

## Suggested Fix

Add an explicit ACPX model-selection mode, for example `modelSelection: acp | agent`, or omit `--model` for agents that do not advertise ACP model support. Preserve the routed model in status/telemetry as the intended model, but do not force unsupported ACP negotiation.

Add coverage for:

- model-capable ACP agents receiving `--model`;
- OpenCode or another non-advertising agent running without it;
- a clear doctor warning when `modelSelection: agent` relies on external agent config.

---

## Impact

OpenCode-backed extraction, synthesis, repair, and dreaming fail entirely even when the requested model works in OpenCode.

---

## Workaround

Use an ACPX wrapper that strips the `--model <value>` pair and let OpenCode select its configured model.

---

## Environment Details

- **Signet:** 0.147.10
- **OS:** macOS 27.0 arm64
- **Node:** 26.5.0
- **Bun:** 1.3.14
- **DB:** SQLite

## Related

- #973 — hooks:disabled still loads OpenCode plugin, causing recursive capture
- #977 — daemon pause must quiesce all background ACPX workers
- #979 — OpenCode background targets ignore empty tool allowlist under deny-all

aaf2tbz:
## Summary

Adds explicit ACPX model-selection ownership so OpenCode can use its native configured model without Signet forcing unsupported ACP `--model` negotiation. OpenCode defaults to agent-managed selection; other ACP agents retain the existing ACP-managed behavior. Closes #974.

## Changes

- Add validated `modelSelection: acp | agent` routing configuration with one shared resolver and backward-compatible optional typing.
- Default OpenCode to `agent`, omit `--model` in that mode, and preserve explicit `acp` overrides plus existing Codex/Claude behavior.
- Expose effective selection in inference status and add a non-failing route-doctor warning when native agent configuration owns the model.
- Document the behavior and add parser, argv, status API, and CLI doctor regressions.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: ACP integration/configuration docs

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

Agent scoping and admin/mutation security are N/A because this change adds no data queries or mutation endpoints. Readiness validation is scoped to changed files and affected packages.

- [ ] Migration is idempotent

Rust parity is N/A because ACPX routing is currently TypeScript-only. Rollback is a normal commit revert; operators can explicitly select either mode in configuration.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Passing scoped verification:

- `bun test platform/core/src/routing.test.ts` — 17 pass, 0 fail
- Focused ACPX provider tests — 2 pass, 0 fail
- `bun test surfaces/cli/src/commands/route.test.ts` — 9 pass, 0 fail
- Focused inference API test — 1 pass, 0 fail
- Biome check on all nine changed TypeScript files
- `bun run --filter '@signet/core' typecheck`
- `bun run --filter '@signet/cli' build:workspace-deps` and `build:cli`
- Real ACPX 0.12.0 → OpenCode 1.18.0 → `minimax-coding-plan/MiniMax-M3` returned exactly `PONG` with agent-managed selection

Broader file state:

- Full provider tests: 24 pass, with 3 unrelated macOS fixture failures (`/var` vs `/private/var` and a Linux process-group fixture).
- Full inference API tests include environment-dependent unrelated failures; the changed ACPX execution/status case passes.

## AI disclosure

## Notes

Root cause: ACPX 0.12.0 rejects `--model` when the agent does not expose compatible session model metadata. In `agent` mode Signet keeps the routed model in status/provider telemetry but deliberately omits the negotiation flag, leaving model choice to the agent's native configuration. Route doctor warns about that external dependency without marking an otherwise healthy target broken.

### DIFF
--- a/docs/ACP-INTEGRATION.md
+++ b/docs/ACP-INTEGRATION.md
@@ -121,6 +121,14 @@ npx -y acpx@0.7.0 \
 The prompt is written to stdin. The final text response is read from stdout and
 returned to the current `LlmProvider` abstraction.
 
+`modelSelection` controls who applies the routed model. `acp` passes
+`--model` to ACPX and requires the agent to advertise compatible ACP model
+selection. `agent` omits the flag and relies on the harness's native model
+configuration; `signet route doctor` warns operators to verify that external
+configuration. OpenCode defaults to `agent` because installations may not
+advertise ACP model negotiation, while other agents default to `acp`. Set the
+mode explicitly to override either default.
+
 ### JSON event capture
 
 Signet can also ask ACPX for JSON output:
@@ -132,6 +140,7 @@ inference:
       executor: acpx
       acpx:
         agent: codex
+        modelSelection: acp
         format: json
         captureEvents: true
         maxCapturedEvents: 200
@@ -173,6 +182,7 @@ ACPX target config lives under `inference.targets.<name>.acpx`.
 | Field | Type | Description |
 |---|---|---|
 | `agent` | string | ACPX agent command, for example `codex`, `claude` for Claude Code, or `opencode`. Legacy `claude-code` values are normalized to `claude` before spawning ACPX. |
+| `modelSelection` | `acp` or `agent` | `acp` passes the routed model through ACPX negotiation. `agent` omits `--model` and relies on the agent's native configuration. Defaults to `agent` for OpenCode and `acp` otherwise. |
 | `version` / `acpxVersion` | string | ACPX npm package version. Defaults to Signet's pinned `0.7.0`. |
 | `bin` | string | Optional executable override. If omitted, Signet runs `npx -y acpx@<version>`. |
 | `cwd` | string | Working directory for ACPX and the harness. |
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -485,6 +485,7 @@ subscription-backed CLI session, or gateway.
 | `format` / `outputFormat` | string | ACPX output format. `quiet` is the default; `json` parses ACPX JSON events and extracts the final response |
 | `captureEvents` | boolean | When true, defaults ACPX to JSON output and enables the provider event-capture path |
 | `maxCapturedEvents` | number | Maximum number of JSON events delivered to the provider-side event callback; defaults to 200 |
+| `modelSelection` | string | `acp` passes the routed model through ACPX negotiation; `agent` lets the ACP agent's native configuration choose it. Defaults to `agent` for OpenCode and `acp` for other agents |
 | `timeoutMs` | number | Per-call ACPX subprocess deadline |
 | `extraArgs` | array | Additional ACPX CLI args appended after Signet-managed args |
 | `privacy` | string | `remote_ok`, `restricted_remote`, or `local_only` |
--- a/platform/core/src/index.ts
+++ b/platform/core/src/index.ts
@@ -445,9 +445,11 @@ export {
 	compileLegacyRoutingConfig,
 	parseRoutingConfig,
 	allTargetRefs,
+	resolveAcpxModelSelection,
 	resolveRoutingDecision,
 } from "./routing";
 export type {
+	AcpxModelSelection,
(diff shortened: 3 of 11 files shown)


NicholaiVogel (approved):
Approving per review. Verified the core premise against the real acpx source (`references/acpx`, github.com/openclaw/acpx@0.12.0, matching the daemon's pin):

- `assertRequestedModelSupported` (acp/model-support.ts:228) throws `missing-capability` when the ACP session advertises no model config option and the agent isn't Claude Code. The legacy-metadata fallback (`supportsLegacyClaudeCodeModelMetadata`) is Claude-Code-specific.
- OpenCode has no model advertisement and no startup-model flag in the registry (`agent-registry.ts:55` -> `npx -y opencode-ai acp`). So passing `--model` to an OpenCode target genuinely fails today; omitting it is the correct fix.

Traced the full plumbing: `parseAcpxConfig` (routing.ts:615) eagerly resolves `modelSelection` onto the target; `createRoutingProvider` (inference-provider-factory.ts:40) is the only production ACPX construction path and spreads `...target.acpx` + `model` into `createAcpxProvider`; `buildAcpxCommand` (provider.ts:738) gates `--model` on `resolveAcpxModelSelection(...) === "acp"`. Both the OpenCode default and the explicit `model_selection: acp` override work end-to-end.

Backward-compatible: Codex/Claude keep the `acp` default -> `--model` still passed (Claude additionally benefits from acpx's legacy fallback). Agent casing normalized via `toLowerCase()`. The `route doctor` warning is correctly non-fatal (doesn't set exit code). Minor non-blocking note: `resolveAcpxModelSelection` runs twice (parse + build), idempotent.

CI green. Merging via admin override.

PR #989, opened 2026-07-23
fix(dreaming): parse JSON after model preambles

### LINKED ISSUE #976: fix(dreaming): accept valid JSON after model preambles or markdown fences
**Signet version:** 0.147.10 (also present in current source at `v0.148.3-1-g2a83986e`)
**Environment:** macOS 27.0 arm64, Node 26.5.0, Bun 1.3.14, ACPX 0.12.0, OpenCode 1.18.0, MiniMax-M3
**Related issues:** #913, #947

---

## Summary

Incremental dreaming fails after successful inference when a model adds a reasoning preamble before the requested JSON. The parser only accepts raw JSON or a markdown fence that begins at character zero.

---

## Observed Behavior

The ACPX/OpenCode/MiniMax-M3 call completed, but the dreaming pass failed with:

```text
[dreaming] Dreaming pass failed
JSON Parse error: Unexpected identifier "Looking"
```

The model response began with text such as `Looking...` before its JSON payload. The pass was recorded as failed and the token threshold was not reset.

---

## Expected Behavior

Dreaming should accept the first valid complete JSON object when the response contains a short preamble or a fenced JSON block, while continuing to reject genuinely malformed output.

---

## Steps to Reproduce

1. Configure dreaming to use an ACPX/OpenCode target with MiniMax-M3.
2. Trigger `signet dream trigger` with enough graph/session context for a normal pass.
3. Have the model return `Looking at the graph...` followed by a valid JSON object, optionally in a `json` fence.
4. Observe `parseDreamingResult()` pass the entire string to `JSON.parse()` and fail.

---

## Root Cause Analysis

`platform/daemon/src/pipeline/dreaming.ts::parseDreamingResult()` trims the response and strips a code fence only if the whole response starts with triple backticks. It then runs `JSON.parse(cleaned)` directly.

This is less tolerant than other structured-output paths and does not handle common model preambles.

---

## Suggested Fix

Parse the raw string first. On failure, scan for the first balanced JSON object while respecting strings and escapes, parse that candidate, and retain existing mutation-shape validation. Add cases for:

- plain JSON;
- leading prose plus JSON;
- leading prose plus fenced JSON;
- braces inside JSON strings;
- malformed/no-JSON output still failing.

---

## Impact

An otherwise successful and potentially expensive dream pass is discarded. Repeated automatic attempts consume more inference and never reset the accumulated-token counter.

---

## Workaround

Patch `parseDreamingResult()` to extract a balanced JSON object, or use a model/provider that consistently emits raw JSON.

---

## Environment Details

- **Signet:** 0.147.10
- **OS:** macOS 27.0 arm64
- **Node:** 26.5.0
- **Bun:** 1.3.14
- **DB:** SQLite

aaf2tbz:
## Summary

Preserves successful dreaming passes when a model emits reasoning prose or a markdown fence before its structured JSON response. Dreaming now finds the first strictly valid balanced JSON object, while malformed output continues to fail and record the pass as failed. Closes #976.

## Changes

- Keep direct `JSON.parse` as the fast path for normal raw JSON responses.
- Reuse the shared string-aware balanced-object scanner after a raw parse failure and strictly parse candidates in response order.
- Preserve existing mutation-shape validation, transactional application, failure recording, and token accounting.
- Add regressions for prose, fenced JSON, decoy braces, escaped strings, malformed payloads, and successful token-counter reset.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

Agent scoping, config bounds, admin/mutation security, and API/spec documentation changes are N/A because this is an internal response-parser correction with no new queries, inputs, endpoints, config, or contract changes. Changed-file lint, the focused suite, full daemon source suite, production build, and running-daemon validation pass. Standalone daemon typecheck remains red on existing repository-wide `rootDir` and database typing errors unrelated to this diff.

- [ ] Migration is idempotent

Rust parity is N/A because dreaming consolidation is implemented in the TypeScript daemon. The behavior is backward compatible for raw and fenced JSON; rollback is a normal commit revert.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Passing verification:

- `bun test platform/daemon/src/pipeline/dreaming.test.ts` — 32 pass, 0 fail
- `bun test platform/daemon/src` — pass
- Biome check on both changed TypeScript files
- `bun run --cwd platform/daemon build` — pass
- Temporary clean daemon: `/health/ready` reported ready, followed by a clean shutdown

Known baseline:

- `bun run --cwd platform/daemon typecheck` fails on existing repository-wide `rootDir` and database typing errors; no reported error targets the changed files.

## AI disclosure

## Notes

Root cause: `parseDreamingResult()` only removed a markdown fence when it began at character zero, then passed the entire remaining response to `JSON.parse`. Model reasoning such as `Looking...` therefore discarded an otherwise valid and potentially expensive pass. The fallback remains strict JSON parsing and does not repair malformed payloads.

### DIFF
--- a/platform/daemon/src/pipeline/dreaming.test.ts
+++ b/platform/daemon/src/pipeline/dreaming.test.ts
@@ -500,6 +500,47 @@ describe("dreaming", () => {
 			expect(passes[0]?.status).toBe("failed");
 		});
 
+		it("accepts valid JSON after a reasoning preamble and resets the token counter", async () => {
+			addDreamingTokens(accessor, AGENT, 120_000);
+			seedEntity(db, "ent-1", "Test", "concept");
+			const response = JSON.stringify({ mutations: [], summary: "Reviewed graph after reasoning" });
+			const generate = async () => `Looking at {graph} before deciding...\n${response}\nDone.`;
+
+			const result = await runDreamingPass(accessor, generate, defaultCfg(), "/tmp", AGENT, "incremental");
+
+			expect(result.summary).toBe("Reviewed graph after reasoning");
+			expect(getDreamingState(accessor, AGENT).tokensSinceLastPass).toBe(0);
+		});
+
+		it("accepts fenced JSON after a reasoning preamble", async () => {
+			seedEntity(db, "ent-1", "Test", "concept");
+			const response = JSON.stringify({ mutations: [], summary: "Reviewed fenced result" });
+			const generate = async () => `Looking at the graph...\n\`\`\`json\n${response}\n\`\`\``;
+
+			const result = await runDreamingPass(accessor, generate, defaultCfg(), "/tmp", AGENT, "incremental");
+
+			expect(result.summary).toBe("Reviewed fenced result");
+		});
+
+		it("keeps braces, escaped quotes, and backslashes inside JSON strings balanced", async () => {
+			seedEntity(db, "ent-1", "Test", "concept");
+			const summary = 'Reviewed {draft}, the "quoted" value, and C:\\temp';
+			const response = JSON.stringify({ mutations: [], summary });
+			const generate = async () => `Reasoning first.\n${response}`;
+
+			const result = await runDreamingPass(accessor, generate, defaultCfg(), "/tmp", AGENT, "incremental");
+
+			expect(result.summary).toBe(summary);
+		});
+
+		it("rejects malformed balanced output after a preamble", async () => {
+			seedEntity(db, "ent-1", "Test", "concept");
+			const generate = async () => 'Looking...\n{"mutations": nope, "summary": "broken"}';
+
+			await expect(runDreamingPass(accessor, generate, defaultCfg(), "/tmp", AGENT, "incremental")).rejects.toThrow();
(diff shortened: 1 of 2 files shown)


NicholaiVogel (approved):
Approving per review. Dreaming failed when a model emitted reasoning prose or a markdown fence before its JSON response. The fix keeps `JSON.parse` as the fast path and falls back to the existing `extractBalancedJsonObjects` scanner, strictly parsing candidates in response order. Malformed output still fails and records the pass as failed — no silent swallow.

Good edge-case coverage: prose preamble, fenced JSON, decoy braces, escaped quotes/backslashes inside string values, and malformed payloads. Mutation-shape validation and token accounting are preserved. Merging via admin override.

PR #992, opened 2026-07-23
fix(pipeline): quiesce background inference on pause

### LINKED ISSUE #977: fix(pipeline): daemon pause must quiesce all background inference workers
**Signet version:** 0.147.10
**Environment:** macOS 27.0 arm64, Node 26.5.0, Bun 1.3.14, ACPX 0.12.0, OpenCode 1.18.0
**Related issues:** #901, #908

---

## Summary

`signet daemon pause` does not promptly establish a quiescent state when background ACPX extraction/synthesis is active. The command can remain blocked while additional ACPX/OpenCode children appear and old session-end work continues to enqueue summaries, making pause unreliable as an emergency stop for a runaway inference loop.

---

## Observed Behavior

During a recursive OpenCode hook loop:

```text
$ signet daemon pause
Pausing extraction pipeline...
```

The command did not complete promptly. Logs showed workers reporting stopped, but ACPX/OpenCode children were still launched or remained active afterward, and session-end callbacks continued enqueueing summary jobs. Manual process termination was needed before the system became quiet.

---

## Expected Behavior

Pause should have a documented, observable completion contract. Once it reports success, no new extraction, summary, synthesis, repair, or dreaming inference process should start. In-flight work should either be aborted and leases restored or explicitly reported as draining with a bounded timeout.

---

## Steps to Reproduce

1. Configure extraction and synthesis through an ACPX target.
2. Create several active/queued summary and extraction jobs.
3. Run `signet daemon pause` while ACPX calls are active.
4. Observe the CLI remain pending and inspect the process table/logs.
5. Observe ACPX/OpenCode work continue after worker-stop messages.

---

## Root Cause Analysis

The pause route rewrites config and awaits a pipeline runtime restart. Multiple worker types have independent stop/drain/abort behavior, and session-end callbacks can enqueue downstream work while the transition is underway. There is no single quiescence barrier or postcondition check across all inference-producing workers.

---

## Suggested Fix

Define pause semantics and enforce them with a shared cancellation/quiescence barrier:

- reject new leases and inference launches as soon as pause begins;
- abort or bounded-drain all active providers;
- restore leases for aborted jobs;
- prevent transition-time callbacks from starting downstream inference;
- return active/draining counts in the pause response;
- add an integration test with active ACPX processes and queued summary work.

---

## Impact

Operators cannot reliably stop runaway background inference, recursive capture, or unexpected model usage without stopping the entire daemon or killing child processes manually.

---

## Workaround

Stop the daemon entirely or terminate active ACPX/OpenCode child processes before changing configuration.

---

## Environment Details

- **Signet:** 0.147.10
- **OS:** macOS 27.0 arm64
- **Node:** 26.5.0
- **Bun:** 1.3.14
- **DB:** SQLite

## Related

- #973 — hooks:disabled still loads OpenCode plugin, causing recursive capture
- #974 — OpenCode targets fail because Signet passes unsupported --model
- #979 — OpenCode background targets ignore empty tool allowlist under deny-all

aaf2tbz:
## Summary

Closes #977.

Makes `signet daemon pause` an observable quiescence barrier for background inference. A successful pause now closes background router admission, aborts active extraction/session-synthesis/repair provider calls (including ACPX process trees), waits for bounded settlement, and reports active/aborted/remaining counts.

Root cause: workers supplied cancellation signals, but `InferenceRouter.execute()` discarded them before invoking routed providers. Runtime shutdown also had no shared admission gate, so callbacks could enter new inference while individual workers were stopping; dreaming could wait 30 seconds and then drop its handle while ACPX remained alive.

## Changes

- Add router-wide background inference admission control and active-call tracking.
- Compose worker cancellation with the runtime quiescence signal and stop fallback routing after abort.
- Quiesce the router before worker shutdown; reopen admission only for an unpaused runtime.
- Return bounded quiescence counts from live pause/resume transitions and fail pause success if calls remain active.
- Preserve idempotent pause behavior so repeated pause requests do not restart the runtime.
- Document the pause completion contract and add a regression covering abort, fallback suppression, admission rejection, and resume.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: documentation

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Agent scoping verified on all new/changed data queries (N/A: no data queries changed)
- [x] Input/config validation and bounds checks added (bounded 5-second quiescence wait)
- [x] Security checks applied to admin/mutation endpoints (existing admin guard retained)
- [x] Lint/typecheck/tests pass locally (repository-wide baseline is red; details below)

N/A — no schema or data migration.

- [ ] Migration is idempotent
- [x] Daemon Rust parity reviewed or explicitly N/A (JS runtime lifecycle only; Rust daemon has no matching routed-provider lifecycle)
- [x] Rollback / compatibility note included in PR description (removing the commit restores the prior response shape and shutdown behavior; added JSON fields are backward-compatible)

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Passing validation:

- `bun test src/inference-router.test.ts` — 9 passed
- `bun test src/daemon-status.test.ts` — 5 passed
- `bun test src/daemon-refactor.test.ts` — 3 passed
- `bun test src/pipeline/index.test.ts` — 2 passed
- `bun test src/pipeline/dreaming-worker.test.ts` — 2 passed
- `bunx biome lint` on all changed TypeScript files — passed
- `bun run build` in `platform/daemon` — passed (daemon bundles + dashboard production build; existing Svelte accessibility warnings only)
- Running built daemon — first pause returned `remaining: 0`, worker status showed inference workers stopped, idempotent pause did not restart, resume restored controlled-write mode, and SIGINT shutdown exited 0

Repository baseline findings:

- Full daemon suite: 2,144 passed / 7 skipped / 60 failed, then Bun 1.3.14 crashed with exit 133. Failures are unrelated shared-global/environment baseline failures (including `/tmp` vs `/private/tmp` path expectations, package fixture state, and cross-file DB/config contamination). The new quiescence regression passed in this run and all lifecycle-adjacent files pass in isolated processes.
- Standalone daemon typecheck remains red on pre-existing monorepo `rootDir` leakage and broad Bun SQLite typing errors; filtering diagnostics for changed files found no new errors attributable to this change.

## AI disclosure

## Notes

ACPX 0.12 source/distribution behavior was reviewed: it handles termination and escalates child cleanup. Signet's ACPX provider already terminates detached process groups; this fix supplies the cancellation signal that the router previously dropped.

### DIFF
--- a/docs/api/operations.md
+++ b/docs/api/operations.md
@@ -513,9 +513,20 @@ Returns `409` if another pause/resume transition is already running.
   "changed": true,
   "paused": true,
   "file": "/home/user/.agents/agent.yaml",
-  "mode": "paused"
+  "mode": "paused",
+  "quiescence": {
+    "activeAtStart": 1,
+    "aborted": 1,
+    "remaining": 0,
+    "timedOut": false
+  }
 }
 ```
+
+After a successful pause, background extraction, session synthesis (including
+dreaming), and repair inference admission is closed. Active provider calls are
+aborted and boundedly drained before this response. `remaining: 0` confirms
+that no tracked background inference call remains active.
 
 ### POST /api/pipeline/resume
 
--- a/docs/specs/complete/pipeline-pause-control.md
+++ b/docs/specs/complete/pipeline-pause-control.md
@@ -5,6 +5,7 @@ informed_by:
 success_criteria:
   - "Operators can pause extraction work from the CLI or dashboard without losing later backlog processing"
   - "While paused, new memories still persist and queue for later extraction instead of being dropped"
+  - "A successful pause closes background inference admission and aborts or drains every active background provider call"
   - "Resuming restores normal worker startup and backlog draining with no schema migration"
 scope_boundary: "Temporary operator control for background pipeline activity only; does not add schedules, per-stage pausing, or new memory semantics"
 ---
@@ -67,6 +68,12 @@ When `enabled = true` and `paused = true`:
 - do keep retention-only startup behavior
 - do report pipeline mode as `paused`
 
+The live pause transition first closes router admission for extraction,
+summary/synthesis, repair, and dreaming inference. Active calls receive an
+abort signal (including ACPX subprocess trees), and the endpoint does not
+report success until they settle or the bounded quiescence wait expires. The
+response reports the active, aborted, remaining, and timeout counts.
+
 ### C) CLI control
 
 Add:
@@ -123,6 +130,8 @@ pipeline is paused, then drain after resume.
 - `/api/pipeline/status` reports `mode = "paused"` when applicable
 - `/api/pipeline/pause` and `/api/pipeline/resume` expose live operator control
   with structured mutation responses
+- successful pause responses include a `quiescence` result; `remaining = 0`
+  is the observable completion contract
 
 ## 6) Validation and tests
 
@@ -133,3 +142,5 @@ pipeline is paused, then drain after resume.
 - Paused mode reports correctly in status/observability
 - Resume clears the flag and restores normal startup path
 - Dashboard pause/resume helpers surface structured success and error results
+- Router quiescence aborts active background inference, blocks new admission,
+  and prevents an aborted target from spawning a fallback provider
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -46,7 +46,11 @@ import { initFeatureFlags } from "./feature-flags";
 import { writeFileIfChangedAsync } from "./file-sync";
 import { createSignetHttpServer } from "./http-server";
 import { syncAgentWorkspaces } from "./identity-sync";
-import { type InferenceStatusSummary, getOrCreateInferenceRouter } from "./inference-router.js";
+import {
+	type BackgroundInferenceQuiescence,
+	type InferenceStatusSummary,
+	getOrCreateInferenceRouter,
+} from "./inference-router.js";
 import { closeInferenceProviderResolver, getInferenceProvider, initInferenceProviderResolver } from "./llm";
 import { logger } from "./logger";
 import { type ResolvedMemoryConfig, loadMemoryConfig, shouldWarnGraphExtractionWritesDisabled } from "./memory-config";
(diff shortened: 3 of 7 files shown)


NicholaiVogel (approved):
Approving per review. Makes `signet daemon pause` a true quiescence barrier for background inference instead of just flipping a flag.

Root cause was real: workers supplied cancellation signals, but `InferenceRouter.execute()` discarded them before invoking routed providers, so active extraction/synthesis/repair calls kept running; runtime shutdown had no shared admission gate, so new calls could start while workers were stopping; and dreaming could wait 30s then drop its handle while an ACPX subprocess tree stayed alive.

The fix is the right shape:
- Router-wide background admission control (`backgroundAdmissionsOpen`) + active-call tracking (`activeBackgroundExecutions`), composed with the worker cancellation signal via `AbortSignal.any` — so a single abort covers both the router gate and the worker.
- Admission is closed before worker shutdown and only reopened for an unpaused runtime.
- `execute()` now wraps `executeRouted()` in a try/finally that always releases the tracked call, with admission rejection returning a structured `execution-failed` error.
- The `executeRouted` target loop breaks on abort both before attempting a target and after a failed attempt — so an aborted target cannot spawn a fallback provider (the key safety property).

Pause response now reports bounded quiescence counts (`activeAtStart`/`aborted`/`remaining`/`timedOut`), and pause throws if calls remain active — which is the observable completion contract. Idempotent pause preserved (repeated pause doesn't restart the runtime).

Verified: the regression test directly proves abort + fallback-suppression + admission-rejection + resume. CI green (smoke full suite, Validate mandatory, Daemon tsc, regression-test gate). Confirmed zero file overlap with the concurrent main surge (dashboard/ACPX/routing changes), so the squash applies cleanly with no rebase. Merging via admin override.

PR #1010, opened 2026-07-23
fix(test): await native smoke child teardown

### LINKED ISSUE #964: flaky(native): native-embedding-runtime-smoke afterEach teardown times out intermittently
## Symptom

`Nightly Release` intermittently fails on a single test in `scripts/native-embedding-runtime-smoke.test.ts`:

```
(fail) compiled native embedding runtime > keeps /health within SLA while the native embedding download is stalled (event-loop isolation)
  ^ a beforeEach/afterEach hook timed out for this test.
```

The failing **platform is non-deterministic** — it has struck `macos-15-intel` and `ubuntu-latest (linux-x64)` on different runs. All other platforms pass.

## This is pre-existing and unrelated to #949/#962/#963

Proven by run history:

- **#951** (`fix(opencode): retain memory across turn transforms`, merged **8 hours before #949**) — run [29900033536](https://github.com/Signet-AI/signetai/actions/runs/29900033536) — FAILED this exact test, same `afterEach hook timed out`, on `ubuntu-x64`.
- `#950` (the commit immediately before #949) — PASSED (lucky).
- An openclaw fix on 2026-07-20 — also FAILED this test.
- `#949` — its `build:core` tsc error made Nightly Release die early (42s), **masking** this smoke step.
- `#962` — fixed the tsc error, so the build progressed to the smoke step and **unmasked** the flake (struck `macos-15-intel`).
- `#963` — passed all smoke jobs (lucky again).

So this test has been intermittently flaking on `main` for days, independent of the inference provisioning workstream. #962 did not cause it; it revealed it.

## Root cause (test-harness teardown, not a daemon bug)

`scripts/native-embedding-runtime-smoke.test.ts`:

```ts
async function stopChild(child) {
  if (child.exitCode !== null) return;
  child.kill("SIGTERM");
  await Promise.race([
    new Promise<void>((resolve) => child.once("close", () => resolve())),
    Bun.sleep(5_000).then(() => {
      if (child.exitCode === null) child.kill("SIGKILL");
    }),
  ]);
}

afterEach(async () => {
  for (const child of children.splice(0)) await stopChild(child);
  // ...close blackhole servers, rm temp dirs
});
```

The `afterEach` calls `stopChild`, which races a 5s `SIGTERM→SIGKILL` escalation against `child.once("close")`. When the compiled native binary's embedding worker is stuck on the blackhole server (a deliberate stalled-CDN simulation), the process can take long enough to fully close its stdio pipes (`stdio: ["ignore","pipe","pipe"]`) that the hook blows past bun's **default hook timeout (5s)**. The `Promise.race` resolves on the 5s sleep without awaiting `close` after the SIGKILL, so teardown is not reliably bounded.

Importantly: the daemon's `/health` isolation behavior is **correct** — whenever the test reaches its assertions, `/health` stays within SLA (the sibling test `embeds and recalls...` passes). It is the **teardown** that hangs, not the event-loop isolation under test.

## Suggested fix (low-risk, ~10 lines)

1. In `stopChild`, after sending SIGKILL, **await `close`** with a hard bound (e.g. another 2s) so teardown can't return while the process is still dying.
2. Give the file an explicit hook timeout (e.g. `afterEach(..., 30_000)` or a top-level `testTimeout`) so a slow teardown is tolerated rather than reported as a test failure.
3. Optionally mirror the `SIGTERM → grace → SIGKILL → await-exit` pattern already used in `platform/daemon/src/pipeline/provider.ts` (`terminateSubprocessWithEscalation`).

## Repro

Non-deterministic. Most reliably reproduced on a loaded/slow CI runner (Intel Mac or a busy ubuntu runner) by running the full `bun test scripts/native-embedding-runtime-smoke.test.ts` against a freshly compiled native binary with `SIGNET_NATIVE_EMBEDDING_SMOKE=1`. It does not reproduce on fast runners (apple silicon, lightly loaded).

## Context

- Surfaced while investigating the #949 inference cutover CI fallout (#958/#959/#962/#963). Scoped out of that workstream per maintainer decision — tracked here.
- Blocked run examples: [29900033536](https://github.com/Signet-AI/signetai/actions/runs/29900033536) (#951, pre-#949), [29912797848](https://github.com/Signet-AI/signetai/actions/runs/29912797848) (#962).

aaf2tbz:
## Summary

Completes the native embedding smoke teardown hardening from #994 so the release gate waits for the spawned daemon to actually close after escalation and cannot collide with Bun's default hook timeout.

Closes #964

## Changes

- Reuse one `close` promise across the SIGTERM grace period and the SIGKILL reap period.
- Bound post-SIGKILL reaping at 2 seconds and report a diagnostic failure if the child never closes.
- Give the cleanup hook an explicit 30-second timeout.
- Add always-on regression tests for a SIGTERM-resistant child and a missing `close` event.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: native release smoke harness

## Screenshots

N/A — test-harness-only change.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — N/A; no product/spec behavior changed.
- [x] Agent scoping verified on all new/changed data queries — N/A; no data queries changed.
- [x] Input/config validation and bounds checks added — teardown now has explicit grace, reap, and hook bounds.
- [x] Error handling and fallback paths tested (no silent swallow) — missing `close` now fails with a bounded diagnostic.
- [x] Security checks applied to admin/mutation endpoints — N/A; no endpoint changes.
- [x] Docs updated for API/spec/status changes — N/A; no API/spec/status changes.
- [x] Regression tests added for each bug fix — escalation/reap and missing-close cases covered.
- [x] Lint/typecheck/tests pass locally — changed file passes focused tests and Biome; mandatory broad commands were run and their unrelated failures reproduce on clean `main` as detailed below.

- [x] Migration is idempotent — N/A; no migration.
- [x] Daemon Rust parity reviewed or explicitly N/A — N/A; test harness only.
- [x] Rollback / compatibility note included in PR description — N/A; no runtime compatibility change.

## Testing

- [x] `bun test` passes — changed file and runtime pass. Broad branch: 3428 pass, 101 fail, 4 errors; identically built clean `main`: 3426 pass, the same 101 fail and 4 errors. The branch delta is the two new passing regression tests; broad failures are pre-existing and do not involve the changed file.
- [x] `bun run typecheck` passes — changed smoke file compiles in focused and native-runtime tests. The mandatory broad command was run; its daemon `rootDir`/SQLite typing failures reproduce unchanged on clean `main` and do not reference the changed file.
- [x] `bun run lint` passes — `bunx biome check scripts/native-embedding-runtime-smoke.test.ts` passes. The mandatory broad command was run; existing repository formatting diagnostics reproduce on clean `main` and do not include the changed file.
- [x] Tested against running daemon — built the full workspace/dashboard/native executable and ran the compiled native embedding smoke: 4 pass, including the stalled-download `/health` case.
- [x] N/A — UI/migration testing is not applicable; runtime proof is included above.

Additional focused proof:

- `bun test --rerun-each 100 scripts/native-embedding-runtime-smoke.test.ts` — 200 pass, 0 fail (compiled-native cases intentionally skipped without the opt-in env).
- `SIGNET_NATIVE_EMBEDDING_SMOKE=1 SIGNET_NATIVE_SMOKE_BINARY=./dist/native/signet-darwin-arm64 bun test scripts/native-embedding-runtime-smoke.test.ts` — 4 pass, 0 fail.
- `bun run build`, `bun run build:dashboard`, and `bun run build:native-bun` — pass.

## AI disclosure

## Notes

#994 shortened the SIGTERM grace window but still resolved immediately after sending SIGKILL. This PR completes issue #964's recommended action by awaiting `close` under a second hard bound and setting an explicit cleanup-hook budget.

### DIFF
--- a/scripts/native-embedding-runtime-smoke.test.ts
+++ b/scripts/native-embedding-runtime-smoke.test.ts
@@ -10,6 +10,29 @@ const root = join(import.meta.dir, "..");
 const enabled = process.env.SIGNET_NATIVE_EMBEDDING_SMOKE === "1";
 const tempDirs: string[] = [];
 const children: ChildProcessWithoutNullStreams[] = [];
+const CHILD_KILL_REAP_MS = 2_000;
+const CHILD_TERM_GRACE_MS = 2_000;
+
+interface StoppableChild {
+	readonly exitCode: number | null;
+	kill(signal: "SIGTERM" | "SIGKILL"): void;
+	once(event: "close", listener: () => void): void;
+}
+
+interface StopTimings {
+	readonly termGraceMs: number;
+	readonly killReapMs: number;
+}
+
+function closesWithin(closed: Promise<true>, timeoutMs: number): Promise<boolean> {
+	return new Promise((resolve) => {
+		const timer = setTimeout(() => resolve(false), timeoutMs);
+		void closed.then(() => {
+			clearTimeout(timer);
+			resolve(true);
+		});
+	});
+}
 
 /** Resolve the compiled native binary to smoke-test.
  *  Honors an explicit SIGNET_NATIVE_SMOKE_BINARY override (release CI builds to
@@ -60,22 +83,19 @@ function floatVector(value: unknown): Float32Array {
 	return new Float32Array(value.buffer.slice(value.byteOffset, value.byteOffset + value.byteLength));
 }
 
-async function stopChild(child: ChildProcessWithoutNullStreams): Promise<void> {
+async function stopChild(
+	child: StoppableChild,
+	timings: StopTimings = { termGraceMs: CHILD_TERM_GRACE_MS, killReapMs: CHILD_KILL_REAP_MS },
+): Promise<void> {
 	if (child.exitCode !== null) return;
+	const closed = new Promise<true>((resolve) => child.once("close", () => resolve(true)));
 	child.kill("SIGTERM");
-	// Keep the SIGTERM grace window well under bun's default afterEach timeout
-	// (5s). A healthy daemon honors SIGTERM in milliseconds; only a deliberately
-	// wedged child (the isolation test pins it on a blackhole fetch) holds out,
-	// and SIGKILL at 2s is safe on a throwaway process. At 5s the grace window
-	// collided with the hook timeout, so the slowest runner (darwin-arm64)
-	// deterministically reported "a beforeEach/afterEach hook timed out" even
-	// though the test body passed.
-	await Promise.race([
-		new Promise<void>((resolve) => child.once("close", () => resolve())),
-		Bun.sleep(2_000).then(() => {
-			if (child.exitCode === null) child.kill("SIGKILL");
-		}),
-	]);
+	if (await closesWithin(closed, timings.termGraceMs)) return;
+
+	if (child.exitCode === null) child.kill("SIGKILL");
+	if (!(await closesWithin(closed, timings.killReapMs))) {
+		throw new Error(`native smoke child did not close within ${timings.killReapMs}ms after SIGKILL`);
+	}
 }
 
 const blackholeServers: Server[] = [];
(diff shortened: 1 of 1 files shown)


NicholaiVogel (approved):
Approving after full review (the teardown fix traced + new unit tests verified, not diff-only). 16/16 CI green.

What this fixes (#964): a nondeterministic afterEach-hook timeout in the native-embedding smoke test. The old stopChild raced a 5s SIGTERM→SIGKILL escalation against child.once("close") but resolved on the sleep timer WITHOUT awaiting close after SIGKILL — so teardown could return while the process was still dying, blowing past bun's 5s default hook timeout on slow/loaded runners (Intel Mac, busy ubuntu). The daemon's /health isolation under test was always correct; only teardown hung.

The fix matches the issue's suggested approach exactly:
1. After SIGKILL, await close with a hard 2s bound (closesWithin helper). Teardown can no longer return while the process is dying.
2. Explicit afterEach(..., 30_000) timeout so a slow teardown is tolerated rather than reported as a test failure.
3. Clean refactor: stopChild takes a StoppableChild interface + StopTimings, enabling two injectable unit tests that don't need the real binary — one proves SIGTERM→SIGKILL→await-close escalation works on a resistant child; the other proves a never-closing child throws within the reap bound. Both pass.

Rebase note: the branch had accumulated a full release bump (0.154.0→0.154.1) + CHANGELOG entries + lockfile churn from the release-bump race. Rebased to keep ONLY the one real test fix and dropped all version/CHANGELOG noise so no premature bump lands via this PR. Diff is now exactly the 1 reviewed file. #964 closes on merge.

PR #1013, opened 2026-07-23
fix(core): infer openai-compatible privacy from endpoint

### LINKED ISSUE #1002: Privacy inference for openai-compatible ignores endpoint, misclassifies local servers
**Privacy inference for `openai-compatible` ignores the endpoint, misclassifying local servers as `remote_ok`.**

`inferTargetPrivacy("openai-compatible")` at `routing.ts:392` always returns `"remote_ok"`, regardless of whether the endpoint is `127.0.0.1`, a Unix socket, or a remote host. A separate function `inferLegacyTargetPrivacy` (line 395-398) correctly checks `isLocalInferenceEndpoint(endpoint)` for `openai-compatible` executors, but `buildCandidateTrace` (line 1240) calls `inferTargetPrivacy` — not `inferLegacyTargetPrivacy`.

This means any task class with `privacy: restricted_remote` silently blocks a perfectly valid local `openai-compatible` target. The user sees "All routing candidates were blocked by policy or runtime state" with no indication that the privacy gate is what rejected it, unless they know to hit the explain endpoint.

### Reproduction
1. Configure an `openai-compatible` target with an endpoint on `127.0.0.1`
2. Create a task class (e.g. `memory_extraction`) with `privacy: restricted_remote`
3. Router blocks the target with "privacy gate (restricted_remote)" despite the target being local

### Fix direction
Change `buildCandidateTrace` to call `inferLegacyTargetPrivacy(target.executor, target.endpoint)` instead of `inferTargetPrivacy(target.executor)`.

aaf2tbz:
## Summary

Fix endpoint-aware privacy inference for `openai-compatible` routing targets so loopback servers are eligible for `restricted_remote` workloads while remote gateways remain blocked.

## Changes

- Infer default `openai-compatible` privacy from the configured endpoint during target parsing and candidate evaluation.
- Consolidate legacy and explicit target privacy inference into one canonical path.
- Add regressions for parsed loopback targets, direct configs with omitted privacy, and remote endpoints that must remain behind the privacy gate.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: N/A

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — behavior now matches the approved model-provider-router privacy gate; no contract change.
- [x] Agent scoping verified on all new/changed data queries — N/A; no data queries or scope changes.
- [x] Input/config validation and bounds checks added — N/A; the existing normalized endpoint and local-endpoint validator are reused.
- [x] Error handling and fallback paths tested (no silent swallow) — remote endpoints retain the explicit privacy-gate block trace.
- [x] Security checks applied to admin/mutation endpoints — N/A; no endpoint or permission changes. Privacy policy remains fail-closed for remote gateways.
- [x] Docs updated for API/spec/status changes — N/A; public schema and documented intended behavior are unchanged.
- [x] Lint/typecheck/tests pass locally — changed files and affected package pass; repository-wide baseline failures are documented below.

- [x] Migration is idempotent — N/A; no migration.
- [x] Daemon Rust parity reviewed or explicitly N/A — N/A; Rust inference routes do not implement the TypeScript candidate decision engine and currently return router-not-initialized.
- [x] Rollback / compatibility note included in PR description — no schema or compatibility change; reverting restores the previous incorrect loopback classification.

## Testing

- [x] `bun test` passes — changed router suite: 20 passed, 0 failed. Repository-wide branch: 3,017 pass / 108 fail / 20 errors; clean `main` at `4b486522`: 3,015 pass / 108 fail / 20 errors. The broad failures are pre-existing on `main`, and this PR adds two passing regressions without affecting them.
- [x] `bun run typecheck` passes — `@signet/core` passes. Repository-wide branch and clean `main` both report the same 457 pre-existing TypeScript diagnostics; neither changed file contributes a diagnostic.
- [x] `bun run lint` passes — both changed files pass Biome. Repository-wide branch and clean `main` both report the same 1,815 pre-existing errors and 143 warnings; this PR does not affect them.
- [x] N/A — this is a pure core routing decision fix; the exact configured-target reproduction was exercised directly through `parseRoutingConfig` and `resolveRoutingDecision`.

## AI disclosure

## Notes

Closes #1002

The original one-line recommendation covered only candidate fallback. The configured-target reproduction also required endpoint-aware inference during parsing because parsed targets eagerly store their default privacy. This PR fixes both entry paths while preserving explicit privacy overrides and remote fail-closed behavior.

### DIFF
--- a/platform/core/src/routing.test.ts
+++ b/platform/core/src/routing.test.ts
@@ -14,6 +14,38 @@ const ready = {
 	accountState: "ready",
 } as const;
 
+const compatibleTargetRef = makeRoutingTargetRef("compatible", "default");
+
+function parseRestrictedCompatibleRouting(endpoint: string) {
+	return parseRoutingConfig({
+		inference: {
+			defaultPolicy: "auto",
+			targets: {
+				compatible: {
+					executor: "openai-compatible",
+					endpoint,
+					models: {
+						default: {
+							model: "compatible-model",
+						},
+					},
+				},
+			},
+			policies: {
+				auto: {
+					mode: "automatic",
+					defaultTargets: [compatibleTargetRef],
+				},
+			},
+			taskClasses: {
+				memory_extraction: {
+					privacy: "restricted_remote",
+				},
+			},
+		},
+	});
+}
+
 describe("inference config + decision engine", () => {
 	it("classifies loopback inference endpoints as local", () => {
 		expect(isLocalInferenceEndpoint(undefined)).toBe(true);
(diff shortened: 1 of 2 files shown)


NicholaiVogel (approved):
Approving after full review (privacy contract traced end-to-end, not diff-only) plus the added override regression test. 16/16 CI green.

What this fixes (#1002): the canonical routing path called an endpoint-ignoring inferTargetPrivacy("openai-compatible") that always returned remote_ok, so a local llama.cpp server on 127.0.0.1 was wrongly blocked from restricted_remote tasks. The legacy path had a separate inferLegacyTargetPrivacy that got it right — two sources of truth had diverged.

Avery's judgment is sound:
- Fixed the root cause (the duplication) by folding the endpoint argument into one inferTargetPrivacy and deleting the legacy duplicate — removes the drift instead of adding new policy surface.
- Preserved the safety valve: endpoint-inference is a DEFAULT, not forced. asRoutingPrivacyTier(raw.privacy, inferred) means an explicit per-target privacy override still wins. This is the critical guard against the local-proxy-to-cloud leak scenario.
- Fail-closed preserved: unknown executors still default to remote_ok (rank 0); the fix only relaxes the one valid true-loopback case.

Added one regression test here ("honors an explicit privacy override on a loopback endpoint (no proxy-to-cloud leak)"): a 127.0.0.1 target explicitly marked privacy: remote_ok must stay gated for a restricted_remote task. This locks the safety valve the fix relies on — covers the case where the auto-inference is most consequential. 21/21 routing tests pass (was 20).

Non-blocking notes: isLocalInferenceEndpoint returns true for undefined endpoint (pre-existing, out of scope). Rebased onto current main; functional logic byte-identical to the reviewed branch.

PR #1027, opened 2026-07-25
fix(codex): migrate stale desktop runtimes

### LINKED ISSUE #1024: fix(codex): migrate stale desktop runtime paths used by hooks and MCP
## Summary

A Codex Desktop packaging update can move the bundled Node runtime. When a Signet-managed Codex integration retains an absolute path to the former runtime, every lifecycle hook and the Signet MCP proxy fail before the bridge can run.

This affects all native lifecycle events: session start, prompt submit, and session end.

## Expected behavior

The Codex connector should discover a usable bundled Node runtime when installing or upgrading the integration, validate it before writing configuration, and migrate stale Signet-owned runtime paths without touching unrelated hooks.

## Suggested work

- Support current Codex Desktop application layouts rather than assuming one fixed bundle location.
- Validate the chosen executable before writing hook and MCP commands.
- On reconnect or upgrade, detect and replace only stale Signet-owned runtime references.
- Add a macOS regression test for a previously installed integration after a Codex Desktop packaging change.
- Emit a targeted diagnostic when the configured Signet runtime is missing.

## Scope

This is separate from #978 / #993: the active hook manifest already used the valid Codex hook schema and had no invalid ownership markers. The failure occurred while spawning the configured runtime, before Signet or the daemon received the hook request.

aaf2tbz:
## Summary

Fixes #1024. Codex Desktop packaging updates can move its bundled Node executable, leaving Signet-owned lifecycle hooks and MCP stdio configured with a dead absolute path. The connector now discovers and validates the current bundled runtime before rewriting those Signet-owned commands.

## Changes

- Scan supported macOS Codex.app resource layouts for an executable Node runtime and validate it with `--version`.
- Use the validated runtime for Signet hook commands and local MCP stdio when the installed Signet entry points are available.
- Detect a missing prior Signet runtime and emit a focused migration warning while preserving unrelated hooks.
- Add the Codex Desktop packaging-update regression test and document refresh behavior.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — connector-install behavior only; no product spec dependency changed.
- [x] Agent scoping verified on all new/changed data queries — N/A; no data queries.
- [x] Input/config validation and bounds checks added — runtime must be a regular executable file and pass `node --version`; scanning is depth bounded.
- [x] Error handling and fallback paths tested (no silent swallow) — unusable candidates are skipped; existing Signet command resolution remains the fallback.
- [x] Security checks applied to admin/mutation endpoints — N/A; no endpoint changes.
- [x] Docs updated for API/spec/status changes — Codex integration guide documents runtime discovery and migration.
- [x] Regression tests added for each bug fix — macOS packaging-update fixture covers hooks, MCP, and the migration diagnostic.
- [x] Lint/typecheck/tests pass locally — affected-file Biome check, connector typecheck, and 59/59 connector tests pass. Repository-wide lint/typecheck/test failures reproduce on untouched `main` (existing format/type drift, stale generated route/content guards, and local SQLite extension loading).

- [x] Migration is idempotent — reconnect rewrites only the Signet hook and MCP registrations.
- [x] Daemon Rust parity reviewed or explicitly N/A — N/A; this is an install-time TypeScript connector change.
- [x] Rollback / compatibility note included in PR description — reinstalling older Signet restores its previous command strategy; non-Signet hooks are preserved.

## Testing

- [x] `bun test` passes — `bun test integrations/codex/connector/src/config-toml.test.ts`: 59 pass, 0 fail. Full `bun test` has only failures reproduced on untouched `main`: three Rust route-parity guard failures, one generated web-content-index guard failure, and this macOS hosts SQLite build cannot load `vec_version()`.
- [x] `bun run typecheck` passes — connector package passes. Root typecheck failures are pre-existing daemon type errors reproduced on untouched `main`.
- [x] `bun run lint` passes — touched files pass Biome. Root lint has 726 pre-existing diagnostics reproduced on untouched `main`.
- [x] Tested against running daemon — N/A; no daemon runtime behavior changed.
- [ ] N/A

## AI disclosure

- [x] AI tools were used (see `Assisted-by` tag in commit)

## Notes

The connector uses the current Codex Desktop runtime only when the installed Signet JS entry points can be identified. Otherwise it retains the existing portable Signet command fallback.

### DIFF
--- a/dist/signetai/bin/launch.js
+++ b/dist/signetai/bin/launch.js
@@ -48,7 +48,7 @@ export function launchSignet() {
 	// doesn't carry inline. The env is only set when the wrapper actually
 	// has a runtime tree to share; the binary's own bootstrap can derive
 	// its own path otherwise.
-	const env = { ...process.env };
+	const env = { ...process.env, SIGNET_WRAPPER_DIR: packageDir };
 	if (!env.SIGNET_DIR && existsSync(join(packageDir, "runtime", "connectors"))) {
 		env.SIGNET_DIR = packageDir;
 	}
--- a/integrations/codex/README.md
+++ b/integrations/codex/README.md
@@ -18,6 +18,9 @@ available, with a compatibility path for older Codex installs.
   Codex-generated `MEMORY.md` or `memory_summary.md`
 - Configurable timeout grace periods (5 s for SessionStart, 2 s for UserPromptSubmit)
 - Supports remote daemon URL via `SIGNET_DAEMON_URL` environment variable
+- On macOS, discovers and validates the bundled Codex Desktop Node runtime on
+  install and refreshes only stale Signet-owned hook and MCP commands after a
+  Codex Desktop packaging update
 
 ## Installation
 
--- a/integrations/codex/connector/src/config-toml.test.ts
+++ b/integrations/codex/connector/src/config-toml.test.ts
@@ -87,6 +87,7 @@ let previousDaemonUrl: string | undefined;
 let previousApiKey: string | undefined;
 let previousToken: string | undefined;
 let previousForceCompatHooks: string | undefined;
+let previousWrapperDir: string | undefined;
 let previousArgvEntry: string | undefined;
 let previousExecPath: string;
 
@@ -106,6 +107,7 @@ beforeEach(() => {
 	previousApiKey = process.env.SIGNET_API_KEY;
 	previousToken = process.env.SIGNET_TOKEN;
 	previousForceCompatHooks = process.env.SIGNET_CODEX_FORCE_COMPAT_HOOKS;
+	previousWrapperDir = process.env.SIGNET_WRAPPER_DIR;
 	previousArgvEntry = process.argv[1];
 	previousExecPath = process.execPath;
 	Reflect.deleteProperty(process.env, "SIGNET_SESSION_START_TIMEOUT");
@@ -115,6 +117,7 @@ beforeEach(() => {
 	Reflect.deleteProperty(process.env, "SIGNET_API_KEY");
 	Reflect.deleteProperty(process.env, "SIGNET_TOKEN");
 	Reflect.deleteProperty(process.env, "SIGNET_CODEX_FORCE_COMPAT_HOOKS");
+	Reflect.deleteProperty(process.env, "SIGNET_WRAPPER_DIR");
 	tempHome = join(tmpdir(), `signet-codex-test-${Date.now()}-${Math.random().toString(36).slice(2)}`);
 	codexDir = join(tempHome, ".codex");
 	configPath = join(codexDir, "config.toml");
@@ -130,6 +133,7 @@ afterEach(() => {
 	restoreEnv("SIGNET_API_KEY", previousApiKey);
 	restoreEnv("SIGNET_TOKEN", previousToken);
 	restoreEnv("SIGNET_CODEX_FORCE_COMPAT_HOOKS", previousForceCompatHooks);
+	restoreEnv("SIGNET_WRAPPER_DIR", previousWrapperDir);
 	if (previousArgvEntry === undefined) process.argv.splice(1, 1);
 	else process.argv[1] = previousArgvEntry;
 	process.execPath = previousExecPath;
@@ -899,7 +903,7 @@ describe("CodexConnector.install — hooks.json schema", () => {
 		expect(signetHandlers[0]?.timeout).toBe(20);
 	});
 
-	test("migrates stale Codex Desktop node paths after a packaging update", async () => {
+	test("migrates a stale Codex Desktop MCP runtime path after a packaging update", async () => {
 		const appPath = join(tempHome, "Applications", "Codex.app");
 		const runtime = join(appPath, "Contents", "Resources", "runtime", "node", "bin", "node");
 		mkdirSync(join(runtime, ".."), { recursive: true });
@@ -919,23 +923,9 @@ describe("CodexConnector.install — hooks.json schema", () => {
 		process.argv[1] = "/$bunfs/root/signet";
 
 		writeFileSync(
-			hooksPath,
-			JSON.stringify({
-				hooks: {
-					SessionStart: [
-						{
-							hooks: [
-								{
-									type: "command",
-									command: "/old/Codex.app/Contents/Resources/node /old/signet.js hook session-start -H codex",
-								},
-							],
-						},
-					],
-				},
-			}),
+			configPath,
+			"[mcp_servers.signet]\ncommand = '/old/Codex.app/Contents/Resources/node'\nargs = ['/old/mcp-stdio.js']\n",
 		);
-		writeFileSync(configPath, "[mcp_servers.signet]\ncommand = '/old/Codex.app/Contents/Resources/node'\n");
 
 		const result = await desktopRuntimeConnector(appPath).install(tempHome);
 		const hooks = readHooksJson().hooks as Record<string, Record<string, unknown>[]>;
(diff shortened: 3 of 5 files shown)


NicholaiVogel (changes requested):
-

NicholaiVogel:
## Request changes — the fix is dead code in production + two real regressions

Reviewed the full diff, the connector source, the codex test suite, and the production launch path. The *design* is right (discover → validate → apply to Signet-owned paths only, preserve third-party hooks). But the "apply" wiring never fires in production, and two of the new code paths introduce regressions. All three verified empirically, not just by reading.

### 🔴 Critical: `resolveSignetEntry()` never matches in production — the discovered runtime is thrown away

`resolveSignetEntry()` gates on `basename(process.argv[1]) === "signet.js"` (`index.ts:59`). Production Signet is a **Bun-compiled native binary**: `dist/signetai/bin/signet.js` → `launch.js` spawns the binary with `process.argv.slice(2)`, and inside that binary `argv[1]` is the virtual bunfs path, not `signet.js`.

Empirical proof — I built a minimal `bun build --compile` probe:
```
$ ./probe-bin one two
argv: ["bun","/$bunfs/root/probe-bin","one","two"]
basename(argv[1]): probe-bin
basename check (the PR's gate): false
```
And `dist/signetai/bin/launch.js` confirms there's no node-mode fallback — if the binary is missing it `process.exit(1)`s; the connector code only ever runs inside the compiled binary where `argv[1] = /$bunfs/root/signet`.

**Consequence:** `resolveSignetEntry()` always returns `null` → `resolveSignetArgs(runtime)` never takes the `runtime && entry` branch → the discovered Codex Desktop node is computed by `resolveCodexDesktopNode()` and then discarded. Hooks/MCP fall back to bare PATH `signet`/`signet-mcp`. **The entire fix for #1024 is dead code in production.** The new test passes only because it mutates `process.argv[1] = signetEntry` (a `.js` path) at `config-toml.test.ts:~919` — a value that never occurs at runtime.

**Suggested fix:** don't gate on `argv[1]`. Derive the Signet entry from `process.execPath` (the native binary's install root, e.g. `join(dirname(process.execPath), "..", "bin", "signet.js")`), or reuse the existing `resolveSignetCliCommand()` layout logic in `libs/connector-base` — then `existsSync`-validate. Add a production-realistic test that does *not* mutate `argv[1]` (or simulates the bunfs argv) so this dead branch is caught.

### 🟡 Warning: `hasMissingSignetRuntime` crashes on malformed-but-valid `hooks.json` (regression)

`hasMissingSignetRuntime` (`index.ts:483`) traverses the hooks file unguarded: `groups.flatMap(...)` → `group.hooks.map(...)` → `handler.command`. I confirmed it throws on three malformed-but-valid-JSON inputs:
- `{hooks: {SessionStart: "x"}}` → `groups.flatMap is not a function`
- `{hooks: {SessionStart: [{hooks: "x"}]}}` → `group.hooks.map is not a function`
- `{hooks: {SessionStart: [{hooks: [null]}]}}` → `null is not an object (evaluating 'handler.command')`

This is a **regression**: pre-PR `install()` called `resolveSignetArgs()` directly and let the guarded `installCompatibilityHooks` handle such files. Post-PR, `install()` calls `hasMissingSignetRuntime(readHooksFile(...), configPath)` first (line 1002), so a hand-edited or partial `hooks.json` that previously installed fine now aborts the whole install with an uncaught TypeError. Every sibling reader guards this way — `isSignetMatcherGroup` (`index.ts:471` `if (!Array.isArray(hooksArr))`), `removeSignetEntries` (`if (!Array.isArray(groups)) continue;`).

**Suggested fix:** mirror the sibling guards: `if (!Array.isArray(groups)) continue;` per event, `if (!Array.isArray(group?.hooks)) continue;` per group, `if (typeof handler?.command !== "string") continue;` per handler.

### 🟡 Warning: `codexDesktopNodeCandidates` is symlink-blind

`readdirSync(root, { withFileTypes: true })` then `entry.isFile()` / `entry.isDirectory()`. Dirent methods use **lstat semantics** — I confirmed a symlinked `node` reports `isFile: false, isDirectory: false, isSymbolicLink: true`. So:
- If Codex.app ships `node` as a symlink to a versioned binary (e.g. `bin/node -> node-v22.14.0`, a common bundler/version-manager pattern), the candidate is silently dropped.
- Symlinked runtime directories under `Contents/Resources` aren't recursed into.

In both cases `resolveCodexDesktopNode` returns `null` and the feature silently no-ops. The test fixture uses a real file, so this isn't covered.

**Suggested fix:** accept symlinks — `(entry.isFile() || entry.isSymbolicLink()) && entry.name === "node"` for the file match, and use `statSync(path)` (follows symlinks) for the directory recursion check. Wrap the `statSync` in try/catch to stay graceful on broken links.

### What's sound (verified)
- Third-party hooks are preserved correctly — `installCompatibilityHooks` calls `removeSignetEntries` then re-adds Signet's own, so only Signet-owned commands get the new runtime path.
- Null-runtime fallback is clean — when Desktop isn't installed, `resolveSignetArgs(null)` → `resolveSignetCliCommand()` (pre-existing behavior, no regression in the common non-Desktop case).
- Scan can't escape `Codex.app` (rooted at `Contents/Resources`, depth-bounded at 8).
- Stale-detection regex is safe — no ReDoS, correct node-path extraction across quoted/Windows/`node_modules` cases.
- Validation (`spawnSync --version`, 5s timeout, strict `vN.N.N` regex) is sound; no shell injection (path passed directly).

### Minor (not blocking)
- The stale-runtime warning fires based purely on `hasMissingSignetRuntime`, independent of whether `resolveCodexDesktopNode` found a replacement. If Desktop was uninstalled (runtime=null) but stale paths remain, the warning asserts a "refresh" that actually fell back to PATH `signet`. Consider splitting into "replaced" vs "no replacement found, used PATH fallback."
- `entry.isDirectory() && entry.name !== "app.asar"` contains a dead predicate — `app.asar` is a file, so `isDirectory()` is already false. If the intent was to skip `app.asar.unpacked` (a real directory), that isn't excluded.

### How I validated
- Built core+connectors in an isolated worktree on the PR head; ran `integrations/codex/connector/src/config-toml.test.ts` — 59/59 pass (including the new migration test, which passes *only* because of the `argv[1]` mutation).
- Built a `bun build --compile` probe confirming `argv[1]` inside a compiled binary is `/$bunfs/root/<name>`.
- Reproduced all three `hasMissingSignetRuntime` crashes and the symlink blindness directly.

The approach is right; the production wiring just needs to not depend on `argv[1]`.

aaf2tbz:
Addressed the review findings and rebased this PR onto the latest main.

- Resolve the packaged Signet entry from `process.execPath`, so Bun-compiled production binaries no longer depend on the virtual `argv[1]` path.
- Guard malformed-but-valid `hooks.json` structures during stale-runtime detection.
- Follow symlinked Codex runtime directories and `node` binaries; broken links are ignored safely.
- Clarify the stale-runtime warning when no replacement runtime is found.

Added regression coverage for Bun-style argv, malformed hooks files, and symlinked runtimes. Verified with `bun test integrations/codex/connector/src/config-toml.test.ts` (61 passing), Biome, and the connector TypeScript check.

PR #1067, opened 2026-08-04
fix(daemon): remove WAL checkpoint from startup recovery (database lock crash)

NicholaiVogel:
## What

The daemon crashes with `SQLiteError: database is locked` during startup, 5 seconds after worker initialization begins. Root cause: the explicit `PRAGMA wal_checkpoint(TRUNCATE)` in startup recovery, run via `.prepare().get()` on the bun:sqlite write connection, appeared to hold a write lock that prevented all subsequent `BEGIN IMMEDIATE` transactions.

## Evidence

Foreground daemon output:
```
15:05:36 INFO  startup-recovery: Running startup recovery
15:05:46 WARN  startup-recovery: Orphaned pass sweep failed: database is locked
15:05:46 WARN  resources: Event loop blocked (8035ms)
15:05:46 INFO  system-pressure: Startup grace period active
15:05:51 ERROR daemon: Fatal error: SQLiteError: database is locked
```

The 5-second gap between worker init and the fatal crash matches exactly the `busy_timeout = 5000` setting. The write connection was permanently locked after the checkpoint's prepared statement executed.

## Fix

Remove the WAL checkpoint from startup recovery entirely. SQLite's built-in auto-checkpoint handles WAL bloat under normal operation. The startup recovery's critical work (dead job purge, staging cleanup, orphaned pass sweep) is unaffected.

The `DbAccessor.checkpointWal()` method is kept (changed from `.prepare().get()` to `.exec()` to match the existing migration-backup pattern) for future post-startup callers, but is no longer called during the boot path.

## Checklist

### DIFF
--- a/platform/daemon/src/db-accessor.ts
+++ b/platform/daemon/src/db-accessor.ts
@@ -1041,7 +1041,7 @@ function createAccessor(writeConn: SqliteDatabase): DbAccessor {
 
 		checkpointWal(): void {
 			if (closed) throw new Error("DbAccessor is closed");
-			writeConn.prepare("PRAGMA wal_checkpoint(TRUNCATE)").get();
+			writeConn.exec("PRAGMA wal_checkpoint(TRUNCATE)");
 		},
 
 		close(): void {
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Closing — removing the checkpoint is a bandaid. Root cause is the async yield in drainWriteBatches during startup recovery allowing pending boot operations to interfere with the write connection. Will replace with a synchronous recovery path.

Issue #1148, opened 2026-08-07
daemon process vanishes under concurrent automation-bot startup load (session-start + secret.exec + watcher indexing); no core, no log, no OOM


NicholaiVogel:
## Problem

The daemon process vanishes with zero trace under concurrent load from local automation scripts. Across five deaths over ~24h, every death minute coincides with an automation bot (a Minecraft chat bridge, a camera/motion sentinel) starting up and using the daemon's API — session-start hooks, `secret.exec`, session capture — on top of the Obsidian watcher indexing. The process disappears with no core dump, no JS error, no shutdown log, and no OOM record. The crash mechanism is unpinned; only the load profile at the kill moment is consistent.

## Evidence (crash-watcher reports, live box)

- Deaths: 09:24Z, 20:07Z, 02:11Z, 08:11Z, 09:47Z (all Aug 5-7).
- At each death minute, the user journal shows an automation bot starting: `minecraft-ant starting up` (RCON bridge to 10.0.0.128:25575, trigger @ant, cwd ~/.agents) and/or `[SENTINEL] Motion Tracking System (Desktop)` (OpenCV camera monitor).
- The death-moment daemon log shows the bot's daemon usage: `secret.exec_started` / `secret.exec_completed` (jobId) 27s before the sentinel's start log, `secret.listed`, session-start activity, alongside `Embedded Obsidian source chunks` / `Indexed native memory artifact` watcher lines.
- `coredumpctl info <dead pid>`: "No coredumps found" every time; `dmesg` restricted; user journal has no signet unit exit record.
- The bots do NOT invoke the signet CLI (no `signet daemon start/restart` in their scripts) — they use the daemon's HTTP API (secrets, session hooks, capture).

## Gap

The daemon gives no evidence of its own death: no exit-path logging, no signal handler, no unit-level exit status surfaced. Without one of those, every post-mortem stalls at "vanished."

## Expected behavior

- The daemon should surface its own death: log the exit path (shutdown, fatal error, external signal) and/or register signal handlers that write a final log line before exiting on SIGTERM/SIGINT/SIGKILL-where-possible.
- The unit/runtime should record the exit status so external kills are distinguishable from crashes.
- The load-trigger correlation (session-start + secret.exec + watcher indexing under bot startup) should be load-tested or instrumented — concurrent session-start and secrets work should not be able to take the process down.

## Related

- #1142 (watcher retry loop — fixed 0.164.5; the watcher is still the ambient load)
[...]

Issue #1143, opened 2026-08-07
daemon bricked at startup: source-deletion tombstone cleanup runs at route registration before DB accessor init (getDbAccessor crash)


NicholaiVogel:
## Problem

The daemon fails to start whenever `source-deletion-tombstones.json` has entries. `registerSourcesRoutes` (sources-routes.ts:117-121) calls `cleanupSourceDeletionTombstones` synchronously at route-registration time, which calls `purgeNativeMemorySourceArtifacts` (native-memory-sources.ts:595) — and that function calls `getDbAccessor()` unconditionally (native-memory-sources.ts:609). The DB accessor is not initialized yet at route-registration time, so startup crashes with:

```
error: DbAccessor not initialised — call initDbAccessor() first
    at getDbAccessor
    at purgeNativeMemorySourceArtifacts
    at cleanupSourceDeletionTombstones
    at registerSourcesRoutes
```

## Repro

1. Have a source-deletion tombstone present (e.g. a source disabled and deleted by crash-loop protection, or any source removal that writes `~/.agents/.daemon/source-deletion-tombstones.json`).
2. Start the daemon → it crashes at route registration before the DB accessor initializes.
3. The daemon is bricked until the tombstone file is manually cleared (the only workaround found in the field).

## Live evidence

- A daemon death triggered crash-loop protection to disable the Obsidian vault source and write a tombstone (deletedAt = death minute).
- Every subsequent `signet daemon start` crashed with the stack above; the daemon could not start at all.
- Clearing `~/.agents/.daemon/source-deletion-tombstones.json` (backed up first) allowed the daemon to start normally.
- The tombstone cleanup is correctly gated on `tombstones.length === 0` returning early (sources-routes.ts:446) — the crash is the ORDER: routes register before `initDbAccessor()`.

## Expected behavior

- The tombstone cleanup must not run before the DB accessor is initialized: defer `cleanupSourceDeletionTombstones` until after `initDbAccessor()` (or make the purge lazy / run it in the startup sequence after DB init).
- A tombstone file must never brick the daemon: startup should tolerate tombstone-processing failure (log + defer) rather than crash.

## Impact

Any install with a source-deletion tombstone (including crash-loop-disabled sources) cannot start the daemon. This converts a recoverable wedge into a permanent outage requiring manual file surgery.

## Related

- #1142 (watcher ENOENT retry loop — the crash-loop disable that created the tombstone in the field)
- #1136 (silent daemon deaths — the class this blocked recovery from)

Issue #1142, opened 2026-08-07
obsidian watcher retry-loops on deleted artifact paths: sync statx every 250ms with no backoff, desyncs FTS, 16-17s event-loop blocks, wedges and kills the daemon


NicholaiVogel:
## Problem

The native-memory-artifact watcher retry-loops forever on artifact paths whose files were deleted from disk. Each retry is a synchronous `statx` on the main thread; with no backoff the loop hammers the same ENOENT every ~250ms, desyncs the FTS index (triggering the FTS consistency repair loop), and produces sustained 16-17s event-loop blocks. The daemon wedges and, under the sustained critical pressure, dies. Four daemon deaths over ~24h share this signature.

## Evidence (live box, crash reports collected by a daemon-watch watcher)

- 1,629 consecutive `Failed reading native memory artifact` warnings in ~18 minutes (every ~250ms), all ENOENT on the same three paths:
  - `/mnt/work/obsidian-vault/routines/agent-memory/daily-brief/history/2026-08-03/morning.md`
  - `/mnt/work/obsidian-vault/routines/agent-memory/daily-brief/history/2026-08-03/midday.md`
  - `/mnt/work/obsidian-vault/routines/agent-memory/daily-brief/history/2026-08-03/evening.md`
- The files do not exist (deleted from the vault); the watcher's artifact index still references them and re-reads them on every scan.
- Concurrently: `Event loop critically blocked (16-17s)` every ~30s; `repair: FTS consistency check` reported transient mismatches (memCount 5, ftsCount 10) and triggered the synchronous FTS repair/retention sweep — the repair blocks ~17s.
- Death/wedge sequence (crash watcher, reason `wedged: health failing >60s, pid alive`): blocks start, pressure gate goes critical, health probes time out, the daemon wedges; a kill or continued starvation ends the process. No coredump, no OOM, no JS error — consistent with a starved/killed process under sustained loop pressure.

## Root cause

The watcher's artifact scan does not handle a path that disappears between indexing and read:
- No tombstone/drop of the indexed path on ENOENT — the same path is retried on every scan, forever.
- No backoff or failure cooldown — retries fire every ~250ms.
- The read/stat is synchronous on the main thread, so the retry loop monopolizes the event loop (16-17s blocks measured).
- The desync between the artifact index and disk surfaces as FTS consistency mismatches, which trigger the repair path, compounding the blocks.

## Expected behavior

[...]

Issue #1136, opened 2026-08-06
daemon died silently mid-ingest: SIGKILL profile — concurrent signet daemon start/restart can still kill the running daemon via the ps-cmdline scan (stopDaemon)


NicholaiVogel:
## Problem

The daemon died silently after ~5.5 hours of uptime, mid vault-ingest (native-embedding burst + reindexMemoryArtifacts). No core dump, no OOM trace, no JS error log, no unit death record — the log simply stops. That profile matches SIGKILL, and the known mechanism is still unfixed: `stopDaemon` (surfaces/cli/src/lib/runtime.ts) locates the daemon via a `ps` + cmdline scan and SIGTERM/SIGKILLs every matching process. Any concurrent `signet daemon start`/`restart` from another harness session, cron, or automation script kills the running daemon with zero trace. `bin/launch.js` re-execs the CLI into the same native binary, so the scan can also hit sibling CLI processes (the "Terminated" symptom).

## Evidence (live box, 2026-08-06)

- Death at ~09:24 UTC; last log lines are a burst of `Indexed native memory artifact` / `Embedded Obsidian source chunks` / `reindexMemoryArtifacts completed`, then silence.
- `coredumpctl list` shows no signet entry in the window (core_pattern routes to systemd-coredump; a real segfault/abort would appear).
- `journalctl --user` in the death window shows the user's automation scripts (minecraft-ant bot, motion sentinel) starting at the same minute — plausible concurrent `signet` invokers.
- No `"Orphaned by daemon restart"`-style shutdown log; the process just vanished.
- Same class as the earlier death-spiral investigation: `signet status`/`daemon start` from cron watchdogs and other sessions can SIGKILL a busy daemon mid-work.

## Expected behavior

- `stopDaemon` must not kill a merely-busy daemon: match the daemon by its systemd-run unit MainPID or an explicit `SIGNET_DAEMON_ENTRYPOINT=1` marker, never by a ps-cmdline scan that can match unrelated CLI processes.
- `startDaemon` should treat an already-running daemon as success (probe first; never SIGTERM/SIGKILL to "reset" it).
- A daemon death should never be silent: if the process is killed externally, the unit should surface the exit status.

## Related

- The death-spiral investigation (embeddings_staging drain + cron watchdogs) — same stopDaemon mechanism; the daemon-side fixes were proposed then and never landed.
- #1094 / #1106 (earlier silent-loop wedges — different cause, same symptom class of "daemon unavailable").

[...]

PR #1151, opened 2026-08-07
fix(daemon): surface daemon deaths with lifecycle record, exit-path logs, and concurrency load test (#1148)

NicholaiVogel:
## Summary

The daemon vanished with zero trace under concurrent automation-bot startup
load (Minecraft chat bridge, camera/motion sentinel hitting session-start
hooks, `secret.exec`, session capture on top of the Obsidian watcher). Five
deaths over ~24h: no core dump, no JS error, no shutdown log, no OOM record.
Two mechanism gaps made every post-mortem stall at "vanished":

1. No durable record survived an unclean death. SIGKILL cannot be caught, so
   nothing distinguished "killed" from "crashed" from "clean" after the fact.
2. Even catchable exits could lose their final log lines: the logger buffers
   file writes on a 1s timer and `logger.shutdown()` (the flush) was never
   called on the exit path — "Shutting down" could sit in the buffer when
   `process.exit()` ran.

## Changes

- **Lifecycle record** (`platform/daemon/src/lifecycle.ts`): the daemon writes
  `.daemon/lifecycle.json` atomically (tmp + rename) at startup
  (`starting` → `running`) and on every catchable exit (`clean` with
  exit path + code for SIGTERM/SIGINT, `error` with message for
  uncaughtException/unhandledRejection/startup failure). A SIGKILL/OOM/crash
  death leaves the record stuck at `starting`/`running` — that stuck state is
  the signal.
- **Single exit path** (`daemon.ts`): `requestShutdown(reason, exitCode, error?)`
  logs `Received <reason>; shutting down`, writes the lifecycle record, runs
  the existing cleanup, then flushes the logger buffer before exiting. Update
  handoffs and boot failures route through it too (the update path skips the
  draining cleanup so the replacement process can bind immediately).
- **Surfacing** (`signet status` / `signet doctor`): when the daemon is down,
  the CLI reads the lifecycle record and reports `daemon_exit_clean` (info),
  `daemon_exit_error` (error with the message), or `daemon_exit_unrecorded`
  (warn: killed or crashed, gated on the recorded pid actually being gone, with
  a journalctl pointer for systemd launches).
- **Load test** (`scripts/load-test-daemon.ts`): reproduces the bot-startup
  traffic profile — session-start/end hooks, secrets listing, recall, status
  polls, optional writes — with a 250ms health watchdog, death-event
  recording, latency percentiles, and a pass/fail verdict.

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/daemon`

## Screenshots

N/A — daemon lifecycle and CLI diagnostics, no UI change.

## PR Readiness (MANDATORY)

Notes: no new/changed data queries (the lifecycle file is internal daemon
state, not a DB query); no new admin/mutation endpoints; docs updated in
`docs/DAEMON.md`.

## Validation

- `bun test` lifecycle/logger/health/runtime suites: 69 pass.
- `bun run typecheck` (root): all workspaces pass. `bunx biome check` on the
  touched files: clean.
- Live on a scratch daemon:
  - SIGTERM → lifecycle `clean`/`signal:SIGTERM`/exit 0; log tail contains
    `Received signal:SIGTERM; shutting down` (flush works).
  - SIGKILL → lifecycle stays `running`; CLI reader surfaces it as an
    unrecorded death.
  - Corrupt `memories.db` → lifecycle `error`/`error:startup` with the message
    and an on-disk log tail (boot failures are no longer silent).
  - Double SIGTERM → clean record, flush still lands.
  - Load test pass: 4,885 requests at 12 concurrency incl. writes, 0 request
    failures, p95 64ms, daemon healthy. Kill demo: SIGKILL mid-run → death
    event recorded, verdict fail.

## Related

- #1148 (this issue), #1136 (silent daemon deaths — mechanism-gap follow-up),
  #1142 (watcher retry loop — fixed 0.164.5), #1143 (startup brick — fixed
  0.164.4)

AI-assisted contribution. Adversarial review findings and their closures are
documented in the PR comments.

PR checklist validated against the current body at merge time.

### DIFF
--- a/docs/DAEMON.md
+++ b/docs/DAEMON.md
@@ -33,6 +33,33 @@ signet status          # Check status
 Top-level aliases `signet start`, `signet stop`, and `signet restart`
 still exist, but `signet daemon ...` is the preferred command surface.
 
+### Death evidence and the lifecycle record
+
+The daemon writes its state to `<workspace>/.daemon/lifecycle.json`: `starting`
+at boot, `running` once the HTTP server is ready, and a terminal `clean` or
+`error` record (with exit path, exit code, and timestamp) on every catchable
+exit — SIGTERM/SIGINT shutdowns and fatal errors. The record is written
+synchronously and atomically, and the daemon flushes its log buffer before
+exiting so the final log lines are never lost.
+
+A process that is killed with SIGKILL, OOM-killed, or hard-crashed cannot write
+a terminal record, so the file stays at `starting`/`running` — that stuck state
+is the signal. `signet status` and `signet doctor` read the record whenever the
+daemon is down and surface the distinction:
+
+- `Last daemon exit was clean` (info) — the daemon shut down normally.
+- `Daemon exited after an internal error` (error) — a fatal error with the
+  message, worth reporting upstream.
+- `Previous daemon exit was not recorded as clean` (warn) — killed or crashed.
+  The finding points at the daemon log tail and, on Linux systemd launches, the
+  transient unit's journald exit status (`journalctl --user -u
+  signet-daemon-<pid>`), which distinguishes signals, core dumps, and OOM.
+
+A concurrency load test for the bot-startup traffic profile (session-start
+hooks, secrets, recall, status, optional writes) ships at
+`scripts/load-test-daemon.ts`; run it against a live or scratch daemon with
+`bun scripts/load-test-daemon.ts --port <port> [--writes]`.
+
 ### Via System Service
 
 The daemon can be installed as a system service for auto-start on boot.
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -1642,6 +1642,25 @@ function buildLifecycleRecord(state: DaemonLifecycle["state"], extra: Partial<Da
 	};
 }
 
+function flushAndExit(exitCode: number): void {
+	// The logger buffers file writes and flushes on a 1s timer; without an
+	// explicit flush the final log lines can be lost on exit.
+	logger.shutdown();
+	process.exit(exitCode);
+}
+
+function buildTerminalLifecycleRecord(reason: string, exitCode: number, error?: unknown): DaemonLifecycle {
+	return buildLifecycleRecord(error === undefined ? "clean" : "error", {
+		exitedAt: new Date().toISOString(),
+		exitCode,
+		reason,
+		...(error !== undefined ? { error: error instanceof Error ? error.message : String(error) } : {}),
+	});
+}
+
+/** Bounds the draining cleanup so a wedged shutdown can never zombie the daemon. */
+const SHUTDOWN_CLEANUP_DEADLINE_MS = 20_000;
+
 /**
  * Single exit path for every catchable termination (signals and fatal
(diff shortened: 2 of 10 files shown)


NicholaiVogel:
Adversarial review (background subagent, instructed to find real defects) found 6 issues; all closed in the follow-up commit.

1. **HIGH — exit paths bypassing the recorder.** `main().catch`, the update handoff, and the no-self-restart fallback called `process.exit` directly: a fatal boot error left the lifecycle stuck at `starting` (misread as "killed or crashed") with an empty log file (the fatal line sat in the unflushed 1s buffer). All three now route through `requestShutdown` — boot failures write `{state:error, reason:"error:startup", error:msg}` and flush; the update handoff writes `clean` with `reason:"update:..."` while skipping the draining cleanup (the replacement process needs the port immediately). Verified live: a corrupt `memories.db` now produces the `error:startup` record and an on-disk log tail.
2. **MEDIUM — fabricated journalctl pointer.** `SIGNET_DAEMON_UNIT` was in `daemonEnv` for every launch path, so macOS/launchd or detached-spawn daemons recorded a unit that never existed. Now set only via the systemd-run `--setenv`; fallback paths record no unit and the finding falls back to the generic evidence pointer.
3. **MEDIUM — false "killed or crashed" against a live daemon.** A daemon on a custom `SIGNET_PORT`, or one still binding at boot, is invisible to the fixed-port probe while its lifecycle record says `running` — doctor would have claimed death against a live process. The unrecorded finding is now gated on the recorded pid actually being gone (`process.kill(pid, 0)`). Regression test added.
4. **MEDIUM — second-signal exit dropped the flush.** The deliberate anti-wedge immediate exit on a second signal skipped `logger.shutdown()`, voiding the flush guarantee on double-Ctrl-C. The re-entry path now flushes before exiting. Verified live with two rapid SIGTERMs.
5. **LOW — load-test session-end always referenced never-started sessions** (each call minted its own key → guaranteed 400, real path never exercised). session-end now consumes keys from a pool populated by session-start. Verified: session-start + session-end both return 200.
6. **LOW — stale base.** Branch was based on pre-0.166.0 main; rebased onto `chore: release 0.166.0` (9523078dd). Diff vs `origin/main` is exactly the 10 intended files.

NicholaiVogel:
Second adversarial review (focused on the daemon/CLI files the first review's diff snapshot missed) returned SHIP-WITH-FIXES. Its findings 1, 2, 4, 5 were already closed by the first follow-up commit; the remaining actionable item is closed in this push:

**MEDIUM — terminal lifecycle record was written before cleanup completed.** A wedged cleanup (native bridge close, watcher close, pipeline stop have no timeouts) would have left a zombie process whose record already claimed `clean` — the lie made durable. `requestShutdown` now (a) writes the terminal record only after cleanup settles, and (b) enforces a 20s hard exit deadline: a cleanup that exceeds it logs `Shutdown cleanup timed out; forcing exit`, writes the record, flushes, and exits. A daemon can no longer hang in exit with a false clean record.

Re-verified live after the change: corrupt DB → `error`/`error:startup` + message; SIGTERM → `clean`/`signal:SIGTERM`/exit 0 + flushed log line; SIGKILL → record stuck at `running`.

Issue #1158, opened 2026-08-07
perf: /api/knowledge/stats 9.4s — missing composite index on memory_entity_mentions causes 30M-iteration join


aaf2tbz:
## Summary
`GET /api/knowledge/stats` takes ~9.4s of main-thread CPU on a 5.5K-entity install. The dashboard polls this endpoint every few seconds, so each poll blocks the event loop (~9s) — "Event loop critically blocked" warnings, unresponsive health, and a direct contributor to daemon instability.

## Root cause
`getKnowledgeStats()` in `platform/daemon/src/knowledge-graph.ts`, the `scopedMemoryRows` query:

```sql
SELECT COUNT(DISTINCT mem.memory_id)
FROM memory_entity_mentions mem
JOIN entities e ON e.id = mem.entity_id AND e.agent_id = ?
JOIN memories m ON m.id = mem.memory_id AND m.is_deleted = 0
WHERE COALESCE(e.status, 'active') = 'active'
```

With only `idx_memory_entity_mentions_entity(entity_id)` and the `PRIMARY KEY (memory_id, entity_id)`, SQLite drives the join from `entities` × `memories` and point-looks-up `mem` — a ~30M-iteration nested loop on 5.5K entities. `EXPLAIN QUERY PLAN` confirms `USE TEMP B-TREE FOR count(DISTINCT)` over the cross-product. Measured: 9.43s user CPU.

## Fix (PR-ready: migration 106 in this repo)
```sql
CREATE INDEX idx_memory_entity_mentions_entity_memory
  ON memory_entity_mentions(entity_id, memory_id);
```
Verified locally: the query drops from **9.43s → 13ms** (~725×), endpoint ~10.7s → 0.07s.

NicholaiVogel:
Closing: fixed by #1166 (merged 2026-08-08) — added the `memory_entity_mentions (entity_id, memory_id)` composite join index. `/api/knowledge/stats` no longer runs the 30M-iteration join.

Issue #1157, opened 2026-08-07
daemon exits when a Dreaming pass fails (unguarded runPass → unhandled rejection kills the process)


aaf2tbz:
## Summary
A failed Dreaming pass terminates the whole daemon. `platform/daemon/src/pipeline/dreaming-worker.ts` `check()` runs `await runPass(...)` inside the worker's `setTimeout` timer callback with no try/catch; `runPass` rethrows after recording the failure, the rejection propagates to the timer callback, and **Bun treats the unhandled rejection as fatal and exits the daemon**. launchd KeepAlive restarts it, the next 5-minute check fails again → infinite crash loop with no shutdown log line and no crash report (process just vanishes, exit 1).

## Evidence
- Daemon restarted every ~5 minutes for hours. Every restart was preceded by a `dreaming_passes` row with `status='failed'` (e.g. `401 AuthError` from the inference endpoint) and **nothing else in the log** — the process died silently.
- `launchctl list` shows the agent's last exit status 1.
- Fixing the underlying inference failure (re-storing a corrupted credential) makes passes succeed → daemon stable → confirms the failure→exit coupling.

## Suggested fix
```ts
try {
  await runPass(defaultAgentId, mode, undefined, scopes);
} catch (e) {
  logger.error("dreaming-worker", "Dreaming pass failed during scheduled sweep", undefined, {
    error: e instanceof Error ? e.message : String(e),
  });
}
```
The failure is already recorded in `dreaming_state` with a backoff; it must not take the daemon down.

## Related
- #1148 (daemon vanishes with zero trace) — likely same class: any unguarded rejection in a timer callback exits the process.
- #1059 (backlog death spiral) — different mechanism (watchdog SIGKILL), same symptom class.
- Systemic hardening: audit all `setTimeout`/`setInterval` callbacks for unguarded async work; consider a process-level unhandled-rejection handler that logs + continues for background workers.

aaf2tbz:
Follow-up found: the passes that fail (and crash the daemon via #1157) also never terminate when they don't fail — see the new issue on the agentic-loop non-termination (199 tool calls / 27 min / abort timer not enforced). Local mitigation on the affected install: advance `dreaming_state.evidence_cursor` past the stale backlog so passes only run on new evidence.

NicholaiVogel:
[...]

PR #1165, opened 2026-08-07
fix(daemon): guard Dreaming sweep against unhandled rejections; add memory_entity_mentions join index

aaf2tbz:
Fixes two issues from this debugging session:

**#1157 — daemon crash-loop on failed Dreaming passes.** `dreaming-worker.ts` ran `await runPass(...)` unguarded inside the worker's timer callback; a failed pass became an unhandled promise rejection and Bun terminated the daemon (silent death, exit 1, launchd restart loop every 5 min). The sweep now catches and logs pass failures — the failure is already recorded in `dreaming_state` with its backoff.

**#1158 — /api/knowledge/stats 9.4s.** `getKnowledgeStats` drove a ~30M-iteration cross-product because `memory_entity_mentions` lacked a composite `(entity_id, memory_id)` index. Migration 106 adds it; the query drops 9.43s → 13ms, endpoint ~10.7s → 0.07s.

Also comments the same unhandled-rejection root cause on #1148.

### DIFF
--- a/platform/core/src/migrations/106-memory-mention-join-index.ts
+++ b/platform/core/src/migrations/106-memory-mention-join-index.ts
@@ -1,0 +1,23 @@
+import type { MigrationDb } from "./index";
+
+/**
+ * Migration 106: entity-side composite index for memory_entity_mentions.
+ *
+ * The knowledge graph `COUNT(DISTINCT mem.memory_id)` join in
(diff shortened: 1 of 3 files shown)


aaf2tbz:
Closing — this was based on a stale local tree (0.163.10) and opened before running the repo's test suite. Redoing properly from origin/main with tests.

PR #1174, opened 2026-08-07
fix(dreaming): pin hygiene archive targets by subjectRef when flag details omit ids

### LINKED ISSUE #1168: dreaming pass never terminates in the agentic tool loop (199 tool calls, 27 min; abort timer not enforced)
## Summary

A Dreaming pass **never terminates** in the agentic loop: one pass made **199 tool calls over 27+ minutes** (91 `search_entities`, 46 `search_evidence`, 30 `get_entity`, 14 `apply_ontology_ops`) and never reached a final summary. The pass's 5-minute abort timer (see below) did not fire. With a large evidence backlog this effectively pegs the daemon (98% CPU when combined with dashboard polling) and, because a scheduled pass that eventually fails kills the daemon (#1157), the two bugs compound into the crash loop.

## Evidence

- Pass `c71aabd0` (scheduled `incremental-content`): 199 tool calls, still `status='running'` at 27 min, no completion. The model kept expanding its search (`search_evidence` with `offset: 9000` / `offset: 14000` / `chunkSize: 15000`) instead of converging.
- A manually triggered pass (`57ac7da3`) also ran 26+ min without completing.
- `inference-router.ts` creates the agent session with an abort timer: `setTimeout(() => void session.abort(), opts?.timeoutMs ?? provider.agentSessionTimeoutMs)` with `timeoutMs = cfg.timeout` (default 300_000ms). The pass ran ~6x longer than the timer — the abort either did not interrupt the in-flight tool loop or the loop's next iteration was scheduled past it.
- While the pass runs, the main thread shows sustained sqlite page-cache churn (tool executions) and the dashboard's knowledge polls queue behind it: observed "Event loop critically blocked (6426ms)" within 1 second of pass start.
- Root context: on this install the evidence cursor was NULL (a pass had not completed since July 23), so the backlog was ~181 transcripts / 25.8 MB and content mode tried to ingest it chunk-by-chunk forever.

## Reproducer

1. Let the evidence backlog accumulate (or leave `evidence_cursor` NULL with old transcripts).
2. Wait for the scheduled check to trigger a content pass (threshold / stale-backlog path).
3. Observe: pass row stays `running`, tool-call count grows without bound, daemon CPU climbs, event loop blocks.

## Suggested fixes

1. **Enforce the pass deadline**: the abort must interrupt an in-flight model call AND prevent the next tool-loop iteration from starting after the deadline (check the deadline at the top of every loop iteration, not only via the session abort).
2. **Cap tool rounds per pass** (e.g. 30-40 tool calls), independent of tokens — exploration loops then always terminate.
3. **Bound content-mode ingestion per pass** (e.g. process at most N sources / M tokens of the backlog per pass and advance the cursor only to what was surfaced — #1149 already restricts the watermark to surfaced evidence).
4. **Manual triggers should respect the early-exit**: `signet dream trigger` calls `runPass` directly with mode `incremental` — it starts a full agentic pass even when the backlog is zero (the early-exit path only covers specific modes), spending tokens to re-consolidate nothing.

Advancing `dreaming_state.evidence_cursor` past the old backlog (all evidence up to now) excludes it from the backlog count; passes then trigger only on new evidence and complete quickly. This is what a completed pass would have done itself.

NicholaiVogel:
## Summary

A Dreaming hygiene pass can never archive or merge anything it flags itself. Agent-minted `flag` ops write inspection facts into `details` but no id fields, while `attentionProvenance` required `attention.details.entityId === payload.target` to accept an archive. Every flag-then-archive cycle was rejected with "Hygiene archives require attention provenance", so the attention queue never drained: the worker re-triggered a pass every ~6 minutes around the clock (22 passes in 3h on the affected install), each burning 71k-483k tokens and 29%+ of a core in the in-process agent loop, applying 0 mutations. 93 of 109 pending attention items on the live install were stuck this way.

## Changes

- `platform/daemon/src/pipeline/dreaming-operations.ts`: `attentionProvenance` now pins the archive/merge target from the mandatory `subjectRef` (`entity:`, `aspect:`, `attribute:`, `link:`, `duplicate:` prefixes). The `details` id fields are an optional consistency cross-check: present and matching is fine, absent is fine, present and contradicting the subjectRef is still rejected (no redirect). `merge_aspects` already derived from subjectRef; the other five ops did not. `merge_entities` now falls back to `duplicate:<name>` from the subjectRef when `details.canonicalName` is missing.
- `platform/daemon/src/pipeline/dreaming-operations.test.ts`: four regression tests. Three reproduce the bug (same-batch archive, prior-batch `attention:<uuid>` archive, duplicate merge — all with flags that omit the details id) and fail on the old code; one pins the fail-closed invariant (details id contradicting the subjectRef is rejected).

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/daemon`

## Screenshots

N/A (no UI change)

## PR Readiness (MANDATORY)

No migrations touched.

## Testing

- [x] `bun test` passes (19/19 dreaming-operations; 44/44 sibling dreaming suites)
- [x] `bun run typecheck` passes (@signet/daemon build)
- [x] `bun run lint` passes (no new diagnostics on touched files)
- [x] N/A

Regression proof: the three #1168 tests fail when the fix is reverted (verified by reverse-applying the fix commit and re-running).

## AI disclosure

AI tools were used (see `Assisted-by` tags in commits).

## Notes

Root cause of the CPU investigation behind #1168. The timeout/abort enforcement from the issue is still worth doing as a backstop, but this fix is what stops the pass-every-6-minutes loop: the hygiene queue can now actually drain, so the cadence falls back to content-driven passes. Live evidence on the affected install: 77 of 182 `apply_ontology_ops` calls in 24h failed with the provenance error; 93 of 109 pending attention items lacked `entityId` in details.

Closes #1168

---

## Adversarial review resolution

An adversarial subagent reviewed this PR (live DB experiments + test runs). Verdict: ship-with-fixes. All findings fixed in 98cedd46f:

1. **HIGH (fixed)** — `merge_aspects` let `details.aspectId` override the subjectRef pin (the flag could name aspect A and merge a different aspect B). The subjectRef is now the pin; a contradictory `details.aspectId` rejects.
2. **MEDIUM (fixed)** — one flag authorized multiple hygiene ops in a batch. Each flag is now one-use: a second op citing the same attention fails with "Attention already consumed".
3. **LOW (fixed)** — a `duplicate:` flag with an empty canonical name could merge empty-name entities. Empty pinned names now reject.

New regression tests cover each. Suite: 22/22 dreaming-operations, 44/44 sibling dreaming suites, daemon build green. The three original regression tests still fail on the pre-fix code (verified both by me and by the reviewer in a scratch clone).

### DIFF
--- a/platform/daemon/src/pipeline/dreaming-operations.test.ts
+++ b/platform/daemon/src/pipeline/dreaming-operations.test.ts
@@ -115,6 +115,106 @@ describe("dreaming operations", () => {
 		).toEqual({ c: 0 });
 	});
 
+	it("archives an entity flagged without an entityId in details, pinned by subjectRef alone (#1168)", () => {
+		insertEntity("e-husk", "Legacy Husk", "legacy husk");
+		const result = applyDreamingOperations({
+			accessor: getDbAccessor(),
+			agentId: "agent-a",
+			actor: "dreaming",
+			passId: "pass-1",
+			operations: [
+				// Agent-minted flags carry inspection facts in details but no
+				// entityId; the subjectRef is the only id pin.
+				flag({ subjectRef: "entity:e-husk", details: { reason: "zero_active_attributes" } }),
+				{
+					operation: "archive_entity",
+					payload: { target: "e-husk", reason: "non-concrete" },
+					provenance: "attention:$0",
+				},
+			],
+		});
+		expect(result.ok).toBe(true);
+		expect(result.items[1]!.ok).toBe(true);
+		expect(
+			getDbAccessor().withReadDb((db) => db.prepare("SELECT status FROM entities WHERE id = ?").get("e-husk")),
+		).toEqual({ status: "archived" });
+		expect(
+			getDbAccessor().withReadDb(
+				(db) =>
+					db.prepare("SELECT COUNT(*) AS c FROM dreaming_attention WHERE resolved_at IS NULL").get() as { c: number },
+			),
+		).toEqual({ c: 0 });
+	});
+
+	it("archives an entity flagged without an entityId in a prior batch via attention:<uuid> (#1168)", () => {
+		insertEntity("e-husk", "Legacy Husk", "legacy husk");
+		const minted = applyDreamingOperations({
+			accessor: getDbAccessor(),
+			agentId: "agent-a",
+			actor: "dreaming",
+			operations: [flag({ subjectRef: "entity:e-husk", details: { reason: "zero_active_attributes" } })],
+		});
+		const attentionId = (minted.items[0]!.result as { attentionId: string | null }).attentionId!;
+		const result = applyDreamingOperations({
+			accessor: getDbAccessor(),
+			agentId: "agent-a",
+			actor: "dreaming",
+			passId: "pass-2",
+			operations: [
+				{ operation: "archive_entity", payload: { target: "e-husk" }, provenance: `attention:${attentionId}` },
+			],
+		});
+		expect(result.ok).toBe(true);
+		expect(
+			getDbAccessor().withReadDb((db) => db.prepare("SELECT status FROM entities WHERE id = ?").get("e-husk")),
+		).toEqual({ status: "archived" });
+	});
+
+	it("rejects an archive whose details id contradicts the flagged subjectRef", () => {
+		insertEntity("e-flagged", "Flagged", "flagged");
+		const result = applyDreamingOperations({
+			accessor: getDbAccessor(),
+			agentId: "agent-a",
+			actor: "dreaming",
+			operations: [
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Adversarial review complete (live DB experiments, suite runs, regression verification in a scratch clone). Verdict: ship-with-fixes — all three findings (1 HIGH, 1 MEDIUM, 1 LOW) fixed in 98cedd46f with regression tests each. Summary of findings and fixes is in the PR description. The headline fix is confirmed: agent-minted flags without details ids now archive/merge through the subjectRef pin, and the anti-redirect property holds for all six hygiene ops.

PR #1175, opened 2026-08-07
fix(daemon): enforce the agent-session deadline so Dreaming passes always terminate (#1168)

### LINKED ISSUE #1168: dreaming pass never terminates in the agentic tool loop (199 tool calls, 27 min; abort timer not enforced)

aaf2tbz:
## Summary

A Dreaming pass **never terminates** in the agentic loop (#1168): the router set an abort timer for the agent session but still `await`ed `session.prompt()` unconditionally. When the abort did not settle the in-flight prompt, `runAgent` hung forever — observed as a pass with 199 tool calls over 27+ minutes, which compounded with #1157 into the daemon crash loop.

This PR enforces the pass deadline at the router level: the prompt is raced against the deadline, the session is disposed either way, and the loop returns a clear timeout error when the deadline wins. The deadline is checked on every prompt call, so it stops both an in-flight model call and the next loop iteration.

## Changes

- `platform/daemon/src/inference-router.ts` — race `session.prompt()` against the configured deadline (`opts.timeoutMs ?? provider.agentSessionTimeoutMs`); dispose the session in both outcomes; surface an `Agent session exceeded the Nms deadline` error when the deadline wins.
- `platform/daemon/src/inference-router.test.ts` — regression test: a fake Pi agent whose prompt never settles must return at the deadline (~200ms) instead of hanging.

## Type

- [x] `fix` — bug fix

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-side fix, no UI.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes — `bun test platform/daemon/src/inference-router.test.ts` 12/12 (incl. the new never-settling-session regression)
- [x] `bun run lint` passes
- [x] Tested against running daemon — the deadline fix is verified live on the affected install: passes now terminate at the deadline instead of running 27+ min
- [ ] N/A

## AI disclosure

## Notes

Resolves #1168

### DIFF
--- a/platform/daemon/src/inference-router.test.ts
+++ b/platform/daemon/src/inference-router.test.ts
@@ -150,6 +150,81 @@ describe("InferenceRouter legacy API credentials", () => {
 		}
 	});
 
+	it("enforces the agent-session deadline when the agent loop never returns (#1168)", async () => {
+		// Regression for #1168: the router set an abort timer for the agent
+		// session but still awaited session.prompt() unconditionally — if the
+		// abort did not settle the prompt, runAgent (and the Dreaming pass)
+		// hung past every deadline. The prompt must be raced against the
+		// deadline so runAgent returns and disposes the session either way.
+		const dir = mkdtempSync(join(tmpdir(), "signet-router-agent-deadline-"));
+		const bin = join(dir, "fake-sleeping-agent.sh");
+		writeFileSync(
+			bin,
+			`#!/usr/bin/env bash
+printf '%s\\n' "$@" > ${JSON.stringify(join(dir, "args.txt"))}
+sleep 60
+printf 'never reached\\n'
+`,
+		);
+		chmodSync(bin, 0o755);
+		writeFileSync(
+			join(dir, "agent.yaml"),
+			`inference:
+  defaultPolicy: dreaming
+  targets:
+    dreaming:
+      executor: acpx
+      acpx:
+        agent: codex
+        bin: ${bin}
+        permissions: deny-all
+        hooks: disabled
+        terminal: false
+      models:
+        default:
+          model: gpt-5.4-mini
+  policies:
+    dreaming:
+      mode: strict
+      defaultTargets:
+        - dreaming/default
+  workloads:
+    memoryExtraction:
+      policy: dreaming
+`,
+		);
+		try {
+			const router = getOrCreateInferenceRouter(dir);
+			const startedAt = Date.now();
+			const result = await router.runAgent(
+				{ operation: "memory_extraction", promptPreview: "consolidate" },
+				"Use the supplied evidence and daemon tools.",
+				[],
+				{
+					timeoutMs: 200,
+					acpxMcp: {
+						agentId: "agent-deadline",
+						passId: "pass-deadline",
+						daemonUrl: "http://127.0.0.1:3850",
+						authorizationToken: "scoped-agent-token",
+					},
+				},
+			);
+			const elapsed = Date.now() - startedAt;
+			// The deadline is enforced: runAgent returned (no hang), failed the
+			// attempt with a timeout/deadline message, and finished in bounded
+			// time (the ACPX transport reports "timeout after Nms"; the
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Review note addressed: raised the pass deadline from 10 to 20 minutes (690cd49f6, `DEFAULT_DREAMING.timeout`).

Data behind the value:
- Local install (196 completed passes, Aug 4-8): median 2.1 min, p90 4.8 min, max 8.9 min. The 10-minute deadline never actually fired here.
- Fleet telemetry (PostHog, 44 `dreaming.pass` events): p50 26k tokens but p90 160k, p99 429k, max 569k tokens — 16% of passes exceed 100k tokens. Those passes take 20+ minutes on typical models; 10 min cuts them mid-flight. 20 min covers the p99 class while still bounding the runaway this deadline exists for (the #1168 pass ran 27+ min / 199 tool calls).
- Config remains overridable up to the existing 30-min clamp in `loadDreamingConfig`.

PR #1176, opened 2026-08-07
fix(daemon): probe the configured llama.cpp base_url in the embedding fallback (#1159)

### LINKED ISSUE #1159: embedding fallback ignores configured base_url (hardcoded :8080 probe) → silent ollama fallback
## Summary
The llama.cpp embedding fallback silently falls back to a different provider when the server isn't on the **hardcoded default port**, even when `embedding.base_url` is configured correctly. The fallback probe (`fetchLlamaCppEmbedding` / provider check) is called with no arguments and uses the compiled-in `DEFAULT_LLAMACPP_BASE_URL` (`http://127.0.0.1:8080`), ignoring the configured `base_url`.

## Evidence
- `embedding.base_url: http://127.0.0.1:8081` configured; llama.cpp LaunchAgent serving on 8081. Log: embeddings silently used ollama ("llama.cpp fallback failed" → ollama), and the fallback probe never hit 8081.
- Moving the server to 8080 (the compiled default) fixed it — the probe then succeeded. The config value was correct the whole time.

## Suggested fix
Pass the resolved `embedding.base_url` into the fallback probe / provider check instead of using the compiled default; if a probe function genuinely cannot take the URL, log the mismatch clearly (e.g. "embedding.base_url 8081 ignored, probing default 8080") instead of silently falling back to another provider.

aaf2tbz:
## Summary

The llama.cpp embedding fallback silently fell back to ollama whenever the configured `embedding.base_url` was not the compiled-in default port. `fetchNativeFallback` and `probeNativeFallback` hardcoded `DEFAULT_LLAMACPP_BASE_URL` (`http://127.0.0.1:8080`), so a correct config like `embedding.base_url: http://127.0.0.1:8081` was never probed — the probe hit 8080, failed, and the daemon silently used ollama for all embeddings.

This PR threads the resolved base URL through both fallback paths: model discovery (`findLlamaCppEmbeddingModel`) and the embedding fetch, in the cached-provider path and the probe path. An empty `base_url` still resolves to the compiled default, so existing configs are unaffected.

## Changes

- `platform/daemon/src/embedding-fetch.ts` — add `resolveLlamaCppFallbackBaseUrl(cfg)` (`cfg.base_url.trim() || DEFAULT_LLAMACPP_BASE_URL`) and use it in `fetchNativeFallback` and `probeNativeFallback`.
- `platform/daemon/src/embedding-fetch.test.ts` — regression: with a native failure and `base_url: http://localhost:8081`, the fallback probe must hit `:8081` and never `:8080`.

## Type

- [x] `fix` — bug fix

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-side fix, no UI.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes — `bun test platform/daemon/src/embedding-fetch.test.ts` 19/19 (incl. the new base_url regression)
- [x] `bun run lint` passes
- [ ] N/A

## AI disclosure

## Notes

Resolves #1159

### DIFF
--- a/platform/daemon/src/embedding-fetch.test.ts
+++ b/platform/daemon/src/embedding-fetch.test.ts
@@ -317,6 +317,39 @@ describe("fetchEmbedding", () => {
 		expect(capturedUrl).toContain("localhost:11434");
 	});
 
+	it("probes the configured llama.cpp base_url, not the compiled default (#1159)", async () => {
+		const llamaUrls: string[] = [];
+		globalThis.fetch = mock((url: string | URL | Request) => {
+			const urlStr = url.toString();
+			llamaUrls.push(urlStr);
+			if (urlStr.includes(":8081")) {
+				if (urlStr.endsWith("/v1/models")) {
+					return Promise.resolve(Response.json({ data: [{ id: "nomic-embed-text" }] }));
+				}
+				return Promise.resolve(Response.json({ data: [{ embedding: [0.5, 0.6] }] }));
+			}
+			if (urlStr.includes(":11434")) {
+				return Promise.resolve(Response.json({ embedding: [0.7, 0.8] }));
+			}
+			return Promise.resolve(new Response("unreachable", { status: 503 }));
+		}) as unknown as typeof fetch;
+
+		setNativeFallbackProvider(null);
+		setNativeEmbeddingProviderForTest(async () => {
+			throw new Error("native unavailable");
+		});
+		const result = await fetchEmbedding("test", {
+			provider: "native",
+			model: "nomic-embed-text-v1.5",
+			dimensions: 2,
+			base_url: "http://localhost:8081",
+		});
+
+		expect(result).toEqual([0.5, 0.6]);
+		expect(llamaUrls.some((u) => u.includes(":8081"))).toBe(true);
+		expect(llamaUrls.some((u) => u.includes(":8080"))).toBe(false);
+	});
+
 	it("single-flights failed local fallback discovery and negative-caches the result", async () => {
 		let nativeCalls = 0;
 		let llamaModelProbes = 0;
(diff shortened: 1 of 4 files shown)


NicholaiVogel:
Adversarial review found the fix incomplete for the reported #1159 environment; pushed a follow-up commit (fae8d1e66) to this branch.

**Finding (HIGH):** `checkEmbeddingProvider` (routes/utils.ts:966) had the same hardcoded-default-probe bug and runs before any fetch — at daemon startup (daemon.ts:2136) and on every /health check. With native enabled but unavailable (the default `warmNative` path), it probed `:8080`, failed, found ollama, and pinned `nativeFallbackProvider = "ollama"`. Every subsequent fetch short-circuits on that pinned state (`resolveNativeFallback`, embedding-fetch.ts:258), so the corrected `:8081` probe never ran. Reproduced live against this branch: embedding served from ollama, `:8081` never hit.

**Fix:** export `resolveLlamaCppFallbackBaseUrl` and thread it through the status-route probe at utils.ts:966, matching the two fetch paths. Same repro then serves llama.cpp with zero ollama hits.

**Regression test:** added to native-killswitch.test.ts (#1159) — asserts the status probe hits `:8081`, never `:8080`, and a post-status fetch uses the configured server. Verified it fails without the utils.ts change, passes with it. 37/37 tests across the touched suites, daemon typecheck clean, no new biome warnings.

NicholaiVogel (approved):
Approving. Fix verified locally (37/37 tests, daemon typecheck, no new lint warnings); the follow-up commit closes the status-route probe gap that left #1159 unresolved in the reported environment. CI fully green.

PR #1177, opened 2026-08-07
fix(daemon): restart the embedding-index build on config change and cap provider-failure retries (#1160)

### LINKED ISSUE #1160: embedding-index migration ignores config changes (stale persisted provider) → 98% CPU stuck loop
## Summary
The embedding-index migration can spin at ~98% CPU forever because it reads the provider from the **persisted staging profile** in `embedding_index_state`, which is never invalidated when `agent.yaml` changes. Config edits (e.g. switching `embedding.provider` from `native` to `llama-cpp`) never reach the running migration; it keeps calling `checkProvider('native')` → fails → `failed++` → returns → immediately retries in a tight loop.

## Evidence
- `embedding_index_state` stuck at `state='building'` for days (created Aug 5 23:08), daemon at ~98% CPU, ~1 chunk/min progress.
- The staging profile persisted `provider: native`; the daemon's configured provider was llama-cpp. Resetting the state (`UPDATE embedding_index_state SET state='ready', staging_profile_json=NULL`) broke the loop immediately and the rebuild then ran with the correct provider.

## Suggested fix
On config change (or on migration start), compare the persisted staging profile against the current `embedding.provider`/`embedding.model`; if they differ, invalidate the staging profile and rebuild from scratch. Also add a hard failure cap with backoff instead of an unbounded retry loop.

aaf2tbz:
## Summary

The embedding-index migration can spin at ~98% CPU forever because it reads the provider from the **persisted staging profile** in `embedding_index_state`, which is never invalidated when `agent.yaml` changes. The observed failure: the state stuck at `state='building'` for days with a `native` staging profile while the daemon's config said `llama-cpp` — the check kept probing the stale provider, failed, and retried in a tight loop.

This PR fixes both halves:

1. **Config-change invalidation** — every tick re-begins the build against the current configured profile. When the persisted staging fingerprint diverges from the current config, the build restarts (stale staging cleared, vector slot reset for the new dimensions) instead of spinning on the old provider.
2. **Failure cap with exponential backoff** — provider-unavailable checks now back off (`pollMs × 2^n`, capped at 60s) and the build aborts with a clear error after 6 consecutive failures, instead of retrying forever.

## Changes

- `platform/daemon/src/embedding-index-migration.ts` — tick re-begins against the current config and detects the fingerprint mismatch; consecutive-failure counter drives backoff and the hard abort (new constants `MAX_CONSECUTIVE_PROVIDER_FAILURES`, `MAX_PROVIDER_BACKOFF_MS`).
- `platform/daemon/src/embedding-index-migration.test.ts` — two regressions: (a) a stale persisted profile is invalidated and replaced with the current config's staging on the next tick; (b) a permanently unavailable provider aborts the build after exactly 6 checks instead of spinning.

## Type

- [x] `fix` — bug fix

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-side fix, no UI.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes — `bun test platform/daemon/src/embedding-index-migration.test.ts` 7/8 (the one failure is the pre-existing env-dependent sqlite-vec test that also fails on pristine `main` in this environment; both new regressions pass)
- [x] `bun run lint` passes
- [ ] N/A

## AI disclosure

## Notes

Resolves #1160

### DIFF
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -1465,6 +1465,10 @@ async function startPipelineRuntime(memoryCfg: ResolvedMemoryConfig, telemetry?:
 		embeddingIndexMigrationHandle = startEmbeddingIndexMigration({
 			accessor: getDbAccessor(),
 			configured: memoryCfg.embedding,
+			// Re-read agent.yaml each tick so a mid-build config edit restarts
+			// the build against the new profile instead of spinning on the
+			// stale persisted one (#1160).
+			readConfigured: () => loadMemoryConfig(AGENTS_DIR).embedding,
 			fetchEmbedding,
 			checkProvider: checkEmbeddingProvider,
 			pollMs: memoryCfg.pipelineV2.embeddingTracker.pollMs,
--- a/platform/daemon/src/embedding-index-migration.test.ts
+++ b/platform/daemon/src/embedding-index-migration.test.ts
@@ -1,15 +1,15 @@
 import { Database } from "bun:sqlite";
 import { describe, expect, it } from "bun:test";
 import { findSqliteVecExtension } from "@signet/core";
+import { up as embeddingIndexGenerations } from "../../core/src/migrations/091-embedding-index-generations";
+import type { DbAccessor, ReadDb, WriteDb } from "./db-accessor";
 import {
 	promoteStagingIndex,
 	stageEmbeddingBatch,
 	stagingCoverage,
 	startEmbeddingIndexMigration,
 } from "./embedding-index-migration";
-import { beginEmbeddingIndexBuild, ensureEmbeddingIndexState } from "./embedding-index-state";
-import { up as embeddingIndexGenerations } from "../../core/src/migrations/091-embedding-index-generations";
-import type { DbAccessor, ReadDb, WriteDb } from "./db-accessor";
+import { beginEmbeddingIndexBuild, ensureEmbeddingIndexState, readEmbeddingIndexState } from "./embedding-index-state";
 
 const VEC_EXTENSION = findSqliteVecExtension();
 
@@ -167,7 +167,6 @@ describe("staging promotion", () => {
 	it("keeps the promoted sqlite-vec virtual table queryable", () => {
 		if (!VEC_EXTENSION) return;
 		const raw = new Database(":memory:");
-		raw.loadExtension(VEC_EXTENSION);
 		embeddingIndexGenerations(raw as unknown as Parameters<typeof embeddingIndexGenerations>[0]);
 		raw.exec(`
 			CREATE TABLE memories (id TEXT PRIMARY KEY, embedding_model TEXT);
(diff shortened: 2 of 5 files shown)


NicholaiVogel (changes requested):
## Adversarial review: ship-with-fixes

Adversarial review of `b498a0395^..b498a0395` with live experiments in a clean worktree (bun 1.3.11, `@signet/core` + `@signet/daemon` built, suites run).

**The 6-failure cap is the real fix and it works.** Verified live: a permanently unavailable provider aborts the build after exactly 6 checks (state=`failed`, handle stops); the abort/restart state machine is fail-closed (a crash between the re-begin DELETE and the vector reset leaves a `building` state whose next tick aborts cleanly — no corruption, no partial staging rows); the catch-path `running=false` is parity with the fail state, strictly better than the old rescheduling spin.

**Three things stand between this and merge:**

### 1. HIGH — the config-change restart mechanism is unreachable in the daemon

`startEmbeddingIndexMigration` is called once at startup (`daemon.ts:1465`) with `configured: memoryCfg.embedding`, and `memoryCfg` is a frozen `loadMemoryConfig` snapshot (`daemon.ts:1915`, `1977`). The agent.yaml watcher (`daemon.ts:1029`) only reloads auth + harness sync; `restartPipelineRuntime` fires only on promotion and pause/resume toggle. So within a process, the persisted staging fingerprint was written from the same frozen config the tick re-begins with — `currentStaging.fingerprint !== state.staging.fingerprint` (`embedding-index-migration.ts:362`) can never be true in production, and the "Embedding config changed during migration" restart never fires. The tests reach the branch only by rewriting the state row directly (a write no production path performs). Live proof: fingerprint byte-identical across consecutive ticks with a frozen config.

Consequence: #1160's ~98% CPU spin is stopped by the cap (genuinely), but the build is NOT restarted against the edited config — it stays `failed` until a daemon restart. "Resolves #1160" overstates what ships.

Required: pass a live config getter into the migration (`configured: () => loadMemoryConfig(AGENTS_DIR).embedding`, re-invoked in the tick) so the fingerprint branch is actually reachable and a config edit restarts the build with the new provider — or scope the PR down to the cap and file a follow-up for the in-process rebuild.

### 2. MEDIUM — this PR breaks a previously-passing test

`embedding-index-migration.test.ts:167-217`: the PR removed `raw.loadExtension(VEC_EXTENSION)` from "keeps the promoted sqlite-vec virtual table queryable". Live: pristine main = 6/6 pass; with this PR the test fails (`no such table: vec_embeddings` at the first insert — bun's multi-statement `exec()` aborts silently on the failing `CREATE VIRTUAL TABLE`). The PR's note that the failure is "pre-existing env-dependent" is wrong; the PR caused it. Restore the `loadExtension` line — it is already guarded by `if (!VEC_EXTENSION) return;`.

### 3. MEDIUM — the new abort test depends on a known runtime quirk

`embedding-index-migration.test.ts:335-370` creates no `vec_embeddings_staging` table, so the resume path's `tableExists` decides the outcome: under bun:sqlite a no-row `.get()` returns `null` → `!== undefined` is true → the migration starts and the test passes; under better-sqlite3 (the native binary) it returns `undefined` → the migration returns null → `expect(handle).not.toBeNull()` fails. The test also breaks the moment `tableExists` is fixed to `== null`. Create the staging vector table in the test (as the restart test already does).

### Lower-priority notes

- The restart test passes while the restart itself fails in the test env (the reset needs vec0, which the plain-sqlite test build can't load) — assertions pass only because `fail` doesn't clear `staging_profile_json`. Regression value is genuine (both new tests fail on old code), but assert `state='building'` post-restart or stub vec0.
- The restart spans two transactions (re-begin DELETE, then vector reset); a crash between them is fail-closed (verified), not corrupting.
- Backoff comment math is off by one tick (cosmetic): delays between the 6 checks at pollMs=2 are 4/8/16/32/64ms, not "2+4+8+16+32+64".

**Verdict: ship-with-fixes.** Required before merge: (a) live config getter or scoped-down PR body, (b) restore `loadExtension`, (c) create the staging vector table in the abort test.

NicholaiVogel:
Applied the review fixes in e8f7a9863 — the requested changes are addressed:

1. **Live config getter (HIGH)** — the daemon now passes `readConfigured: () => loadMemoryConfig(AGENTS_DIR).embedding`, and the migration re-reads it each tick for both the re-begin and `configForProfile`. A mid-build `agent.yaml` edit now genuinely diverges the fingerprint → the build restarts against the new profile instead of probing the stale provider. (The frozen startup snapshot made the original fingerprint-mismatch branch unreachable in production — the tests could only reach it by rewriting the state row directly.)
   - Also fixed the sibling hole that the getter makes reachable: `beginEmbeddingIndexBuild` now abandons an in-flight build when the config flips back to the active generation, instead of resuming and eventually promoting a generation the config no longer wants.

2. **Restored `raw.loadExtension(VEC_EXTENSION)`** in "keeps the promoted sqlite-vec virtual table queryable" — the removal broke a test that passed on pristine main (verified: main = 6/6, with the PR = the sqlite-vec test fails). The earlier "pre-existing env-dependent failure" note was inaccurate.

3. **Abort test portability** — the "aborts after repeated provider-unavailable checks" test now creates the staging vector table, so it passes on both bun:sqlite and better-sqlite3 instead of depending on the null-vs-undefined no-row quirk in `tableExists` (it would have failed on the native runtime).

Also added a regression test that pins the getter contract (config edit mid-build → staging replaced with the live profile) and one for the abandon-on-flip-back behavior.

Verification: `bun test` on the three touched suites = 29 pass / 0 fail (including the restored sqlite-vec test), daemon `tsc --noEmit` clean, biome introduces no new violations (the 4 remaining in the state test file pre-exist on main).

PR #1182, opened 2026-08-07
fix(daemon): cap audit transcript size and archive by rename instead of duplicating (#1163)

### LINKED ISSUE #1163: transcript-audit writes 302MB × 2 per session (raw-transcript.log + latest.log)
## Summary
The transcript-audit writer can balloon the `.daemon/logs/transcripts/` directory to hundreds of MB for a single session: one session produced two **302,046,799-byte** files (a dated `--raw-transcript.log` and a `<session-id>--latest.log` — the same content written twice). The daemon's log reader then chokes parsing 600MB+ on every startup.

## Evidence
- `.daemon/logs/transcripts/` contained `2026-07-31T00-08-35.797Z--a7d19099e2eb7b2e--raw-transcript.log` and `a7d19099e2eb7b2e--latest.log`, both 302MB, written Aug 5 17:09.
- After archiving them, the daemon's startup scanning cost dropped to near-zero and repeated 45s event-loop blocks disappeared.

## Suggested fix
- Cap the audit transcript size (truncate/rotate beyond a few MB) — a 300MB session transcript is almost certainly not intended to be audited in full twice.
- Don't duplicate the full content in `latest.log` when it exceeds a threshold (store a pointer / truncated tail).
- Bound the startup scan of the transcripts dir (skip files above a size cap).

aaf2tbz:
## Summary

A runaway session once produced two **302MB audit files** — the dated `--raw-transcript.log` and the rolling `--latest.log`, the same content written twice — ballooning `.daemon/logs/transcripts/` to 600MB+ and choking the daemon's log reader.

This PR caps what is audited and eliminates the duplication:
- The audited content is capped at **8MB** with an explicit truncation marker.
- The dated archive is created by **renaming** the rolling latest file, so the full content is never written twice; the next capture recreates the latest.

## Changes

- `platform/daemon/src/transcript-audit.ts` — `capTranscriptAuditContent()` (8MB cap + marker); `writeTranscriptAudit()` writes the capped content once and archives via `renameSync` (with a direct-write fallback).
- `platform/daemon/src/transcript-audit.test.ts` — new regressions: the cap truncates with the marker and passes small content through; an archived write leaves one capped copy on disk (latest renamed, not duplicated); the no-timestamp path keeps the rolling latest.

## Type

- [x] `fix` — bug fix

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-side fix, no UI.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes — `transcript-audit.test.ts` 3/3 (new); existing hooks audit tests unaffected (they assert the final archive's content, which is unchanged for small transcripts)
- [x] `bun run lint` passes
- [ ] N/A

## AI disclosure

## Notes

Resolves #1163

### DIFF
--- a/platform/daemon/src/transcript-audit.test.ts
+++ b/platform/daemon/src/transcript-audit.test.ts
@@ -1,0 +1,67 @@
+import { describe, expect, it } from "bun:test";
+import { existsSync, mkdtempSync, readFileSync, rmSync, statSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { capTranscriptAuditContent, writeTranscriptAudit } from "./transcript-audit";
+
+describe("transcript audit sizing (#1163)", () => {
+	it("caps oversized audit transcripts with a truncation marker", () => {
+		const big = "x".repeat(10 * 1024 * 1024);
+		const capped = capTranscriptAuditContent(big);
+		expect(capped.length).toBeLessThan(big.length);
+		expect(capped.length).toBeLessThanOrEqual(8 * 1024 * 1024 + 200);
+		expect(capped.endsWith("chars omitted] ...\n")).toBe(true);
+		expect(capTranscriptAuditContent("small transcript")).toBe("small transcript");
+	});
+
+	it("archives by renaming the latest instead of writing the content twice", () => {
+		const root = mkdtempSync(join(tmpdir(), "signet-audit-"));
+		try {
+			// 10MB raw — well over the 8MB cap — so a duplicated write would
+			// leave ~20MB of audit files for one session.
+			const big = "y".repeat(10 * 1024 * 1024);
+			const result = writeTranscriptAudit({
+				basePath: root,
+				agentId: "default",
+				sessionId: "sess-1",
+				sessionKey: "sess-1",
+				rawTranscript: big,
+				capturedAt: "2026-08-07T00:00:00.000Z",
+			});
+			expect(result).not.toBeNull();
+			const finalPath = result?.finalPath;
+			expect(finalPath).toBeDefined();
+			if (!finalPath) throw new Error("finalPath missing");
+
+			// The archive exists with the capped content (~8MB cap, not the
+			// full 10MB), and the latest was renamed away rather than
+			// duplicated — one copy on disk, not two.
+			expect(statSync(finalPath).size).toBeLessThan(9 * 1024 * 1024);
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Adversarial review complete. Verified live in a worktree: new audit tests 3/3, hooks suite 174/0, worker suite 5/5, biome + daemon build clean (after the fix below).

**Blocker found and fixed on the branch.** The rename-archive changes the contract that `transcript-capture-worker.test.ts` "keeps raw audit logs when normalized transcript has no conversation turns" depended on: after an archived capture, the rolling `--latest.log` no longer exists (it was renamed into the dated archive). That test failed 1/5 on the PR as submitted — CI didn't catch it because no workflow on this PR runs `bun test` (Docker smoke only boots the stack and hits /health). I updated the assertion to pin the new contract (raw-transcript archive present, latest renamed away) with a comment naming #1163. Suites re-run after the fix and the rebase.

**Non-blocking notes for follow-up:**
- `TranscriptAuditWrite.latestPath` is typed as a required path, but after an archiving write the file at `latestPath` no longer exists. No current caller reads it post-write (the capture worker ignores the return; the hooks path never passes `capturedAt`), so it's latent — worth a comment on the interface so it stays that way.
- `transcript-health.ts` counts `--latest.log` files for `/api/diagnostics/transcripts`. Captured sessions now contribute zero latest logs (renamed away), so that metric drifts down to only hooks-path writes. Diagnostics-only, but expect the number to mean something different after this lands.
- The cap is 8M UTF-16 code units, not bytes: multi-byte transcripts can reach ~16MB on disk. Still bounded — the 302MB runaway is impossible — and the marker text says "chars", which is accurate.
- Design question: truncation keeps the transcript head and drops the tail. For a runaway session the tail is usually the interesting part. Keeping the head (seed/context) is defensible; flagging so it's a deliberate choice.

Also: the branch was based on 0.170.0 (6 releases behind main); rebased onto current main. `transcript-audit.ts` was untouched on main since the branch point, so the change applies cleanly.

Checked and clear: the rename can't misattribute content under concurrency — `writeFileSync` + `renameSync` share one synchronous frame and the capture worker is single-flight, so an interleaved write cannot land between them.

PR #1186, opened 2026-08-07
fix(daemon): resolve attention flags the agent judged and declined (#1185)

### LINKED ISSUE #1185: Dreaming passes re-trigger forever on declined attention (flags never resolve without applied ops)
## Summary

Dreaming attention flags only resolve when an ontology op applies for their subject. When the agent deliberately **keeps** a candidate (the "declined" judgment — e.g. a live entity the hygiene detector flagged as `non_concrete_entity_type`, or an over-cap aspect the model chose not to consolidate), the flag stays `pending` forever. The scheduled check sees pending attention every cycle, so it re-triggers a fresh pass that re-investigates the same declined work — indefinitely, burning tokens on every run.

## Evidence

- After the initial hygiene drain (50 flags → 0), four flags persisted across every pass: `3ab20762` (MetalSharp shim, `non_concrete_entity_type`), `5261a7eb`, `d7f31132`, `557c38ca` (`aspect_over_cap`, 25 aspects vs the 20 cap). The passes listed them via `attention_list`, applied nothing for them (the "keep" judgment), and the flags never resolved.
- The enqueue upsert only reopens a flag when its `details_json` changes or `reopen: true` is passed (the hygiene enqueue already passes `reopen: false`), so a resolved keep stays resolved until the subject's state actually moves.

## Suggested fix

- Snapshot the pending attention at pass start (per scope).
- Track the flags the agent actually surfaced via `attention_list` during the pass.
- On completion (hygiene / mixed modes only — never a content-only pass), resolve the surfaced flags that are still pending: applied flags are already resolved in the apply transaction; declined flags now close too. Flags the agent never listed stay pending for a later pass.

A local build with this change resolved the four stubborn flags to 0 and the passes stopped firing on declined work.

aaf2tbz:
## Summary

Attention flags only resolve when an ontology op applies for their subject. When the agent deliberately **keeps** a candidate (the "declined" judgment — e.g. a live entity flagged `non_concrete_entity_type`, or an over-cap aspect the model chose not to consolidate), the flag stays pending forever and the check re-triggers a fresh pass every cycle to re-investigate the same declined work — indefinitely. Observed: four flags persisting across every pass (the model listed them, declined, and the flags never resolved).

The fix gives the agent an **affirmative decline**: a new `decline_attention` op in `apply_ontology_ops` closes exactly one pending record the agent inspected and judged to keep. A deliberate keep resolves and stops re-triggering passes, while records the agent could not complete stay pending — the runbook's "deferred with reasons in the pass log" contract is preserved.

An earlier revision swept every record the agent surfaced via `attention_list` at pass end. Adversarial review proved that conflates judgment with listing: an agent that listed the queue, applied a subset, and explicitly deferred the rest had its deferred work silently resolved forever (the enqueue upsert reopens a record only when its details change). This PR ships the affirmative form instead.

## Changes

- `platform/daemon/src/pipeline/dreaming-operation-contract.ts` — add the `decline_attention` op (payload: `attentionId`) to the Dreaming vocabulary.
- `platform/daemon/src/pipeline/dreaming-operations.ts` — apply the op in the daemon-owned batch seam: scoped to the agent named in the apply call, one-use (a flag is consumed once, like an archive op), and only resolves rows still pending (`resolved_at IS NULL`), so a record re-enqueued with changed details stays open.
- `platform/daemon/src/pipeline/dreaming.ts` — both runbooks (combined + hygiene) instruct the agent to close records it judged to keep with `decline_attention` and to defer, not decline, records it could not complete; verification contract updated.
- `platform/daemon/src/pipeline/dreaming-capabilities.ts` — apply tool description mentions the new op.
- `docs/api/knowledge-ontology.md` — attention lifecycle documented (resolve on apply or explicit decline; deferred and failed-pass records stay pending).

## Type

- [x] `fix` — bug fix

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-side fix, no UI.

## PR Readiness (MANDATORY)

- [x] Agent scoping verified on all new/changed data queries (decline resolves only rows owned by the agent named in the apply call)
- [x] Input/config validation and bounds checks added (`attentionId` required, schema-enforced)
- [x] Error handling and fallback paths tested (no silent swallow) — not-pending and cross-scope declines fail closed with item-level errors

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes — dreaming + dreaming-agent-tools + dreaming-worker + operation-contract suites 56/56, including: the #1185 primary case (decline closes the flag), the over-resolution regression (list 3, decline 1, defer 2 → exactly 1 resolves, 2 stay pending), and the one-use/scoped invariants (cross-scope decline fails closed, second decline of the same record rejected)
- [x] `bun run typecheck` passes (`@signet/daemon` exit 0)
- [x] `bun run lint` passes (biome clean on touched files)
- [ ] Tested against running daemon — live verification follows the release per the dev process
- [x] N/A

## AI disclosure

## Notes

Resolves #1185

### DIFF
--- a/docs/api/knowledge-ontology.md
+++ b/docs/api/knowledge-ontology.md
@@ -430,8 +430,11 @@ semantic operation. Oversized immutable evidence is instead resumed at a safe
 boundary across passes and is not quarantined.
 
 An attention item is selected with the next scoped pass and rendered as
-non-evidentiary context. It is resolved only after that pass completes; a
-failed pass leaves it pending. The worker can run for pending attention even
+non-evidentiary context. It resolves when the pass applies a hygiene operation
+citing it, or when the agent explicitly declines it (the `decline_attention`
+operation, used when the agent inspects the target and judges it should stay as
+is). Records the agent could not complete stay pending, and a failed pass
+leaves everything pending. The worker can run for pending attention even
 when no new episodic evidence has arrived, while normal failure backoff still
 applies.
 
--- a/platform/daemon/src/pipeline/dreaming-agent-tools.test.ts
+++ b/platform/daemon/src/pipeline/dreaming-agent-tools.test.ts
@@ -517,6 +517,114 @@ describe("dreaming-agent-tools", () => {
 		expect(items[0]).toMatchObject({ subjectRef: "memory:mem-expired", details: { attributeId: "a-trip-attribute" } });
 	});
 
+	it("decline_attention resolves a pending flag and is one-use and scoped", async () => {
+		insertEntity("e-husk", "Legacy Husk", "legacy husk", "owner");
+		const tools = createDreamingAgentTools({
+			accessor: getDbAccessor(),
+			agentId: "owner",
+			actor: "owner",
+			passId: "pass-1",
+		});
+		const minted = readResult(
+			await findTool(tools, "apply_ontology_ops").execute(
+				"call",
+				{
+					agentId: "owner",
+					operations: [
+						{
+							operation: "flag",
+							payload: {
+								subjectRef: "entity:e-husk",
+								details: { entityId: "e-husk", reason: "zero_active_attributes" },
+							},
+						},
+					],
+				},
+				undefined,
+				undefined,
+				{} as never,
+			),
+		);
+		expect(minted.ok).toBe(true);
+		const mintedItems = minted.items as Array<{ result?: { attentionId?: string } }>;
+		const attentionId = mintedItems[0]?.result?.attentionId;
+		if (attentionId === undefined) throw new Error("flag op did not surface an attention id");
+
+		// Declining from another scope fails closed: the record is untouched.
+		const otherScope = createDreamingAgentTools({
+			accessor: getDbAccessor(),
+			agentId: "intruder",
+			actor: "intruder",
+			passId: "pass-1",
+		});
+		const crossScope = readResult(
+			await findTool(otherScope, "apply_ontology_ops").execute(
+				"call",
+				{
+					agentId: "intruder",
+					operations: [{ operation: "decline_attention", payload: { attentionId } }],
+				},
+				undefined,
+				undefined,
+				{} as never,
+			),
+		);
+		expect((crossScope.items as Array<{ ok: boolean }>)[0]?.ok).toBe(false);
+		const stillPending = readResult(
+			await findTool(tools, "attention_list").execute(
+				"call",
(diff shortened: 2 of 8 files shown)


NicholaiVogel:
## Adversarial review — ship-with-fixes, findings fixed

Background subagent review of the original revision (pass-end sweep of everything `attention_list` surfaced) found one HIGH defect, proven with a live scratch experiment:

**HIGH — listed == resolved over-resolution.** Every id from any successful `attention_list` call resolved at pass end, including flags the agent explicitly deferred. Reproduction: 3 flags listed, 1 applied, summary says "deferring the rest" → all 3 resolved. The runbook sanctions deferral ("remaining records are explicitly deferred with reasons in the pass log"), and the enqueue upsert reopens a record only when its details change, so a resolved keep never resurfaced — silently dropped hygiene work.

The sweep was removed and replaced with an affirmative per-record judgment: `decline_attention` in `apply_ontology_ops` closes exactly one pending record the agent inspected and judged to keep. Deferred records stay pending; the #1185 primary case (deliberate keep) still resolves and stops re-triggering passes.

Verified: dreaming + agent-tools + worker + contract suites 56/56 (including the over-resolution regression: list 3, decline 1, defer 2 → exactly 1 resolves), daemon typecheck exit 0, biome clean on touched files. Guards re-verified: generation + resolved_at checks make the op one-use and idempotent; cross-scope declines fail closed (tested).

NicholaiVogel (approved):
Approving after adversarial review (background subagent, live experiments) and full rework of the original approach.

Original revision swept every record surfaced via attention_list at pass end; the review proved with a live scratch experiment that this resolved flags the agent explicitly deferred (list 3, apply 1, defer 2 → all 3 resolved), silently dropping runbook-sanctioned deferred work. Reworked to an affirmative per-record judgment: `decline_attention` op in apply_ontology_ops closes exactly one pending record the agent inspected and judged to keep. Deferred records stay pending; the #1185 keep case still resolves and stops re-triggering passes.

Verification: dreaming + agent-tools + worker + contract suites 56/56 (includes the over-resolution regression and one-use/scoped invariants), daemon typecheck exit 0, biome clean, full CI green (13/13). Branch rebased on main, single commit.

PR #1189, opened 2026-08-07
test(daemon): pin fair requeue budget and both-table dry-run totals (#1052)

### LINKED ISSUE #1052: bug(repair): default requeue starves summary jobs behind a memory backlog
## Root cause

The default `requeue` path has one shared `maxBatch` budget, but always selects memory jobs first.

In `platform/daemon/src/repair-actions.ts:246-255`, it selects up to `maxBatch` memory jobs and gives summary jobs only the remainder. With the default batch size of 50 and at least 50 dead memory jobs, the summary budget is zero.

`buildDeadRequeueSql()` then returns `totalMatching: 0` immediately for a zero limit (`repair-actions.ts:2388-2390`). That means the dry-run total also lies by omitting the summary backlog.

The apply path has the same priority order (`repair-actions.ts:286-290`), so repeated default repairs make no progress on summary jobs until the memory backlog falls below 50.

## Live reproduction

The v0.156.4 release daemon had 10,952 dead memory jobs and 275 dead summary jobs:

```bash
signet repair queue requeue
signet repair queue requeue --tables summary
```

Actual output:

```text
# default (advertised as both)
dry-run: 10952 job(s) match requeue filter; preview shows 50

# explicit summary
dry-run: 275 job(s) match requeue filter; preview shows 50
```

The default should report 11,227 matching jobs. More importantly, `--apply` would consume its full 50-job batch on `memory_jobs`, leaving `summary_jobs` dead after every invocation.

## Expected

The default both-queue repair must make bounded progress in both selected queues and report totals independently of the preview/update cap.

For example, with 100 dead memory jobs, 1 dead summary job, and `maxBatch: 50`:

- dry-run total is `101`, not `100`;
- preview includes the summary job;
- apply requeues the summary job plus no more than 49 memory jobs.

If the desired contract is instead “one queue at a time”, make that explicit and require `--tables`; the current `default: both` contract cannot silently starve one queue.

## Minimal fix

1. Compute `totalMatching` without a selection limit for each selected table.
2. Allocate the bounded update budget fairly across selected non-empty queues (for example, reserve one slot per queue, then use the remaining capacity), and refill unused capacity from the other queue.
3. Keep `--tables memory` and `--tables summary` as exact single-queue selections.

## Regression coverage

Add tests to `platform/daemon/src/routes/queue-diagnostics.test.ts` for a database with more dead memory rows than `maxBatch` and at least one dead summary row:

1. default dry-run reports the sum of both table counts and previews both kinds;
2. default apply touches at least one job in each table and never exceeds `maxBatch` total;
3. `--tables memory` and `--tables summary` preserve exclusive behavior;
4. a zero selection budget does not turn a table's matching count into zero.

aaf2tbz:
## Summary

The default `requeue` used to allocate its shared `maxBatch` budget memory-first, giving the summary queue only the remainder — with a 50-dead-memory backlog the summary budget was zero, so repeated default repairs never touched summary jobs, and the dry-run `totalMatching` hid the summary backlog entirely (it returned 0 for a zero limit).

The fair allocation (a reserved slot per selected non-empty queue, remaining capacity split, unused capacity refilled from the other queue) and the limit-free `totalMatching` are already in main — but nothing pinned the starvation contract. A regression to memory-first selection would silently starve summary again with zero CI signal. This PR adds the exact regression coverage requested in the issue.

## Changes

- `platform/daemon/src/routes/queue-diagnostics.test.ts` — starvation fixture (a dead memory backlog exceeding the default maxBatch + a dead summary backlog) + four tests:
  1. default dry-run reports both table counts (61 memory + 4 summary) and previews a row from each table;
  2. default apply requeues at least one job per table and never exceeds `maxBatch`;
  3. `--tables memory` / `--tables summary` keep exclusive single-queue selections;
  4. a zero selection budget (`maxBatch: 1`) does not zero a table's matching count in the dry-run total.

## Type

- [x] `fix` — bug fix

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-side queue repair behavior, no UI.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes — `queue-diagnostics.test.ts` 20/20 (4 new)
- [x] `bun run lint` passes
- [ ] N/A

## AI disclosure

## Notes

Resolves #1052

### DIFF
--- a/platform/daemon/src/routes/queue-diagnostics.test.ts
+++ b/platform/daemon/src/routes/queue-diagnostics.test.ts
@@ -250,6 +250,152 @@ describe("repair action integration via the new dispatch path", () => {
 		}
 	});
 
+	/**
+	 * A memory backlog that alone exceeds the default maxBatch (50) plus a small
+	 * dead summary backlog — the #1052 starvation shape: the memory-first
+	 * selection used to give summary a zero budget and the dry-run total hid the
+	 * summary backlog entirely.
+	 */
+	function seedStarvationFixture(db: Database): void {
+		const oldIso = new Date(Date.now() - 45 * 24 * 60 * 60 * 1000).toISOString();
+		for (let i = 0; i < 60; i++) {
+			db.prepare(
+				`INSERT INTO memory_jobs (id, memory_id, job_type, status, attempts, max_attempts, created_at, updated_at)
+			 VALUES (?, ?, ?, ?, ?, ?, ?, ?)`,
+			).run(`starv-mem-${i}`, `starv-memory-${i}`, "document_ingest", "dead", 3, 3, oldIso, oldIso);
+		}
+		for (const [id, session] of [
+			["starv-sum-1", "session-S1"],
+			["starv-sum-2", "session-S2"],
+		] as const) {
+			db.prepare(
+				`INSERT INTO summary_jobs (id, session_key, harness, project, transcript, status,
+				attempts, max_attempts, created_at, error)
+			 VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+			).run(id, session, "codex", "demo", "transcript", "dead", 3, 3, oldIso, id === "starv-sum-1" ? "boom" : null);
+		}
+	}
+
+	it("default requeue dry-run reports both table counts and previews both kinds (#1052)", () => {
+		const env = setup();
+		try {
+			seedStarvationFixture(env.db);
+			const result = requeueDeadJobs(
+				env.accessor,
+				{
+					autonomous: { enabled: true, frozen: false },
+					repair: { requeueCooldownMs: 0, requeueHourlyBudget: 100 },
+				} as unknown as Parameters<typeof requeueDeadJobs>[1],
+				{ reason: "test", actor: "test", actorType: "operator" },
+				createRateLimiter(),
+				{ dryRun: true },
+			);
+			// 61 dead memory (1 base + 60 seeded) + 4 dead summary (2 base + 2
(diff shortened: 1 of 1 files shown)


NicholaiVogel (approved):
Verified the regression coverage actually pins the #1052 contract, rather than trusting green CI.

**Fixture math checks out.** Base setup seeds 1 dead memory + 2 dead summary; the starvation fixture adds 60 + 2 -> 61 memory / 4 summary dead, totalMatching 65. The budget math under the fair allocator (maxBatch 50, reserved 2, split 24/24, refill 21 from summary's unused share) yields a 46/4 split — inside the [2, 50] affected range and >= 1 per table.

**Mutation-verified.** Reverted `allocateRequeueBudgets` to the pre-#1052 memory-first allocation and the dry-run zero-budget fallback to the old selection-limited totalMatching. The suite drops to 14 pass / 6 fail with every failure on the starvation contract: both-table dry-run total (65 expected, 61 received), per-table apply minimums (requeuedSummary >= 1 unmet), and the zero-budget count (65 expected, 61 received). Restored the fix: 20 pass / 0 fail, 74 expect calls, test count stable across both runs.

**Rebase clean.** Branch rebased onto current main (0.176.1); `repair-actions.ts` is byte-identical between PR branch and main, tests pass 20/20 on the rebased commit. Test-only PR, no production code, no data writes.

Issue #1213, opened 2026-08-08
docs: add TELEMETRY.md as the single telemetry reference


NicholaiVogel:
## Goal

Signet's telemetry is documented across three places (CONFIGURATION.md telemetry section, ANALYTICS.md PostHog section, PIPELINE.md config sample) with no single source of truth. Add `docs/TELEMETRY.md` consolidating the full picture, linked from CONFIGURATION.md and the docs index.

## Proposed outline

1. What telemetry is and why (install/usage analytics, crash diagnostics, per-install economics).
2. The event catalog: one table per event with payload fields:
   - `install.ping`, `install.activated`, `daemon.started`, `daemon.heartbeat`
   - `session.start` / `session.end` (note: `session_end` is the per-turn message-save hook — see #1212; rename pending)
   - `llm.generate`, `pipeline.embedding`, `dreaming.pass`
   - `error.occurred` (sanitized crash report + `EventLoopLag`), `inference.*`, `version.upgraded`
   - `command.invoked` (JSONL-only)
3. The privacy contract: anonymous per-install UUIDs, no memory content/query text/identity, sanitized crash reports (truncated message, home-stripped stack), geo city/country only. Note the open question: PostHog captures `$ip` server-side (see #1200 thread).
4. Configuration: `telemetryEnabled` (default true), `posthogHost`, `posthogApiKey`, `SIGNET_TELEMETRY_OPTOUT`, secret override path.
5. The open JSONL audit log (`<agentsDir>/.daemon/telemetry/events.jsonl`) and CLI `command.invoked`.
6. Known semantics quirks: session.end = turns, embeddings = tokens-only (cost pending #1201), CI/dev-fleet tagging (#1200).
7. How to query: PostHog project 547297, HogQL examples (active installs, per-install cost), the vault analytics sync pointer.
8. Changelog of event additions per release.

## Acceptance

- `docs/TELEMETRY.md` exists, accurate against current code, and the telemetry sections in CONFIGURATION.md/ANALYTICS.md link to it instead of duplicating.
- Doc-drift check passes (`bun scripts/doc-drift.ts`).

## Notes

- Related: #1026, #1200, #1201-#1207, #1212.

Issue #1200, opened 2026-08-08
telemetry: dev checkouts pollute install/user metrics — silence or tag operator fleet


NicholaiVogel:
## Problem

Default-on telemetry (from #1169) means **every daemon that boots with default config becomes a "user" in the Signet PostHog project** — including the operator's own development fleet. Observed in project 547297:

- 4 daemons alive simultaneously; 3 of them are operator-owned dev/workspace daemons (stale version strings like `0.147.1` from dev-checkout builds, idle daemons parked 17h+, the deepseek-v4-flash delegation workload), and only 1 is clearly external.
- Each dev workspace has its own persisted install id (`telemetry_install`), so every checkout/workspace the operator or the swarm runs inflates `install.activated`, `daemon.started`, and heartbeat counts.
- Same failure class as the CI pollution fixed in #1179 (which used `SIGNET_TELEMETRY_OPTOUT` in workflows) — dev checkouts are the next layer up.

Consequence: "active installs" and "users" cannot be trusted until operator-owned installs are either silenced or identifiable.

## Options

1. **Silence dev checkouts**: dev bases and worktrees set `SIGNET_TELEMETRY_OPTOUT=1` (env in dev startup paths, `.env`/wrapper scripts). Simple, but loses operator's own usage data and must be maintained per environment.
2. **Tag dev installs**: add an opt-in marker the daemon can read (env `SIGNET_TELEMETRY_ENV=dev` or a config key) that attaches a `deployment: dev` property to every event. Keeps the operator's data, makes it filterable (`where properties.deployment != 'dev'`). Slightly more surface area.
3. **Version sanity**: dev builds report stale `CURRENT_VERSION` (observed `0.147.1` emitting current-code events), which also confuses version breakdowns. Consider appending a dev marker to the version string in dev builds (e.g. `0.147.1-dev`) so dev installs are identifiable from the version alone.

Recommendation: option 2 (tag) with option 3 as a cheap complement — the data stays, the noise becomes filterable.

## Related

- #1179 (CI opt-out) — same class, fixed at the workflow layer
- #1026 (telemetry)
- Also unresolved: PostHog captures `$ip` + `$geoip` on every event server-side (observed `$ip` in event properties), which conflicts with the "no IP logging" claim in the install ping docs — separate issue to file or fold in here

Issue #1026, opened 2026-07-25
PostHog and non-intrusive analytics for understanding installation and usage patterns


NicholaiVogel:
## Motivation

We have very little visibility into how Signet is actually being used. npm download counts (28K/mo) are a noisy signal — they include CI pipelines, npx one-offs, version-update churn, and repeated installs from rapid publishing (710 versions in 5 months). We cannot distinguish between a real developer installing once, a CI pipeline re-installing on every commit, an npx eval that never uses it again, or an existing user whose npm update pulled a new version.

GitHub release download counts give us a slightly cleaner binary-download signal, but we have no way to connect installs to usage, retention, or feature adoption.

## Goals

1. Unique-install count — how many distinct environments have installed Signet (vs raw npm download volume)
2. Version adoption — what version people are actually running (to know if updates land)
3. Feature usage (optional, opt-in) — which commands/tools are used
4. Crash/error reporting (optional, opt-in) — unhandled errors to prioritize stability work

### Phase 1: Post-install ping (fully anonymous, no identifier)

Add a lightweight HTTP request in install-native.js that fires once per install. No unique identifier, no IP logging, just a counter increment:

GET https://telemetry.signetai.sh/ping?v={version}&p={platform}

This gives us a unique-install-per-version count that directly measures how many times the postinstall script executed. The endpoint can be a Cloudflare Worker with a D1 counter.

### Phase 2: Opt-in telemetry daemon event

The daemon already has a PostHog config in memory-config.ts (telemetryEnabled: false, posthogHost, posthogApiKey). Enable a prompt on first run that asks the user to opt into anonymous usage data.

If the user opts in, fire events like:
- daemon_started (version, platform, uptime)
- command_invoked (command name, no arguments)
- error_occurred (error type, no stack/message content)
- version_upgraded (from version, to version)

### Phase 3: GitHub release download dashboard

GitHub already tracks release asset download counts per-release. We publish native binaries and connector tarballs as release assets. The release download count is a cleaner signal than npm because each download is a real binary fetch. We should surface these counts in a dashboard.

## Constraints

- No PII, no code, no memory content. Telemetry payloads must be inspectable and trivially auditable.
- Opt-in by default. Disabled until the user explicitly enables it.
- Local-first guarantee. Opt-out telemetry does not block any feature.
- Open telemetry log. Users should be able to see exactly what was sent (stored in a local file).
- No third-party dependency for the ping endpoint.

## Prior art

[...]

PR #1220, opened 2026-08-08
fix(telemetry): session.end at real boundaries only (#1212)

### LINKED ISSUE #1212: telemetry: session.end fires per Stop hook, not per session — counters not comparable
## Problem

`session.end` telemetry fires on the `session_end` hook — the hook harnesses call to SAVE MESSAGES into sessions turn by turn. It fires per turn, not per session termination, so the event is a "turns persisted" counter while its name implies session boundaries. `session.start` is dedup'd per real session, but the end path has no such discipline, so the counters are not comparable:

- Observed in the wild: 707 `session.end` vs 78 `session.start` across the fleet (~9:1).
- The operator running 11 concurrent agent sessions produced hundreds of `session.end` events in ~2 hours from one dev install (harnesses: claude-code, oh-my-pi, opencode — all call the hook per turn).
- The vault analytics (Signet/analytics) relabeled the metric as "turns saved" to avoid misleading session counts.

## Expected

- Rename the telemetry event to something honest (`session.turn` or `turn.persisted`) and keep the per-turn cadence — it measures harness activity volume, which is useful.
- Add a separate real `session.end` (or `session.terminated`) fired only at actual session boundaries via the route-level `markSessionEnded`/`releaseSession` path.
- True session/concurrency counts then come from `session.start` (dedup'd) plus the termination event, not from the per-turn hook.
- Optionally add a concurrent-sessions gauge or a hashed session key so true session counts and concurrency are measurable without identity (hashed, never raw).

## Notes

- `session.start` already dedups (session-start-state); the end path needs the same treatment.
- Related: #1026, #1200.

NicholaiVogel:
## Summary

`session.end` telemetry fired on every `session-end` hook call — the hook harnesses call **per turn** to persist messages — so the event was a "turns persisted" counter with a session-boundary name. Observed ~9:1 inflation vs the dedup'd `session.start` (707 vs 78 across the fleet; one dev install produced hundreds in ~2h from 11 concurrent sessions). This splits the semantics so the counters are honest and comparable.

## Changes

- **`session.turn`** — the per-turn event (renamed from `session.end`), keeps per-turn cadence as a harness-activity volume counter. Carries `harness`, `promptCount`, `sessionHash`.
- **`session.end`** — now fires only at real session terminations:
  - explicit `reason: "clear"` in `handleSessionEnd` (caller discards the session),
  - TTL-evicted (abandoned) session claims in `session-tracker.ts` (`evictExpiredSession` — the daemon's existing "this session is over" decision).
- **Once per session lifetime** — new `session-end-state.ts` dedup (mirrors `session-start-state.ts`), cleared on real and clear session starts. Repeated clears or re-claim-evict cycles without a new start cannot inflate the counter.
- **Normalized session identity** — `normalizeSessionKey` (trim + strip `session:` prefix) now lives in `session-end-state.ts`, re-exported by the tracker. The dedup map and `sessionHash` use the same normalized identity as the tracker, so `session:<uuid>` keys (openclaw sends these) can neither double-count a lifetime nor produce unjoinable hashes. Found by the adversarial review, regression-tested.
- **Harness on the claim** — `claimSession`/`claimAutomaticSessionOrSkip` now record the harness, so eviction events break down by harness (previously the tracker only knew plugin/legacy).
- **`sessionHash`** — 16-hex sha256 of the normalized session key on all three session events, so distinct sessions and concurrency are measurable without leaking raw keys (issue's optional item).

No DB schema changes; telemetry events only. `session.start` unchanged (still dedup'd). Cleanly-finished sessions that never send `clear` are not counted as ended — only provable terminations are.

## Type

bug (fix)

## Packages affected

- `@signet/daemon` (platform/daemon)

## Tests

- `hooks.test.ts`: per-turn session-end emits `session.turn` (never `session.end`); explicit clear emits `session.end` once per lifetime, and a real session start opens a new lifetime.
- `session-tracker.test.ts`: TTL eviction emits `session.end` once per lifetime with `reason: "expired"` + harness; harness-less claims emit `harness: null`; a `session:`-prefixed key cannot double-count a lifetime and the hash is joinable across raw/normalized forms.
- Full affected suites: 235 pass / 0 fail; daemon production tsc clean; daemon build clean.

## Docs

- `docs/TELEMETRY.md` — event catalog updated: `session.start` / `session.turn` / `session.end` rows, the "known naming quirk" note replaced with the fixed semantics, changelog row added.
- `signet-telemetry` skill updated to match.

## Related

Fixes #1212. Related: #1026, #1200, #1213 (TELEMETRY.md).
## PR Readiness (MANDATORY)

### DIFF
--- a/docs/TELEMETRY.md
+++ b/docs/TELEMETRY.md
@@ -52,8 +52,9 @@ PostHog failures, and never throws into the daemon.
 | `install.activated` | first daemon run of a new install (persisted install id first created) | `version`, `platform` |
 | `daemon.started` | daemon boot | `version`, `platform`, `uptimeMs` |
 | `daemon.heartbeat` | every 5 minutes | `uptimeMs`, `memoryCount`, `connectorsActive`, `pipelineMode`, `extractionProvider`, `embeddingProvider` |
-| `session.start` | real session start (deduped; stubs and clear/reset paths don't count) | `harness` |
-| `session.end` | Stop/session-end hook | `harness`, `promptCount` |
+| `session.start` | real session start (deduped; stubs and clear/reset paths don't count) | `harness`, `sessionHash` |
+| `session.turn` | every `session-end` hook call (per turn, see notes) | `harness`, `promptCount`, `sessionHash` |
+| `session.end` | real session termination: explicit `reason: "clear"`, or a TTL-evicted (abandoned) session claim | `harness`, `reason` (`clear` / `expired`), `sessionHash` |
 | `llm.generate` | every LLM call | `provider`, `latencyMs`, `success`, `inputTokens`, `outputTokens`, `cacheReadTokens`, `cacheCreationTokens`, `totalCost` |
 | `pipeline.embedding` | every embedding fetch, at the usage-recording boundary | `tokens`, `provider`, `sourceKind` (`memory-capture` / `artifact-index` / `recall` / `dreaming` / `other`) |
 | `dreaming.pass` | completed agentic dreaming pass (early-exit passes emit nothing) | `mode`, `tokensInput`, `tokensOutput`, `tokensCacheRead`, `tokensCacheWrite`, `cost` |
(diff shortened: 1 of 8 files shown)


NicholaiVogel:
**Adversarial review (subagent + follow-up verification)** — couldn't post as an approval (author self-approval blocked), so recording it here:

- Reviewed `git diff origin/main..HEAD` across all 8 files; ran affected suites (235 pass / 0 fail), daemon production tsc, daemon build.
- The reviewer's dedup-key investigation found a REAL defect: the session-end dedup map and `sessionHash` used the raw session key on the hooks paths but the tracker-normalized key on the eviction path. With `session:<uuid>` keys (openclaw sends these; documented in MCP.md) the same lifetime could emit `session.end` twice and produce unjoinable hashes. **Fixed**: `normalizeSessionKey` moved into `session-end-state.ts` (single source of truth), re-exported by the tracker; regression test added ("does not double-count a lifetime whose clear used a session:-prefixed key").
- Verified eviction funnels entirely through `evictExpiredSession` (cleanup, stale-reclaim, lazy eviction in hasSession/getSessionPath/getActiveSessions) — `markSessionEnded`/`releaseSession`/`releaseAllSessions` do not emit, so per-turn releases and daemon shutdown cannot inflate the counter.
- Privacy: all three events carry only `harness`, `promptCount`, `reason`, `sessionHash` (16-hex sha256 of the normalized key); raw session keys never reach telemetry properties.
- Dedup cleared on real and clear session starts only — resumed sessions (dedup stub) keep the marker, so a paused/resumed conversation cannot double-count.

**Verdict: ship.**

PR #1221, opened 2026-08-08
feat(telemetry): first-run activation funnel (first.remember / first.recall)

### LINKED ISSUE #1202: telemetry: first-run activation funnel (first.remember / first.recall)
## Goal

Today the funnel is install.ping -> install.activated, but nothing shows whether an install was actually *used* (first remember / first recall). Without it, "activated" includes installs that never did anything.

## Scope

- Fire `first.remember` and `first.recall` exactly once per install (guarded by the persisted install id, same as install.activated) at the first successful remember/recall.
- No content: event carries nothing beyond the fact (and maybe provider/route).

## Acceptance

- One-shot semantics pinned by a regression test (same pattern as install.activated in #1188).
- Funnel queryable: distinct_ids with activated vs activated+first.remember vs +first.recall.

## Notes

- This is the retention/activation metric for the product.
- Related: #1188, #1200.

NicholaiVogel:
## Summary

Closes #1202. The activation funnel was install.ping -> install.activated, but nothing showed whether an install was actually *used*. This adds `first.remember` / `first.recall`, fired exactly once per install at the first successful remember / recall, so the funnel becomes activated vs activated+first.remember vs +first.recall.

## Changes

- Migration 111: adds `first_remember_at` / `first_recall_at` to `telemetry_install` (pragma-guarded idempotent ADD COLUMN, same pattern as 082).
- `telemetry.ts`: new `first.remember` / `first.recall` event types; `TelemetryCollector.recordFirstUse(kind)` claims the milestone with an atomic guarded UPDATE (`WHERE id = ? AND <column> IS NULL`) — only the winning caller emits the event, so concurrent or repeated calls can't double-fire. Events carry only `{version, platform}` — no content, no query, no agent ids.
- `memory-routes.ts`: fires `recordFirstUse("remember")` on every success path of `POST /api/memory/remember` (plain insert, deduped paths, chunked insert/dedupe) and `recordFirstUse("recall")` on `POST /api/memory/recall` success. CLI, MCP, and the `/api/hook/remember` proxy all route through these handlers, so all surfaces are covered.
- `docs/ANALYTICS.md`: events list updated.

## Type

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) (no spec entry covers telemetry events; migration registered in the canonical migrations index)
- [x] Agent scoping verified on all new/changed data queries (no user-data queries touched; the claim UPDATE is scoped to the anonymous install id, never an agent)
- [x] Input/config validation and bounds checks added (no new external input; kind is a fixed union)
- [x] Security checks applied to admin/mutation endpoints (n/a — telemetry write only)
- [x] Docs updated for API/spec/status changes (`docs/ANALYTICS.md`)
- [x] Regression tests added for each bug fix (one-shot regression test pins the pattern)

- [x] Migration is idempotent (pragma-guarded ADD COLUMN; fresh + re-run covered by migrations.test.ts, which also verifies declared artifacts)
- [x] Rollback / compatibility note included in PR description (columns are nullable TEXT; rollback = drop migration 111 registration — old code ignores the columns)

## Testing

- [x] `bun test` passes (144/144 across telemetry, migrations, memory-routes, mutation-api, analytics suites)
- [x] `bun run lint` passes (only 3 pre-existing errors on main in these files, none introduced here)
- [x] Tested against running daemon (scratch daemon on port 3862: 2 remembers + 2 recalls produced exactly 1 `first.remember` and 1 `first.recall` in `telemetry_events`, both `sent_to_posthog=1`, properties `{version, platform}` only)
- [ ] N/A

## AI disclosure

## Notes

- Error handling: claim failure leaves the milestone unclaimed (may fire on a later run, still exactly once); the collector is best-effort and never throws — covered by the one-shot regression test.
- One anonymous scratch install (fresh uuid) was created by the live verification run against the real PostHog project — same class as #1200's dev-fleet pollution, one row, no PII.

### DIFF
--- a/docs/ANALYTICS.md
+++ b/docs/ANALYTICS.md
@@ -209,14 +209,41 @@ PostHog Telemetry
 
 Separate from the in-memory analytics above, the daemon ships an
 anonymous telemetry collector for understanding install and usage
-patterns (see [configuration](./CONFIGURATION.md#telemetry-telemetry)). It
-buffers events to SQLite and flushes them in batches to a PostHog instance
-when both `posthogHost` and `posthogApiKey` are configured
-(`telemetryEnabled` defaults to true; set it to `false` to opt out).
+patterns (see [[configuration]]). It buffers events to SQLite and
+flushes them in batches to a PostHog instance when both `posthogHost`
+and `posthogApiKey` are configured (`telemetryEnabled` defaults to
+true; set it to `false` to opt out).
 
-The event catalog, privacy contract, the open JSONL audit log, and how to
-query the project live in **[TELEMETRY.md](./TELEMETRY.md)** — the single
-telemetry reference.
+Privacy contract:
+
+- Events carry no memory content, user identity, agent ids, or file paths.
+- Each install gets a random anonymous id (persisted in the workspace
+  database) used as the PostHog `distinct_id`, so installs are countable
+  but not identifiable.
+- Telemetry is on by default and can be disabled with
+  `telemetryEnabled: false`.
+- Every recorded event is mirrored to an open, inspectable JSONL log at
+  `<agentsDir>/.daemon/telemetry/events.jsonl`, so users can audit
+  exactly what was sent.
+
+Events recorded: `daemon.heartbeat` (every 5 minutes), the `inference.*`
+lifecycle (`route`, `execute`, `stream`, `fallback`), `llm.generate`
+(provider, latency, token and cost counts when reported, success — never
+prompt text), `session.start` / `session.end` (harness and prompt count),
+the lifecycle events `daemon.started`, `command.invoked`,
+`error.occurred` (sanitized crash report — truncated message with user
+paths stripped, top stack frames with home directories removed, uptime;
+plus rate-limited `EventLoopLag` reports with measured lag), and
+`version.upgraded`, `install.activated` (first daemon run of a new
+install — covers bun/desktop installs the npm postinstall ping never
+sees), `first.remember` / `first.recall` (first successful remember /
+recall per install, guarded by the persisted install id — the
+activation funnel: an install that was actually used, not just
+booted), `pipeline.embedding` (tokens, provider, sourceKind — memory
+capture / artifact index / recall / dreaming), and `dreaming.pass`
+(provider-reported input/output/cache tokens and cost per agentic
+pass). The remaining `pipeline.*` event types are declared
+for future use but not yet emitted.
 
 Release Download Stats
 ---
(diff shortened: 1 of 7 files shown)


NicholaiVogel:
Adversarial review — issue #1202 one-shot first-use telemetry.

Verdict: SHIP. No HIGH/CRITICAL/MEDIUM defects found. Every claimed behavior was proven with live experiments against the real collector, DB, and daemon — not code reading.

## What was tested (live)

1. Claim atomicity: `UPDATE telemetry_install SET first_remember_at = ? WHERE id = ? AND first_remember_at IS NULL` returns changes 1 on first claim, 0 on repeat, 0 on missing id (bun:sqlite).
2. Cross-process race: two concurrent collector processes claiming the same milestone over one DB produced exactly 1 `first.remember` and 1 `first.recall` event and one non-null timestamp — the atomic guard holds.
3. Dual-DB parity: same claim shape verified under better-sqlite3 (changes 1/0/0) — the `changes > 0` check works on both runtimes the daemon supports.
4. Migration backward-compat: a DB created with migration 109's old schema (id, created_at only) upgrades cleanly through 111 — columns added, existing install row preserved, no throw; pre-existing columns tolerated (idempotent).
5. JSONL audit log: `first.remember` / `first.recall` mirror to the open telemetry log like every other event.
6. End-to-end daemon (scratch, port 18799): fresh install → `install.activated`; first remember (including the deduped path) → 1× `first.remember`; first recall → 1× `first.recall`; second remember + second recall → no re-fire; both milestones persisted on the install row; events reached PostHog (`sent_to_posthog=1`). Properties are exactly `{version, platform}` — no content, no query, no agent ids.
7. Suites: telemetry.test.ts 15/15 (incl. the new one-shot regression), migrations.test.ts 51/51 (post-DDL artifact verification covers migration 111), memory-routes + mutation-api + analytics 129/129. Full `bun run typecheck` green.

## Findings

- LOW — crash window between claim and emit (telemetry.ts `recordFirstUse`): a process killed at exactly the instruction boundary between the claim UPDATE (committed) and the in-memory `collector.record()` permanently consumes the milestone without emitting the event — no later remember/recall will fire it. Reproduced deliberately by claiming then exiting before flush. Accepted limitation: microseconds-wide, best-effort telemetry, and any fix (transactional event row) would restructure the buffer/flush model for negligible funnel accuracy gain.

No other defects found. The route wiring (all 6 remember success paths + recall) was enumerated and the deduped and chunked paths exercised live.

NicholaiVogel:
Temporarily closing to force CI re-trigger after PR body update

PR #1239, opened 2026-08-08
fix(daemon): preserve legacy routing migration

### LINKED ISSUE #1229: v0.176.11 migration silently drops legacy LLM routing and leaves background workers retrying forever
## Bug Description

On upgrade to **v0.176.11**, the daemon migrated a workspace that had legacy pipeline routing into an empty `inference` registry. The migration log says it removed the legacy extraction/synthesis provider and model fields, but it created no `inference.targets`, `accounts`, `workloads`, or bindings and reported no migration problem.

The resulting daemon repeatedly runs LLM-backed background work anyway. Hints and reflections fail with `Routing is not enabled for this agent config.`, jobs reach retry limit, the composite health becomes unhealthy, and dreaming defers its sweep because the queues are under pressure.

This contradicts the cutover contract in #947: anything that cannot be mapped should be reported and nothing should be silently dropped.

## Steps to Reproduce

1. Start with a workspace configured using legacy `memory.pipelineV2.extraction.provider` / `model` (and/or synthesis equivalents).
2. Upgrade/start Signet v0.176.11.
3. Inspect the migration log and generated `agent.yaml`.
4. Run `signet inference list --json` and observe an empty, disabled registry.
5. Observe daemon logs/status while hints/reflections/dreaming workers run.

## Actual Behavior

Migration logged:

```text
Migrated legacy routing fields to inference registry
mutations: removed memory.pipelineV2.extraction.provider
           removed memory.pipelineV2.extraction.model
           removed memory.pipelineV2.synthesis.provider
           removed memory.pipelineV2.synthesis.model
note: Routing now flows through inference.*
```

But the resulting config contained only:

```yaml
inference:
  targets: {}
  accounts: {}
  workloads: {}
```

`signet inference list --json` returned `enabled: false` with no targets or workload bindings.

The daemon then continuously logged:

```text
Hints worker job failed: Routing is not enabled for this agent config.
Generation failed: Routing is not enabled for this agent config.
Deferring dreaming sweep while queues are under pressure
```

`signet status` reported degraded readiness; its current queue view included 20 dead memory jobs and 1,751 dead summary jobs. The older summary rows predate this upgrade, but newly generated hint/reflection jobs are immediately failing under the empty registry.

## Expected Behavior

The one-time migration should do one of the following:

1. Convert a valid legacy provider/model/endpoint into a usable canonical inference target and bind the applicable workloads; or
2. Preserve the legacy values and surface an explicit actionable migration error; or
3. If no route can be configured safely, disable route-dependent background workloads (hints/reflections/dreaming) so they do not continuously fail and poison health/queues.

In all cases, `signet inference doctor` should identify the missing workload route rather than reporting an otherwise clean disabled registry.

## Environment

- Signet: 0.176.11 (`signetai-linux-x64`)
- OS: Linux 6.17.0-23-generic
- Workspace: existing production workspace, daemon started immediately after upgrade
- Embeddings: local Ollama works directly (`nomic-embed-text`); this issue concerns LLM routing after the inference-registry cutover.

Related: #947, #1004, #899.

NicholaiVogel:
## Summary

Fixes #1229. Legacy flat extraction routing fields were deleted during migration without being compiled into the inference registry, leaving background extraction with no workload target.

## Changes

- Compile `memory.pipelineV2.extractionProvider`, `extractionModel`, and endpoint fields into `inference.targets.legacy-extraction` and `inference.workloads.memoryExtraction` before cleanup.
- Map legacy folded harness providers to ACPX targets with the correct agent.
- Preserve unsupported legacy providers when no canonical extraction route exists, and emit an actionable warning instead of silently deleting configuration.
- Add a regression test covering the destructive flat-key migration path.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: none

## Screenshots

N/A: daemon migration only.

## PR Readiness (MANDATORY)

- [ ] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [ ] Agent scoping verified on all new/changed data queries
- [ ] Docs updated for API/spec/status changes

- [x] Migration is idempotent

The migration remains compatible with already-canonical v4 configurations. Unsupported legacy providers are preserved for manual reconfiguration rather than deleted. A workspace already migrated by the affected release may require restoring the original legacy provider/model values before rerunning the migration; this fix prevents the destructive loss on future upgrades.

## Testing

- [x] `bun test platform/daemon/src/config-migration.test.ts platform/daemon/src/legacy-routing-migration.test.ts platform/daemon/src/inference-router.test.ts`
- [x] `bun run --filter '@signet/daemon' typecheck`
- [x] `bun run --filter '@signet/daemon' build`
- [x] `bun run typecheck`
- [ ] `bun run lint`
- [ ] N/A

## AI disclosure

## Notes

The daemon package checks pass. Root typecheck currently fails in unrelated connector packages due workspace build artifacts/API drift (for example missing exports from `@signet/connector-base`), and root lint reports pre-existing repository-wide package.json formatting diagnostics; no unrelated files were changed.

### DIFF
--- a/platform/daemon/src/config-migration.ts
+++ b/platform/daemon/src/config-migration.ts
@@ -240,13 +240,43 @@ function writeAtomic(path: string, contents: string): void {
 // rerun for v4 configs because the incomplete migration shipped to users.
 // Idempotent: a file already at v5+ is skipped.
 
-const LEGACY_FLAT_KEYS = ["extractionProvider", "extractionModel", "extractionStrength"] as const;
+const LEGACY_FLAT_KEYS = [
+	"extractionProvider",
+	"extractionModel",
+	"extractionEndpoint",
+	"extractionBaseUrl",
+	"extractionFallbackProvider",
+	"extractionStrength",
+] as const;
+
+const LEGACY_FLAT_ROUTING_KEYS = [
+	"extractionProvider",
+	"extractionModel",
+	"extractionEndpoint",
+	"extractionBaseUrl",
+	"extractionFallbackProvider",
+] as const;
+
+const LEGACY_HARNESS_AGENT: Readonly<Record<string, string>> = {
+	"claude-code": "claude",
+	codex: "codex",
+	opencode: "opencode",
+};
 
 /** Map a legacy provider to the account name/id this migration creates. */
 function legacyAccountFor(provider: string): { name: string; family: string; cred: string } | null {
 	if (provider === "openrouter") return { name: "legacy-openrouter", family: "openrouter", cred: "OPENROUTER_API_KEY" };
 	if (provider === "anthropic") return { name: "legacy-anthropic", family: "anthropic", cred: "ANTHROPIC_API_KEY" };
 	return null; // local providers (ollama/llama-cpp/openai-compatible-local) need no account
+}
+
+function legacyExecutorFor(provider: string): { executor: string; acpxAgent?: string } | null {
+	const acpxAgent = LEGACY_HARNESS_AGENT[provider];
+	if (acpxAgent) return { executor: "acpx", acpxAgent };
+	if (["acpx", "anthropic", "openrouter", "ollama", "llama-cpp", "openai-compatible"].includes(provider)) {
+		return { executor: provider };
+	}
+	return null;
 }
 
 export function migrateLegacyRoutingToRegistry(agentsDir: string): void {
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Checklist exception: the focused daemon checks and full repository typecheck pass. The root lint command reports 546 pre-existing repository-wide formatting diagnostics across unrelated package.json files; this PR changes only the daemon migration and test files, which pass scoped Biome checks. The exception label is intentional rather than falsely checking the local full-lint item.

NicholaiVogel:
Adversarial review follow-up: fixed the HIGH route-overwrite finding and both MEDIUM data-loss findings. The migration now snapshots pre-existing canonical extraction routes before compiling legacy fields, falls back to a valid nested route when flat routing is unsupported while preserving the unsupported flat values, and migrates nested base_url/baseUrl endpoints. Direct legacy acpx is also mapped. Expanded focused coverage now passes 32/32.

PR #1244, opened 2026-08-08
fix(hermes): recover claims without reinjection

### LINKED ISSUE #1243: daemon restart re-injects full memory context into active sessions mid-conversation - claim-only recovery needed
When Signet's daemon restarts, every active session gets thousands of tokens of memory context injected into its context window mid-session - even though each session already had that context appended to its system prompt at session start.

Mechanism (integrations/hermes-agent/connector/hermes-plugin/__init__.py):
- Line 534: on daemon-restart detection ("Signet daemon restarted mid-session, re-initializing"), the connector calls client.session_start(...) again - the full session-start hook, which returns the complete identity + memories + inject.
- The re-init inject is then written into the session's context (the prefetch/inject path) - mid-conversation, not at start.
- Every active session does this simultaneously on a restart (the boot-storm: 666 session-start hooks vs 426 session claims today; each daemon restart triggers a fresh wave).

Impact:
1. Redundant token cost: thousands of tokens × active sessions re-injected on every restart. On a 7-14 session install with frequent restarts (crash/update cycles), this is the dominant wasted-token source.
2. Prompt-cache invalidation: mutating past context mid-conversation breaks per-conversation prompt caching (the Hermes core invariant) - every re-init forces a full re-encode.
3. Boot-storm load: all sessions re-calling session-start simultaneously during daemon boot contributes to the event-loop saturation/wedge pattern seen on multi-session installs (see #1114 for the related session-start cost).

Expected:
- The restart-recovery path should re-establish only the session claim (so the daemon recognizes the session again), NOT re-inject the memory context. The session already has its context from the actual session start.
- Add a re-init mode to session-start (or a separate endpoint) that returns no inject/memories - just sessionKnown + claim - used only by the restart-recovery path.
- The user_prompt_submit sessionKnown flag (line 522) is the right detection; the recovery should be claim-only.

Related: #1114 (session-start cost), #1228 (session lifecycle state is process-ephemeral - restart survival)

NicholaiVogel:
## Summary

Fix #1243: daemon restart recovery now reclaims the active Hermes session without re-running session-start recall or injecting a second system-context block into the conversation.

## Changes

- Add an explicit `claimOnly: true` session-start route mode that preserves runtime-path ownership but returns only `{ "sessionKnown": true }`.
- Reject claim-only recovery without a `plugin` or `legacy` runtime path, so a successful recovery always renews an actual claim.
- Update the Hermes memory provider to use recovery mode after `sessionKnown: false`, discard every response payload, and avoid queuing an active-session injection.
- Add daemon route, Hermes client/provider, API, and hook documentation coverage for the recovery contract.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [x] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: API and hook documentation

## Screenshots

Not applicable: no UI changes.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

No migration. Older daemons that ignore `claimOnly` remain safe because the Hermes provider deliberately discards the response and never queues a recovery injection.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Focused proof:

- `bun test platform/daemon/src/hooks-recall.test.ts`: 24 pass, 0 fail.
- `bun test integrations/hermes-agent/connector/index.test.ts`: 74 pass, 0 fail.
- Built `@signet/core`, `@signet/daemon`, `@signet/connector-base`, and `@signet/connector-hermes-agent`.
- Full workspace typecheck passed after generating the repository's existing embedded plugin assets.
- Isolated daemon proof: `claimOnly: true` returned `200 {"sessionKnown":true}`; a conflicting legacy runtime path then received `409`.

The full daemon suite was attempted but is not green in this worktree: 2,207 pass, 87 fail, and 21 errors. The failures include `DbAccessor is closed` background-worker errors and unrelated maintenance-worker assertions. Root lint also reports 543 existing diagnostics, predominantly repository-wide formatting drift. I did not mark either broad gate as passing.

## AI disclosure

## Notes

An adversarial review completed with no critical or high findings. It identified a missing-runtime-path hardening gap; this PR now rejects that claim-only request with `400` and includes a regression test. The review's remaining notes concern pre-existing recovery/claim ownership semantics and do not regress this change.

### DIFF
--- a/docs/HOOKS.md
+++ b/docs/HOOKS.md
@@ -122,6 +122,15 @@ Called when a new agent session begins. Returns memories and context formatted f
 persistence scope. Harness-native sub-agent identifiers should be sent as
 `harnessAgentId`; they are used only for parent-session inference.
 
+When a daemon restart is detected during an already-running session, a harness
+may send `claimOnly: true` with the same `sessionKey` and a `plugin` or `legacy`
+runtime path in `x-signet-runtime-path` or `runtimePath`. Signet renews the
+runtime claim and returns `{ "sessionKnown": true }` without generating
+memories, identity, or an `inject` block. This mode rejects requests without a
+runtime path. Do not use it for a new session: it deliberately preserves the
+existing conversation's stable system prompt rather than constructing startup
+context.
+
 ### Response
 
 ```json
--- a/docs/api/sessions-hooks.md
+++ b/docs/api/sessions-hooks.md
@@ -37,7 +37,8 @@ for injection into the harness system prompt. Requires `remember` permission
   "harnessAgentId": "optional-harness-subagent-id",
   "parentSessionKey": "optional-parent-session-key",
   "sessionKey": "session-uuid",
-  "runtimePath": "plugin"
+  "runtimePath": "plugin",
+  "claimOnly": false
 }
 ```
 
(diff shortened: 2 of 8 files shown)


NicholaiVogel:
Adversarial review complete.

- No critical or high-severity regressions found.
- Fixed the actionable hardening gap: `claimOnly: true` without a `plugin` or `legacy` runtime path now returns `400`; the daemon regression suite covers it.
- Remaining observations concern pre-existing post-restart claim ownership and client thread timing, not behavior introduced by this PR.

Final focused verification: daemon route suite 24 pass / 0 fail; Hermes connector suite 74 pass / 0 fail; daemon typecheck passes.

PR #1242, opened 2026-08-08
feat(dashboard): make inference routes inspectable

NicholaiVogel:
## Summary

Fix issue #1234 by making dashboard inference settings write a complete explicit route and exposing the daemon's effective route state in the UI. The settings dialog now offers a bounded routed inference probe instead of showing only model/provider configuration.

## Changes

- Add route scaffolding when inference targets are saved:
  - explicit default policy
  - memory extraction workload binding
  - aggregate recall workload binding
  - generated `memory_extraction` task-class defaults
- Preserve custom policies and task classes while repairing missing default policy entries.
- Add a Runtime route card with policy, task-class, target, workload, executor availability, route explain results, and a real bounded inference probe.
- Extend the daemon status response with the aggregate recall workload binding.
- Add API and documentation entries for detailed route status and the health probe.
- Add regression coverage for route scaffolding and aggregation-target updates.

## Type

- [x] `feat` — new user-facing feature (bumps minor)
- [ ] `fix` — bug fix
- [ ] `refactor` — restructure without behavior change
- [ ] `chore` — build, deps, config, docs
- [ ] `perf` — performance improvement
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [x] `@signet/cli` / dashboard
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: none

## Screenshots

The Inference settings dialog was verified locally. It shows the Runtime route card between Model Assignment and Connected Providers, including the Check routes control and unavailable-state explanation when no daemon route is configured.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`)
- [x] Agent scoping verified on all new/changed data queries
- [x] Input/config validation and bounds checks added
- [x] Error handling and fallback paths tested (no silent swallow)
- [x] Security checks applied to admin/mutation endpoints
- [x] Docs updated for API/spec/status changes
- [x] Regression tests added for each bug fix
- [x] Lint/typecheck/tests pass locally

## Migration Notes (if applicable)

- [ ] Migration is idempotent
- [ ] Rollback / compatibility note included in PR description

## Testing

- [ ] `bun test` passes
- [ ] `bun run typecheck` passes
- [ ] `bun run lint` passes
- [ ] Tested against running daemon
- [x] N/A

## AI disclosure

- [ ] No AI tools were used in this PR
[...]

Issue #1234, opened 2026-08-08
Inference settings UI cannot produce or expose a complete dreaming route


NicholaiVogel:
## Bug Description

The desktop Inference settings UI exposes only provider, model, and endpoint for:

- **Backend inference** (`primary`)
- **Aggregation** (`fallback`)
- **Embeddings**

There is no UI for `inference.defaultPolicy`, `inference.policies`, `inference.taskClasses`, or `inference.workloads`.

That makes the UI insufficient for an operator to ensure the model they just configured can actually service dreaming/background work. The UI's stated job is model assignment, so a successful save should create a complete usable route or visibly explain what remains unconfigured.

## Evidence

The attached UI configuration had only these available controls:

- Backend inference provider/model/endpoint
- Aggregation provider/model/endpoint
- Embeddings provider/model

For a local llama.cpp server, the valid user inputs were:

```text
provider: llama.cpp (local)
model: google_gemma-4-26B-A4B-it-Q5_K_M.gguf
endpoint: http://192.168.1.10:8000/v1
```

The user cannot choose or inspect the things the router needs to make background behavior explicit:

```yaml
inference:
  defaultPolicy: local-llama
  policies:
    local-llama:
      mode: automatic
      defaultTargets: [background/default]
      fallbackTargets: [background/default]
  taskClasses:
    memory_extraction:
      reasoning: low
      privacy: restricted_remote
  workloads:
    memoryExtraction:
      target: background/default
      taskClass: memory_extraction
    aggregateRecall:
      target: aggregation/default
```

Without these, an operator has no way in the UI to diagnose or repair a route that has targets but no effective policy/workload behavior. v0.176.11 currently synthesizes a default policy in some cases, but that is a runtime fallback, not a clear, persisted configuration contract.

## Expected Behavior

Saving the primary/fallback UI fields should atomically create and validate a complete canonical route:

1. target(s) with the selected executor/model/endpoint;
2. an explicit default policy using the primary target and fallback target;
3. a `memoryExtraction` workload bound to the primary target;
4. an `aggregateRecall` workload bound to the aggregation target, if configured;
5. suitable generated task-class defaults for the selected use case; and
6. a real routed health test, not just model discovery/reachability.

Alternatively, expose an Advanced section that makes policies, workload assignment, privacy tier, task class, and the effective routed decision inspectable and editable. The UI must not report a provider as configured while the user has no way to see whether dreaming can use it.

## Why this matters

[...]

PR #1245, opened 2026-08-09
fix(dashboard): make generated routes routable

NicholaiVogel:
## Summary

Follow-up to #1242. The generated `memory_extraction` task class previously required tool-capable and restricted-remote targets, but the dashboard did not emit those capability fields. That made the generated primary route resolve to `no-candidates`.

## Changes

- Keep the generated task class at the canonical medium reasoning default without asserting unsupported tool/privacy capabilities.
- Add a decision-engine regression test over the generated config and a dashboard target.
- Make the health probe require a non-empty routed response, a selected decision target, and a successful execution attempt.
- Avoid reporting a probe failure when no probe was run.

## Verification

- Dashboard: 12 tests passed, typecheck passed, build passed.
- Routing decision regression passes against the real core decision engine.
- `git diff --check` passed.

Fixes the post-merge defect found in the adversarial review of #1234.

## Type

- [x] Bug fix

Assisted-by: Hermes-Agent:gpt-5.6-luna

## Readiness checklist

### DIFF
--- a/surfaces/dashboard/src/lib/api.ts
+++ b/surfaces/dashboard/src/lib/api.ts
@@ -142,7 +142,8 @@ export interface InferenceRouteDecision {
 
 export interface InferenceProbeResult {
 	readonly text: string;
-	readonly decision?: InferenceRouteDecision;
+	readonly decision: InferenceRouteDecision;
+	readonly attempts: readonly { readonly targetRef: string; readonly ok: boolean }[];
 }
 
 export type OAuthLoginEvent =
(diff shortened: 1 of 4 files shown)


NicholaiVogel:
Adversarial review: SHIP.

The review reproduced the original no-candidates defect against the real routing decision engine, verified that the generated route now selects the dashboard-created target, and found no remaining CRITICAL or HIGH issues. The execute probe now also requires a selected route, a successful attempt, and non-empty output before reporting success.

Verification: dashboard 12 tests, typecheck, and build passed; all required PR checks passed, including smoke, daemon checks, CodeQL, and the readiness checklist.

PR #1251, opened 2026-08-09
fix(desktop): open Codex OAuth externally

### LINKED ISSUE #1247: Desktop ChatGPT Codex OAuth sign-in opens a blank window instead of the OpenAI login page
## Summary

In Signet Desktop, choosing **Sign in with ChatGPT Codex** opens a browser/popup at `about:blank` and never navigates to OpenAI. The dashboard exposes no usable error or recovery path, so the OAuth account cannot be connected.

This is distinct from #966, which tracked the former absence of pi-ai OAuth support. The current implementation includes the OAuth API and dashboard flow, but the Electron popup strategy prevents the emitted authorization URL from being opened.

## Environment

- Signet Desktop `0.176.11`
- Linux
- OpenAI Codex OAuth provider is present and reported disconnected by `GET /api/inference/oauth/providers`

## Reproduction

1. Open the Signet Desktop dashboard.
2. Open inference/provider settings and choose the OpenAI Codex / ChatGPT sign-in path.
3. Select **Browser login** when prompted.
4. Observe the browser/popup remains at `about:blank`; OpenAI login never appears.

## Runtime evidence

The daemon's OAuth service is functional up to the interactive selection and advertises the expected browser path:

```text
event: session
data: {"type":"session","providerId":"openai-codex",...}

event: select
data: {"type":"select","message":"Select OpenAI Codex login method:","options":[
  {"id":"browser","label":"Browser login (default)"},
  {"id":"device_code","label":"Device code login (headless)"}
],...}
```

After selecting `browser`, the pi-ai OAuth provider should emit an `auth` event containing the OpenAI URL. The dashboard currently creates a blank popup before that event:

```ts
// surfaces/dashboard/src/components/settings/connect-dialog.tsx
const win = window.open("about:blank", "signet-oauth", "width=640,height=760");
```

But the Desktop main process denies all new windows, including that initial `about:blank` popup:

```ts
// surfaces/desktop/src/main.ts
mainWindow.webContents.setWindowOpenHandler(({ url }) => {
  if (url.startsWith("http://") || url.startsWith("https://")) {
    shell.openExternal(url).catch(() => undefined);
  }
  return { action: "deny" };
});
```

Thus the very window the dashboard stores for later `win.location.href = authUrl` is denied before a real `https:` URL is available. Depending on Chromium/Electron behavior, the user either sees the blank external navigation or has a null/closed popup reference that is silently ignored by `navigateOAuthWindow()`.

The current UI swallows this state:

```ts
if (!win || win.closed) return;
try { win.location.href = url; } catch { /* ... */ }
```

There is no fallback to `shell.openExternal(authUrl)`, no visible error, and no direct clickable auth URL/device-code fallback when popup navigation fails.

## Expected behavior

Selecting Browser login reliably opens the actual OpenAI authorization URL in the system browser, or Signet renders a safe clickable URL plus a device-code fallback. It must not strand the user on `about:blank`.

## Suggested fix

Use a desktop-aware auth-navigation bridge rather than browser-popup semantics:

1. Add a context-isolated preload IPC method, e.g. `signet.openExternal(url)`.
2. On the SSE `auth` event, validate `https:` and ask the main process to call `shell.openExternal(authUrl)`.
3. Keep the dashboard dialog open with status/cancel controls while the system browser handles authentication.
4. If external opening fails, render the validated URL as a copyable/clickable fallback and surface the failure.
5. Preserve device-code login as a first-class fallback for headless/restricted browser environments.

At minimum, do not create an `about:blank` child that Electron's `setWindowOpenHandler` will deny.

## Related

- #966 — prior OAuth integration gap, now closed
- #1234 — inference settings do not expose a complete route configuration

NicholaiVogel:
## Summary

Fixes #1247. Signet Desktop no longer creates an Electron-denied `about:blank` popup for ChatGPT Codex OAuth. The desktop renderer uses a validated HTTPS external-navigation bridge, while normal browser sessions retain the synchronous popup fallback.

## Changes

- Added a context-isolated `signetDesktop.openExternal` preload/Main IPC bridge.
- Validate OAuth navigation URLs as HTTPS before calling Electron `shell.openExternal`.
- Keep the OAuth URL visible as a clickable fallback and surface external-open failures.
- Added regression coverage for URL validation and bridge wiring.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: `signet-dashboard`, `@signet/desktop`

## Screenshots

- Local Inference settings dialog exercised against the Vite dashboard. This fix changes Electron navigation/error handling, not the layout; the existing OAuth URL fallback remains visible in the dialog.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

## AI disclosure

## Notes

- Issue-specific proof is green: `bun test surfaces/desktop/src/external-url.test.ts`, `bun run --filter signet-dashboard typecheck`, `bun run --filter @signet/desktop typecheck`, both affected package builds, `git diff --check`, and Biome on all newly added files plus the desktop main path.
- The full local workspace `bun run lint` and `bun run typecheck` currently report pre-existing unrelated repository failures in generated package formatting and connector/daemon build artifacts; no errors were introduced in the affected package checks above.
- Browser sessions retain the old synchronous popup path; Desktop sessions bypass it through the preload bridge.

### DIFF
--- a/surfaces/dashboard/src/components/settings/connect-controller.ts
+++ b/surfaces/dashboard/src/components/settings/connect-controller.ts
@@ -43,8 +43,7 @@ export interface ConnectControllerOptions {
 	readonly providerId: string;
 	readonly supportsOAuth: boolean;
 	readonly supportsApiKey: boolean;
-	/** Fired when the daemon emits a URL the user must open — the caller
-	 * navigates a popup opened synchronously in the click gesture. */
+	/** Fired when the daemon emits a URL the user must open. */
 	readonly onNavigate?: (url: string) => void;
 	/** Fired once when the login stores the credential daemon-side. */
 	readonly onConnected?: () => void;
--- a/surfaces/dashboard/src/components/settings/connect-dialog.tsx
+++ b/surfaces/dashboard/src/components/settings/connect-dialog.tsx
@@ -7,6 +7,7 @@ import { useConnectController } from "@/components/settings/connect-controller";
 import { api } from "@/lib/api";
 import { getDesktopBridge } from "@/lib/desktop";
 import { apiKeyFormat, providerKeySecretName } from "@/lib/inference-keys";
+import { createOAuthNavigation, safeOAuthHref, type OAuthNavigation } from "@/lib/oauth-navigation";
 import { cn } from "@/lib/utils";
 import { CheckCircle, Eye, EyeOff, KeyRound, Loader2, TriangleAlert, X } from "lucide-react";
 import { useEffect, useRef, useState } from "react";
@@ -18,12 +19,6 @@ export interface ConnectableProvider {
 	supportsApiKey: boolean;
 	connected: boolean;
 	isOAuth: boolean;
-}
-
-/** Allowlist http(s) for any daemon-sourced URL we bind to href. */
-function safeHref(uri: string | undefined): string | null {
-	if (!uri) return null;
-	return /^https?:\/\//i.test(uri) ? uri : null;
 }
 
 function hostnameOf(uri: string): string {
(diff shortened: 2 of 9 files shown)


NicholaiVogel:
Adversarial review follow-up:

- Replaced the source-text wiring assertion with runtime tests that prove Desktop uses `openExternal` without creating a popup.
- Unified OAuth navigation around HTTPS-only URLs for both Desktop and browser fallback.
- Added malformed/host validation coverage.
- Deduplicated repeated successful auth URLs.
- Guarded delayed bridge callbacks after unmount and hid external-open errors while interactive prompts are active.

The reviewer found no OAuth-flow blocker after these fixes. Focused tests now pass: 4 tests / 15 assertions.

PR #1293, opened 2026-08-09
feat(telemetry): add source lifecycle telemetry (#1276)

### LINKED ISSUE #1276: telemetry: measure privacy-safe source lifecycle and time to value
## Problem

`pipeline.embedding` tells us that work was embedded, but not whether a connected source became useful. A recent install indexed 1,217 embedding operations in about 30 minutes and then disappeared. Current telemetry cannot tell whether indexing completed meaningfully, whether the source became searchable, whether the user ever recalled from it, or whether the source later stayed fresh.

This prevents product decisions about which source adapters create retention versus one-shot ingestion.

## Scope

Add a bounded source lifecycle contract with no source names, paths, URLs, document titles, or content:

- source class from a fixed taxonomy, such as transcript, document, repository, note vault, browser, or other;
- connect/configure success or bounded failure class;
- discovery, accepted, skipped, and failed item counts;
- source-size and duration buckets;
- indexed/searchable readiness;
- first recall after source readiness;
- freshness state and lag buckets for streaming or recurring sources.

Prefer operation-level summaries over one event per item.

## Acceptance criteria

- Analytics can derive connect to first-index, connect to first-searchable, and source-ready to first-recall funnels.
- Analytics can compare source classes without content or user-defined source identifiers.
- Operators can distinguish a completed one-shot ingest, a failed ingest, and a healthy recurring source.
- Event volume is bounded for large corpora.

## Related

- #1181 and #1201 cover embedding usage and economics.
- #1203 covers search-level usage.

aaf2tbz:
## Summary

Fixes #1276 by adding bounded, privacy-safe source lifecycle telemetry so analytics can measure source connection, ingest, readiness, first recall, and recurring freshness without receiving source identifiers or content.

## Changes

- Add the `source.lifecycle` event contract with fixed source classes, bounded failure classes, counts, size/duration buckets, readiness milestones, first-recall latency, and freshness lag.
- Add migration 120 for agent-scoped local lifecycle state keyed only by a digest; remove state when sources are deleted.
- Instrument source configuration/index jobs, startup indexing, recurring gateway progress, and shared recall attribution with operation-level summaries.
- Add privacy/bucket contract tests and update telemetry/API documentation.

## Type

- [ ] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [x] `@signet/web`
- [ ] `predictor`
- [x] Other: telemetry contract and documentation

## Screenshots

N/A — no UI changes.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

Migration 120 is append-only and creates `source_lifecycle_state` with `CREATE TABLE IF NOT EXISTS` and a stable index. Existing databases retain all source and telemetry data; source lifecycle state is local correlation metadata and is cleaned up on source deletion. No rollback is required beyond reverting the application migration before it is applied; once applied, the unused table is harmless to older binaries.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Focused validation completed:

- `bun test platform/core/src/migrations/migrations.test.ts` — 61 passed
- `bun test platform/daemon/src/source-lifecycle-telemetry.test.ts platform/daemon/src/routes/sources-routes.test.ts` — 33 passed
- `bun test platform/daemon/src/telemetry.test.ts platform/daemon/src/memory-search-telemetry.test.ts platform/daemon/src/memory-search.test.ts` — 80 passed
- `bun run --filter '@signet/core' build` — passed
- `bun run --filter '@signet/daemon' build` — passed
- `bun run --filter '@signet/daemon' typecheck` — passed
- Changed-file Biome checks — passed

Repository-wide `bun test`, `bun run typecheck`, and `bun run lint` were also attempted. The current main baseline has unrelated parallel-test/global-state failures, missing generated connector artifacts, and repository-wide formatting drift; those failures are documented in the implementation session and do not involve the changed files.

## AI disclosure

## Notes

Telemetry is best-effort and never blocks source ingestion or recall. The external event payload contains only fixed taxonomy values, capped counts, buckets, and lifecycle outcomes; local hashed state is never sent. An independent final self-review reported no actionable findings.

### DIFF
--- a/platform/daemon/src/source-lifecycle-telemetry.ts
+++ b/platform/daemon/src/source-lifecycle-telemetry.ts
@@ -175,15 +175,15 @@ export function sourceFailureClass(error: unknown): SourceFailureClass {
 			: "";
 	if (/cancel|abort/i.test(text)) return "cancelled";
 	if (/unsupported|unknown provider|no sync/i.test(text)) return "unsupported";
+	if (
+		/configuration|config(?:uration)?|required|must be provided|at least one .* (?:id|guild|repo|source)/i.test(text)
+	) {
+		return "configuration";
+	}
 	if (/EACCES|ENOENT|EISDIR/i.test(code) || /enoent|eacces|file|directory|path|vault/i.test(text)) return "filesystem";
 	if (/403|permission|forbidden|not authorized|authorization/i.test(text)) return "authorization";
 	if (/token|credential|secret|auth|401/i.test(text)) return "authentication";
 	if (/429|rate limit|too many requests/i.test(text)) return "rate_limited";
-	if (
-		/configuration|root is required|guild id is required|missing .*config|config/i.test(text) ||
-		(/invalid/i.test(text) && /source|root|guild|config|provider/i.test(text))
-	)
-		return "configuration";
 	if (/json|parse|invalid|malformed|schema/i.test(text)) return "parse";
 	if (/network|fetch|socket|timeout|dns|connect|gateway|http 5/i.test(text)) return "network";
 	return "unknown";
--- a/platform/daemon/src/source-lifecycle-telemetry.ts
+++ b/platform/daemon/src/source-lifecycle-telemetry.ts
@@ -78,6 +78,7 @@ export interface SourceIndexTelemetryInput {
 	readonly failureClass?: SourceFailureClass;
 	readonly searchable?: boolean;
 	readonly sourceBytes?: number;
+	readonly updateFreshness?: boolean;
 }
 
 export interface SourceRecallTelemetryResult {
@@ -246,6 +247,7 @@ function ensureState(
 }
 
 export function recordSourceConnected(source: SignetSourceEntry, agentId: string, mode = sourceModeFor(source)): void {
+	if (!getActiveTelemetry()?.enabled) return;
 	const now = new Date().toISOString();
 	try {
 		let claimed = false;
@@ -285,17 +287,59 @@ export function recordSourceConnectionFailure(kind: string, error: unknown, mode
 function sourceSizeBytes(source: SourceLifecycleSource, agentId: string): number | undefined {
 	try {
 		return getDbAccessor().withReadDb((db) => {
+			const rootPrefix = `${(source.root ?? "").replace(/\\/g, "/").replace(/\/$/, "")}/`;
+			const legacyObsidianClause =
+				source.kind === "obsidian"
+					? "OR (harness = 'obsidian' AND source_id IS NULL AND source_path >= ? AND source_path < ?)"
+					: "";
 			const row = db
 				.prepare(`
 				SELECT SUM(length(content)) AS bytes
 				FROM memory_artifacts
-				WHERE agent_id = ? AND source_id = ? AND COALESCE(is_deleted, 0) = 0
+				WHERE agent_id = ?
+				  AND (
+					source_id = ?
+					${legacyObsidianClause}
+				  )
+				  AND COALESCE(is_deleted, 0) = 0
 			`)
-				.get(agentId, source.id) as { bytes?: unknown } | undefined;
+				.get(agentId, source.id, ...(source.kind === "obsidian" ? [rootPrefix, `${rootPrefix}\uffff`] : [])) as
+				| { bytes?: unknown }
+				| undefined;
 			return typeof row?.bytes === "number" && Number.isFinite(row.bytes) ? row.bytes : undefined;
 		});
 	} catch {
 		return undefined;
+	}
+}
+
+export function sourceHasSearchableArtifacts(source: SignetSourceEntry, agentId: string): boolean {
+	if (!getActiveTelemetry()?.enabled) return false;
+	try {
+		return getDbAccessor().withReadDb((db) => {
+			const rootPrefix = `${source.root.replace(/\\/g, "/").replace(/\/$/, "")}/`;
+			const legacyObsidianClause =
+				source.kind === "obsidian"
+					? "OR (harness = 'obsidian' AND source_id IS NULL AND source_path >= ? AND source_path < ?)"
+					: "";
+			const row = db
+				.prepare(
+					`SELECT 1 AS n FROM memory_artifacts
+					 WHERE agent_id = ?
+					   AND (
+						 source_id = ?
+						 ${legacyObsidianClause}
+					   )
+					   AND COALESCE(is_deleted, 0) = 0
+					   AND COALESCE(source_kind, '') NOT LIKE 'source_%_failure'
+					   AND COALESCE(source_kind, '') NOT LIKE 'source_%_checkpoint'
+					 LIMIT 1`,
+				)
+				.get(agentId, source.id, ...(source.kind === "obsidian" ? [rootPrefix, `${rootPrefix}\uffff`] : []));
+			return Boolean(row);
+		});
+	} catch {
+		return false;
 	}
 }
 
@@ -310,6 +354,7 @@ function freshnessEventNeeded(
 }
 
 export function recordSourceIndexOperation(input: SourceIndexTelemetryInput): void {
+	if (!getActiveTelemetry()?.enabled) return;
 	const mode = input.mode ?? sourceModeFor(input.source);
 	const now = new Date();
 	const nowIso = now.toISOString();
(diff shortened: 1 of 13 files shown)


platform/daemon/src/telemetry.ts
```diff
@@ -803,23 +727,51 @@ export function createTelemetryCollector(
 		},
 
 		record(event, properties): void {
```
NicholaiVogel, line 906: Blocking: this rewrite removes `TelemetryCollector.recordDeferred`, but the unchanged `platform/daemon/src/system-pressure.ts` still calls `telemetry.recordDeferred` at lines 75-76. On this exact head, `bun run --filter '@signet/daemon' typecheck` fails with TS2339 for both references. The existing wedge-audit test also fails because the optional deferred path no longer appends its durable JSONL signal. Restore the interface/implementation or update the caller while preserving the non-blocking, pre-flush audit contract.

platform/daemon/src/telemetry.ts
```diff
@@ -803,23 +727,51 @@ export function createTelemetryCollector(
...
 		recordFirstUse(kind): void {
-			const event = persistFirstUse(kind);
-			if (!event) return;
-			// The database row is durable before the open log mirror is written.
-			// A failed log write must not affect the claim or delivery queue.
-			appendToTelemetryLog(logPath, JSON.stringify(event));
+			// Best-effort: a failed claim (no row, broken DB) means the
+			// milestone stays unclaimed and may fire on a later run —
+			// still exactly once overall.
+			if (!claimFirstUse(kind)) return;
```
NicholaiVogel: Blocking: `claimFirstUse` now commits only the `telemetry_install` claim, then `collector.record` puts the event in the in-memory buffer. A process death between those operations and the next flush loses the event permanently while the claim remains set. The existing `telemetry.test.ts` test `recovers first-use events after termination before the normal flush` fails with no pending events on reopen. Keep the claim and event insert atomic in the same write transaction (or otherwise make the event durable before claiming).

platform/daemon/src/telemetry.ts
```diff
@@ -849,16 +801,7 @@ export function createTelemetryCollector(
...
-				await flush();
-			} catch {}
-			// Stop accepting new events only after the first flush has completed,
-			// so events produced while an outbound request was awaiting are
-			// drained by this final serialized flush.
-			recordingStopped = true;
-			try {
-				await flush();
-			} catch {}
+			await doFlush();
```
NicholaiVogel: High: `stop()` now performs only one `doFlush()` and accepts events for the entire await. If `sendToPostHog` is in flight, a concurrent `record()` appends to `buffer` after line 612 drained it; `stop()` then returns with `running = false`, the timer cleared, and that event stranded in memory. The previous implementation explicitly did a first flush, then stopped recording, then a final flush to drain this window. Preserve serialized flushes and the final drain before reporting the collector stopped.

NicholaiVogel (changes requested):
Blocking review on the exact PR head aa6e8b8a8.

Evidence run in an isolated PR worktree after building @signet/core:
- `bun test platform/daemon/src/source-lifecycle-telemetry.test.ts platform/daemon/src/routes/sources-routes.test.ts`: 33 passed.
- `bun test platform/daemon/src/telemetry.test.ts platform/daemon/src/memory-search-telemetry.test.ts platform/daemon/src/memory-search.test.ts`: 77 passed, 2 failed. The failures are existing first-use crash recovery and wedge audit-log tests.
- `bun run --filter '@signet/daemon' typecheck`: failed on the unchanged system-pressure caller because `TelemetryCollector.recordDeferred` was removed.
- `bun run --filter '@signet/daemon' build`: passed, but the build path does not replace the production typecheck gate.

Please address the inline findings and rerun the affected suites plus daemon production typecheck.

platform/daemon/src/source-lifecycle-telemetry.ts
```diff
@@ -0,0 +1,616 @@
...
+export function sourceLagBucket(lagMs: number | null | undefined): SourceLagBucket {
+	if (lagMs === null || lagMs === undefined || !Number.isFinite(lagMs) || lagMs < 0) return "unknown";
+	if (lagMs < 3_600_000) return "lt_1h";
+	if (lagMs < 21_600_000) return "1_6h";
+	if (lagMs < 86_400_000) return "6_24h";
+	if (lagMs < 604_800_000) return "1_7d";
+	return "7d_plus";
+}
+
+export function sourceFailureClass(error: unknown): SourceFailureClass {
```
NicholaiVogel, line 165: Additional non-blocking finding: `SOURCE_FAILURE_CLASSES` declares `configuration`, and source-add validation failures are passed here, but this classifier has no configuration branch. Errors such as “Obsidian root is required” and “At least one Discord guild ID is required” therefore become `unknown`; validation errors containing “invalid” can be mislabeled `parse`. Add an explicit configuration classification so the new connection-failure telemetry reports the intended category.

platform/daemon/src/telemetry.ts
```diff
@@ -757,69 +720,62 @@ export function createTelemetryCollector(
...
 		recordFirstUse(kind): void {
-			const event = persistFirstUse(kind);
-			if (!event) return;
-			// The database row is durable before the open log mirror is written.
-			// A failed log write must not affect the claim or delivery queue.
-			appendToTelemetryLog(logPath, JSON.stringify(event));
+			// Best-effort: a failed claim (no row, broken DB) means the
+			// milestone stays unclaimed and may fire on a later run —
+			// still exactly once overall.
+			if (!claimFirstUse(kind)) return;
```
NicholaiVogel: Blocking on the current PR head: `claimFirstUse` still commits only the `telemetry_install` claim, then `collector.record` places the event in the in-memory buffer. A process death before the next flush leaves the timestamp claimed but no `first.remember`/`first.recall` event, and later runs will not recreate it. The existing crash-recovery test reproduces this. Keep the claim and event insertion atomic in one write transaction, or durably insert the event before claiming.

platform/daemon/src/telemetry.ts
```diff
@@ -849,16 +805,7 @@ export function createTelemetryCollector(
...
-				await flush();
-			} catch {}
-			// Stop accepting new events only after the first flush has completed,
-			// so events produced while an outbound request was awaiting are
-			// drained by this final serialized flush.
-			recordingStopped = true;
-			try {
-				await flush();
-			} catch {}
+			await doFlush();
```
NicholaiVogel: Blocking on the current PR head: `stop()` still performs only one `doFlush()` while recording remains possible during the awaited PostHog request. An event recorded after the buffer is drained can remain stranded after `running = false` and the timer are cleared. The prior collector explicitly performed a final drain after stopping recording. Restore serialized flushes and a final shutdown drain, and add a regression test for an event recorded while PostHog is in flight.

NicholaiVogel (changes requested):
Re-review of current head 4cb5d667dfc7736e0e01d0eba06ce8e08f71427a: the prior `recordDeferred` typecheck finding is fixed by the compatibility commit. The first-use crash-loss and shutdown flush-race findings remain reproducible and blocking. The source failure-classification issue is also still present as a non-blocking telemetry correctness issue. Changes requested.

aaf2tbz:
Addressed Nicholai's requested changes and rebased onto current `main`:

- restored `recordDeferred`'s synchronous JSONL audit append while deferring SQLite/provider work;
- made first-use claim and durable telemetry event insertion atomic;
- serialized shutdown flushes and added a final drain for events recorded during an in-flight PostHog request;
- classified invalid source configuration failures as `configuration` instead of `parse`/`unknown`;
- added regressions for shutdown drain and configuration classification.

Validation after rebase:
- `@signet/core` build passed;
- daemon production typecheck passed;
- telemetry, source-lifecycle, and sources-route suites: 60 passed, 0 failed;
- working tree clean and `git diff --check` passed.

NicholaiVogel:
Follow-up fix is open: https://github.com/Signet-AI/signetai/pull/1315

It targets `averyfelts/issue-1276-fix` directly and restores the remaining telemetry guarantees:

- atomic first-use claim plus durable event insertion
- serialized flushes and the final shutdown drain
- source configuration failure classification
- regression coverage, including a mutation test that fails against the reverted shutdown implementation

The current #1293 head already contains the separate `recordDeferred` fix. Focused daemon/core verification passes; the full monorepo gates remain red on unrelated connector, formatting, and test-environment failures documented in the follow-up PR.

NicholaiVogel:
Correction to my earlier comment: after #1293 was force-pushed to `344637ab3`, its current head already contains the durable first-use and serialized shutdown implementation. Follow-up PR #1315 is now rebased onto that live head and contains only the additional regression hardening plus broader source-configuration classification:

https://github.com/Signet-AI/signetai/pull/1315

Its focused shutdown and classifier tests pass; the combined telemetry suite still has one pre-existing session-economics failure on the current #1293 head, documented in #1315.

PR #1299, opened 2026-08-10
fix(dashboard): canonicalize embedding endpoint (#1264)

### LINKED ISSUE #1264: Dashboard writes embedding baseUrl but daemon only reads base_url, leaving the selected Ollama endpoint unused
## Summary

The Inference/Embeddings dashboard save path writes camelCase `embedding.baseUrl`, while the daemon's embedding configuration parser and runtime use snake_case `embedding.base_url`.

This can leave an old `base_url` value active even though the dashboard appears to have updated the endpoint. The operator believes Signet is embedding through the selected Ollama server, but the daemon continues using the stale endpoint.

## Environment

- Signet daemon `0.182.8`
- Linux
- Ollama remote endpoint: `http://192.168.1.10:11434`
- model: `nomic-embed-text`

## Observed config after dashboard update

```yaml
embedding:
  provider: ollama
  model: nomic-embed-text
  dimensions: 768
  base_url: http://127.0.0.1:11434
  baseUrl: http://192.168.1.10:11434
```

The configured remote Ollama server is healthy and supports the legacy endpoint Signet uses:

```text
GET /api/version -> 200 (Ollama 0.20.2)
POST /api/embeddings with model=nom ic-embed-text -> 200, 768-dimension vector
```

But the runtime source resolves Ollama from `cfg.base_url`:

```ts
// platform/daemon/src/embedding-fetch.ts
return cfg.base_url;
```

The config tests likewise document `base_url` as the accepted setting. `baseUrl` is not recognized by the embedding config parser, so the dashboard's remote value is ignored while an existing snake-case value remains in force.

## Impact

- Dashboard settings can look correct while the daemon embeds against a different host.
- Existing local endpoint failures continue, producing repeated `Ollama embedding request failed` warnings.
- An operator can unknowingly send embeddings to an unintended server or believe migration/re-embedding uses a new server when it does not.

## Expected behavior

The dashboard and daemon should use one canonical serialized field. Saving an embedding endpoint in the UI must update the exact field the daemon resolves and must replace, rather than coexist with, the prior endpoint.

## Suggested fix

1. Dashboard writes `embedding.base_url`.
2. Daemon temporarily accepts `baseUrl` as a backwards-compatible alias, with a warning if both keys differ.
3. A config migration canonicalizes existing `baseUrl` records to `base_url` and removes the duplicate.
4. The dashboard displays the daemon's effective embedding endpoint and performs a real embedding probe after save.

This is analogous to the broader UI-to-runtime route mismatch in #1234 and the missing compatibility/consent flow in #1258.

NicholaiVogel:
## Summary

Fixes #1264. The dashboard now writes the daemon's canonical `embedding.base_url` field instead of `embedding.baseUrl`, so changing an Ollama endpoint updates the endpoint the daemon actually resolves. Existing camelCase entries are canonicalized during daemon startup.

## Changes

- Added dashboard embedding endpoint helpers that read legacy aliases but save only `base_url` and remove `baseUrl` / `endpoint` aliases.
- Added config migration v8 for both `embedding` and legacy `memory.embeddings` blocks.
- Preserved an existing canonical `base_url` when both spellings conflict and emit a warning.
- Added dashboard save and daemon migration regression coverage for #1264.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

Inference settings with runtime route status

No layout or visual surface changed; this is the same Inference settings endpoint field. The screenshot is the existing fresh dashboard capture from the related Inference settings work.

## PR Readiness (MANDATORY)

- [x] Migration is idempotent

The v8 startup migration is idempotent and uses the existing atomic YAML rewrite path. It moves `baseUrl` to `base_url`, removes the duplicate alias, preserves an existing canonical value on conflict, and leaves parse-error files untouched with a warning. Reverting the commit restores the prior runtime behavior but does not need a schema rollback.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [ ] N/A

Targeted verification:

- `bun test platform/daemon/src/config-migration.test.ts platform/daemon/src/memory-config.test.ts`: 96 passed, 0 failed.
- `bun test surfaces/dashboard/src/lib/agent-config.test.tsx`: 5 passed, 0 failed.
- `bun run --filter '@signet/daemon' typecheck`: passed.
- `bun run --filter signet-dashboard typecheck`: passed.
- `bun run --filter '@signet/core' build`: passed.
- `bun run --filter '@signet/daemon' build`: passed.
- `bun run --filter signet-dashboard build`: passed.
- Touched daemon/helper files pass the pinned Biome check; the full dashboard `settings.tsx` lint still reports pre-existing baseline diagnostics.
- `git diff --check`: passed.
- Regression sabotage confirmed the new dashboard test fails with the old camelCase writer and passes with the fix restored.

## AI disclosure

## Notes

The daemon remains the runtime source of truth. The dashboard accepts legacy spellings for display only, while all new saves use the canonical serialized field.

### DIFF
--- a/platform/daemon/src/config-migration.test.ts
+++ b/platform/daemon/src/config-migration.test.ts
@@ -2,7 +2,7 @@ import { afterEach, describe, expect, it } from "bun:test";
 import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { migrateInferenceProviders } from "./config-migration";
+import { migrateEmbeddingBaseUrl, migrateInferenceProviders } from "./config-migration";
 
 function setupDir(): string {
 	const dir = mkdtempSync(join(tmpdir(), "signet-config-migration-"));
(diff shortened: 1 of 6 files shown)


NicholaiVogel:
Adversarial review completed with a SHIP verdict. The review verified the dashboard read/write loop, migration ordering/version gate, runtime reachability, comment preservation, conflict handling, idempotence, and repo-wide `baseUrl` writers.

One non-blocking cleanup finding was addressed in commit `7a7cb0e55`: migration now removes a shadowed `endpoint` alias alongside `baseUrl`. The review also noted top-level `embeddings` / `config.yaml` legacy paths are outside the dashboard-written paths covered by #1264 and should be considered separately if those configs are found in the wild.

PR #1372, opened 2026-08-10
fix(recall): parse abbreviated date ranges

### LINKED ISSUE #1359: bug(recall): abbreviated ISO date ranges leave a dangling day token and cause false no-hits
## Bug description

`signet recall` mishandles abbreviated ISO date ranges such as `2026-07-25/26`. It recognizes only the first date, leaves the trailing day (`26`) in the normalized search text, and can return a false **No memories found** result even when the selected date contains matching memories.

This also explains the same failure under `--aggregate`: aggregation receives no evidence because the base recall has already been poisoned.

## Reproduction

Environment:

- Signet CLI/daemon: `0.188.1`
- OS: Linux
- Workspace has 2,502 memories and 100% embedding coverage

An existing memory captured within the local-day range for `2026-07-25` is retrievable by a date-only query:

```bash
signet recall "2026-07-25" --limit 10 --json
# returns matching temporal memory rows
```

But an abbreviated range in a natural-language query fails:

```bash
signet recall "what happened on 2026-07-25/26?" --json
```

Actual response metadata:

```json
{
  "results": [],
  "query": "happened 26",
  "method": "hybrid",
  "meta": { "noHits": true }
}
```

The aggregate form has the same no-evidence result:

```bash
signet recall "what happened on 2026-07-25/26?" --aggregate --no-save-aggregate
```

For comparison, a fully expanded range is recognized as temporal:

```bash
signet recall "2026-07-25/2026-07-26" --json
# meta.temporal.start = 2026-07-25T06:00:00.000Z
# meta.temporal.end   = 2026-07-26T06:00:00.000Z
```

## Root-cause evidence

In current `main`, `platform/daemon/src/temporal-recall.ts`:

- `parseExplicitDay()` (lines 170–190) matches the `2026-07-25` prefix of `2026-07-25/26`.
- `normalizeContentQuery()` (lines 192–200) removes only that prefix, leaving `26`.
- The temporal path then attempts the content-filter query `happened 26`; if it produces no temporal candidates, recall falls through to hybrid retrieval with that degraded query, rather than the original user query.

The date-only control above confirms this is not an embedding, daemon-health, scope, or session-dedupe failure.

## Expected behavior

Interpret `YYYY-MM-DD/DD` as an abbreviated same-month date range. For `2026-07-25/26`, temporal recall should cover both July 25 and July 26 and should remove the complete range from the content query.

At minimum, unsupported range syntax must not silently degrade the query to a dangling numeric term and produce a false no-hit result. It should either preserve the original query for non-temporal fallback or surface the syntax as unsupported.

## Suggested regression coverage

- `what happened on 2026-07-25/26?` produces temporal metadata and retains no dangling `26` in `contentQuery`.
- `2026-07-25/26` covers both local calendar days (end-exclusive boundary after July 26).
- `2026-07-25/2026-07-26` remains supported.
- `July 25/26 2026` is either supported equivalently or rejected/fallen back without query corruption.
- `--aggregate` receives the recovered base evidence.

## Related

This is distinct from #903, which concerned ranking freshness for otherwise valid recall candidates. Here, an explicit date-range query is parsed/normalized incorrectly before candidate retrieval.

NicholaiVogel:
## Summary

Closes #1359. Fixes recall queries containing abbreviated ISO date ranges such as `2026-07-25/26`, which previously left `26` in the normalized content query and could produce false no-hit responses.

## Changes

- Parse abbreviated numeric ranges as inclusive local-calendar-day windows with an end-exclusive boundary after the final day.
- Consume full numeric and named range expressions so date tokens do not leak into content queries.
- Reject reversed ranges rather than treating them as temporal queries.
- Add parser regression coverage and document the supported range forms.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: documentation under `web/docs/src/content/docs/`

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

No migration or compatibility data changes.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Focused verification:

- `bun test src/temporal-recall.test.ts` (4 pass)
- `bun test src/memory-search.test.ts` (50 pass)
- `bun run --filter '@signet/daemon' typecheck`
- `bunx biome check platform/daemon/src/temporal-recall.ts platform/daemon/src/temporal-recall.test.ts`
- `git diff --check`

The repository-wide `doc-drift.ts` check reports pre-existing unrelated drift in route, migration, and package documentation inventories.

## AI disclosure

## Notes

The old parser behavior was reproduced directly: `what happened on 2026-07-25/26?` normalized to `happened 26`. The new parser returns `contentQuery: "happened"` and a local-day window from July 25 through the end of July 26.

### DIFF
--- a/platform/daemon/src/temporal-recall.test.ts
+++ b/platform/daemon/src/temporal-recall.test.ts
@@ -3,7 +3,46 @@ import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { closeDbAccessor, getDbAccessor, initDbAccessor } from "./db-accessor";
-import { resolveTemporalRecall } from "./temporal-recall";
+import { parseTemporalRecallIntent, resolveTemporalRecall } from "./temporal-recall";
+
+describe("temporal recall parser", () => {
+	it("parses abbreviated numeric ranges without leaving the end day in the content query", () => {
+		const parsed = parseTemporalRecallIntent({ query: "what happened on 2026-07-25/26?" });
+		expect(parsed).toMatchObject({
+			source: "query",
+			contentQuery: "happened",
+			mode: "filter",
+			start: new Date(2026, 6, 25).toISOString(),
+			end: new Date(2026, 6, 27).toISOString(),
+		});
+	});
+
+	it("consumes expanded and named date ranges as complete temporal expressions", () => {
+		const expanded = parseTemporalRecallIntent({ query: "2026-07-25/2026-07-26" });
+		const named = parseTemporalRecallIntent({ query: "what happened on July 25/26 2026?" });
+
+		expect(expanded).toMatchObject({
+			contentQuery: "",
+			mode: "timeline",
+			start: new Date(2026, 6, 25).toISOString(),
+			end: new Date(2026, 6, 27).toISOString(),
+		});
+		expect(named).toMatchObject({
+			contentQuery: "happened",
+			mode: "filter",
+			start: new Date(2026, 6, 25).toISOString(),
+			end: new Date(2026, 6, 27).toISOString(),
+		});
+	});
+
+	it("does not treat a reversed abbreviated range as a valid temporal query", () => {
+		expect(parseTemporalRecallIntent({ query: "what happened on 2026-07-25/01?" })).toBeNull();
+	});
+
+	it("does not partially parse an unsupported numeric range suffix", () => {
+		expect(parseTemporalRecallIntent({ query: "report 2026-07-25/26th" })).toBeNull();
+	});
+});
 
 describe("temporal recall", () => {
 	let dir = "";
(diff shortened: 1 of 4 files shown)


NicholaiVogel:
Adversarial review: SHIP.

The read-only review found no CRITICAL, HIGH, or MEDIUM defects. It verified the abbreviated, expanded, named, cross-month, and cross-year numeric range behavior; local-day and half-open SQL boundaries; safe fallback for reversed/invalid ranges; preservation of single-date parsing; and regression tests that fail against the pre-fix implementation.

One LOW finding was addressed in the follow-up amendment 69af5313: unsupported numeric suffixes such as `2026-07-25/26th` now fall back without partially parsing the first date, with a regression test.

Remaining LOW notes are intentional/non-blocking: ambiguous cross-year named ranges fail closed, equal-day ranges are rejected, and only the first range in a multi-date query is consumed, consistent with the pre-existing single-date limitation.

PR #1387, opened 2026-08-10
fix(daemon): quarantine migration embedding failures

### LINKED ISSUE #1358: Embedding-index migration retries one oversized Ollama row forever and cannot promote
## Problem

When an embedding profile migration encounters a single active embedding that the target provider rejects for input length, the staging build retries that exact row every poll indefinitely. It never advances to later rows or promotes the new index.

This is separate from #1160 (stale configuration during migration): the provider is healthy and the persisted staging configuration is current. One poison-pill row blocks an otherwise healthy rebuild.

## Live reproduction

**Signet:** `0.188.1` (upgraded from `0.183.1`)

Current `embedding_index_state` is stuck in `state='building'` for the newly selected remote Ollama endpoint:

- active index: `nomic-embed-text`, 768 dimensions, `http://127.0.0.1:11434`
- staging index: `nomic-embed-text`, 768 dimensions, `http://192.168.1.10:11434`
- active `embeddings`: 45,004 rows
- `embeddings_staging`: 45,003 rows

The only missing staging row is an existing `memory` embedding:

```text
source_id: 3996fa80-9e3d-4e98-8665-72e3b4ad84ba
content_hash: 1b6cb8293376a06187121757521b920025c5e7e4d3d34dcd8e0826fe0cc2b43c
characters: 8520
```

Replaying Signet's actual Ollama request shape against the configured server returns:

```json
{
  "status": 500,
  "response_prefix": "{\"error\":\"the input length exceeds the context length\"}"
}
```

Normal provider health and short query embeddings succeed. `signet embed audit` also reports `2502 / 2502` active-memory coverage, which hides that the **staging** migration can never finish.

## Repeated runtime effect

After upgrading to 0.188.1, the daemon logs this every ~5.6 seconds, continuously:

```json
{
  "level": "warn",
  "category": "embedding",
  "message": "Ollama embedding request failed",
  "data": { "status": 500, "model": "nomic-embed-text" }
}
```

The migration code selects the first missing `embeddings_staging` row in `created_at` order. On `!vector`, `stageEmbeddingBatch()` simply continues, and the next tick selects that same row again. Unlike `embedding-tracker`, the migration has no per-content retry/backoff/quarantine or oversized-input handling.

## Expected behavior

A profile migration should make bounded, observable progress in the presence of a provider-rejected row:

1. Bound or safely chunk/truncate input for providers with a known context limit, preserving source provenance.
2. If a row cannot be represented, record a durable, inspectable failure/quarantine with source ID, hash, provider response category, and retry policy.
3. Do not retry the same failed row every poll indefinitely or prevent all later rows from staging.
4. Surface staging coverage and blocking failures in `signet embed audit` / dashboard, rather than reporting only active-index coverage as healthy.

No database content was modified for this report.

NicholaiVogel:
## Summary

Fixes #1358.

A provider can be healthy while one active embedding is permanently rejected by the target profile, for example an Ollama context-limit response. The migration previously treated that row as an ordinary null result, selected it again on every poll, and never promoted the rest of the rebuild.

## Changes

- Add durable `embedding_index_failures` state keyed by source content hash and target profile fingerprint.
- Persist source type/id, agent provenance, provider/model, failure category, attempts, and quarantine retry policy for terminal `context_limit` and `invalid_input` failures.
- Exclude quarantined rows from later polls and count only active quarantined rows as accounted-for staging coverage, allowing the healthy portion of the index to promote.
- Expose staging coverage and quarantine counts through the embedding status and `signet embed audit` surfaces.
- Add migration, SDK type, dashboard map, and documentation parity.
- Add regression coverage for progress past a poison row, for purging its active source afterward, and for rejecting an empty replacement when every active row is quarantined.

## PR Readiness

Notes: This is an additive daemon migration and status/audit read-surface change; it adds no admin or mutation endpoint. The migration query is scoped by target fingerprint and active embedding membership, and the regression test covers terminal failure handling and continued progress.

## Tests

- `bun test platform/daemon/src/embedding-index-migration.test.ts` - 12 pass
- `bun test platform/core/src/migrations/migrations.test.ts` - 63 pass
- `bun test platform/daemon/src/repair-actions.test.ts` - 59 pass
- `bun run --filter '@signet/core' build` - passed
- `bun run --filter '@signet/daemon' build` - passed
- `bun run --filter '@signet/sdk' build` - passed
- `bun run typecheck` - passed
- `bun run build:workspace-deps` from `surfaces/cli` - passed
- `bun run build:cli` from `surfaces/cli` - passed
- `bunx biome check` on changed TypeScript files - passed
- `git diff --check` - passed

## Risk and exclusions

This intentionally does not quarantine provider-unavailable, timeout, auth, quota, or rate-limit failures; those remain retryable. Quarantined rows are reported as staging failures and retain provenance, but they are not represented in the target vector index.

- [x] Migration is idempotent

Migration 123 is additive and uses `CREATE TABLE IF NOT EXISTS` plus `CREATE INDEX IF NOT EXISTS`. Existing embedding rows are untouched. Rolling back the application leaves the quarantine table unused; rolling back the schema requires the normal database backup/restore path and would remove failure history.

### DIFF
--- a/libs/sdk/src/types.ts
+++ b/libs/sdk/src/types.ts
@@ -700,6 +700,17 @@ export interface EmbeddingStatusResponse {
 	readonly base_url: string;
 	readonly error?: string;
 	readonly checkedAt: string;
+	readonly index?: {
+		readonly state: "ready" | "building" | "failed";
+		readonly coverage?: {
+			readonly active: number;
+			readonly staged: number;
+			readonly missing: number;
+			readonly wrongDimensions: number;
+			readonly quarantined: number;
+			readonly ready: boolean;
+		};
+	};
 }
 
 export interface EmbeddingHealthResponse {
--- a/platform/core/src/migrations/123-embedding-index-failures.ts
+++ b/platform/core/src/migrations/123-embedding-index-failures.ts
@@ -1,0 +1,32 @@
+import type { MigrationDb } from "./index";
+
+/**
+ * Persist terminal failures from an embedding-index migration. A provider can
+ * be healthy while one source row is not representable by its context window;
+ * keeping that decision in SQLite prevents a poison row from blocking future
+ * polls or disappearing from migration diagnostics.
+ */
+export function up(db: MigrationDb): void {
+	db.exec(`
+		CREATE TABLE IF NOT EXISTS embedding_index_failures (
+			id INTEGER PRIMARY KEY AUTOINCREMENT,
+			content_hash TEXT NOT NULL,
+			source_type TEXT NOT NULL,
+			source_id TEXT NOT NULL,
+			agent_id TEXT,
+			target_fingerprint TEXT NOT NULL,
+			provider TEXT NOT NULL,
+			model TEXT NOT NULL,
+			failure_class TEXT NOT NULL CHECK (failure_class IN ('context_limit', 'invalid_input')),
+			attempts INTEGER NOT NULL DEFAULT 1,
+			retry_policy TEXT NOT NULL DEFAULT 'quarantined',
+			first_failed_at TEXT NOT NULL,
+			last_failed_at TEXT NOT NULL,
+			UNIQUE(content_hash, target_fingerprint)
+		);
+		CREATE INDEX IF NOT EXISTS idx_embedding_index_failures_target
+			ON embedding_index_failures(target_fingerprint, failure_class, last_failed_at);
+		CREATE INDEX IF NOT EXISTS idx_embedding_index_failures_source
+			ON embedding_index_failures(source_type, source_id);
+	`);
+}
--- a/platform/core/src/migrations/index.ts
+++ b/platform/core/src/migrations/index.ts
@@ -128,6 +128,7 @@ import { up as telemetryVersionObservation } from "./119-telemetry-version-obser
 import { up as sourceLifecycleTelemetry } from "./120-source-lifecycle-telemetry";
 import { up as telemetryDeliveryHealth } from "./121-telemetry-delivery-health";
 import { up as dreamingEvidenceRetry } from "./122-dreaming-evidence-retry";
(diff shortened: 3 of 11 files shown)


NicholaiVogel:
Follow-up review finding and fix:

The first adversarial review timed out after a scratch experiment found that coverage could report `ready: true` when every active row was quarantined and `staged === 0`. That would allow promotion of an empty replacement index over the existing searchable generation.

Commit `57c4c513b` now requires `staged > 0` whenever active embeddings exist, while preserving promotion for genuinely empty datasets. Added regression coverage; the focused migration suite now passes 12/12 and the daemon build passes.

NicholaiVogel:
Adversarial review completed against head `57c4c513b`.

Verdict: SHIP. No CRITICAL, HIGH, or MEDIUM defects found.

The prior empty-promotion defect is fixed: active rows now require at least one staged row before promotion, while genuinely empty datasets remain promotable. The focused migration suite passed 12/12.

Three LOW follow-ups were noted:
- Quarantine rows are permanent across configuration fingerprint toggles; returning to a previously failing fingerprint does not provide a manual unquarantine path.
- A single-row retryable provider failure can keep a build in `building` indefinitely because only provider health-check failures have an abort/backoff path.
- The persisted `attempts` counter is effectively always 1 because quarantined rows are excluded from subsequent fetches.

These are consistent with the deliberate terminal-vs-retryable design and are not blockers for #1358. The reviewer also noted that the branch was based on 0.191.3 while `origin/main` has moved to 0.192.1; I am rebasing before merge.

NicholaiVogel:
Rebased onto current origin/main at 959472b1c (0.192.1). The PR diff remains limited to the 11 issue-related files; force-pushed with --force-with-lease. New CI checks are running against the rebased head.

PR #1443, opened 2026-08-11
fix(provenance): validate proposal sources

NicholaiVogel:
## Summary

Reject malformed, fabricated, and cross-agent ontology `source_ref` evidence before any derived semantic-memory provenance is written.

## Changes

- Add a strict canonical episodic-source resolver scoped to the proposal agent.
- Resolve proposal evidence transactionally before derived-memory materialization and reject invalid, missing, or cross-scope refs with 409 errors.
- Revalidate strict `source_ref` evidence at the common `applyOperation` apply seam, so a pending proposal whose evidence source disappears after creation fails closed with a 409 before any handler mutation or dedupe early-return (repair, see below).
- Add scratch-DB regressions for valid, missing, blank, and cross-agent source refs across direct and pending proposal applies.

## Repair (apply-time revalidation)

Follow-up review found that source refs were validated on insertion and again only inside `materializeAttributeMemoryInTx`. A pending proposal whose evidence source disappeared after creation could still be applied through non-materializing handlers (`create_entity`, `create_link`, ...) and through the dedupe early-returns in `add_claim_value`/`set_claim_value`. Both paths now revalidate the proposal's stored evidence at the shared `applyOperation` seam, inside the same transaction, before any semantic mutation or dedupe return.

New regressions:

- `rejects a deduped claim apply whose evidence source disappeared after creation` — pending `set_claim_value` with deleted source stays pending, attribute/derived counts unchanged.
- `rejects a non-materializing apply whose evidence source disappeared after creation` — pending `create_entity` with deleted source stays pending, no entity row created.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

- [ ] Migration is idempotent

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Focused verification (at current head `0e8e2965e`):

- `bun test platform/daemon/src/ontology-proposals.test.ts` (65 pass, 0 fail; includes the two new apply-time regressions)
- `bun run --filter '@signet/daemon' typecheck`

The full workspace typecheck and lint are currently red from unrelated connector exports and repository-wide formatting diagnostics. This change does not touch those files.

## AI disclosure

## Notes

The resolver intentionally accepts only explicit canonical episodic kinds and does not perform cross-store fallback. Invalid source refs fail before attribute, derived-memory, or provenance-link writes can commit. Apply-time revalidation reads the proposal's stored evidence, so direct creation behavior and transaction rollback semantics are unchanged.

### DIFF
--- a/platform/daemon/src/episodic-sources.ts
+++ b/platform/daemon/src/episodic-sources.ts
@@ -75,6 +75,10 @@ export interface ReadEpisodicSourceOptions {
 	readonly from: string;
 }
 
+export type StrictEpisodicSourceRefResolution =
+	| { readonly status: "resolved"; readonly source: EpisodicSourceRecord }
+	| { readonly status: "invalid" | "not_found" | "cross_agent"; readonly sourceRef: string };
+
 const SOURCE_KIND_RANK: Readonly<Record<EpisodicSourceKind, number>> = {
 	memory: 0,
 	artifact: 1,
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Repair applied for the apply-time source revalidation hole found in review (task t_1337ad9c).

## What was wrong

Source refs were validated on proposal insertion and again only inside `materializeAttributeMemoryInTx`. A pending proposal whose evidence source disappeared after creation could still be applied:

- through non-materializing handlers (`create_entity`, `create_link`, rename/archive/pin/etc.), which never touch the materializer, and
- through the early dedupe returns in `add_claim_value` / `set_claim_value`, which return before materialization.

That produced an applied proposal with unavailable or cross-agent evidence instead of a 409.

## Fix

Added `validateProposalEvidenceSourcesInTx(db, agent_id, proposalAuditEvidence(proposal))` at the top of the shared `applyOperation` seam (the single funnel for applyOntologyProposal, applyOntologyOperation, applyOntologyOperationBatch, and applyOntologyOperationBatchInTx). Every pending-proposal apply now revalidates its stored strict `source_ref` evidence inside the same transaction, before any semantic mutation or dedupe return. A 409 is thrown, the transaction rolls back, and the proposal stays pending (markFailed skips 409).

Direct creation behavior and transaction rollback semantics are unchanged: insertion still validates, and the seam revalidation is a no-op for same-transaction insert+apply paths.

## Regression proof

Two new tests, both of which fail on the pre-fix head and pass now:

- `rejects a deduped claim apply whose evidence source disappeared after creation` — creates an applied claim, then a pending duplicate whose evidence source is deleted; apply throws 409, proposal stays pending, attribute/derived counts unchanged.
- `rejects a non-materializing apply whose evidence source disappeared after creation` — pending `create_entity` whose evidence source is deleted; apply throws 409, proposal stays pending, no entity row created.

## Verification

- `bun test platform/daemon/src/ontology-proposals.test.ts`: 65 pass, 0 fail (was 63 at reviewed head; +2 regressions)
- `bun run --filter '@signet/daemon' typecheck`: pass
- `bun run --filter '@signet/daemon' build`: pass
- `git diff --check` clean; rebased onto current origin/main (`72ca35781`)

Head: `0e8e2965e322b4597fcb3d3a75cd0be4a931edaf` (force-pushed, PR branch was rebased onto live main).

PR #1491, opened 2026-08-11
fix(daemon): enforce Bun single-instance lock

### LINKED ISSUE #1448: bug(macos): daemon single-instance flock guard is a no-op under Bun — launchd respawn can start a second SQLite writer
## Summary
The daemon's single-instance guard (`daemon.lock` + `flock`) is a silent no-op under Bun: `tryLockSync` returns `true` ("allow startup") whenever `node:fs.flockSync` is missing — and it **is** missing on the runtime this daemon actually runs on.

Verified on Bun 1.3.14 (darwin arm64):
```
node:fs.flockSync -> undefined
Bun.flockSync     -> undefined
```

## Affected code
- `platform/daemon/src/daemon.ts:2512-2521` (`tryLockSync`) — fails open by design
- `platform/daemon/src/daemon.ts:1941-1950` — lock acquired before `initDbAccessorAsync`; the comment states the lock exists precisely to prevent a respawn (systemd, launchd `KeepAlive`, or `signet daemon start`) from starting a second instance that fights the first for the SQLite write lock ("SQLiteError: database is locked").

## macOS impact
After an auto-update the daemon spawns a detached replacement and exits; the LaunchAgent plist has `KeepAlive=true` (`service.ts:158-159`), so launchd immediately respawns a third instance. Nothing stops it: the flock never locks, the respawned instance opens its own write connection to the same WAL database and loops in `bindWithRetry` EADDRINUSE forever while the replacement holds the port. Two concurrent write connections → SQLITE_BUSY on `BEGIN IMMEDIATE`, duplicate background workers, WAL contention.

## Recommended fix
Fail closed instead of open:
1. Replace the flock with a portable exclusive lockfile: `openSync(lockPath, "wx")` with stale-PID detection and retry (the repo already uses this pattern at `surfaces/cli/src/lib/native-sync.ts:107-119`), or
2. `dlopen` libc and call `flock(fd, LOCK_EX|LOCK_NB)` (same pattern already used at `resource-monitor.ts:32-50`), or
3. `PRAGMA locking_mode=EXCLUSIVE` on the write connection.

At minimum, log a warning and exit when no locking primitive is available — never silently proceed.

NicholaiVogel:
## Summary

Fixes #1448.

The daemon's previous `flock` guard failed open under Bun because `node:fs.flockSync` is unavailable. This allowed launchd or another supervisor to start a second daemon and create competing SQLite writers. This PR replaces that path with a portable, fail-closed lockfile that works under Bun on Linux and macOS.

## Changes

- Add an atomic `openSync(..., "wx")` lock with owner PID recording and dead-owner reclamation.
- Keep the lock descriptor open for the daemon lifetime and remove the lock during normal process exit.
- Refuse startup when another owner is alive or lock acquisition is otherwise unavailable.
- Add a concurrent-process regression covering refusal of a second holder and recovery after the owner exits.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

Not applicable. No UI changes.

## PR Readiness (MANDATORY)

Not applicable. No migrations or durable schema changes.

## Testing

- `bun test platform/daemon/src/single-instance-lock.test.ts` passed: 1 test, 2 expectations.
- `bunx biome check platform/daemon/src/daemon.ts platform/daemon/src/single-instance-lock.ts platform/daemon/src/single-instance-lock.test.ts` passed.
- `git diff --check` passed.
- `bun run --filter '@signet/daemon' typecheck` passed after rebuilding the workspace core package.
- Full workspace test/lint commands were not run. The affected package build, typecheck, focused regression, touched-file lint, and diff checks were run.

## AI disclosure

## Notes

The lock is acquired in `main()` before database initialization, preserving the invariant established by PR #1076 while removing the Bun-incompatible flock fallback. No merge, review, or maintainer comment has been performed by this task.

### DIFF
--- a/platform/daemon/src/daemon-restart.test.ts
+++ b/platform/daemon/src/daemon-restart.test.ts
@@ -1,0 +1,9 @@
+import { describe, expect, it } from "bun:test";
+import { resolveDaemonRestartMode } from "./daemon-restart";
+
+describe("daemon update restart handoff", () => {
+	it("lets launchd KeepAlive restart after the old process releases the lock", () => {
+		expect(resolveDaemonRestartMode({ SIGNET_DAEMON_SERVICE: "launchd" })).toBe("service-manager");
+		expect(resolveDaemonRestartMode({})).toBe("replacement");
+	});
+});
--- a/platform/daemon/src/daemon-restart.ts
+++ b/platform/daemon/src/daemon-restart.ts
@@ -1,0 +1,12 @@
+/**
+ * Selects who owns the daemon restart after an update.
+ *
+ * launchd's KeepAlive starts the next process as soon as this process exits.
+ * Spawning a replacement ourselves races the single-instance lock, so the
+ * old process must exit and release the lock before launchd respawns it.
+ */
+export type DaemonRestartMode = "service-manager" | "replacement";
+
+export function resolveDaemonRestartMode(env: NodeJS.ProcessEnv = process.env): DaemonRestartMode {
+	return env.SIGNET_DAEMON_SERVICE === "launchd" ? "service-manager" : "replacement";
+}
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -7,7 +7,7 @@
 import "./bun-socket-polyfill";
 import { spawn } from "node:child_process";
 import { createHash } from "node:crypto";
-import { closeSync, existsSync, mkdirSync, openSync, readFileSync, statSync, unlinkSync, writeFileSync } from "node:fs";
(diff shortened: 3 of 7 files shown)


NicholaiVogel:
Live production profiling follow-up: #1513. The 0.199.17 daemon reproduced the startup wedge; the main thread was dominated by SQLite page reads during synchronous startup integrity recovery. #1513 includes the exact syscall evidence and source path.

PR #1503, opened 2026-08-12
fix(daemon): defer legacy VACUUM conversion

NicholaiVogel:
## Summary

Move the legacy SQLite auto-vacuum conversion out of synchronous database initialization. Startup now records durable pending state after migrations, reports readiness, and runs the full VACUUM through a single-flight post-ready worker.

## Changes

- Removed the full legacy VACUUM from `finishDbAccessorInit` while preserving migration gating.
- Added durable conversion state with pending, running, completed, and bounded failed states.
- Started the conversion worker after daemon readiness and stopped its timer during runtime shutdown.
- Added conversion status to `GET /api/pipeline/status` and documented the response.
- Added regression coverage for startup behavior, exactly-once conversion, interrupted conversion recovery, and bounded failures.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: daemon operations documentation

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

No migration file was added. The durable state table is created after existing migrations complete, so schema migration failure still blocks startup before the new maintenance state is recorded.

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A

Focused verification completed locally:

- `bun test platform/daemon/src/db-vacuum.test.ts platform/daemon/src/db-accessor.test.ts` passed: 35 tests, 68 expectations.
- `bun run --filter '@signet/daemon' typecheck` passed.
- `bunx biome check` passed for all changed TypeScript files.
- `git diff --check` passed.

The full root test, typecheck, and lint commands were not run. The existing `platform/daemon/src/pipeline/maintenance-worker.test.ts` suite has five baseline failures at `origin/main`, confirmed separately against a clean HEAD checkout.

## AI disclosure

## Notes

The conversion is still a synchronous SQLite VACUUM on the existing database connection, but it starts only after readiness is recorded and is protected by durable single-flight state. During conversion, database and event-loop latency can increase. A maximum of three attempts is persisted. Interrupted or failed attempts retry on later startup while budget remains; exhausted attempts remain visibly failed instead of creating an unbounded startup retry loop.

### DIFF
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -50,6 +50,7 @@ import {
 import { listConnectors } from "./connectors/registry";
 import { clearAllPresence, reconcileAcpDeliveries } from "./cross-agent";
 import { closeDbAccessor, getDbAccessor, getVectorRuntimeStatus, initDbAccessorAsync } from "./db-accessor";
+import { type VacuumConversionHandle, startVacuumConversionWorker } from "./db-vacuum";
 import { getQueueDiagnosticsSnapshot, getQueuePressureSnapshot } from "./diagnostics-queue";
 import { fetchEmbedding } from "./embedding-fetch";
 import { type EmbeddingIndexMigrationHandle, startEmbeddingIndexMigration } from "./embedding-index-migration";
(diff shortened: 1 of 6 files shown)


NicholaiVogel:
Live production profiling follow-up: #1513. The 0.199.17 daemon reproduced the startup wedge; the main thread was dominated by SQLite page reads during synchronous startup integrity recovery. #1513 includes the exact syscall evidence and source path.

PR #1516, opened 2026-08-12
fix(daemon): capture startup stderr in health test

### LINKED ISSUE #1515: flaky(ci): macos-15-intel source daemon exits before native /health isolation
## Summary

`Nightly Release` fails intermittently on the `macos-15-intel` native build leg in the source-daemon regression test:

```text
(fail) native embedding event-loop isolation (e2e) > /health stays within SLA while the embedding worker is stuck on a model download
error: daemon exited before health (status 1)
```

The failure occurs before the test can assert the `/health` SLA. The other native matrix legs pass.

## Evidence

- Run [31632700056](https://github.com/Signet-AI/signetai/actions/runs/31632700056), commit `b3e634718`: macos-15-intel failed; macos-14 arm64, Linux x64, Linux arm64, and Windows passed.
- Run [31633665615](https://github.com/Signet-AI/signetai/actions/runs/31633665615), commit `fb5490cd1`: the same macos-15-intel test failed with the same early-exit error; every other native leg passed.
- Rerunning run 31633665615 reproduced the failure at 19:59Z. The retry used the same `bun test src/native-embedding-eventloop-isolation.test.ts` step and failed after about two seconds.
- The same test passes locally on Linux after a clean dependency install and workspace builds: `1 pass`, `23 expect() calls`.
- The failure also appears in earlier main runs 31627690703 and 31628370221, before the two commits above. This makes a regression from the wrapper marker change in #1505 or the dashboard change in #1512 unlikely.

## Scope

This is the source-daemon test, not the compiled native runtime smoke test tracked by #964. The test intentionally starts a daemon with a native embedding configuration and redirects the model fetch to a local TCP blackhole. The logs currently expose only the child exit status, so the platform-specific daemon stderr is not included in the failure output.

Please capture the child stderr or lifecycle record on failure, then determine why the source daemon exits before binding `/health` on the Intel macOS runner. The likely class is macOS Intel startup or timing sensitivity, but the exact exit cause is not established by the current logs.

## Related

- #964 tracks a separate intermittent teardown timeout in the compiled native embedding smoke test.
- #1514 defers and bounds the startup integrity scan. Its PR check for native `/health` isolation passed, but this issue is present on main before that PR is merged.

NicholaiVogel:
## Summary

Fixes #1515. Capture the source daemon child lifecycle and stderr so early startup exits report the actual exit status, signal, and daemon diagnostics instead of only `status 1`.

## Changes

- Capture child stderr chunks and the `close` lifecycle event.
- Include captured stderr, exit status, and signal in the pre-health failure.
- Add regressions proving startup stderr and signal termination are preserved in the reported error.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: none

## Screenshots

Not applicable. No UI changes.

## PR Readiness (MANDATORY)

Not applicable. No migrations touched.

- [ ] Migration is idempotent

## Testing

- [ ] `bun test` passes
- [ ] `bun run lint` passes
- [x] N/A for the full-repo commands. Focused daemon verification is listed below.

Additional verification:

- `bun test platform/daemon/src/native-embedding-eventloop-isolation.test.ts` passes: 3 tests, 0 failures, 26 expect calls, 8.33s.
- `bun run --filter '@signet/daemon' build` passed, including daemon TypeScript compilation and dashboard build.
- `./node_modules/.bin/biome check platform/daemon/src/native-embedding-eventloop-isolation.test.ts` passed.
- `git diff --check` passed.

The targeted e2e test exercises the real source daemon and passed locally on Linux. The macOS Intel failure remains platform-specific and is the reason for the diagnostic capture. Full-repo typecheck and lint were not run because the PR only changes this daemon test; the daemon package build and focused formatter check cover the touched area.

## AI disclosure

## Notes

The diagnostic regressions launch child processes that write `startup failure` to stderr and exit with status 1, or terminate with SIGTERM. The production path drains stdout, captures stderr, waits for the close event, and reports the captured diagnostics, status, and signal when the daemon exits before `/health` becomes available.

### DIFF
--- a/platform/daemon/src/native-embedding-eventloop-isolation.test.ts
+++ b/platform/daemon/src/native-embedding-eventloop-isolation.test.ts
@@ -18,16 +18,50 @@
  */
 
 import { afterEach, describe, expect, it } from "bun:test";
-import { type ChildProcessWithoutNullStreams, spawn } from "node:child_process";
+import { type ChildProcessByStdio, spawn } from "node:child_process";
+import type { Readable } from "node:stream";
 import { mkdtempSync, readFileSync, readdirSync, rmSync, writeFileSync } from "node:fs";
 import { type Server, createServer } from "node:net";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 
+type TestChild = ChildProcessByStdio<null, Readable, Readable>;
+
 const daemonScript = join(import.meta.dir, "daemon.ts");
 const tempDirs: string[] = [];
-const children: ChildProcessWithoutNullStreams[] = [];
+const children: TestChild[] = [];
 const servers: Server[] = [];
+
+type ChildExit = {
+	readonly code: number | null;
+	readonly signal: NodeJS.Signals | null;
+};
+
+type ChildLifecycle = {
+	readonly stderr: Buffer[];
+	readonly closed: Promise<ChildExit>;
+};
+
+function captureChildLifecycle(child: TestChild): ChildLifecycle {
+	const stderr: Buffer[] = [];
+	const closed = new Promise<ChildExit>((resolve) => {
+		child.once("close", (code, signal) => resolve({ code, signal }));
+	});
+	child.stderr.on("data", (chunk: Buffer) => stderr.push(chunk));
+	return { stderr, closed };
+}
+
+function formatChildExitError(status: number | null, signal: NodeJS.Signals | null, stderr: string): Error {
+	const signalDetails = signal === null ? "" : `, signal ${signal}`;
+	const details = stderr.trim() || "<empty>";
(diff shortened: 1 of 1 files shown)


NicholaiVogel:
Follow-up: #1518 tracks the Windows-specific signal assertion failure after this change. #1519 tracks preserving daemon stdout, `.daemon/logs`, and `lifecycle.json` so the remaining macOS Intel pre-health exit can be diagnosed instead of reported as blank stderr.

PR #1584, opened 2026-08-15
fix(daemon): reap escaped ACPX children on darwin

NicholaiVogel:
## Summary

Reap escaped ACPX agent child processes on Darwin/macOS. Cleanup previously only walked the Linux `/proc` tree and early-returned on every other platform, so an ACPX agent that forked a long-lived child (e.g. a Codex ACP process) leaked it on macOS. This adds a Darwin sweep that finds candidate children by command basename and confirms ownership by matching the injected `SIGNET_ACPX_RUN_ID` in the process environment, using `launchctl procinfo <pid>` (macOS has no portable `ps -o environment=`).

## Changes

- `platform/daemon/src/pipeline/provider.ts`
  - `cleanupAcpxAgentProcesses(agent, runId)` no longer early-returns on non-Linux platforms; it now routes to the platform-specific sweep.
  - Added a `launchctl procinfo`-based environment matcher that tolerates both `NAME = value` and `NAME:value` separators so the run-id match does not silently no-op on macOS variants.
  - Candidate listing parses `ps -axo pid=,command=` output line-by-line rather than assuming positional order.
  - Added a cleanup-platform test seam (mirroring `SIGNET_ACPX_PROC_ROOT`) so the Darwin path is selectable in Linux CI.
- `platform/daemon/src/pipeline/provider.test.ts`
  - New regression test "reaps escaped Codex ACP children on darwin" driving fake `ps` + fake `launchctl procinfo` behavior, asserting our run-id child is killed and a foreign process with the same basename is left alone.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: 

## Screenshots

N/A — daemon-only change, no UI.

## PR Readiness (MANDATORY)

N/A — no migrations touched.

## Testing

- [x] `bun test` passes (provider suite 41/41)
- [x] `bun run typecheck` passes (`@signet/daemon` + root)
- [x] `bun run lint` passes (touched files)
- [ ] N/A

## AI disclosure

## Notes

Real macOS behavior is exercised on this machine; the Linux CI test validates the Darwin sweep logic through injected fake `ps` / `launchctl procinfo` commands, since the runner is Ubuntu-only. Linux `/proc` behavior is unchanged (verified by the pre-existing cleanup test in the same suite).

### DIFF
--- a/platform/daemon/src/pipeline/provider.test.ts
+++ b/platform/daemon/src/pipeline/provider.test.ts
@@ -681,6 +681,106 @@ printf 'ok\\n'
 		}
 	});
 
+	it("reaps escaped Codex ACP children on darwin via a bounded ps sweep (#1459)", async () => {
+		const root = join(tmpdir(), `signet-acpx-darwin-sweep-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+		mkdirSync(root, { recursive: true });
+		const childPid = 4242;
+		const foreignPid = 4243;
+		const runidFile = join(root, "runid.txt");
+		const fakePs = join(root, "ps");
+		const fakeProcinfo = join(root, "procinfo");
+		const bin = join(root, "fake-acpx.sh");
+		// fake ps: the full process listing for `-axo pid=,command=`. Two
+		// same-named agent processes; only one is bound to our run id.
+		writeFileSync(
+			fakePs,
+			`#!/usr/bin/env bash
+if [[ "$1" == "-axo" && "$2" == "pid=,command=" ]]; then
+  printf '%s\\n' "${childPid} /usr/local/bin/codex-acp"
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
This is a duplicate of #1583, so I am closing it in favor of that PR.

The duplicate comparison found #1583 stronger: it bounds the Darwin candidate scan, detects and logs ps/procinfo failures, and fails closed instead of silently treating unavailable checks as an empty sweep. The identical happy-path test does not distinguish those behaviors. No behavior from #1584 is being ported; #1583 is the winner and will receive the missing failure/cap regressions and checklist repair.

NicholaiVogel:
false

PR #1634, opened 2026-08-16
fix(daemon): defer startup tombstone cleanup

### LINKED ISSUE #1630: daemon startup can stall on awaited source tombstone cleanup
## Summary

`fff9c3c67e786dcb6fddc5ad45340263f2e51c0e` changed daemon startup to await `cleanupSourceDeletionTombstones()`. That cleanup now awaits `removeSourceLifecycleState()`, which uses the bounded asynchronous DB writer. A stalled writer therefore stalls the startup sequence before the daemon reaches its ready state.

Before this commit, the startup caller invoked the synchronous cleanup function and returned without waiting for the asynchronous writer queue. The behavior changed from synchronous completion to a Promise that remains pending while the writer is blocked.

## Severity

P2. A blocked or saturated async SQLite writer can keep the daemon in a degraded-ready or not-ready startup window. The source-deletion cleanup is part of `main()` before the HTTP server and workers are brought up, so this can prevent normal startup rather than merely delaying telemetry.

## Exact reproduction

Use the parent and the commit introducing the behavior:

```text
git worktree add /tmp/signet-regression-parent 8fb746f37678
git worktree add /tmp/signet-regression-head fff9c3c67e786dcb6fddc5ad45340263f2e51c0e
```

In each checkout, run a Bun probe that:

1. Creates one unconfigured `obsidian` source-deletion tombstone.
2. Initializes the daemon DB accessor.
3. Replaces `getDbAccessor().withWriteTxAsync` with an async function blocked on a Promise.
4. Calls `cleanupSourceDeletionTombstones(tempAgentsDir, () => 0)`.
5. Waits 50 ms before releasing the blocked writer and records whether cleanup settled.

The probe's relevant code is:

```ts
const accessor = getDbAccessor();
const originalAsync = accessor.withWriteTxAsync;
let release!: () => void;
const gate = new Promise<void>((resolve) => { release = resolve; });
accessor.withWriteTxAsync = async (fn) => {
  await gate;
  return originalAsync(fn);
};
const cleanup = cleanupSourceDeletionTombstones(tempAgentsDir, () => 0);
let settled = false;
Promise.resolve(cleanup).then(() => { settled = true; });
await new Promise((resolve) => setTimeout(resolve, 50));
console.log({ pendingBeforeRelease: !settled, returnsPromise: cleanup instanceof Promise });
release();
await cleanup;
```

Observed results:

```text
8fb746f37678: { pendingBeforeRelease: false, returnsPromise: false }
fff9c3c67e78: { pendingBeforeRelease: true,  returnsPromise: true }
```

The probe was run against both exact commit checkouts and exited successfully. The only behavior difference is the range change from synchronous cleanup to awaited async lifecycle-state removal.

## Expected behavior

A failure or unavailable async writer during startup source-tombstone cleanup must not prevent the daemon from continuing to its ready state. Cleanup should either remain outside the readiness-critical path or catch and defer lifecycle cleanup failures, while retaining the tombstone for a later retry.

## Introducing commit

`fff9c3c67e786dcb6fddc5ad45340263f2e51c0e` (`fix(daemon): async pipeline maintenance`)

A later follow-up, `12e07cca3`, added a try/catch around this startup await and explicitly logged that cleanup failure should continue startup. That follow-up is consistent with the failure demonstrated above.

NicholaiVogel:
## Summary

Closes #1630. A daemon startup tombstone cleanup can wait on the bounded async DB writer before HTTP binding. This PR moves that cleanup into the serialized post-ready runtime lane so the daemon can publish readiness without waiting on writer availability.

## Changes

- Remove the readiness-critical `await cleanupSourceDeletionTombstones(...)` from the pre-bind startup sequence.
- Run tombstone cleanup after deferred integrity work and before deferred pipeline startup, preserving the existing serialization and retry behavior.
- Log unexpected deferred cleanup failures loudly while retaining tombstones for retry.
- Add the #1630 blocked-writer regression probe and production-wiring assertions.
- Keep tombstoned sources out of the ready daemon's list, health, and snapshot read paths until deferred cleanup completes, while preserving the delete retry handle.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: none

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

N/A. No migrations touched.

## Testing

- [ ] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Focused proof:

- `bun test platform/daemon/src/routes/sources-routes.test.ts` passed: 40 tests, 233 assertions.
- `bun run --filter '@signet/daemon' build` passed, including daemon `tsc --noEmit`.
- `bun run lint` exited 0 with pre-existing repository warnings.
- `bun run typecheck` remains blocked by unrelated pre-existing connector errors for missing generated bundles and stale `@signet/connector-base` exports; `@signet/daemon` itself passed.
- `git diff --check` passed.

## AI disclosure

## Notes

The cleanup still runs and remains in the existing post-ready deferred lane before pipeline startup, so it does not race deferred pipeline writes. A blocked writer leaves the cleanup pending but no longer prevents HTTP readiness; releasing the writer completes the purge and removes the tombstone. Unexpected cleanup rejection is logged by the deferred startup path. Commit: `eb4cf82e5c855726bab8f4734f4792bec31936c9`.

### DIFF
--- a/platform/core/src/index.ts
+++ b/platform/core/src/index.ts
@@ -366,6 +366,7 @@ export {
 	parseDiscordSettings,
 	parseGitHubSettings,
 	removeSource,
+	removeSourceIfGeneration,
 	saveSourcesConfig,
 } from "./sources-config";
 export type {
--- a/platform/core/src/sources-config.test.ts
+++ b/platform/core/src/sources-config.test.ts
@@ -17,6 +17,7 @@ import {
 	parseDiscordSettings,
 	parseGitHubSettings,
 	removeSource,
+	removeSourceIfGeneration,
 } from "./sources-config";
 
 let dir = "";
@@ -470,6 +471,25 @@ describe("sources-config", () => {
 		if (removed.ok === false) throw new Error(removed.error);
 		expect(removed.source.id).toBe(added.source.id);
 		expect(loadSourcesConfig(agentsDir).sources).toEqual([]);
+	});
+
+	it("does not remove a re-added source when the old generation is stale", () => {
+		const agentsDir = tmp();
+		const vault = join(agentsDir, "vault");
+		mkdirSync(vault, { recursive: true });
+		const first = addObsidianSource({ root: vault, name: "Vault A", now: "2026-01-01T00:00:00.000Z" }, agentsDir);
+		expect(first.ok).toBe(true);
+		if (first.ok === false) throw new Error(first.error);
+		const removed = removeSource(first.source.id, agentsDir);
+		expect(removed.ok).toBe(true);
+		const readded = addObsidianSource({ root: vault, name: "Vault A2", now: "2026-01-02T00:00:00.000Z" }, agentsDir);
+		expect(readded.ok).toBe(true);
+		if (readded.ok === false) throw new Error(readded.error);
+
+		const stale = removeSourceIfGeneration(first.source.id, first.source.generation, agentsDir);
+
+		expect(stale).toEqual({ ok: true, removed: false, source: readded.source });
+		expect(loadSourcesConfig(agentsDir).sources).toEqual([readded.source]);
 	});
 
 	it("uses unique temp files and leaves no stale lock or tmp files after sequential mutations", () => {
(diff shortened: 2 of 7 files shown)


NicholaiVogel:
Closing as superseded: the diagnostics fix landed in #1638 (merged) and the follow-on work concluded in #1639 (merged). Nothing further to land here.

PR #1640, opened 2026-08-17
fix(secrets): require keyring native addon by absolute path in compiled builds

nyashkn:
## Problem

`signet secret exec` (and any keyring-backed secret op) fails with `Error: The native keyring module is not installed for this platform` in every distribution of the compiled binary — npm install, `signet install`, and the raw GitHub release binary, since they all ship the same Bun `--compile` artifact.

Root cause: `@napi-rs/keyring`'s own package entry does `const { createRequire } = require("node:module"); require = createRequire(__filename)`, reassigning its local `require` binding. That defeats Bun's static bundling of the package, so `require("@napi-rs/keyring")` / `import("@napi-rs/keyring")` throw `Cannot require module @napi-rs/keyring` inside the compiled binary — even though the identical call resolves fine from source. This is unrelated to sync vs async require; both fail identically once the package fails to bundle.

## Fix

Requiring a native `.node` addon by absolute filesystem path is unaffected — it's a runtime `dlopen` of a real file, not a bundle-time module resolution.

- `scripts/build-native-bun.ts`: resolve the exact target package's `.node` file (using the GNU Linux package suffix where required) (two-hop `require.resolve` through `@napi-rs/keyring`'s own install location) and embed it as a base64 runtime asset, following the existing embedded-asset pattern used for workers/wasm/templates.
- `platform/daemon/src/native-runtime-assets.ts`: add `materializeEmbeddedNativeAddon()` to write the embedded asset to a content-hashed path in `tmpdir()`, same idiom as `resolveEmbeddedWorkerPath`/`materializeEmbeddedWasmAssets`.
- Compiled-binary bootstrap (`cli-native.ts` template): materialize the addon and set `SIGNET_KEYRING_NATIVE_MODULE_PATH` to the absolute path before anything imports the addon.
- `platform/core/src/secrets-keyring.ts`: `loadModule()`/`loadModuleSync()` check that env var first and `require()` the absolute path directly, using normal package resolution only when the override is absent; malformed or missing overrides fail closed.

<details>
<summary>Why not `NAPI_RS_NATIVE_LIBRARY_PATH`?</summary>

`@napi-rs/keyring` has its own env-var override for the native library path, but it's read *inside* the package's `index.js` — which never loads in the first place inside a compiled binary. Setting that env var has zero effect. Confirmed via a minimal standalone repro (`bun build --compile` on a script doing bare `require("@napi-rs/keyring")` with the env var set — still throws).
</details>

## Test plan

- [x] Added `native-runtime-assets.test.ts` coverage for `materializeEmbeddedNativeAddon` (writes real file, returns null when asset/name absent).
- [x] Windows-safe publish retries replacement errors and preserves complete-file publication for concurrent callers; corrupted destination replacement is covered by the materialization path.
- [x] Targeted native asset suite green: `bun test platform/daemon/src/native-runtime-assets.test.ts` — 7/7 pass (33 expectations), including fenced sweep regression coverage for older/newer valid hashes plus corrupt/stray cleanup.
- [x] `@signet/core` and `@signet/daemon` production builds pass; Biome and `git diff --check` pass.
- [x] `tsc --noEmit` clean on `platform/core` and `platform/daemon`.
- [ ] Compiled-binary smoke remains unproven: no actual compiled keyring secret operation was run in this checkout; no smoke pass is claimed.
- [ ] Full platform suite not run; focused coverage and package builds were run.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) — secrets/keyring spec entries checked, no contract change, packaging-only fix
- [x] Agent scoping verified on all new/changed data queries — no data queries touched (native binary loading, no DB access)
- [x] Input/config validation and bounds checks added — override path is trimmed/empty-checked, wrapped in try/catch with fallback
- [x] Error handling and fallback paths tested (no silent swallow) — override failure is reported as unsupported without package fallback
- [x] Security checks applied to admin/mutation endpoints — no endpoints touched
- [x] Docs updated for API/spec/status changes — no public API/spec surface changed, internal build-time asset only
- [x] Regression tests added for each bug fix — `materializeEmbeddedNativeAddon` coverage includes the concurrent older/newer hash fence, corrupt prefix cleanup, and stray temp cleanup in `native-runtime-assets.test.ts`
- [x] Lint/typecheck/tests pass locally — daemon/core builds clean, targeted native asset suite 6/6, Biome clean, and `git diff --check` clean

## r6 verification

The startup sweep now treats valid hash-keyed addon files as immutable: it validates each well-formed candidate by content hash and never removes a content-valid file regardless of version. Invalid hash-keyed files and prefix strays remain eligible for cleanup. Superseded-version cleanup is intentionally deferred to a separately scoped maintenance operation; no age/version heuristic runs on startup.

- [x] r6 regression: materialize newer content A, then older content B; A remains present and valid.
- [x] r6 sweep cleanup: corrupt hash-keyed prefix files and stray temp files are removed; valid differently hashed files remain.
- [x] Local focused suite: 7/7 pass (33 expectations).
- [x] Local core and daemon typechecks, scoped Biome, and `git diff --check` pass.
- [ ] CI status not claimed here: checks were queued at update time, with known intermittent CodeQL/GitHub service outages.

### DIFF
--- a/platform/core/src/secrets-keyring.ts
+++ b/platform/core/src/secrets-keyring.ts
@@ -70,7 +70,9 @@ function tryRequireOverride(): typeof import("@napi-rs/keyring") | null {
 	// This variable is a compiled-runtime bootstrap contract. A value supplied
 	// by the launcher must never fall back to package resolution: that would
 	// make a broken compiled bundle appear healthy by loading a host addon.
-	if (!override.startsWith("/")) return null;
+	const isAbsolute =
+		override.startsWith("/") || /^\\\\[^\\]+\\[^\\]+/.test(override) || /^[A-Za-z]:[\\/]/.test(override);
+	if (!isAbsolute) return null;
 	try {
 		return require(override) as typeof import("@napi-rs/keyring");
 	} catch {
--- a/platform/core/src/secrets-keyring.ts
+++ b/platform/core/src/secrets-keyring.ts
@@ -61,16 +61,33 @@ function classifyError(error: unknown): SecretKeyringResult {
 	return { state: "corrupt", message };
 }
 
+// `@napi-rs/keyring`'s own `createRequire(__filename)` breaks Bun `--compile`
+// bundling; require the `.node` addon by absolute path instead (path set by
+// native-runtime-assets.ts in compiled builds, unset in source/dev runs).
+function tryRequireOverride(): typeof import("@napi-rs/keyring") | null {
+	const override = process.env.SIGNET_KEYRING_NATIVE_MODULE_PATH?.trim();
+	if (!override) return null;
+	try {
+		return require(override) as typeof import("@napi-rs/keyring");
+	} catch {
+		return null;
+	}
+}
+
 async function loadModule(): Promise<typeof import("@napi-rs/keyring") | null> {
-	modulePromise ??= import("@napi-rs/keyring").catch(() => null);
+	modulePromise ??= (async () => tryRequireOverride() ?? (await import("@napi-rs/keyring").catch(() => null)))();
 	return modulePromise;
 }
 
 function loadModuleSync(): typeof import("@napi-rs/keyring") | null {
 	if (syncModule !== undefined) return syncModule;
+	const override = tryRequireOverride();
+	if (override) {
+		syncModule = override;
+		return syncModule;
+	}
 	try {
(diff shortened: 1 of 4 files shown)


NicholaiVogel:
Thank you! @nyashkn

PR #1643, opened 2026-08-17
fix(tests): own and reap db-owner-worker test fixtures

### LINKED ISSUE #1641: fix(tests): db-owner-worker processes outlive their test harness and wedge the host
## Summary

`db-owner-worker` processes spawned during test runs are not reaped when the run ends. They accumulate across sessions and eventually wedge the whole host.

## Evidence (2026-08-17)

- 223 orphaned `db-owner-worker` processes found on the dev host, ages 17-21h, spawned from worktrees whose tasks had long since wrapped (including deleted worktrees).
- ~77MB RSS each = ~17GB pinned RAM plus file descriptors and SQLite lock contention.
- Symptom: the daemon/agent appears to hang on main-thread DB reads; the actual cause is host-level resource exhaustion, not the DB engine.
- Two Kanban repair runs on PR #1639 independently blocked on this ("severe host-wide DB-owner leakage from concurrent worktrees"); after killing the orphans, the same suite that had been failing for lifecycle-timeout reasons passed clean with no code change. I.e. the leak also manufactures false test failures.

## Impact

1. Host wedge under normal multi-worktree Kanban operation.
2. Flaky/false test failures attributed to the code under test (lifecycle timeouts that vanish on a clean host).
3. Cross-contamination between concurrent test runs sharing a database directory.

## Direction

Likely fix at the harness/owner layer: the spawning process must own reaping (process group kill / bounded wait on exit), tests must await closeDbAccessor/owner shutdown in teardown, and stale workers should self-exit when their parent/spawner dies (PPID watch or pipe-to-parent). A CI/host hygiene guard (fail or warn when a test run leaves live db-owner-worker processes behind) would catch regressions.

Note: operational mitigation in the meantime is `pkill -f db-owner-worker.ts` between runs.

NicholaiVogel:
Closes #1641. Implements harness-owned reaping (process-group kill + bounded wait in teardown), teardown awaits for closeDbAccessor/owner shutdown, PPID self-exit safety net, and a loud hygiene guard failing runs that leave live db-owner-worker processes. Focused suites green. (PR created by orchestrator; full wrap comment follows on the issue.)

### DIFF
--- a/platform/daemon/src/db-owner-client.test.ts
+++ b/platform/daemon/src/db-owner-client.test.ts
@@ -60,6 +60,15 @@ describe("DB owner client", () => {
 		directory = null;
 	});
 
+	function processExists(pid: number): boolean {
+		try {
+			execFileSync("kill", ["-0", String(pid)], { stdio: "ignore" });
+			return true;
+		} catch {
+			return false;
+		}
+	}
+
 	test("reaps every owner process during harness teardown", async () => {
 		const database = makeDb();
 		directory = database.directory;
@@ -67,19 +76,14 @@ describe("DB owner client", () => {
 		await client.start();
 		const ownerPid = client.health().pid;
 		if (ownerPid === null) throw new Error("owner did not publish a pid");
+		// Capture the harness spawn scope at spawn time. Do not scan the host and
+		// intersect it with health().pid after teardown: that misses any other
+		// owner this harness spawned and can never detect that leak. This scoped
+		// PID set also avoids false positives from other users' workers.
+		const spawnedPids = new Set([ownerPid]);
 		await client.close();
 		client = null;
-		let processes = "";
-		try {
-			processes = execFileSync("pgrep", ["-f", "db-owner-worker\\.(ts|js|mjs)"], { encoding: "utf8" });
-		} catch {
-			// pgrep exits 1 when teardown left no matching process.
-		}
-		const survivors = processes
-			.split("\\n")
-			.filter((line) => line.length > 0)
-			.map(Number)
-			.filter((pid) => pid === ownerPid);
+		const survivors = [...spawnedPids].filter(processExists);
 		expect(survivors).toEqual([]);
 	});
 
--- a/platform/daemon/src/db-owner-client.test.ts
+++ b/platform/daemon/src/db-owner-client.test.ts
@@ -71,6 +71,13 @@ describe("DB owner client", () => {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
R1 review follow-up:

- Fixed the hygiene guard in 9b84bde55. It now captures the harness-owned worker PID at spawn time and checks that scoped PID set after teardown. It no longer scans host workers and intersects the result with `client.health().pid`, so the assertion can detect a surviving harness child without false-positive attribution to other users' workers. The assertion is live: a surviving captured PID makes the test fail.
- The recall-lane `<200ms` assertion was not changed or widened. Five repeated runs on the review worktree measured 217.8-244.4ms and failed, while the timing test and its implementation are unchanged from `origin/main`. A direct main checkout run was blocked by missing `@signet/core` workspace resolution after bootstrap, so I did not claim a clean main comparison. This is host-load sensitivity/pre-existing test fragility, not a regression from the hygiene-guard change; please treat the existing strict budget as a separate stabilization decision rather than silently widening it.

Verification: scoped survivor test pass, Biome pass on touched files, `@signet/daemon` typecheck pass. Pushed to `signetai/fix1641-db-owner-reaping`.

NicholaiVogel:
R2 follow-up pushed in 50f2db476.

The regression test now intentionally skips client.close(), captures the spawned owner PID, asserts the live survivor is detected, and kills/awaits that PID in finally so no orphan is left behind. The processExists helper treats Linux zombies as exited.

Verification: focused survivor test passed; daemon typecheck passed; Biome passed on the touched file; full db-owner-client suite had 16/17 passing, with the known pre-existing recall timing assertion failing at 224.6ms (<200ms), matching the documented load-sensitive flake. git diff --check passed.

NicholaiVogel:
R3 follow-up: the orphan regression now exercises the survivor-scope guard itself. After intentionally skipping client.close(), the test passes the scoped owner PID set to assertNoSurvivors(), which fails loudly and includes the leaked PID; the finally block still SIGKILLs and awaits that PID.

Mutation verification: I locally disabled the guard's detection body and ran the targeted test; it went red because the expected PID-bearing failure was absent. The guard was restored afterward.

Verification: `bun test platform/daemon/src/db-owner-client.test.ts --test-name-pattern 'detects an owner survivor'` passed. The full focused file was 16 pass, 1 fail: the pre-existing recall timing test exceeded 200ms (214.76ms); all other tests passed. Production db-owner-client.ts remains byte-identical.

Issue #1642, opened 2026-08-17
[Feature]: Bot provisioning + agent-management UX: wire grounded multitenancy to per-profile Hermes Bot installs


NicholaiVogel:
Before proposing a new core surface, check whether the capability belongs in an existing route, connector, plugin, skill, or MCP server. Prefer the smallest surface that solves the problem correctly.

## Problem and use case

Signet already grounds the multitenancy primitives: `signet agent add <name> --memory isolated|shared|group [--group <g>]` (plus `remove`/`purge`/`list`), backed by the daemon's `POST/GET/DELETE /api/agents` routes that carry `read_policy`/`policy_group`. The hermes-agent connector already ships `installPlugin`/`uninstallPlugin`, `memory.provider` config patching, required-tool health checks, and `ensureNamedAgentRegistered`. So the *primitives* exist.

What does not exist is a way to **point those primitives at a Hermes Bot profile**. Hermes "Bots" are per-profile homes (`~/.hermes/profiles/<bot>/`) with their own `config.yaml`, `.env`, and plugin dir. Today, giving each Bot its own Signet memory scope means:

1. hand-copying the four plugin files (`__init__.py`, `client.py`, `plugin.yaml`, `README.md`) into each profile's provider dir, and
2. hand-editing each profile's `.env` with `SIGNET_AGENT_ID=<bot>` and its `config.yaml` with `memory.provider: signet`.

The connector keys every target to the **ambient** home — `resolveHermesHomePath()` returns `HERMES_HOME || ~/.hermes` with no profile argument (`platform/core/src/identity.ts:321`; `getHermesHome()` in the connector). There is no `--profile <bot>` targeting, no per-profile scope introspection, and **no agent-management/mutation surface at all** (no policy/group-change verb; the dashboard agents view does not surface memory policy or group).

Who's blocked: anyone running a multi-agent (Bot) Hermes install who wants shared or scoped Signet memory per agent without per-profile copy/paste.

## Proposed solution

A **Bot-provisioning** flow that reuses the existing connector primitives, scoped to a profile — not new multitenancy:

[...]

PR #1650, opened 2026-08-17
feat(agent): enforce memory policy mutations

NicholaiVogel:
## Summary
Implements issue #1642 slice 3 agent policy mutation and introspection.

- Adds daemon PATCH `/api/agents/:name` with strict centralized policy validation.
- Adds `signet agent set` and canonical `signet agent show`; retains `info`.
- Returns resolved effective scope and prevents local `agent.yaml` fallback for mutation.
- Routes existing agent creation through the same validator.
- Updates API documentation.

## POST validation decision
Changing POST validation is backward-compatible for valid requests and intentionally rejects malformed policy/group combinations that previously persisted invalid state. This is required so all daemon entry points enforce the same invariant; callers already using isolated/shared policies without a group are unchanged.

## Verification
- `bun install` passed: 3056 packages installed.
- `bun run --filter '@signet/daemon' build` passed, including `tsc --noEmit` and Vite dashboard build.
- `bun run --filter '@signet/cli' build` passed.
- `git diff --check` passed.

AI usage disclosed per `AI_POLICY.md`; commit includes Assisted-by metadata.

### DIFF
--- a/platform/core/src/agents.test.ts
+++ b/platform/core/src/agents.test.ts
@@ -1,0 +1,27 @@
+import { describe, expect, test } from "bun:test";
+import { resolveAgentMemoryPolicy } from "./agents";
+
+describe("resolveAgentMemoryPolicy", () => {
+	test.each([
+		["isolated", undefined, "agent"],
+		["shared", undefined, "global"],
+		["group", "writers", "group"],
+	] as const)("accepts %s policy", (readPolicy, policyGroup, effectiveScope) => {
+		expect(resolveAgentMemoryPolicy(readPolicy, policyGroup)).toMatchObject({
+			readPolicy,
+			policyGroup: policyGroup ?? null,
+			effectiveScope,
+		});
+	});
+
+	test.each([
+		[undefined, undefined],
+		["unknown", undefined],
+		[123, undefined],
+		["group", undefined],
+		["isolated", "extra"],
+		["shared", 123],
+	])("rejects invalid policy combinations", (readPolicy, policyGroup) => {
+		expect(() => resolveAgentMemoryPolicy(readPolicy, policyGroup)).toThrow();
+	});
+});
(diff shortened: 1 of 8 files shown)


NicholaiVogel:
Repair pass pushed in 6beae4d5e.

Findings addressed:
- HIGH POST validation bypass: present malformed read_policy values are now validated as raw input instead of being silently defaulted to isolated. policy_group is likewise validated at the route boundary.
- MEDIUM legacy-row GET failure: invalid persisted policy rows now return a structured 500 JSON error identifying the invalid persisted policy instead of escaping as an unstructured exception.
- LOW coverage: added shared policy resolver regression coverage and CLI tests for exact PATCH shape and daemon-backed show output. Existing CLI offline behavior remains covered.

Verification:
- bun install --frozen-lockfile at repository root: passed
- bun test platform/core/src/agents.test.ts surfaces/cli/src/commands/agent.test.ts: 13 passed, 0 failed
- bunx biome check on touched files: passed
- bun run --filter '@signet/core' build: passed
- bun run --filter '@signet/daemon' build: passed (daemon typecheck and dashboard build)
- bun run --filter '@signet/cli' build: passed
- git diff --check: passed

NicholaiVogel:
Implemented and pushed dc98fa29e.

Actual verification output:
- bun test platform/daemon/src/agent-routes-policy.test.ts: 3 pass, 0 fail, 15 expect() calls
- bun test surfaces/cli/src/commands/agent.test.ts: 4 pass, 0 fail, 14 expect() calls
- bun test platform/core/src/agents.test.ts: 9 pass, 0 fail, 9 expect() calls
- bun run --filter '@signet/daemon' typecheck: Exited with code 0\n- bunx biome check --write ...: Checked 1 file in 11ms; final check across 6 files: Checked 6 files in 16ms. No fixes applied.\n- git diff --check: exit 0\n\nThe route suite exercises raw POST/PATCH omission, non-string, unknown, valid, group-policy combinations, not-found, and legacy persisted invalid GET response. POST/PATCH now validate present raw fields before resolving defaults.

PR #1651, opened 2026-08-17
feat(dashboard): implement Agents view

### LINKED ISSUE #1642: [Feature]: Bot provisioning + agent-management UX: wire grounded multitenancy to per-profile Hermes Bot installs

NicholaiVogel:
## Summary
Implements the dashboard Agents view for #1642.

- Wires `/api/agents` roster loading with explicit loading, empty, and error states.
- Renders daemon-provided policy, group, and effective scope badges.
- Enables policy/group editing through `PATCH /api/agents/:name` and displays the daemon-returned effective scope after refresh.
- Enables Agents navigation only now that the view is production-wired.

## Verification
- `bun install` (repo root): passed, 3056 packages installed.
- `bun run --filter '@signet/core' build`: passed.
- `bun run typecheck` (surfaces/dashboard): passed.
- `bun test` (surfaces/dashboard): passed, 38 tests / 0 failures / 125 expect calls.
- `bun run build` (surfaces/dashboard): passed; Vite emitted the production bundle.
- `git diff --check`: passed.

Browser smoke against a running real daemon was not run in this environment. The implementation uses the same-origin dashboard API client and daemon route contract; no mocked smoke result is claimed.

## Notes
The daemon remains the authority for policy validation and effective scope resolution. The dashboard sends the requested memory policy/group and renders the returned response without inferring scope locally.

Closes #1642


NicholaiVogel:
Repair update pushed at c74582617 after rebasing onto origin/main d8750fc64.

Changes:
- kept the landed PATCH /api/agents/:name contract unchanged and verified the dashboard client sends {memory, group} with the daemon response consumed as the authoritative result
- replaced direct PATCH from the inline Confirm button with a required pre-submit confirmation dialog
- removed roster effective-scope display because GET /api/agents does not provide it; post-save status uses only PATCH-returned effective_scope

Verification:
- bun install: passed
- bun test surfaces/dashboard/src --reporter=dot: 38 pass, 0 fail, 125 expect calls
- bun run --filter signet-dashboard build: passed
- bunx biome check surfaces/dashboard/src/views/agents.tsx surfaces/dashboard/src/lib/api.ts: passed
- git diff --check: passed

A real daemon-backed browser smoke was not completed in this run; no claim is made for that gate.

NicholaiVogel:
Re-review r2 fixes pushed at d0dc7b037.

- Fixed the HIGH group -> shared/isolated transition: dashboard PATCH now sends explicit `group: null` for non-group policies, matching the daemon route contract.
- Added regression coverage for both shared and isolated transitions in `surfaces/dashboard/src/lib/api.scope.test.ts`.
- Verified: focused test 1 pass; dashboard typecheck pass; Biome on touched files pass; git diff --check pass.
- Smoke evidence: attempted the repo-native launch `SIGNET_PATH=/mnt/work/hermes-scratch/pr1651-smoke bun platform/daemon/src/daemon.ts --help`; initial run exposed stale core dist exports, then the required core build command timed out after 420s without producing a runnable daemon. No user daemon or ~/.hermes state was touched; browser smoke remains blocked by this exact build/runtime issue.

NicholaiVogel:
Repair pushed at ca0d221f5e62602c5db9453f6e3a93a6016f99f3.

Test-only fix: the fetch mock now captures the actual request body from a Request (with options.body fallback), preserving the group:null assertions for both grouped -> shared and grouped -> isolated transitions.

Verification:
- `bun test src/lib/api.scope.test.ts`: 1 pass, 0 fail
- `bun test`: 38 pass, 1 fail. The remaining failure is the same scope test receiving two undefined fetch arguments when the full dashboard suite runs; the focused test captures both actual bodies and passes. No production files changed.

NicholaiVogel:
Repaired the test-only fetch mock in `surfaces/dashboard/src/lib/api.scope.test.ts` to capture the actual options body and restore the prior global fetch after the test. Both shared and isolated transitions still assert `group: null`.

Verification at pushed head `fbd54ec32`: `cd surfaces/dashboard && bun test`
```
39 pass
0 fail
128 expect() calls
Ran 39 tests across 9 files.
```

Only the test file changed; no production changes.

NicholaiVogel:
Repaired the remaining dashboard suite-order mock failure and pushed commit 8dd724a45.

Root cause: `agent-config.test.tsx`, `sources.test.tsx`, and `dreaming.test.tsx` each installed a global `fetch` stub in `beforeAll`, then left `globalThis.fetch` as `undefined` in `afterAll`. In the full suite, that leaked module-global state into `api.scope.test.ts`; its spy therefore observed calls with an undefined request init/body. The isolated test passed, while the full suite reproduced 38 pass / 1 fail. Each polluting suite now captures the pre-test fetch and restores it in `afterAll`. The scope test uses Bun's spy/restore path and keeps both shared and isolated body assertions.

Verification, three consecutive full runs from `surfaces/dashboard`:

Run 1:
39 pass
0 fail
128 expect() calls
Ran 39 tests across 9 files. [529.00ms]

Run 2:
39 pass
0 fail
128 expect() calls
Ran 39 tests across 9 files. [513.00ms]

Run 3:
39 pass
0 fail
128 expect() calls
Ran 39 tests across 9 files. [515.00ms]

Focused `bun test src/lib/api.scope.test.ts`: 1 pass, 0 fail.

PR #1658, opened 2026-08-18
fix(cli): honor Hermes profile targets

### LINKED ISSUE #1654: regression: Hermes profile connect/disconnect ignores explicit target
## Summary

Commit `f28a8e790bb070f31af04df65c4bc6c547b01eeb` adds `connect hermes --profile` and `disconnect hermes --profile`, but the CLI does not construct the connector with the requested profile target.

## Reproduction / proof

In `surfaces/cli/src/commands/connector.ts`:

- `installHermes()` calls `new HermesAgentConnector().install(...)` and passes `profile` inside install options.
- `disconnectHermes()` calls `new HermesAgentConnector().uninstall({ profile: options.profile })`.

In `integrations/hermes-agent/connector/src/index.ts`, both operations resolve the target from `this.target` and `getHermesHome()`. The install/uninstall options are assigned to an undeclared/unused `targetOptions` field and do not set `this.target`. Therefore `this.target` remains undefined and the connector falls back to ambient Hermes home resolution.

## Impact

The explicit profile argument is ignored. A user intending to connect or disconnect one Hermes profile can instead install into or mutate the ambient/default Hermes configuration. This defeats profile isolation and can change the wrong profile's plugin, `.env`, provider config, or provider backup.

## Fix direction

Construct `HermesAgentConnector` with a validated profile target before calling install/uninstall, or make the connector's public operation options the single target-resolution path and test both CLI commands against two isolated profile homes. Add a regression test that proves the selected profile changes while the ambient profile remains untouched.

## Scope

This is a regression introduced by `f28a8e790`; no credentials or personal data are included here.

NicholaiVogel:
Fixes #1654

## Problem

`connect hermes --profile <name>` and `disconnect hermes --profile <name>` passed the profile in the install/uninstall options bag, but the connector resolved its Hermes home from its constructed target. The profile was therefore ignored and the ambient Hermes profile could be mutated.

## Approach

- Construct `HermesAgentConnector` with the validated profile target at both CLI call sites.
- Keep install/uninstall operation options for registration and policy data only; profile resolution is owned by connector construction.
- Update the existing profile ownership regressions to exercise the constructor target path.

## Verification

- `bun test integrations/hermes-agent/connector/index.test.ts`: 82 pass, 0 fail.
- `bun run --filter '@signet/connector-hermes-agent' typecheck`: pass.
- `bunx biome check integrations/hermes-agent/connector/src/index.ts integrations/hermes-agent/connector/index.test.ts surfaces/cli/src/commands/connector.ts`: pass.
- `git diff --check`: pass.
- CLI package build was attempted but is blocked by the worktree's missing workspace packages and missing `@napi-rs/keyring` declarations, unrelated to this diff.

## Risk

Scoped to Hermes connector profile target selection. Ambient installs remain unchanged; explicit profile installs and disconnects now resolve under `~/.hermes/profiles/<name>`.

Assisted-by: OpenAI Codex: gpt-5.6-luna

### DIFF
--- a/integrations/hermes-agent/connector/index.test.ts
+++ b/integrations/hermes-agent/connector/index.test.ts
@@ -1664,7 +1664,7 @@ describe("HermesAgentConnector profile ownership boundaries", () => {
 		symlinkSync(victim, join(home, ".hermes", "profiles", "escape"));
 		process.env.HOME = home;
 		process.env.HERMES_HOME = undefined;
-		await expect(new HermesAgentConnector().install(join(home, ".agents"), { profile: "escape" })).rejects.toThrow(
+		await expect(new HermesAgentConnector({ profile: "escape" }).install(join(home, ".agents"))).rejects.toThrow(
 			/(symlinked|escapes validated root)/,
 		);
 		expect(existsSync(join(victim, "plugins"))).toBe(false);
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Merged PR #1658 at `2f25f9240174e815bde957bee9b1ebd10a36164b` via exact-head rebase merge from `c0b48af63142e25244b4de3cb181928ddc411fb0`.

All checks for the merged head were green: Biome check, mandatory PR checklist validation, regression-test guard, smoke, CodeQL, version, migration, and language analysis checks.

PR #1660, opened 2026-08-18
fix(daemon): accept null policy_group on POST /api/agents

### LINKED ISSUE #1653: regression: POST /api/agents rejects explicit null policy_group
## Regression

`POST /api/agents` rejects `policy_group: null` for valid isolated/shared agents, even though the field is nullable in the persisted model and the PATCH route accepts explicit `group: null` for the same scope transitions.

## Introduced by

- SHA: `9490d93a3`
- Current release: `v0.211.0` (`017fe1203`)

## Proof

The POST validator now checks:

```ts
if (rawGroup !== undefined && typeof rawGroup !== "string") {
  return c.json({ error: "group must be a string" }, 400);
}
```

For a request such as:

```json
{"name":"worker","read_policy":"shared","policy_group":null}
```

`rawGroup` is `null`, so the route returns HTTP 400. Before `9490d93a3`, `parseOptionalString(body.policy_group) ?? null` normalized this valid nullable form and accepted it. The PATCH route explicitly permits `null`, and the dashboard sends explicit `group: null` when clearing a group.

## Blast radius

Clients that serialize optional nullable fields as `null` instead of omitting them cannot create isolated/shared agents. This includes generic JSON clients and agent-add flows that send a complete object with `policy_group: null`. The failure is a backward-compatibility break at the REST boundary.

## Fix direction

Treat `null` like omission for `policy_group` on POST, while continuing to reject other non-string values. Add a regression test covering POST with `policy_group: null` for both `isolated` and `shared` policies.

NicholaiVogel:
## Summary

Fixes #1653.

The POST `/api/agents` validator now distinguishes an omitted `policy_group` from an explicit `null`: both persist as no group, while non-null non-string, empty, and oversized values remain invalid. This restores the nullable REST contract without loosening validation.

## Evidence

- Regression matrix covers omitted, explicit null for isolated and shared policies, valid group, number, empty string, and 129-character oversize input.
- Focused daemon route suite: 38 passed, 0 failed, 82 expectations.
- Daemon typecheck: passed.
- `git diff --check`: passed.

## Checklist exception

- [x] Tests added for the regression matrix
- [x] Focused tests and typecheck passed
- [x] Biome checked touched files; existing warnings in `misc-routes.ts` remain unrelated (`Context` import and pre-existing unused `result`)

### DIFF
--- a/platform/daemon/src/agent-routes-policy.test.ts
+++ b/platform/daemon/src/agent-routes-policy.test.ts
@@ -71,6 +71,12 @@ describe("agent route memory policy contracts", () => {
 		expect((await request("PATCH", "/api/agents/omitted", { read_policy: "group", policy_group: "team" })).status).toBe(
 			200,
 		);
+		expect((await request("PATCH", "/api/agents/omitted", { read_policy: "group", policy_group: "" })).status).toBe(
+			400,
+		);
+		expect(
+			(await request("PATCH", "/api/agents/omitted", { read_policy: "group", policy_group: "x".repeat(129) })).status,
+		).toBe(400);
 		expect((await request("PATCH", "/api/agents/not-found", {})).status).toBe(404);
 	});
 });
--- a/platform/daemon/src/daemon-auth-guard-colocation.test.ts
+++ b/platform/daemon/src/daemon-auth-guard-colocation.test.ts
@@ -643,6 +643,32 @@ inference:
 			});
 			expect(readFileSync(join(tmpDir, "agent.yaml"), "utf-8")).toBe(original);
 		});
+		it("POST /api/agents distinguishes omitted, null, valid, and invalid policy_group", async () => {
+			const app = await makeApp();
+			const { registerMiscRoutes } = await import("./routes/misc-routes");
+			registerMiscRoutes(app);
+			const cases = [
+				[{ name: "agent-omitted" }, 201],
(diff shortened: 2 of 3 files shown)


NicholaiVogel:
Addressed the review finding: PATCH /api/agents now shares the same validator layer and 1-128 character bound as POST for policy_group/group values. This closes the POST/PATCH contract parity gap for oversized groups without adding route-specific duplicate checks.

Regression coverage added for PATCH with read_policy=group and policy_group="" (400), plus policy_group="x".repeat(129) (400).

Verification:
- POST policy matrix: 3 pass, 0 fail in agent-routes-policy.test.ts (omitted, invalid, valid; existing daemon colocation matrix also covers null, number, empty, oversized: 38 pass, 0 fail)
- PATCH policy matrix: 3 pass, 0 fail in agent-routes-policy.test.ts, including omitted/invalid/valid and empty/oversized regressions
- Focused daemon suites: 3 pass, 0 fail; 38 pass, 0 fail
- Daemon typecheck: pass
- Biome touched files: pass

Commit: 3a4eeafd2

NicholaiVogel:
Merged via rebase as `8b3b18b162d66d350ae73b2a2ef54bf983051dd1`. The shared bounded validator now preserves omitted/null/invalid policy_group semantics across POST and PATCH. Verification: 41/41 focused checks from r2, CI green at exact head `3a4eeafd22fa8db616c1bf4a0fffac98653e9856`, and checklist-exception applied.

Issue #1665, opened 2026-08-18
no shared embedding-provider circuit breaker: a provider outage becomes a per-file error burst and event-loop starvation instead of quiescing


NicholaiVogel:
## Summary

There is no shared embedding-provider circuit breaker. When the embedding provider is down, three independent consumers each keep issuing provider calls on their own schedules with incompatible backoff:

- the embedding-index migration worker (its own in-process 6-check gate)
- the native-source Obsidian embedding path (per-file backoff only)
- the memory-artifact reindex path (no provider backoff, and it does not consult event-loop pressure)

A broad artifact/vault sweep therefore produces a tight burst of connection-refused embed attempts, one per file, before the per-file backoff map fills in. Combined with the reindex path never yielding under pressure, this drives the `event loop critically blocked`, DB-owner deadline, and memory-growth behavior behind daemon OOM kills.

## Root cause

- `platform/daemon/src/embedding-index-migration.ts` has a local consecutive-provider-failure abort and backoff, but only for the migration worker.
- `platform/daemon/src/native-memory-sources.ts` / `obsidian-source-embeddings.ts` back off per file via an in-memory `sourceEmbeddingFailures` map. Against a sweep of thousands of changed files the effective protection is one attempt per file per sweep before the map populates, so a provider outage still produces a large error burst and repeated sweeps.
- `platform/daemon/src/memory-lineage.ts` `doReindex` scans the full canonical file list and issues batch DB writes without calling `isSystemPressureHigh` / `awaitPressureClear`, so it keeps churning while the event loop is degraded.

None of these share a single provider-down signal or retry clock, so the system cannot quiesce as a whole and self-heal when the provider returns.

## Repro

1. Configure an embedding provider and point the daemon at a workspace with a large set of native-source artifacts (tens of thousands), then stop the provider.
2. Trigger a broad reindex (e.g. any event that invalidates the artifact stat cache, or a large set of changed files).
3. Observed: a tight burst of `Embedding fetch error: Unable to connect` lines (many within milliseconds, one per file), logs of `Event loop critically blocked`, DB write queue pressure, and RSS growth that can end in the kernel OOM kill.

## Expected behavior

- A single process-wide provider-down signal that all embedding consumers consult before issuing a provider call.
- While the breaker is open, consumers skip to an "embedding pending" state instead of attempting per-file embeds, emitting at most one consolidated "provider down, retrying in Xs" line.
- The reindex path yields on event-loop pressure between batches.
[...]

PR #1666, opened 2026-08-18
fix(daemon): make embedding provider failures durable

### LINKED ISSUE #1664: embedding-index migration 'failed' state is not durable: restart resurrects it to 'building' and re-runs the full build
## Summary

When the embedding provider is unavailable, the embedding-index migration aborts to `state = 'failed'` after its consecutive-provider-failure threshold (#1160 machinery), but that terminal state is not durable. The very next daemon start resurrects it to `'building'`, deletes the staging table, and re-runs the full artifact build from scratch. On an install with a large artifact set this means the full build re-runs on every boot for as long as the provider is down, which drives event-loop stalls, DB-owner deadline failures, and memory growth.

## Root cause

`beginEmbeddingIndexBuild` in `platform/daemon/src/embedding-index-state.ts` handles three cases:

- `ready` and the configured profile already matches the active generation -> nothing to build
- `building` with the same staging fingerprint -> resume
- `building` but the config flipped back to the active profile -> abandon the build

It has no branch for `state === 'failed'`. When the persisted state is `failed`, control falls through to the unconditional block that does:

```
DELETE FROM embeddings_staging;
UPDATE embedding_index_state SET staging_profile_json = ?, state = 'building', ...;
```

So the failure written by the migration worker (`platform/daemon/src/embedding-index-migration.ts`, the `MAX_CONSECUTIVE_PROVIDER_FAILURES` abort path) is wiped on the next daemon start. Compounding this, the consecutive-failure counter is in-process only and resets to 0 on every boot, and each restart calls `beginEmbeddingIndexBuild` again which immediately returns the state to `building`. The migration can therefore never reach a stable terminal state; the full build restarts each boot and startup recovery keeps logging "Skipping staging cleanup, embedding index migration is in progress".

## Repro

1. Configure the embedding provider (e.g. Ollama) and point the daemon at a workspace with a large artifact set (tens of thousands of indexed memory artifacts).
2. Stop the provider.
3. Start the daemon and let the migration's consecutive-provider-failure abort fire so `embedding_index_state` becomes `state='failed'`.
4. Restart the daemon.
5. Observed: `state` is `'building'` again, `embeddings_staging` was dropped, and the full artifact build restarts, re-issuing an embedding call per artifact against the still-down provider.

## Expected behavior

- A `failed` state caused by provider unavailability persists across restarts with a durable retry/resume time.
- On startup, if the failure reason is provider-unavailable and the resume time has not passed, the migration stays in `failed` and does not re-run the build or re-emit the staging-cleanup loop.
- The build resumes only when the resume time passes, or when a config change actually flips the embedding profile, or via an explicit operator repair command.

## Proposed fix direction

- Persist a `next_attempt_at` (or encode it in `last_error`) when the build fails on provider unavailability.
- In `beginEmbeddingIndexBuild`, when `state === 'failed'` and the cause is provider-unavailable and `now < next_attempt_at`, do not reset to `building`.
- Keep the existing `MAX_CONSECUTIVE_PROVIDER_FAILURES` / `MAX_PROVIDER_BACKOFF_MS` constants as the backoff source.

## Notes

Reported with AI assistance. Companion issue #1665 covers the missing shared embedding-provider circuit breaker that this interacts with.

NicholaiVogel:
Fixes #1664

## Root cause
`beginEmbeddingIndexBuild` did not handle persisted `state = failed`, so startup fell through to staging deletion and restarted the full build. The provider-failure counter was also process-local.

## Fix
Persist provider-unavailable failures in `last_error` with a cause marker and `nextAttemptAt`, and keep the matching failed build durable until that time. At or after the retry time, the existing build resumes; config changes and non-provider failures retain existing retry semantics. The existing provider failure threshold and capped backoff constants are shared from the embedding-index state module.

## Regression evidence
- Focused state + migration suite: 28 pass, 0 fail, 109 expect calls.
- Regression covers restart before retry time, staging preservation, and resume at retry time.
- Biome check on touched files: pass.
- Daemon typecheck/build remain blocked by pre-existing workspace dependency/type errors: missing `@signet/lifecycle-proof` and existing dashboard Hono type mismatch.

## Companion
#1665 remains separate and is not implemented here; the shared failure threshold/backoff design remains compatible with that future circuit-breaker work.

### DIFF
--- a/platform/daemon/src/embedding-index-migration.ts
+++ b/platform/daemon/src/embedding-index-migration.ts
@@ -10,7 +10,13 @@ import {
 	type PersistedEmbeddingProfile,
 	readEmbeddingIndexState,
 } from "./embedding-index-state";
-import { beginEmbeddingIndexBuild, failEmbeddingIndexBuild } from "./embedding-index-state";
+import {
+	beginEmbeddingIndexBuild,
+	failEmbeddingIndexBuild,
+	MAX_CONSECUTIVE_PROVIDER_FAILURES,
+	MAX_PROVIDER_BACKOFF_MS,
+	persistEmbeddingIndexFailure,
+} from "./embedding-index-state";
 import { embeddingProfileFingerprint, recommendedEmbeddingProfileId, type EmbeddingRole } from "./embedding-profile";
 import { logger } from "./logger";
 import type { EmbeddingConfig } from "./memory-config";
@@ -21,13 +27,6 @@ const ACTIVE_VECTOR_TABLE = "vec_embeddings";
 const VECTOR_REBUILD_BATCH_SIZE = 50;
 const VECTOR_REBUILD_RETRY_DELAY_MS = 100;
 const MAX_VECTOR_REBUILD_RETRIES_WITHOUT_CANCELLATION = 3;
-
-// #1160: after this many consecutive provider-unavailable checks the build is
-// aborted (state='failed') instead of retrying forever; the daemon restarts
-// the build on the next config change / daemon restart.
-const MAX_CONSECUTIVE_PROVIDER_FAILURES = 6;
-// Backoff grows pollMs * 2^n between failed checks, capped at this delay.
-const MAX_PROVIDER_BACKOFF_MS = 60_000;
 
 interface ActiveEmbeddingRow {
 	readonly content_hash: string;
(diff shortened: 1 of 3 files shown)


NicholaiVogel:
Repairing the three adversarial findings from review:

1. MEDIUM restart durability: replaced the in-memory same-connection test with a file-backed SQLite scratch DB. The test closes the writer, reopens a fresh connection, and asserts failed state, staging fingerprint/row, and persisted nextAttemptAt. A second fresh connection at the deadline asserts resume to building.
2. MEDIUM load-bearing red/green: ran the same file-backed test against origin/main (57df21aef). Verbatim base result: 8 pass, 1 fail; the restart test failed with `SyntaxError: JSON Parse error: Unexpected identifier "provider"` because pre-fix code rebuilt/cleared the state before the persisted failure assertion. At corrected HEAD a061e54ca: 9 pass, 0 fail, 34 expect() calls.
3. LOW misleading title/fixture: title now explicitly says file-backed restart and the fixture uses mkdtempSync plus close/reopen.

Additional evidence:
- `bun test platform/daemon/src/embedding-index-state.test.ts`: 9 pass, 0 fail.
- `bun test platform/daemon/src/embedding-index-migration.test.ts`: 19 pass, 0 fail.
- Biome on all three changed files: pass.
- Daemon typecheck remains blocked by pre-existing baseline errors: missing @signet/lifecycle-proof declarations and the existing dashboard Hono type mismatch; no errors in the changed test.

Pushed commit a061e54ca. No merge performed.

NicholaiVogel:
Landed PR #1666 via rebase merge.

Final PR head: 82da40b31758c9ab476402ebbc1b2dbb670c9f90
Merge commit on main: f2e185cbb21a5dd6380dff403e3a2cbe557b7dde

The branch rebased cleanly onto current main before merge. Head-specific checks were green: mandatory checklist, Biome, regression guard, daemon production typecheck, async DB accessor boundary, smoke, version check, migration guard, and CodeQL analyses. Issue #1664 was closed with the defect, durable retry-state fix, file-backed restart proof, and review-chain summary.

PR #1668, opened 2026-08-18
fix(daemon): share embedding provider circuit breaker

### LINKED ISSUE #1665: no shared embedding-provider circuit breaker: a provider outage becomes a per-file error burst and event-loop starvation instead of quiescing

NicholaiVogel:
## Summary

Fixes #1665. A provider outage previously caused independent embedding consumers to keep issuing calls, creating per-file error bursts and starving the event loop.

## Changes

- Add a process-shared embedding circuit breaker using the existing retry-backoff policy and migration limits.
- Gate embedding index migration and embedding tracker provider checks through the shared breaker.
- Gate native Obsidian source embedding work and preserve the pending state while the provider is unavailable.
- Make memory-artifact reindexing honor system pressure between batches.
- Keep the API small and compatible with the #1664 durable failed-state lane.

## Testing

- `git diff --check` passed.
- Daemon TypeScript diagnostics for changed files passed; repository typecheck still reports pre-existing missing `@signet/lifecycle-proof` and dashboard type errors.
- Focused migration and memory-lineage tests were run; the suite exposed unrelated existing fixture/process-pressure failures and did not pass cleanly.

## AI disclosure

Implementation assisted by AI.

### DIFF
--- a/platform/daemon/src/embedding-circuit-breaker.test.ts
+++ b/platform/daemon/src/embedding-circuit-breaker.test.ts
@@ -1,0 +1,33 @@
+import { afterEach, describe, expect, it } from "bun:test";
+import {
+	awaitEmbeddingProviderAvailable,
+	recordEmbeddingProviderFailure,
+	resetEmbeddingCircuitBreakers,
+	shouldEmitEmbeddingProviderNotice,
+} from "./embedding-circuit-breaker";
+
+describe("embedding provider circuit breaker", () => {
+	afterEach(() => resetEmbeddingCircuitBreakers());
+
+	it("single-flights overlapping provider checks", async () => {
+		let calls = 0;
+		let release!: (available: boolean) => void;
+		const check = () => {
+			calls++;
+			return new Promise<boolean>((resolve) => {
+				release = resolve;
+			});
+		};
+		const first = awaitEmbeddingProviderAvailable("native:test", check, 10_000);
+		const second = awaitEmbeddingProviderAvailable("native:test", check, 10_000);
+		release(true);
+		expect(await Promise.all([first, second])).toEqual([{ available: true }, { available: true }]);
+		expect(calls).toBe(1);
+	});
+
+	it("emits one notice per active provider-down window", () => {
+		recordEmbeddingProviderFailure("native:test", 10_000);
+		expect(shouldEmitEmbeddingProviderNotice("native:test", Date.now())).toBe(true);
+		expect(shouldEmitEmbeddingProviderNotice("native:test", Date.now())).toBe(false);
+	});
+});
--- a/platform/daemon/src/embedding-circuit-breaker.test.ts
+++ b/platform/daemon/src/embedding-circuit-breaker.test.ts
@@ -30,4 +30,18 @@ describe("embedding provider circuit breaker", () => {
 		expect(shouldEmitEmbeddingProviderNotice("native:test", Date.now())).toBe(true);
 		expect(shouldEmitEmbeddingProviderNotice("native:test", Date.now())).toBe(false);
 	});
+
+	it("does not let a reset circuit be clobbered by a stale check", async () => {
+		let release!: (available: boolean) => void;
+		const slow = awaitEmbeddingProviderAvailable(
+			"native:stale",
+			() => new Promise<boolean>((resolve) => (release = resolve)),
+			10_000,
+		);
+		resetEmbeddingCircuitBreakers();
+		expect(await awaitEmbeddingProviderAvailable("native:stale", async () => true, 10_000)).toEqual({ available: true });
+		release(false);
+		expect(await slow).toEqual({ available: true });
+		expect(await awaitEmbeddingProviderAvailable("native:stale", async () => true, 10_000)).toEqual({ available: true });
+	});
 });
--- a/platform/daemon/src/embedding-circuit-breaker.test.ts
+++ b/platform/daemon/src/embedding-circuit-breaker.test.ts
@@ -39,9 +39,13 @@ describe("embedding provider circuit breaker", () => {
 			10_000,
 		);
 		resetEmbeddingCircuitBreakers();
-		expect(await awaitEmbeddingProviderAvailable("native:stale", async () => true, 10_000)).toEqual({ available: true });
+		expect(await awaitEmbeddingProviderAvailable("native:stale", async () => true, 10_000)).toEqual({
+			available: true,
+		});
 		release(false);
 		expect(await slow).toEqual({ available: true });
-		expect(await awaitEmbeddingProviderAvailable("native:stale", async () => true, 10_000)).toEqual({ available: true });
+		expect(await awaitEmbeddingProviderAvailable("native:stale", async () => true, 10_000)).toEqual({
+			available: true,
+		});
 	});
 });
--- a/platform/daemon/src/embedding-circuit-breaker.ts
+++ b/platform/daemon/src/embedding-circuit-breaker.ts
@@ -3,8 +3,13 @@ import { computeRetryBackoffMs } from "./embedding-repair-state";
 export const MAX_CONSECUTIVE_PROVIDER_FAILURES = 6;
 export const MAX_PROVIDER_BACKOFF_MS = 60_000;
 
-interface CircuitState { failures: number; retryAt: number; lastNoticeAt: number }
+interface CircuitState {
+	failures: number;
+	retryAt: number;
+	lastNoticeAt: number;
+}
 const circuits = new Map<string, CircuitState>();
+const inFlightChecks = new Map<string, Promise<EmbeddingProviderGateResult>>();
 
 export interface EmbeddingProviderGateResult {
 	readonly available: boolean;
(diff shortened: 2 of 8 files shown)


NicholaiVogel:
## r1 repair update

Addressed the five adversarial findings:

1. HIGH native-source integration: provider-unavailable fetch outcomes now call `recordEmbeddingProviderFailure` on the shared provider key, so native-source failures open the shared breaker.
2. HIGH `doReindex`: batch pressure yields now consult the shared breaker and stop further reindex work while the provider gate is open.
3. HIGH behavioral regressions: added focused breaker single-flight/consolidated-notice coverage; the existing native-source provider-down regression now exercises the shared failure wiring and quiescence.
4. MEDIUM concurrency: shared breaker checks now single-flight overlapping health checks and reset in-flight state safely.
5. MEDIUM notice throttle: notice emission is tied to the active failure window and resets on provider recovery.

Evidence at commit `1331c9da7`:
- `bun test platform/daemon/src/embedding-circuit-breaker.test.ts platform/daemon/src/obsidian-source-embeddings.test.ts`: 12 pass, 0 fail.
- Biome check on touched breaker/source/test files: pass.
- `git diff --check`: pass.
- daemon typecheck remains blocked by pre-existing missing `@signet/lifecycle-proof` declarations and dashboard Hono incompatibility.

NicholaiVogel:
r2 repair pushed at 293008236.

Findings addressed:
- Native-source owner path now probes the configured provider through fetchEmbedding instead of async () => true; provider failures also record the shared breaker.
- Reindex no longer claims to gate embedding work it does not perform; it retains only event-loop pressure yielding. Embedding-producing workers remain breaker-gated.
- Added generation-fenced stale-check regression and single-flight coverage.
- shouldEmitEmbeddingProviderNotice is now production-wired through the breaker callback in native source, tracker, and index migration, giving one consolidated notice per active down window.
- Obsidian provider-down regression remains green and verifies burst quiescence at the consumer boundary (one provider attempt, subsequent calls skipped).

Evidence:
- bun test platform/daemon/src/embedding-circuit-breaker.test.ts platform/daemon/src/obsidian-source-embeddings.test.ts: 13 pass, 0 fail.
- git diff --check: pass.
- daemon typecheck remains blocked by pre-existing missing @signet/lifecycle-proof declarations and dashboard Hono incompatibility.

No merge performed.

NicholaiVogel:
R3 repair pushed in b3269de0c.

Quoted r3 findings and fixes:

1. HIGH, doReindex did not respect the shared circuit. Fixed memory-lineage.ts to derive the configured provider key, consult awaitEmbeddingProviderAvailable(providerKey, undefined, ...), stop scanning when the shared circuit is open, and skip pending upsert flushes while open. This uses no fake provider probe callback.

2. HIGH, native-source failures did not emit notices at the real recording points. Fixed both owner and legacy source embedding failure paths to call shouldEmitEmbeddingProviderNotice(providerKey) after recordEmbeddingProviderFailure and emit the consolidated consumer logger warning.

3. HIGH, production consumer evidence. The existing provider-down Obsidian regression remains green and now exercises the shared breaker through the real consumer path; the focused consumer output shows one warning for the provider-down window and the second call performs no provider fetch. The breaker suite remains green with single-flight, one-notice, and stale-reset coverage.

4. MEDIUM, Biome. Ran biome --write and a clean biome check across all seven touched files: exit 0.

Test evidence:
- bun test platform/daemon/src/embedding-circuit-breaker.test.ts platform/daemon/src/obsidian-source-embeddings.test.ts: 13 pass, 0 fail.
- memory-lineage.test.ts was run separately: 11 pass, 19 fail, 15 errors. The failures reproduce the documented pre-existing Promise-vs-number, closed-DB, undefined lock, event-loop, and timeout/environment failures; reindex incremental cases that execute remain green. No attempt was made to hide these failures.
- git diff --check: pass.

No merge requested. No release action taken.

NicholaiVogel:
R4 findings addressed in d81db28eb:

- HIGH doReindex pre-open circuit: added an initial non-probing awaitEmbeddingProviderAvailable gate before listCanonicalFiles/scanning or accumulation. A seeded open circuit sets breakerOpen and returns before any scan or final flush. Existing final upsert flush remains gated by !breakerOpen.
- HIGH consumer regression coverage: extended the provider-down Obsidian production-caller regression to assert the shared circuit is open through awaitEmbeddingProviderAvailable after the failure path at obsidian-source-embeddings.ts:351-360. The existing second-call path exercises the consolidated backoff behavior.

Verification:
- bun test platform/daemon/src/embedding-circuit-breaker.test.ts platform/daemon/src/obsidian-source-embeddings.test.ts: 13 pass, 0 fail, 51 expect() calls.
- bunx biome check platform/daemon/src/memory-lineage.ts platform/daemon/src/obsidian-source-embeddings.test.ts: clean.
- git diff --check: clean.

The requested dedicated seed-open doReindex integration test and many-file notice sweep are not added in this round; the implementation fix and shared-circuit assertion are pushed for review.

NicholaiVogel:
Added the two requested load-bearing regression tests in e30eaf1d6:

- `memory-lineage.test.ts`: seeds the shared provider circuit open, runs a small (<50 item) reindex, and asserts no artifact upserts are flushed.
- `obsidian-source-embeddings.test.ts`: sweeps five provider-failing files and asserts one consolidated provider-unavailable notice.

Verification at head:
```
$ bun test platform/daemon/src/memory-lineage.test.ts -t 'halts a small scan'
1 pass, 0 fail

$ bun test platform/daemon/src/obsidian-source-embeddings.test.ts
11 pass, 0 fail

$ bun test platform/daemon/src/embedding-circuit-breaker.test.ts
3 pass, 0 fail

$ bunx biome check platform/daemon/src/memory-lineage.test.ts platform/daemon/src/obsidian-source-embeddings.test.ts
Checked 2 files in 25ms. No fixes applied.
```

Base-red capture for the seeded-open regression (`origin/main` memory-lineage implementation with the new test retained):
```
BASE_RC=1
Expected: 0
Received: 2
(fail) memory-lineage > reindexMemoryArtifacts incremental > halts a small scan before accumulating or flushing when the shared provider circuit is already open
```

NicholaiVogel:
Merged as 128e6582d5067530aa2c4b3fe2cd9596c3af6ed6 after five adversarial review rounds. Closes #1665.

PR #1673, opened 2026-08-18
fix(daemon): handle degenerate statfs block sizes

### LINKED ISSUE #1672: Daemon bricked at startup: statfsSync returns bsize: 0, space preflight misreads it as "disk full"
# Daemon bricked at startup: `statfsSync` returns `bsize: 0`, space preflight misreads it as "disk full"

## Environment
- macOS (darwin), Apple Silicon
- Signet CLI/daemon `v0.211.7` (nightly), Bun-compiled binary at `~/.local/bin/signet`
- Memory DB: `~/.agents/memory/memories.db`, size **330,649,600 bytes (330 MB)**
- Disk: `df -H /` reports **807 GB free** of 1000 GB (2% used). Disk is NOT full.

## Symptom
`signet daemon restart` (and startup) fails:

```
[db-accessor] sqlite-vec extension not found — vector search disabled   (xN, non-fatal warn)
✖ Failed to restart daemon
```

The actual fatal error, from `~/.agents/.daemon/logs/signet-<date>.log`:

```
"Fatal error"  name=DbSpacePreflightError
"[migration_backup] migration backup blocked: insufficient disk space.
 Database size: 330649600 bytes; free: 0 bytes; required: 330649600 bytes.
 Free disk space and retry."
```

Stack: `preflightMigrationBackupSpace` → `prepareMigrationBackup` → `backupBeforeMigrationAsync` → `initDbAccessorAsync` → `main3`.

Note: the `sqlite-vec extension not found` lines are `console.warn` (non-fatal) and are NOT the cause of the brick. The cause is the `DbSpacePreflightError` below.

## Reproduction (the space measurement)
The space-availability computation multiplies two `statfs` fields:

- `platform/daemon/src/db-accessor.ts`, `availableBytes()`:
  `const stat = deps.statfsSync(path); return stat.bavail * stat.bsize;`
- `preflightMigrationBackupSpace()`: `freeBytes = availableBytes(dirname(dbPath), deps)`, then
  `if (freeBytes < dbBytes) throw new DbSpacePreflightError("migration_backup", metrics);`

Probing the **same Bun runtime** the daemon uses, against the real paths:

```
$ bun run /tmp/statfs_probe.js
root   : {"bavail":244199454,"bsize":0,"free":0}
home   : {"bavail":244199454,"bsize":0,"free":0}
memory : {"bavail":244199454,"bsize":0,"free":0}
```

`bavail` is populated (a plausible block count) but **`bsize` is 0**. The product
`bavail * bsize` is therefore `0`, and that `0` is reported verbatim as `free: 0 bytes`.
The disk is not full; a statfs field is being populated with a degenerate value, and the
space computation trusts it.

## Root cause (category error)
The check conflates two distinct conditions:

1. **"The disk is genuinely full"** — a verified measurement. Fail-closed is correct here.
2. **"The space measurement returned a degenerate/broken value"** — a measurement failure.

`preflightMigrationBackupSpace` treats condition (2) as if it were condition (1): a
zero produced by `bavail * bsize` with `bsize=0` is indistinguishable in the code from a
real full disk, so it throws the fatal error and aborts startup.

The existing escape hatch does not cover this: the write-probe fires only when
`freeBytes === 0 && dbBytes <= MAX_ZERO_FREE_SPACE_PROBE_BYTES` (1 MiB). The memory DB is
330 MB, so the probe never runs and the degenerate-0 is taken at face value.

## What the fix must NOT be (no bandaid)
Reject, in particular:
- `if (freeBytes === 0) skip the check` — masks a broken measurement; a genuinely full
  disk would then be backed up and silently fail mid-copy.
- Hardcoding an assumed-free constant, or special-casing darwin/Bun.
- Widening the existing size threshold — that tunes a knob; it does not fix the measurement.

## Required root-cause fix
1. **`availableBytes` must not multiply a statfs field that can be degenerate.**
   Validate the reading: if `bsize` is not a positive finite number (or the product is
   degenerate), do not return `bavail * bsize`. Return `null` ("could not measure") or
   compute free space from an independent capacity source (`df` / `stat -f` / a
   platform-appropriate syscall). Never fabricate a zero from a broken field.
2. **`preflightMigrationBackupSpace` must distinguish "could not measure" from "measured full."**
   A `null`/degenerate measurement is a different outcome than a verified-full reading.
   Only throw `DbSpacePreflightError` when a *verified* measurement says full.
3. **Generalize the write-probe to be the authority on measurement failure.**
   Gate the actual-size write-probe on measurement degeneracy, not on
   `dbBytes <= 1 MiB`. Whenever the statfs reading is untrustworthy, measure reality by
   attempting the required-size write — regardless of DB size.

## Upstream angle (secondary)
The degenerate `bsize: 0` (with a populated `bavail`) looks like a Bun `statfsSync`
field-population defect on darwin, or a macOS statfs field that Bun propagates. Worth a
separate Bun-side report. Independently, application code should not be brittle to a
runtime statfs quirk — the Signet-side fix above is required regardless of where the
`bsize=0` originates.

## Expected outcome after fix
`signet daemon restart` succeeds; the space gate either passes (disk has 807 GB free)
or correctly reports a verified-full condition. Vector-search degradation (the
non-fatal `sqlite-vec` warn) is a separate matter and is not what bricks the daemon.

NicholaiVogel:
Fixes #1672

## Summary
- Treat non-positive, non-finite, or degenerate `statfsSync` values as unmeasurable instead of fabricating zero available bytes.
- Use the required-size write probe whenever space measurement is untrustworthy, regardless of database size.
- Preserve fail-closed behavior when a verified statfs reading reports insufficient space.

The key distinction is between a verified-full measurement and a measurement failure. A `bsize: 0` reading with populated `bavail` is not evidence that the disk is full, so startup proceeds only after the actual required write probe succeeds.

## Test matrix
- Degenerate `{ bavail: 244199454, bsize: 0 }` with a large database: write probe runs and backup proceeds.
- Verified-full `{ bavail: 0, bsize: 4096 }`: `DbSpacePreflightError` remains enforced.
- Existing small zero-space probe behavior now uses the same degenerate-measurement path.
- Focused daemon suite: 35 passed, 0 failed.

## Verification
- `bun test platform/daemon/src/db-accessor.test.ts`
- `bunx biome format --write platform/daemon/src/db-accessor.test.ts`
- `git diff --check`

The observed `bsize: 0` behavior may also warrant a separate Bun/macOS upstream report; this PR keeps Signet resilient regardless of the runtime cause.

Checklist exception: requested by task coordination; no merge performed.

### DIFF
--- a/platform/daemon/src/db-accessor.test.ts
+++ b/platform/daemon/src/db-accessor.test.ts
@@ -482,6 +482,32 @@ describe("DbAccessor", () => {
 		expect(operations).toEqual(["probe", "unlink", "backup"]);
 	});
 
+	test("does not fabricate free space when the degenerate write probe fails", () => {
+		const dbPath = tmpDbPath();
+		cleanupDirs.push(join(dbPath, ".."));
+		writeFileSync(dbPath, "database");
+
+		let error: DbSpacePreflightError | undefined;
+		try {
+			backupBeforeMigration({ exec: () => {} }, dbPath, 65, {
+				copyFileSync: () => {
+					throw new Error("probe failed");
+				},
+				readdirSync: () => [],
+				statSync: () => ({ mtimeMs: 0, size: 1024 }),
+				statfsSync: () => ({ bavail: 1, bsize: 0 }),
+				unlinkSync: () => {},
+				now: () => 7000,
+				log: () => {},
+			});
+		} catch (err) {
+			error = err as DbSpacePreflightError;
+		}
+		expect(error).toBeInstanceOf(DbSpacePreflightError);
+		if (!error) throw new Error("expected DbSpacePreflightError");
+		expect((error.metrics as unknown as { freeBytes: number | null }).freeBytes).toBeNull();
+	});
+
 	test("still blocks a genuinely verified-full migration backup", () => {
 		const dbPath = tmpDbPath();
 		cleanupDirs.push(join(dbPath, ".."));
--- a/platform/daemon/src/db-accessor.test.ts
+++ b/platform/daemon/src/db-accessor.test.ts
@@ -442,7 +442,7 @@ describe("DbAccessor", () => {
 			},
 			readdirSync: () => Array.from(files.keys()),
 			statSync: (path) => ({ mtimeMs: files.get(String(path).slice(dbDir.length + 1)) ?? 0, size: 8 }),
-			statfsSync: () => ({ bavail: 0, bsize: 1 }),
+			statfsSync: () => ({ bavail: 0, bsize: 0 }),
 			unlinkSync: (path) => {
 				const name = String(path).slice(dbDir.length + 1);
 				operations.push(`unlink:${name}`);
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Closed both adversarial review findings from review-1673-findings.md.

- HIGH rebase: rebased signetai/fix-1672-statfs-bsize onto current origin/main. New HEAD: a6c2f5ef0. `git diff origin/main..HEAD` now contains exactly the intended two files: platform/daemon/src/db-accessor.ts and platform/daemon/src/db-accessor.test.ts.
- MEDIUM metrics: the degenerate statfs/write-probe failure now preserves `freeBytes: null` in the DbSpacePreflightError metrics payload instead of fabricating zero. Added a regression assertion for the unmeasured state.
- Verification: `bun test platform/daemon/src/db-accessor.test.ts` => 36 pass, 0 fail, 65 expect() calls. `bunx biome check platform/daemon/src/db-accessor.ts platform/daemon/src/db-accessor.test.ts` => no errors; one pre-existing unused-function warning for resolveVecEmbeddingDimensions. `git diff --check` clean.
- `checklist-exception` label is present on the PR.

NicholaiVogel:
Steward closeout: PR #1673 merged via rebase as `cd7195a00aa34ac8c5656bffd74929601c66a462`.

Verified before merge: exact head `a6c2f5ef09f33f47acd477a9ba4f82018958fbf4`, open PR with `checklist-exception`, all 19 head-specific checks completed successfully, clean merge-tree against current `main`, and exactly the two intended daemon accessor files in the PR range. Issue #1672 was closed with the defect, fix, review-chain, and Bun upstream context documented.

PR #1677, opened 2026-08-18
fix(daemon): create db owner parent directory

### LINKED ISSUE #1678: bug(daemon): release-native daemon startup crash — DB-owner child dies with SQLITE_CANTOPEN (missing memory directory)
## Symptom
Release-native daemon startup fails before `/health`: the spawned SQLite DB-owner child exits (`DbOwnerDiedError`) in the compiled smoke suite (`scripts/native-embedding-runtime-smoke.test.ts`, `SIGNET_DB_OWNER_SMOKE=1`), blocking the release train.

## Root cause
Daemon startup created `.daemon` and log directories but not the `memory/` parent directory (`dirname(MEMORY_DB)`). The child therefore failed with `SQLITE_CANTOPEN` before readiness. Direct worker probes did not reproduce it because their temporary DB directories already existed — only the release-equivalent compiled path hit it.

## Evidence
- Compiled smoke before fix: 6 pass / 4 fail (all four daemon-start tests exit pre-`/health`).
- Captured child stderr (previously discarded) shows `SQLITE_CANTOPEN`.
- After creating the parent directory before spawn: 9 pass / 1 fail (remaining failure is a separate post-readiness `signet sync` non-exit, tracked separately).

## Fix direction
Create `dirname(MEMORY_DB)` before spawning the DB owner; propagate the child's trailing stderr into death/transport/timeout diagnostics.

NicholaiVogel:
## Summary
Fix release-native daemon startup failures where the DB-owner child exited before readiness.

## Root cause
Daemon startup created `.daemon` and log directories but not the `memory/` parent directory. The spawned SQLite DB-owner child therefore failed with `SQLITE_CANTOPEN` before `/health`; direct worker probes did not reproduce it because their temporary DB directories already existed.

## Changes
- Create `dirname(MEMORY_DB)` before spawning/starting the DB owner.
- Capture the child's last 8 KiB of stderr and include it in DB-owner death, transport, and startup-timeout diagnostics.
- Preserve the existing worker environment candidate only as-is; the verified fix is directory creation.

## Verification
- `bunx biome check platform/daemon/src/db-owner-client.ts platform/daemon/src/daemon.ts` passed.
- Exact compiled smoke: 9 passed, 1 failed. All four prior DB-owner daemon-start failures passed after the fix. The remaining failure occurred after `Daemon ready` in the install/sync subprocess and timed out at 30 seconds; it is unrelated to DB-owner startup.

## Type
- [x] Bug fix

## Checklist
- [x] Tests or evals added/updated where appropriate
- [x] Documentation updated where appropriate

Fixes #1678

### DIFF
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -2004,6 +2004,7 @@ async function main() {
 
 	mkdirSync(DAEMON_DIR, { recursive: true });
 	mkdirSync(LOG_DIR, { recursive: true });
+	mkdirSync(dirname(MEMORY_DB), { recursive: true });
 
 	// Acquire an exclusive lock to prevent multiple daemon instances from
 	// competing for the SQLite write lock. Without this, a respawn (systemd,
--- a/platform/daemon/src/db-owner-client.test.ts
+++ b/platform/daemon/src/db-owner-client.test.ts
@@ -317,6 +317,24 @@ describe("DB owner client", () => {
 		expect(rows).toEqual([{ id: "m1" }]);
 	});
 
+	test("retires a broken stdin without starving the queue in a retry loop", async () => {
+		const database = makeDb();
+		directory = database.directory;
+		const workerPath = join(database.directory, "broken-stdin-worker.js");
+		await Bun.write(
+			workerPath,
+			'process.stdout.write(JSON.stringify({ type: "ready", pid: process.pid }) + "\\n"); setTimeout(() => process.stdin.destroy(), 25);',
+		);
+		client = createDbOwnerClient({ dbPath: database.path, workerPath });
+		await client.start();
+		const handle = client.submit(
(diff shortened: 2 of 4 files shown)


NicholaiVogel:
Round 5 verification notes (orchestrator):

- The write-rejection handler now calls `retireOwner(...)` synchronously (the exact fix r4's finding prescribed), so the `child === owner` guard falls through and `dispatch(jobId)` requeues the job onto the fresh owner. Verified by diff inspection at 72eac410f plus the full focused suite (17 pass / 1 pre-existing saturation flake).
- Dedicated regression test for this specific path was attempted and withdrawn: a destroyed-peer stdin does not deterministically produce a write-callback error on Linux (the write buffers and reports OK; only the deadline backstop fires). The existing SIGABRT-recovery test covers write rejection against a dead owner; the alive-owner transport-error path is verified by inspection. Injecting a `write()` seam solely for this test was judged out of scope for this PR.

PR #1682, opened 2026-08-18
fix(daemon): degrade when sqlite-vec load fails

### LINKED ISSUE #1681: bug(daemon): macOS DB owner dies loading sqlite-vec (bun sqlite3 lacks extension support) — daemon fatal on startup
## What

Release runs 32183044773 (a1a9f7c62) and 32184801343 (410f455f7) fail the "Native embedding /health isolation (source daemon)" step on BOTH macOS legs (macos-14/arm64, macos-15-intel/x64). Linux legs pass.

DB owner child exits code 1 at startup:

```
platform/daemon/src/db-owner-worker.ts:82
  db.loadExtension(vecExtension);
error: This build of sqlite3 does not support dynamic extension loading
```

The daemon treats owner death at startup as fatal (`error:startup; shutting down`), so the daemon never becomes healthy on macOS.

## Why it surfaced now

Previously masked: before the memory-dir mkdir fix (#1677), the owner died earlier with SQLITE_CANTOPEN and never reached the loadExtension call. The mkdir fix exposed this next failure in the chain. Latent since the owner-worker introduced the loadExtension call; macOS bun builds compile sqlite3 without dynamic extension loading.

## Impact

Any macOS user starting the new daemon (release-native or source) hits a fatal startup crash once #1677 ships. KNN recall is affected by any fallback, but a degraded-but-alive daemon beats a fatal one (same doctrine as #1543).

## Suggested direction (read the code first)

`db-owner-worker.ts` already has a null path (`if (vecExtension !== null)`). Make extension-load failure degrade to that path instead of killing the owner: catch the load error, log a warning (include platform + bun sqlite3 capability), continue without vec. Do NOT widen timeouts or skip tests. A capability probe in `findSqliteVecExtension` (return null when the runtime cannot load extensions) is an acceptable alternative if cleaner.

## Test strategy

Cannot repro on Linux CI directly (Linux bun loads extensions fine). Add a unit test that forces the failure path (e.g. make `findSqliteVecExtension` return a path that fails to load, or mock the capability probe) and assert the owner still reaches ready and serves queries. The macOS release legs then verify end-to-end.

NicholaiVogel:
## Summary

Fixes #1681. The DB owner no longer dies when Bun's SQLite build cannot dynamically load sqlite-vec, which lets macOS daemon startup continue with degraded KNN recall.

## Changes

- Catch sqlite-vec load failures in the DB owner worker.
- Log the platform, load error, and explicit KNN degradation warning.
- Keep the database open and serving ordinary queries when vec is unavailable.
- Add regression coverage for a failed extension load followed by owner readiness and a plain query.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: none

## Screenshots

N/A. No UI changes.

## PR Readiness (MANDATORY)

Not applicable. No migrations touched.

## Testing

- [ ] `bun test` passes
- [x] `bun run lint` passes for touched files via `bunx biome check platform/daemon/src/db-owner-worker.ts platform/daemon/src/db-owner-client.test.ts`
- [ ] N/A

Focused test result: 17 passed, 1 pre-existing timing-sensitive failure in `recall lane completes while maintenance lane is saturated` (received 214-223ms against a 200ms threshold). The new regression test passes.

Daemon build was attempted and is blocked by the existing missing workspace package `/home/nicholai/signet/signetai/node_modules/@signet/lifecycle-proof`.

## AI disclosure

## Notes

The fallback is intentionally narrow: only extension initialization is degraded. Database open, owner readiness, and ordinary SQL service continue normally. KNN-dependent operations remain unavailable and are explicitly called out in the warning log.

### DIFF
--- a/platform/daemon/src/db-owner-client.test.ts
+++ b/platform/daemon/src/db-owner-client.test.ts
@@ -16,6 +16,7 @@ import {
 import { findSqliteVecExtension } from "@signet/core";
 import { closeDbAccessor, initDbAccessor } from "./db-accessor";
 import { recallThroughDbOwner } from "./db-owner-recall";
+import { loadSqliteVecIfAvailable } from "./db-owner-worker";
 
 function makeDb(): { readonly directory: string; readonly path: string } {
 	const directory = mkdtempSync(join(tmpdir(), "signet-db-owner-"));
@@ -77,6 +78,23 @@ describe("DB owner client", () => {
 			throw new Error(`DB owner survivor(s) detected: ${survivors.join(", ")}`);
 		}
 	}
+
+	test("keeps the owner alive when sqlite-vec cannot be loaded", async () => {
+		const database = makeDb();
+		directory = database.directory;
+		const sqlite = new Database(database.path);
+		expect(loadSqliteVecIfAvailable(sqlite, join(database.directory, "missing-vec-extension"))).toBe(false);
+		sqlite.close();
+		client = createDbOwnerClient({ dbPath: database.path });
+		await client.start();
+		await expect(
+			client.submit<readonly { readonly value: number }[]>(
+				{ kind: "query", statement: { sql: "SELECT 1 AS value", result: "all" } },
+				{ operation: "vec-degraded-plain-query", lane: "read", deadlineMs: 1_000 },
+			).result,
+		).resolves.toEqual([{ value: 1 }]);
+		expect(client.health().state).toBe("ready");
+	});
 
 	test("detects an owner survivor when harness teardown is skipped", async () => {
 		const database = makeDb();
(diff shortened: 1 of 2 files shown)


NicholaiVogel:
Merged by orchestrator. Review chain: r1 verified mechanism + caught non-load-bearing test; r2 replaced it with a real end-to-end test (SIGNET_VEC_PATH override, red on main / green with fix — orchestrator-verified at both revisions). Unblocks the release train; next release run must go green on all five platforms.

Issue #1671, opened 2026-08-18
bug(native-sources): unbounded source synchronization remains coupled to the daemon event loop


NicholaiVogel:
## Summary

The native-memory bridge still couples broad source synchronization to the daemon runtime. It starts after readiness, performs repeated full-tree scans, and performs database/indexing work from the same execution environment that serves HTTP. The filesystem traversal is written with async APIs, but that does not isolate the per-file SQLite, graph, cleanup, and embedding operations that follow it.

This is not a request for another scan throttle or another yield. The scanner needs a durable worker/process architecture with bounded streaming, resumable progress, and an explicit database execution boundary.

## Reproduction evidence

Observed with Signet `0.211.7` against a production-scale database containing approximately 103k memories and 98k embeddings:

- the daemon logged `Daemon ready`;
- the HTTP listener opened;
- CPU then ramped toward 100%;
- `/health/live` and `/api/status` stopped responding;
- live tracing showed continuous random-offset 4 KiB `pread64` calls against the SQLite database;
- the parent daemon thread was inside native `pread64` when inspected with `gdb`.

A long Bun CPU profile from the same runtime family included `runScan` and `syncExisting` among the named application functions, along with glob matching and synchronous filesystem/resource operations. The direct native stack does not expose a useful Bun symbol, but it establishes that the daemon is spending the failure window in native database reads, not waiting on an HTTP client or merely rendering the dashboard.

Disabling the configured Obsidian source did not prevent the reproduction. The built-in Codex, Claude Code, and Hermes sources, database initialization, or their shared indexing path can therefore trigger the same class of failure. The Obsidian source is not the root cause by itself.

## Code path and current behavior

The relevant implementation is in `platform/daemon/src/daemon.ts` and `platform/daemon/src/native-memory-sources.ts`.

The daemon schedules `startNativeMemoryBridge(...)` after readiness and immediately calls `syncExisting()`. The bridge is configured with a ten-second polling interval. The bridge currently:

1. builds the active source list, including built-in sources and enabled configured sources;
2. recursively walks source trees;
3. performs a full counting/path-index pass;
4. performs another full indexing pass;
5. matches paths against glob patterns, including recursive `**` matching;
6. reads and fingerprints files;
7. performs artifact existence, graph, embedding, cleanup, and indexing operations through the database layer;
8. repeats the scan when a resync is requested or the polling interval fires.

[...]

PR #1688, opened 2026-08-19
test(daemon): Phase D stability acceptance harness

### LINKED ISSUE #1543: rearchitect daemon for event-loop isolation: worker-pool off-main-thread execution, scale to 10M+ memories without hitching
Rearchitect the daemon around event-loop isolation: no unbounded synchronous IO, native work, or SQLite work on the main thread. Target: a server that serves 10,000,000+ memories without hitching on startup, recall, ingestion, or background maintenance.

## The architectural defect

The recent wedge fixes are converging on one missing invariant: the daemon has no enforced boundary between asynchronous scheduling and synchronous execution. A Promise, a queue, or a `setImmediate` yield does not make the work off-main-thread. The current database accessor makes this explicit:

- `platform/daemon/src/db-accessor.ts:156-188` exposes synchronous `withWriteTx` and `withReadDb` as the normal primitives. `withWriteTxAsync` is optional and is documented as something callers "should prefer", not an invariant enforced by the accessor.
- `platform/daemon/src/db-accessor.ts:1183-1195` executes `BEGIN IMMEDIATE`, the callback, and `COMMIT` synchronously on the singleton write connection.
- `platform/daemon/src/db-accessor.ts:1217-1256` bounds admission to 64 queued jobs and yields between jobs, but `drainWriteQueue` still calls `job.run()` synchronously. The queue limits how much work is waiting; it does not move the work off the event loop or bound the duration of one job.
- `platform/daemon/src/db-accessor.ts:1146-1161` throws when 16 read connections are in use instead of applying backpressure through an awaitable pool. That turns load into request failure, while the underlying query still runs synchronously on whichever event-loop callback acquired the connection.

This is the direct code smell behind the recurring incidents: the code calls these paths async because the caller can await the result, while the expensive native operation remains synchronous. The architecture therefore has neither worker ownership nor a hard execution budget.

## Direct code evidence

1. Startup still performs collection-size-dependent synchronous work.

`initDbAccessorAsync` is not an asynchronous SQLite initialization boundary. In `platform/daemon/src/db-accessor.ts:795-799`, it opens the connection and then calls the synchronous `finishDbAccessorInit`. That function runs migrations, FTS repair/backfill, vector-table setup, and vector backfill synchronously (`:842-900`). The FTS recovery path can execute `INSERT INTO memories_fts ... SELECT ... FROM memories` over the whole collection (`:938-956`). Vector recovery selects all missing rows and inserts them in one synchronous transaction (`:1019-1115`), with a comment recording that one anti-join already cost about 6.5 seconds on a 43k-row local database (`:1023-1028`). A 10M-memory workspace turns this from an edge case into a guaranteed startup hazard.

`platform/daemon/src/startup-recovery.ts:10-17,42-59,68-83` is even more explicit: startup recovery is "fully synchronous", must not yield, and drains up to 50,000 dead jobs and 200,000 staging rows in synchronous batches. The batch size bounds SQL result size, but there is no wall-clock deadline, worker boundary, or loop-budget check. "Bounded batch" is not the same as bounded event-loop occupancy.

2. The native source scanner is only half-isolated.

`platform/daemon/src/native-memory-sources.ts:696-724` correctly uses async filesystem reads, but the scan then crosses back into synchronous SQLite and indexing work for every file. `nativeArtifactContentHash`, `nativeArtifactCapturedAt`, graph existence checks, embedding existence checks, and artifact cleanup use `getDbAccessor().withReadDb` or `withWriteTx` (`:542-694, 787-886, 905-981`). `indexExternalMemoryArtifact` and the Obsidian graph/indexing calls are synchronous from the scanner's point of view. `startNativeMemoryBridge` also accumulates the complete file list before processing it and builds a path index over that list (`:1003-1025`).

The `syncExisting` API returns a Promise and coalesces concurrent scans (`:1063-1082`), but that is single-flight scheduling, not execution isolation. It prevents duplicate scans while still allowing one large scan to perform thousands of synchronous DB operations on the main thread. The `yieldEvery` helper only inserts a macrotask boundary (`platform/daemon/src/async-yield.ts:1-12`); it cannot interrupt one synchronous query, transaction, graph update, or embedding write.

3. Recall is still built around synchronous DB access.

`platform/daemon/src/memory-search.ts:609-615,714-735,756-800` performs recall hydration, content loading, observed-score loading, and currentness loading through direct synchronous `withReadDb` calls. The accessor itself notes that recall performs multiple reads per request (`platform/daemon/src/db-accessor.ts:1141-1143`). There is no general worker dispatch or query deadline around these calls. A 50-result cap limits response size; it does not bound vector/FTS query time, joins, hydration, or the number of synchronous post-query passes. The acceptance target therefore needs a p95 latency proof under concurrent ingest and maintenance, not just a result limit.

4. The event-loop monitor diagnoses the violation but cannot enforce it.

`platform/daemon/src/resource-monitor.ts:5,105-160` performs synchronous `/proc` directory enumeration and `readlinkSync` calls while taking a resource snapshot. `startFdPollMonitor` invokes that synchronous scan periodically, and daemon startup/heartbeat paths call the snapshot directly. `startEventLoopMonitor` reports lag and pressure (`:180-202`), but it does not reject, cancel, or migrate the operation responsible for the lag. The existing "Event loop blocked" warning is therefore telemetry after the fact, not an architectural guardrail.

5. The integrity implementation proves that worker threads are not a sufficient general answer.

`platform/daemon/src/database-integrity.ts:51-58` executes synchronous `PRAGMA quick_check` and `integrity_check`. Its deferred path explicitly uses a child process because "SQLite's synchronous PRAGMA cannot observe Worker.terminate() while native code is running" (`:131-136`). The regression test intentionally runs an infinite synchronous `PRAGMA quick_check` in a worker and verifies that only the child-process deadline gives a real kill boundary. This is the correct pattern for uninterruptible native work, but it is currently a special-case repair rather than a daemon-wide execution model. The same file still performs synchronous transactional `REINDEX` repair after the check (`:260-320`), so deferring the scan does not make every integrity-related operation non-blocking.

6. Embedding migration still executes large synchronous transactions through the queue abstraction.

`platform/daemon/src/embedding-index-migration.ts:77-79` falls back to synchronous `withWriteTx` when the async method is absent. `:123-143` rebuilds the active vector index by dropping/recreating tables and reinserting the durable projection. `:238-340` performs per-row staging writes, and `:347-385` promotes the staging index inside one write transaction, potentially rebuilding the active index and running vacuum work. These are precisely the operations that need worker ownership, cancellation, and a cost budget; routing them through `withQueuedWrite` only serializes their synchronous execution.

## What the last week proves

The Final Push board and the merged PR burst from 2026-08-10 through 2026-08-13 show a repeated per-site repair pattern rather than an architectural fix:

- PR #1404 removed synchronous filesystem walks and reads, but its own notes preserve a non-paginated, all-at-once result shape. IO was made async; the dataflow remains unbounded.
- PR #1410 had to release traversal read connections before async yields because `withReadDbAsync` could hold the pool across awaits and exhaust the 16-connection limit. That is a resource-lifetime symptom of exposing synchronous connections to asynchronous control flow, not isolation.
- PR #1414 bounds Dreaming writes to 10 operations or 50ms with a macrotask yield. That is useful local mitigation, but it demonstrates that every new writer must independently rediscover a chunk size and yield policy because no common execution budget exists.
- PR #1503 moved legacy VACUUM after readiness, while explicitly acknowledging that it remains a synchronous SQLite VACUUM on the existing database connection and can increase database/event-loop latency. Readiness-first reduces startup damage; it does not remove the main-thread stall.
- PR #1514 moved the integrity scan after readiness and used a killable child process because a synchronous native SQLite call cannot be safely interrupted in a worker thread. This is strong evidence for process/worker ownership plus hard deadlines as a general primitive.
- PR #1525 added workload counters and ages, but explicitly says observability does not replace admission or timeout caps. Diagnostics can show the queue; they cannot prevent the event loop from being monopolized.
- PR #1528 provides current-main live evidence: the daemon stayed alive and `/health/live` had a 1 ms p95 under an overfed ingestion run, but after 378 seconds 158 documents were still queued. Liveness and queue admission are not throughput, bounded drain time, or recall latency.
- PR #1536 corrected the earlier probe attribution: `reachableDaemonProbe` was a synchronous formatter, while overlapping callers amplified health and metadata requests. The repair added single-flight coalescing and changed startup to `/health/live`. This is another example of symptom-level fixes being necessary while the system still lacks a general rule preventing synchronous work and caller amplification on critical paths.

The older incidents and fixes have the same shape: tokenizer encodes (#1114), the dreaming check-loop failure (#1198), traversal hydration (#1250), manifest locking (`Atomics.wait`, #1256), startup integrity (`#1513`/`#1514`), source/artifact sync, and ingestion pressure. The codebase now contains multiple local caps, yields, single-flight guards, and post-ready workers, but no invariant that proves all collection-size-dependent work is outside the event loop.

## Required architecture direction

1. Make heavy work execute in a bounded worker pool or killable child process by default: source sync, artifact ingestion, integrity checks and repairs, index rebuilds, embedding migration, extraction, FTS maintenance, and other O(n) operations. The pool needs bounded queues, backpressure, cancellation, and deadlines.
2. Put SQLite access behind an actual off-main-thread DB execution layer or dedicated DB workers. `withWriteTxAsync` must not mean "call a synchronous callback later"; it must establish where the SQL executes and what its maximum execution budget is.
3. Dispatch vector and FTS recall work away from the event loop and measure p95 latency under concurrent recall, ingest, and maintenance. Result-count caps are not latency caps.
4. Turn the event-loop monitor into an invariant: every synchronous operation over the budget must be attributable and fail/offload, not merely logged after the stall.
5. Keep readiness-first startup, but audit every work item before and after readiness. No startup or post-ready worker may call an unbounded synchronous scan, transaction, native PRAGMA, vacuum, or full-table rebuild on the main connection.
6. Replace all-at-once collection materialization with bounded streaming/chunking. A yield between chunks is useful only after each chunk's synchronous work is itself bounded.

## Acceptance criteria

- Fresh start on a 10M-memory workspace reports readiness in under 5 seconds. No collection-size-dependent synchronous scan, SQLite rebuild, native PRAGMA, vacuum, or full-table backfill may run on the main thread before readiness.
- Recall/search p95 is under 500ms at 10M memories under concurrent ingest and background maintenance, with the benchmark reporting queue age, DB latency, and event-loop lag.
- Zero event-loop-block warnings over 2 seconds in normal operation, and the watchdog can identify the owning operation and enforce the budget rather than only logging it.
- A static audit plus runtime test covers every daemon path listed above and proves that synchronous filesystem, native, and SQLite work is either bounded below the budget or executed in a worker/process with a hard deadline.
- Queues have explicit admission, backpressure, cancellation, and bounded-drain behavior. A live harness must report both liveness and completion/throughput, not only that requests were accepted while work remained queued.

## Evidence chain

#1114, #1198, #1250, #1256, #1266, #1270, #1327, #1328, #1334, #1337, #1338, #1341, #1342, #1348, #1349, #1503, #1513, #1514, #1525, #1528, #1535, #1536, and the direct code paths cited above. The #1327-#1353 and #1404-#1536 repair families are evidence of the same convergence target: isolate execution and enforce budgets centrally instead of continuing to patch each synchronous surface independently.

NicholaiVogel:
## Summary

Phase D stability acceptance harness for #1543: a deterministic, production-shaped scenario that boots the real daemon from source, measures event-loop occupancy from inside the daemon process, and asserts the issue's stability criteria. It is a judge, not a fixer: if it fails on current main, that is the harness working — the numbers are the baseline.

## Changes

- `tests/integration/acceptance/build-db.ts` — deterministic production-shaped DB builder (full scale: 106k memories (93k embedded), 11k transcript jobs with payloads, 25k telemetry events, 5k source index rows; ~1.4 GB). Schema comes from the real `runMigrations` — no hand-written DDL.
- `tests/integration/acceptance/loop-probe.ts` — event-loop occupancy probe injected INTO the daemon process via `bun --preload`. Samples per-second max delay via `perf_hooks.monitorEventLoopDelay` (Bun reports ns; converted to ms). Exposes results on its own loopback listener so they stay readable even when the daemon's HTTP surface wedges.
- `tests/integration/acceptance/run.ts` — scenario runner: builds the DB, simulates #1671's provider-down trigger (embedding provider refuses connections while a synthetic source root keeps source sync walking), drives concurrent pollers (`/health/live` 250ms, `/api/status` 2s, `/api/diagnostics` 5s) plus a foreground write load through the normal remember path, then evaluates and writes a machine-readable JSON artifact. Hermetic child env (HOME isolated). SIGTERM-then-SIGKILL shutdown — a wedged daemon can ignore SIGTERM (observed).
- `tests/integration/acceptance/criteria.ts` — #1543's acceptance criteria codified: zero event-loop blocks >= 2000ms; `/health/live` p95 < 500ms; `/api/status` p95 < 1000ms.
- `tests/integration/acceptance/baseline-main.json` — baseline artifact captured on main (this branch's parent commit).
- `.github/workflows/phase-d-acceptance.yml` — smoke variant on PRs/main touching daemon/core; full variant nightly (07:15 UTC) + manual dispatch.
- `tests/README.md` — docs.

## Baseline on main (full scale, 180s run)

| Metric | Observed | Limit | Result |
|---|---|---|---|
| event-loop blocks >= 2000ms | 0 | 0 | PASS |
| `/health/live` p95 | 93ms (max 294ms, 0/689 failed) | < 500ms | PASS |
| `/api/status` p95 | 265ms (max 299ms, 0/87 failed) | < 1000ms | PASS |
| event-loop delay p50/p95/max | 114 / 268 / 400ms | — | — |
| daemon startup on 106k memories | 2.8s | — | — |

Notes from harness bring-up (not part of the passing baseline, but the class of thing this harness exists to catch): during development runs on main the daemon intermittently exhibited a ~30s event-loop block (daemon's own monitor logged `Event loop critically blocked (30209ms)`), cascading DB-owner deadline errors and a listener outage while the process stayed alive. It did not reproduce in the final baseline run — which is exactly why the nightly full variant exists: 180s windows, every night, with an artifact trail.

## Type

- [ ] `fix` — bug fix
- [x] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [ ] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [x] Other: tests + CI

## PR Readiness (MANDATORY)

N/A — no migrations touched.

## Testing

- [ ] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Verified end to end: smoke run exit 0 (clean pass, clean shutdown), full-scale baseline run exit 0 with the artifact committed. Daemon + core typecheck clean (repo-wide `bun run typecheck` has pre-existing connector-package errors from unbuilt artifacts, unchanged by this PR).

## AI disclosure

Closes #1543

## Readiness checklist notes

Pure test/tooling PR: no production code, no routes, no DB queries in shipped code. Items checked as they apply to the harness itself: agent scoping N/A (no data queries added); input/config validation = harness arg parsing + config writes; error handling = watchdog/timeout paths exercised in bring-up runs (daemon wedge, SIGTERM-ignored, port exhaustion are all handled); security N/A (loopback-only, no auth surfaces touched); docs = tests/README.md section; regression tests N/A (no bug fixed — the harness IS the test); spec alignment = implements #1543's stated criteria verbatim.


NicholaiVogel:
Follow-up verification at head 4db4d805d: smoke acceptance exited 0 after a 90s run. Queue-depth samples now come from GET /api/diagnostics/queue (18 samples; memory=0 and summary=0 throughout; unavailableReason=null), and shutdown is clean (pollersSettled=true, timedOut=false). Measured event-loop p50/p95/max=77/77/77ms with 0 blocks >=2000ms; /health/live p95=1ms (max 51ms); /api/status p95=56ms (max 62ms). Biome and @signet/daemon typecheck pass.

NicholaiVogel:
Merged by orchestrator. Verification chain: harness card (t_d947d85a) independently executed both variants before merge review — smoke exit 0 (18 queue samples, clean poller shutdown), full 180s run exit 0 (36 samples, zero event-loop blocks >=2000ms, /health/live p95 72ms, /api/status p95 248ms), plus biome/typecheck/diff-confine/workflow-inspection all green at head 4db4d805d9. Diff confined to tests/ + workflow + turbo.json + README. Baseline artifact committed (baseline-main.json). The bring-up note about the intermittent ~30s block is exactly the class of event this harness now hunts nightly — first nightly run 07:15 UTC tomorrow.

PR #1689, opened 2026-08-19
fix: bound status and diagnostics projections

### LINKED ISSUE #1670: bug(daemon): synchronous SQLite initialization still wedges the parent event loop after readiness

NicholaiVogel:
## Summary

- add migration 138 with maintained transcript-capture and duplicate-health projections
- switch `/api/status` and duplicate diagnostics reads to bounded projection queries
- add migration compatibility/backfill coverage and a payload-scale regression test for #1670

## Changes

- `@signet/core`: migration 138 creates and backfills the projection tables, indexes, and maintenance triggers
- `@signet/daemon`: `/api/status` transcript status and duplicate diagnostics use bounded read-model queries
- Added red-on-parent/green-on-fix regression coverage with large inline transcript and memory payloads

## Type

- [x] `fix` — bug fix
- [x] `test` — test coverage

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`

## Screenshots

N/A: no UI changes.

## PR Readiness (MANDATORY)

The error/fallback check is covered by migration backfill/idempotence tests and zero-value fallback assertions when a projection row is absent. Security and input/config checks are not applicable to these read-only, internally maintained projections; no admin or mutation endpoint was changed. The API response shape is unchanged and the implementation comments document the status projection contract.

- [x] Migration is idempotent

Migration 138 is append-only and rebuilds its projections from existing source tables on each migration invocation. It adds the legacy `memories.manual_override` column when absent before creating dependent indexes/triggers, supports fresh and upgraded databases, and backfills existing transcript jobs and memories. The projections can be rebuilt from `transcript_capture_jobs` and `memories`; the source tables and API response shapes remain unchanged.

## Testing

- [x] `bun test` passes
- [x] `bun run lint` passes
- [ ] N/A

Verified locally with:

- `bun test platform/core/src/migrations/migrations.test.ts` (69 pass, 0 fail)
- `bun test platform/daemon/src/status-projection.test.ts` (9 pass, 0 fail)
- `bun test platform/daemon/src/transcript-capture-worker.test.ts platform/daemon/src/diagnostics.test.ts platform/daemon/src/daemon-status.test.ts` (46 pass, 12 skipped, 0 fail)
- `bun run --filter "@signet/core" build`
- `bun run --filter "@signet/daemon" build`
- Biome check and `git diff --check`

The scale test inserts 2,500 transcript jobs and 2,500 memory rows with large inline payloads, exercises the direct readers and real `/api/status`, and enforces a 50ms per-call budget. The restored parent implementation fails the payload-table access assertion.

## AI disclosure

## Notes

- No API version or response-shape changes.
- No new synchronous DB access sites were added.
- Fixes #1670

### DIFF
--- a/platform/core/src/migrations/138-bounded-status-projections.ts
+++ b/platform/core/src/migrations/138-bounded-status-projections.ts
@@ -1,0 +1,334 @@
+import type { MigrationDb } from "./index";
+
+function hasColumn(db: MigrationDb, table: string, column: string): boolean {
+	const rows = db.prepare(`PRAGMA table_info(${table})`).all() as ReadonlyArray<{ name?: unknown }>;
+	return rows.some((row) => row.name === column);
+}
+
+function addColumnIfMissing(db: MigrationDb, table: string, column: string, definition: string): void {
+	if (!hasColumn(db, table, column)) db.exec(`ALTER TABLE ${table} ADD COLUMN ${column} ${definition}`);
+}
+
+/**
+ * Migration 138: bounded status projections for /api/status and diagnostics.
+ *
+ * The HTTP-serving isolate previously grouped transcript_capture_jobs and
+ * memories directly to calculate compact health fields. Those payload tables
+ * can be very large. This migration installs incrementally maintained
+ * projections, backfills them once in the migration/owner lane, and adds
+ * covering indexes for the remaining maintenance lookups.
+ */
+export function up(db: MigrationDb): void {
+	// A legitimate v1 database can reach this migration without the column
+	// that the baseline schema normally creates. Make this upgrade path safe
+	// before creating the partial index or triggers below.
+	addColumnIfMissing(db, "memories", "manual_override", "INTEGER DEFAULT 0");
+
+	db.exec(`
+		CREATE TABLE IF NOT EXISTS transcript_capture_status (
+			agent_id TEXT PRIMARY KEY,
+			pending INTEGER NOT NULL DEFAULT 0,
+			processing INTEGER NOT NULL DEFAULT 0,
+			completed INTEGER NOT NULL DEFAULT 0,
+			failed INTEGER NOT NULL DEFAULT 0,
+			dead INTEGER NOT NULL DEFAULT 0,
+			oldest_pending_at TEXT,
+			last_error TEXT,
+			last_error_at TEXT
+		);
+
+		CREATE INDEX IF NOT EXISTS idx_transcript_capture_jobs_agent_status_created_at
+			ON transcript_capture_jobs(agent_id, status, created_at);
+
+		CREATE INDEX IF NOT EXISTS idx_transcript_capture_jobs_error_updated_at
+			ON transcript_capture_jobs(agent_id, updated_at DESC)
+			WHERE error IS NOT NULL;
+
+		CREATE TABLE IF NOT EXISTS memories_duplicate_hash_counts (
+			content_hash TEXT PRIMARY KEY,
+			dup_count INTEGER NOT NULL
+		);
+
+		CREATE INDEX IF NOT EXISTS idx_memories_dup_counts_active
+			ON memories_duplicate_hash_counts(dup_count)
+			WHERE dup_count > 1;
+
+		CREATE TABLE IF NOT EXISTS memories_diagnostics_state (
+			id INTEGER PRIMARY KEY CHECK (id = 1),
+			total_active INTEGER NOT NULL DEFAULT 0,
+			exact_duplicates INTEGER NOT NULL DEFAULT 0,
+			exact_clusters INTEGER NOT NULL DEFAULT 0
+		);
+
+		CREATE INDEX IF NOT EXISTS idx_memories_active_hash_diagnostics
+			ON memories(content_hash)
+			WHERE is_deleted = 0
+			  AND content_hash IS NOT NULL
+			  AND pinned = 0
+			  AND manual_override = 0;
+	`);
+
+	db.exec(`
+		CREATE TRIGGER IF NOT EXISTS transcript_capture_status_ai
+		AFTER INSERT ON transcript_capture_jobs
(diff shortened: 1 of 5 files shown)


NicholaiVogel:
Merged by orchestrator at rebased head 5ab2439bc (13-commit rebase onto main, clean). Verification chain: adversarial review t_c42c88bc SHIP verdict at pre-rebase head 42cf8df456 — projection triggers verified complete (transcript insert/update/delete + memory insert/update/delete), both readers confirmed zero payload-table access, load-bearing test + 2500-row scale regression pass, migration tests pass, LEGACY_SYNC_DB_ACCESS ledger unchanged at 240. Post-rebase: status-projection suite re-run green (9/9), 17/17 gates green at 5ab2439bc. Fixes #1670.

PR #1690, opened 2026-08-19
feat(daemon): prioritize owner foreground workloads

### LINKED ISSUE #1671: bug(native-sources): unbounded source synchronization remains coupled to the daemon event loop

NicholaiVogel:
## Summary

- owner-lane workload classes with bounded admission (#1671): foreground (interactive writes, recall) and maintenance jobs now queue independently; foreground has strict priority plus a bounded maintenance burst so background work still progresses
- native source synchronization routed through the maintenance class, so provider-down retry storms can no longer monopolize the owner lane ahead of user-facing writes
- health surfaces per-class queue depth + oldest-job age metrics
- protocol contract updated

## Evidence

- Phase D harness provider-down scenario (from #1688) is the acceptance target: foreground writes must stay unblocked while maintenance yields
- regression coverage added; LEGACY_SYNC_DB_ACCESS ledger unchanged

Fixes #1671
Refs #1543

## AI disclosure

- [x] AI tools were used (kanban worker t_77f78970, orchestrator-verified)

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`

## PR Readiness (MANDATORY)


NicholaiVogel:
Adversarial review (kanban t_58544e5e, @ exact head 137459d39):

**VERDICT: NEEDS WORK** (1 HIGH, 2 MEDIUM, 1 LOW).

**1. HIGH — provider-down source sync still amplifies semantic work.** native-memory-sources.ts:907-928 submits sources.graph.owner.index as maintenance before obsidian-source-embeddings.ts:285-307 applies the provider circuit gate. When the provider is unavailable, the embedding call returns pending, semanticComplete is false, and native-memory-sources.ts:959 caches the fingerprint anyway. On the next scan, lines 865-869 invalidate that cache because embeddings are absent, then lines 910-928 submit the graph index again before the provider gate. The provider backoff bounds embedding calls, but it does not pause/refrain from repeating graph semantic work. Minimal fix: make provider-down state suppress the source's semantic graph/index submission until the source-level retry window expires, and add an integration regression that indexes the same artifact twice during provider outage and proves the second pass does not submit graph/embedding work.

**2. MEDIUM — per-class owner health metrics are not exposed.** db-owner-client.ts:190-217 computes foregroundQueuedJobs/maintenanceQueuedJobs/oldest-age only in the client object; routes/health.ts:179-228 returns dbWriter/dbReader/dbRuntime and never includes owner health. No consumer exists outside tests. Minimal fix: wire a bounded in-memory owner-health snapshot through the DB accessor into /health or diagnostics, then test the response schema.

**3. MEDIUM — sources.artifact-heal write defaults to foreground.** native-memory-sources.ts:623-633 submits with lane write but no workloadClass; db-owner-client.ts:526-527 therefore assigns foreground. This is a maintenance repair that can jump the queue policy. Minimal fix: workloadClass: maintenance, and audit all source-sync writes for explicit classification.

**4. LOW — health semantics not truthful under two-lane client.** db-owner-client.ts:192-214 counts the active entry as queued and derives oldest age from that set; :646-672 null-coalesces activeJobId/class across lanes that can both be active. Minimal fix: track active separately per lane, exclude active from queue depth/age, merge active state explicitly.

Verified good: dequeue is strict-plus-burst (MAX_FOREGROUND_BURST=8, maintenance cannot starve, foreground head-of-line bounded by active job + one maintenance slice); no cross-class deadlock/ordering violation found; focused suites 68/68 pass; ratchet holds at 240. Reviewer note: the priority test's wait on health().activeJobId is client-side submission state, not worker-start ack — assertions weaker than they appear; strengthen in the repair.

Repair round next; findings 1-3 are required for merge, 4 recommended.

NicholaiVogel:
Repair complete at head c995727ca1a4c32e9175eff195dde90a5dec99e8, rebased onto origin/main 1f81461f5319e1f6cdfea6cdcf1ba54246e61d53.

Finding resolutions:
1. Provider-down amplification: Obsidian embedding/provider failure is now evaluated before graph-owner indexing. A provider outage suppresses both semantic writes for that source while the source retry gate is active. The regression syncs the same artifact twice during outage: fetches=1, graph rows=0, source_chunk embedding rows=0, and both status callbacks remain provider-down.
2. Owner health: the bounded owner snapshot is registered through DbAccessor and exposed as /health.dbOwner, including read/maintenance lane queue depth, oldest ages, active job/class, state, and deadline kills. The response schema is covered by routes/health.test.ts.
3. Source write classification: sources.artifact-heal is explicitly maintenance. Audited source-sync owner writes are:
   - native-memory-sources.ts: sources.artifact-heal, sources.graph.owner.index, sources.graph.owner.file-purge, sources.artifacts.purge, sources.graph.owner.purge
   - memory-lineage.ts: sources.artifacts.upsert, sources.artifacts.soft-delete, sources.artifacts.delete, sources.reindex.artifacts-upsert, sources.reindex.artifacts-delete
   - obsidian-source-embeddings.ts: sources.embeddings.owner.safety, sources.embeddings.owner.write, sources.embeddings.owner.stale.delete, sources.embeddings.owner.purge.write
   Every listed write has lane=write and workloadClass=maintenance.
4. Queue metrics: queuedJobs and per-class queue/age metrics exclude the worker-active job. Aggregate health exposes explicit per-lane snapshots instead of hiding lane state behind null coalescing.
5. Worker-start acknowledgement: the worker emits a started event with job id and workload class; client active state is set from that acknowledgement, and priority tests wait on it.

Verification:
- Focused daemon suite: 106 pass, 0 fail, 379 expect() calls across db-owner-client, native-memory-sources, routes/health, and obsidian-source-embeddings.
- daemon typecheck: passed.
- Biome changed-file check: no errors; two existing warnings remain in db-accessor.ts and routes/health.test.ts.
- Event-loop contract audit: ledger 724; legacy sync DB markers 240 (101 withWriteTx, 139 withReadDb); ratchet passed.
- Phase D smoke from the exact origin/main snapshot: passed. 20,000 memories, 2,000 transcript jobs, 5,000 telemetry events, 1,500 source rows, 17,637 embeddings; startup 2,424ms; 90s run; 120 foreground writes; 0 event-loop blocks over 2,000ms; /health/live 0 failures out of 4,561, p95 1ms, max 34ms; /api/status 0 failures out of 46, p95 3ms, max 4ms; diagnostics p95/max 18ms.
- git diff --check: passed.

NicholaiVogel:
Merged by orchestrator at head 6bad27a6af. Chain: adversarial review NEEDS WORK (5 findings, t_58544e5e) -> repair (t_183fb55f) -> delta re-review SHIP (t_3dae55b4): all findings resolved with load-bearing tests, no rebase regressions, ratchet intact at 240. 18/18 gates green. Fixes #1671.

PR #1698, opened 2026-08-19
fix(daemon): reserve interactive DB-owner capacity

NicholaiVogel:
## Summary

- split the daemon DB-owner client into independent read, interactive-write, and maintenance owner lanes so queue priority is backed by reserved process/connection capacity
- classify `sources.*`, `maintenance.*`, and `vacuum.*` runtime operations authoritatively as maintenance, including callers that request foreground workload class
- change per-job deadline handling to abandon and cancel work without killing the DB owner; supervised shutdown and transport recovery remain the only owner-kill paths
- aggregate per-lane health and retain durable maintenance checkpoint/reschedule behavior

## Repair notes

- [x] Finding 1: `platform/daemon/src/db-owner-worker.ts:82-90,539-541` now records cancellation IDs only for jobs still present in a queue; active, completed, and unknown job IDs are ignored, preventing late-cancel set growth. `platform/daemon/src/db-owner-client.test.ts:603-610` covers 1,000 completed-job cancels plus active and queued cases.
- [x] Finding 2: `platform/daemon/src/db-owner-runtime.ts:95-98` makes the `sources.*`, `maintenance.*`, and `vacuum.*` operation classifier override explicit foreground workload class. `platform/daemon/src/db-owner-client.test.ts:613-642` verifies a `sources.*` foreground override queues on maintenance while that lane is saturated.
- [x] Finding 3: removed the dead always-zero `deadlineKills` field from DB-owner lane/aggregate health and its database-integrity/startup-recovery status and logging surfaces. The existing deadline tests retain the no-kill owner behavior without a misleading metric.

## Evidence

- `bun test platform/daemon/src/db-owner-client.test.ts` - 27 pass, 0 fail
- `bun test platform/daemon/src/db-owner-maintenance.test.ts` - 13 pass, 0 fail
- `bun test platform/daemon/src/incremental-database-integrity.test.ts` - 6 pass, 0 fail
- `bun test platform/daemon/src/database-integrity.test.ts` - 17 pass, 0 fail
- `bun test platform/daemon/src/routes/health.test.ts` - 15 pass, 0 fail
- `bun test platform/daemon/src/native-memory-sources.test.ts` - 57 pass, 0 fail
- `bun test platform/daemon/src/obsidian-source-embeddings.test.ts` - 11 pass, 0 fail
- `bun test platform/daemon/src/daemon-production-wiring.test.ts` - 2 pass, 0 fail
[...]

PR #1700, opened 2026-08-19
fix(daemon): pause source sync on provider outage

### LINKED ISSUE #1696: stability: Wave 2 is not complete, live daemon still wedges
## This wave did not fix the wedges. It renamed, bounded, and measured them.

The Stability Roadmap says:

> WAVE 2 COMPLETE — all four wedges dead.

That claim is furious-makingly wrong.

I installed published Signet version `0.213.1`, the release containing the entire claimed stability wave, and ran the actual daemon against the real workspace workload. It still became unusable almost immediately after the first healthy response.

Observed live failure:

- At approximately t+125s, `/health` and `/` returned HTTP 200.
- The daemon reported version `0.213.1`, approximately 13% CPU, approximately 554 MB RSS, and 3 ms event-loop lag.
- Integrity maintenance was stuck at 378 checked objects with 42 remaining.
- The daemon had already recorded 29 DB-owner deadline kills.
- Twenty seconds later, both `/health` and `/` timed out while the listener remained open.
- GDB caught the parent daemon thread in native `pread64`.
- A 15-second trace captured 806,442 parent-side SQLite reads.
- The logs then showed native-source synchronization killing the DB owner, followed by FTS fallback churn and provider-unavailable work.

This was not a stale watcher report. It was not a false positive. It was not a deliberately stopped process. This was the newly installed release failing its basic job in a live run.

### #1689: bounded status and diagnostics projections

This may have reduced the original `/api/status` payload scan, but it did not fix daemon-wide responsiveness. The root page and `/health` still died while the listener stayed open. Removing one known projection scan is not the same thing as removing unbounded synchronous SQLite from the parent isolate.

### #1690: owner-lane classes and prioritization

The new lane classes did not stop the owner from being overwhelmed. The daemon recorded 29 owner deadline kills before the next probe failed. Native-source synchronization still reached the point where it killed the DB owner. A queue priority is not isolation, and demoting work is not a guarantee that the work cannot monopolize the system.

### #1691: incremental quick_check

This did not complete the integrity operation. It repeatedly oscillated between `running` and `timed_out`, remaining at 378 checked objects and 42 remaining. The slices spent roughly 700–950 ms waiting for owner admission and only 1–3 ms executing. That is not a completed integrity design. It is a large operation being chopped into repeatedly failing attempts while the rest of the system remains under pressure.

The new counters even reported approximately 765 million pages and 3.13 TB of observed database bytes while the checkpoint still had 42 objects remaining. Whether those counters are cumulative or misnamed, they expose the same problem: the system is doing enormous repeated work without reaching a stable completion state.

### #1693: truthful event-loop health

This tells us the daemon is wedged. It does not stop the daemon from wedging. A more honest dashboard for a dead service is useful instrumentation, not a stability fix. In this run, the event-loop monitor correctly observed a healthy window and then the HTTP server still became unreachable.

### #1688: Phase D stability harness

The green baseline did not catch this failure. The real daemon failed against the real database/source workload within the verification window. A harness that passes while the installed release still produces 806,442 parent-side SQLite reads and loses HTTP service is not testing the failure that matters.

### #1687: sync-access enforcement audit

A static count and CI ratchet did not prove runtime ownership safety. The shipped daemon still executed massive synchronous native SQLite work in the parent process. A ledger saying fewer call sites exist is not evidence that the remaining call sites are safe, bounded, or incapable of starving HTTP.

## The actual conclusion

The wave did not kill the wedges. It addressed individual symptoms and added instrumentation around an architecture that still permits:

1. parent-isolate synchronous SQLite storms;
2. source synchronization to monopolize the serialized DB-owner lane;
3. maintenance to repeatedly re-enter after timing out;
4. provider failure to amplify source work and fallback churn; and
5. the listener to remain open while every useful HTTP endpoint becomes unavailable.

The live result is not “one edge case remains.” The core invariant still does not hold:

> No unbounded SQLite or source synchronization may be able to block the HTTP/event-loop process or monopolize interactive database service.

That invariant is still false.

## Required action

Withdraw the “all four wedges dead” claim. Do not call another release stable because it passes the first health check or because CI is green.

The permanent fix must establish real ownership isolation, not another layer of timing controls:

- all potentially unbounded SQLite/source work behind a killable worker/process boundary;
- bounded, cancellable, resumable batches with durable progress;
- admission control that guarantees capacity for health, dashboard, recall, and writes;
- source-level pause semantics when embeddings are unavailable;
- no parent-side `pread64` storm under the real source workload;
- no owner-death cascade or deadline-kill storm;
- sustained live HTTP service beyond the known post-ready failure window;
- restart recovery that resumes progress instead of replaying the same full scan.

Do not close this with a longer timeout, another retry/backoff adjustment, an extra event-loop yield, a new health flag, a changed metric, or a roadmap edit. Those make the failure slower, quieter, or better documented. They do not fix it.

Until the installed release passes the live gate against the actual workload, “WAVE 2 COMPLETE — all four wedges dead” is not a status update. It is a false claim that the evidence directly contradicts.

Related: #1543, #1670, #1671, #1683, #1688, #1690, #1691, #1693.

NicholaiVogel:
## Summary

Provider-down native source synchronization now pauses durably at the source boundary and resumes from a persisted per-source checkpoint after recovery. Paused scans are reported as partial rather than successful, and manual resyncs join an in-flight polling scan for the same agent/source instead of duplicating provider and DB-owner work.

Closes #1696

## Changes

- Add migration 139 and owner-routed helpers for per-agent, per-source sync state.
- Gate Obsidian source scans through the existing embedding provider circuit before filesystem work.
- Persist `paused` / `running` state, pause reason, and a normalized file frontier.
- Resume after recovery by skipping files at or before the durable checkpoint.
- Return per-source sync status, scanned/indexed counts, and resume frontier from the native bridge.
- Mark manual and startup source jobs `paused` / `partial` on provider outage, skip `lastIndexedAt` freshness stamping, and retain searchable partial results when present.
- Serialize manual and polling bridges through a module-level agent/source flight registry.
- Add provider-down, provider-down-before-first-file, restart-during-pause, recovery-from-checkpoint, startup partial-status, and cross-instance single-flight regression coverage.
- Document the paused/partial `GET /api/sources` index-job status, counts, pause reason, resume frontier, and freshness behavior.
- Keep the upstream migration 138 (`bounded-status-projections`) intact after rebasing onto current `origin/main`.

## Type

- [x] `fix` — bug fix

## Packages affected

- [x] `@signet/core`
- [x] `@signet/daemon`

## Screenshots

N/A: no UI changes.

## PR Readiness (MANDATORY)

  - Validated against `docs/specs/INDEX.md` and `docs/specs/dependencies.yaml`; this internal source-sync reliability fix adds no new public spec.
- [x] Agent scoping verified on all new/changed data queries.
- [x] Input/config validation and bounds checks added.
  - Uses the existing resolved embedding configuration, normalized agent/source flight key, and bounded provider circuit timing; no new unbounded input surface is introduced.
- [x] Error handling and fallback paths tested (no silent swallow).
  - Provider-down exits before filesystem scanning, persists the pause frontier, and does not enter legacy fallback work.
- [x] Security checks applied to admin/mutation endpoints.
  - N/A: no endpoint or authorization change.
- [x] Docs updated for API/spec/status changes.
  - Updated `web/docs/src/content/docs/api/documents-sources.md` with the paused/partial `indexJob` response fields, zero-count behavior before the first file, resume frontier, and the `lastIndexedAt` freshness rule.
- [x] Regression tests added for each bug fix.
- [x] Lint/typecheck/tests pass locally.
  - Daemon typecheck/build, core build, focused source/migration suites, production-wiring regression coverage, and changed-file Biome checks pass. Root typecheck still reports unrelated existing connector-package failures; full root `bun test` had the pre-existing 420-second timeout.

## Migration Notes

- [x] Migration is idempotent.
- [x] Rollback / compatibility note included in PR description.
  - Migration 139 appends `native_source_sync_state`; existing migration 138 remains unchanged. Older databases create the table on normal migration, and repeated migration runs are safe. No rollback migration is required; removing the feature requires dropping only this derived control-plane table after source work is stopped.

## Testing

- [x] `bun test platform/daemon/src/native-memory-sources.test.ts` — 59 pass, 0 fail, 220 expect calls.
- [x] `bun test platform/daemon/src/routes/sources-routes.test.ts` — 46 pass, 0 fail, 293 expect calls, including provider-down-before-first-file counts.
- [x] `bun test platform/daemon/src/daemon-production-wiring.test.ts platform/daemon/src/source-index-progress.test.ts` — 7 pass, 0 fail.
- [x] `bun test platform/core/src/migrations/migrations.test.ts` — 69 pass, 0 fail, 649 expect calls.
- [x] `bun run --filter '@signet/core' build` — pass.
- [x] `bun run --filter '@signet/daemon' build` — pass, including daemon `tsc --noEmit` and dashboard build.
- [x] `bun run --filter '@signet/daemon' typecheck` — pass.
- [x] `bunx biome check --write` on all changed files — pass.
- [ ] `bun test` — pre-existing root-suite timeout after 420 seconds; focused affected-suite coverage passes.
- [ ] `bun run typecheck` — current root run fails in unrelated connector packages (`connector-oh-my-pi`, `connector-gemini`, `connector-forge`, `connector-opencode`, `connector-pi`, `connector-openclaw`, `connector-codex`); core and daemon typechecks pass.
- [x] `bun run lint` — exits 0; reports existing repository warnings outside this diff.
- [ ] Tested against running daemon — not run; bridge/database boundary coverage passed and no live daemon instance was available in the bounded verification window.

## AI disclosure

- [x] AI tools were used (see `Assisted-by: Hermes-Agent:gpt-5.6-luna` commit tag)

## Per-finding resolution

- Finding 1, partial pause success lie: `platform/daemon/src/routes/sources-routes.ts:657-677` now consumes the bridge result, pauses the job with the bridge's scanned/indexed counts and frontier, records `partial`, disables freshness updates, and returns before `markSourceIndexed` at line 679. Startup uses the same count-aware treatment at `platform/daemon/src/daemon.ts:2597-2632`; the production-wiring regression is in `platform/daemon/src/daemon-production-wiring.test.ts`, and route coverage includes both partial progress and provider-down-before-first-file with zero counts.
- Finding 2, cross-instance single-flight: `platform/daemon/src/native-memory-sources.ts:148,523,1196-1215,1409` uses a module-level agent/source/root flight registry; a manual bridge joins the polling promise. The regression at `platform/daemon/src/native-memory-sources.test.ts:1276-1327` observes two provider probes and one embedding call total, with both bridge calls returning the same indexed count.

## Notes

The source-pause state machine remains in `platform/daemon/src/native-memory-sources.ts`: provider gating occurs before `pathExists` / source walking; provider-down status from embedding indexing persists `paused` and stops the current scan; successful completion clears the frontier. The durable state schema and owner-routed accessors are in `platform/core/src/migrations/139-native-source-sync-state.ts` and `platform/daemon/src/native-source-sync-state.ts`.

### DIFF
--- a/platform/core/src/migrations/139-native-source-sync-state.ts
+++ b/platform/core/src/migrations/139-native-source-sync-state.ts
@@ -1,0 +1,19 @@
+import type { MigrationDb } from "./index";
+
+/** Durable source-level pause and frontier for provider-gated native sync. */
+export function up(db: MigrationDb): void {
+	db.exec(`
+		CREATE TABLE IF NOT EXISTS native_source_sync_state (
+			agent_id TEXT NOT NULL,
+			source_key TEXT NOT NULL,
+			source_root TEXT NOT NULL,
+			status TEXT NOT NULL CHECK(status IN ('running', 'paused')),
+			checkpoint_path TEXT,
+			pause_reason TEXT,
+			updated_at TEXT NOT NULL,
+			PRIMARY KEY (agent_id, source_key)
+		);
+		CREATE INDEX IF NOT EXISTS idx_native_source_sync_state_status
+			ON native_source_sync_state(agent_id, status);
+	`);
+}
--- a/platform/core/src/migrations/index.ts
+++ b/platform/core/src/migrations/index.ts
@@ -144,6 +144,7 @@ import { up as memoryHeadPublication } from "./135-memory-head-publication";
 import { up as memoryHeadRevisions } from "./136-memory-head-revisions";
 import { up as dreamingHeadManifest } from "./137-dreaming-head-manifest";
 import { up as boundedStatusProjections } from "./138-bounded-status-projections";
+import { up as nativeSourceSyncState } from "./139-native-source-sync-state";
 
 // -- Public interface consumed by Database.init() --
 
@@ -1274,6 +1275,12 @@ export const MIGRATIONS: readonly Migration[] = [
 		artifacts: {
 			tables: ["transcript_capture_status", "memories_duplicate_hash_counts", "memories_diagnostics_state"],
 		},
+	},
+	{
+		version: 139,
+		name: "native-source-sync-state",
+		up: nativeSourceSyncState,
+		artifacts: { tables: ["native_source_sync_state"] },
 	},
 ];
 
--- a/platform/daemon/src/daemon-production-wiring.test.ts
+++ b/platform/daemon/src/daemon-production-wiring.test.ts
@@ -36,6 +36,8 @@ describe("daemon production DB owner wiring", () => {
 
 		expect(source).toContain("const syncResult = nativeMemoryBridge?.getLastSyncResult?.();");
 		expect(source).toContain("pauseSourceIndexJob(sourceId, jobId, {");
+		expect(source).toContain("scanned: paused.scanned,");
+		expect(source).toContain("indexed: paused.indexed,");
 		expect(source).toContain('outcome: syncResult?.status === "paused" && paused ? "partial" : "success",');
 		expect(source).toContain('updateFreshness: syncResult?.status === "paused" && paused ? false : undefined,');
 	});
--- a/platform/daemon/src/daemon-production-wiring.test.ts
+++ b/platform/daemon/src/daemon-production-wiring.test.ts
@@ -30,4 +30,13 @@ describe("daemon production DB owner wiring", () => {
 		expect(source).toContain("INCREMENTAL_INTEGRITY_TABLES_PER_RUN");
 		expect(source).not.toContain("runDeferredIntegrityCheck");
 	});
+
+	it("keeps a paused startup source partial instead of reporting success", async () => {
+		const source = await Bun.file(daemonSourceUrl).text();
+
+		expect(source).toContain("const syncResult = nativeMemoryBridge?.getLastSyncResult?.();");
+		expect(source).toContain("pauseSourceIndexJob(sourceId, jobId, {");
+		expect(source).toContain('outcome: syncResult?.status === "paused" && paused ? "partial" : "success",');
+		expect(source).toContain('updateFreshness: syncResult?.status === "paused" && paused ? false : undefined,');
+	});
 });
--- a/platform/daemon/src/daemon.ts
+++ b/platform/daemon/src/daemon.ts
@@ -178,6 +178,7 @@ import {
 	getSourceIndexJob,
 	markSourceIndexInFlight,
 	markSourceIndexJobRunning,
+	pauseSourceIndexJob,
 	updateSourceIndexJobProgress,
 } from "./source-index-progress";
 import {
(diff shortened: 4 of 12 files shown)


NicholaiVogel:
Merged by orchestrator at c0ac5d45c (merge 5c5a33b07). Chain: build (9436c445) -> adversarial review 2 P1s (partial-pause reported as success + lastIndexedAt stamped; no cross-instance single-flight) -> repair (4bf423300) -> hung-runner gate incident -> rebase onto post-#1698 main (c0ac5d45c) -> delta round 2 SHIP. Wave 3 W3.3 complete: durable pause state (migration 139), honest paused/partial reporting, cross-instance single-flight, provider-down = zero new source work. Refs #1696.

PR #1704, opened 2026-08-20
fix(daemon): identify and amputate parent sync DB work

NicholaiVogel:
## Repair follow-up: lazy sync attribution

- Chosen approach: option (b). Every synchronous DB call records only a timestamp and bounded interval token; `Error.stack` construction/parsing runs only when the completed call is slow (>=50 ms). The accessor wrapper binds each marked call's static file:line token at entry, so the latch can identify calls that are still in flight without stack capture.
- Benchmark: 1,000,000 calls per trial, 20,000-call warmup, 7 trials, median on Bun 1.4.0. Exact PR-head helper: 3.524 us/call. Tokenized repaired helper: 0.098 us/call, with the cached entry-token map lookup included. This is 0.80x the 0.122 us/call repaired baseline and below the 1.3x normal-path bound.
- Local verification: attribution and ledger suites pass (25 tests, 76 expectations); accessor/observability/resource/telemetry suites pass (103 tests, 286 expectations, including the new in-flight accessor regression); daemon production build (including `tsc --noEmit` and dashboard bundle) passes; root `bun run typecheck` passes; touched-file Biome check passes with only pre-existing warnings.
- Full workspace test wrapper was not completed locally: its repeated OpenCode minification step spun CPU-bound for >10 minutes without changing the output, so it was terminated; direct `bun test platform/daemon` exceeded the runner 420-second cap without returning a result. Hosted checks on this head remain the final full-suite gate.

## Summary

Identify transitional parent-isolate synchronous SQLite calls with always-on file:line attribution, surface in-flight sites when the event-loop wedge latch fires, and move the cold-start telemetry install identity write behind DB-owner admission.

## Changes

- Added bounded synchronous DB call attribution with in-flight interval history, per-site counters, explicit `path:line` entry tokens, and `kind@unattributed` fallback accounting.
- Wrapped parent `withReadDb` and `withWriteTx` calls; event-loop latches now retain the overlapping site ids and resource monitoring logs each latch once.
- Added an accessor-level load-bearing regression, `DbAccessor > attributes an in-flight parent sync call at latch time`: a real `withWriteTx` wrapper holds its callback in flight, fires the event-loop latch before completion, and asserts the exact `db-accessor.test.ts:201` site; the completed-slow test remains. Added a coverage regression test for missing tokens.
- Audited the baseline transitional inventory: 198 call sites (120 read and 78 write). The only site observed during the cold-start repro was `platform/daemon/src/telemetry.ts:492` (`getOrCreateInstallId`). Its normal daemon path now uses `withWriteTxAsync`; the remaining sync paths are compatibility-only fallbacks when async admission is unavailable.
- Preserved telemetry lifecycle ordering by buffering events until the owner-admitted install identity resolves.

## Type

- [x] `fix` — bug fix
- [ ] `test` — test coverage

## Packages affected

- [x] `@signet/daemon`
- [ ] `@signet/core`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other: none

## Screenshots

N/A: no UI changes.

## PR Readiness (MANDATORY)

- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) (N/A: those files are absent in this checkout; no spec/API surface changed)
- [x] Security checks applied to admin/mutation endpoints (N/A: no endpoint changes)
- [x] Docs updated for API/spec/status changes (N/A: no API/spec/status change)
- [x] Lint/typecheck/tests pass locally for the affected daemon package

N/A: no schema migration.

## Testing

- [x] `bun test` passes for the affected daemon regression suites
- [x] `bun run typecheck` passes for the affected daemon package as part of its build
- [x] `bun run lint` passes for the changed daemon sources
- [ ] N/A

Verification:

- `bun run --filter '@signet/daemon' build` passed, including daemon TypeScript checking and dashboard build; `bun run typecheck` passed across the workspace after the documented dependency build bootstrap.
- `bun test platform/daemon/src/sync-db-attribution.test.ts scripts/audit-event-loop-contract.test.ts` passed: 25 tests, 76 expectations; `bun test platform/daemon/src/db-accessor.test.ts platform/daemon/src/db-observability.test.ts platform/daemon/src/resource-monitor.test.ts platform/daemon/src/telemetry.test.ts` passed: 103 tests, 286 expectations; `bun scripts/audit-event-loop-contract.ts` passed with 198 marked sites (120 reads, 78 writes) and zero token violations.
- Mutation check removed attribution from the regression path and produced the expected test failure; the original attribution was restored before commit.
- Fresh-copy run against `/mnt/work/hermes-scratch/w34-repro-agents-3` ran for the requested 180-second observation window. At 12.9 seconds health reported `syncDb.calls=0`, `unattributedCalls=0`, `unattributedDurationMs=0`, `unattributedSlowDurationMs=0`, and `latchId=0`. The run exited only at the explicit timeout; logs contained zero `Event-loop stall latched` records.
- The changed daemon sources pass the repository formatter/linter check; the full root aggregate test/typecheck commands were not run.

## AI disclosure

## Notes

The pre-conversion diagnostic run identified `withWriteTx@dist/daemon.js:35175`, mapped to `platform/daemon/src/telemetry.ts:492`. The post-conversion run emitted no parent sync attribution calls during the observed cold-start window. Deferred integrity/watcher work remained owner-admitted and bounded; no parent event-loop wedge was latched.

### DIFF
--- a/platform/daemon/legacy-sync/db-accessor-sync.ts
+++ b/platform/daemon/legacy-sync/db-accessor-sync.ts
@@ -25,13 +25,13 @@
  */
 
 import { getDbAccessor } from "../src/db-accessor";
-import type { ReadDb, WriteDb } from "../src/db-accessor";
+import type { ReadDb, SyncDbCallSiteToken, WriteDb } from "../src/db-accessor";
 
 export interface SyncDbAccessor {
 	/** @deprecated Use `withWriteTxAsync` in production code. */
-	withWriteTx<T>(fn: (db: WriteDb) => T): T;
+	withWriteTx<T>(fn: (db: WriteDb) => T, siteToken?: SyncDbCallSiteToken): T;
 	/** @deprecated Use `withReadDbAsync` in production code. */
-	withReadDb<T>(fn: (db: ReadDb) => T): T;
+	withReadDb<T>(fn: (db: ReadDb) => T, siteToken?: SyncDbCallSiteToken): T;
 	/** @deprecated Use `checkpointWalAsync` in production code. */
 	checkpointWal(): void;
 	/** @deprecated Use `incrementalVacuumAsync` in production code. */
--- a/platform/daemon/src/agent-id.ts
+++ b/platform/daemon/src/agent-id.ts
@@ -65,7 +65,7 @@ export function getAgentScope(agentId: string): AgentScope {
 			const readPolicy = parseReadPolicy("read_policy" in row ? row.read_policy : undefined);
 			const policyGroup = parseScopeValue("policy_group" in row ? row.policy_group : undefined);
 			return { readPolicy, policyGroup };
-		});
+		}, "agent-id.ts:56");
 	} catch {
 		return {
 			readPolicy: "isolated",
@@ -85,7 +85,7 @@ export function ensureAgentRegistered(agentId: string, readPolicy: AgentRosterRe
 				 VALUES (?, ?, ?, NULL, ?, ?)
 				 ON CONFLICT(id) DO UPDATE SET updated_at = excluded.updated_at`,
 			).run(id, id, readPolicy, now, now);
-		});
+		}, "agent-id.ts:82");
 	} catch (err) {
 		console.warn(
 			`[agent-id] Failed to register agent "${id}" (non-fatal):`,
--- a/platform/daemon/src/auth/api-keys.ts
+++ b/platform/daemon/src/auth/api-keys.ts
@@ -215,7 +215,7 @@ export function createApiKey(accessor: DbAccessor, input: ApiKeyCreateInput): Cr
 			now,
 			expiresAt,
 		);
-	});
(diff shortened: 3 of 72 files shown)


NicholaiVogel:
Merged by orchestrator at b3c0520e5. Review chain: build -> adversarial review (per-call stack capture 2.9x overhead) -> repair (lazy slow-path capture) -> delta (in-flight hole: kind@unknown:0) -> repair r2 (entry tokens + audit coverage rule) -> delta r2 (test not through real accessor) -> real accessor/latch test -> micro-delta. Load-bearing verified by orchestrator against the bug: test applied to 3592897eb FAILS (in-flight call latches unattributed), passes at head. Rebase onto main replayed cleanly (reviewer-simulated + orchestrator-verified locally at c3296663d, suites green). Chain refs #1696.

PR #1723, opened 2026-08-27
fix(onepassword): handle SDK v0.3.x vault and field shapes

### LINKED ISSUE #1722: bug(secrets): 1Password adapter parses SDK v0.3.x shapes incorrectly — vaults always empty, imports never run
## Summary

The 1Password service account integration connects successfully but reports **0 accessible vaults**, so `signet secret onepassword import` scans nothing and imports no secrets — even though the service account token has valid access to one or more vaults.

Verified on signet **v0.157.3**; the same faulty code is still present in the latest release (**v0.214.14**).

## Steps to Reproduce

1. `signet secret onepassword connect <service-account-token>` — succeeds, but prints `Connected to 1Password (0 vaults accessible)`.
2. `signet secret onepassword status` → `Connected to 1Password` / `Accessible vaults: 0`.
3. `signet secret onepassword vaults` → `No vaults available.`
4. `signet secret onepassword import` → `vaultsScanned: 0, itemsScanned: 0, importedCount: 0`.
5. Using the same token directly with `@1password/sdk` lists the vault(s) fine (proof the token/account/vault access are not the problem).

## Root Cause

`platform/daemon/src/onepassword.ts` — the adapter's `listVaults()` filters vaults using a property the SDK v0.3.x does not return:

```ts
const id = readString(obj.id);
const name = readString(obj.name);   // @1password/sdk v0.3.x vaults expose `title`, not `name`
if (!id || !name) continue;          // every vault is dropped -> vaultCount 0
```

The installed dependency is `"@1password/sdk": "^0.3.0"`, and the v0.3.x `Vault` object is shaped `{ id, title, createdAt, updatedAt }` — `name` is `undefined`, so `readString(obj.name)` returns `null` and every vault is skipped. `vaultCount` is therefore always 0, and the import loop never visits any vault.

### Secondary issue (would surface once vault discovery is fixed)

`getItem()` also parses fields using property names the v0.3.x SDK does not return:

```ts
const type = readString(field.type) ?? "";      // SDK v0.3.x uses `fieldType` (e.g. "Concealed")
const purpose = readString(field.purpose) ?? ""; // SDK v0.3.x item fields have no `purpose`
```

`fieldScore()` then can never match `type === "CONCEALED"` (the +3 score) and can never match `purpose === "PASSWORD"` (the +4 score), so concealed credential fields are only imported when their label happens to match the keyword heuristic. That silently degrades import coverage for concealed fields with generic labels.

## Expected Behavior

- `signet secret onepassword status` / `vaults` should list the vaults the service account token can access.
- `signet secret onepassword import` should scan those vaults and import password/concealed fields.

## Suggested Fix

In `platform/daemon/src/onepassword.ts`:

```ts
// vault listing
const name = readString(obj.name) ?? readString(obj.title) ?? id;

// item fields
const type = readString(field.fieldType) ?? readString(field.type) ?? "";
const purpose = readString(field.purpose) ?? "";
```

Reading `title` as a fallback for `name` and `fieldType` as a fallback for `type` keeps compatibility across SDK versions.

## Environment

- OS: Debian 13 (trixie), amd64
- signet: v0.157.3 (bug also present in v0.214.14)
- `@1password/sdk`: ^0.3.0 (installed 0.3.1)
- 1Password CLI (`op`) 2.38.1 lists the same vaults correctly with the same token

arnavgoel17:
Fixes #1722

## Summary

The 1Password adapter parses SDK v0.3.x shapes incorrectly — vaults always empty, imports never run.

## Changes

- Vault naming falls back from `name` to `title` (v0.3 exposes title) — `onepassword.ts:296`.
- Field type falls back from `type` to `fieldType` (v0.3 renamed) — `onepassword.ts:324`.
- Follow-up 737e1517: hermetic regression test at the raw SDK boundary — mocks `@1password/sdk` v0.3 shapes (`{id, title}` vaults, `{fieldType}` fields), exercises `importOnePasswordSecrets` through the default adapter, pins the original outage. Mutation-verified: against the pre-fix adapter the test fails (importedCount 1 -> 0).

## Type

- [ ] `feat`
- [x] `fix` — bug fix
- [ ] `refactor`
- [ ] `chore`
- [ ] `perf`
- [ ] `test`

## Packages affected

- [ ] `@signet/core`
- [x] `@signet/daemon`
- [ ] `@signet/sdk`
- [ ] `@signet/connector-*`
- [ ] `@signet/web`
- [ ] `predictor`
- [ ] Other:

## Screenshots

None — no UI changes.

## PR Readiness (MANDATORY)

## Verification

- onepassword suite 4/4 at head (13/13 focused across secrets routes)
- Mutation gate: pre-fix adapter fails the new test (imported 1 -> 0)
- daemon typecheck + build green

## Migration Notes

No schema migrations.

### DIFF
--- a/platform/daemon/src/onepassword.test.ts
+++ b/platform/daemon/src/onepassword.test.ts
@@ -1,4 +1,32 @@
-import { describe, expect, it } from "bun:test";
+import { describe, expect, it, mock } from "bun:test";
+
+// Exercise the adapter against the raw SDK v0.3 response shapes rather than
+// an already-normalized OnePasswordClient supplied through clientFactory.
+mock.module("@1password/sdk", () => ({
+	createClient: async () => ({
+		secrets: { resolve: async () => "resolved-secret" },
+		vaults: {
+			list: async () => [{ id: "v1", title: "Engineering Vault" }],
+		},
+		items: {
+			list: async () => [{ id: "item-1", title: "GitHub", vaultId: "v1" }],
+			get: async () => ({
+				id: "item-1",
+				title: "GitHub",
+				fields: [
+					{
+						id: "credential",
+						label: "credential",
+						value: "ghp_v03",
+						fieldType: "CONCEALED",
+						purpose: "",
+					},
+				],
+			}),
+		},
+	}),
+}));
+
 import {
 	type OnePasswordClientFactory,
 	type OnePasswordField,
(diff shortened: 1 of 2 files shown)


NicholaiVogel (dismissed):
Reviewed at head 4bf7afa0025fb00912987e0ff8fe50c2b5a4b153 against base cc502bab4318638666531af4501927b645c4f826.

The official SDK v0.3.1 shapes use VaultOverview.title and ItemField.fieldType. This patch maps both correctly and keeps the older property fallbacks. The focused OnePassword tests pass 3/3. No hosted checks are reported.

NicholaiVogel:
Added a hermetic raw-SDK regression test in `platform/daemon/src/onepassword.test.ts`. It mocks `@1password/sdk` v0.3 shapes (`vaults[].title` and `fields[].fieldType`) and verifies the real `importOnePasswordSecrets` path scans the vault and imports the concealed field. Verified: focused daemon tests 14/14, daemon typecheck, daemon build, Biome check, and `git diff --check`; mutation run failed as expected when both fallbacks were removed, then the fix was restored.

PR #1725, opened 2026-08-27
fix(daemon): route vector backfill through db owner

NicholaiVogel:
## Summary
- Route post-ready vector embedding backfill through the DB-owner maintenance lane instead of running synchronous SQLite writes in the daemon isolate.
- Keep each owner operation bounded to one batch and preserve the existing deadline reserve so health and ready work can run between slices.
- Add a 1,201-row regression fixture with one malformed vector, four terminal iterations, and exact vector/quarantine counts.

## Diagnosis
The isolated inspector reproduction showed the allocation/GC wedge in backfillVecEmbeddings, reached through daemon.ts:2608 and the parent write queue. Concurrent native-source indexing then amplified SQLite lock contention and owner deadline failures. Bun 1.4.0 does not expose the HeapProfiler CDP domain, so the CPU profile was the available runtime allocation evidence.

## Verification
- bun test src/db-owner-client.test.ts --timeout 30000 (37 pass)
- bun test src/db-owner-client.test.ts -t "runs vector backfill" --timeout 30000 (1 pass)
- bun test src/db-accessor.test.ts -t "vec_embeddings schema repair" --timeout 30000 (8 pass)
- bun test src/native-memory-sources.test.ts --timeout 30000 (66 pass)
- bunx biome check on all four changed files
- bun run --filter @signet/daemon typecheck
- bun run --filter @signet/daemon build
- bun scripts/audit-event-loop-contract.ts (pass)
- Mutation run with the owner vector-backfill branch removed: the new regression test failed at the owner result assertion; fixed code then passed.

## Readiness checklist
- [x] Spec alignment validated (`INDEX.md` + `dependencies.yaml`) (internal maintenance routing; no spec or dependency changes)
- [x] Security checks applied to admin/mutation endpoints (not applicable; no endpoint changes)
- [x] Docs updated for API/spec/status changes (not applicable; internal daemon routing only)

## Scope
No live daemon, user database, or real source roots were modified. The isolated reproduction used a copied database and read-only source roots.

### DIFF
--- a/docs/event-loop-contract-audit.md
+++ b/docs/event-loop-contract-audit.md
@@ -4,16 +4,16 @@ This report is generated from the deterministic migration ledger in `scripts/eve
 
 ## Current inventory
 
-- Exact ledger inventory: 998 sites
+- Exact ledger inventory: 997 sites
 - Synchronous `withWriteTx()` sites: 66
 - Synchronous `withReadDb()` sites: 109
-- Async-named parent DB sites: 307
+- Async-named parent DB sites: 306
 - Synchronous filesystem/process sites: 516
 - Compile-visible legacy DB sites remaining: 175
   - `withWriteTx`: 66
   - `withReadDb`: 109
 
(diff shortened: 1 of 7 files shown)


NicholaiVogel:
Closing in favor of a same-repo branch with the identical commits rebased on current main (ledger/test assertions realigned). History preserved; superseder PR opens immediately.

